PR #161, opened 2018-01-04
Fixed dependencies and added them to the top level stack.yaml

tiago-loureiro:
Some dependencies were not building so they were added to the top level stack.yaml too to make it easier to check that all dependencies build correctly

### DIFF
--- a/libs/api-bot/src/Network/Wire/Bot/Monad.hs
+++ b/libs/api-bot/src/Network/Wire/Bot/Monad.hs
@@ -858,6 +858,7 @@ randUser (Email loc dom) (BotTag tag) = do
         , newUserPict           = Nothing
         , newUserAssets         = []
         , newUserAccentId       = Nothing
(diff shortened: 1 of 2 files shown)


libs/api-bot/src/Network/Wire/Bot/Monad.hs
```diff
@@ -858,6 +858,7 @@ randUser (Email loc dom) (BotTag tag) = do
         , newUserPict           = Nothing
         , newUserAssets         = []
         , newUserAccentId       = Nothing
+        , newUserEmailCode      = Nothing
```
elverkilde: where's this coming from?

tiago-loureiro: This new parameter was introduced [back here](https://github.com/wireapp/wire-server/commit/03aa0c7d05ce791c8909f78a62e2b14626351e6e)

PR #162, opened 2018-01-04
Feature/integration test keyspace from config

tiago-loureiro:
Read used keyspace from environment variable/config file to ensure that the integration tests use the same DB as the service itself. Also removed some non-exhaustive pattern matches from one of the tests

### DIFF
--- a/services/gundeck/test/integration/API.hs
+++ b/services/gundeck/test/integration/API.hs
@@ -250,7 +250,7 @@ sendMultipleUsers gu ca _ _ = do
     ntfs3 <- listNotifications uid3 Nothing gu
     liftIO $ do
         assertBool "Not at least 2 notifications" (length ntfs3 >= 2)
-        let (n1:nx) = ntfs3
+        let (n1,nx) = checkNotifications ntfs3
         -- The first notification must be the test payload
         let p1 = view queuedNotificationPayload n1
         assertEqual "Wrong events in 1st notification" pload p1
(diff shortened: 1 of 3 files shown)


services/gundeck/test/integration/API.hs
```diff
@@ -259,6 +259,9 @@ sendMultipleUsers gu ca _ _ = do
             let p2 = fromJSON (Object (List1.head (n^.queuedNotificationPayload)))
             in assertEqual "Wrong events in notification" (Success (PushRemove tok)) p2
   where
+    checkNotifications []     = error "No notifications received!"
```
elverkilde: this seems redundant, since there's already a `length >= 2` assertion above?

tiago-loureiro: Yes, but it causes a
```
wire-server/services/gundeck/test/integration/API.hs:253:13: warning: [-Wincomplete-uni-patterns]
    Pattern match(es) are non-exhaustive
    In a pattern binding: Patterns not matched: []
```
which would be nice to avoid imho

elverkilde: ok, makes sense. I don't know if we get a better error message if we use `assertFail` instead of `error`, they are probably similar.

tiago-loureiro: Yeah, `error` is a bit more convenient in this case as `assertFailure` returns `IO ()`

PR #170, opened 2018-01-10
Release/20180110

tiago-loureiro:
Highlights:

- [proxy]: allow proxying of /giphy/v1/gifs/
- [brig]: welcome emails after successful team creation/join
- [brig/galley]: bots in team conversations, filter bot by prefix
- defaults for files in tests, read vars from config/env instead of hardcoded for tests
- fix some of the dependencies, ensure top level stack.yaml builds everything


tiago-loureiro:
> 👍 - could you just also add this commit to the release?

👍

PR #171, opened 2018-01-10
Upgrade to tasty-1.0.0.1

tiago-loureiro:
The latest version of tasty is pretty useful as it allows using [pattern tests](https://github.com/feuerbach/tasty#patterns) as _awk expressions_.

Unfortunately it's pretty bleeding edge so not present in any LTS.


tiago-loureiro:
> The root stack.yaml also needs to be updated.

Right, thank you!

PR #174, opened 2018-01-12
Expire team invitations after a certain amount of time

tiago-loureiro:
Allow setting an expiration time for team invitations


services/brig/brig.yaml.example
```diff
@@ -72,6 +72,7 @@ turn:
 
 optSettings:
   setActivationTimeout: 5
+  setTeamInvitationTimeout: 10
```
jschaul: Doesn't 5 seconds suffice for testing purposes?

tiago-loureiro: > Doesn't 5 seconds suffice for testing purposes?
 
👍

PR #175, opened 2018-01-16
Optionally filter services by provider

tiago-loureiro:
Add an option when starting `brig` to list only services from a specific providerID on the `GET "/services"` endpoint.

Adding tests for this is tricky but I ran them without the option (and all tests succeed) and then added `--provider-id-search-filter <random uuid_v4>` on the Procfile which causes the provider/service search test to fail (which is what you'd expect - no services are returned).
The happy path was checked locally by launching brig with `--provider-id-search-filter=<known_providerid>` and verified that only services for that provider were returned from `GET /services`. Clearly room for improvement...

### DIFF
--- a/services/brig/src/Brig/Provider/DB.hs
+++ b/services/brig/src/Brig/Provider/DB.hs
@@ -410,16 +410,19 @@ paginateServiceTags :: MonadClient m
     -> Maybe ProviderId
     -> m ServiceProfilePage
 paginateServiceTags tags start size providerFilter = liftClient $ do
-    let start' = Name (maybe "" (toLower . fromName) start)
-    let size'  = size + 1
-    let tags'  = unpackTags tags
-    p <- maybeFilterbyProvider providerFilter . filterPrefix (fromName start') <$> queryAll start' size' tags'
+    let size' = size + 1
+    let tags' = unpackTags tags
+    p <- filterResults <$> queryAll start' size' tags'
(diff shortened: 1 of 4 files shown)


services/brig/src/Brig/Provider/DB.hs
```diff
@@ -407,15 +407,25 @@ paginateServiceTags :: MonadClient m
...
+paginateServiceTags tags start size providerFilter = liftClient $ do
     let start' = Name (maybe "" (toLower . fromName) start)
     let size'  = size + 1
     let tags'  = unpackTags tags
-    p <- filterPrefix (fromName start') <$> queryAll start' size' tags'
+    p <- maybeFilterbyProvider providerFilter . filterPrefix (fromName start') <$> queryAll start' size' tags'
     r <- mapConcurrently resolveRow (result p)
     return $! ServiceProfilePage (hasMore p) (catMaybes r)
   where
+    maybeFilterbyProvider :: Maybe ProviderId -> Page TagRow -> Page TagRow
```
elverkilde: you could also define `filterByProvider :: ProviderId -> Page TagRow -> Page TagRow`, then `maybeFilterbyProvider = maybe id filterByProvider`. Not sure if that reads better?

tiago-loureiro: Thought the same... I made a little change though, think it reads better now

PR #177, opened 2018-01-17
Speed up test for too many tokens

tiago-loureiro:
With this change we no longer use `brig` for user/client creation and create only 1 push token per user, halving the number of token creation requests to SNS. In addition, we also do not clean up SNS/cassandra.

Locally, the test now runs in ~25 secs (vs. the previous 120)

### DIFF
--- a/services/gundeck/test/integration/API.hs
+++ b/services/gundeck/test/integration/API.hs
@@ -13,7 +13,7 @@ import Bilge
 import Bilge.Assert
 import Control.Arrow ((&&&))
 import Control.Concurrent
-import Control.Concurrent.Async       (Async, async, wait, replicateConcurrently)
+import Control.Concurrent.Async       (Async, async, wait)
 import Control.Concurrent.STM.TChan
 import Control.Lens                   ((&), (.~), (^.), (^?), view)
 import Control.Monad.IO.Class         (MonadIO)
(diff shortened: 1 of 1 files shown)


services/gundeck/test/integration/API.hs
```diff
@@ -614,33 +614,19 @@ testRegisterPushToken g _ b _ = do
 
 -- TODO: Try to make this test more performant, this test takes too long right now
 testRegisterTooManyTokens :: TestSignature ()
-testRegisterTooManyTokens g _ b _ = do
+testRegisterTooManyTokens g _ _ _ = do
     -- create tokens for reuse with multiple users
-    apsTok <- Token . T.decodeUtf8 . B16.encode <$> randomBytes 32
     gcmTok <- Token . T.decodeUtf8 . toByteString' <$> randomId
-    
+    uids   <- liftIO $ replicateConcurrently 55 randomId
```
elverkilde: does this actually speed things up? I would imagine most of the time is spent in the http calls? Would it not make sense to try parallelizing those instead?

services/gundeck/test/integration/API.hs
```diff
@@ -614,33 +614,19 @@ testRegisterPushToken g _ b _ = do
...
     -- create 55 users with these tokens, which should succeed
-    userClients <- replicateM 55 $ registerToken 201 apsTok gcmTok
-
+    mapM_ (registerToken 201 gcmTok) uids
```
elverkilde: Can't these be done in parallel?

jschaul:
> we also do not clean up SNS

is that a good idea, though? Would it be wise to have another SNS cleanup task, perhaps?

services/gundeck/test/integration/API.hs
tiago-loureiro: > does this actually speed things up?

Not really, I will actually remove that.

services/gundeck/test/integration/API.hs
tiago-loureiro: Unfortunately these cannot, as concurrent updates would break the test.

tiago-loureiro:
> is that a good idea, though? Would it be wise to have another SNS cleanup task, perhaps?

Yes, a clean up task is a good idea; I'd rather have that instead since cleanups during the test make them slower + if a test fails then those resources are "left lingering"

PR #184, opened 2018-01-23
Read twilio and nexmo credentials from files

tiago-loureiro:
Read secrets from files instead of directly from command line arguments


tiago-loureiro:
Ah, well spotted: thanks! Fixed with: 8f819ab4c19c47968a4218e97a51d0c50000e42e

tiago-loureiro:
Made some changes based on feedback, let me know if this reads better

services/brig/test/resources/nexmo.txt
```diff
@@ -0,0 +1,2 @@
+key: "dummy"
```
jschaul: is this file `yaml`? maybe change the extension from `.txt` to `.yaml`?

tiago-loureiro: 👍

PR #187, opened 2018-01-24
Feature/fake sqs galley

jschaul:
This change will by default use fake-SQS instead of the real one during tests in order to not have interference when multiple people run galley integration tests at the same time.
Backward compatibility is maintained by supporting both config files and ENV variables. (I'm planning another PR to remove ENV variables in tests and only rely on config files)

Requires to run the `fake-sqs.sh` script or similar when testing locally. (Alternatively a pod can be started with fake-sqs on kubernetes)


services/galley/fake-sqs.sh
```diff
@@ -0,0 +1,19 @@
...
+  docker kill $(cat DOCKER_ID)
+  rm DOCKER_ID
+  printf done
+}
+
+trap cleanup EXIT
+
+docker run -d --cidfile DOCKER_ID -p 4568:4568 airdock/fake-sqs 
+sleep 2
+curl http://localhost:4568 -d "Action=CreateQueue&QueueName=integration-team-events.fifo"
```
tiago-loureiro: Instead of a `sleep n`, could we instead run `curl` until it succeeds, sth like
```
while true
do
  curl --silent http://localhost:4568 -d "Action=CreateQueue&QueueName=integration-team-events.fifo"
  if [ $? -eq 0 ]; then
    echo "Queue created, all set!"
    break
  else
    echo "Waiting for the queue to be created..."
  fi
  sleep 1
done
```

services/galley/Procfile
```diff
@@ -1 +1 @@
-web: LOG_LEVEL=Info ./dist/galley -p $PORT --brig-host=$BRIG_WEB_HOST --brig-port=$BRIG_WEB_PORT --gundeck-host=$GUNDECK_WEB_HOST --gundeck-port=$GUNDECK_WEB_PORT --cassandra-host=$GALLEY_CASSANDRA_HOST --cassandra-port=$GALLEY_CASSANDRA_PORT --cassandra-keyspace $GALLEY_CASSANDRA_KEYSPACE --team-events-queue-name $GALLEY_SQS_TEAM_EVENTS --aws-region $KHAN_REGION
+web: LOG_LEVEL=Info ./dist/galley -p $PORT --brig-host=$BRIG_WEB_HOST --brig-port=$BRIG_WEB_PORT --gundeck-host=$GUNDECK_WEB_HOST --gundeck-port=$GUNDECK_WEB_PORT --cassandra-host=$GALLEY_CASSANDRA_HOST --cassandra-port=$GALLEY_CASSANDRA_PORT --cassandra-keyspace $GALLEY_CASSANDRA_KEYSPACE --team-events-queue-name $GALLEY_SQS_TEAM_EVENTS --aws-region $KHAN_REGION --fake-sqs-host $FAKE_SQS_HOST --fake-sqs-port $FAKE_SQS_PORT
```
tiago-loureiro: You could add a line here:
```
sqs: ./fake-sqs.sh
```
which would start `fake-sqs` together with `galley`

services/galley/fake-sqs.sh
jschaul: Good idea, thanks!

services/galley/Procfile
jschaul: I tried that; it doesn't work due to timing issues, i.e. it will start galley and start attempting to connect to SQS before SQS is up and the queue created. Adding a `sleep` before starting galley also doesn't work, as it then shows errors about `/i/monitoring` returning a 500. It seems keiretsu is quite eager. 
I propose we investigate using some other form of scheduling for all of the dependencies (cassandra, elasticsearch, redis, and now fake-sqs, ...) whether with docker compose or other means. That is not directly related to this PR though.

tiago-loureiro:
Made a few changes, there is a single option `endpoint` which can be configured to point to wherever we want

PR #192, opened 2018-02-01
Issue 489

tiago-loureiro:
This PR migrates `brig` from the `aws` package to `amazonka` and also allows overriding the AWS endpoints (from AWS to anything else we might want to use). All tests pass using real AWS endpoints but still a few details to iron out namely:

* [x] tests that depend on SQS can now use `fake-sqs.sh`; blacklisting does not work due to the lack of feedback from SES
* [x] try out local SES
* [x] try out local DynamoDB
* [x] clean up some types
* [x] resolving any pending TODO


services/brig/.env
```diff
@@ -4,3 +4,7 @@ BRIG_CASSANDRA_PORT=9042
 BRIG_CASSANDRA_KEYSPACE=brig_test
 BRIG_ELASTICSEARCH_URL=http://localhost:9200
 BRIG_ELASTICSEARCH_USER_INDEX=directory_test
+# AWS_SQS_ENDPOINT=http://localhost:4568
```
jschaul: Unfortunately these `.env` files don't support any comments (at least when loaded from keiretsu). Perhaps `localhost` should be the default in this `.env` files, i.e. the default when testing.

services/brig/Makefile
```diff
@@ -89,6 +89,10 @@ integration-list: install
 integration-%: db-migrate index-reset
 	LOG_LEVEL=Info keiretsu --run '$(EXE_IT) --pattern="$*"' --env $(KEIRETSU_ENV)
 
+test-fast:
+	stack install --fast --test --local-bin-path=dist
+	cp "$(shell stack path --dist-dir)/build/$(NAME)-integration/$(NAME)-integration" dist/
```
jschaul: I don't think copying is needed anymore (integration test is defined as executable now in .cabal file). Also `test:` should be sufficient as a target.

services/brig/fake-sqs.sh
```diff
@@ -0,0 +1,29 @@
...
+
+trap cleanup EXIT
+
+docker run -d --cidfile DOCKER_ID -p 4568:4568 airdock/fake-sqs 
+while true
+do
+  curl --silent http://localhost:4568 -d "Action=CreateQueue&QueueName=integration-brig-events"
+  if [ $? -eq 0 ]; then
+    echo "Queue created!"
+    curl --silent http://localhost:4568 -d "Action=CreateQueue&QueueName=integration-brig-events-internal"
```
jschaul: let's consolidate this with the galley one, and move this script one level up to `services`, creating queues for both brig and galley. There'll be interference otherwise.

tiago-loureiro: Yup, was planning to do so, thanks!

services/brig/Makefile
tiago-loureiro: Yup, agreed

services/brig/.env
tiago-loureiro: 👍

tiago-loureiro:
This is now ready for review. You can run it against local SQS/DynamoDB except that it will cause the [blacklist](https://github.com/wireapp/wire-server/blob/develop/services/brig/test/integration/API.hs#L78) test to fail so I decided to default to AWS services until that can be automated as well (idea would be to use a local SES and, in that test, push bounce/complaint notifications to the local queue - will do that in a subsequent PR).

PR #196, opened 2018-02-06
Issue 631

tiago-loureiro:
Make conversation size configurable and make use of that to make tests that test limits to run faster. Lowered some limits to speed up other tests as well.


services/galley/src/Galley/API/Update.hs
```diff
@@ -126,31 +126,31 @@ addMembers (zusr ::: zcon ::: cid ::: req ::: _) = do
         Data.deleteConversation cid
         throwM convNotFound
     let mems = botsAndUsers (Data.convMembers conv)
-    to_add <- rangeChecked (toList $ invUsers body) :: Galley (Range 1 128 [UserId])
-    let new_users = filter (notIsMember conv) (fromRange to_add)
+    toAdd <- rangeChecked (toList $ invUsers body) :: Galley (Range 1 128 [UserId])
```
jschaul: This still has 128 hardcoded in its type

PR #197, opened 2018-02-07
Feature/cassandra prefix search

tiago-loureiro:
This PR introduces the possibility of searching for services using only a prefix, i.e., the `start` parameter. For that, we use an index based on the first character of the service name, similarly to how we use `tags` (which can still be used together with `start`).

It also introduces some `batch`ing when updating services vs. the service index; i.e., on `name` and/or `tag` changes, both the `service` and the `indexes` are updated in a batch.

### DIFF
--- a/services/brig/src/Brig/Provider/API.hs
+++ b/services/brig/src/Brig/Provider/API.hs
@@ -196,7 +196,7 @@ routes = do
     get "/services" (continue listServiceProfilesByTag) $
         accept "application" "json"
         .&> zauth ZAuthAccess
-        .&> query "tags"
+        .&> opt (query "tags")
         .&. opt (query "start")
         .&. def (unsafeRange 20) (query "size")
 
@@ -469,9 +469,10 @@ deleteService (pid ::: sid ::: req) = do
         throwStd badCredentials
     svc  <- DB.lookupService pid sid >>= maybeServiceNotFound
     let tags = unsafeRange (serviceTags svc)
-    DB.deleteServiceTags pid sid (serviceName svc) tags
+        name = serviceName svc
+    DB.deleteServiceTags pid sid name tags
     lift $ RPC.removeServiceConn pid sid
-    DB.deleteService pid sid
+    DB.deleteService pid sid name
     return empty
 
 deleteAccount :: ProviderId ::: Request -> Handler Response
@@ -485,9 +486,10 @@ deleteAccount (pid ::: req) = do
     forM_ svcs $ \svc -> do
         let sid  = serviceId svc
         let tags = unsafeRange (serviceTags svc)
+            name = serviceName svc
         DB.deleteServiceTags pid sid (serviceName svc) tags
         lift $ RPC.removeServiceConn pid sid
-        DB.deleteService pid sid
+        DB.deleteService pid sid name
     DB.deleteKey (mkEmailKey (providerEmail prov))
     DB.deleteAccount pid
     return empty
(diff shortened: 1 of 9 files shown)


services/brig/src/Brig/Provider/API.hs
```diff
@@ -509,10 +507,15 @@ getServiceProfile (pid ::: sid) = do
     s <- DB.lookupServiceProfile pid sid >>= maybeServiceNotFound
     return (json s)
 
-listServiceProfilesByTag :: QueryAnyTags 1 3 ::: Maybe Name ::: Range 10 100 Int32 -> Handler Response
-listServiceProfilesByTag (tags ::: start ::: size) = do
+searchServiceProfiles :: Maybe (QueryAnyTags 1 3) ::: Maybe Name ::: Range 10 100 Int32 -> Handler Response
+searchServiceProfiles (Nothing ::: Just start ::: size) = do
```
jschaul: `start :: Name` is used as a prefix here, right? If it's not a full name, wouldn't `Text` be a better type here?

services/brig/src/Brig/Provider/DB.hs
```diff
@@ -339,33 +361,75 @@ updateServiceConn pid sid url tokens keys enabled = retry x5 $ batch $ do
...
     cqlEnabled = "UPDATE service SET enabled = ? WHERE provider = ? AND id = ?"
 
+--------------------------------------------------------------------------------
+-- Service "Indexes" (tag and prefix)
+
+insertServiceIndexes :: MonadClient m
+                     => ProviderId
+                     -> ServiceId
+                     -> Name
+                     -> Range 0 3 (Set.Set ServiceTag)
```
jschaul: How about a type alias for `Range 0 3 (Set.Set ServiceTag)`?

services/brig/src/Brig/Provider/API.hs
```diff
@@ -509,10 +507,15 @@ getServiceProfile (pid ::: sid) = do
...
+    ss <- DB.paginateServiceNames start (fromRange size) =<< setProviderSearchFilter <$> view settings
```
jschaul: `start` can be `Just ""` in which case the cql lookup won't work (return no results). Or is there code handling the empty string elsewhere?

services/brig/test/integration/API/Provider.hs
```diff
@@ -358,27 +358,73 @@ testListServicesByTagAndPrefix config db brig = do
...
+        liftIO $ assertEqual ("get page with tag, size: " ++ show qry) (length expectsNames) (length _ls)
+        let _names = map serviceProfileName _ls
+        liftIO $ assertEqual ("get page with tag, str: " ++ show qry) expectsNames _names
+        -- Search with name only
+        _ls <- getPage uid Nothing (Just qry) size
+        liftIO $ assertEqual ("get page no tag, size: " ++ show qry) (length expectsNames) (length _ls)
         let _names = map serviceProfileName _ls
-        liftIO $ assertEqual ("str: " ++ show qry) expects _names
+        liftIO $ assertEqual ("get page no tag: " ++ show qry) expectsNames _names
+        return _ls
```
jschaul: How about a testcase searching with an empty prefix?

services/brig/src/Brig/Provider/DB.hs
tiago-loureiro: 👍

services/brig/src/Brig/Provider/API.hs
tiago-loureiro: Right... this was overlooked. I thought the `FromByteString` instance checked the size as well but only the `FromJSON` does.
I think the right here is actually to use a `Range 1 128 Text` type, which also solves what you described [earlier on]( 
https://github.com/wireapp/wire-server/pull/197/files/96f4cf6d190f5e43d498b96aeb8be01b7aa3df9a#r167524840)

services/brig/test/integration/API/Provider.hs
tiago-loureiro: I didn't add it because searching with no tags and no prefix (at least that was the intention of the API), we should return a 400.
Alternatively, we can do what I was aiming for at first which was: if no prefix and no tag, return random services, potentially from a whitelisted provider. Any thoughts?

elverkilde: I think we should deny it on the API level, i.e. return 400. My only concern is, I don't think that's what the code currently does, that's why I was thinking a test could prove me wrong.

tiago-loureiro: That was fixed with the [previous commit](https://github.com/wireapp/wire-server/pull/197/commits/7885d789cda94d9dcace1de9b02f5b7fd833078b) and I just [added a test for this case](https://github.com/wireapp/wire-server/pull/197/commits/8e4e84f4146631e9d8a39cb69033b46847c74516)

PR #198, opened 2018-02-08
Feature/external conversation joining

jschaul:
Introduces `code` (and an implicit `team-invite`) conversation access modes as discussed in issue-674. This PR only addresses part of issue-674.


services/galley/deb/etc/sv/galley/run
```diff
@@ -46,5 +46,6 @@ exec chpst -u $USER \
     --http-pool-size=${GALLEY_HTTP_POOL_SIZE:-128} \
     --team-events-queue-name=${GALLEY_SQS_TEAM_EVENTS?'unset'} \
     --aws-sqs-endpoint=${GALLEY_SQS_ENDPOINT?'unset'} \
-    --conv-team-max-size=${GALLEY_CONV_TEAM_MAX_SIZE:-128} \
+    --conv-team-max-size=${CONV_AND_TEAM_MAX_SIZE:-128} \
+    --conversation-code-uri=${GALLEY_CONVERSATION_code_URI?'unset'} \
```
tiago-loureiro: Should we add `PREFIX` to make it (even) more explicit?

services/galley/schema/src/V25.hs
```diff
@@ -0,0 +1,19 @@
...
+import Text.RawString.QQ
+
+migration :: Migration
+migration = Migration 25 "Add conversation_codes table" $
+    schema' [r|
+        CREATE TABLE conversation_codes (
+            key           ascii, -- opaque conversation ID
+            conversation  uuid,
+            value         ascii, -- secret value
+            PRIMARY KEY (key, conversation)
```
tiago-loureiro: Perhaps it would be useful to add a `scope` here too to allow for other types of codes, similarly to the `vcodes` in `brig`?
Also, if `(key, conversation)` is the `PRIMARY KEY` then we can create the same `key` for different conversations so we should use only `key` as the `PRIMARY KEY`. Potentially, if `scope` is used, then a `PRIMARY KEY` like `(key, scope)` would make sense too.

services/galley/src/Galley/API/Update.hs
```diff
@@ -105,13 +113,114 @@ unblockConv (usr ::: conn ::: cnv) = do
...
+returnCode :: Code -> Galley Response
+returnCode c = do
+    urlPrefix <- view $ options . optSettings . setConversationCodeURI
+    let res = mkConversationCode (codeKey c) (codeValue c) urlPrefix
+    return $ setStatus status200 . json $ res
+
+joinConversationByCode :: UserId ::: ConnId ::: Request ::: JSON -> Galley Response
+joinConversationByCode (zusr ::: zcon ::: req ::: _) = do
+    body <- fromBody req invalidPayload
+    cnv <- codeConversation <$> (Data.lookupCode (conversationKey body) >>= ifNothing convNotFound)
```
tiago-loureiro: Should this return `codeNotFound` instead?

services/galley/src/Galley/API/Update.hs
```diff
@@ -105,13 +113,114 @@ unblockConv (usr ::: conn ::: cnv) = do
...
+    let targetAccess = Set.fromList (toList (cupAccess body))
+    -- checks and balances
+    when (PrivateAccess `elem` targetAccess) $
+        throwM invalidTargetAccess
+    (bots, users) <- botsAndUsers <$> Data.members cnv
+    unless (usr `isMember` users) $
+        throwM convNotFound
+    conv <- Data.conversation cnv >>= ifNothing convNotFound
+    let currentAccess = Set.fromList (toList $ Data.convAccess conv)
+    case Data.convType conv of
```
tiago-loureiro: You have `ensureGroupConv` [here](https://github.com/wireapp/wire-server/blob/5beb3d8a1753fc3489507d679dfffbaeab55592c/services/galley/src/Galley/API/Update.hs#L413-L418) already.

services/galley/src/Galley/API/Update.hs
```diff
@@ -105,13 +113,114 @@ unblockConv (usr ::: conn ::: cnv) = do
...
+        SelfConv    -> throwM invalidSelfOp
+        One2OneConv -> throwM invalidOne2OneOp
+        ConnectConv -> throwM invalidConnectOp
+        _           -> return ()
+    -- remove conversation codes if CodeAccess is revoked
+    when (CodeAccess `elem` currentAccess && CodeAccess `notElem` targetAccess) $ do
```
tiago-loureiro: Since only team members can change access for team conversations, better if we do the team membership first otherwise we might revoke codes in team conversations

services/galley/src/Galley/API/Update.hs
```diff
@@ -105,13 +113,114 @@ unblockConv (usr ::: conn ::: cnv) = do
...
+        Data.deleteCode key
+    -- special case TeamConversation
+    case Data.convTeam conv of
+        Nothing     ->
+            when (TeamAccess `elem` targetAccess) $
+                throwM $ invalidTargetAccess
+        Just tid    -> do
+            tMembers <- Data.teamMembers tid
+            -- only team members can change access mode
+            unless (usr `elem` (memId <$> users)) $
```
tiago-loureiro: I think we're checking team membership here, so
```
unless (usr `elem` (view userId <$> tMembers))
```
?

services/galley/src/Galley/API/Update.hs
```diff
@@ -105,13 +113,114 @@ unblockConv (usr ::: conn ::: cnv) = do
...
+                throwM $ accessDenied
+            -- remove non-team users if the only mode left is TeamAccess
+            when (targetAccess == Set.fromList [TeamAccess]) $
+                handleTeamOnly tid tMembers users bots conv
+    -- update cassandra & send event
+    now <- liftIO getCurrentTime
+    let e = Event ConvAccessUpdate cnv usr now (Just $ EdConvAccessUpdate body )
+    unless (currentAccess == targetAccess) $ do
```
tiago-loureiro: Maybe we can check this in the very beginning and make the whole handler a no-op?

tiago-loureiro:
Another thought: since users cannot join managed conversations, we should also ensure that `CodeAccess` cannot be set (one should not be able to generate links for such conversations)

services/galley/src/Galley/API.hs
```diff
@@ -469,6 +469,85 @@ sitemap = do
...
+
+    put "/conversations/:cnv/access" (continue updateConversationAccess) $
+        zauthUserId
+        .&. zauthConnId
+        .&. capture "cnv"
+        .&. request
+        .&. contentType "application" "json"
+
+    document "PUT" "updateConversationAccess" $ do
+        summary "Join a conversation using a code"
```
tiago-loureiro: This endpoint is to change the access mode, right?

services/galley/src/Galley/API/Update.hs
```diff
@@ -105,13 +113,114 @@ unblockConv (usr ::: conn ::: cnv) = do
     conv' <- acceptOne2One usr conv conn
     setStatus status200 . json <$> conversationView usr conv'
 
-joinConversation :: UserId ::: ConnId ::: ConvId ::: JSON -> Galley Response
-joinConversation (zusr ::: zcon ::: cnv ::: _) = do
+updateConversationAccess :: UserId ::: ConnId ::: ConvId ::: Request ::: JSON -> Galley Response
+updateConversationAccess (usr ::: zcon ::: cnv ::: req ::: _ ) = do
+    body <- fromBody req invalidPayload
+    let targetAccess = Set.fromList (toList (cupAccess body))
```
tiago-loureiro: The semantics of the access mode update endpoint are that clients should specify the full set of access modes, right?
In that case, does an access mode of `Invite | TeamOnly | Code` make sense, for instance?

jschaul: > The semantics of the access mode update endpoint are that clients should specify the full set of access modes, right?
Yes

> In that case, does an access mode of Invite | TeamOnly | Code make sense, for instance?
Well, it's `team`, which is meant to mean "team members can join". It's not "team-only".

tiago-loureiro: Right, my bad there. But so wouldn't it make sense to have a conversation access as `Invite | Team` that would also cause non-team members to be kicked out? Or by having a `Team` access only we mean that users can still be added to the conversation as long as they are part of the same team?

PR #199, opened 2018-02-09
Issue 667

tiago-loureiro:
With this PR we introduce the possibility of changing emails and resetting/changing the password for providers. In addition, provider tokens should also have a more limited lifetime which we should change in the config.

This is better viewed comparing it against a [different branch](https://github.com/wireapp/wire-server/compare/feature/cassandra_prefix_search...issue-667) - that PR should be merged in first.


services/brig/test/integration/API/Provider.hs
```diff
@@ -8,7 +8,7 @@ module API.Provider (tests, Config) where
 
 import Bilge hiding (accept, timeout, head)
 import Bilge.Assert
-import Brig.Types
+import Brig.Types hiding (NewPasswordReset (..), CompletePasswordReset(..), EmailUpdate (..), PasswordReset (..), PasswordChange (..))
```
elverkilde: Oh dear, that's a lot of hiding. Did you try qualifying one of the imports instead?

tiago-loureiro: I thought the same... but importing qualified results in ~100 errors (either with `Brig.Types` or `Brig.Types.Provider`). Also importing only the needed types results in an even "uglier" import :/

tiago-loureiro:
Fixed the PR after rebasing.

elverkilde:
👍

PR #201, opened 2018-02-13
Fake ses feedback

tiago-loureiro:
This PR changes `brig`s test defaults to use fake AWS services; however, due to the feedback `SNS -> SQS`, we need to fake these events when using faked services but I think this is fine for testing purposes.

Since localstack is quite a large dependency, I tried migrating dynamo and sqs but couldn't get it all running smoothly so this setup worked nicely and I got all tests passing locally in 80 secs.


services/brig/test/integration/Main.hs
```diff
@@ -76,6 +80,27 @@ runTests iConf bConf = do
   where
     mkRequest (Endpoint h p) = host (encodeUtf8 h) . port p
 
+    mkLocalAWSEnv :: Logger.Logger -> Opts.AWSOpts -> Manager -> IO (Maybe AWS.Env)
+    mkLocalAWSEnv l o m
+        | "amazonaws.com" `isSuffixOf` (decodeLatin1 ((Opts.sesEndpoint o)^.awsHost)) = return Nothing
```
tiago-loureiro: We only really need an AWS env for SES but it's easier to create full `AWSOpts` so that we can reuse code from `Brig.AWS`

PR #206, opened 2018-02-20
Pull dist path from environment

elverkilde:
This allows building all executables in one folder and use that for running integration tests (i.e. top-level stack install).


tiago-loureiro:
👍  When running `keiretsu --env <env>`, requires `DIST` to be defined on that `env` file, right?

PR #207, opened 2018-02-20
Ensure 'Just Range' is within Range

tiago-loureiro:
This PR changes the behaviour of `rangeCheckedMaybe` (which is used in the creation of every type of conversation) to verify that if we have a `Just`, then this `Just` is within the required `Range`.
Thus, when creating conversations, if `Name` is specified, then it must be within the `[1, 256]` range.

Also fixed one test which was failing after the configurable options for max connections was introduced.

### DIFF
--- a/services/galley/test/integration/API/Teams.hs
+++ b/services/galley/test/integration/API/Teams.hs
@@ -7,7 +7,7 @@ import Bilge hiding (timeout)
 import Bilge.Assert
 import Control.Concurrent.Async (mapConcurrently)
 import Control.Lens hiding ((#), (.=))
-import Control.Monad (void, replicateM)
+import Control.Monad (void)
 import Control.Monad.IO.Class
 import Data.Aeson hiding (json)
 import Data.ByteString.Conversion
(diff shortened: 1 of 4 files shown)


services/galley/test/integration/API/Teams.hs
```diff
@@ -888,20 +888,21 @@ postCryptoBroadcastMessageProto g b c a = do
...
     (bob,   bc) <- randomUserWithClient b (someLastPrekeys !! 1)
     connectUsers b alice (list1 bob [])
     let msg = [(bob, bc, "ciphertext1")]
     Util.postOtrBroadcastMessage id g alice ac msg !!! const 404 === statusCode
 
-postCryptoBroadcastMessage100 :: Galley -> Brig -> Cannon -> Maybe Aws.Env -> Http ()
-postCryptoBroadcastMessage100 g b c a = do
+postCryptoBroadcastMessageMaxConns :: Galley -> Brig -> Cannon -> Maybe Aws.Env -> TestSetup -> Http ()
+postCryptoBroadcastMessageMaxConns g b c a s = do
+    let maxConns = fromIntegral (maxConvTeamSize s)
```
jschaul: At the moment, the default for maxConvTeamSize is 128, while the default for the maximum number of connections is 1000. While this test could benefit from a configurable amount of connected users, I think we shouldn't mix these two constants here.

tiago-loureiro: Good point. However, since `galley` does not know what this number is (and we need to since creating more connections will cause the test to fail), I suggest we create as many connections as possible (i.e., until `brig` returns us a `403`) or we reach 100 users, whichever comes first.
https://github.com/wireapp/wire-server/pull/207/commits/34446804188851a3599560c02dd7d40a5b4ae998

PR #211, opened 2018-02-22
Allow SQS/SNS endpoints to be configured

tiago-loureiro:
This PR allows SNS/SQS to be configured for gundeck as well. Sadly, [these](https://github.com/wireapp/wire-server/blob/develop/services/gundeck/test/integration/API.hs#L116-L119) test do not pass due to the fact that the mocked SNS service returns a slightly different text in the exception (and SNS does not provide very granular responses when errors occur as can be seen [here](https://github.com/wireapp/wire-server/blob/5cdae7198e6ad38c04ad803cc43898262bc7dd28/services/gundeck/src/Gundeck/Aws.hs#L410-L431).

It's still useful to be able to run gundeck without AWS resources, only native push notifications will not work (and perhaps we can PR a change in localstack to make this work nicer, it seems that subscriptions SNS -> SQS are indeed implemented).


tiago-loureiro:
I added an extra commit https://github.com/wireapp/wire-server/pull/211/commits/7947e18dfea765199e3ad51cb103e09c210f10bd so that tests default to using local AWS fake services.
One can easily then just run the tests by running `make integration-fake-aws`

jschaul:
Which implementations of fake SNS did you try? Only the localstack one? (which seems not very good/stable/complete) Perhaps it would be an idea to try different implementations, maybe one of e.g. 

https://github.com/p4tin/goaws
https://github.com/s12v/sns
https://github.com/Comcast/cmb
(maybe there are others)

behaves more closely like real the real AWS?

tiago-loureiro:
> It would be good to provide such an explicit configuration option then - i.e. irrespective of testing, it should be possible to start gundeck in a websocket-only mode (without requiring a fake SNS/SQS to be up).

Agreed, but I'd rather do that once we move on to use config files only? It's a bit cumbersome to do that with command line args

tiago-loureiro:
I tried a few, namely [s12n](https://github.com/s12v/sns) but

> https://github.com/Comcast/cmb
> https://github.com/p4tin/goaws

seem promising

PR #208, opened 2018-02-20
Fetch emails from wire-emails

lipis:
### In short we are going from [this](https://github.com/wireapp/wire-server/tree/develop/services/brig/deb/opt/brig/templates) to [this](https://github.com/wireapp/wire-server/tree/new-emails/services/brig/deb/opt/brig/templates)

---

In order to fetch the latest emails/translations to the project:

1. Edit `services/brig/deb/opt/brig/new-version` with the latest/desired tag from [wire-emails](https://github.com/wireapp/wire-emails) in new PR
2. Wait until TravisCI will fetch and update the same PR with the latest emails

PR #217, opened 2018-02-27
Add travis file that builds & pushes docker images

jschaul:
Adds a few docker images and Makefile targets. 'wireserver' on dockerhub becomes the default namespace (I'll convert this shortly to an organization).

`docker-builder` is currently uploaded manually due to it taking >50 minutes (which is the travis hard limit)

Example build: https://travis-ci.org/wireapp/wire-server/builds/346911997
Resulting docker images: https://hub.docker.com/r/wireserver/

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,9 +6,9 @@ services:
 env:
   global:
     - DOCKER_TAG=0.1.${TRAVIS_BUILD_NUMBER}
-    - DOCKER_PUSH=1
+    - DOCKER_PUSH=1 # if DOCKER_PUSH is defined, `make docker-*` uploads images to dockerhub.
   matrix:
-    # - TARGETS="docker-builder" # currently takes ~ 2h (needs to compile GHC) which is more than travis permits.
+    # - TARGETS="docker-builder" # currently takes ~ 2h (needs to compile GHC) which is more than travis permits. This image will, until a more recent ghc is available as an alpine package, be uploaded manually/through wire-internal CI"
     - TARGETS="docker-deps"
     - TARGETS="docker-intermediate docker-migrations"
     - TARGETS="docker-intermediate
@@ -46,6 +46,6 @@ script:
 notifications:
   email: false
 
-#branches:
-#only:
-#  - develop
+branches:
+  only:
+    - develop
--- a/.travis.yml
+++ b/.travis.yml
@@ -46,6 +46,6 @@ script:
 notifications:
   email: false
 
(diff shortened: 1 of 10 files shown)


.travis.yml
```diff
@@ -0,0 +1,51 @@
...
+              "
+
+script:
+  - docker login -u ${DOCKER_USERNAME} -p ${DOCKER_PASSWORD}
+  - make ${TARGETS[@]}
+
+notifications:
+  email: false
+
+#branches:
```
tiago-loureiro: So with the current config it will trigger a build for all branches, right?

jschaul: I'll uncomment this before merging; I just needed travis builds to run on this PR to test them.

.travis.yml
```diff
@@ -0,0 +1,51 @@
...
+script:
+  - docker login -u ${DOCKER_USERNAME} -p ${DOCKER_PASSWORD}
+  - make ${TARGETS[@]}
+
+notifications:
+  email: false
+
+branches:
+  only:
+    - develop
```
lipis: This YAML is not valid.. (notice the `script`, `notification`, etc are not green as `service`, `env` keys.. 

https://codebeautify.org/yaml-validator/cb9d3cdc

jschaul: Travis seems to parse it fine, it builds the intended things, and both https://yamlvalidator.com/ and http://www.yamllint.com/ say "valid yaml".

I didn't try what happens if there's a 

```yaml
env:
    matrix:
        - >
          TARGETS="..."
```
bit, which might make your quoted validator happy but might confuse travis, I'm not sure.

lipis: @jschaul I think is not valid as on travis it [shows two targets](https://travis-ci.org/wireapp/wire-server/builds/349934386?utm_source=github_status&utm_medium=notification) while you have three. Maybe close the `"` on the same line instead of a separate (Line 39)

jschaul: the build you quoted (84) was a manual test with a different config, trying the yaml multi-line string syntax ">", which failed. 

The travis file as-is on the develop branch is valid for travis' purposes, builds three targets, see e.g. [build 91](https://travis-ci.org/wireapp/wire-server/builds/350267424)

I can put all the targets space-separated into the same line, though that will make one long, unreadable line. Would this help you in your environment?

lipis: I'm just trying to figure out how can I run my stuff in #208 and I can't find out a way :)

PR #222, opened 2018-03-02
Release/20180302

tiago-loureiro:
Highlights:
 * [galley]:
   * fake AWS services
   * document error responses
   * configurable team/conversation sizes
   * joining conversations
   * error when conversation name is too large instead of silently failing

 * [brig]:
    * migrate to amazonka
    * fake AWS services
    * fixes for provider/services prefix searches

* [gundeck]:
    * fake AWS services

* [generic]:
    * docker-compose to bring up all external dependencies up
    * travis builds
    * upgrade to HsOpenSSL version that supports OpenSSL 1.1


services/brig/schema/src/V47.hs
```diff
@@ -0,0 +1,20 @@
+{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE QuasiQuotes       #-}
+
+module V47 (migration) where
+
+import Cassandra.Schema
+import Text.RawString.QQ
+
+migration :: Migration
+migration = Migration 47 "Create and populate prefix table" $
```
elverkilde: This migration doesn't actually populate the table, we need to perform a manual step here.

PR #226, opened 2018-03-06
Expose 'expires_at' json field for ephemeral users

jschaul:
Ephemeral users are defined as users created with neither email nor phone
The optional field 'expires_at' is available for these users on
  - /register
  - /self (SelfProfile)
  - /users/:Id (public UserProfile)
expiry time is creation time + SessionTokenTimeout (default: 1 day). JSON example of user profile: 
```
{
   (... other user fields)
   "expires_at" : "2018-03-06T20:36:12.000Z"
}
```

### DIFF
--- a/libs/brig-types/src/Brig/Types/Intra.hs
+++ b/libs/brig-types/src/Brig/Types/Intra.hs
@@ -24,6 +24,7 @@ data AccountStatus
     = Active
     | Suspended
     | Deleted
+    | Ephemeral
     deriving Eq
 
 instance FromJSON AccountStatus where
@@ -31,12 +32,14 @@ instance FromJSON AccountStatus where
         "active"    -> pure Active
         "suspended" -> pure Suspended
         "deleted"   -> pure Deleted
+        "ephemeral" -> pure Ephemeral
         _           -> fail $ "Invalid account status: " ++ Text.unpack s
 
 instance ToJSON AccountStatus where
     toJSON Active    = String "active"
     toJSON Suspended = String "suspended"
     toJSON Deleted   = String "deleted"
+    toJSON Ephemeral = String "ephemeral"
 
 newtype AccountStatusUpdate = AccountStatusUpdate
     { suStatus :: AccountStatus }
--- a/libs/brig-types/src/Brig/Types/User.hs
+++ b/libs/brig-types/src/Brig/Types/User.hs
@@ -17,12 +17,13 @@ import Data.Aeson
 import Data.Aeson.Types (Parser, Pair)
 import Data.ByteString.Conversion
 import Data.Id
-import Data.Json.Util ((#))
+import Data.Json.Util ((#), UTCTimeMillis (..))
 import Data.Misc (PlainTextPassword (..))
 import Data.Maybe (isJust)
 import Data.Range
 import Data.Text (Text)
 import Data.Text.Ascii
+import Data.Time (UTCTime)
 import Galley.Types.Bot (ServiceRef)
 import Galley.Types.Teams hiding (userId)
 
@@ -96,6 +97,7 @@ connectedProfile u = UserProfile
     , profileService  = userService u
     , profileLocale   = Just (userLocale u)
     , profileDeleted  = userDeleted u
+    , profileExpire   = userExpire u
     }
 
 publicProfile :: User -> UserProfile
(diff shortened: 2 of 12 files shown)


libs/brig-types/src/Brig/Types/User.hs
```diff
@@ -154,6 +159,7 @@ instance ToJSON User where
         # "locale"    .= userLocale u
         # "service"   .= userService u
         # "handle"    .= userHandle u
+        # "expire"    .= (UTCTimeMillis <$> userExpire u)
```
tiago-loureiro: Could we call this `expires_at` instead?

services/brig/src/Brig/Data/User.hs
```diff
@@ -71,23 +77,37 @@ newAccount u inv = do
...
+    return (UserAccount (user uid (locale defLoc) expiry) status, passwd)
   where
     ident         = newUserIdentity u
     pass          = newUserPassword u
     name          = newUserName u
     pict          = fromMaybe noPict (newUserPict u)
     assets        = newUserAssets u
+    status        = case ident of
+                        Nothing -> -- any user registering without either an email or a phone is Ephemeral,
+                                   -- i.e. can be deleted after a short amount of time (24h)
```
tiago-loureiro: Can we say `sessionTokenTimeout` instead `24h`?

libs/brig-types/src/Brig/Types/User.hs
jschaul: We certainly can.

services/brig/src/Brig/Data/User.hs
jschaul: sure.

services/brig/Makefile
```diff
@@ -90,18 +90,18 @@ integration-list: install
 integration-%: db-migrate index-reset
 	LOG_LEVEL=Info keiretsu --run '$(EXE_IT) --pattern="$*"' --env $(KEIRETSU_ENV)
 
+
+.PHONY: fast
+fast:
```
fisx: shouldn't this be:

```
.PHONY:
fast: .PHONY
```

services/brig/test/integration/API.hs
```diff
@@ -447,6 +450,40 @@ testMultipleUsers brig = do
...
+  where
+    ensureExpiry :: Maybe UTCTime -> String -> Http ()
+    ensureExpiry expiry s = case expiry of
+                       Nothing -> liftIO $ assertFailure ("user must have an expiry" <> s)
+                       Just a -> do
+                          now <- liftIO getCurrentTime
+                          let diff = diffUTCTime a now
+                              minExp = 20 :: Integer -- 20 seconds
+                              maxExp = 60 * 60 * 24 * 10 :: Integer -- 10 days
+                          liftIO $ assertBool "expiry in the future" (diff > fromIntegral minExp)
```
fisx: [nit-pick] which way does the text go here?  should it be the one printed in case of failure, or the condition that is tested here?

libs/brig-types/src/Brig/Types/Intra.hs
```diff
@@ -24,19 +24,22 @@ data AccountStatus
     = Active
     | Suspended
     | Deleted
+    | Ephemeral
```
fisx: I would have added the ttl to the 'Ephemeral' constructor intsead of adding an extra field to the status-carrying types.

- This way an ephemeral user can never have 'Nothing' in its ttl.
- Makes 'toUsers', 'toUserAccount' easier: no need to pass a ttl if it's not needed.
- Complicates storage in cassandra a little.
- Not sure if it works at all.  Or if it's worth it.

Just an idea.  (-:

services/brig/Makefile
jschaul: I've not seen this, I thought `.PHONY: fast` only to mean the `fast` target is always out-of-date, i.e. `make fast` will run even if there is a local file with the name `fast`. That's the explanation found on [stackoverflow](https://stackoverflow.com/questions/2145590/what-is-the-purpose-of-phony-in-a-makefile) at least.

services/brig/test/integration/API.hs
jschaul: Good point, thanks!

libs/brig-types/src/Brig/Types/Intra.hs
jschaul: This might be an idea for a separate PR.

PR #230, opened 2018-03-08
Garbage collect ephemeral users on profile lookup

jschaul:


services/brig/src/Brig/API.hs
```diff
@@ -1513,3 +1509,6 @@ activate (Activate tgt code dryrun)
 
 validateHandle :: Text -> Handler Handle
 validateHandle = maybe (throwE (StdError invalidHandle)) return . parseHandle
+
+ifNothing :: Utilities.Error -> Maybe a -> Handler a
+ifNothing e = maybe (throwStd e) return
```
tiago-loureiro: We could move this to `types-common`, we already use it in [galley](https://github.com/wireapp/wire-server/blob/d256479f94815da4f351c122286ef483be966c54/services/galley/src/Galley/App.hs#L221) and [gundeck](https://github.com/wireapp/wire-server/blob/9482a5b0cb95ebae3c0c584f3bd7733148fd30bd/services/gundeck/src/Gundeck/Monad.hs#L100)

PR #231, opened 2018-03-12
Release/20180312

jschaul:
* ephemeral user / `expired_at` support & GC
* Minor refactoring in galley


jschaul:
Closing. We should add https://github.com/wireapp/wire-server/pull/238

PR #244, opened 2018-03-15
Feature/test using config files

jschaul:
This PR changes a few things:

* fixes example configuration files to work in integration tests
* service binaries and integration test binaries can now be run with configuration files, rather than environment variables. Configuration files are the new default.
* Removes the need for `keiretsu`, `Procfile`s, `Intfile`s. There is a replacement shell script for integration tests on this repository that is less generic than keiretsu was; but should be simple enough and make it a little clearer how integration tests are run.
* streamlines the various service Makefiles:
    * `make` (default target) runs `make fast`, i.e. `stack install --fast --pedantic --local-bin-path=dist`. By default we want shorter compilation times. 
    * `make integration` depends on `make fast` instead `make install` to have faster feedback cycles when testing
* new top-level make targets:
    * `make integration` ensures all binaries are in the correct location and runs all of the integration tests (for brig,galley,gundeck,cargohold)
* streamline dist folders: whether running `make` from the top-level or from within a service's directory, the binaries like `brig` and `brig-schema` and `brig-integration` will all be put under the top-level `wire-server/dist` folder. That $(DIST) location is no longer configurable. 

The actual code places using e.g. `optOrEnv brig iConf (local . read) "BRIG_WEB_PORT"` can be changed to only accept configuration files. That work should be done in a separate PR.

PR #245, opened 2018-03-15
Updated documentation on building & integration tests

jschaul:
This documentation only becomes correct after #244 is merged.


fisx:
(we probably want to update https://github.com/zinfra/backend-wiki/wiki/Development-Setup and reference the README.md rather than duplicate it?)

PR #248, opened 2018-03-18
Team id user profile

tiago-loureiro:
This PR adds a field in the `user`'s table for the binding `team_id` of that user; it exposes it over the `/users` and `/self` endpoints as part of the user's public profile.

A migration to backfill `team_id`s for existing users will be done after.

### DIFF
--- a/services/brig/src/Brig/API/User.hs
+++ b/services/brig/src/Brig/API/User.hs
@@ -241,7 +241,9 @@ createUser new@NewUser{..} = do
                                                 , Maybe (Team.Invitation, Team.InvitationInfo)
                                                 , Maybe TeamId
                                                 )
-    handleTeam (Just (NewTeamMember i))  e = findTeamInvitation e i
+    handleTeam (Just (NewTeamMember i))  e = findTeamInvitation e i >>= return . \case
+        Just (inv, info, tid) -> (Nothing, Just (inv, info), Just tid)
+        Nothing               -> (Nothing, Nothing         , Nothing)
     handleTeam (Just (NewTeamCreator t)) _ = (Just t, Nothing, ) <$> (Just . Id <$> liftIO nextRandom)
     handleTeam Nothing                   _ = return (Nothing, Nothing, Nothing)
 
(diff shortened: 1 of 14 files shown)


services/brig/src/Brig/API/User.hs
```diff
@@ -227,15 +229,21 @@ createUser new@NewUser{..} = do
...
-    handleTeam (Just (NewTeamMember i))  e = (Nothing, ) <$> findTeamInvitation e i
-    handleTeam (Just (NewTeamCreator t)) _ = return (Just t, Nothing)
-    handleTeam Nothing                   _ = return (Nothing, Nothing)
+    handleTeam :: Maybe NewTeamUser 
+               -> Maybe UserKey
+               -> ExceptT CreateUserError AppIO ( Maybe BindingNewTeamUser
+                                                , Maybe (Team.Invitation, Team.InvitationInfo)
+                                                , Maybe TeamId
+                                                )
+    handleTeam (Just (NewTeamMember i))  e = findTeamInvitation e i
```
jschaul: `(Nothing, ) <$> findTeamInvitation e i` and remove the first return type of findTeamInvitation? findTeamInvidation never returns a BindingNewTeamUser, so it's odd to have that in the type signature.

tiago-loureiro:
> Try top-level compilation also (missing line 

👍 https://github.com/wireapp/wire-server/commit/0c842160ca6aa394584677746f22c093a6c7b70a

services/brig/src/Brig/API/User.hs
tiago-loureiro: Unfortunately that's not as easy as it seems since `(Nothing, ) <$> findTeamInvitation e i` would then return `(Maybe a, (Maybe b, Maybe c))`

So I could either have a little `do` block or leave as-is, any particular preference?

jschaul: I have a preference for a do block, to keep the type signature of `findTeamInvitation` cleaner.

jschaul:
You could also change https://github.com/wireapp/wire-server/blob/develop/services/brig/src/Brig/App.hs#L121 to ensure brig does not start up without the schema migration.

services/brig/src/Brig/API/User.hs
tiago-loureiro: 👍 https://github.com/wireapp/wire-server/pull/248/commits/f0fae1f45d934b4a3d3a7ccd49866677538e813a

tiago-loureiro:
> You could also change https://github.com/wireapp/wire-server/blob/develop/services/brig/src/Brig/App.hs#L121 to ensure brig does not start up without the schema migration.

Ah yes, good point 👌 https://github.com/wireapp/wire-server/commit/a6f9d15c4447265d7cae662daa7b54b1e49a01c9

PR #252, opened 2018-03-21
New retention types and policies for assets

tiago-loureiro:
Adds other retention possibilities, depending on the desired use case. Adjusts existing retention policies.

### DIFF
--- a/services/cargohold/src/CargoHold/S3.hs
+++ b/services/cargohold/src/CargoHold/S3.hs
@@ -195,10 +195,14 @@ mkKey :: V3.AssetKey -> S3AssetKey
 mkKey (V3.AssetKeyV3 i r) = S3AssetKey $ "v3/" <> retention <> "/" <> key
   where
     key = UUID.toText (toUUID i)
+    -- This is the prefix used to store on S3, which currently is in sync
+    -- with the JSON representation but does not necessarily need to be
     retention = case r of
-        V3.AssetEternal    -> "eternal"
-        V3.AssetPersistent -> "persistent"
-        V3.AssetVolatile   -> "volatile"
+        V3.AssetEternal                 -> "eternal"
+        V3.AssetPersistent              -> "persistent"
+        V3.AssetVolatile                -> "volatile"
+        V3.AssetEternalInfrequentAccess -> "eternal-infrequent_access"
+        V3.AssetExpiring                -> "expiring"
 
 -------------------------------------------------------------------------------
 -- Resumable Uploads
--- a/services/cargohold/src/CargoHold/S3.hs
+++ b/services/cargohold/src/CargoHold/S3.hs
@@ -195,14 +195,7 @@ mkKey :: V3.AssetKey -> S3AssetKey
 mkKey (V3.AssetKeyV3 i r) = S3AssetKey $ "v3/" <> retention <> "/" <> key
   where
     key = UUID.toText (toUUID i)
-    -- This is the prefix used to store on S3, which currently is in sync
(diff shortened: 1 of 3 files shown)


services/cargohold/src/CargoHold/S3.hs
```diff
@@ -195,10 +195,14 @@ mkKey :: V3.AssetKey -> S3AssetKey
 mkKey (V3.AssetKeyV3 i r) = S3AssetKey $ "v3/" <> retention <> "/" <> key
   where
     key = UUID.toText (toUUID i)
+    -- This is the prefix used to store on S3, which currently is in sync
+    -- with the JSON representation but does not necessarily need to be
```
fisx: It would make it easier if it had to be, though, no?  I would require the two to always match and use a single function to create the strings.

services/cargohold/test/integration/API/V3.hs
```diff
@@ -76,45 +76,55 @@ tests s = testGroup "v3"
...
-    r1 <- uploadSimple (c . path "/assets/v3") uid sets bdy <!!
-        const 201 === statusCode
-
-    let      loc = decodeHeader "Location" r1 :: ByteString
-    let Just ast = decodeBody r1 :: Maybe V3.Asset
-    let Just tok = view V3.assetToken ast
+    let def  = V3.defAssetSettings
+    let rets = [ V3.AssetVolatile
+               , V3.AssetExpiring
+               , V3.AssetEternalInfrequentAccess
```
fisx: make `AssetRetention` `deriving (Bounded, Enum)` and define `rets = [minBound ..]`?

services/cargohold/test/integration/API/V3.hs
```diff
@@ -76,45 +76,55 @@ tests s = testGroup "v3"
...
+        -- Delete (allowed for creator)
+        deleteAsset c uid (view V3.assetKey ast) !!! const 200 === statusCode
+        r4 <- get (c . path loc . zUser uid . header "Asset-Token" (toByteString' tok) . noRedirect) <!!
+            const 404 === statusCode
+        let Just date' = C8.unpack <$> lookup "Date" (responseHeaders r4)
+        let utc' = parseTimeOrError False defaultTimeLocale rfc822DateFormat date' :: UTCTime
+        liftIO $ assertBool "bad date" (utc' >= utc)
 
 testSimpleTokens :: TestSignature ()
 testSimpleTokens c = do
```
fisx: When I read `simpleRoundtrip`, I expected the trivial aeson encode-decode loop and comparison.  Should this test be added for all json-deriving data types nearby, while we're at it?

services/cargohold/src/CargoHold/S3.hs
tiago-loureiro: Yeah, was about to do so :)

tiago-loureiro: https://github.com/wireapp/wire-server/pull/252/commits/7a9f01f98420e6224c266dbe4687300c699dd5d4

services/cargohold/test/integration/API/V3.hs
tiago-loureiro: Good idea https://github.com/wireapp/wire-server/pull/252/commits/dc3e9a8af5250c0d045e96a31aa23c255b4e01a3

services/cargohold/test/integration/API/V3.hs
tiago-loureiro: I think that's a good idea, let me deal with that in a separate

PR #249, opened 2018-03-20
Fetch emails from wire-emails

lipis:
### In short we are going from [this](https://github.com/wireapp/wire-server/tree/develop/services/brig/deb/opt/brig/templates) to [this](https://github.com/wireapp/wire-server/tree/new-mails/services/brig/deb/opt/brig/templates)

---

In order to fetch the latest emails/translations to the project:

1. Edit `services/brig/deb/opt/brig/new-version` with the latest/desired tag from [wire-emails](https://github.com/wireapp/wire-emails)
2. Run `python services/brig/deb/opt/brig/bin/fetch.py`
3. Commit changes

---

All email previews can be found here: https://wireapp.github.io/wire-emails/dist/

PR #269, opened 2018-04-03
Emails from wire emails

tiago-loureiro:
Replaces https://github.com/wireapp/wire-server/pull/249, LGTM


services/brig/src/Brig/Team/Template.hs
```diff
@@ -62,23 +61,24 @@ loadTeamTemplates o = readLocalesDir defLocale (templateDir gOptions) "team" $ \
...
-                <*> readText (fp <> "/email/sender.txt"))
+                <*> readText fp "email/sender.txt")
   where
     gOptions = general (emailSMS o)
     tOptions = team (emailSMS o)
     tUrl     = template $ tInvitationUrl tOptions
 
     defLocale = setDefaultLocale (optSettings o)
     readTemplate = readTemplateWithDefault (templateDir gOptions) defLocale "team"
+    readText = readTextWithDefault (templateDir gOptions) defLocale "team"
```
fisx: Looks like `readText` is defined on two scope levels?  Do we allow that?

fisx: sorry, scratch that.  `readTextWithDefault` used to be called `readText`, but it's not any more.

PR #270, opened 2018-04-03
Order TurnURI by scheme and transport

tiago-loureiro:
Change the order of how TURN servers are advertised: `/calls/config` now advertises all known URIs and orders them in a specific order: first the ones with `turn` scheme and then `UDP`

### DIFF
--- a/libs/brig-types/test/unit/Test/Brig/Types/TURN.hs
+++ b/libs/brig-types/test/unit/Test/Brig/Types/TURN.hs
@@ -2,16 +2,44 @@
 
 module Test.Brig.Types.TURN where
 
-import Brig.Types.TURN hiding (turnURI)
+import Brig.Types.TURN
 import Data.Aeson
+import Data.List (sort)
+import Data.Misc
+import System.Random.Shuffle
 import Test.Brig.Types.Arbitrary ()
 import Test.Tasty
+import Test.Tasty.HUnit
 import Test.Tasty.QuickCheck
 
 tests :: TestTree
 tests = testGroup "TURN"
     [ testProperty "TurnURI: decode . encode = id" turnURIid
+    , testGroup "Order"
+        [ testCase "Turn order example" turnURIOrder
+        ]
     ]
 
 turnURIid :: TurnURI -> Property
 turnURIid t = Just t === (decode . encode) t
+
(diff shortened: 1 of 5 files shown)


libs/brig-types/test/unit/Test/Brig/Types/TURN.hs
```diff
@@ -2,16 +2,44 @@
...
+    , testGroup "Order"
+        [ testCase "Turn order example" turnURIOrder
+        ]
     ]
 
 turnURIid :: TurnURI -> Property
 turnURIid t = Just t === (decode . encode) t
+
+turnURIOrder :: IO ()
+turnURIOrder = go (100 :: Int)
```
jschaul: Perhaps you could add a test case ensuring that shuffling a list of turnURIs with different hosts keeps the random order of the hosts, even if sorting the schemes and transports?

services/brig/src/Brig/TURN/API.hs
```diff
@@ -61,17 +60,17 @@ getCallsConfig (_ ::: _ ::: _) = json <$> lift newConfig
     newConfig = do
         env  <- liftIO =<< readIORef <$> view turnEnv
         let (sha, secret, tTTL, cTTL, prng) = (env^.turnSHA512, env^.turnSecret, env^.turnTokenTTL, env^.turnConfigTTL, env^.turnPrng)
-        uris <- liftIO $ randomize (unsafeRange 2) (env^.turnServers)
+        uris <- List1.sort <$> liftIO (randomize (env^.turnServers))
```
jschaul: The purpose of `List1.sort <$> liftIO (randomize` isn't very clear for people not knowing that we wish to sort by scheme and transport, but randomize by host and port - perhaps add a comment here?

tiago-loureiro: 👍

libs/brig-types/test/unit/Test/Brig/Types/TURN.hs
tiago-loureiro: Ah yes, good idea 👍

tiago-loureiro:
Will be replaced with a more appropriate fix: https://github.com/wireapp/wire-server/pull/273

PR #273, opened 2018-04-09
Turn v2

tiago-loureiro:
Replaces https://github.com/wireapp/wire-server/pull/270 by having a specific endpoint `/calls/v2` which will return all available `TURN URIs`; the existing `/calls/config` remains as-is, returning only endpoints with `turn` scheme and `udp` transport (with the difference that _all_ known valid URIs are now returned).

### DIFF
--- a/libs/brig-types/src/Brig/Types/TURN.hs
+++ b/libs/brig-types/src/Brig/Types/TURN.hs
@@ -47,6 +47,7 @@ import qualified Data.ByteString.Conversion as BC
 import           Data.List1
 import           Data.Misc                  (IpAddr, Port (..))
 import           Data.Monoid
+import           Data.Ord
 import           Data.Text                  (Text)
 import           Data.Text.Ascii
 import qualified Data.Text.Encoding         as TE
@@ -92,13 +93,18 @@ data TurnURI = TurnURI
     , _turiTransport :: Maybe Transport
     } deriving (Eq, Show, Generic)
 
+-- | Relies on Ord instances for `Scheme` and `Transport`.
+-- | First, prefer `turn` scheme and then prefer UDP.
+instance Ord TurnURI where
+    compare = comparing _turiScheme <> comparing _turiTransport
+
 data Scheme = SchemeTurn
             | SchemeTurns
-            deriving (Eq, Show, Generic)
+            deriving (Eq, Ord, Show, Generic)
 
-data Transport = TransportTCP
-               | TransportUDP
-               deriving (Eq, Show, Generic)
+data Transport = TransportUDP
+               | TransportTCP
+               deriving (Eq, Ord, Show, Generic)
 
 -- future versions may allow using a hostname
 newtype TurnHost = TurnHost IpAddr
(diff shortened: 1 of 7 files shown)


services/brig/src/Brig/TURN/API.hs
```diff
@@ -50,35 +51,61 @@ routes = do
         .&. header "Z-Connection"
 
     document "GET" "getCallsConfig" $ do
-        Doc.summary "Retrieve TURN server addresses and credentials."
+        Doc.summary "Retrieve TURN server addresses and credentials \
+                    \ for scheme `turn` and transport `udp` "
```
neongreen: double space :trollface: 

(and a trailing space, too)

libs/brig-types/src/Brig/Types/TURN.hs
```diff
@@ -96,8 +96,8 @@ data Scheme = SchemeTurn
             | SchemeTurns
             deriving (Eq, Show, Generic)
 
-data Transport = TransportTCP
-               | TransportUDP
+data Transport = TransportUDP
```
neongreen: Since the order is apparently important, I'd appreciate a comment here (maybe even a link to the relevant issue).

services/brig/src/Brig/TURN/API.hs
```diff
@@ -50,35 +51,61 @@ routes = do
...
+        return $ List1.list1 f fs
 
-        genUsername :: Word32 -> MWC.GenIO -> IO TurnUsername
-        genUsername ttl prng = do
-            rnd <- view (packedBytes . utf8) <$> replicateM 16 (MWC.uniformR (97, 122) prng)
-            t   <- fromIntegral . (+ ttl) . round <$> getPOSIXTime
-            pure $ turnUsername t rnd
+    genUsername :: Word32 -> MWC.GenIO -> IO TurnUsername
+    genUsername ttl prng = do
+        rnd <- view (packedBytes . utf8) <$> replicateM 16 (MWC.uniformR (97, 122) prng)
```
neongreen: `(ord 'a', ord 'z')` would've been nicer

libs/brig-types/src/Brig/Types/TURN.hs
tiago-loureiro: It _was_ important, not anymore 😊 See https://github.com/wireapp/wire-server/pull/273/commits/f7fa0877c83f9f538c06c04ae610d05b7c209ea6 and let me know if it's clearer ?

tiago-loureiro:
> Minor thing: perhaps you could add an example ?transport=.. line to turn/servers.txt?

You mean as our default? Your suggestion is tested [here](https://github.com/wireapp/wire-server/blob/develop/services/brig/test/integration/API/TURN.hs#L70)

services/brig/src/Brig/TURN/API.hs
tiago-loureiro: 👍 :)

libs/brig-types/src/Brig/Types/TURN.hs
neongreen: > ```diff haskell
> +-- | Relies on Ord instances for `Scheme` and `Transport`.
> +-- | First, prefer `turn` scheme and then prefer UDP.
> ```

It looks like it wasn't important and now it is important. So, still not clear.

jschaul:
> You mean as our default? 

exactly. I was referring to the text file `turn/servers.txt` (I understand there is a test; still, the current line's format in the servers.txt is kind of outdated)

PR #275, opened 2018-04-09
Endpoint that allows checking for set password

tiago-loureiro:
Depending on the user's state, our clients may end up in a situation where they need to ask the user to either "Enter a password" or "Set a password";

In such occasions, it would be convenient for clients to be able to check if the _self_ user has a password registered with the backend.

@deanrobertcook

### DIFF
--- a/services/brig/src/Brig/API.hs
+++ b/services/brig/src/Brig/API.hs
@@ -430,6 +430,16 @@ sitemap o = do
 
     ---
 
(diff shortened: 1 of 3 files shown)


jschaul:
Rather than a separate endpoint, couldn't this information be part of the `SelfProfile` ?

marcoconti83:
I like @jschaul 's suggestion too!

PR #276, opened 2018-04-10
arch docs

jschaul:


neongreen:
Let's also add the `icu` and `openssl` packages, it's possible that I got them by installing other things (I have them on my system)

neongreen:
done

PR #279, opened 2018-04-12
Release/20180412

tiago-loureiro:
Highlights:
 * Updated docs
 * Fixed subject on activation emails
 * Endpoint that allows self user to check whether a password has been set in the account


services/brig/src/Brig/TURN/API.hs
```diff
@@ -70,7 +70,11 @@ getCallsConfigV2 :: JSON ::: UserId ::: ConnId -> Handler Response
 getCallsConfigV2 (_ ::: _ ::: _) = json <$> lift (newConfig Nothing)
 
 getCallsConfig :: JSON ::: UserId ::: ConnId -> Handler Response
-getCallsConfig (_ ::: _ ::: _) = json <$> lift (newConfig (Just ([SchemeTurn], [Just TransportUDP, Nothing])))
+getCallsConfig (_ ::: _ ::: _) = json . dropTransport <$> lift (newConfig (Just ([SchemeTurn], [Just TransportUDP, Nothing])))
+  where
+    -- In order to avoid being backwards incompatible, remove the `transport` query param from the URIs
```
fisx: Doesn't that mean that parts of the config are silently ignored?  Wouldn't it be better to require the config to not contain transports?

tiago-loureiro: We want the `/calls/config` endpoint to return the same "type of config" that it currently does; thus, we take that into account when changing the internal representation of the config.

However, when using `/calls/config/v2`, we do want to return _all_ of those configs

PR #280, opened 2018-04-16
Split alpine-builder

neongreen:
Pros:
* The next time we want to change something about the way we build Haskell stuff, we won't have to wait for the GHC build every. single. time.

Cons:
* A split based on components (e.g. GHC in a separate docker image) would be nicer conceptually.

### DIFF
--- a/build/alpine/Dockerfile.builder
+++ b/build/alpine/Dockerfile.builder
@@ -2,10 +2,17 @@
 
 FROM wireserver/alpine-prebuilder
 
-# download stack indices and compile/cache dependencies to speed up subsequent container creation
+# Download stack indices and compile/cache dependencies to speed up subsequent
+# container creation.
+#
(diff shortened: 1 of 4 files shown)


build/alpine/Dockerfile.builder
```diff
@@ -1,104 +1,11 @@
...
 
 # download stack indices and compile/cache dependencies to speed up subsequent container creation
-# TODO: make this caching step optional?
 RUN apk add --no-cache git && \
     mkdir -p /src && cd /src && \
     git clone https://github.com/wireapp/wire-server.git && \
     cd wire-server && \
     stack update && \
-    stack build --pedantic --haddock --test --dependencies-only
+    stack build --pedantic --haddock --test --no-run-tests
```
neongreen: In addition to splitting the builder, we also build the full project now (to make sure that `extra-deps` get into the image).

PR #282, opened 2018-04-17
Pass prebuilder as an arg

neongreen:


neongreen:
I can't build it on Concourse otherwise.

PR #289, opened 2018-04-23
[WIP] Morphing Identity

neongreen:


neongreen:
This place is a message… and part of a system of messages… pay attention to it! Sending this message was important to us. We considered ourselves to be a powerful culture.

This place is not a place of honor… no highly esteemed deed is commemorated here… nothing valued is here.

What is here is dangerous and repulsive to us.

This message is a warning about danger.

PR #291, opened 2018-04-25
Break up / clean up brig test module structure

fisx:
In `/services/brig/test/integration`:

- move `API.hs` => `API/User.hs`
- split up `API/User.hs` into `API/User/*` for the test groups.
- factor out `API/User/Prelude.hs` for the common utilities.
- call `API/User/Auth.hs` tests from `API/User.hs`, not `Main.hs`.

Code has not been changed, just shuffled around.

I am proposing this because my IDE is slowing down the original file size of ~2.5k LOC.  During the change I also noticed that the structure is now more logical.  I found a test case that was located in the wrong section, and moved it the the correct sub-module.  Sub-modules are providing tests in pure functions (`TestTree` rather than `IO TestTree`).


services/brig/test/integration/API/User/Property.hs
```diff
@@ -0,0 +1,166 @@
...
+-- import Galley.Types
+-- import Gundeck.Types.Notification
+-- import Gundeck.Types.Push.V2
+-- import OpenSSL.EVP.Digest (getDigestByName, digestBS)
+import Test.Tasty hiding (Timeout)
+-- import Test.Tasty.Cannon hiding (Cannon)
+-- import Test.Tasty.HUnit
+-- import Safe hiding (at)
+-- import System.Random (randomIO)
+-- import Web.Cookie (parseSetCookie, setCookieName)
```
jschaul: remove commented out lines?

fisx: oops, thanks!

services/brig/test/integration/API/User/Util.hs
```diff
@@ -0,0 +1,341 @@
...
+toAddressBook :: [(Text, [Text])] -> IO AddressBook
+toAddressBook xs = do
+    Just sha <- liftIO $ getDigestByName "SHA256"
+    return . AddressBook $ fmap (toCard sha) xs
+  where
+    toCard sha (cardId, entries) = Card (Just $ CardId cardId)
+                                        (map (Entry . digestBS sha . T.encodeUtf8) entries)
+
+setHandleAndDeleteUser :: Brig -> Cannon -> User -> [UserId] -> (UserId -> HttpT IO ()) -> Http ()
+setHandleAndDeleteUser brig cannon u others execDelete = do
```
jschaul: This long helper function would perhaps be better placed in the only file it's used, i.e. `integration/API/User/Account.hs`

PR #292, opened 2018-04-26
sso brig integration

fisx:
Integration of sso-sp service into brig (part 1).

- new data type `UserSSOId`, can be used instead of email or phone.
- `NewTeamMemberSSO` case to `NewTeamUser` that carries a `TeamId`.
- deletion: users with email and/or phone can still delete themselves, even with a `UserSSOId`.
- deletion: sso service will not be asked or informed (does not exist yet).
- deletion: change `isOnlyOwner`: team owner definition is now "has perms *and* email".
- refactor data types `User`, `NewUser` to make some inconsistent values impossible.
- roundtrip json tests for brig-types, brig-common.

less importantly / cleanup:

- some use of `HasCallStack`.
- rm .ghci (should not be under git control)
- rm trailing whitespace.
- galley-types: introduce `Internal` modules that export constructors for opaque data types.
- un-pin QuickCheck and fix cannon tests.
- derive `Show`, `Eq` for many types


services/brig/src/Brig/API.hs
```diff
@@ -180,12 +184,15 @@ sitemap o = do
     post "/i/users/blacklist" (continue addBlacklist) $
         param "email" ||| param "phone"
 
+    get "/i/users/:uid/can-be-deleted/:tid" (continue canBeDeleted) $
```
tiago-loureiro: Why do we pass the teamID here?

services/brig/src/Brig/API/User.hs
```diff
@@ -315,6 +318,19 @@ createUser new@NewUser{..} = do
...
+    addUserToTeamSSO account tid ident = do
+        let uid = userId (accountUser account)
+        added <- lift $ Intra.addTeamMember uid tid
+        unless added $
+            throwE TooManyTeamMembers
+        lift $ do
+            activateUser uid ident
+            void $ onActivated (AccountActivated account)
+            Log.info $ field "user" (toByteString uid)
+                     . field "team_id" (toByteString tid)
```
tiago-loureiro: Please use field `team` here

services/brig/src/Brig/API/User.hs
```diff
@@ -653,29 +669,50 @@ mkPasswordResetKey ident = case ident of
...
+            ownerSituation <- lift $ Team.isOnlyTeamOwner uid tid
+            case ownerSituation of
+               Team.IsOnlyTeamOwner       -> throwE DeleteUserOnlyOwner
+               Team.IsOneOfManyTeamOwners -> pure ()
+               Team.IsNotTeamOwner        -> pure ()
+               Team.NoTeamOwnersAreLeft   -> do
+                   Log.warn $ field "user" (toByteString uid)
+                            . field "team" (toByteString tid)
+                            . msg (val "Team.NoTeamOwnersAreLeft")
+                   pure ()
```
tiago-loureiro: I don't think this `pure ()` is needed here?

services/brig/src/Brig/Team/Util.hs
```diff
@@ -1,15 +1,24 @@
...
+
+-- | A team owner is a team member with full permissions *and* an email address.
+isOnlyTeamOwner :: UserId -> TeamId -> AppIO IsOnlyTeamOwner
+isOnlyTeamOwner uid tid = isOnlyTeamOwner' uid . fmap (^. userId) <$> Intra.getTeamOwners tid
+
+isOnlyTeamOwner' :: UserId -> [UserId] -> IsOnlyTeamOwner
+isOnlyTeamOwner' _ [] = NoTeamOwnersAreLeft
+isOnlyTeamOwner' uid owners
+    | uid `notElem` owners   = IsNotTeamOwner
+    | null (owners \\ [uid]) = assert (owners == nub owners) IsOnlyTeamOwner
```
tiago-loureiro: Why this assertion?

services/brig/src/Brig/API/User.hs
```diff
@@ -191,6 +189,10 @@ createUser new@NewUser{..} = do
                 return (True, Just $ CreateUserTeam (Team.inTeam inv) nm)
         Nothing -> return (False, Nothing)
 
+    case (ident, tid) of
```
tiago-loureiro: Note that here the user also joins a team (notice the `joinedTeam` variable down below)

libs/brig-types/src/Brig/Types/Common.hs
```diff
@@ -159,37 +160,69 @@ data UserIdentity
...
 phoneIdentity (FullIdentity  _ phone) = Just phone
 phoneIdentity (PhoneIdentity   phone) = Just phone
 phoneIdentity (EmailIdentity _      ) = Nothing
+phoneIdentity (SSOIdentity _ _ (Just phone)) = Just phone
+phoneIdentity (SSOIdentity _ _ Nothing) = Nothing
+
+ssoIdentity :: UserIdentity -> Maybe UserSSOId
+ssoIdentity (SSOIdentity ssoid _ _) = Just ssoid
+ssoIdentity _ = Nothing
+
```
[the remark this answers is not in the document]
fisx: I can give `UserSSOId` the shape we want to use in spar (current working title of the wire-sso-sp-service).  Then we can move the definition to `/libs/spar-types` later once there is such a thing.

Is that what you suggested?

libs/brig-types/src/Brig/Types/User.hs
```diff
@@ -229,41 +243,82 @@ data NewUser = NewUser
...
+    result <- case (invcode, teamcode, team, ssoid, teamid) of
+        (Just a,  Nothing, Nothing, Nothing, Nothing) -> return . Just . NewUserOriginInvitationCode $ a
+        (Nothing, Just a,  Nothing, Nothing, Nothing) -> return . Just . NewUserOriginTeamUser $ NewTeamMember a
+        (Nothing, Nothing, Just a,  Nothing, Nothing) -> return . Just . NewUserOriginTeamUser $ NewTeamCreator a
+        (Nothing, Nothing, Nothing, Just _,  Just t)  -> return . Just . NewUserOriginTeamUser $ NewTeamMemberSSO t
+        (Nothing, Nothing, Nothing, Nothing, Nothing) -> return Nothing
+        (_, _, _, _, _) -> fail $ "team_code, team, invitation_code, sso_id are mutually exclusive\
+                                  \ and sso_id, team_id must be either both present or both absent."
+    case (result, pass, uid) of
+        (_, _, Just SSOIdentity {}) -> pure result
```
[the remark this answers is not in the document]
fisx: Yes, which raises an interesting question.  We have users who strongly object to their team members to go and create passwords.  Will it be enough to disable password-login for those use cases?

services/brig/src/Brig/Team/Util.hs
fisx: 'cause it's true!  (-:

Would you like me to remove it?

I'll remove it.

services/brig/src/Brig/API/User.hs
fisx: You mean I'm not doing that, but should?  Will check!

libs/brig-types/src/Brig/Types/Common.hs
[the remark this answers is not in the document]
tiago-loureiro: 👍

services/brig/src/Brig/API/User.hs
tiago-loureiro: > You mean I'm not doing that, but should? Will check!

Yes ;)

services/brig/src/Brig/API/User.hs
fisx: done, but only for logging.  json (`team_id`) and cql (`already `team`) unchanged.  is that what you wanted?

services/brig/src/Brig/API.hs
fisx: to check whether the team would be orphaned.  brig could ask galley for this, but that would mean two more network roundtrips, and it seemed to @jschaul and me that it is more elegant to provide it explicitly.

fisx:
I think I've done everything you asked.  Could you double-check again?  Thanks!  (And thanks for spotting the `joinedTeam` error...)

libs/brig-types/src/Brig/Types/Common.hs
```diff
@@ -159,37 +164,82 @@ data UserIdentity
...
+-- | TODO: once we have @/libs/spar-types@ for the wire-sso-sp-server called spar, this type should
+-- move there.
+data UserSSOId = UserSSOId { userSSOIdTenant :: UUID, userSSOIdSubject :: UUID }
+    deriving (Eq, Show)
```
tiago-loureiro: Would it make sense to define `Show` in terms of `userSSOIdToText` ?

libs/brig-types/src/Brig/Types/User.hs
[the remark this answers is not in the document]
tiago-loureiro: > Will it be enough to disable password-login for those use cases?

Let's leave all options open and then restrict things if necessary.

libs/brig-types/src/Brig/Types/User.hs
```diff
@@ -229,41 +243,82 @@ data NewUser = NewUser
...
 
 newUserPhone :: NewUser -> Maybe Phone
 newUserPhone = phoneIdentity <=< newUserIdentity
 
+newUserSSOId :: NewUser -> Maybe UserSSOId
+newUserSSOId = ssoIdentity <=< newUserIdentity
+
 instance FromJSON NewUser where
       parseJSON = withObject "new-user" $ \o -> do
+          ssoid                 <- o .:? "sso_id"
```
tiago-loureiro: Any reason not to parse `sso_id` in `parseIdentity` ?

libs/types-common/src/Data/Misc.hs
```diff
@@ -220,7 +220,7 @@ instance Cql (Fingerprint a) where
 
 newtype PlainTextPassword = PlainTextPassword
     { fromPlainTextPassword :: Text }
-    deriving (ToJSON)
+    deriving (Eq, Show, ToJSON)
```
tiago-loureiro: I'd advise against deriving a `Show` instance here, just in case a `PlainTextPassword` ends up in logs

services/brig/src/Brig/API.hs
```diff
@@ -1467,6 +1475,19 @@ addBlacklist emailOrPhone = do
...
+canBeDeleted (uid ::: tid) = do
+    onlyOwner <- lift (Team.isOnlyTeamOwner uid tid)
+    case onlyOwner of
+       Team.IsOnlyTeamOwner       -> throwStd noOtherOwner
+       Team.IsOneOfManyTeamOwners -> pure ()
+       Team.IsNotTeamOwner        -> pure ()
+       Team.NoTeamOwnersAreLeft   -> do
+           Log.warn $ Log.field "user" (toByteString uid)
+                    . Log.msg (Log.val "Team.NoTeamOwnersAreLeft")
+           pure ()
```
tiago-loureiro: Wasn't this removed at some point?

services/brig/src/Brig/API/User.hs
```diff
@@ -653,29 +674,49 @@ mkPasswordResetKey ident = case ident of
...
+            Suspended -> ensureNotOnlyOwner account >> go a
+            Active    -> ensureNotOnlyOwner account >> go a
             Ephemeral -> go a
   where
-    ensureNotOnlyOwner = do
-        onlyOwner <- lift $ Team.isOnlyTeamOwner uid
-        when onlyOwner $
-            throwE DeleteUserOnlyOwner
+    ensureNotOnlyOwner :: Maybe UserAccount -> ExceptT DeleteUserError (AppT IO) ()
+    ensureNotOnlyOwner acc = case userTeam . accountUser =<< acc of
```
tiago-loureiro: Any reason to use `Maybe UserAccount` instead of `UserAccount` ?

libs/brig-types/src/Brig/Types/User.hs
fisx: we could have parsed it in both `parseIdentity` and `parseNewUserOrigin` (further down in the same do block).  but i like the fact that it's only parsed once, so the parser only needs to be changed in one place if it has to be changed.  and i thought it is clearer that way.

see also: https://github.com/wireapp/wire-server/blob/feature/sso-brig-integration/libs/brig-types/src/Brig/Types/User.hs#L164

libs/types-common/src/Data/Misc.hs
fisx: :+1:

embarassing.  thanks for catching this!

services/brig/src/Brig/API.hs
fisx: i think this may be another one.  (-:  i'll remove that, too.

PR #293, opened 2018-04-26
Do not use newlines on email subject or sender

tiago-loureiro:
otherwise it may cause the _from_ name to look like this: `From: =?utf-8?Q?Wire=0A?= <...@wire.com>`, leading to delivery failures in some providers.


neongreen:
```bash
$ for s in $( fd | grep subject ) ; do ag "\n" "$s"; done
$ for s in $( fd | grep sender ) ; do ag "\n" "$s"; done
$
```

neongreen:
Would be nice to have a (slightly more rigorous) CI check for this, but no pressure

jschaul:
Let's add a test for this to avoid potential regressions here.

PR #295, opened 2018-04-26
Release/20180426

tiago-loureiro:
Highlights:
 * Endpoint to check password
 * Separate endpoints for TURN v2 (`/calls/config/v2`)
 * Fixes for templates
 * Some test code refactoring


fisx:
failing test in  3c14d7d, but it doesn't appear to be related (gundeck fails, nothing but brig was touched):

```
        serialise/ok:               FAIL (0.03s)
          *** Failed! Falsifiable (after 58 tests): 
          APNSSandbox
          Use --quickcheck-replay=977893 to reproduce.
```

tiago-loureiro:
> Also I assume the email newline fix happens in another repo?

Yes, I will also add a new commit to strip newlines on `subject/from`... just in case.

PR #296, opened 2018-04-27
Ensure user assets get deleted

tiago-loureiro:


services/brig/test/integration/Util.hs
```diff
@@ -46,9 +46,10 @@ import qualified Data.UUID as UUID
 import qualified Data.UUID.V4 as UUID
 import qualified Test.Tasty.Cannon as WS
 
-type Brig    = Request -> Request
-type Cannon  = Request -> Request
-type Galley  = Request -> Request
+type Brig      = Request -> Request
+type Cannon    = Request -> Request
+type CargoHold = Request -> Request
```
jschaul: For consistency: Use `Cargohold` (with lowercase `h`)?

services/brig/brig.integration.yaml
```diff
@@ -12,6 +12,10 @@ elasticsearch:
   url: http://127.0.0.1:9200
   index: directory_test
 
+cargohold:
+  host: 127.0.0.1
+  port: 8084
```
jschaul: Could you update the diagram from https://github.com/wireapp/wire-server/tree/develop/doc/arch using e.g. https://www.draw.io/ to reflect that brig needs routing to cargohold?

tiago-loureiro: 👍

services/brig/test/integration/Util.hs
tiago-loureiro: We actually use camelcase in some cases as in `import ... CargoHold.Types.V3` since cargo hold is made up of 2 words. We do use it as `cargohold` in a few places though (binary, package-name, cargohold as a variable...)

So I think it actually makes sense to keep it camelcase here? We can then for consistency's sake indeed either move everything to camelcase or rename `CargoHold` to `Cargohold` everywhere.

services/brig/brig.integration.yaml
tiago-loureiro: https://github.com/wireapp/wire-server/pull/296/commits/afac44a67437d1d9e089ec6bf8ee84cbea386a3d

services/brig/test/integration/Util.hs
jschaul: Ok, nevermind. I hadn't seen the `import ... CargoHold.Types.V3`. So it's consistent already. All good!

PR #300, opened 2018-04-30
In Brig, read max conv size from CLI options in all places

neongreen:


neongreen:
ЮБИЛЕЙНЫЙ

neongreen:
Here are all remaining 128s, you can browse through them to see if I missed anything: https://gist.github.com/neongreen/50ddd7708c9ed453af1c530d76ec36e1.

PR #303, opened 2018-05-02
Fix makefile

fisx:


neongreen:
I can approve this one if you explain what exactly the change in `Makefile` accomplished because I don't know a thing about makefiles

fisx:
The licence check appears to be hanging?

PR #312, opened 2018-05-08
Guide `make fast` behavior during development via shell env.

fisx:
Tweak the `make fast` rule such that it lets you ignore warnings during development without risking to accidentally commit your compiler settings:

```shell
$ WIRE_STACK_OPTIONS="--ghc-options=-Wwarn" make fast
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,7 +14,7 @@ install: init
 
 .PHONY: fast
 fast: init
-	stack install --pedantic --test --local-bin-path=dist --fast
+	stack install --pedantic --test --local-bin-path=dist --fast $(WIRE_STACK_OPTIONS)
 
 .PHONY: clean
 clean:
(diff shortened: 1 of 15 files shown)


fisx:
Yes, `extra-dep` does that.  This is ok as long as the ci runs `make fast` from the top level at some point (or some other goal that builds all stack deps), as the libs are still non-extra deps there.

fisx:
Some more thoughts:

- `--no-run-tests` is also a good candidate for adding to `$WIRE_STACK_OPTIONS`.  This may be a good compromise for the discussion in #305.  I'd certainly be happy with it.  (-:
- About the changes to `stack.yaml`: stack will still re-build brig-types if you build brig and something changed in brig-types.  A nice side-effect of the `extra-dep` flag is that the modules from brig-types won't be interpreted into ghci any more.

PR #315, opened 2018-05-09
Allow arbitrary awk patterns in tasty integration tests

fisx:


fisx:
This is just to demonstrate it works in brig.  If there is a consensus that this is a good idea, I will make sure all integration tests work the same way.  Would be a small & fast change!

services/brig/test/integration/Main.hs
```diff
@@ -70,7 +71,13 @@ runTests iConf bConf = do
     teamApis    <- Team.tests bConf mg b c g
     turnApi     <- TURN.tests mg b turnFile
 
-    defaultMain $ testGroup "Brig API Integration"
+    -- you can now do this (see <https://github.com/feuerbach/tasty#patterns>):
+    -- `WIRE_TASTY_PATTERN='$NF == "post /register - 201 + no email"' make integration`
+    otherArgs' <- lookup "WIRE_TASTY_PATTERN" <$> getEnvironment >>= pure . \case
```
neongreen: This seems to be the same as `lookupEnv "WIRE_TASTY_PATTERN" <&> \case ...`

neongreen:
Seems ok to me. Is there some place where this could be documented and made discoverable? Currently it's just a comment in the code.

fisx:
> Is there some place where this could be documented and made discoverable?

very good point.  should we start `/doc/Development.md` and reference it from `/README.md`?

services/brig/test/integration/Main.hs
fisx: cool.  i've been meaning to find that thing, thanks!  will fix.

fisx:
Ok, this is how I would like to merge it.  Any more concerns?

PR #319, opened 2018-05-14
Add a tool for generating users

neongreen:
I noticed that `users.txt` from S3 was out of date (I couldn't login with any of the users from there) so I wrote a tool for mass-registering users.


neongreen:
Though I'm not sure we actually need it, given that https://github.com/wireapp/wire-server/blob/develop/deploy/services-demo/create_test_user.sh exists

fisx:
Is there anything this can do that `post /i/users/` can't?  Would we need that?  If no I don't think we should add this.

neongreen:
#323

PR #321, opened 2018-05-14
Fix compilation with GCC 8

### LINKED ISSUE #320: Doesn't compile with GCC 8.1.0
```
src/http/ngx_http_script.c: In function ‘ngx_http_script_add_full_name_code’:
src/http/ngx_http_script.c:1296:18: error: cast between incompatible function types from ‘size_t (*)(ngx_http_script_engine_t *)’ {aka ‘long unsigned int (*)(struct <anonymous> *)’} to ‘void (*)(ngx_http_script_engine_t *)’ {aka ‘void (*)(struct <anonymous> *)’} [-Werror=cast-function-type]
     code->code = (ngx_http_script_code_pt) ngx_http_script_full_name_len_code;
                  ^
cc1: all warnings being treated as errors
make[3]: *** [objs/Makefile:947: objs/src/http/ngx_http_script.o] Error 1
make[3]: Leaving directory '/home/yom/wire/server/services/nginz/src'```

neongreen:
Fixes #320


neongreen:
No, but it's the same thing that the original patch uses, and presumably they care about supporting older versions of GCC :slightly_smiling_face: If I'm wrong, CI will tell me that.

neongreen:
Hm, though it doesn't look like CI actually builds nginx

neongreen:
Okay, I checked and it compiles with GCC 6

neongreen:
plz approve

neongreen:
Okay, when nginx 1.14.1 comes out I guess we can upgrade to that.

PR #322, opened 2018-05-14
Journal user events

tiago-loureiro:
Allow to optionally journal the following events over protobuf to SQS:
 * User activation
 * User deletion
 * User updates (for locale and email)

### DIFF
--- a/deploy/services-demo/conf/brig.demo.yaml
+++ b/deploy/services-demo/conf/brig.demo.yaml
@@ -27,7 +27,6 @@ gundeck:
 aws:
   sesQueue: integration-brig-events
   internalQueue: integration-brig-events-internal
-  journalQueue: integration-user-events
   blacklistTable: integration-brig-userkey-blacklist
   prekeyTable: integration-brig-prekeys
   sesEndpoint: http://localhost:4569 # https://email.eu-west-1.amazonaws.com
--- a/deploy/services-demo/conf/brig.demo.yaml
+++ b/deploy/services-demo/conf/brig.demo.yaml
@@ -27,9 +27,10 @@ gundeck:
 aws:
   sesQueue: integration-brig-events
   internalQueue: integration-brig-events-internal
+  journalQueue: integration-user-events
   blacklistTable: integration-brig-userkey-blacklist
   prekeyTable: integration-brig-prekeys
-  sesEndpoint: https://email.eu-west-1.amazonaws.com # http://localhost:4569 # https://email.eu-west-1.amazonaws.com
+  sesEndpoint: http://localhost:4569 # https://email.eu-west-1.amazonaws.com
   sqsEndpoint: http://localhost:4568 # https://sqs.eu-west-1.amazonaws.com
   dynamoDBEndpoint: http://localhost:4567 # https://dynamodb.eu-west-1.amazonaws.com
 emailSMS:
--- a/libs/types-common-journal/proto/UserEvents.proto
+++ b/libs/types-common-journal/proto/UserEvents.proto
@@ -7,22 +7,13 @@ message UserEvent {
     required EventType event_type = 1;
     required bytes user_id = 2;
     required int64 utc_time = 3;
-    optional ConsentData consent = 4;
-    optional bytes email = 5;
-    optional bytes phone = 6;
-
-    enum ConsentType {
-        MARKETING = 0;
-    }
-
-    message ConsentData {
-        required ConsentType consentType = 1;
-        required bool allow = 2;
-    }
+    optional bytes email = 4;
+    optional bytes locale = 5;
 
     enum EventType {
-        USER_ACTIVATE = 0;
-        USER_DELETE   = 1;
-        USER_UPDATE   = 2;
+        USER_ACTIVATE     = 0;
+        USER_DELETE       = 1;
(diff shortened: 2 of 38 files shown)


libs/types-common-journal/proto/UserEvents.proto
```diff
@@ -0,0 +1,24 @@
...
+    enum EventType {
+        USER_ACTIVATE     = 0;
+        USER_DELETE       = 1;
+        USER_UPDATE       = 2;
+        // --^ When used, the new values for optional fields are
+        //     whatever is set (email, locale) unless they are 
+        //     null / not set (protobuf does not distinguish)
+        USER_EMAIL_REMOVE = 3;
+        // --^ When an email is removed, the email field is set
+        //     to the value of the old email address
```
jschaul: > the email field is set to the value of the old email address

User A registers with email; then adds a phone number, then removes his email (all valid operations w.r.t the API). There would be no "old value" for anyone to set. This comment is not very clear.

deploy/services-demo/conf/brig.demo.yaml
```diff
@@ -27,9 +27,10 @@ gundeck:
 aws:
   sesQueue: integration-brig-events
   internalQueue: integration-brig-events-internal
+  journalQueue: integration-user-events
```
jschaul: journaling is not needed in demo-mode, let's comment it out here.

tiago-loureiro: 👍

libs/types-common-journal/proto/UserEvents.proto
tiago-loureiro: In that case, the value of the old email (the one that the user just removed) will be on the event https://github.com/wireapp/wire-server/blob/develop/services/brig/src/Brig/API/User.hs#L416

services/brig/test/integration/Util/AWS.hs
```diff
@@ -0,0 +1,62 @@
...
+isRealSESEnv env = case view AWS.journalQueue env of
+    Just url | "amazonaws.com" `Text.isSuffixOf` url -> True
+    _                                                -> False
+
+purgeJournalQueue :: AWS.Env -> IO ()
+purgeJournalQueue env = for_ (view AWS.journalQueue env)
+                      $ SQS.execute (view AWS.amazonkaEnv env)
+                      . SQS.purgeQueue
+
+assertEmptyUserQueue :: AWS.Env -> IO ()
```
fisx: Why `UserQueue`, not `JournalQueue`?

fisx: (same for `assertUserQueue`.)

services/brig/test/integration/Main.hs
```diff
@@ -64,8 +67,11 @@ runTests iConf bConf otherArgs = do
     lg <- Logger.new Logger.defSettings
     db <- defInitCassandra casKey casHost casPort lg
     mg <- newManager tlsManagerSettings
-
-    userApi     <- User.tests bConf mg b c ch g =<< mkLocalAWSEnv lg awsOpts mg
+    awsEnv <- AWS.mkEnv lg awsOpts mg
+    
+    -- If we're journaling, make sure we have a clean slate
+    for_ (awsEnv^.AWS.journalQueue) $ SQS.execute (view AWS.amazonkaEnv awsEnv) . SQS.purgeQueue
```
fisx: this is happening before each test, too, so it should not be strictly necessary here, no?  but i don't mind if it remains there.

libs/types-common-journal/proto/UserEvents.proto
fisx: what about `when an email address X is removed, the email field is set to X`?

services/brig/src/Brig/IO/Intra.hs
```diff
@@ -143,6 +146,15 @@ updateSearchIndex orig e = case e of
...
 
+journalEvent :: UserId -> UserEvent -> AppIO ()
+journalEvent orig e = case e of
+    UserActivated acc                 -> Journal.userActivate (accountUser acc)
+    UserLocaleUpdated _ loc           -> Journal.userUpdate orig Nothing (Just loc)
+    UserIdentityUpdated _ (Just em) _ -> Journal.userUpdate orig (Just em) Nothing
+    UserIdentityRemoved _ (Just em) _ -> Journal.userEmailRemove orig em
+    UserDeleted{}                     -> Journal.userDelete orig
+    _                                 -> return ()
+
```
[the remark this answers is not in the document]
tiago-loureiro: > This means we have email addresses of not-activated users piling up in brig that the journal does not know about.

Yes, I think we should not journal unverified email addresses; once they get verified, we will journal them (this also avoids tracking, on our end, email addresses that are not verified/that you do not own). Note that unverified email/phone number get automatically deleted after [a certain amount of time](https://github.com/wireapp/wire-server/blob/develop/services/brig/src/Brig/Options.hs#L342-L345)

libs/types-common-journal/proto/UserEvents.proto
tiago-loureiro: > what about when an email address X is removed, the email field is set to X?

👍

PR #323, opened 2018-05-14
Extend the create_test_user.sh script

neongreen:
* can create many users
* can create users in `users.txt` format

### DIFF
--- a/deploy/services-demo/create_test_user.sh
+++ b/deploy/services-demo/create_test_user.sh
@@ -1,16 +1,56 @@
 #!/usr/bin/env bash
 
+set -e
+
 #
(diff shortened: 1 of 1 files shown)


neongreen:
Btw, I think I should also update the `users.txt` file I generated

neongreen:
@tiago-loureiro 

> Btw, I think I should also update the users.txt file I generated

is it okay if I just upload it to S3? i.e. I have a `users.txt` file with valid staging users, while the existing one is invalid

tiago-loureiro:
> is it okay if I just upload it to S3? i.e. I have a users.txt file with valid staging users, while the existing one is invalid

Definitely 👍

PR #328, opened 2018-05-17
Fix docker nginz build

neongreen:
I checked locally and it works

### DIFF
--- a/libs/libzauth/libzauth-c/Makefile
+++ b/libs/libzauth/libzauth-c/Makefile
@@ -4,8 +4,8 @@ ARCH    := amd64
 BUILD   ?= 1
 OS      := $(shell uname -s | tr '[:upper:]' '[:lower:]')
 # If we can install libzauth globally, we'll install it there, otherwise
-# it'll go into ~/.wire-dev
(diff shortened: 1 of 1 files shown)


tiago-loureiro:
Can we instead make the following change:
```
-# it'll go into ~/.wire-dev
-PREFIX  := $(shell [ -w /usr/local ] && echo /usr/local || echo "$(HOME)/.wire-dev")
+# it'll go into ~/.wire-dev (unless it's overriden)
+PREFIX  ?= $(shell [ -w /usr/local ] && echo /usr/local || echo "$(HOME)/.wire-dev")
```
in https://github.com/wireapp/wire-server/blob/develop/libs/libzauth/libzauth-c/Makefile#L8 ?

Notice that we are setting the `PREFIX` to be `/usr` in [nginz's Dockefile](https://github.com/wireapp/wire-server/blob/develop/services/nginz/Dockerfile#L7)

It should also work locally for you (it does for me)

neongreen:
@tiago-loureiro @tiago-loureiro @tiago-loureiro 

image

PR #329, opened 2018-05-18
Backdoor login

neongreen:
This adds an internal endpoint that lets you login as any (existing, non-suspended) user.

### DIFF
--- a/libs/brig-types/src/Brig/Types/Swagger.hs
+++ b/libs/brig-types/src/Brig/Types/Swagger.hs
@@ -559,7 +559,7 @@ login = defineModel "Login" $ do
         description "The login code for an SMS login."
         optional
     property "label" string' $ do
-        description "A label to associate with the returned a cookie. \
+        description "A label to associate with the returned cookie. \
                     \Every client should have a unique and stable (persistent) label \
                     \to allow targeted revocation of all cookies granted to that \
                     \specific client."
@@ -568,17 +568,10 @@ login = defineModel "Login" $ do
 backdoorLogin :: Model
 backdoorLogin = defineModel "BackdoorLogin" $ do
     description "Payload for performing a credential-less login."
-    property "email" string' $ do
-        description "Email address"
-        optional
-    property "phone" string' $ do
-        description "Phone number"
-        optional
-    property "handle" string' $ do
-        description "User handle"
-        optional
+    property "user" bytes' $
+        description "User ID"
     property "label" string' $ do
-        description "A label to associate with the returned a cookie. \
+        description "A label to associate with the returned cookie. \
                     \Every client should have a unique and stable (persistent) label \
                     \to allow targeted revocation of all cookies granted to that \
                     \specific client."
--- a/libs/brig-types/src/Brig/Types/Swagger.hs
+++ b/libs/brig-types/src/Brig/Types/Swagger.hs
@@ -565,6 +565,25 @@ login = defineModel "Login" $ do
                     \specific client."
         optional
 
+backdoorLogin :: Model
+backdoorLogin = defineModel "BackdoorLogin" $ do
+    description "Payload for performing a credential-less login."
+    property "email" string' $ do
+        description "Email address"
+        optional
+    property "phone" string' $ do
+        description "Phone number"
+        optional
+    property "handle" string' $ do
+        description "User handle"
+        optional
+    property "label" string' $ do
+        description "A label to associate with the returned a cookie. \
+                    \Every client should have a unique and stable (persistent) label \
+                    \to allow targeted revocation of all cookies granted to that \
+                    \specific client."
+        optional
+
 accessToken :: Model
 accessToken = defineModel "AccessToken" $ do
     description "An API access token."
--- a/libs/brig-types/src/Brig/Types/User/Auth.hs
+++ b/libs/brig-types/src/Brig/Types/User/Auth.hs
@@ -1,6 +1,7 @@
 {-# LANGUAGE FlexibleInstances          #-}
 {-# LANGUAGE GeneralizedNewtypeDeriving #-}
 {-# LANGUAGE OverloadedStrings          #-}
+{-# LANGUAGE LambdaCase                 #-}
 
 module Brig.Types.User.Auth where
 
@@ -17,6 +18,8 @@ import Data.Text (Text)
 import Data.Text.Lazy.Encoding (decodeUtf8, encodeUtf8)
 import Data.Time.Clock (UTCTime)
 import Data.Word
+
+import qualified Data.Aeson.Types as Aeson
 
 -----------------------------------------------------------------------------
 -- Login / Authentication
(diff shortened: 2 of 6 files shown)


neongreen:
(Hold your pitchforks, 112 people who watch this repository! This is something we could always do, and it doesn't break the security properties in any way. This specific endpoint is required to implement single sign-on.)

libs/brig-types/src/Brig/Types/User/Auth.hs
```diff
@@ -43,10 +46,15 @@ newtype LoginCodeTimeout = LoginCodeTimeout
...
     deriving (Eq, Show)
 
+-- | Different kinds of logins.
 data Login
     = PasswordLogin !LoginId !PlainTextPassword !(Maybe CookieLabel)
     | SmsLogin !Phone !LoginCode !(Maybe CookieLabel)
 
+-- | A special kind of login that is only used for an internal endpoint.
+data BackdoorLogin
+    = BackdoorLogin !LoginId !(Maybe CookieLabel)
```
fisx: `UserId` would be enough here.

services/brig/src/Brig/User/API/Auth.hs
```diff
@@ -145,6 +145,28 @@ routes = do
 
     -- Internal
 
+    post "/i/users/backdoor-login" (continue backdoorLogin) $
```
fisx: Why not `/i/login`, in line with `/login`?  `/i/backdoor-login` would be ok with me, too.  @tiago-loureiro?

services/brig/test/integration/API/User/Auth.hs
```diff
@@ -90,30 +95,18 @@ testEmailLogin brig = do
...
-        assertBool "expiry" (cookie_expiry_time ck > cookie_creation_time ck)
-        assertBool "domain" (cookie_domain ck /= "")
-        assertBool "path" (cookie_path ck /= "")
-
-    -- Check basic access token attributes
-    let tk = decodeToken  _rs
     liftIO $ do
-        assertEqual "user" (userId u) (ZAuth.accessTokenOf tk)
-        assertBool "expiry" (ZAuth.tokenExpiresUTC tk > now)
+        assertSanePersistentCookie (decodeCookie _rs)
```
fisx: If we are using `_rs`, it should be renamed to `rs`, no?

fisx: (same below)

services/brig/test/integration/API/User/Auth.hs
```diff
@@ -237,6 +230,46 @@ testThrottleLogins conf b = do
...
+    backdoorLogin brig loginSpec PersistentCookie
+        !!! const 403 === statusCode
+
+-- | Check that @/backdoor-login@ fails if the user doesn't exist.
+testNoUserBackdoorLogin :: Brig -> Http ()
+testNoUserBackdoorLogin brig = do
+    -- Try to login as wrong@wire.com and see if we fail
+    let loginSpec = BackdoorLogin (LoginByEmail (Email "wrong" "wire.com")) Nothing
+    backdoorLogin brig loginSpec PersistentCookie
+        !!! const 403 === statusCode
```
fisx: we need to have a way of distinguishing `user does not exist` from `something else happened`, so that we can decide whether to create the user and try again.  this should also be tested here.

libs/brig-types/src/Brig/Types/User/Auth.hs
neongreen: Are you saying that `CookieLabel` is not needed for your purposes, or is that it is actually bad to have here? If it's just the former then I would rather leave it in place.

services/brig/test/integration/API/User/Auth.hs
neongreen: It's used in about 30 places in code. I think in those cases it's supposed to mean "this is temporary" and not "this is unused", but I agree that it's slightly weird.

services/brig/src/Brig/User/API/Auth.hs
neongreen: Fixed.

services/brig/test/integration/API/User/Auth.hs
neongreen: You can do that by inspecting the `"label"` in the response. I added explicit tests for that.

libs/brig-types/src/Brig/Types/User/Auth.hs
fisx: sorry, i meant: `LoginId` is too much, `UserId` would be better instead.  emails and phone numbers should not be used to identify users on this end-point for now.  we certainly need the cookie label.

services/brig/test/integration/API/User/Auth.hs
fisx: :+1:

libs/brig-types/src/Brig/Types/Swagger.hs
```diff
@@ -565,6 +565,25 @@ login = defineModel "Login" $ do
...
+        description "Email address"
+        optional
+    property "phone" string' $ do
+        description "Phone number"
+        optional
+    property "handle" string' $ do
+        description "User handle"
+        optional
+    property "label" string' $ do
+        description "A label to associate with the returned a cookie. \
```
fisx: this probably should read: `A label to associate with the returned a cookie.`?

services/brig/src/Brig/User/Auth.hs
```diff
@@ -181,3 +183,14 @@ validateTokens ut at = do
                 unless (e == ZAuth.Expired) (throwE e)
     ck <- lift (lookupCookie ut) >>= maybe (throwE ZAuth.Invalid) return
     return (ZAuth.userTokenOf ut, ck)
+
+-- | Allow to login as any user without having the credentials.
+backdoorLogin :: BackdoorLogin -> CookieType -> ExceptT LoginError AppIO Access
+backdoorLogin (BackdoorLogin uid label) typ = do
+    let pw = PlainTextPassword ""
+    Data.authenticate uid pw `catchE` \case
```
tiago-loureiro: How about instead using `Data.reauthenticate` with `Nothing` as the password and catching `ReAuthMissingPassword` instead?

fisx: why?  to me this makes semantically less sense by far.

neongreen: It makes less sense semantically but doesn't require "making up" a password. I can't say that I like either choice much

fisx: (i'm ok with this change.  it's not an improvement in my opinion, but it's still perfectly readable and correct.)

neongreen: Maybe the rightest thing is to change `authenticate` to accept `Nothing`

tiago-loureiro: I think that if we want to be explicit, it'd be better to have a different function altogether and call it `authenticateInternal` or something like that, that explicitly does not check for anything else other than the checks that we want to do; I just think that passing an empty string as a password felt off

PR #332, opened 2018-05-18
Delete all 'stack.yaml's

neongreen:
Building each individual service is possible by doing `stack build <service name>`. Not having separate `stack.yaml`s makes things simpler and apparently less buggy. (For instance, now `stack test` followed by `stack build` doesn't rebuild 21 packages. Change detection also works better.)

This PR also changes `make integration` to avoid running things twice:

```
.PHONY: integration
integration: fast
	# We run "i" instead of "integration" to avoid useless rebuilds
	# (since after "fast" everything will be built already)
	$(MAKE) -C services/cargohold i
	$(MAKE) -C services/galley i
	$(MAKE) -C services/brig i
	$(MAKE) -C services/gundeck i-fake-aws
```

All the `i`s were used to be `integration`s (`make integration = make fast && make i`).

I'm not 100% sure that this PR doesn't break Docker.


neongreen:
NB. I think I realize now what was the problem with re-compilation: contrary to what people often expect, Stack creates `.stack-work` in all package directories, even when there's only one `stack.yaml`. So, two different `stack.yaml`s were trying to claim the same `.stack-work` directory.

fisx:
Nice!

I haven't thought about the implications for ci / ops, but please consider this approved by me.

Do I have to delete `.stack-work` anywhere / everywhere?  (Does not look like it.)

neongreen:
You don't have to delete anything AFAIK

neongreen:
It uses some old version of the source, apparently:

```
bash-4.4# find -name 'stack.yaml'
./.stack-work/downloaded/dyD7MDUbZ9a6/stack.yaml
./.stack-work/downloaded/zOjCGa44_NLM/stack.yaml
./.stack-work/downloaded/L6gqNECf0paG/stack.yaml
./libs/http-client-openssl-ext/stack.yaml
./libs/sodium-crypto-sign/stack.yaml
./libs/metrics-core/stack.yaml
```

neongreen:
And this, apparently, happens because /src/wire-server exists already :unamused:

services/nginz/Makefile
```diff
@@ -100,7 +100,8 @@ src: $(TMPDIR)/$(NGINX_BUNDLE)
 	#Find keys on https://nginx.org/en/pgp_keys.html
 	gpg --verify $(TMPDIR)/$(NGINX_BUNDLE).asc $(TMPDIR)/$(NGINX_BUNDLE)
 	tar zxf $(TMPDIR)/$(NGINX_BUNDLE)
-	mv -vn nginx-$(NGINX_VERSION) src
+	rm -rf src
```
tiago-loureiro: Doesn't this cause nginz to be built every time we do `make services`? 🤔

neongreen:
@tiago-loureiro I have rewritten the makefile – now it doesn't recompile nginx. The downside is that now nginx and swagger tarballs are lying around (because the makefile lists them as "inputs"), but it doesn't seem to be a very bad thing.

neongreen:
Merging because the only failing test is the `throttle` one, which is fixed in develop. If my nginx makefile is wrong, please yell.

PR #333, opened 2018-05-22
Release/20180522

tiago-loureiro:
Highlights: 
 * Better instructions for using fake AWS services / fixes for simulations CLI
 * Ensure profile assets get deleted on profile deletion
 * Minio instead of fake_s3 for fake AWS
 * Journal user events
 * Fix compilation with GCC 8


tiago-loureiro:
To be replaced with slightly better naming

PR #336, opened 2018-05-22
Never user unsafe async

neongreen:
This doesn't change any behavior whatsoever. Just a precaution.


neongreen:
Okay, looks like it's just a flaky test.

Issue #339, opened 2018-05-23
Error make services


sd166:
When performing the  "**make services**", an error occurs.

```
make services
mkdir -p dist
stack install --pedantic --test --local-bin-path=dist
bonanza-3.6.0: test (suite: bonanza-tests)
brig-types-1.35.0: test (suite: brig-types-tests)
gundeck-1.45.0: test (suite: gundeck-tests)
types-common-0.16.0: test (suite: tests)
zauth-0.10.3: test (suite: zauth-unit)
cannon-0.31.0: test (suite: cannon-tests)
types-common-0.16.0: Test suite tests passed
zauth-0.10.3: Test suite zauth-unit passed
brig-types-1.35.0: Test suite brig-types-tests failed
gundeck-1.45.0: Test suite gundeck-tests passed
cannon-0.31.0: Test suite cannon-tests passed
bonanza-3.6.0: Test suite bonanza-tests passed
Completed 6 action(s).
Log files have been written to: /root/wire-server/.stack-work/logs/
Test suite failure for package brig-types-1.35.0
    brig-types-tests:  exited with: ExitFailure 1
Full log available at /root/wire-server/.stack-work/logs/brig-types-1.35.0-test.log


    Tests
      Common (types vs. aeson)
        Handle:                        OK (0.38s)
          +++ OK, passed 100 tests.
        Name:                          OK (0.71s)
          +++ OK, passed 100 tests.
        ColourId:                      OK (0.15s)
          +++ OK, passed 100 tests.
        Email:                         OK (0.80s)
          +++ OK, passed 100 tests.
        Phone:                         OK (0.31s)
          +++ OK, passed 100 tests.
        UserIdentity:                  OK (0.71s)
          +++ OK, passed 100 tests.
        UserSSOId:                     OK (0.16s)
          +++ OK, passed 100 tests.
        AssetSize:                     OK (0.15s)
          +++ OK, passed 100 tests.
        Asset:                         OK (0.34s)
          +++ OK, passed 100 tests.
      TURN
        TurnURI: decode . encode = id: OK (3.33s)
          +++ OK, passed 100 tests.
      User (types vs. aeson)
        parseIdentity
          FullIdentity:                OK (0.04s)
          EmailIdentity:               OK (0.04s)
          PhoneIdentity:               OK
          SSOIdentity:                 OK
          Bad phone:                   OK (0.18s)
          Bad email:                   OK (0.03s)
          Nothing:                     OK
[...]

PR #340, opened 2018-05-23
Fix ToJSON and arbitrary instance for PlainTextPassword

tiago-loureiro:
Fix for https://github.com/wireapp/wire-server/issues/339


tiago-loureiro:
Both the `FromJSON` and `ToJSON` instances are required since different services use it (e.g., brig reading client input or galley passing through a password in case of re-authentication.

>  I'd expect that the argument that applies to "Show PlainTextPassword"

I'd argue that the `Show` instance is not necessary and should probably be avoided if we prefer that scenario (I am all in favour of that); as for the others they indeed need to stay

fisx:
Ok, sorry about that.

I would like to see a test though that breaks when ToJSON does that.

(It's possible that we have such a test I didn't run it.  I was hoping for the ci, but the ci does not always appear to be running.)

fisx:
(Please do not delay merging, though.  I'm happy to add the test elsewhere later if it is indeed missing.)

PR #344, opened 2018-05-24
Add a new, more efficient connections-status endpoint

neongreen:
The old endpoint (`GET /i/users/connections-status`) sinned in two ways:

  * It took the list of users as a query parameter. This doesn't fly for
    256 users.

  * It gave N^2 responses (connections between all users), while most of the
    time we only need to know connection status for connections from one
    user to a bunch of other users.

The new `POST /i/users/connections-status` endpoint is pure as snow
in comparison.

### DIFF
--- a/services/galley/src/Galley/API/Util.hs
+++ b/services/galley/src/Galley/API/Util.hs
@@ -11,6 +11,7 @@ import Control.Lens (view, (&), (.~))
 import Control.Monad
 import Control.Monad.Catch
 import Control.Monad.IO.Class
+import Control.Concurrent.Async.Lifted.Safe
 import Data.ByteString.Conversion
 import Data.Id
 import Data.Foldable (find, for_, toList)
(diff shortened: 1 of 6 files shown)


neongreen:
WIP don't review until I fix a bug

services/galley/src/Galley/API/Util.hs
```diff
@@ -48,22 +48,18 @@ ensureAccessRole role users mbTms = case role of
...
+-- The connection has to be bidirectional (e.g. if A connects to B and later
+-- B blocks A, the status of A-to-B is still 'Accepted' but it doesn't mean
+-- that they are connected).
 ensureConnected :: UserId -> [UserId] -> Galley ()
 ensureConnected _ []   = pure ()
 ensureConnected u uids = do
-    conns <- getConnections u uids (Just Accepted)
-    unless (all (isConnected u conns) uids) $
+    connsFrom <- getConnections [u] uids (Just Accepted)
+    connsTo   <- getConnections uids [u] (Just Accepted)
```
tiago-loureiro: I think we could/should do these 2 requests concurrently here (FWIW, the first request will be pretty fast, the 2nd might take longer) or even move this logic to brig

PR #346, opened 2018-05-24
cleanup, fix tasty interface

fisx:


neongreen:
basically yesterday I discovered that you can use `TASTY_PATTERN` and it works already, so `WIRE_TASTY_PATTERN` is not even needed (I think)

and I was going to make a PR about that but I should've told you first and I forgot

neongreen:
> Every option can be passed via environment. To obtain the environment variable name from the option name, replace hyphens `-` with underscores `_`, capitalize all letters, and prepend `TASTY_`. For example, the environment equivalent of `--smallcheck-depth` is `TASTY_SMALLCHECK_DEPTH`.

fisx:
yes, that would have saved some time.  (-:

no matter.  i'll work this into this one.  there is still the README that i think is concise enough to be justified, yet very useful.  i've consulted it several times already.

fisx:
(It is ok to keep the `WIRE_STACK_OPTIONS`, though?)

neongreen:
(Yep.)

fisx:
There is still some noise that I can remove if it makes anybody happy, but I would like to merge it as is, ideally.

neongreen:
You could also mention `TASTY_PATTERN` or give a link to the official guide, it seems simpler than `--test-arguments=`

fisx:
thanks, done.

Issue #341, opened 2018-05-24
Test failure: Gundeck -> Main -> Native -> serialise/ok


sd166:
When performing the "make integration", an error occurs.

```
make integration
mkdir -p dist
stack install --pedantic --test --local-bin-path=dist --fast
bonanza-3.6.0: test (suite: bonanza-tests)
cannon-0.31.0: test (suite: cannon-tests)
gundeck-1.45.0: test (suite: gundeck-tests)
types-common-0.16.0: test (suite: tests)
zauth-0.10.3: test (suite: zauth-unit)
brig-types-1.35.0: test (suite: brig-types-tests)
types-common-0.16.0: Test suite tests passed
zauth-0.10.3: Test suite zauth-unit passed
brig-types-1.35.0: Test suite brig-types-tests passed
gundeck-1.45.0: Test suite gundeck-tests failed
cannon-0.31.0: Test suite cannon-tests passed
bonanza-3.6.0: Test suite bonanza-tests passed
Completed 6 action(s).
Log files have been written to: /root/wire-server/.stack-work/logs/
Test suite failure for package gundeck-1.45.0
    gundeck-tests:  exited with: ExitFailure 1
Full log available at /root/wire-server/.stack-work/logs/gundeck-1.45.0-test.log


    Main
      Native
        serialise/ok:               FAIL (0.36s)
          *** Failed! Falsifiable (after 65 tests):
          APNSSandbox
          Use --quickcheck-replay=215561 to reproduce.
        serialise/size-limit:       OK (0.49s)
          +++ OK, passed 100 tests.
        crypto/block-size-16:       OK (0.03s)
          +++ OK, passed 100 tests.
        crypto/aes-decrypt:         OK (0.03s)
          +++ OK, passed 100 tests.
        crypto/encrypt-then-mac:    OK (0.04s)
          +++ OK, passed 100 tests.
      DelayQueue
        enqueue/limit:              OK (0.03s)
          +++ OK, passed 100 tests.
        enqueue/unique:             OK
          +++ OK, passed 100 tests.
        enqueue/cancel:             OK
          +++ OK, passed 100 tests.
        dequeue/delay:              OK (42.54s)
          +++ OK, passed 100 tests.
        dequeue/order:              OK
          +++ OK, passed 100 tests.
      JSON
        push/recipient:             OK (0.07s)
          +++ OK, passed 100 tests.
        BulkPush
          PushTarget example:       OK
          BulkPushRequest example:  OK
[...]

PR #348, opened 2018-05-25
Document #341.

fisx:
This does not fix this issue, but sheds some light on it in case it pops up again in the future.

### DIFF
--- a/services/gundeck/test/unit/Native.hs
+++ b/services/gundeck/test/unit/Native.hs
@@ -50,8 +50,8 @@ serialiseOkProp t = ioProperty $ do
     c <- aes256
     d <- sha256
     m <- randMessage n c d
-    r <- either (const Nothing) Just <$> serialise m a
-    let sn = r >>= decode' . LT.encodeUtf8
+    r <- serialise m a
+    let sn = either (const Nothing) Just r >>= decode' . LT.encodeUtf8
     let equalTransport = fmap snsNotifTransport sn == Just t
     equalNotif <- case snsNotifBundle <$> sn of
         Nothing                -> return False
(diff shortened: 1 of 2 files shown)


services/gundeck/test/unit/Native.hs
```diff
@@ -32,8 +32,10 @@ import qualified OpenSSL.EVP.Digest      as EVP
 tests :: TestTree
 tests = testGroup "Native"
     [ testProperty "serialise/ok" $
+        -- this may fail sporadically, but that's not a production issue.
+        -- see <https://github.com/wireapp/wire-server/issues/341>.
         forAll genTransport serialiseOkProp
-    , testProperty "serialise/size-limit" $
+      testProperty "serialise/size-limit" $
```
neongreen: Accidental?

services/gundeck/test/unit/Native.hs
```diff
@@ -63,7 +65,8 @@ serialiseOkProp t = ioProperty $ do
             plain <- EVP.cipherBS c (encKeyBytes encKey) iv EVP.Decrypt dat'
             let n' = plainNotif <$> decodeStrict' plain
             return $ n' == Just n && mac == cipherMac p
-    return $ equalTransport && equalNotif
+    let debugInfo = (t, a, n, {- c, d, m, -} r, sn, equalTransport, equalNotif)
```
neongreen: Consider adding variable names, e.g.

```haskell
let debugInfo = intercalate ", " $ zipWith (\name var -> name ++ "=" ++ var)
      [ ("something something", t)
      , ("dunno what", a)
      , ... ]
return $ counterexample debugInfo (equalTransport && equalNotif)
```

fisx: i'd rather not spend the time.  this only happens if a test case fails, in which case we need to take a closer look at the source anyway.

services/gundeck/test/unit/Native.hs
fisx: yes, that's broken.  will fix.  thanks!

tiago-loureiro:
@fisx I updated the branch to get the flaky limit test in

neongreen:
Let's merge this in?

PR #350, opened 2018-05-28
Fixed flaky cookie limit test

tiago-loureiro:
This test relied on the fact that subsequent logins could be made after one second (thus the call to `wait`, which pauses for a second). This value though will not be sufficient if `Retry-After` is greater than 1 second.

I have modified the test so that it waits for `Retry-After` before logging in again and also run in parallel both tests for `Persistent` and `Session` cookies.

In order for this test to finish, we need to wait roughly for:
 * `setUserCookieLimit` logins, attempted once per second
 * `RetryAfter` * `value of carry` in the test

With a `RetryAfter` of 5 seconds, `setUserCookieLimit` of 5 and `carry = 2`, the test takes about 20 seconds to finish.

### DIFF
--- a/services/brig/test/integration/API/User/Auth.hs
+++ b/services/brig/test/integration/API/User/Auth.hs
@@ -17,6 +17,7 @@ import Control.Monad.IO.Class
 import Control.Monad.Trans.Control
 import Data.Aeson
 import Data.Aeson.Lens
+import Data.ByteString (intercalate)
 import Data.ByteString.Conversion
 import Data.Id
 import Data.List (sort, partition)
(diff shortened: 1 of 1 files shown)


services/brig/test/integration/API/User/Auth.hs
```diff
@@ -540,11 +558,18 @@ getCookieId c = maybe (error "no cookie value")
...
 
 listCookies :: HasCallStack => Brig -> UserId -> Http [Auth.Cookie ()]
-listCookies b u = do
-    rs <- get (b . path "/cookies" . header "Z-User" (toByteString' u)) <!!
+listCookies b u = listCookiesWithLabel b u []
+
+listCookiesWithLabel :: HasCallStack => Brig -> UserId -> [CookieLabel] -> Http [Auth.Cookie ()]
+listCookiesWithLabel b u l = do
+    rs <- get (b . path "/cookies"
+                 . queryItem "labels" labels
```
neongreen: Does everything still work if `labels` is empty?

tiago-loureiro: Yup, that's the [default](https://github.com/wireapp/wire-server/blob/develop/services/brig/src/Brig/User/Auth/Cookie.hs#L138)

PR #351, opened 2018-05-29
Extend loadtest

neongreen:
* Changed command-line syntax: instead of `-min` and `-max` options, there is now a range parser.
* Added the `extended` library, containing `optparse-applicative` and `async` utilities.
* More fine-grained control over conversations: you can specify how many bots will/will not be sending messages. Useful when testing team size (sending 256^2 messages is bad)
* Added `--parallel-requests` for specifying the maximum number of requests that can happen in parallel (per conversation)
* Set user-agent for all requests originating from api-client
* Added `--clients` for testing with more than one client per bot

### DIFF
--- a/libs/extended/src/Control/Concurrent/Async/Lifted/Safe/Extended.hs
+++ b/libs/extended/src/Control/Concurrent/Async/Lifted/Safe/Extended.hs
@@ -1,0 +1,48 @@
+{-# LANGUAGE TypeOperators       #-}
+{-# LANGUAGE FlexibleContexts    #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+
+-- | A version of "Control.Concurrent.Async.Lifted.Safe" with extra utilities.
+module Control.Concurrent.Async.Lifted.Safe.Extended
+    ( module Control.Concurrent.Async.Lifted.Safe
+    -- * Pooled functions (using at most T threads)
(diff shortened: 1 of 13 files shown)


neongreen:
I will give candy, make tea, or bring an exquisite paper napkin from the kitchen to whoever reviews this PR (optionally with me standing over the prospective reviewer's shoulder and explaining everything)

libs/extended/src/Control/Concurrent/Async/Lifted/Safe/Extended.hs
```diff
@@ -0,0 +1,48 @@
...
+    -> Int
+    -> m a
+    -> m [a]
+replicatePooled t n = sequencePooled t . replicate n
+
+-- | A concurrent variant of 'sequence' that uses at most T threads.
+sequencePooled
+    :: forall m a . (MonadBaseControl IO m, Forall (Pure m))
+    => Int -> [m a] -> m [a]
+sequencePooled t xs =
```
tiago-loureiro: I briefly remember that @neongreen mentioned there was a bug with `Async.Pool` that got fixed recently, what was it?

tiago-loureiro: Never mind, already read your comment on the `stack.yaml` file

PR #358, opened 2018-06-05
Route geocode requests to googlemaps too

tiago-loureiro:
i.e., requests to `[...]/proxy/googlemaps/maps/api/geocode/json?latlng=0,0` will be proxied to `maps.googleapis.com/maps/api/geocode/json?key=...&latlng=0,0`

TODO: It'd be nicer if we could match on a prefix like `/googlemaps` and route any request in the form of `/googlemaps/maps/api/:a/:b/:c/...` to `maps.googleapis.com/maps/api/:a/:b/:c`

### DIFF
--- a/services/proxy/src/Proxy/API.hs
+++ b/services/proxy/src/Proxy/API.hs
@@ -68,6 +68,10 @@ sitemap e = do
         (proxy e "key" "secrets.googlemaps" Static "/maps/api/staticmap" googleMaps)
         return
 
+    get "/proxy/googlemaps/maps/api/geocode/:path"
+        (proxy e "key" "secrets.googlemaps" Prefix "/maps/api/geocode" googleMaps)
+        return
+
     get "/proxy/giphy/v1/gifs/:path"
         (proxy e "api_key" "secrets.giphy" Prefix "/v1/gifs" giphy)
         return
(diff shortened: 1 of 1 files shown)


services/proxy/src/Proxy/API.hs
```diff
@@ -68,6 +68,10 @@ sitemap e = do
         (proxy e "key" "secrets.googlemaps" Static "/maps/api/staticmap" googleMaps)
         return
 
+    get "/proxy/googlemaps/maps/api/geocode/:path"
+        (proxy e "key" "secrets.googlemaps" Prefix "/maps/api/geocode" googleMaps)
```
jschaul: Change to either

```haskell
    get "/proxy/googlemaps/maps/:path"
        (proxy e "key" "secrets.googlemaps" Prefix "/maps" googleMaps)
```
(using `Prefix`)

or 

```haskell
    get "/proxy/googlemaps/maps/api/geocode/:path"
        (proxy e "key" "secrets.googlemaps" Static "/maps/api/geocode" googleMaps)
```
(using `Static`)

Alternatively, ask clients to call `[...]/proxy/googlemaps/api/geocode/json?latlng=0,0` without the `/maps/` segment (which is consistent with the inconsistent existing googlemaps/api/static endpoint but inconsistent with regards to the other endpoints) and combine the two `/proxy/googlemaps/`.

PR #359, opened 2018-06-05
Fix: wait for dynamo to come up before attempting to delete stale tables.

fisx:
rebased and tested on two machines.  it seems to work at least as good as before, and i could not reproduce the race condition that motivated this.

### DIFF
--- a/deploy/docker-ephemeral/init.sh
+++ b/deploy/docker-ephemeral/init.sh
@@ -1,4 +1,4 @@
-#!/usr/bin/env sh
+#!/usr/bin/env bash
 
 exec_until_ready() {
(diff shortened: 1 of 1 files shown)


fisx:
still testing this locally, but it takes so long...

fisx:
"no space left on device"

fisx:
not sure this is working.

fisx:
[see changed description]

fisx:
ah, ci now does not run any more if the changes are not in the code that is tested.  nice!

fisx:
[shellcheck](http://hackage.haskell.org/package/ShellCheck) also likes the change.

PR #360, opened 2018-06-05
Tweak user sso id type

fisx:
Relax UserSSOId type to contain any strings (not just UUID).

Store the values as json blobs on cassandra.

Introduces string-conversions, one of my favorite packages.

### DIFF
--- a/libs/brig-types/src/Brig/Types/Common.hs
+++ b/libs/brig-types/src/Brig/Types/Common.hs
@@ -185,7 +185,7 @@ instance ToJSON UserIdentity where
         SSOIdentity si em ph -> go em        ph        (Just si)
       where
         go :: Maybe Email -> Maybe Phone -> Maybe UserSSOId -> Value
-        go em ph si = object $ ["email" .= em, "phone" .= ph, "sso_id" .= si]
+        go em ph si = object ["email" .= em, "phone" .= ph, "sso_id" .= si]
 
 newIdentity :: Maybe Email -> Maybe Phone -> Maybe UserSSOId -> Maybe UserIdentity
 newIdentity email    phone    (Just sso) = Just $! SSOIdentity sso email phone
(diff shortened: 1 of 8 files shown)


libs/brig-types/src/Brig/Types/Common.hs
```diff
@@ -214,33 +211,16 @@ ssoIdentity _ = Nothing
...
-    case Text.splitAt ix raw & _2 %~ Text.splitAt 1 of
-        (tenant, (":", subject)) -> case (fromText tenant, fromText subject) of
-            (Nothing, _)      -> Left "UserSSOId: tenant"
-            (Just _, Nothing) -> Left "UserSSOId: subject"
-            (Just t, Just s)  -> pure $ UserSSOId t s
-        bad -> Left $ "UserSSOId: impossible: " <> show bad
-
-userSSOIdToText :: UserSSOId -> Text
-userSSOIdToText (UserSSOId tenant subject) = toText tenant <> ":" <> toText subject
+data UserSSOId = UserSSOId { userSSOIdTenant :: Text, userSSOIdSubject :: Text }
```
tiago-loureiro: So `Tenant` and `Subject` used to be `UUID`s... is that no longer true?

fisx: correct.  tenant is a URL (except for in the centrify example, there it is just a name, which violates the specs).  subject is a NameID, which may be a URI or a UUID or something else, with 3 optional qualifying namespace-URIs that all have their own meaning.

this is not great, but UUID didn't work out with the sample xml snippets i've found.

fisx: - Uniqueness of subject inside the tenant namespace is the IdP's responsibility.
- Uniqueness of the tenant is our responsibility.  Currently, we only have a config file and need to verify this manually.  As soon as we are going to start storing the IdP information in cql, we will need to make sure the newly created IdPs do not delete old ones.  If anybody tries, they will get an error.

is that a good plan?

PR #362, opened 2018-06-06
Turn hostnames

tiago-loureiro:
This PR adds support for hostnames in TURN URIs by allowing hostnames in the form of either `IPv4` addresses or _valid hostnames_. To be noted that _valid hostnames_ also _include_ `IPv4` addresses.

`TurnHost` can only be constructed using `parseTurnHost` (unless you import an `Internal` module) and it tries to first parse a given TURN address as an IP address; if that fails, the fallback is to parse it as a [valid hostname](https://en.wikipedia.org/wiki/Hostname#Restrictions_on_valid_hostnames).

`/calls/config` remains as-(is/was); `/calls/config/v2` was not used until now so we have changed that endpoint to advertise turn uri that have a valid hostname.

The turn file then may contain both plain IP address and hostnames; at the application level, we filter `TurnHost` based on `isHostName`

### DIFF
--- a/libs/brig-types/brig-types.cabal
+++ b/libs/brig-types/brig-types.cabal
@@ -31,6 +31,7 @@ library
         Brig.Types.Swagger
         Brig.Types.Team.Invitation
         Brig.Types.TURN
+        Brig.Types.TURN.Internal
         Brig.Types.User
         Brig.Types.User.Auth
 
--- a/libs/brig-types/brig-types.cabal
+++ b/libs/brig-types/brig-types.cabal
@@ -48,6 +48,7 @@ library
       , galley-types          >= 0.45.7
       , errors                >= 1.4
       , hashable
+      , hostname-validate
       , iproute               >= 1.5
       , iso3166-country-codes >= 0.2
       , iso639                >= 0.1
@@ -90,6 +91,7 @@ test-suite brig-types-tests
       , bytestring
       , currency-codes
       , galley-types
+      , hostname-validate
       , iproute
       , iso639
       , lens
--- a/libs/brig-types/src/Brig/Types/TURN.hs
+++ b/libs/brig-types/src/Brig/Types/TURN.hs
@@ -25,8 +25,7 @@ module Brig.Types.TURN
     , Transport (..)
 
     , TurnHost (..)
-    , _TurnHost
-
+    , isHostName
     , TurnUsername
     , turnUsername
     , tuExpiresAt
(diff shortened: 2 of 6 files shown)


tiago-loureiro:
Make some changes that significantly simplify things: we rely on 2 different files to advertise TURN servers.
V1 will advertise servers with their IPv4 address and no transport: it expects the file to have the format `turn:<IPv4 address>:<port`.
V2 will advertise servers using their hostname and (potentially) transport: it expects the file to have the URIs in any of the formats described [here](https://tools.ietf.org/html/rfc7065)

Also, the config changes need to go in before this is merged: after this PR, we already assume that `TURN_SERVERS` -> `TURN_SERVERS_V1` (also in the config).
We could keep the old name around, etc., but I think it's fine since we just need to keep the build in sync with the config.

services/brig/deb/etc/sv/brig/run
```diff
@@ -125,7 +125,8 @@ exec chpst -u $USER \
     --zauth-provider-token-timeout=${BRIG_ZAUTH_PROVIDER_TOKEN_TIMEOUT?'unset'} \
     --elasticsearch-url=${BRIG_ELASTICSEARCH_URL?'unset'} \
     --elasticsearch-user-index=${BRIG_ELASTICSEARCH_USER_INDEX?'unset'} \
-    --turn-servers=${BRIG_TURN_SERVERS?'unset'} \
+    --turn-servers-v1=${BRIG_TURN_SERVERS_V1?'unset'} \
+    --turn-servers-v2=${BRIG_TURN_SERVERS_V2?'unset'} \
```
jschaul: How about
```
    --turn-servers=${BRIG_TURN_SERVERS?'unset'} \
    --turn-servers-v2=${BRIG_TURN_SERVERS_V2?'unset'} \
```
(consistent with endpoint naming), to avoid having to lock the configuration commit to the code commit? (Otherwise, earlier config won't work with newer code and newer config won't work with earlier code - making any kind of rollback impossible)

services/brig/brig.integration.yaml
```diff
@@ -73,7 +73,8 @@ zauth:
 
 turn:
   # Replace this with the list of supported URIs
-  servers: test/resources/turn/servers.txt
+  serversV1: test/resources/turn/servers-v1.txt
+  serversV2: test/resources/turn/servers-v2.txt
```
jschaul: Let's use `servers` and `serversV2` (consistent with endpoint naming) to avoid having to lock the configuration commit to the code commit?

PR #366, opened 2018-06-07
Make UTCTimeMillis abstract and fix Eq, Show instances.

fisx:
`UTCTimeMillis` is currently used like this:

```haskell
data User = User { ..., timestamp :: UTCTime, ... }

instance ToJSON User where
    toJSON u = object [ ..., "timestamp" .= UTCTimeMillis (timestamp u), ... ]
```

This makes sure that the client will never see more than 3 decimal digits in time stamps.  The representation on the server side is not ideal, though:

1. The roundtrip-property `decode . encode === id` will fail on all timestamps with more digits.
2. The law that if `show x == show y` then `x == y` is broken.

This PR makes the data type opaque and adds a smart constructor and `Eq`, `Show`, `{To,From}JSON` instances that satisfy the above laws.

The implementation of `(==)` is not very fast, but I'm not sure it's worth optimizing?  I think we are using this only for serialization anyway.

### DIFF
--- a/libs/types-common/src/Data/Json/Util.hs
+++ b/libs/types-common/src/Data/Json/Util.hs
@@ -34,7 +34,6 @@ import Data.Text (pack)
 import qualified Data.Text.Encoding
 import qualified Data.Text.Encoding.Error
 import GHC.Generics
-import GHC.Prim
 import GHC.Stack
 
 append :: Pair -> [Pair] -> [Pair]
--- a/libs/types-common/src/Data/Json/Util.hs
+++ b/libs/types-common/src/Data/Json/Util.hs
@@ -20,7 +20,6 @@ import qualified Data.ByteString.Lazy.Char8 as L8
 import qualified Data.ByteString.Base64.Lazy as EL
 import Data.Char (isUpper)
 import Data.Fixed
-import Data.Maybe (fromMaybe)
 import Data.String
 import Data.Time.Clock
 import Data.Time.Format (formatTime, parseTimeM)
@@ -55,7 +54,10 @@ newtype UTCTimeMillis = UTCTimeMillis { fromUTCTimeMillis :: UTCTime }
 
 {-# INLINE toUTCTimeMillis #-}
 toUTCTimeMillis :: HasCallStack => UTCTime -> UTCTimeMillis
-toUTCTimeMillis = UTCTimeMillis . (TL.seconds %~ MkFixed . (* (10^9)) . (`div` (10^9)) . coerce @Pico @Integer)
+toUTCTimeMillis = UTCTimeMillis . (TL.seconds %~ MkFixed . roundit . coerce @Pico @Integer)
+  where
+    roundit = (* onebillion) . (`div` onebillion)
+    onebillion = 10 ^ (9 :: Integer)
 
 {-# INLINE showUTCTimeMillis #-}
 showUTCTimeMillis :: UTCTimeMillis -> String
--- a/libs/types-common/src/Data/Json/Util.hs
+++ b/libs/types-common/src/Data/Json/Util.hs
@@ -1,7 +1,8 @@
 {-# LANGUAGE DeriveGeneric        #-}
 {-# LANGUAGE FlexibleInstances    #-}
-{-# LANGUAGE TypeSynonymInstances #-}
+{-# LANGUAGE NumDecimals          #-}
 {-# LANGUAGE TypeApplications     #-}
+{-# LANGUAGE TypeSynonymInstances #-}
 
 module Data.Json.Util
     ( append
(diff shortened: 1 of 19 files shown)


neongreen:
Can we make `toUTCTimeMillis` return `Value` and thus get rid of `UTCTimeMillis` altogether? Or is it still useful for something?

fisx:
So, that was a great idea, but it breaks json rountrip on the types containing `UTCTimeMillis` values.

I am inclined to revert 4c460aa for that reason.

fisx:
Actually i take that back: the roundtrip property was broken before, and i think we should actually keep `UTCTimeMillis` values in types like `User`.  @tiago-loureiro I think you were opposed to that a while back?  I forgot your reasoning, though.

fisx:
now i think i am happy with this, but i should probably also go through everything again next week.  there might be conversions between rounded and un-rounded time that could be moved to better places, not sure.

fisx:
ready for review!  please only read the total changes, no the individual commits.

fisx:
@jschaul wanna take a look at concourse with me?

libs/types-common/src/Data/Json/Util.hs
```diff
@@ -42,18 +48,37 @@ infixr 5 #
...
 newtype UTCTimeMillis = UTCTimeMillis { fromUTCTimeMillis :: UTCTime }
-  deriving (Eq)
 
-instance Show UTCTimeMillis where
-    showsPrec d = showParen (d > 10) . showString . showUTCTimeMillis
+{-# INLINE toUTCTimeMillis #-}
+toUTCTimeMillis :: HasCallStack => UTCTime -> UTCTimeMillis
+toUTCTimeMillis = UTCTimeMillis . (TL.seconds %~ MkFixed . roundit . coerce @Pico @Integer)
+  where
+    roundit = (* onebillion) . (`div` onebillion)
```
neongreen: Fun fact: there is an extension (`-XNumDecimals`) that lets stuff like `1e9` have integral types.

libs/types-common/src/Data/Json/Util.hs
```diff
@@ -42,18 +48,37 @@ infixr 5 #
...
+-- Construct values using 'toUTCTimeMillis'; deconstruct with 'fromUTCTimeMillis'.
+-- Unlike with 'UTCTime', 'Show' renders ISO string.
 newtype UTCTimeMillis = UTCTimeMillis { fromUTCTimeMillis :: UTCTime }
-  deriving (Eq)
 
-instance Show UTCTimeMillis where
-    showsPrec d = showParen (d > 10) . showString . showUTCTimeMillis
+{-# INLINE toUTCTimeMillis #-}
+toUTCTimeMillis :: HasCallStack => UTCTime -> UTCTimeMillis
+toUTCTimeMillis = UTCTimeMillis . (TL.seconds %~ MkFixed . roundit . coerce @Pico @Integer)
```
neongreen: Heh, I feel like there should be an easier way to do this but I can't think of it off the top of my head :/

libs/types-common/src/Data/Json/Util.hs
```diff
@@ -42,18 +48,37 @@ infixr 5 #
...
+showUTCTimeMillis (UTCTimeMillis t) = formatTime defaultTimeLocale formatUTCTimeMillis t
+
+readUTCTimeMillis :: String -> Maybe UTCTimeMillis
+readUTCTimeMillis = fmap toUTCTimeMillis . parseTimeM True defaultTimeLocale formatUTCTimeMillis
+
+formatUTCTimeMillis :: String
+formatUTCTimeMillis = "%FT%T%QZ"
+
+{-# INLINE eqUTCTimeMillis #-}
+-- (this implementation is not very fast)
```
neongreen: Wouldn't it be better to do something like this?

``` haskell
eqUTCTimeMillis = (==) `on` (roundToMillis . fromUTCTimeMillis)
```

Where `roundToMillis :: UTCTime -> UTCTime` would be an internal helper function.

services/brig/src/Brig/Data/Connection.hs
```diff
@@ -27,19 +27,20 @@ import Control.Monad.IO.Class
...
 connectUsers from to = do
-    now <- liftIO getCurrentTime
+    now <- toUTCTimeMillis <$> liftIO getCurrentTime
     retry x5 $ batch $ do
         setType BatchLogged
         setConsistency Quorum
         forM_ to $ \(u, c) -> do
-            addPrepQuery connectionInsert (from, u, Accepted, now, Nothing, c)
-            addPrepQuery connectionInsert (u, from, Accepted, now, Nothing, c)
+            addPrepQuery connectionInsert (from, u, Accepted, fromUTCTimeMillis now, Nothing, c)
```
neongreen: Perhaps we could have an instance of `Cql` for `UTCTimeMillis`?

fisx: https://github.com/wireapp/wire-server/pull/366/commits/bb3c2559843bc3b661c54052395a430f38293e65

libs/types-common/src/Data/Json/Util.hs
fisx: https://github.com/wireapp/wire-server/pull/366/commits/8f68b934da298907c3bdb3a5a2c6de2f01aeb688

libs/types-common/src/Data/Json/Util.hs
```diff
@@ -42,18 +49,35 @@ infixr 5 #
 
 -- | A newtype wrapper for 'UTCTime' that formats timestamps in JSON with
 -- millisecond precision instead of the default picosecond precision.
+-- Construct values using 'toUTCTimeMillis'; deconstruct with 'fromUTCTimeMillis'.
+-- Unlike with 'UTCTime', 'Show' renders ISO string.
+--
+-- It would be nice to have a Cql instance for this type to make storing slightly more
+-- straight-forward.  This would require cassandra-utils as a dependency here, as both brig and
+-- cassandra (and possibly others in the future) are using this type.
```
jschaul: You can surround Cql instances with 

```haskell
{-# LANGUAGE CPP                        #-}
#ifdef WITH_CQL
...
#endif
```
As is already done in other modules of types-common, e.g. `types-common/src/Data/Id.hs`

Cassandra, by default, stores timestamps in microseconds.

PR #369, opened 2018-06-11
Add pkg-config of MacOS(which is indeed needed) to Dependencies.md

JacobChengZhang:


tiago-loureiro:
Thank you @JacobChengZhang

PR #371, opened 2018-06-12
Fix: client parser require timestamps to have exactly 3 millisec digits.

fisx:


libs/types-common/src/Data/Json/Util.hs
```diff
@@ -63,7 +63,10 @@ toUTCTimeMillis = UTCTimeMillis . (TL.seconds . coerced @Pico @_ @Integer %~ (*
 
 {-# INLINE showUTCTimeMillis #-}
 showUTCTimeMillis :: UTCTimeMillis -> String
-showUTCTimeMillis (UTCTimeMillis t) = formatTime defaultTimeLocale formatUTCTimeMillis t
+showUTCTimeMillis (UTCTimeMillis t) = formatTime defaultTimeLocale format t
```
neongreen: I think you can write it like this:

```haskell
showUTCTimeMillis (UTCTimeMillis t) = formatTime defaultTimeLocale "%FT%T.%03qZ" t
```

fisx: thanks!  (-:

PR #372, opened 2018-06-12
Switch from SQS to ActiveMQ for the internal event queue [partially done]

neongreen:
## Code

* [x] **Add a timeout to `enqueue`.**
  There's a 500ms timeout and six tries in total.

* [x] **Add a timeout to `listen`.**
  Both reading from the queue and sending an ACK now have a 1s timeout. Establishing the connection also has a 1s timeout.

* [x] **Put a timeout on `onEvent`.**
  `onEvent` has a 1min timeout. When the handler times out, an exception will be thrown and picked up by `listen`. When `listen` catches an exception, it will reestablish the connection and get a new message to process. After failing on the same message several times, the message will go into the Dead Letter Queue.

* [x] **Log exceptions in `listen`.**
  They are logged by providing a custom list of handlers to `recovering`.

* [ ] **Allow several instances (for failover).**
  Amazon MQ doesn't do failover automatically – instead it just provides two URLs and when one breaks, another starts working. We have to allow having two URLs in the config and support failover when one of them stops working.

## Checks

* [ ] **Check that the backend can't hold up a particular task indefinitely.**
  This has to be [configured server-side](http://blog.klocwork.com/open-source/activemq-to-abort-or-ackabort/). In addition, we should make sure that the backend doesn't process tasks in parallel, because the server-side configuration is "time since last ACK", and if the backend is stuck at one task while the rest are doing fine, it will not get marked as a slow consumer.

* [ ] **Check that ActiveMQ will put tasks into a Dead Letter Queue after several tries.**
  This has to be [configured server-side](http://activemq.apache.org/message-redelivery-and-dlq-handling.html#MessageRedeliveryandDLQHandling-BrokerRedelivery(v5.7)).

* [x] **Check that ActiveMQ retains tasks when it's restarted.**
  With `persistent:true` in the message header, it does. I checked that with local ActiveMQ it works.

* [ ] **Check that failover works.**

## Migration from SQS

* [ ] **Remove SQS-related code.**

* [ ] **Remove SQS mocks from docker-ephemeral.**

* [ ] **Delete SQS queues.**

## Other stuff

* [x] **Figure out whether heartbeats should be enabled.**
  It looks like we don't need them (connections will stay open for a minute) and by default they are not sent.

* [ ] **Set up notifications when anything ends up in the dead letter queue.**

## Bureaucracy

* [x] **Keep going with bureaucracy.**
  I think I've made quite a bureaucracy here.


neongreen:
This can be reviewed now. The tests pass with both SQS and STOMP.

neongreen:
By the way, I decided on a somewhat different config file format after all, hope that's okay

neongreen:
CLI options are not backwards-compatible; maybe that's why tests fail? @tiago-loureiro should I just add some extra code to handle both old and new CLI options, or is there some other way to fix this?

tiago-loureiro:
> CLI options are not backwards-compatible; maybe that's why tests fail? @tiago-loureiro should I just add some extra code to handle both old and new CLI options, or is there some other way to fix this?


The tests fail because the yaml config is not compatible; the CLI options will cause a different problem, which we will see on staging only. My suggestion is that we replicate the queue value, namely:
`BRIG_AWS_INTERNAL_QUEUE` and `BRIG_INTERNAL_EVENTS_QUEUE` should have the same value.
Also,
`--stomp-credentials=${BRIG_STOMP_CREDENTIALS?'unset'} ` is optional so I think using `unset` here will cause `brig` not to start. Feel free to change yaml config to the configuration you are suggesting here (I haven't had the chance to double check that)

tiago-loureiro:
We merged https://github.com/wireapp/wire-server/pull/442 in instead, this remains for hist{e,o}rical reasons

PR #373, opened 2018-06-12
Upgrade to LTS-11.13

tiago-loureiro:
It's mostly working (most tests pass, but requires a bit more fiddling with `http-client-openssl-ext` which still has an `undefined` but I'd like to open this up so we get some early feedback from the CI


neongreen:
Might become mergeable after we solve the TLS problem (we need to perform an action when a connection is established, but only once per connection). The current idea for the workaround is to fork `http-client` and make `withConnection` tell us whether a connection is fresh or not.

Also see https://github.com/snoyberg/http-client/issues/163.

neongreen:
> tools/api-simulations/users-staging.txt

not sure it should be here

neongreen:
https://github.com/wireapp/wire-server/pull/373/commits/673f88da6764813c5ac14bd575c0c5395654724c -- using forked http-client. It compiles but I haven't run it.

neongreen:
The tests pass but no idea if it means anything.

neongreen:
TODO: delete all your code

neongreen:
hehehe

services/brig/test/integration/API/Provider.hs
```diff
@@ -1101,38 +1141,42 @@ waitFor t f ma = do
 
 -- | Run a test case with an external service application.
 withTestService
```
neongreen: I split this function into two but didn't make any logic changes

neongreen:
Okay, here's a commit that adds the "bad fingerprint" test: https://github.com/wireapp/wire-server/pull/373/commits/079cf90478a5243e0fde1a13b97299f8edcd860f

services/brig/test/integration/API/Provider.hs
```diff
@@ -599,8 +616,31 @@ testMessageBot config crt db brig galley cannon = withTestService config crt db
 
     testMessageBotUtil uid uc cid pid sid sref buf brig galley cannon
 
-testAddRemoveBotTeam :: Maybe Config -> FilePath -> DB.ClientState -> Brig -> Galley -> Cannon -> Http ()
-testAddRemoveBotTeam config crt db brig galley cannon = withTestService config crt db brig defServiceApp $ \sref buf -> do
+testBadFingerprint :: Config -> DB.ClientState -> Brig -> Galley -> Cannon -> Http ()
```
neongreen: Here's the test that checks for bad fingerprints, copypasted from `testMessageBot`

services/brig/test/integration/API/Provider.hs
```diff
@@ -105,17 +98,41 @@ tests conf p db b c g = do
...
+    { privateKey   :: FilePath
+    , publicKey    :: FilePath
+    , cert         :: FilePath
+    , botHost      :: Text
+    , botPort      :: Int
+    } deriving (Show, Generic)
+
+instance FromJSON Config
+
+-- | Get the config from environment variables (and some defaults)
```
neongreen: I moved all `getEnv`s into one place so that I would be able to rely on the `Config` being a `Just` in all places

services/brig/test/integration/API/Provider.hs
```diff
@@ -558,8 +575,8 @@ testDeleteService config db brig = do
     getServiceProfile brig uid pid sid !!!
         const 404 === statusCode
 
-testAddRemoveBot :: Maybe Config -> FilePath -> DB.ClientState -> Brig -> Galley -> Cannon -> Http ()
```
neongreen: The certificate path can be taken from the config so there's no need to pass it around

services/brig/src/Brig/Provider/RPC.hs
```diff
@@ -60,10 +64,23 @@ data ServiceError
...
+            -- TODO: not sure if 'Reuse' is the right thing to do here --
+            -- maybe instead of returning the connection to the pool we want
+            -- to close it
+            Http.withConnection' req man Http.Reuse $ \mConn -> do
+                -- If we see this connection for the first time, verify fingerprints
+                let conn = Http.managedResource mConn
+                    seen = Http.managedReused   mConn
+                case (seen, fromDynamic @SSL.SSL (Http.connectionRaw conn)) of
+                    (True, _) -> pure ()
+                    (_, Nothing) -> error "TODO"
```
tiago-loureiro: So this should not be possible as services [only allow SSL connections](https://github.com/wireapp/wire-server/blob/develop/libs/brig-types/src/Brig/Types/Provider.hs#L430): Maybe we can fail in this case?

services/galley/src/Galley/External.hs
```diff
@@ -118,9 +122,23 @@ urlPort (HttpsUrl u) = do
...
+        let conn = Http.managedResource mConn
+            seen = Http.managedReused   mConn
+        case (seen, fromDynamic @SSL.SSL (Http.connectionRaw conn)) of
+            (True, _) -> pure ()
+            (_, Nothing) -> error "TODO"
+            (_, Just ssl) -> verifyFingerprints fprs ssl
+        -- Make a request using this connection and return it back to the pool
+        Http.withResponse req{Http.connectionOverride = Just mConn} man
+            (const $ return ())
+        -- TODO: this code is duplicated here and in Brig.Provider.RPC and
```
tiago-loureiro: Most definitely!

services/brig/src/Brig/Provider/RPC.hs
```diff
@@ -60,10 +64,23 @@ data ServiceError
...
-    mg <- ($ toList (sconFingerprints scon)) <$> view extGetManager
+    let fprs = toList (sconFingerprints scon)
+    (man, verifyFingerprints) <- view extGetManager
     extHandleAll onExc $ do
-        rs <- lift $ recovering x3 httpHandlers $ const $
-            liftIO $ Http.httpLbs req mg
+        rs <- lift $ recovering x3 httpHandlers $ const $ liftIO $
+            -- TODO: not sure if 'Reuse' is the right thing to do here --
+            -- maybe instead of returning the connection to the pool we want
+            -- to close it
```
tiago-loureiro: > maybe instead of returning the connection to the pool we want to close it

In case of popular services, keeping a connection open can/should be done; I expect the manager to do the right thing here and keep connections open as needed, etc; by default, it keeps up to 512 connections open across all hosts which sounds reasonable for our use case

tiago-loureiro:
> @tiago-loureiro @neongreen Did you know amazonka 1.6.0 has a slight issue? brendanhay/amazonka#466
> 
> https://github.com/blender/Rome/blob/ab0e93de16c9ade32d362ba680d8bda0a361199f/stack.yaml#L38-L47

No, I did not notice that. Thank you for the heads up!

services/brig/src/Brig/Provider/RPC.hs
neongreen: I'd rather keep the `error` here if it's indeed not possible. I changed the error message though.

neongreen:
This can be reviewed now.

neongreen:
(The tests pass locally.)

neongreen:
After this we might also want to do some permutation of these:

* switch to lts-12
* start using extensible snapshots
* rebuild alpine-builder

tiago-loureiro:
> (The tests pass locally.)

They seem to pass on our integration environment as well ;)

> After this we might also want to do some permutation of these:
> switch to lts-12
> start using extensible snapshots
> rebuild alpine-builder

Yup. I think that, as @jschaul mentioned, we can try upgrading alpine to [version 3.8](https://alpinelinux.org/posts/Alpine-3.8.0-released.html), which includes ghc-8.4 and will probably simplify our build pipeline.

I will review this PR and let's aim to merge it soon so that we can let it brew on our staging environment for some time.

Would also be good to finally get https://github.com/wireapp/wire-server/pull/203#issuecomment-393028321 in, it's been brewing here for way too long.

neongreen:
I will rebase this today and afterwards I shall beg @tiago-loureiro for a review

neongreen:
*sigh*

I really don't understand how on earth the `post /register` test can take 150 seconds, and I thought it was a thing that only happened in the lts-12 branch, but now it's also happening here.

neongreen:
BY GOLLY the day has finally come where I used strace and it wasn't in vain

I'm so happy

Ahem. Anyway, we're purging the user journal queue for 150s, and it only takes so long if other tests were run before this one.

services/brig/src/Brig/Blacklist.hs
```diff
@@ -0,0 +1,58 @@
+{-# LANGUAGE OverloadedStrings   #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TypeFamilies        #-}
+
+module Brig.Blacklist
```
neongreen: Not sure why this module is here at all :thinking:

services/brig/test/resources/turn/servers-v2.txt
```diff
@@ -1,2 +1 @@
-turn:localhost:3478
-turn:localhost:3479
\ No newline at end of file
+turn:localhost:3478
```
neongreen: Apparently this file is modified like this automatically whenever I run integration tests. Or at least sometimes.

Is this okay?

services/cargohold/src/CargoHold/S3.hs
```diff
@@ -500,15 +501,22 @@ completeResumable r = do
...
+                -> Aws.Env
+                -> S3Configuration Aws.NormalQuery
+                -> Bucket
+                -> Seq S3Chunk
+                -> ConduitT () ByteString (ResourceT IO) ()
+    chunkSource man env s3c b cs = case Seq.viewl cs of
         EmptyL  -> mempty
         c :< cc -> do
-            (src, fin) <- lift $ do
+            -- (src, fin) <- lift $ do
```
neongreen: Why is this commented out? Is there some TODO that should've been done here?

services/brig/test/resources/turn/servers-v2.txt
fisx: it's debatable whether this is ok, but it's expected behavior.  if you want to fix it, do it elsewhere!  :-)

services/cargohold/src/CargoHold/S3.hs
fisx: looks like it's safe to delete this.

services/brig/test/resources/turn/servers-v2.txt
fisx: (fviw, i would change the tests so they never change files under version control.  if they want to write to something that is under version control, they should copy it to another place first that is covered by .gitignore, and write to it there.)

tiago-loureiro:
>BY GOLLY the day has finally come when I used strace and it wasn't in vain
>
>I'm so happy111
>
>Ahem. Anyway, the investigation showed that we're purging the user journal queue for 150s, and it > only takes so long if other tests were ran before this one. Hm.

OK, I will have a look at this.

neongreen:
Ha-ha!

It gets stuck after exactly 99 open connections:

```
# during a long wait

[yom:~] $ ss -nap | grep $(pidof brig-integration) | wc -l
99
```

neongreen:
https://github.com/brendanhay/amazonka/issues/490

neongreen:
@tiago-loureiro Okay, this works locally now. Should we deploy it to staging? Or merge into develop? Or..?

tiago-loureiro:
> Okay, this works locally now. Should we deploy it to staging? Or merge into develop? Or..?

Let's see the integration tests run and give it a final review.

neongreen:
Now it's only missing a final review + it shouldn't be merged until we release the current `develop` to production (which I'm doing right now, just have to set up khan)

neongreen:
Closing in favor of https://github.com/wireapp/wire-server/pull/478.

PR #380, opened 2018-06-18
Add a per-conversation message timer

neongreen:
### Features

* [x] Add the `message_timer` field to the database
* [x] Allow providing an initial `message_timer` for a conversation
* [x] Add an endpoint for setting `message_timer`
* [x] Add an endpoint for getting `message_timer` or make sure it's provided as a part of some other endpoint
* [x] Send out an event when the timer is changed
* [x] Round negative values of the timer to zero

### Documentation

* [x] Make sure that the GET endpoint is documented in Swagger
* [x] Make sure that the PUT endpoint is documented in Swagger
* [x] Make sure that the event is documented in Swagger

### Testing

* [x] Test that providing an initial `message_timer` works
* [x] Test that setting and getting `message_timer` works
* [x] Test that the event is sent out when the timer is changed
* [x] Test that a guest can not change the timer
* [x] Test that a timer can not be set for a 1-to-1 conversation
* [x] Test that a timer can be unset

### Open questions

* [x] Should the backend check the range of the timer? (No.)
* [x] Should the `message-timer-update` event be renamed? (No.)

### Bureaucracy

* [x] Introduce bureaucracy

### DIFF
--- a/libs/api-client/src/Network/Wire/Client/API/Conversation.hs
+++ b/libs/api-client/src/Network/Wire/Client/API/Conversation.hs
@@ -106,6 +106,6 @@ createConv users name = sessionRequest req rsc readBody
     req = method POST
         . path "conversations"
         . acceptJson
-        . json (NewConv users name mempty Nothing Nothing)
+        . json (NewConv users name mempty Nothing Nothing Nothing)
         $ empty
     rsc = status201 :| []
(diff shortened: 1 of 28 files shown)


neongreen:
NB. We don't prohibit `"message_timer": ...` from appearing in 1:1 conversation creation requests, but it's not going to be parsed or written in the database.

PR #381, opened 2018-06-18
Richer user events

tiago-loureiro:
We also want to journal user name's changes (and their respective teamID's); the actual change set is rather small, there's a fair amount of refactor of the tests which should make further changes a lot simpler.


dkovacevic:
I would merge it

PR #386, opened 2018-06-28
Added missing config to brig when running the demo

tiago-loureiro:
Missed this one when turnV2 was added

### DIFF
--- a/deploy/services-demo/conf/brig.demo.yaml
+++ b/deploy/services-demo/conf/brig.demo.yaml
@@ -69,6 +69,7 @@ zauth:
     providerTokenTimeout: 604800 # 7 days
 
 turn:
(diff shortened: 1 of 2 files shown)


deploy/services-demo/conf/brig.demo.yaml
```diff
@@ -69,6 +69,7 @@ zauth:
     providerTokenTimeout: 604800 # 7 days
 
 turn:
+  serversV2: resources/turn/servers.txt
```
jschaul: should read: `servers-v2.txt`

tiago-loureiro: 🤦‍♂️  Thank you.

PR #392, opened 2018-07-04
Change end-point name from /i/backdoor-login to /i/sso-login.

fisx:
The old name introduced in https://github.com/wireapp/wire-server/pull/329 was confusing and prone to create misunderstandings.

This end-point will be used for SSO login.  Neither does it violate our attack model, nor is security deteriorated in any way.  Without it, admins with access to the cassandra nodes on the server would still be able to impersonate any user by temporarily replacing their password.


fisx:
@neongreen @jschaul @raphaelrobert

fisx:
:+1: to your changes @neongreen

PR #393, opened 2018-07-04
Spar routing in demo setup

jschaul:
Depends on 

* [x] #390 
* [x] Updated `migrations` docker image: https://github.com/wireapp/wire-server/commit/deb925a2fa1d5e12defa7ca8f6c24f09366471c6


deploy/services-demo/conf/spar.demo.yaml
```diff
@@ -0,0 +1,30 @@
+saml:
+  version:     SAML2.0
+  log_level:   Debug
+
+  sp_host:     0.0.0.0
+  sp_port:     8088
+  sp_app_uri:  http://localhost:8080/ # <--- change this to point to a reachable web app
```
fisx: ah, right, the web-app is not included here.  i wonder how hard that would be?  :-)

deploy/services-demo/demo.sh
```diff
@@ -108,6 +109,7 @@ run_haskell_service gundeck ${blue} Info
 run_haskell_service cannon ${orange} Info
 run_haskell_service cargohold ${purpleish} Info
 run_haskell_service proxy ${redish} Info
+run_haskell_service spar ${orange} Info
```
fisx: Yes, double-orange is better than red, as in the integration tests.  I'm going to change that there, too.

PR #394, opened 2018-07-04
Send team activation emails on /activate/send

neongreen:
Fixes https://github.com/wearezeta/backend-issues/issues/711. I haven't added a regression test for this, but teams.wire.com will hopefully maybe switch to the new flow at some point and make the existence of a regression test irrelevant.


gregor:
@neongreen We have changing the flow on our list but it hasn't gotten priority by product yet.

PR #395, opened 2018-07-05
Spar integrationt tests

fisx:
Added more integration tests and set them to pending.  Some small refactorings all over the place that I needed along the way.

Notes to the reviewer:

- I think this one can best be reviewed commit-by-commit.
- I'm most interested in the question whether any integration tests might be missing.  I think if we implement all the pending ones and get them to green, we may be ok.


services/spar/src/Spar/Intra/Brig.hs
```diff
@@ -118,6 +119,17 @@ confirmUserId buid = do
   maybe (pure Nothing) (const . pure . Just $ buid) (Brig.userTeam =<< usr)
 
 
+assertIsTeamOwner :: (HasCallStack, MonadError ServantErr m, MonadSparToBrig m) => Maybe UserId -> TeamId -> m ()
+assertIsTeamOwner mbuid tid = do
+  let reject = throwError err403 { errBody = encode [aesonQQ|{"error":"you need to be team admin to create an IdP"}|] }
```
jschaul: Instead of returning errors that are `{"error": "simple strings"}`, could you follow the error structure used in other services, that is to say, include a `code` and a descriptive, stable `label` (the idea being that clients can pattern match on the `label` field which should never change once introduced, and we can still change the exact textual error description while remaining backwards-compatible). Example:

```
{
  "code": 400,
  "message": "[path] 'id' invalid: Failed reading: Invalid UUID",
  "label": "client-error"
}
```

See e.g. https://github.com/wireapp/wire-server/blob/develop/services/galley/src/Galley/API/Error.hs#L45 for some more examples.

services/spar/test-integration/Test/Spar/APISpec.hs
```diff
@@ -167,45 +176,75 @@ spec = do
...
-      context "zuser is a team member, not a team admin" $ do
+      context "zuser is a team member, but not a team admin" $ do
         it "responds with 'forbidden' and a helpful message" $ do
-          pending
+          env <- ask
+          (_owner, tid) <- call $ createUserWithTeam (env ^. teBrig) (env ^. teGalley)
+          nobody <- let Just perms = Galley.newPermissions mempty mempty
+                    in call $ createTeamMember (env ^. teBrig) (env ^. teGalley) tid perms
+          callIdpCreate' (env ^. teSpar) (Just nobody) (env ^. teNewIdp)
+            `shouldRespondWith` check (== 403) [aesonQQ|{"error":"you need to be team admin to create an IdP"}|]
```
jschaul: Similar to the previous comment, tests here could match on `label`s of errors, not on the exact string representation. Instead of duplicating the error strings both in the src and in the tests, defining errors in one place and re-using them in the tests would be preferable.

services/brig/src/Brig/API.hs
```diff
@@ -1382,11 +1388,21 @@ canBeDeleted (uid ::: tid) = do
...
        Team.IsNotTeamOwner        -> pure ()
-       Team.NoTeamOwnersAreLeft   -> do
+       Team.NoTeamOwnersAreLeft   -> do  -- (keeping the user won't help in this case)
            Log.warn $ Log.field "user" (toByteString uid)
                     . Log.msg (Log.val "Team.NoTeamOwnersAreLeft")
     return empty
 
+isTeamOwner :: UserId ::: TeamId -> Handler Response
+isTeamOwner (uid ::: tid) = do
+    onlyOwner <- lift (Team.isOnlyTeamOwner uid tid)
```
jschaul: `Team.isOnlyTeamOwner` could use a better name. Perhaps `getOwnerState` or something better. (a function `isTeamOwner` calling `isOnlyTeamOwner` is misleading)

services/spar/test-integration/Test/Spar/APISpec.hs
```diff
@@ -228,6 +235,7 @@ spec = do
 
       context "some URLs are not https" $ do
```
jschaul: Not sure if you're aware of the `HttpsUrl` type defined [here](https://github.com/wireapp/wire-server/blob/develop/libs/types-common/src/Data/Misc.hs#L194-L227) and if that could be useful.

fisx:
all good feedback, no objections.  will process soon!

fisx:
processing done (mostly).  i would like to give this another round though and remove some more `pending`s.

services/brig/src/Brig/API.hs
```diff
@@ -1377,16 +1383,26 @@ addBlacklist emailOrPhone = do
...
        Team.IsOnlyTeamOwner       -> throwStd noOtherOwner
        Team.IsOneOfManyTeamOwners -> pure ()
        Team.IsNotTeamOwner        -> pure ()
-       Team.NoTeamOwnersAreLeft   -> do
+       Team.NoTeamOwnersAreLeft   -> do  -- (keeping the user won't help in this case)
            Log.warn $ Log.field "user" (toByteString uid)
                     . Log.msg (Log.val "Team.NoTeamOwnersAreLeft")
     return empty
 
+isTeamOwner :: UserId ::: TeamId -> Handler Response
```
tiago-loureiro: Is this really necessary? This simply looks up in galley and checks the status; or was it added here because atm spar talks only to brig?

services/spar/src/Spar/API.hs
```diff
@@ -133,37 +140,62 @@ withDebugLog msg showval action = do
...
--- [aesonQQ|{"error":"not a SAML metainfo URL"}|]
--- [aesonQQ|{"error":"invalid or unresponsive request URL"}|]
--- [aesonQQ|{"error":"public keys in request body and metainfo do not match"}|]
 
--- Type families to convert spar's 'API' type into an "outside-world-view" API type
+type MonadValidateIdP m = (MonadHttp m, MonadIO m)
+
+validateNewIdP :: forall m. (HasCallStack, MonadError SparError m, MonadValidateIdP m)
+               => NewIdP -> m ()
+validateNewIdP newidp = if True then pure () else do
```
tiago-loureiro: `if True then pure ()` ? :)

fisx: will be removed once the remainder of this function won't crash unconditionally any more.  :-).  sorry, should have added a TODO to that end.

services/brig/src/Brig/API.hs
fisx: I think `getZUsrTeam` returns a `teamid` even if the user is only member, not owner.  But you are correct if you mean that we should probably make that `getZUserOwnedTeam` and simplify the code accordingly.  I'll do that immediately!

fisx: b344ade720d06f9f9034abb42e3c9462072304d8

services/spar/spar.integration.yaml
```diff
@@ -14,6 +14,10 @@ saml:
       surname: Girlfriend
       email: email:president@evil.corp
 
+spInfo:
+  meta_uri:  https://app.wire.com/sso/metadata         # corresponds to 'APIMeta'
```
tiago-loureiro: What's the rationale behind mixing camelcase with underscore here?

services/spar/src/Spar/API.hs
```diff
@@ -133,37 +149,70 @@ withDebugLog msg showval action = do
   SAML.logger SAML.Debug $ "leaving " ++ msg ++ mconcat [": " ++ fromJust mshowedval | isJust mshowedval]
   pure val
 
-authorizeIdP :: (HasCallStack, MonadError ServantErr m, Brig.MonadSparToBrig m)
+-- | Called by get, delete handlers.
+authorizeIdP :: (HasCallStack, MonadError SparError m, SAML.SP m, Brig.MonadSparToBrig m)
```
tiago-loureiro: Do we need to return an `IdP` here?

services/spar/src/Spar/Options.hs
```diff
@@ -24,13 +23,15 @@ import GHC.Types (Symbol)
...
+import Spar.Types (IdPExtra, SPInfo)
 
 import qualified Data.Yaml as Yaml
 import qualified SAML2.WebSSO.Config as SAML
 
 
 data Opts = Opts
-    { saml           :: !(SAML.Config TeamId)
+    { saml           :: !(SAML.Config IdPExtra)
+    , spInfo         :: SPInfo
```
tiago-loureiro: Is the laziness here on purpose?

tiago-loureiro:
Looks good in general, I think some things would help out in having spar a bit more in sync with other services, for instance,
```
import qualified Data.Id as Brig
```

That's a bit odd; `Data.Id` is rather generic and it's IMHO a bit confusing to read `Brig.TeamId` vs simply `TeamId`.

services/spar/spar.integration.yaml
fisx: lazyness + ignorance of wire coding policy.  :-)

will fix.

fisx:
thanks for the comments, will fix everything.  yes, i don't feel i'm particularly in sync with the rest of the code.  any pointers to where specifically i'm off are greatly appreciated.

PR #399, opened 2018-07-10
Add an IdpGetAll endpoint, return meta-url and login-url

neongreen:


services/spar/test/Arbitrary.hs
```diff
@@ -27,6 +27,11 @@ instance Arbitrary NewIdP where
     _nidpPublicKey  <- arbitrary
     pure $ NewIdP {..}
 
+instance Arbitrary IdPList where
+  arbitrary = do
+    _idplProviders <- arbitrary
+    pure $ IdPList {..}
```
fisx:
```haskell
arbitrary = IdPList <$> arbitrary
```

but it's both good.

services/spar/spar.integration.yaml
```diff
@@ -14,6 +14,9 @@ saml:
       surname: Girlfriend
       email: email:president@evil.corp
 
+metaUrl:        https://app.wire.com/sso/metadata         # corresponds to 'APIMeta'
```
fisx: Should be under `spInfo`.  There are at least two metadata urls, one for IdP, one for SP, and it's not clear from the name here which is meant.

Also, should the comment be in the source code?  That's what it refers to.

fisx:
I will continue review later, and probably make the changes to I can merge myself if you have no objections.

Looking really good so far, thanks!

neongreen:
I 140% don't mind any changes

services/spar/src/Spar/Types.hs
```diff
@@ -23,7 +23,40 @@ import URI.ByteString
...
+  , _idpeSPInfo :: SPInfo
+  }
+  deriving (Eq, Show, Generic)
+
+makeLenses ''IdPExtra
+deriveJSON deriveJSONOptions ''IdPExtra
+
+-- | A list of 'IdP's, returned by some endpoints. Wrapped into an object to
+-- allow extensibility later on.
+data IdPList = IdPList
```
tiago-loureiro: I think that similarly to https://github.com/wireapp/wire-server/pull/402, we should make this an object

neongreen:
Isn't it already one? If not, it should be, yes

fisx:
All good (with my nit-picky changes).  I'm going to merge this into #395 myself now.

fisx:
thanks!

fisx:
(missing: idp-put endpoint; tests.  i'll add that now to the parent PR.)

PR #402, opened 2018-07-11
Expose push token list as a json object

tiago-loureiro:
Which is more flexible than using a plain list


libs/gundeck-types/src/Gundeck/Types/Swagger.hs
```diff
@@ -40,6 +41,12 @@ push = defineModel "Push" $ do
         description "Fallback transport"
         optional
 
+pushList :: Model
```
neongreen: Perhaps you could rename the model to `PushTokenList`? For greppability.

tiago-loureiro: 👍

PR #403, opened 2018-07-12
Fix haddock

fisx:


fisx:
(Running integration tests with haddock turned on even only for our own packages considerably drains battery and time.  I am strongly in favor of leaving it turned off.)

PR #405, opened 2018-07-17
Alternatively use SMTP for email

tiago-loureiro:
This PR introduces an optional way to send out emails: by using SMTP instead of forcing SES. The basic functionality works, want to improve a few things namely, we need to ensure that

- [x] - config is backwards compatible (Currently only works if config files are used)
- [x] - SMTP connection can be (re)established (use a pool)
- [x] - integration tests work with both configs
- [x] - general code cleanup

Feedback on how the config file could look easier to read is welcome!

### DIFF
--- a/services/brig/brig.integration.yaml
+++ b/services/brig/brig.integration.yaml
@@ -30,6 +30,10 @@ aws:
   userJournalQueue: integration-user-events.fifo
   # ^ Comment this out if you don't want to journal user events
   blacklistTable: integration-brig-userkey-blacklist
+  # ^ NOTE: blacklisting of emails (processing of bounces and complaints)  is only done
+  #         automatically IF sesQueue/sesEndpoint are used. If SMTP is used directly, the
+  #         operator must handle these notifications "manually" (there are internal endpoints)
+  #         that may be used for this
   prekeyTable: integration-brig-prekeys
   sqsEndpoint: http://localhost:4568 # https://sqs.eu-west-1.amazonaws.com
   dynamoDBEndpoint: http://localhost:4567 # https://dynamodb.eu-west-1.amazonaws.com
--- a/services/brig/brig.integration.yaml
+++ b/services/brig/brig.integration.yaml
@@ -38,11 +38,12 @@ emailSMS:
   # or you can use SMTP directly (blacklisting of email/phone must be otherwise handled by
   # the operator).
   email:
-    # sesQueue: integration-brig-events
-    # sesEndpoint: http://localhost:4569 # https://email.eu-west-1.amazonaws.com
-    smtpEndpoint: email-smtp.eu-west-1.amazonaws.com
-    smtpUser: foo
-    smtpPassword: bar
+    sesQueue: integration-brig-events
+    sesEndpoint: http://localhost:4569 # https://email.eu-west-1.amazonaws.com
+    # smtpEndpoint: email-smtp.eu-west-1.amazonaws.com
+    # smtpUsername: <username>
+    # smtpPassword: <password>
+    # smtpConnType: tls
 
   general:
     templateDir: deb/opt/brig/templates
--- a/services/brig/brig.integration.yaml
+++ b/services/brig/brig.integration.yaml
@@ -41,11 +41,13 @@ emailSMS:
     sesQueue: integration-brig-events
     sesEndpoint: http://localhost:4569 # https://email.eu-west-1.amazonaws.com
     # If you prefer to use SMTP directly, uncomment the following lines
-    # and set the correct credentials
-    # smtpEndpoint: email-smtp.eu-west-1.amazonaws.com
-    # smtpUsername: foo
-    # smtpPassword: <path_to_file_with_password>
-    # smtpConnType: tls | ssl | plain
+    # and set the correct credentials.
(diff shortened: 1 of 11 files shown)


tiago-loureiro:
This is now ready for review. Tests were run with both configs, SMTP was tested using Amazon's SMTP server.

services/brig/brig.integration.yaml
```diff
@@ -26,16 +26,27 @@ gundeck:
...
   dynamoDBEndpoint: http://localhost:4567 # https://dynamodb.eu-west-1.amazonaws.com
 emailSMS:
+  # You can either use SES directly (in which case, ensure a feedback queue is configured)
+  # or you can use SMTP directly (blacklisting of email/phone must be otherwise handled by
+  # the operator).
+  email:
+    sesQueue: integration-brig-events
+    sesEndpoint: http://localhost:4569 # https://email.eu-west-1.amazonaws.com
+    # If you prefer to use SMTP directly, uncomment the following lines
+    # and set the correct credentials
```
jschaul: From the code it seems when both options are defined, SES takes precedence. Perhaps add this to the configuration comments?

services/brig/test/integration/Main.hs
```diff
@@ -89,18 +90,25 @@ runTests iConf bConf otherArgs = do
...
                     $ fromByteString (BS.pack e)
 
+    parseEmailAWSOpts :: IO (Maybe Opts.EmailAWSOpts)
+    parseEmailAWSOpts = case Opts.email . Opts.emailSMS <$> bConf of
+        Just (Opts.EmailAWS aws) -> return (Just aws)
+        Just (Opts.EmailSMTP  _) -> return Nothing
+        _                        -> do
+            sesEnd   <- parseEndpoint <$> getEnv "AWS_SES_ENDPOINT"
+            sesQueue <- pack          <$> getEnv "AWS_USER_SES_QUEUE"
+            return . Just $ Opts.EmailAWSOpts sesQueue sesEnd
```
jschaul: I'm not sure we need to keep supporting environment variables in tests; we're not using this anywhere. (removing environment variable support could be done later, too)

services/brig/test/integration/API/User/Account.hs
```diff
@@ -299,14 +299,19 @@ testCreateUserBlacklist brig aws =
                                                , "password" .= defPassword
                                                ]
 
-    publishMessage :: Text -> Email -> AWS.Env -> Http ()
-    publishMessage typ em env = do
+    -- If there is no queue available, we need to force it either by publishing an event or using the API
+    forceBlacklist :: Text -> Email -> Http ()
+    forceBlacklist typ em = case aws^.AWS.sesQueue of
+        Just queue -> publishMessage typ em queue
+        Nothing    -> Bilge.post (brig . path "i/users/blacklist" . queryItem "email" (toByteString' em)) !!! const 200 === statusCode
```
jschaul: While this is fine in the tests; how will backlisting work with an SMTP server configured? (This is a little cheating: the tests testing blacklisting will pass even when SMTP instead of SES is configured; however no blacklisting is ever done when SMTP is used in production)

services/brig/src/Brig/SMTP.hs
```diff
@@ -0,0 +1,68 @@
...
+
+    create = do
+      (ok, conn) <- connect
+      if ok
+        then Logger.log lg Logger.Debug (msg $ val "Established connection to: " +++ host)
+        else Logger.log lg Logger.Warn (msg $ val "Failed to established connection, check your credentials to connect to: " +++ host )
+      return conn
+
+    destroy _ =
+      Logger.log lg Logger.Debug (msg $ val "Closing connection to: " +++ host)
```
jschaul: Shouldn't `destroy` take an action on the `SMTPConnection` object, like calling `closeSMTP`; or is that already done implicitly? (if so, where?)

services/brig/brig.integration.yaml
tiago-loureiro: Yes, good idea 👍

services/brig/test/integration/Main.hs
tiago-loureiro: > I'm not sure we need to keep supporting environment variables in tests; we're not using this anywhere. (removing environment variable support could be done later, too)

Good point. I actually added this "after" just for completeness.

> I'm not sure we need to keep supporting environment variables in tests; we're not using this anywhere. (removing environment variable support could be done later, too)

Let's do that then, totally agree to remove env variable usage at least in the tests.

jschaul:
> config is backwards compatible

The yaml config format is currently not backwards compatible (that's why the concourse tests fail). Did you mean to make it backwards compatible? Or did you just want to keep the command-line parameter syntax backwards compatible?

jschaul:
> Yes, given that we don't yet use yaml files on staging/production, I'd rather simplify things by not being backwards compatible when using a yaml file but only when using command line args

:+1:

services/brig/src/Brig/SMTP.hs
tiago-loureiro: When using a resource pool, the connection gets anyway closed when the resource is destroyed (also verified using `lsof`). However, it does sound like a good idea to try and disconnect gracefully with `closeSMTP` 👍

PR #408, opened 2018-07-18
Fix Spar: expect the correct response body type.

fisx:
- https://github.com/wireapp/wire-server/blob/develop/services/brig/src/Brig/API.hs#L131
- https://github.com/wireapp/wire-server/blob/develop/services/brig/src/Brig/API.hs#L1121

### DIFF
--- a/services/spar/src/Spar/Error.hs
+++ b/services/spar/src/Spar/Error.hs
@@ -27,7 +27,7 @@ data SparCustomError
 
   | SparBadUserName LT
   | SparNoBodyInBrigResponse
-  | SparCouldNotParseBrigResponse
+  | SparCouldNotParseBrigResponse LT
   | SparCouldNotRetrieveCookie
   | SparCassandraError LT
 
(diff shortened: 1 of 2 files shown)


services/spar/src/Spar/Error.hs
```diff
@@ -60,7 +60,7 @@ sparToWaiError (SAML.CustomError SparNotInTeam)                 = Wai.Error stat
 sparToWaiError (SAML.CustomError SparNotTeamOwner)              = Wai.Error status403 "forbidden" "You need to be team owner to create an IdP."
 sparToWaiError (SAML.CustomError (SparBadUserName msg))         = Wai.Error status400 "client-error" ("Bad UserName in SAML response: " <> msg)
 sparToWaiError (SAML.CustomError SparNoBodyInBrigResponse)      = Wai.Error status400 "server-error" "Brig response without body."
-sparToWaiError (SAML.CustomError SparCouldNotParseBrigResponse) = Wai.Error status400 "server-error" "Could not parse brig response body."
+sparToWaiError (SAML.CustomError (SparCouldNotParseBrigResponse msg)) = Wai.Error status400 "server-error" ("Could not parse brig response body: " <> msg)
 sparToWaiError (SAML.CustomError SparCouldNotRetrieveCookie)    = Wai.Error status400 "server-error" "Brig response contained no Set-Cookie header."
```
jschaul: server errrors should be status500, not 400.

services/spar/src/Spar/Error.hs
```diff
@@ -60,7 +60,7 @@ sparToWaiError (SAML.CustomError SparNotInTeam)                 = Wai.Error stat
...
 sparToWaiError (SAML.CustomError (SparCassandraError msg))      = Wai.Error status500 "server-error" ("Cassandra error: " <> msg)
 sparToWaiError (SAML.CustomError (SparNewIdPBadMetaUrl msg))    = Wai.Error status400 "client-error" ("bad or unresponsive metadata url: " <> msg)
```
jschaul: instead of having a range of different errors labelled `client-error`s, couldn't they have more descriptive labels, similar to what's done [e.g. in galley](https://github.com/wireapp/wire-server/blob/develop/services/galley/src/Galley/API/Error.hs#L94)?

services/spar/src/Spar/Intra/Brig.hs
```diff
@@ -101,7 +101,7 @@ createUser suid (Id buid) teamid = do
     . path "/i/users"
     . json newUser
     . expect2xx
-  parseResponse resp
+  userId . selfUser <$> parseResponse @SelfProfile resp
```
jschaul: `Spar.Intra.Brig.createUser` seems unused in any tests. How about adding this call to a test case?

PR #426, opened 2018-08-01
Use only Cassandra for handling blacklisted user keys

tiago-loureiro:
We have, for some time, been writing/reading to both DynamoDB and C* when looking at user keys. Once the migration of Dynamo to Cassandra has been finalised, this PR can be merged as it removes any references to DynamoDB (related to the blacklist table that is).


tiago-loureiro:
> Once the migration of Dynamo to Cassandra has been finalised

This has now been done so this PR can now be review and merged.

PR #427, opened 2018-08-01
Upgrade to LTS-12

neongreen:
```
Error: While constructing the build plan, the following exceptions were encountered:

In the dependencies for hedgehog-quickcheck-0.1:
    QuickCheck-2.11.3 from stack configuration does not match >=2.7 && <2.11 
                      (latest matching version is 2.10.1)
needed due to spar-0.1 -> hedgehog-quickcheck-0.1

In the dependencies for multihash-0.1.5:
    base-4.11.1.0 from stack configuration does not match >=4.7 && <4.11
                  (latest matching version is 4.10.1.0)
needed due to brig-1.35.0 -> multihash-0.1.5
```


neongreen:
It compiles and the tests pass, but some of them take awfully long time:

```
Brig API Integration
  user
    account
      post /register - 201 (with preverified):                         2018-08-02T09:12:22Z, I, known hosts: [datacenter1:rack1:127.0.0.1:9042]
q[galley] W, logger=cassandra.galley, 127.0.0.1:9042, read-loop: AsyncCancelled
[brig] W, logger=cassandra.brig, 127.0.0.1:9042, read-loop: AsyncCancelled
[gundeck] W, logger=cassandra.gundeck, 127.0.0.1:9042, read-loop: AsyncCancelled
[spar] W, logger=cassandra.spar, 127.0.0.1:9042, read-loop: AsyncCancelled
OK (151.36s)
```

This needs to be investigated.

neongreen:
The `read-loop: AsyncCancelled` thing is fixed by switching to newer `cql-io`. The test taking 150 seconds remains unexplained, though.

fisx:
does this deprecate #373?

neongreen:
No, it's based on that branch and it's not ready yet (LTS 11 can be merged
but this can't be)

neongreen:
* [ ] New `redis-io` works with GHC 8.4, so we should also consider upgrading to that instead of using a fork.

neongreen:
I'll try building the Docker images locally first.

neongreen:
Alright, and now this is blocked by #508, because new hscim entails a new authentication mechanism.

neongreen:
Let's see if CI passes!

Extensible snapshots will be done in a separate PR.

neongreen:
I'm pushing alpine-builder:lts-12 to Quay so that this PR can be tested on CI properly.

build/alpine/README.md
```diff
@@ -26,4 +26,4 @@ cd services/brig && make docker
 * `Dockerfile.intermediate` - based on `Dockerfile.deps`/`Dockerfile.builder`, this is an intermediate image compiling all dynamically linked binaries (obtained when running `make install` in the top-level directory).
 * `Dockerfile.executable` - based on `Dockerfile.deps`/`Dockerfile.intermediate`, this extracts a single executable from the intermediate image, yielding a small image with a single dynamically linked binary.
 * `Dockerfile.migrations` - same as `Dockerfile.executable`, with a fixed set of database migration binaries.
-* `Dockerfile.prebuilder` - dependencies of `Dockerfile.builder` that are expected to change very rarely (GHC, system libraries).
+* `Dockerfile.prebuilder` - dependencies of `Dockerfile.builder` that are expected to change very rarely (GHC, system libraries). Currently we're able to use system GHC, but if we require a newer version of GHC than the one provided by Alpine, we could build GHC in `Dockerfile.prebuilder` (as it has been done before).
```
fisx: link to https://github.com/wireapp/wire-server/releases/tag/v2018-11-28 here?

libs/types-common/src/Data/Id.hs
```diff
@@ -250,12 +250,14 @@ newtype RequestId = RequestId
...
-instance Monoid RequestId where
-    mempty = RequestId "N/A"
-    mappend a b
+instance Semigroup RequestId where
+    (<>) a b
         | a == mempty = b
         | otherwise   = a
 
+instance Monoid RequestId where
+    mempty = RequestId "N/A"
```
fisx: I guess this is fine as long as we're only using this for logging?  Perhaps we should add a comment about the broken Monoid instance to the newtype?

build/alpine/README.md
neongreen: Done.

libs/types-common/src/Data/Id.hs
neongreen: It doesn't seem broken (re/ laws), just weird. I'll make a PR about replacing it with `instance Default RequestId`.

(I'd prefer if it was banished completely and we had a `Maybe RequestId`, but it'd be a more invasive change and this way we at least get rid of a weird instance.)

PR #429, opened 2018-08-01
Change 'getOptions' to allow not having CLI options

neongreen:


neongreen:
Oops, some changes got here that shouldn't have. Fixed.

PR #433, opened 2018-08-03
[archived] Don't parse CLI options

neongreen:
Instead of using CLI options and env variables, we'd like to switch to YAML configs everywhere (if only to avoid having to change many different places whenever a single new option is added). The first step is to banish option-parsing code from wire-server.

Problems (I'm writing them down here so that I won't forget):

* The Cannon config uses `$(hostname -i)`, we'll have to switch to "external host file". Look at the diff for `run`.

* The Cargohold config uses `KHAN_DOMAIN` (that comes from `khan metadata`) and neither me nor Tiago know why. Same for Gundeck and Galley and Brig. Again, look at the diffs for `run`.

* I'm not sure if I should do the same for `galley/journaler` or not.

Also, I'm not going to rid of `LOG_LEVEL` here. It can be done later, or perhaps never.

### DIFF
--- a/services/brig/deb/etc/sv/brig/run
+++ b/services/brig/deb/etc/sv/brig/run
@@ -6,136 +6,17 @@ exec 2>&1
 # defaults
 USER=${USER:-www-data}
 APP=brig
-CONFIG=${CONFIG:-/etc/$APP/.env}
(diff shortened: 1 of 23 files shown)


neongreen:
All commits from this PR will find their way into separate PRs

neongreen:
Please don't delete the branch just yet.

PR #434, opened 2018-08-07
Implement a stub for whitelisted services search

neongreen:
Currently it just returns whatever /services returns. I'll replace it with a fully-functional version later.

### DIFF
--- a/services/brig/src/Brig/Provider/API.hs
+++ b/services/brig/src/Brig/Provider/API.hs
@@ -3,6 +3,7 @@
 {-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE TupleSections     #-}
 {-# LANGUAGE TypeFamilies      #-}
+{-# LANGUAGE TypeApplications  #-}
 {-# LANGUAGE TypeOperators     #-}
 
 module Brig.Provider.API (routes) where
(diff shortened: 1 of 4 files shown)


services/brig/src/Brig/Provider/API.hs
```diff
@@ -583,17 +591,36 @@ getServiceProfile (pid ::: sid) = do
...
+-- purely to make testing easier for clients.
+--
+-- Before going to production, it needs to start considering the whitelist.
+-- It would be also quite nice to have both "start" and "prefix" here to
+-- make pagination work in the future.
+searchTeamServiceProfiles
+    :: UserId ::: TeamId ::: Maybe (Range 1 128 Text) ::: Range 10 100 Int32
+    -> Handler Response
+searchTeamServiceProfiles (u ::: t ::: prefix ::: size) = do
+    member <- lift $ RPC.getTeamMember u t
```
tiago-loureiro: You could instead just fetch the `teamId` in the user profile instead of doing an RPC (using [lookupUser](https://github.com/wireapp/wire-server/blob/develop/services/brig/src/Brig/Data/User.hs#L205) or even better having a new lookup that fetches only the `teamId`)

neongreen: Sure, will do

PR #435, opened 2018-08-09
Add RequestId field to Spar logs

fisx:
there is a small integration test "fix" piggy-backing on this (last commit).

### DIFF
--- a/services/spar/src/Spar/App.hs
+++ b/services/spar/src/Spar/App.hs
@@ -7,6 +7,7 @@
 {-# LANGUAGE ScopedTypeVariables        #-}
 {-# LANGUAGE TypeApplications           #-}
 {-# LANGUAGE TypeFamilies               #-}
+{-# LANGUAGE ViewPatterns               #-}
 
 module Spar.App where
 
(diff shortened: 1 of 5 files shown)


services/spar/src/Spar/App.hs
```diff
@@ -58,11 +60,20 @@ instance HasConfig Spar where
...
+  logger (toLevel -> lv) mg = do
+    lg <- asks sparCtxLogger
+    reqid <- asks sparCtxRequestId
+    let fields, mg' :: Log.Msg -> Log.Msg
+        fields = Log.field "request" (unRequestId reqid)
+        mg'    = Log.msg . condense $ flatten <$> mg
+    Spar . Log.log lg lv $ fields Log.~~ mg'
     where
-      mg' = Log.msg $ flatten <$> mg
+      condense = dropWhile (== ' ') . reverse . dropWhile (== ' ') . reverse . f
```
tiago-loureiro: Looks like a lot of work being done here... is it really necessary to remove all "extra" whitespaces when logging?

fisx: saml2-web-sso dumps large haskell values into the log in pretty-printed form.  while not strictly necessary, it looks much nicer with `condense`.  please pick what i should do:

- [ ] leave the code as is;
- [ ] remove `condense` entirely;
- [ ] rewrite `condense` to use lazy bytestrings, which should be fast enough given that we're about to write this string to a socket, which should dwarf the overhead for string brushing.

tiago-loureiro: Let's try then 3 and perhaps have a 2nd look if the need arises. Particularly since you mention _large_ haskell values.

PR #437, opened 2018-08-09
Add minimal smtp/postfix docker image to the demo

jschaul:
Allows sending real emails when e.g signing up via POST
localhost:8080/register

Note that these emails will most likely land in your spam folder.

WIP note: email sending works, but smoketests don't actually work (when the mailbot is configured to use gmail), since emails land in the spam folder, where the mailbot doesn't find them. Three options:

1. Make the smoketester also look in the spam folder
2. Invest time into using a mail server docker image from which emails are not labelled as spam (might be hard)
3. configure smoketester's mailbox.json to use an account without spam detection (e.g. another docker image running a mail server)

Thoughts?

### DIFF
--- a/deploy/docker-ephemeral/docker-compose.yaml
+++ b/deploy/docker-ephemeral/docker-compose.yaml
@@ -3,18 +3,18 @@ services:
   fake_dynamodb:
     image: cnadiminti/dynamodb-local:2018-04-11
     ports:
-      - 4567:8000
+      - 127.0.0.1:4567:8000
 
   fake_sqs:
     image: airdock/fake-sqs:0.3.1
     ports:
-      - 4568:4568
+      - 127.0.0.1:4568:4568
 
   fake_localstack:
     image: localstack/localstack:0.8.0  # NB: this is younger than 0.8.6!
     ports:
-      - 4569:4579 # ses # needed for local integration tests
-      - 4575:4575 # sns
+      - 127.0.0.1:4569:4579 # ses # needed for local integration tests
+      - 127.0.0.1:4575:4575 # sns
     environment:
       - DEBUG=1
       - DEFAULT_REGION=eu-west-1
@@ -28,7 +28,7 @@ services:
         # smtp connection handling needs to be made more generic to accept
         # alternative ports. See changes introduced in
         # https://github.com/wireapp/wire-server/pull/405
-      - "25:25"
+      - "127.0.0.1:25:25"
     environment:
       - maildomain=mail.wiredemo.example.com
       - smtp_user=dummy:dummy-smtp-password
@@ -36,7 +36,7 @@ services:
   fake_s3:
     image: minio/minio:RELEASE.2018-05-25T19-49-13Z
     ports:
-     - "4570:9000"
+     - "127.0.0.1:4570:9000"
     environment:
       MINIO_ACCESS_KEY: dummykey
       MINIO_SECRET_KEY: dummysecret # minio requires a secret of at least 8 chars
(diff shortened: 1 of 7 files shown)


tiago-loureiro:
> Make the smoketester also look in the spam folder

I think we should try this, I will have a look.

tiago-loureiro:
@jschaul I guess you got ahead https://github.com/wireapp/wire-server/pull/437/commits/f09596e67dc9c3450f37fea10ea976b603a3699d

jschaul:
Optional "check-the-spam-folder" option added. Ready for review.

deploy/docker-ephemeral/docker-compose.yaml
```diff
@@ -13,13 +13,26 @@ services:
...
-      - 4569:4579 # ses
+      - 4569:4579 # ses # needed for local integration tests
       - 4575:4575 # sns
     environment:
       - DEBUG=1
       - DEFAULT_REGION=eu-west-1
       - SERVICES=ses,sns
 
+  basic_smtp: # needed for demo setup
+    # https://github.com/catatnight/docker-postfix
```
fisx: is this image binding the listener to `0.0.0.0`?  if so, how do we avoid receiving spam from the outside world, and possibly relaying it?

deploy/services-demo/README.md
```diff
@@ -68,13 +68,12 @@ In order to view the API, you need to create a regular user. For that purpose, y
...
 
-Then, have a look at what the configuration for the [api-smoketest](../../tools/api-simulations/README.md) should be. Once you have the correct `mailboxes.json`, this should just work from the top level directory (note the `sender-email` must match brig's [sender-email](https://github.com/wireapp/wire-server/blob/develop/services/brig/brig.integration.yaml#L35))
+Example:
 
 ```
-# This assumes, for now, that brig is configured to use a real AWS SES endpoint
-../../dist/api-smoketest --api-host=127.0.0.1 --api-port=8080 --api-websocket-host=127.0.0.1 --api-websocket-port=8081 --mailbox-config=<path_to_mailboxes_file> --sender-email=backend-integration@wire.com --enable-asserts
+../../dist/api-smoketest --api-host=127.0.0.1 --api-port=8080 --api-websocket-host=127.0.0.1 --api-websocket-port=8081 --mailbox-config=<path_to_mailboxes_file> --sender-email=backend-demo@mail.wiredemo.example.com --mailbox-folder INBOX --mailbox-folder '[Gmail]/Spam' --enable-asserts
```
fisx: not really about this PR, but you could break this line up to make it fit into the column and you don't have to scroll horizontally.

deploy/docker-ephemeral/docker-compose.yaml
jschaul: Good catch - people should use firewalls and not expose their dev setup to the world anyway, but I suppose we can provide better defaults for people who're not careful with their setup.

> is this image binding the listener to 0.0.0.0?

Yes, like all other docker containers started with docker-compose, it makes them accessible to the external world by default:

```
sudo netstat -antlp | grep LISTEN | grep 25
tcp6       0      0 :::25                   :::*                    LISTEN      3655/docker-proxy
```

It looks like this can be fixed by prefixing the exposed ports with `127.0.0.1:`, so after [this commit](84b08851c5749a020f1079d98dc8daf0227188bc):

```
sudo netstat -antlp | grep LISTEN | grep 25
tcp        0      0 127.0.0.1:25            0.0.0.0:*               LISTEN      28084/docker-proxy
```

PR #439, opened 2018-08-13
Internal provider activation

tiago-loureiro:
Allows for better automation of provider creation/activation

### DIFF
--- a/libs/types-common/src/Data/Code.hs
+++ b/libs/types-common/src/Data/Code.hs
@@ -60,7 +60,7 @@ deriving instance Cql Value
 
 -- | A key/value pair. This would actually more accurately if the value would actually
 -- be a "value" but since we use "key" and "code" already in quite a few place in the API
--- (but without a type, using plain fields; this will make it easier to re-use a key/value
+-- (but without a type, using plain fields). This will make it easier to re-use a key/value
 -- pair in the API, keeping "code" in the JSON for backwards compatibility
 data KeyValuePair = KeyValuePair
   { kcKey  :: !Key
--- a/libs/types-common/src/Data/Code.hs
+++ b/libs/types-common/src/Data/Code.hs
@@ -1,20 +1,24 @@
 {-# LANGUAGE CPP                        #-}
 {-# LANGUAGE DataKinds                  #-}
+{-# LANGUAGE DeriveGeneric              #-}
 {-# LANGUAGE GeneralizedNewtypeDeriving #-}
 {-# LANGUAGE OverloadedStrings          #-}
 {-# LANGUAGE StandaloneDeriving         #-}
+{-# LANGUAGE TemplateHaskell            #-}
 
 -- | Types for verification codes.
 module Data.Code where
 
 import Data.Aeson hiding (Value)
+import Data.Aeson.TH
 import Data.ByteString.Conversion
 import Data.Int
 import Data.Range
 import Data.Scientific (toBoundedInteger)
+import Data.Json.Util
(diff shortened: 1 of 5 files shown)


services/brig/test/integration/API/Provider.hs
```diff
@@ -169,6 +170,58 @@ testRegisterProvider db brig = do
...
+        const 201 === statusCode
+
+    let Just npr = decodeBody _rs :: Maybe NewProviderResponse
+    -- Since a password was given, none should have been generated
+    liftIO $ assertBool "password" (isNothing (rsNewProviderPassword npr))
+    let pid = rsNewProviderId npr
+
+    -- No login possible directly after registration
+    loginProvider brig email defProviderPassword !!!
+        const 403 === statusCode
```
neongreen: Is there also an error label that we could check for?

services/brig/test/integration/API/Provider.hs
```diff
@@ -169,6 +170,58 @@ testRegisterProvider db brig = do
...
+    activateProvider brig (Code.kcKey pair) (Code.kcCode pair) !!!
+        const 200 === statusCode
+
+    -- Login succeeds after activation (due to auto-approval)
+    loginProvider brig email defProviderPassword !!!
+        const 200 === statusCode
+
+    -- Email address is now taken
+    registerProvider brig new !!!
+        const 409 === statusCode
```
neongreen: And error label here.

services/brig/test/integration/API/Provider.hs
```diff
@@ -169,6 +170,58 @@ testRegisterProvider db brig = do
         assertEqual "description" defProviderDescr (providerDescr p)
         assertEqual "profile" (ProviderProfile p) pp
 
+-- | Step-by-step registration procedure with verification
+-- of pre- and post-conditions, using an internal endpoint
+-- instead of accessing the DB directly
+testRegisterProviderInternal :: Brig -> Http ()
```
neongreen: This test seems to be mostly a duplicate of `testRegisterProvider`. If you don't want to refactor them, it'd be nice to at least have a `-- TODO: duplicate of ...` comment here.

services/brig/test/integration/API/Provider.hs
```diff
@@ -169,6 +170,58 @@ testRegisterProvider db brig = do
...
+        const 409 === statusCode
+
+    -- Retrieve full account and public profile
+    -- (these are identical for now).
+    uid <- randomId
+    _rs <- getProvider brig pid <!! const 200 === statusCode
+    let Just p  = decodeBody _rs
+    _rs <- getProviderProfile brig pid uid <!! const 200 === statusCode
+    let Just pp = decodeBody _rs
+    liftIO $ do
```
neongreen: Can you add a comment to `Provider` saying that there are tests that compare fields one-by-one and therefore they have to be updated when there are fields added to `Provider`?

libs/types-common/src/Data/Code.hs
```diff
@@ -53,3 +57,14 @@ instance FromJSON Timeout where
 deriving instance Cql Key
 deriving instance Cql Value
 #endif
+
+-- | A key/value pair. This would actually more accurately if the value would actually
+-- be a "value" but since we use "key" and "code" already in quite a few place in the API
+-- (but without a type, using plain fields; this will make it easier to re-use a key/value
+-- pair in the API, keeping "code" in the JSON for backwards compatibility
```
neongreen: Missing `)`.

PR #441, opened 2018-08-15
Bump saml2-web-sso, hsaml2 deps.

fisx:
the hsaml2 commit referenced before was destroyed when i merged a branch.  saml2-web-sso is just in here because i was already at it.

from here on, i will only merge PRs that have depencies pointing to commits that have already landed stable branches.


neongreen:
I looked at the diffs and it seems okay

fisx:
the failed test has nothing to do with this PR (something flaky in gundeck integration).

PR #442, opened 2018-08-15
Internal queue v2

tiago-loureiro:
This is basically https://github.com/wireapp/wire-server/pull/372 with `develop` merged in; didn't want to make that PR messy.


stack.yaml
```diff
@@ -66,11 +66,15 @@ packages:
 # services/spar:
 - location:
     git: https://github.com/wireapp/saml2-web-sso
+<<<<<<< HEAD
```
neongreen: merge conflict

services/brig/src/Brig/AWS.hs
```diff
@@ -121,9 +119,8 @@ mkEnv lgr opts emailOpts mgr = do
                      (mkEndpoint SQS.sqs      (Opt.sqsEndpoint opts))
                      (mkEndpoint DDB.dynamoDB (Opt.dynamoDBEndpoint opts))
     sq <- maybe (return Nothing) (fmap Just . getQueueUrl e . Opt.sesQueue) emailOpts
-    iq <- getQueueUrl e (Opt.internalQueue opts)
     jq <- maybe (return Nothing) (fmap Just . getQueueUrl e) (Opt.userJournalQueue opts)
-    return (Env g sq iq jq pk e)
```
neongreen: This is the conflict from the original PR:

```
<<<<<<< internal-queue
    return (Env g sq jq bl pk e)
=======
    return (Env g sq iq jq pk e)
>>>>>>> develop
```

I'm curious where has blacklist (`bl`) gone.

tiago-loureiro: https://github.com/wireapp/wire-server/pull/426 we now use C* for that.

PR #444, opened 2018-08-15
Remove de-whitelisted services from conversations

neongreen:


neongreen:
Hm, crap, this is actually wrong. `deleteBot` also deletes the user, which we don't really want to do.

neongreen:
(It *might* be true currently that every bot user only exists in one conversation, but I wouldn't want to rely on it.)

neongreen:
Oh hmmm, nevermind, it actually seems to be true.

> A bot is created and destroyed as it is added and removed from a conversation, respectively, i.e. it is scoped to a single conversation and its lifecycle bound to its membership in that conversation.

neongreen:
By the way, it's kinda amusing to trace what happens when a bot is dropped from a conversation:

```
GALLEY.deleteBot
    BRIG.botDeleteSelf                                       at /bot/self
        BRIG.deleteBot
            BRIG.removeBotMember
                GALLEY.rmBot                                 at /i/bots
                    GALLEY.push1 (Event MemberLeave ...)
                    GALLEY.removeMember bot conv
                    GALLEY.eraseClients bot
            BRIG.rmClient =<< BRIG.lookupClients bot
            BRIG.updateStatus bot Deleted
```

services/brig/src/Brig/API.hs
```diff
@@ -211,6 +214,15 @@ sitemap o = do
       .&. contentType "application" "json"
       .&. request
 
+    -- provided a list of services, tells you for each whether it's
+    -- whitelisted or not; gets an array of ServiceRefs and returns an array
+    -- of bools
+    post "/i/teams/:tid/services/whitelisted" (continue internalQueryServiceWhitelist) $
```
tiago-loureiro: It'd be good to avoid passing around plain lists in JSON as they are a bit inflexible; given that it's only an internal endpoint, accepting a list is probably not too problematic.

Wrt the result type, I'd avoid a plain list of bools since that relies on an order which is a bit fragile and easy to get wrong on the caller side; the overhead of sending back an object with sth like `pid,sid,whitelisted/notwhitelisted` seems worthwhile.

services/brig/src/Brig/API.hs
```diff
@@ -1037,6 +1049,14 @@ internalListClients (_ ::: _ ::: req) = do
...
     return $ json (UserClients ucs)
 
+internalQueryServiceWhitelist :: TeamId ::: Request -> Handler Response
+internalQueryServiceWhitelist (tid ::: req) = do
+    svcs :: [ServiceRef] <- parseJsonBody req
+    fmap json $ forM svcs $ \svc -> do
+        let pid = svc ^. serviceRefProvider
+            sid = svc ^. serviceRefId
+        getServiceWhitelistStatus tid pid sid
+
```
tiago-loureiro: Given that you want to get the whitelisted services for a single team, you can do this in a single read; with the definition:
```
        create table if not exists service_whitelist
            ( team     uuid
            , provider uuid
            , service  uuid
            , primary key (team, provider, service)
            ) with clustering order by (provider asc, service asc)
```
you can simply do `SELECT pid,sid FROM service_whitelist` and then filter it at the application level.

Actually, assuming that there are only a handful services whitelisted, the most efficient could even be to implement a `GET /teams/:id/services/whitelisted` or something like that although this is something that I don't _really_ suggest :)

neongreen:
Okay, now please re-re-review

services/galley/src/Galley/API/Update.hs
```diff
@@ -424,8 +427,29 @@ postNewOtrMessage usr con cnv val msg = do
         let (toBots, toUsers) = foldr (newMessage usr con (Just cnv) msg now) ([],[]) rs
         pushSome (catMaybes toUsers)
         void . fork $ do
-            gone <- External.deliver toBots
+            dewhitelisted <- getDewhitelisted (map fst toBots)
```
tiago-loureiro: How about checking that `toBots` is not empty? Assuming most conversations _do not_ have bots, we'd be saving an RPC for every message sent.

neongreen:
@tiago-loureiro here you go

PR #445, opened 2018-08-15
Don't allow bots in team-only conversations

neongreen:


services/brig/test/integration/API/Provider.hs
```diff
@@ -510,9 +513,35 @@ testAddRemoveBotTeam config crt db brig galley cannon = withTestService config c
...
+    bid <- addBotConv brig cannon uid1 uid2 cid pid sid buf
+    let buid = botUserId bid
+    -- Make the conversation team-only again and check that the bot has been removed
+    setAccessRole uid1 cid TeamAccessRole
+    WS.bracketR cannon uid1 $ \ws -> do
+        _ <- waitFor (5 # Second) not (isMember galley buid cid)
+        getBotConv galley bid cid !!!
+            const 404 === statusCode
+        svcAssertMemberLeave buf buid [buid] cid
+        wsAssertMemberLeave ws cid buid [buid]
```
neongreen: This one fails: `Exception: Timeout: No matching notification received`. I've no idea why :(

neongreen:
This now passes tests (at least locally).

I was getting another error, by the way -- that was possibly related to the fact that during a conversation access update we try to deliver events to bots even after we remove them from the conversation. It's possible that clients and bots handle that just fine, but for the sake of tidiness I decided to rewrite `updateConversationAccess` to do things properly:

* First it does all the checks.
* Then it figures out which bots and users should be removed.
* Then everyone gets a conversation-access-update event. Hey, if we remove them, they deserve to know why.
* Finally we kick out bots (individually because there's no `deleteBots`), remove users, and send the MemberLeave event to remaining bots + all users.

services/brig/src/Brig/Provider/API.hs
```diff
@@ -689,7 +689,15 @@ addBot (zuid ::: zcon ::: cid ::: req) = do
     maxSize <- fromIntegral . setMaxConvSize <$> view settings
     unless (length (cmOthers mems) < maxSize - 1) $
         throwStd tooManyMembers
-    for_ (cnvTeam cnv) $ ensureNotManagedConv
+
+    -- For team conversations: bots are not allowed in managed and in
+    -- team-only conversations
```
neongreen: These two lines do the trick in this PR.

services/galley/src/Galley/API/Update.hs
```diff
@@ -116,71 +120,101 @@ updateConversationAccess (usr ::: zcon ::: cnv ::: req ::: _ ) = do
     body <- fromBody req invalidPayload :: Galley ConversationAccessUpdate
     let targetAccess = Set.fromList (toList (cupAccess body))
         targetRole = cupAccessRole body
-    -- checks and balances
```
neongreen: Ehhhh, this is where you should switch to split diff mode.

services/brig/src/Brig/Provider/API.hs
```diff
@@ -689,7 +689,15 @@ addBot (zuid ::: zcon ::: cid ::: req) = do
...
+    when (cnvAccessRole cnv == TeamAccessRole) $
+        throwStd invalidConv
+    for_ (cnvTeam cnv) $ \tid -> do
```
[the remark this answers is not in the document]
neongreen: The inner machinations of my mind are an [enigma](https://www.youtube.com/watch?v=KNZSXnrbs_k)

PR #448, opened 2018-08-16
Build profiling versions of libraries in alpine-builder

neongreen:
This is the first step towards having automatic stack traces in integration tests. (The second step will be enabling `-xc` when running integration tests, and the third step will be just fixing makefiles/etc so that locally we would get stack traces as well.)

I checked locally and the image builds successfully; I can also upload it to Quay if nobody objects.

### DIFF
--- a/build/alpine/Dockerfile.builder
+++ b/build/alpine/Dockerfile.builder
@@ -5,18 +5,25 @@ ARG prebuilder=quay.io/wire/alpine-prebuilder
 FROM ${prebuilder}
 WORKDIR /src/wire-server
 
+# Get newer Stack
+
+# later versions don't have static binaries (temporarily)
+ARG STACK_VERSION=1.6.5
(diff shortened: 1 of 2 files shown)


build/alpine/Dockerfile.builder
```diff
@@ -5,18 +5,25 @@ ARG prebuilder=quay.io/wire/alpine-prebuilder
...
-# image – that's because Stack sometimes doesn't realize that local packages
-# should be recompiled, and we really don't want that to happen during CI.
+# We also build profiling versions of all libraries. Due to a bug in Stack,
+# they have to be built in a separate directory. See this issue:
+# https://github.com/commercialhaskell/stack/issues/4032
+
 RUN apk add --no-cache git ncurses && \
     mkdir -p /src && cd /src && \
-    git clone https://github.com/wireapp/wire-server.git && \
+    git clone -b develop https://github.com/wireapp/wire-server.git && \
```
fisx: This is an independent fix, no?

fisx: (same with "newer stack"?)

neongreen: Independent, yes. I didn't want to make three PRs.

PR #450, opened 2018-08-17
Track bots

neongreen:
In this PR:

* Track bots in the `service_user` table
* Additionally track bots in team conversations in the `service_team` table
* Remove bots from conversations when the service is gone, immediately
* Remove bots from conversations when they are de-whitelisted
* Add a migration tool for filling `service_user` and `service_team`
* Add a tool that whitelists all services used by teams

Note that this PR removes an internal endpoint which Brig provided for Galley and which is called every time somebody posts a message in a conversation with bots. I'm not sure what's the procedure for deploying such PRs to staging.


neongreen:
Ideally `paginateC` should be used in a bunch of other places in wire-server, too, but I'm too sleepy.

PR #452, opened 2018-08-22
SSO integration tests; refactoring.

fisx:
This is closely related to https://github.com/wireapp/saml2-web-sso/pull/21

### DIFF
--- a/services/spar/test-integration/Test/Spar/APISpec.hs
+++ b/services/spar/test-integration/Test/Spar/APISpec.hs
@@ -124,22 +124,16 @@ spec = do
 
       context "access granted" $ do
         it "responds with a very peculiar 'allowed' HTTP response" $ do
-          pending
           (idp, privcreds, authnreq) <- negotiateAuthnRequest
           authnresp <- liftIO $ mkAuthnResponse privcreds idp authnreq True
           sparresp <- submitAuthnResponse authnresp
           liftIO $ do
-            print $ unlines
-              [ cs . renderLBS def { rsPretty = True } . fromSignedAuthnResponse $ authnresp
-              , ppShow sparresp
-              , maybe "Nothing" cs (responseBody sparresp)
-              ]
             statusCode sparresp `shouldBe` 200
             let bdy = maybe "" (cs @LBS @String) (responseBody sparresp)
             bdy `shouldContain` "<?xml version=\"1.0\" encoding=\"UTF-8\"?>"
             bdy `shouldContain` "<!DOCTYPE html PUBLIC \"-//W3C//DTD XHTML 1.1//EN\" \"http://www.w3.org/TR/xhtml11/DTD/xhtml11.dtd\">"
-            bdy `shouldContain` "<title>???</title>"
-            bdy `shouldContain` "window.opener.postMessage({type: 'AUTH_ERROR', payload: {label: '???'}}, receiverOrigin)"
+            bdy `shouldContain` "<title>wire:sso:success</title>"
+            bdy `shouldContain` "window.opener.postMessage({type: 'AUTH_SUCCESS'}, receiverOrigin)"
 
         context "unknown user" $ do
           it "creates the user" $ do
--- a/services/spar/test-integration/Test/Spar/APISpec.hs
+++ b/services/spar/test-integration/Test/Spar/APISpec.hs
@@ -21,12 +21,16 @@ import Data.Maybe
 import Data.String.Conversions
 import Data.UUID as UUID hiding (null, fromByteString)
 import Galley.Types.Teams as Galley
+import GHC.Stack
 import Lens.Micro
 import Prelude hiding (head)
 import SAML2.WebSSO as SAML
+import SAML2.WebSSO.Test.Credentials
+import SAML2.WebSSO.Test.MockResponse
 import Spar.Types
 import Util
 
+import qualified Data.ByteString.Lazy as LBS
 import qualified Network.HTTP.Types.Status as HTTP
 import qualified Spar.Intra.Brig as Intra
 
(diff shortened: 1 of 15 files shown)


fisx:
CI is having technical difficulties; but the tests pass locally.

Hints for the reviewer: this should be almost exclusively semantics-preserving refactoring and work on the integration tests.  I'm interested if I've missed any impact on production, though.  That would be worth taking a closer look together.

services/spar/test-integration/Test/Spar/APISpec.hs
```diff
@@ -90,24 +94,46 @@ spec = do
...
+          liftIO $ do
+            -- import Text.XML
+            -- putStrLn $ unlines
+            --   [ cs . renderLBS def { rsPretty = True } . fromSignedAuthnResponse $ authnresp
+            --   , show sparresp
+            --   , maybe "Nothing" cs (responseBody sparresp)
+            --   ]
+            statusCode sparresp `shouldBe` 200
+            let bdy = maybe "" (cs @LBS @String) (responseBody sparresp)
+            bdy `shouldContain` "<?xml version=\"1.0\" encoding=\"UTF-8\"?>"
```
neongreen: Does the body only contain these four lines? If yes, then it's better to use `shouldBe`

services/spar/test-integration/Test/Spar/APISpec.hs
```diff
@@ -148,33 +174,34 @@ spec = do
...
           context "zuser is a team member, but not a team owner" $ do
             it "responds with 'insufficient-permissions' and a helpful message" $ do
               env <- ask
-              (_owner, tid, idp) <- createTestIdP
+              let idpid = env ^. teIdP . idpId
+                  teamid = env ^. teTeamId
               newmember <- let Just perms = newPermissions mempty mempty
-                        in call $ createTeamMember (env ^. teBrig) (env ^. teGalley) tid perms
-              whichone (env ^. teSpar) (Just newmember) idp
+                        in call $ createTeamMember (env ^. teBrig) (env ^. teGalley) teamid perms
```
neongreen: given how often `tid` is used in other services as a shorthand for a variable with type `TeamId`, I'd say that it's better to keep it as `tid`

services/spar/test-integration/Test/Spar/APISpec.hs
```diff
@@ -207,10 +235,11 @@ spec = do
...
         it "responds with 2xx and removes IdP" $ do
           env <- ask
-          (uid, _, idp) <- createTestIdP
-          callIdpDelete' (env ^. teSpar) (Just uid) idp
+          let idpid = env ^. teIdP . idpId
+              userid = env ^. teUserId
+          callIdpDelete' (env ^. teSpar) (Just userid) idpid
             `shouldRespondWith` \resp -> statusCode resp < 300
-          callIdpGet' (env ^. teSpar) (Just uid) idp
+          callIdpGet' (env ^. teSpar) (Just userid) idpid
```
neongreen: ditto

services/spar/test-integration/Test/Spar/APISpec.hs
```diff
@@ -248,47 +277,32 @@ spec = do
...
-            respurl <- endpointToURL (env ^. teMockIdp) "resp"
-            let newidp = (env ^. teNewIdp)
-                  & nidpMetadata   .~ metaurl
-                  & nidpRequestUri .~ respurl
-                  & modnewidp
-            (uid, _) <- call $ createUserWithTeam (env ^. teBrig) (env ^. teGalley)
-            withMockIdP (serveMetaAndResp metafile respstatus) $ do
-              callIdpCreate' (env ^. teSpar) (Just uid) newidp
-                `shouldRespondWith` checkErr (== 400) "client-error"
+            newidp <- makeTestNewIdP =<< liftIO UUID.nextRandom
```
neongreen: No idea what's happening here

services/spar/test-integration/Test/Spar/DataSpec.hs
```diff
@@ -316,18 +315,20 @@ mkAuthnReqMobile idpid = do
...
-                         -> (SAML.IdPId -> ReaderT TestEnv IO ResponseLBS)   -- raw authnreq
-                         -> ReaderT TestEnv IO ( UserId
-                                               , SAML.ResponseVerdict
-                                               , URI                         -- location header
-                                               , [(SBS, SBS)]                -- query params
-                                               )
-prepareAccessVerdictCore isGranted mkAuthnReq = do
-  (uid, _, idpid) <- createTestIdP
+requestAccessVerdict :: HasCallStack
+                     => Bool                                             -- is the verdict granted?
```
neongreen: `-- ^`

services/spar/test-integration/Util.hs
```diff
@@ -308,36 +325,55 @@ shouldRespondWith action proper = do
...
+-- registering with spar.  The issuer id can be used to do this several times without getting the
+-- error that this issuer is already used for another team.
+makeTestNewIdP :: (HasCallStack, MonadReader TestEnv m, MonadIO m) => UUID -> m NewIdP
+makeTestNewIdP issuerid = do
   env <- ask
-  liftIO . runHttpT (env ^. teMgr) $ do
-    (uid, tid) <- createUserWithTeam (env ^. teBrig) (env ^. teGalley)
-    (uid, tid,) . (^. SAML.idpId) <$> callIdpCreate (env ^. teSpar) (Just uid) sampleIdP
+  let Endpoint ephost (cs . show -> epport) = env ^. teTstOpts . to cfgMockIdp
+      mkurl = either (error . show) id . SAML.parseURI' . (("http://" <> ephost <> ":" <> epport) <>)
```
neongreen: If you're doing it like this then leave a comment in the config file saying that the host URL must not have a slash

services/spar/test-integration/Test/Spar/APISpec.hs
fisx: i specifically avoided that because there is a bit more text in there, and some of it really isn't crucial for the correctness, so i'd prefer to keep it like it is.

services/spar/test-integration/Test/Spar/APISpec.hs
fisx: i disagree with this, too.  i think `tid` is almost as bad as the one-letter local variables used in many places.

if you insist, i'll change it, though.

PR #459, opened 2018-08-27
Set LANG in all Makefiles.

fisx:
`find Makefile build/ tools/ services/ libs/ -name Makefile -exec perl -i -pe 'unless ($done) { $done = 1; print "LANG := en_US.UTF-8\n" }' {} \;`


neongreen:
Yes, this should work.

fisx:
I think concourse is acting up again...  :-/  (or did i break it in #452?)

fisx:
the concourse errors are not concourse's fault, but they will be handeled elsewhere: https://github.com/wireapp/wire-server/pull/462, possibly https://github.com/wireapp/wire-server/pull/461

PR #462, opened 2018-08-28
Work on integration tests

fisx:
- make some flaky integration tests less flaky (hopefully...)
- spar integration tests:
    - fix behavior on concourse
    - better error messages
    - assert availability of mock idp
    - make mock idp in spar integration tests serve via SSL

### DIFF
--- a/services/galley/test/integration/API/Teams.hs
+++ b/services/galley/test/integration/API/Teams.hs
@@ -21,6 +21,7 @@ import Data.Range
 import Galley.Types hiding (EventType (..), EventData (..), MemberUpdate (..))
 import Galley.Types.Teams
 import Galley.Types.Teams.Intra
+import GHC.Stack
 import Gundeck.Types.Notification
 import Test.Tasty
 import Test.Tasty.Cannon (Cannon, TimeoutUnit (..), (#))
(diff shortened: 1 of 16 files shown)


fisx:
mock idp should work with TLS now, see last two commits.  now we need to update cailleach again:

- re-align the equivalent of spar.integration.yaml
- re-align the equivalent of integration.yaml
- make sure the mock TLS credentials are mounted into the spar-integration container in some volume
- update staging config

services/galley/test/integration/API/Teams.hs
```diff
@@ -187,7 +188,6 @@ testCreateOne2OneWithMembers g b c a = do
         Util.addTeamMemberInternal g tid mem1
         checkTeamMemberJoin tid (mem1^.userId) wsMem1
         assertQueue "team member join" a $ tUpdate 2 [owner]
-        WS.assertNoEvent timeout [wsMem1]
```
tiago-loureiro: Hmmm, why was this removed?

fisx: sometimes, a `user.activated` events is getting in the way of this.  given that we create two users, we assumed that this is correct behavior and we would have to sort out a race condition.

alternatively, we could pick an event and make sure it is one of (1) not there and (2) the one we're seeing every now and then.

fisx:
there was a brig test failing.  nothing to do with this PR, i just want to check if it is non-deterministic.

services/galley/test/integration/API/Teams.hs
tiago-loureiro: 👍

PR #463, opened 2018-08-29
Allow clients to limit the number of turn servers on `/calls/config/v2`

jschaul:


libs/brig-types/src/Brig/Types/TURN.hs
```diff
@@ -258,3 +264,65 @@ instance ToJSON Transport where
...
+-- with order priority from highest to lowest: UDP -> TLS -> TCP
+-- i.e. (if enough servers of each type are available)
+--   1 -> 1x UDP
+--   2 -> 1x UDP, 1x TLS
+--   3 -> 1x UDP, 1x TLS, 1x TCP
+--   4 -> 2x UDP, 1x TLS, 1x TCP
+--   5 -> 2x UDP, 2x TLS, 1x TCP
+--    ... etc
+-- if not enough servers are available, prefer udp, then tls
+limitServers :: [TurnURI] -> Int -> [TurnURI]
```
tiago-loureiro: Not sure you prefer this "style" but since you asked :)
```
limitServers :: [TurnURI] -> Int -> [TurnURI]
limitServers uris = limitHeuristic [] (splitByType ([],[],[],[]) uris)
  where
    splitByType acc       [] = acc
    splitByType (a,b,c,d) (x:xs)
        | isUdp x   = splitByType (x:a, b  , c  ,  d) xs
        | isTls x   = splitByType (a  , x:b, c  ,  d) xs
        | isTcp x   = splitByType (a  , b  , x:c,  d) xs
        | otherwise = splitByType (a  , b  , c  ,x:d) xs
```

services/brig/src/Brig/TURN/API.hs
```diff
@@ -98,6 +108,15 @@ newConfig env = do
         (f:fs) <- shuffleM (toList xs)
         return $ List1.list1 f fs
 
+    limitedList :: List1 TurnURI -> Int -> List1 TurnURI
```
tiago-loureiro: How about changing `limitedList` to `limitedList :: List1 TurnURI -> Range 1 10 Int -> List1 TurnURI` and call `fromRange` within the function?

jschaul: sure!

libs/brig-types/src/Brig/Types/TURN.hs
jschaul: Ah, thanks! Yes, I prefer that style.

tiago-loureiro:
So I think that `limitServers` could be rewritten as:
```
limitServers :: [TurnURI] -> Int -> [TurnURI]
limitServers uris lim = limitServers' [] lim uris
  where
    limitServers' acc 0 _     = acc -- Already have accumulated enough
    limitServers' acc _ []    = acc -- No more input
    limitServers' acc _ input = do
        let (udps, noUdps) = partition isUdp input
            (udp, forTls)  = (Prelude.take 1 udps, noUdps ++ drop 1 udps)

            (tlss, noTlss) = partition isTls forTls
            (tls, forTcp)  = (Prelude.take 1 tlss, noTlss ++ drop 1 tlss)

            (tcps, noTcps) = partition isTcp forTcp
            (tcp, rest)    = (Prelude.take 1 tcps, noTcps ++ drop 1 tcps)

            new = udp ++ tls ++ tcp
            newAcc = Prelude.take lim $ acc ++ new
        if null new -- Did we find anything interesting? If not, time to go
            then Prelude.take lim $ acc ++ rest
            else limitServers' newAcc (lim - length newAcc) rest
```
Up for debate whether it's more or less readable - it can likely be slightly improved too.

PR #466, opened 2018-08-31
Refactor NewIdP; sort out dsig cert mess.

fisx:
# changes

- remove `NewIdP` in favor of `IdPMetadata`.
- no more metadata URL; all metadata is uploaded by team admin.
- distinguish between metadata certs and authnresponse certs.
- ignore metadata certs altogether (metadata is uploaded through trusted channel and no DSig authentication is needed).
- allow for more than one authnresponse cert.  it's insane, but at least azure provdes more than one cert with any obvious means of telling which will be used.
- to account for above changes:
    - update `validateNewIdP`.
    - update `/services/integration.yaml`.
    - update C* schema.  (note that `frozen<blob>` was rejected by C*; `blob` caused runtime errors; `list<blob>` works fine, at least for the tests I have.)
    - update test cases; eliminate some that no longer apply.  elimiinate mocking (see https://github.com/wireapp/wire-server/pull/466#issuecomment-419396359 for a rationale)

# changes *without the last commit*

- NewIdP type is less redundant now.
- update `/services/integration.yaml` to account for changes in NewIdP type.
- update C* schema to allow for more than one dsig cert for
  authentication response signing.  (at least azure does this, and
  there is no way of knowing which certs will be used, so we need to
  try all of them.)
- update validateNewIdP to account for the above changes.
- mock idp now responds with fresh issuer name on every /meta request.
- remove test case that is no longer valid (cert list for signing authn
  responses should not contain the cert for signing metadata)


fisx:
(Ready for review I guess, but I don't want to merge this before Monday, as it requires changes ci and staging config, and I would like to have some backup people around to fix things when I'm done breaking them.)

fisx:
(this also depends on repo `saml2-web-sso`, branch `tweaks-2018-08-30`.)

neongreen:
ping me when this is ready for review

fisx:
@tiago-loureiro:  what we discussed earlier today won't work: the metadata document does not contain the url under which it can be reached, so we would need to maintain that separately.  in the last commit [here](https://github.com/wireapp/wire-server/pull/466/commits/92a70cda45ac77994e39b18df20f3eba82864dfe) and [in saml2-web-sso](https://github.com/wireapp/saml2-web-sso/commit/fa0e35bed861752152f3c049fab263d30d9953ab), resp., i have tried to get rid of both metadata verification and fetching from idp.  i think this is legal, a great simplification, and i want to go for this now.

let me know if you're ok with this.

@neongreen thanks!  this won't get done today after all, but i'll let you know.

fisx:
I propose to merge this PR without squashing, so we can point to https://github.com/zinfra/cailleach/commit/acfa2beb5fc2076468d0e1b671abb1ac1dbc1044 from somewhere in the code where we'll look if it should ever become necessary for spar to talk to IdPs directly.

If we decide against the last commit, I would have to spend a bit of extra work fixing some (small?) issues with the mock idp or the production code for metadata validation, and then everybody using this feature would have to spend more time figuring out how to authenticate the metadata that spar would pull from a URL the admin provides.  I move to not do that.

@fisx you should probably confirm that at least for okta, azure AD, centrify, it is straight-forward for the team admin to extract the metadata XML and paste it into wire.

@neongreen this is ready for review, but feel free to stall it until I'm back so we can look at it together.

jschaul:
I admit I don't quite follow the reasoning that led to not needing a mockIdp anymore - can you explain?

(about squashing or not squashing, perhaps the commits prior to the `Simplify` one could be squashed, keeping that one separate is of course fine if it may get reverted in the future)

tiago-loureiro:
> I admit I don't quite follow the reasoning that led to not needing a mockIdp anymore - can you explain?

I had the same question 🤔

fisx:
> I admit I don't quite follow the reasoning that led to not needing a mockIdp anymore - can you explain?

The IdP has two end-points that cover the features we want: `get-metadata` and `get-authn-response`.

- `get-authn-response` can mocked by the client (ie., the test suite): it is just a function returning a signed xml document from another xml document and some randomness.  If we implemented this as a mock service, we would just make the structure of the code that tests production code more complicated, but the production code would never touch the mokc service.  so i want to implement it as a function in some monad that is called from the test suite.
- `get-metadata` is intended for the SP to talk to, but i want to go for the simplest solution that works (simple not only in backend code, but also in client code and complexity of user instructions).  so instead of configuring a metadata url in the team admin screens, i want to require the team admin to upload the metadata every time it changes.  this is expected to happen once every few years, or months at most, and it is a simple, easy-to-understand procedure.  **why is sp fetching metadata from idp complex?** we can't rely on the xml we pull from the web to be intact, so we need an independent source of the certificate used for signing.  getting that certificate out of the idp admin UI may be not supported, and it may become necessary for the admin to cut & paste it out of the metadata herself, which is silly considering the url is supposed to make handling the actual metadata unnecessary.  it is not clear when the updates should be pulled, and we are likely to need some button for the admin to trigger an update.  if we update, and the keys for signing metadata change, there may be interesting corner cases, like "we can't verify the new metadata, could you upload the new certificate manually please?".

that's why i think removing support for metadata urls is by far the best option.

thanks for making me explain this to myself again!  :-).  let me know if i actually answered your question.

jschaul:
Ok, thanks!

> i want to require the team admin to upload the metadata every time it changes

Where does the team admin get this information from; is this a copy-paste procedure from one UI to the next?

fisx:
> Where does the team admin get this information from; is this a copy-paste procedure from one UI to the next?

yes, that's correct.  i don't see any other option to that either.  web interfaces is how all these services get touched at all.  makes you shiver when you think about it.

services/spar/test-integration/Util.hs
```diff
@@ -356,25 +346,30 @@ ping :: (Request -> Request) -> Http ()
 ping req = void . get $ req . path "/i/status" . expect2xx
 
 
--- | Create a cloned 'NewIdP' corresponding to the mock idp from the config and suitable for
+makeIssuer :: MonadIO m => m Issuer
+makeIssuer = do
+  uuid <- liftIO UUID.nextRandom
+  either (liftIO . throwIO . ErrorCall . show)
+         (pure . Issuer)
+         (SAML.parseURI' ("https://issuer/_" <> UUID.toText uuid))
```
neongreen: should this be `issuer` or `issuer.net`?

services/spar/test-integration/Util/Types.hs
```diff
@@ -53,31 +49,30 @@ import SAML2.WebSSO.Types.TH (deriveJSONOptions)
...
-  , _teOpts        :: Opts
-  , _teTstOpts     :: IntegrationConfig
+  , _teOpts        :: Opts               -- ^ spar config
+  , _teTstOpts     :: IntegrationConfig  -- ^ integration test config
+
+    -- user, team, idp details created on spar:
+  , _teMetadata    :: SAML.IdPMetadata   -- ^ used for registering the mock idp
+  , _teUserId      :: UserId             -- ^ owner of the idp's home team
+  , _teTeamId      :: TeamId             -- ^ home team of the idp
+  , _teIdP         :: IdP                -- ^ details of the idp
```
neongreen: doesn't `IdP` contain `IdPMetadata`? in which case `_teMetadata` is redundant

services/spar/test-integration/Test/Spar/APISpec.hs
```diff
@@ -140,6 +137,13 @@ spec = do
           it "logs out user B, logs in user A" $ do
             pending
 
+        context "more than one dsig cert" $ do
+          it "accepts the first of two certs for signatures" $ do
+            pending
+
+          it "accepts the second of two certs for signatures" $ do
+            pending
```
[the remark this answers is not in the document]
fisx: I'm working on implementing all pending tests in a different PR, but would like to get spar and QA testing up and working again before i'm done with that.

Let me know if you want to insist, then I could postpone that and do it all in one step later this week.

services/spar/test-integration/Util.hs
fisx: right, it should be consistent with the earlier issuer.net.  will fix!

(beyond that it doesn't matter.  it has to be a valid URL, but it is never used as anything but an identifier.)

PR #469, opened 2018-09-16
Extra muted conversation field

tiago-loureiro:
Currently, clients can only set 2 values related to muting a conversation: on/off and a _ref_ which is a reference of when a conversation was muted (a timestamp).

This PR introduces a new field (_otr_muted_status_, implemented as an _int32_) which allows clients to set a more granular value, per conversation (e.g., 0 may mean mute certain message types and 1 would mean mute other messages or calls). The meaning of each value must be agreed upon between different clients and the backend will make no validation of this value (other than being an int32).

Some TODO's related to _muting conversations_ and its consequences on _fallback notifications_ were added inline as comments.

### DIFF
--- a/libs/galley-types/src/Galley/Types.hs
+++ b/libs/galley-types/src/Galley/Types.hs
@@ -313,6 +313,8 @@ data OtherMember = OtherMember
 instance Ord OtherMember where
     compare a b = compare (omId a) (omId b)
 
+-- Inbound member updates.  This is what galley expects on its endpoint.  See also
+-- 'MemberUpdateData'.
 data MemberUpdate = MemberUpdate
     { mupOtrMute       :: !(Maybe Bool)
     , mupOtrMuteStatus :: !(Maybe MutedStatus)
@@ -390,6 +392,8 @@ data Connect = Connect
     , cEmail     :: !(Maybe Text)
     } deriving (Eq, Show)
 
+-- Outbound member updates.  Used for events (sent over the websocket, etc.).  See also
+-- 'MemberUpdate'.
 data MemberUpdateData = MemberUpdateData
     { misOtrMuted       :: !(Maybe Bool)
     , misOtrMutedStatus :: !(Maybe MutedStatus)
--- a/libs/galley-types/src/Galley/Types.hs
+++ b/libs/galley-types/src/Galley/Types.hs
@@ -46,6 +46,7 @@ module Galley.Types
     , NewConvManaged            (..)
     , NewConvUnmanaged          (..)
     , MemberUpdate              (..)
+    , MutedStatus               (..)
     , TypingStatus              (..)
     , UserClientMap             (..)
     , UserClients               (..)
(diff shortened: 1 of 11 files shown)


libs/galley-types/src/Galley/Types.hs
```diff
@@ -867,6 +879,7 @@ instance FromJSON Member where
         Member <$> o .:  "id"
                <*> o .:? "service"
                <*> o .:? "otr_muted"        .!= False
+               <*> o .:? "otr_muted_status"
```
fisx: [unrelated to this PR, but perhaps easy to fix]

what's the difference between `MemberUpdate` and `MemberUpdateData`?  why do we need `ToJSON` instances for these?  can you explain this in haddocks near the code?

libs/galley-types/src/Galley/Types.hs
```diff
@@ -286,10 +288,16 @@ newtype Accept = Accept
...
+-- the server will not interpret this value in any way.
+newtype MutedStatus = MutedStatus { fromMutedStatus :: Int32 }
+    deriving (Eq, Num, Ord, Show, FromJSON, ToJSON)
+
 data Member = Member
     { memId             :: !UserId
     , memService        :: !(Maybe ServiceRef)
-    , memOtrMuted       :: !Bool
+    , memOtrMuted       :: !Bool -- ^ DEPRECATED, remove it once enough clients use `memOtrMutedStatus`
+    , memOtrMutedStatus :: !(Maybe MutedStatus)
```
fisx: Drop the `Maybe`?  If not, please explain difference between `Nothing` and `Just 0` in comment.

tiago-loureiro: The `Maybe` is here by design; if it's a nothing, then clients have not yet set this value which is the reason why we don't want to use a default; `Just 0` could mean `all message types muted` or it could also mean `no message types muted`.

fisx: Ok, I believe you, just trying to understand: `int` in cassandra can be `NULL`, which means that all bits have their default, which we don't want to know.  Yes, that makes sense, thanks!

libs/galley-types/src/Galley/Types.hs
```diff
@@ -392,6 +392,7 @@ data Connect = Connect
 
 data MemberUpdateData = MemberUpdateData
     { misOtrMuted       :: !(Maybe Bool)
+    , misOtrMutedStatus :: !(Maybe MutedStatus)
```
fisx: You probably would have caught that first-try had there been a comment discussing the difference between this and `MemberUpdate` on both types.  :-)

tiago-loureiro: ;)

libs/galley-types/src/Galley/Types.hs
tiago-loureiro: `MemberUpdateData` is used for events (sent over the websocket, etc.) while `MemberUpdate` is what is expected by `galley` on the endpoint.

That being said, thanks for the question since it brought up an incomplete implementation on my end!

PR #471, opened 2018-09-18
upgrade stack in alpine-prebuilder, alpine-builder

fisx:
I *think* this is what caused [the ci failiue on #470](https://bridge-ie-concourse.zinfra.io/builds/5610), even though I have no idea why this hasn't been caught a log earlier.

### DIFF
--- a/build/alpine/Dockerfile.prebuilder
+++ b/build/alpine/Dockerfile.prebuilder
@@ -97,7 +97,7 @@ RUN cd /tmp/ghc && \
     rm -rf /tmp/ghc
 
 # Upgrade stack (there is no binary for stack-1.7.1 on alpine, so this is the easiest way.)
-run stack update && \
+RUN stack update && \
     stack upgrade && \
     rm -f /usr/local/bin/stack && \
     cp /root/.local/bin/stack /usr/local/bin/stack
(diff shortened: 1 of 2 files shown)


fisx:
(nope, probably not.  libxml2-dev is added in prebuilder.)

fisx:
Ah, builder is running every time a PR is passing through the CI.  In that case this is not an option at all.  I'm going to try to build a new prebuilder.

build/alpine/Dockerfile.prebuilder
```diff
@@ -95,3 +95,9 @@ RUN cd /tmp/ghc && \
     make install &> /dev/null && \
     mv /tmp/config.yaml /root/.stack/ && \
     rm -rf /tmp/ghc
+
+# Upgrade stack (there is no binary for stack-1.7.1 on alpine, so this is the easiest way.)
+run stack update && \
```
jschaul: Is there a reason you do the stack upgrade in the pre-builder, rather than the builder image? Can you use uppercase `RUN` for consistency?

fisx: we're building builder more often, so it saves time to do it here, rather than once a week.  (earlier i thought it's once a ci run, but the argument still holds, even though it is weaker.  i still don't see a strong reason to have the distinction to begin with: i just build a new prebuilder in 45 minutes.)

yes, i'll fix the casing, thanks.

fisx:
> Also, for the sake of reproducibility, how about adding a --binary-version 1.7.1 argument to stack upgrade ?

good idea.  i've also changed the ARGs a little, i hope i have not overlooked any place where this is passed with anything below 1.7.1.

PR #473, opened 2018-09-25
SCIM integration

neongreen:


services/spar/src/Spar/API.hs
```diff
@@ -264,11 +266,18 @@ validateNewIdP _idpMetadata _idpeTeam = do
...
 -- Here we assume the 'spar' service is only accessible from behind the 'nginz' proxy, which
 --   * does not expose routes prefixed with /i/
 --   * handles authorization (adding a Z-User header if requests are authorized)
 --   * does not show the swagger end-point itself
-type OutsideWorldAPI = StripSwagger (StripInternal (StripAuth API))
+--
+-- In addition, we strip the SCIM documentation because SCIM doesn't have
+-- Swagger schemas yet.
+--
+-- TODO: consider exposing the SCIM docs as well.
```
fisx: Yes, i think just having the generic stuff would already by quite helpful.

services/spar/test/Test/Spar/APISpec.hs
```diff
@@ -25,3 +26,4 @@ spec = do
   validateEveryToJSON (Proxy :: Proxy API.IdpGetAll)
   validateEveryToJSON (Proxy :: Proxy API.IdpCreate)
   validateEveryToJSON (Proxy :: Proxy API.IdpDelete)
+  -- SCIM (TODO: do we want to test SCIM schemas here?)
```
fisx: well, anything that can't be tested in hscim?

services/spar/src/Spar/API.hs
```diff
@@ -71,6 +72,7 @@ type API
      = "sso" :> APISSO
   :<|> "identity-providers" :> APIIDP
   :<|> "i" :> APIINTERNAL
+  :<|> "scim" :> "v2" :> SCIM.API
```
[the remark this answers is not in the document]
neongreen: This is the current order 🤔

services/spar/src/Spar/API.hs
neongreen: In this case do you think Swagger schemas should live in hscim, or be defined in Spar? (My vote is for the former but maybe there are some drawbacks)

services/spar/src/Spar/API.hs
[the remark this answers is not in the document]
fisx: still asleep...  let me try again:

My favorite order is sso, identity-providers, scim, i.

services/spar/src/Spar/API.hs
fisx: i agree, they should live in hscim.  if there are any problems with that neither of us can think of, we can refactor later.

neongreen:
The order of the endpoints has been changed locally. I made a note to add Swagger schemas to hscim, but that won't be done in this PR.

stack.yaml
```diff
@@ -122,6 +122,10 @@ packages:
     git: https://github.com/wireapp/hspec-wai
     commit: ca10d13deab929f1cc3a569abea2e7fbe35fdbe3  # https://github.com/hspec/hspec-wai/pull/49
   extra-dep: true
+- location:
+    git: https://github.com/wireapp/hscim
+    commit: a33b475721ba7359a7726bd7b48ed6b6b07acf85
```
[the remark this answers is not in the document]
neongreen: Done.

services/spar/src/Spar/SCIM.hs
```diff
@@ -0,0 +1,304 @@
...
+instance SCIM.UserDB (ReaderT TeamId Spar) where
+  -- | List all users, possibly filtered by some predicate.
+  list mbFilter = do
+    tid <- ask
+    members <- lift $ getTeamMembers tid
+    users <- forM members $ \member ->
+      lift (getUser (member ^. Galley.userId)) >>= \case
+        Just user -> pure (toSCIMUser user)
+        Nothing -> SCIM.throwSCIM $
+          SCIM.serverError "Database is inconsistent"
```
fisx: Put some more details in here?

```haskell
"DB inconsistent: user " <> show userid <> " exists in galley, but not in brig."
```

I guess we need to worry about this leaking to the UI if we do it, but I think even if that happened it would be preferable to us starting at "you got it all wrong"-style errors.

services/spar/src/Spar/SCIM.hs
```diff
@@ -0,0 +1,304 @@
...
+    let check user = case mbFilter of
+          Nothing -> pure True
+          Just filter_ ->
+            let user' = SCIM.Common.value (SCIM.thing user)
+            in case SCIM.filterUser filter_ user' of
+                 Right res -> pure res
+                 Left err  -> SCIM.throwSCIM $
+                   SCIM.badRequest SCIM.InvalidFilter (Just err)
+    -- TODO: once bigger teams arrive, we should have pagination here.
+    SCIM.fromList <$> filterM check users
```
fisx: I'm curious, but too lazy to find out myself: is there a way to push down filtering into the cql statement and run it on cassandra?  If so that would be worth a comment here I guess (not really a TODO, just a remark).

services/spar/src/Spar/SCIM.hs
```diff
@@ -0,0 +1,304 @@
...
+    extId <- maybe (error "We need an externalId!") pure $
+               SCIM.User.externalId user
+    handl <- maybe (error "Failed to parse userName!") pure $
+               parseHandle (SCIM.User.userName user)
+    -- We check the name for validity, but only if it's present
+    mbName <- forM (SCIM.User.displayName user) $ \n ->
+      either error (pure . Name . fromRange) $
+      checkedEitherMsg @_ @1 @128 "displayName" n
+    -- NB: We assume that checking that the user does _not_ exist has
+    -- already been already done before -- the hscim library check does a
```
fisx: `s/already//`

services/spar/src/Spar/SCIM.hs
```diff
@@ -0,0 +1,304 @@
...
+        SCIM.notFound "user" uidText
+    lift (getUser uid) >>= traverse (\user -> do
+      when (userTeam user /= Just tid) $ SCIM.throwSCIM $
+        SCIM.notFound "user" (idToText uid)
+      pure (toSCIMUser user))
+
+  -- | Create a new user.
+  create user = do
+    tid <- ask
+    -- TODO: report these errors as SCIM errors
```
fisx: why is this non-trivial?

services/spar/test-integration/Util/Types.hs
```diff
@@ -74,9 +76,13 @@ data TestEnv = TestEnv
   , _teTstOpts     :: IntegrationConfig  -- ^ integration test config
 
     -- user, team, idp details created on spar:
+    -- TODO: rename to _teOwnerId
```
fisx: :+1:

services/spar/test-integration/Test/Spar/SCIMSpec.hs
```diff
@@ -0,0 +1,261 @@
...
+        it "lists all users in a team" $ do
+            env <- ask
+            -- Create a user via SCIM
+            user <- randomUser
+            storedUser <- createUser user
+            -- Get all users via SCIM
+            users <- listUsers Nothing
+            -- Check that the SCIM user is present
+            liftIO $ users `shouldContain` [storedUser]
+            -- Check that the (non-SCIM-provisioned) team owner is present
```
fisx: That feels wrong.  Why would we want that?

services/spar/test-integration/Test/Spar/SCIMSpec.hs
```diff
@@ -0,0 +1,261 @@
...
+            liftIO $ users `shouldSatisfy`
+                any ((== env^.teUserId) . scimUserId)
+        it "finds a SCIM-provisioned user by username" $
+            pending
+        it "finds a non-SCIM-provisioned user by username" $
```
fisx: Same here.

services/spar/test-integration/Test/Spar/SCIMSpec.hs
```diff
@@ -0,0 +1,261 @@
...
+        it "finds a SCIM-provisioned user" $ do
+            -- Create a user via SCIM
+            user <- randomUser
+            storedUser <- createUser user
+            -- Check that the SCIM-provisioned user can be fetched
+            storedUser' <- getUser (scimUserId storedUser)
+            liftIO $ storedUser' `shouldBe` storedUser
+        it "finds a non-SCIM-provisioned user" $ do
+            pending
+            {- TODO: this test fails because the user has no handle
```
fisx: if you like, you can write `pendingWith "this test fails because..."` here.

fisx:
concourse says (in case you can't seem to be loading the page either; fly was more talkative) -

```
----LOGS FROM spar-c7c7b564b-2t9pq:
service: no or bad config file: AesonException "Error in $: key \"galley\" not present"
```

services/spar/src/Spar/API.hs
```diff
@@ -67,7 +68,7 @@ app ctx = SAML.setHttpCachePolicy
         $ serve (Proxy @API) (hoistServer (Proxy @API) (SAML.nt @SparError @Spar ctx) (api $ sparCtxOpts ctx) :: Server API)
 
 api :: Opts -> ServerT API Spar
-api opts = apiSSO opts :<|> apiIDP :<|> apiINTERNAL
+api opts = apiSSO opts :<|> apiIDP :<|> SCIM.api :<|> apiINTERNAL
```
neongreen: TODO: disable

services/spar/src/Spar/API/Types.hs
```diff
@@ -120,21 +122,32 @@ sparResponseURI = SAML.getSsoURI (Proxy @APISSO) (Proxy @APIAuthResp)
...
+-- | Strip the endpoint that exposes documentation.
 type family StripSwagger api where
-    StripSwagger ("sso" :> "api-docs" :> Get '[JSON] Swagger :<|> b) = StripSwagger b
+    StripSwagger ("sso" :> "api-docs" :> a) = EmptyAPI
     StripSwagger (a :<|> b) = StripSwagger a :<|> StripSwagger b
     StripSwagger x = x
+
+-- | Strip SCIM.
+type family StripSCIM api where
+    StripSCIM ("scim" :> a) = EmptyAPI
```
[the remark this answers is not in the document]
neongreen: I don't think it's hard, but I'd rather do it as a separate task. It's a nice principle to not put undocumented endpoints into `develop`, though, so I'm rather torn on this. If you have a strong heuristic/rule/something about that, I can reconsider.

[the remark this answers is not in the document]
fisx: since you asked: my heuristic is to not merge incomplete, broken features like this one.  :-)

i guess if you take the route out of the servant api type, this won't even be needed, no?  and then you get a type error if you ever enable it and are missing swagger instances?

services/spar/src/Spar/SCIM.hs
neongreen: It's possible to do filtering in Cassandra if you say "ALLOW FILTERING" in the query, which we don't do because generally we like doing explicit pagination instead of coercing Cassandra into doing full table scans.

services/spar/src/Spar/API/Types.hs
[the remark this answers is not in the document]
neongreen: > Was it hard to provide the default instances?

NB: I just tried that (just in case I'm wrong and it turns out to be easy). It's not. Or rather, we can provide default instances and generate very broken docs, but this is worse than no docs.

To do it right, we need to write a custom ToSchema instance for every ToJSON instance we have, and then to test it we need to write sane Arbitrary instances for hscim types (because that's what `validateEveryToJSON` requires), and finally we also need to ask servant-swagger to treat the `SCIM` content-type tag as the `JSON` content-type tag, because otherwise it doesn't even find anything to test.

And then nobody will look at those docs anyway, likely. That's also why I'm not very keen on doing that work.

I will, however, remove the StripSCIM instance and move it to my local stash, though it's less convenient than having it in the code.

neongreen:
NB: I have an infinite capacity for replying to questions without getting tired, which means that if you don't care about the answers much you should just unsubscribe from the thread (and then the answers will be just for reference of whoever happens to read the thread later).

services/spar/test-integration/Test/Spar/SCIMSpec.hs
neongreen: We don't want that at this stage, though we also don't necessarily want to _not_ have that. I see no usecase for hiding non-SCIM-provisioned users; do you?

In case you just object to it on the grounds of "we don't need it":

I'm somewhat sure that in the future we will need it, though, and I'm more than somewhat sure that we'll want two-way sync instead of one-way sync at some point. I'm okay with it being out of scope for now, but I still want some checks to be present so that half a year later we don't end up with an architecture (and implicit invariants all over the code) that prevent us from having two-way sync and so on.

For me, it's the same principle as "try to make your code reasonably generic even if you don't strictly need it", but applied to architecture.

Just to make it perfectly clear, I don't rule out the possibility that my heuristics are wrong. I'm just not willing to allow "we don't strictly need this now" be considered a hard-and-fast rule for not doing something.

neongreen: If it turns out that I have miscalculated the cost of supporting the usecase of "see non-SCIM-provisioned users via SCIM" (in the code, or in tests, etc), we can surely revisit it.

neongreen:
CI should pass now.

PR #474, opened 2018-09-28
Put OpenSSL paths for macOS into the nginz makefile

neongreen:
This should work on Linux as well and it should not harm anything if OpenSSL paths change on macOS in the future. However, I'd like somebody familiar with Makefile syntax to take a closer look.


neongreen:
@tiago-loureiro regarding portability: we use `$(shell)` all over the place so we seem to be tied to GNU Make anyway.

neongreen:
Or did you mean some other portability?

PR #475, opened 2018-09-28
Fix: authentication requests should ask for `unspecified` SubjectIDs.

fisx:


neongreen:
See my question in the saml PR.

In general, I am not qualified to say whether {some set of changes related to SAML} is a complete {set of changes required to accomplish some specific thing}, so I can't in good conscience approve such PRs. I will restrict myself to either commenting on them or rejecting them :trollface:

fisx:
> See my question in the saml PR.

which one?

neongreen:
Oh, sorry, it's not in the same PR, it's here:

> What if we used 8.3.7 Persistent Identifier instead? 🤔

I don't know how this happened.

neongreen:
(Link to documentation for other people who might be looking at this issue: http://docs.oasis-open.org/security/saml/v2.0/saml-core-2.0-os.pdf)

fisx:
> What if we used 8.3.7 Persistent Identifier instead?

Good idea, but Okta offers only "unspecified" and "email" out of the box.

It is very possible that we need to change this in the future.  Then we can keep it downwards-compatible with this PR by keeping "unspecified" the default, and allowing the browser to pick alternatives by adding a query param to the `initiate-login` end-point.

This change does not affect production, and it will hopefully allow me to get staging intact again tomorrow, so I will merge without reviewer approval.

PR #479, opened 2018-10-05
Fix stack yaml

fisx:


neongreen:
That's unfortunate timing :( The alpine-builder has *almost* finished.

I will kill the Concourse jobs created by this PR and then ask you to push an empty commit or amend this one.

neongreen:
Hm, or actually I should be able to just restart the job for you.

fisx:
> Hm, or actually I should be able to just restart the job for you.

Thanks!  :-)  (in case it matters, I won't need this before Monday.  Good night..)

neongreen:
@fisx do you mind if I also merge this after the tests pass?

PR #480, opened 2018-10-05
Allow setting a different endpoint for generating download links

tiago-loureiro:
Currently, the URIs that we hand out to clients to download asset point to an AWS endpoint, namely a S3 or a Cloudfront URL.

In the context of self hosting, and in case users want to host S3 "internally" using minio.io for instance, there should be a publicly addressable URL that cargohold can give out so that they are reachable from the outside world. For instance, let's say that we have the HTTPS endpoint at `wire-https.foo.bar` and `cargohold` is configured to talk to a self hosted minio.io at `http://fake-s3:9000`; obviously these URLs are not useful for clients. Instead, one would need to make an endpoint available at `https://wire-fake-s3.foo.bar` which simply proxies requests to minio.io.

In order to try this out yourselves, you can set `fake-s3` to `localhost` on your machines.


neongreen:
So where will the actual rewriting happen?

tiago-loureiro:
It happens when we sign the URL for downloads, i.e. here: https://github.com/wireapp/wire-server/blob/develop/services/cargohold/src/CargoHold/S3.hs#L181-L182

neongreen:
You can push an empty commit to re-trigger the CI.

PR #481, opened 2018-10-07
Allow setting specific ports for SMTP and use different image for SMTP

tiago-loureiro:
This PR introduces 2 changes:

1. Addresses an old comment, namely:
>     # if binding to port 25 is a problem,
>     # connection handling needs to be made more generic to accept
>     # alternative ports. See changes introduced in
>     # https://github.com/wireapp/wire-server/pull/405

Indeed it's a bit inconvenient to have to use privileged ports so it's always good to allow some flexibility here.

2. It changes the SMTP docker image which seems to work better than the current one; at least, I have been able to deliver emails without a problem with this solution. Note though that the communication between brig and this SMTP container should be done using `plain` protocol and then configure the container for the [desired scenario](https://github.com/namshi/docker-smtp#below-are-scenarios-for-using-this-container) - which is basically no arguments in our typical test-like use case.

### DIFF
--- a/deploy/docker-ephemeral/docker-compose.yaml
+++ b/deploy/docker-ephemeral/docker-compose.yaml
@@ -21,17 +21,10 @@ services:
       - SERVICES=ses,sns
 
   basic_smtp: # needed for demo setup
-    # https://github.com/catatnight/docker-postfix
-    image: catatnight/postfix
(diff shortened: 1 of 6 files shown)


neongreen:
@tiago-loureiro please merge

PR #482, opened 2018-10-08
Allow easily runnings tests against AWS

tiago-loureiro:
Currently we tend to test things only against fake/mocked AWS services; while this is nice from a feedback loop point of view, it would also be helpful to be able to test things against the actual AWS API and get that sort of feedback sooner rather than later.

This PR intends to streamline that workflow making some assumptions:
 * Defined in the environment: `AWS_REGION`, `AWS_ACCESS_KEY_ID`, `AWS_SECRET_ACCESS_KEY`. This `AWS_ACCESS_KEY_ID` has access to the resources specified in the configuration files
 * Access to a file with default settings to override settings in each service's config options. This file can either be downloaded from S3 (if you provide a path to it) or just added locally
 * Whatever AWS resources used, they must be created beforehand

If the preconditions are met, one can simply run `make i-aws` which is equivalent to `make integration` except that it attempts to use real AWS resources instead.


services/cargohold/Makefile
```diff
@@ -64,6 +64,10 @@ $(DEB_IT):
 i:
 	../integration.sh $(EXE_IT) -s $(NAME).integration.yaml -i ../integration.yaml
 
+.PHONY: i-aws
+i-aws:
+	INTEGRATION_USE_REAL_AWS=1 ../integration.sh $(EXE_IT) -s $(NAME).integration.yaml -i ../integration.yaml
```
jschaul: Did you intend to use `-s $(NAME).integration-aws.yaml` ? The script writes merges and writes to that file, but the Makefile doesn't use that. (same for all the Makefiles)

services/gen-aws-conf.sh
```diff
@@ -0,0 +1,19 @@
...
+
+DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )/" && pwd )"
+
+# Ensure that we have a file named integration-aws.yaml in the current
+# dir. If not, fetch it from a known location on S3
+if [ ! -f "${DIR}/integration-aws.yaml" ]
+then
+    echo "Could not find AWS config file to override settings, specify a location on S3 to download the file or add one at ${DIR}/integration-aws.yaml and retry: "
+    read location
+    aws s3 cp s3://$location .
```
jschaul: incorrect path. Should be `aws s3 cp s3://$location "${DIR}/integration-aws.yaml"`

jschaul:
Also, there are a few warnings in the bash script when running [`shellcheck`](https://github.com/koalaman/shellcheck) on it.

services/cargohold/Makefile
tiago-loureiro: Erm, yeah... rebase gone wrong :/ should be all green now

services/gen-aws-conf.sh
tiago-loureiro: 👍

tiago-loureiro:
> Also, there are a few warnings in the bash script when running shellcheck on it.

Oh nice one, thanks for the tip.

tiago-loureiro:
Should be much cleaner now, assumes that the `integration-aws.yaml` file is scoped per service (example added too).

PR #486, opened 2018-10-09
Always build benchmarks

neongreen:


fisx:
I'll see if I can fix it quickly.

fisx:
Hum, not sure what to do.  `import Prelude` makes it compile, but I don't think that's what you want.

Also, I got this when running `make` in `services/gundeck`:

```
Warning: Unable to find a known candidate for the Cabal entry V3, but did find: 
         * V3.hs~ If you are using a custom preprocessor for this module with its own file extension, consider adding the
         file(s) to your .cabal under extra-source-files.
```

?!  `V3.hs` is where it should be.

neongreen:
I'll look at it tomorrow or at some point sometime

neongreen:
It should work after the rebase. Regarding `V3.hs~`: it might be a temp file from your editor, try deleting it.

PR #488, opened 2018-10-09
SSO binding users

fisx:
Also, some other improvements:

- new module Spar.API.Types
- fight potential race conditions with `Control.Retry`.
- improve integration tests
- fixes & cleanup

questions to the reviewer:

- Should brig worry about `UserSSOId` being unique?  Spar certainly does, perhaps that's enough?  If not, is there a thing we can do to the cassandra schema that'll do this?
- What happens if the user that gets its `UserSSOId` updated in brig does not exist?  What *should* happen?

### DIFF
--- a/services/brig/src/Brig/API.hs
+++ b/services/brig/src/Brig/API.hs
@@ -1419,7 +1419,7 @@ updateSSOId :: UserId ::: JSON ::: JSON ::: Request -> Handler Response
 updateSSOId (uid ::: _ ::: _ ::: req) = do
     ssoid :: UserSSOId <- parseJsonBody req
     lift $ Data.updateSSOId uid ssoid
-    return (setStatus status200 empty)
+    return empty
 
 deleteUser :: UserId ::: Request ::: JSON ::: JSON -> Handler Response
 deleteUser (u ::: r ::: _ ::: _) = do
--- a/services/brig/src/Brig/API.hs
+++ b/services/brig/src/Brig/API.hs
@@ -1418,8 +1418,10 @@ isTeamOwner (uid ::: tid) = do
 updateSSOId :: UserId ::: JSON ::: JSON ::: Request -> Handler Response
 updateSSOId (uid ::: _ ::: _ ::: req) = do
     ssoid :: UserSSOId <- parseJsonBody req
-    lift $ Data.updateSSOId uid ssoid
-    return empty
+    status <- lift $ Data.updateSSOId uid ssoid
+    if status
+      then return empty
+      else return $ setStatus status404 empty
 
 deleteUser :: UserId ::: Request ::: JSON ::: JSON -> Handler Response
 deleteUser (u ::: r ::: _ ::: _) = do
--- a/services/brig/src/Brig/API.hs
+++ b/services/brig/src/Brig/API.hs
@@ -209,6 +209,8 @@ sitemap o = do
 
     put "/i/users/:uid/sso-id" (continue updateSSOId) $
       capture "uid"
+      .&. accept "application" "json"
+      .&. contentType "application" "json"
       .&. request
 
     post "/i/clients" (continue internalListClients) $
--- a/services/brig/src/Brig/API.hs
+++ b/services/brig/src/Brig/API.hs
@@ -1418,10 +1418,10 @@ isTeamOwner (uid ::: tid) = do
 updateSSOId :: UserId ::: JSON ::: JSON ::: Request -> Handler Response
 updateSSOId (uid ::: _ ::: _ ::: req) = do
     ssoid :: UserSSOId <- parseJsonBody req
-    status <- lift $ Data.updateSSOId uid ssoid
-    if status
+    success <- lift $ Data.updateSSOId uid ssoid
+    if success
       then return empty
-      else return $ setStatus status404 empty
+      else return . setStatus status404 $ plain "User does not exist or has no team."
 
 deleteUser :: UserId ::: Request ::: JSON ::: JSON -> Handler Response
 deleteUser (u ::: r ::: _ ::: _) = do
--- a/services/brig/src/Brig/API.hs
+++ b/services/brig/src/Brig/API.hs
@@ -59,6 +59,7 @@ import qualified Brig.API.Client               as API
 import qualified Brig.API.Connection           as API
 import qualified Brig.API.Properties           as API
 import qualified Brig.API.User                 as API
+import qualified Brig.Data.User                as Data
 import qualified Brig.Queue                    as Queue
 import qualified Brig.Team.Util                as Team
 import qualified Brig.User.API.Auth            as Auth
@@ -205,6 +206,10 @@ sitemap o = do
     get "/i/users/:uid/is-team-owner/:tid" (continue isTeamOwner) $
       capture "uid"
       .&. capture "tid"
+
+    put "/i/users/:uid/sso-id" (continue updateSSOId) $
+      capture "uid"
+      .&. request
 
     post "/i/clients" (continue internalListClients) $
       accept "application" "json"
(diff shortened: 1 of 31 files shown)


fisx:
it's ready for review now :-)  (note the questions in the description.)

services/brig/src/Brig/API.hs
```diff
@@ -1408,6 +1415,12 @@ isTeamOwner (uid ::: tid) = do
        Team.NoTeamOwnersAreLeft   -> throwStd insufficientTeamPermissions
     return empty
 
+updateSSOId :: UserId ::: JSON ::: JSON ::: Request -> Handler Response
+updateSSOId (uid ::: _ ::: _ ::: req) = do
+    ssoid :: UserSSOId <- parseJsonBody req
+    lift $ Data.updateSSOId uid ssoid
+    return (setStatus status200 empty)
```
tiago-loureiro: you can also simply `return empty` here

services/spar/src/Spar/Types.hs
```diff
@@ -29,6 +29,10 @@ import qualified Data.ByteString.Builder as Builder
 import qualified Data.Text as ST
 
 
+data Void
+
+type BindCookie = SimpleSetCookie "wire.com"
```
tiago-loureiro: Can we make name the cookie sth else plz, sth along the lines of `bind` or `zbind` or some other variant? (to keep in line with `zuid`.
Also, let's also restrict the cookie domain, which probably needs changes in the [web-saml-sso](https://github.com/wireapp/saml2-web-sso) ?

fisx: `zbind` it is.

fixing the domain is simple enough, but i already have `setCookieSameSite == True`, isn't that enough?  if you don't respond i'll just do both.

fisx: Done.  See also: https://github.com/wireapp/saml2-web-sso/blob/5758cb355a1a9956bb5d3ab2402f82b003a5a961/src/SAML2/WebSSO/Cookie.hs#L57-L76

services/brig/test/integration/Util.hs
```diff
@@ -151,18 +151,20 @@ getConnection brig from to = get $ brig
...
             [ "name"            .= name
             , "email"           .= email'
+            , "phone"           .= phone
             , "password"        .= defPassword
             , "cookie"          .= defCookieLabel
             , "sso_id"          .= ssoid
             , "team_id"         .= teamid
             ]
-    post (brig . path "/i/users" . contentJson . body p)
+    -- liftIO $ print (eitherDecode @NewUser (encode p))
```
tiago-loureiro: Can now be removed?

services/spar/src/Spar/App.hs
```diff
@@ -186,47 +200,70 @@ instance Intra.MonadSparToBrig Spar where
...
+
+      case (fromBindCookie, fromSparCass) of
+        (Nothing,  Nothing)   -> createUser userref      -- first sso authentication
+        (Nothing,  Just uid)  -> pure uid                -- sso re-authentication
+        (Just uid, Nothing)   -> bindUser uid userref    -- bind existing user (non-sso or sso) to ssoid
+        (Just uid, Just uid')
+          | uid == uid' -> pure uid                      -- redundant binding (no change to brig or spar)
+          | otherwise -> throwSpar SparBindUserRefTaken  -- attempt to use ssoid for a second wire user
+
+    cky :: SetCookie <- retrySparAction $ Intra.ssoLogin uid
```
tiago-loureiro: This should not be needed, given that there is a write Quorum in `createUser`; indeed it may be harmless but perhaps hiding other assumptions?

services/brig/test/integration/Util.hs
tiago-loureiro: 👍

services/spar/src/Spar/API.hs
```diff
@@ -165,12 +100,36 @@ authreqPrecheck msucc merr idpid = validateAuthreqParams msucc merr
...
-  pure form
+  cky <- initializeBindCookie zusr authreqttl
+  SAML.logger SAML.Debug $ "setting bind cookie: " <> show cky
+  pure $ addHeader cky form
+
+-- | Create bind cookie with 60 minutes life expectancy; *iff* the user is already authenticated,
+-- store it with the bind cookies.  If user already has a 'UserSSOId' (need to ask brig for that),
+-- throw an error.
+initializeBindCookie :: Maybe UserId -> NominalDiffTime -> Spar BindCookie
+initializeBindCookie zusr authreqttl = do
```
tiago-loureiro: I think we should initialise/set the path/domain during spar's startup; these values never change during runtime and this way you will see config errors only when users make requests which is a bit too late

tiago-loureiro:
> Should brig worry about UserSSOId being unique? Spar certainly does, perhaps that's enough?
> If not, is there a thing we can do to the cassandra schema that'll do this?

Yes, `spar` is the source of truth here and just _orders_ it; in `brig` one could only build a reverse index which is, in fact, spar.

> What happens if the user that gets its UserSSOId updated in brig does not exist? What should happen?

Oh, good point... we should protect against this; i.e., `brig` must verify in the update endpoint that a user with such UUID exists otherwise we end up creating a _phantom_ user that can't be parsed (for instance, we would create a user without a name which breaks our model). Can you please make that check @fisx

jschaul:
failing tests:

>       put /i/users/:uid/sso-id:                                        FAIL
>        Exception: Assertions failed:
>         1: 200 =/= 404
>        Response was:
>        Response {responseStatus = Status {statusCode = 404, statusMessage = "Not Found"}, responseVersion = HTTP/1.1, responseHeaders = [("Transfer-Encoding","chunked"),("Date","Thu, 11 Oct 2018 17:23:05 GMT"),("Server","Warp/3.2.22"),("Content-Encoding","gzip"),("Content-Type","text/plain; charset=UTF-8")], responseBody = Nothing, responseCookieJar = CJ {expose = []}, responseClose' = ResponseClose}
>        CallStack (from HasCallStack):
>          error, called at src/Bilge/Assert.hs:73:30 in bilge-0.22.0-pluR4tRfls2bNLdDifsiA:Bilge.Assert
>          <!!, called at src/Bilge/Assert.hs:90:19 in bilge-0.22.0-pluR4tRfls2bNLdDifsiA:Bilge.Assert
>         !!!, called at test/integration/API/User/Account.hs:948:13 in main:API.User.Account

>     sso
>      post /i/users  - 201 internal-SSO:                               FAIL
>        Exception: Error executing request: test/integration/Util.hs:(158,9)-(166,13): Irrefutable pattern failed for pattern Success (p :: NewUser)
>        CallStack (from HasCallStack):
>          error, called at src/Bilge/Assert.hs:83:18 in bilge-0.22.0-pluR4tRfls2bNLdDifsiA:Bilge.Assert            
>        <!!, called at src/Bilge/Assert.hs:90:19 in bilge-0.22.0-pluR4tRfls2bNLdDifsiA:Bilge.Assert
>          !!!, called at test/integration/API/Team.hs:569:5 in main:API.Team

services/spar/src/Spar/App.hs
```diff
@@ -172,8 +172,9 @@ bindUser buid userref = do
   unless (uteamid == Just teamid)
     (throwSpar . SparBindFromWrongOrNoTeam . cs . show $ uteamid)
   insertUser userref buid
-  Intra.bindUser buid userref
-  pure buid
+  Intra.bindUser buid userref >>= \case
+    True  -> pure buid
+    False -> throwSpar SparBindUserDisappearedFromBrig
```
tiago-loureiro: Can we also log something at least as warning here? This should really never happen and so if it does and we only throw a 404, we're likely to miss this.

fisx:
(I just ran all integration tests locally, so I'll just merge without waiting for the CI.)

PR #491, opened 2018-10-12
Cleanup

fisx:


Makefile
```diff
@@ -89,10 +89,8 @@ DOCKER_DEV_VOLUMES := -v `pwd`:/src/wire-server
 DOCKER_DEV_IMAGE   := quay.io/wire/alpine-builder:local
 .PHONY: run-docker-builder
 run-docker-builder:
-	docker run -it $(DOCKER_DEV_NETWORK) $(DOCKER_DEV_VOLUMES) --rm $(DOCKER_DEV_IMAGE) /bin/bash || \
-	( echo "$(DOCKER_DEV_IMAGE) not found.  building locally.  hit ^C to interrupt." && \
-	  make -C build/alpine builder && \
-	  make $@ )
+	@echo "if this does not work, consider 'docker pull', 'docker tag', or 'make -C build-alpine builder'."
```
tiago-loureiro: What's wrong with the previous approach?

fisx: quoting from the commit message:

> docker exited with non-0 here in situations where you didn't want to trigger a local build and then re-run.  now it's more manual, but also easier to understand what the rule does

i don't remember what exactly those conditions were, but i remember being annoyed when i tried to exit the container and it restarted immediately.

is the "the new thing is simpler" argument enough?

tiago-loureiro: > docker exited with non-0 here in situations where you didn't want to trigger a local build and then re-run.

Makes sense, yeah 👍

PR #495, opened 2018-10-17
[WIP] New cassandra

jschaul:
Upgrade to the lastest `cql-io`. There are multiple reasons for upgrading, see [the changelog](https://gitlab.com/twittner/cql-io/blob/develop/CHANGELOG#L1-23). 

One reason is to try out the changes from [this MR](https://gitlab.com/twittner/cql-io/merge_requests/14) relating to the problem described [here](https://gitlab.com/twittner/cql-io/issues/21). To this end, `initialContactsDNS` is no longer used, so that cql-io can re-resolve the DNS upon losing a control connection.

TODO: This PR is marked as WIP since the changes to [the retry policies](https://gitlab.com/twittner/cql-io/issues/13) should get another look. Instead of our `x1` and `x5` retry policies used everywhere, perhaps we should switch to `defRetrySettings` and `eagerRetrySettings`. `eagerRetryHandlers`, however:
> -- are only safe to use for idempotent
> -- queries, or if a duplicate write has no severe consequences in
> -- the context of the application's data model.

Side-effects:

* switch from `MonadBaseControl` and `Control.Concurrent.*` to `UnliftIO.*` everywhere (thanks @neongreen).

### DIFF
--- a/libs/cassandra-util/src/Cassandra/Schema.hs
+++ b/libs/cassandra-util/src/Cassandra/Schema.hs
@@ -40,7 +40,7 @@ import Data.Time.Clock
 import Data.UUID (UUID)
 import Data.Word
 import Database.CQL.IO
-import Database.CQL.Protocol (Request (..), Query (..), Response(..), Result (..))
+import Database.CQL.Protocol (Request (..), Query (..))
 import GHC.Generics hiding (to, from, S, R)
 import Options.Applicative hiding (info)
 import Prelude hiding (log)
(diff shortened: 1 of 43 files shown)


libs/cassandra-util/src/Cassandra/Schema.hs
```diff
@@ -133,21 +133,15 @@ createKeyspace (Keyspace k) rs = void $ schema (cql rs) (params All ())
     pair (dc, ReplicationFactor n) = "'" <> dc <> "': " <> pack (show n)
 
 useKeyspace :: Keyspace -> Client ()
-useKeyspace (Keyspace k) = do
-    r <- qry
-    case r of
-        RsResult _ (SetKeyspaceResult _) -> return ()
-        RsError _ e                      -> throwM e
-        _                                -> throwM (UnexpectedResponse' r)
+useKeyspace (Keyspace k) = void . getResult =<< qry
```
tiago-loureiro: Didn't have a look at the internal changes in cql-io, is it OK to just ignore the result of `qry`?

jschaul: `getResult` will throw an error, if any:

```haskell
-- | Get the 'Result' out of a 'HostResponse'. If the response is an 'RsError',
-- a 'ResponseError' is thrown. If the response is neither
-- 'RsResult' nor 'RsError', an 'UnexpectedResponse' is thrown.
getResult :: MonadThrow m => HostResponse k a b -> m (Result k a b)
```

PR #497, opened 2018-10-23
[WIP] Fix reauth without password

fisx:
galley end-points for deleting team members and deleting teams, resp., ask for a password even if the user calling the end-point doesn't have one.  we change the behavior to be in sync with the end-points for adding and deleting devices.


tiago-loureiro:
> like af46dcf?

Exactly

PR #502, opened 2018-10-29
[WIP] SSO: fix bind cookie handling

fisx:
There are corner cases in which SSO login can fail due to bind cookie checks.  Since binding existing users is not a feature yet, this patch disables bind cookies entirely, cleaning up the corner cases.

This is a work-around, but a solid one without any expected drawbacks.

This also fixes a silly issue with integration tests.


fisx:
I will try to provide a proper fix in a separate PR that will make this unnecessary.  Meanwhile I'll leave this PR here open in case there is a delay there.

Btw, I disagree with the idea that making the availability of a route configurable is cleaner than removing it from the type and leaving the code lying there for the upcoming fix.

fisx:
set to WIP.  sorry for the noise.

fisx:
closed in favour of #505

PR #503, opened 2018-11-02
exectuable -> executable

neongreen:
🙈


jschaul:
> (How did it even work, actually?)

I suppose `cp fileA directoryB/` is the same as `cp fileA directoryB/fileA` (and the misspelled argument results in the empty string) . Thanks for the typo fix.

PR #505, opened 2018-11-06
SSO: fix binding users

fisx:
Obsoletes #502


fisx:
This was not my best work ever, but I think I'm close now.  Would anybody like to pair on the remaining bug that is likely in the test case?

fisx:
ok, i fixed the last issues i know about.  ready for review!

services/spar/src/Spar/API.hs
```diff
@@ -147,8 +147,8 @@ validateRedirectURL uri = do
     throwSpar $ SparBadInitiateLoginQueryParams "url-too-long"
 
 
-authresp :: Maybe BindCookie -> SAML.AuthnResponseBody -> Spar Void
-authresp cky = SAML.authresp sparSPIssuer sparResponseURI $
+authresp :: Maybe ST -> SAML.AuthnResponseBody -> Spar Void
```
neongreen: Hm, if the `BindCookie` type remains, why isn't it used here?

services/spar/src/Spar/Orphans.hs
```diff
@@ -0,0 +1,45 @@
...
+instance ToJSON ServantErr
+
+instance MimeRender PlainText Void where
+  mimeRender _ = error "instance MimeRender HTML Void: impossible"
+
+instance FromJSON (CI SBS) where parseJSON = fmap mk . parseJSON
+instance ToJSON (CI SBS)   where toJSON = toJSON . original
+
+instance FromJSON SBS      where parseJSON = fmap (cs @ST @SBS) . parseJSON
+instance ToJSON SBS        where toJSON = toJSON . cs @SBS @ST
```
neongreen: These instances only work when you have UTF-8-encoded strings. If it's necessary to have them around, at least we can base64-encode the bytestrings. If this is impossible, and having `Text` instead of `ByteString` inside of the relevant datatypes is impossible either, then I'd rather have manually written instances – this way nobody will get bitten.

services/spar/test-integration/Test/Spar/APISpec.hs
```diff
@@ -384,29 +438,69 @@ specBindingUsers = describe "binding existing users to sso identities" $ do
...
+        let check :: HasCallStack => (Cky.Cookies -> Maybe Cky.Cookies) -> Bool -> SpecWith TestEnv
+            check tweakcookies bindsucceeds = do
+              it (if bindsucceeds then "binds existing user" else "creates new user") $ do
+                (uid, _, idp) <- createTestIdP
+                (subj :: NameID, sparrq, sparresp) <- initialBind' tweakcookies uid idp
+                checkGrantingAuthnResp' sparresp
+
+                uid' <- getUserIdViaRef $ UserRef (idp ^. idpMetadata . edIssuer) subj
+                checkGrantingAuthnResp uid' sparrq sparresp
+                liftIO $ if bindsucceeds
```
neongreen: This is duplicated.

services/spar/test-integration/Test/Spar/APISpec.hs
```diff
@@ -384,29 +438,69 @@ specBindingUsers = describe "binding existing users to sso identities" $ do
...
+                  then uid' `shouldBe` uid
+                  else uid' `shouldNotBe` uid
+                liftIO $ (if bindsucceeds then shouldBe else shouldNotBe) uid' uid
+
+            addAtBeginning :: Cky.SetCookie -> Cky.Cookies -> Cky.Cookies
+            addAtBeginning cky = ((Cky.setCookieName cky, Cky.setCookieValue cky):)
+
+            addAtEnd :: Cky.SetCookie -> Cky.Cookies -> Cky.Cookies
+            addAtEnd cky = addAtBeginning cky . reverse
```
neongreen:
```suggestion
            addAtEnd cky = (++ [(Cky.setCookieName cky, Cky.setCookieValue cky)])
```

services/spar/src/Spar/Types.hs
```diff
@@ -44,7 +45,18 @@ import qualified SAML2.WebSSO as SAML
 
 data Void
 
-type BindCookie = SimpleSetCookie "zbind"
+type SetBindCookie = SimpleSetCookie "zbind"
```
neongreen: Our headers are called `Z-Connection`, `Z-User` and so on; maybe for the cookie it also makes sense to call it `Z-Bind`? (And at least according to Google, cookies are case-sensitive.)

services/spar/src/Spar/App.hs
```diff
@@ -234,26 +236,36 @@ verdictHandler cky aresp verdict = do
                -- (this shouldn't happen too often, see 'storeVerdictFormat')
 
 data VerdictHandlerResult
-  = VerifyHandlerDenied [ST]
-  | VerifyHandlerGranted SetCookie UserId
+  = VerifyHandlerGranted SetCookie UserId
+  | VerifyHandlerDenied [ST]
```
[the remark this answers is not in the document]
fisx: What bit of information would you like to add that's not already appearent in the name?

services/spar/src/Spar/App.hs
```diff
@@ -273,16 +285,17 @@ verdictHandlerResult bindCky = \case
 -- - The page broadcasts a message to '*', to be picked up by the app.
 verdictHandlerWeb :: HasCallStack => VerdictHandlerResult -> Spar SAML.ResponseVerdict
 verdictHandlerWeb = pure . \case
-    VerifyHandlerDenied reasons -> forbiddenPage reasons
     VerifyHandlerGranted cky _uid -> successPage cky
+    VerifyHandlerDenied reasons   -> forbiddenPage "forbidden" reasons
```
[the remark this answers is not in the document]
fisx: The rest of the info is in the reasons.  What else would you like to have here?

fisx:
> Hm, so where are the logic changes here? It says that it obsoletes #502, which was supposed to fix some corner cases. I see the new tests (we treat the presence of the cookie and the absence of it differently), but I haven't found what exactly happens differently in the code.

So, the main reason why binding was breaking creating new users on production on some devices was a combination of misunderstandings about how the cookies package works.  I think most of that has been fixed in saml2-web-sso, but some of the changes in spar were also part of that repair work.

I don't remember anything else that I fixed.

Note to self: separate refactoring commits from behavior-changing commits more clearly in the future.

All your comments I have not responded to make sense to me, I will process them tomorrow and ping you when I'm done.  Thanks!

services/spar/src/Spar/Types.hs
fisx: this is not a header name, but a cookie name: https://github.com/wireapp/wire-server/blob/0285578539abae98c122c63558b43bbdd8227e45/services/brig/src/Brig/User/Auth/Cookie.hs#L191

cookie names are keys in the key-value pair list that comes in the http header `Cookie`.

services/spar/src/Spar/Orphans.hs
fisx: turns out they weren't needed anywhere...?  removed, thanks!

fisx:
I don't know what's wrong with the CI, but I ran integration tests locally successfully.

PR #511, opened 2018-11-13
Fix: gundeck bulkpush option.

fisx:
Also:

- add a second cannon instance to integration tests;
- improve tests considerably around the bug;

There is also branch `gundeck-bulkpush-fix`, which is a less organised version of this branch.  both should be removed once this has landed at the latest.

### DIFF
--- a/services/gundeck/src/Gundeck/Push/Websocket.hs
+++ b/services/gundeck/src/Gundeck/Push/Websocket.hs
@@ -41,7 +41,6 @@ import qualified Gundeck.Presence.Data        as Presence
 import qualified Network.HTTP.Client.Internal as Http
 import qualified Network.URI                  as URI
 import qualified System.Logger.Class          as Log
-import qualified System.Logger                as Logger
 
 -- | Send a 'Notification's to associated 'Presence's.  Send at most one request to each Cannon.
 -- Return the lists of 'Presence's successfully reached for each resp. 'Notification'.
@@ -84,7 +83,7 @@ logCannonsGone (_uri, (_err, prcs)) = do
         view monitor >>= Metrics.counterAdd (fromIntegral $ length prcs)
             (Metrics.path "push.ws.unreachable")
         forM_ prcs $ \prc ->
-            Log.info $ logPresence prc
+            Log.warn $ logPresence prc
                 ~~ Log.field "created_at" (ms $ createdAt prc)
                 ~~ Log.msg (val "WebSocket presence unreachable: " +++ toByteString (resource prc))
 
@@ -119,8 +118,7 @@ bulkSend uri req = (uri,) <$> ((Right <$> bulkSend' uri req) `catch` (pure . Lef
 
 bulkSend' :: URI -> BulkPushRequest -> Gundeck BulkPushResponse
 bulkSend' uri (encode -> jsbody) = do
-    logr <- view applog
-    req <- ( check logr
+    req <- ( check
            . method POST
            . contentJson
            . lbytes jsbody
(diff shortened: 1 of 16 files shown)


fisx:

services/gundeck/src/Gundeck/Push/Websocket.hs
```diff
@@ -130,10 +132,13 @@ bulkSend' uri (encode -> jsbody) = do
   where
     submit req = recovering (limitRetries 1) rpcHandlers $ const (rpc' "cannon" req id)
 
-    check req = req { Http.checkResponse = \rq rs ->
-        when (responseStatus rs /= status200) $
+    check loggr req = req { Http.checkResponse = \rq rs ->
```
tiago-loureiro: Let's remove these log lines and handle the `BulkPushResponse` more gracefully; i.e., when we call [bulkPush](https://github.com/wireapp/wire-server/blob/3d0645c8df417d9e739c7fc3455750dda205474a/services/gundeck/src/Gundeck/Push/Websocket.hs#L50-L65) let's ensure that the different response codes are handled independently

services/integration.sh
```diff
@@ -74,7 +74,8 @@ function run() {
     service=$1
     colour=$2
     export LOG_LEVEL=$3
-    (cd ${DIR}/${service} && ${TOP_LEVEL}/dist/${service} -c ${service}.integration${integration_file_extension} || kill_all) \
+    instance=$4
```
tiago-loureiro: I am no bash expert but could we do something else other than waiting for an optional argument to append? Adding a 5 argument would probably confuse us quite a bit :/

fisx: like this?: 992d69093f5a876c747e1d3cd8675a4461173ff4

services/gundeck/src/Gundeck/Push/Websocket.hs
fisx: I think this is not the case where some notifications in the `BulkPushRequest` have failed to be delivered, but the case where something went wrong with the entire request, e.g. "cannon down".

If we get a `BulkPushResponse`, it contains the status codes for the individual notifications, and those will be processed outside of `bulkSend'`.

If that's not the answer, please clarify!  :-)

fisx:
CI doesn't work, local integration tests passed.

services/gundeck/src/Gundeck/Push/Websocket.hs
tiago-loureiro: > If we get a BulkPushResponse, it contains the status codes for the individual notifications, and those will be processed outside of bulkSend'.

True. Still, I was thinking that we should probably handle a 400 (for instance) differently from other response codes?

services/integration.sh
tiago-loureiro: > like this?: 992d690

Yes, I think that's "better" even though maybe a bit more ugly/noisy

services/gundeck/src/Gundeck/Push/Websocket.hs
fisx: I'm not sure what the difference between 5xx, 400, or 339 should be.  We expect 200, and everything else should be handled gracefull, but also treated as a problem.  In fact, perhaps we should make the warn-level logging an err-level logging?

Otherwise I'm happy to distinguish the cases, but I'll need more requirements then.  Or a piece of code where we already get it right.

fisx: after some off-line communication: 2d291b6

PR #512, opened 2018-11-14
Fix: spar session cookie

fisx:


services/spar/src/Spar/Intra/Brig.hs
```diff
@@ -194,14 +194,14 @@ assertIsTeamOwner buid tid = do
...
+-- | Get persistent cookie from brig and redirect user past login process.
 ssoLogin :: (HasCallStack, MonadError SparError m, SAML.HasConfig m, MonadSparToBrig m)
          => UserId -> m SetCookie
 ssoLogin buid = do
   resp :: Response (Maybe LBS) <- call
     $ method POST
     . path "/i/sso-login"
     . json (SsoLogin buid Nothing)
-    . queryItem "persistent" "true"
+    . queryItem "persist" "true"
```
jschaul: Could you add an integration test exercising this function and checking that the resulting cookie is indeed a persistent one, and not a Session cookie (see [CookieType](https://github.com/wireapp/wire-server/blob/df1eb6ab1533c16b1b3b824d3bb3b3027382f115/libs/brig-types/src/Brig/Types/User/Auth.hs#L191-L202))? Such a test would be useful to catch potential future regressions here.

fisx: will do.  interestingly, it's proving to be tricky to get those tests to pass, so i guess thanks for the suggestion.

fisx: so, tests are in place now, and they show it's not working yet.  [brig says:](https://github.com/wireapp/wire-server/blob/d67ff751c0ebbf18e7ea826f03da32910454415d/services/brig/src/Brig/User/API/Auth.hs#L189-L195)

```
(Request {requestMethod = "POST", httpVersion = HTTP/1.1, rawPathInfo = "/i/sso-login", rawQueryString = "?persist=true", requestHeaders = [("Host","127.0.0.1:8082"),("Accept-Encoding","gzip"),("Content-Length","60"),("Content-Type","application/json")], isSecure = False, remoteHost = 127.0.0.1:38666, pathInfo = ["i","sso-login"], queryString = [("persist",Just "true")], requestBody = <IO ByteString>, vault = <Vault>, requestBodyLength = KnownLength 60, requestHeaderHost = Just "127.0.0.1:8082", requestHeaderRange = Nothing},True)
[("Set-Cookie","zuid=b2skW9ekmxec0iwsjODIeUhPT6jGRlPJxDNRc8Q3bSRx0DM6nNw5PdXlnxEna_LfswSKv-b_K7nWNWeML-_zAA==.v=1.k=1.d=1542276007.t=u.l=.u=5297de7e-4b92-4028-9959-db374fc20985.r=427442d4; Path=/access; Expires=Thu, 15-Nov-2018 10:00:07 GMT; Domain=127.0.0.1; HttpOnly"),("Content-Type","application/json")]
```

persist flag is passed and accepted, but the resulting token still has a life expectancy of, what? 
 five minutes?

services/spar/test-integration/Util.hs
```diff
@@ -452,37 +453,59 @@ createTestIdPFrom metadata mgr brig galley spar = do
...
+    then Right $ SimpleSetCookie web
+    else Left $ "bad cookie name.  (found, expected) == " <> show (Web.setCookieName web, SAML.cookieName proxy)
+
+-- | In 'setResponseCookie' we set no expiration date iff cookie is persistent.  So here we test for
+-- expiration date.  Easier than parsing and inspecting the cookie value.
+hasPersistentCookieHeader :: ResponseLBS -> Either String ()
+hasPersistentCookieHeader rsp = do
+  cky <- getCookie (Proxy @"zuid") rsp
+  unless (isNothing . Web.setCookieExpires $ fromSimpleSetCookie cky) $
+    Left $ "expiration date shoult be empty: " <> show cky
```
jschaul: This test tests exactly the opposite of what it should be testing here:

When you log in **without** the `persist=true` query parameter, or with persist=false, you get a `session cookie`, which means it has no expiration date set, and will expire when you close the browser (and on the backend has a validity of max 1 week):
```
POST /login?persist=false HTTP/1.1

Set-Cookie: zuid=(redacted); Path=/access; Domain=zinfra.io; HttpOnly; Secure
```

When you log in **with** `persist=true`, you get a persistent cookie, which means it has *some* expiration date (in prod that's 56 days, in integration test that's a few seconds or minutes) and can be renewed with `POST /access`:
```
POST /login?persist=true HTTP/1.1

Set-Cookie: zuid=(redacted); Path=/access; Expires=Thu, 10-Jan-2019 10:43:28 GMT; Domain=zinfra.io; HttpOnly; Secure
```

Can you adapt your test code here and the comments to reflect this?

PR #517, opened 2018-11-21
Switch Galley to YAML-only config

neongreen:


services/galley/src/Galley/Options.hs
```diff
@@ -5,126 +5,58 @@
...
-                long "intra-device-listing"
-                <> help "Use this option if you want to fetch the device list from brig instead.")
-        <*>
-            (httpsUrlOption $
-                long "conversation-code-prefix"
-                <> metavar "STRING"
-                <> help "URI prefix for conversations with access mode 'code'")
-
+-- Used by the journaler
 journalOptsParser :: Parser JournalOpts
```
jschaul: I suppose this function could move to https://github.com/wireapp/wire-server/blob/develop/services/galley/journaler/src/Options.hs instead?

neongreen: Good idea.

PR #525, opened 2018-11-28
Support global `make i`.

fisx:


Makefile
```diff
@@ -29,25 +29,18 @@ clean:
...
 	$(MAKE) -C services/galley i
 	$(MAKE) -C services/brig i
 	$(MAKE) -C services/gundeck i-fake-aws
 	$(MAKE) -C services/spar i
 
+integration: fast i
+
 .PHONY: integration-aws
 integration-aws: fast
-	# We run "i" instead of "integration" to avoid useless rebuilds
```
neongreen: Did you also want to add running I-aws here?

fisx: > Did you also want to add running I-aws here?

good idea.  especially since the first commit is not just half-baked, but plain wrong.  thanks!

Makefile
```diff
@@ -29,26 +29,26 @@ clean:
...
+i:
 	$(MAKE) -C services/cargohold i
 	$(MAKE) -C services/galley i
 	$(MAKE) -C services/brig i
 	$(MAKE) -C services/gundeck i-fake-aws
 	$(MAKE) -C services/spar i
 
+integration: fast i
+
 .PHONY: integration-aws
```
neongreen:
```suggestion
.PHONY: i-aws
```

fisx: thanks!  (oh man, that shouldn't have been so hard...)

PR #538, opened 2018-12-04
Avoid modifying r->args.data, which breaks the $arg_name in nginx

coercible:
This pull request fixes a bug in nginx-zauth-module. We pass the `&r->args.data` to `ngx_unescape_uri`, the pointer will be modified in `ngx_unescape_uri`, see https://github.com/nginx/nginx/blob/master/src/core/ngx_string.c#L1735. Then the argument variables in query string would be lost in nginx script.

You can confirm the trouble by this way:

1. in nginx.conf, create such a `location`:

```
        location /bug {
            return 200 "bug is $arg_bug";
        }
```

2. visit this endpoint via curl: `curl -i -XGET http://localhost/bug?bug=zauth-bug`. You should get `bug is zauth-bug` however currently the value or this variable is null.

### DIFF
--- a/services/nginz/third_party/nginx-zauth-module/zauth_module.c
+++ b/services/nginz/third_party/nginx-zauth-module/zauth_module.c
@@ -309,7 +309,8 @@ static ngx_int_t zauth_parse_request (ngx_http_request_t * r) {
                         return NGX_ERROR;
                 }
                 u_char* writer = query.data;
-                ngx_unescape_uri(&writer, &r->args.data, r->args.len, 0);
+                u_char* reqargs = r->args.data;
+                ngx_unescape_uri(&writer, &reqargs, r->args.len, 0);
                 query.len = writer - query.data;
                 res = token_from_query(&query, &tkn);
         } else {


raphaelrobert:
This doesn't seem to pose a problem on our end and doesn't break anything. Could you kindly link to where this breaks in your code?

coercible:
We have integrated nginz with `lua-nginx-module`. We need to do some argument-dependent dispatch in nginx script, just something like:

```nginx
location /a {
     content_by_lua {
         local args, err = ngx.req.get_uri_args()
         if args['key1'] and args['key2'] then
             local res = ngx.location.capture('/b', ...)
             ngx.say(res.body)
         else
             local res = ngx.location.capture('/c', ...)
             ngx.say(res.body)
         end
     }
}

location /b {
    proxy_pass http://upstream1;
}

location /c {
    proxy_pass http://upstream2;
}
```

The bug fixed by this pull request leads to inaccessibility to the query arguments.

raphaelrobert:
This is a friendly reminder that Wire's server code is published under AGPL. This means, among other things, that derived work from our code also needs to be published under the same license and the source code needs to be made available.

coercible:
Thanks for merging this PR.

@raphaelrobert Thanks for your kiindful reminder, we will open source our code after certain development and testing.

PR #543, opened 2018-12-10
Get rid of a Monoid RequestId instance

neongreen:
Change base and merge once the LTS-12 PR is merged.


neongreen:
I would also be in favor (+0.5), yeah.

PR #531, opened 2018-11-30
streamline gundeck (--skip-fallbacks)

fisx:

PR #546, opened 2018-12-10
streamline gundeck (--prefer-notice)

fisx:
Similar to #531, this PR should not change the behavior *under the assumption* that `prefer-notice: true`.  (This obsoletes https://github.com/wireapp/wire-server/pull/518)

Exceptions:
- cli argument `--prefer-notice` is not allowed any more.
- ... ?

### DIFF
--- a/libs/gundeck-types/src/Gundeck/Types/Push/V2.hs
+++ b/libs/gundeck-types/src/Gundeck/Types/Push/V2.hs
@@ -225,6 +225,9 @@ data Push = Push
       -- of the originating user, if he is among the recipients.
     , _pushNativeEncrypt :: !Bool
       -- ^ Should native push payloads be encrypted?
+      --
+      -- REFACTOR: this make no sense any more since native push notifications have no more payload.
+      -- https://github.com/wireapp/wire-server/pull/546
     , _pushNativeAps :: !(Maybe ApsData)
       -- ^ APNs-specific metadata.
     , _pushNativePriority :: !Priority


fisx:
the test fails because this does not hold any more: https://github.com/wireapp/wire-server/blob/develop/services/gundeck/test/integration/API.hs#L719

i argue that that's fine, but i don't have enough context.  if we want to sustain the current behaviour, we need to keep feeding the database table, which makes no sense IMO because we never use the keys.

but it's true that this change in behavior can be witnessed by the clients.  we could ask them if they mind?

services/gundeck/src/Gundeck/Push/Native.hs
```diff
@@ -57,12 +57,17 @@ push1 m a = do
...
-    -- * Transient notifications must be delivered "now or never".
-    -- * Others use the default level of the specific platforms, which is 4 weeks
-    --   for both APNS* and GCM
-    ttl = if msgTransient m
-          then Just (Aws.Seconds 0)
-          else Nothing
+    -- TODO: REFACTOR: this smells like a bug that we've dragged along for a while: according to
+    -- current logic (previous to this PR), 'Notice' is never transient.  but we have sent nothing
+    -- but 'Notice' native notifications for a while.  have we sent out native notifications for
+    -- transient messages at all?  that may have led to clients being woken up, pulling the event
```
tiago-loureiro: >  have we sent out native notifications for transient messages at all? 

This used to happen when we used cipher text messages; in such cases, clients could simply decrypt the symmetrically encrypted message and be done with it (didn't have to ping the backend).

>but transient 'Notification's are translated to 'Notice'

The idea is that transient messages must be delivered _now or never_, thus the '0' TTL (e.g., the device is off). Seems to me now that we're changing the current behaviour, are we not?

fisx:
test suite passed locally.

PR #554, opened 2018-12-12
mock gundeck

fisx:
Notes for the reviewer:

- please read commit-by-commit first, ignoring everything about the new modules `MockGundeck`, `Push` in the gundeck unit tests.  then the overall changes in those two modules.
- i am particular fond of the *test gundeck's old pushAny against pushAll* stuff.  this gives strong evidence that the bulk-push code does not change behavior in any undesired way.  :)
- i changed some types from opaque to exposing constructors.  this was mostly for convenience, i think.  let me know if you want to reduce the changes there (and why).

Related: #518, #531, #546, #549


fisx:
updated the description.  this is ready for review, but i will improve haddocks soon, feel free to wait!

fisx:
all tests passed locally.  ready for review!

neongreen:
I've attempted to review the actual logic here, but failed miserably – probably because I don't know well what the original Gundeck does either.

I will take another stab at it later and ask more questions.

libs/gundeck-types/src/Gundeck/Types/Push/V2.hs
```diff
@@ -177,6 +183,8 @@ instance FromJSON ApsData where
 -----------------------------------------------------------------------------
 -- Priority
 
+-- | REFACTOR: do we every ues LowPriority?  easy to test, just remove the constructor.  if it is
```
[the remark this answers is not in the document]
fisx: good poing.  i will refine the comment.

libs/types-common/src/Data/Id.hs
```diff
@@ -66,6 +66,8 @@ newtype Id a = Id
     { toUUID :: UUID
     } deriving (Eq, Ord, Generic, NFData)
 
+-- REFACTOR: non-derived, custom show instances break pretty-show and violate the law
```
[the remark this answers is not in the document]
fisx: Interesting pointer, but does this produce valid `Read` instances as well?  (I think it doesn't.)

services/gundeck/test/unit/MockGundeck.hs
```diff
@@ -0,0 +1,580 @@
...
+data MockEnv = MockEnv
+  { _meRecipients      :: [Recipient]
+  , _meNativeAddress   :: Map UserId (Map ClientId (Address "no-keys"))
+  , _meWSReachable     :: Set (UserId, ClientId)
+  , _meNativeReachable :: Set (Address "no-keys")
+  }
+  deriving (Eq, Show)
+
+data MockState = MockState
+  { _msNonRandomGen    :: StdGen
```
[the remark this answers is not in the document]
fisx: will add a comment:

```haskell
  { _msNonRandomGen    :: StdGen
    -- ^ the "non-random" in the name means that we always run 'MockGundeck' with the same 'StdGen'
    -- state (look at the 'Monoid' instance below).
```

services/gundeck/test/unit/MockGundeck.hs
```diff
@@ -0,0 +1,580 @@
...
+  protoaddrs :: [UserId -> ClientId -> Address "no-keys"]
+    <- do
+      len <- let l = length _meRecipients in choose (l `div` 2, l)
+      vectorOf len genProtoAddress
+
+  let addrs :: [(Recipient, Address "no-keys")]
+      addrs = mconcat $ zipWith go _meRecipients protoaddrs
+        where go rcp@(Recipient uid _ cids) adr = (\cid -> (rcp, adr uid cid)) <$> cids
+
+      _meNativeAddress :: Map UserId (Map ClientId (Address "no-keys"))
```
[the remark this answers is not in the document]
fisx: correct.  comment added.

services/gundeck/test/unit/MockGundeck.hs
```diff
@@ -0,0 +1,580 @@
...
+validateMockEnv env = do
+  forM_ (Map.toList $ env ^. meNativeAddress) $ \(uid, el) -> do
+    forM_ (Map.toList el) $ \(cid, adr) -> do
+      unless (uid == adr ^. addrUser && cid == adr ^. addrClient) $ do
+        throwError (show (uid, cid, adr))
+
+recipientToIds :: Recipient -> [(UserId, ClientId)]
+recipientToIds (Recipient uid _ cids) = (uid,) <$> cids
+
+genPredicate :: forall a. (Eq a, Ord a) => [a] -> Gen (Set a)
```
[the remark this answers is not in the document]
fisx: `genSubsetOf` it is!  (i want the `gen` to stay in the name because it's a 'Gen', and I do that everywhere else.)

services/gundeck/test/unit/MockGundeck.hs
```diff
@@ -0,0 +1,580 @@
...
+
+genRecipient :: HasCallStack => Gen Recipient
+genRecipient = do
+  uid  <- genId
+  cids <- nub <$> listOf1 genClientId
+  pure $ Recipient uid RouteAny cids
+
+genId :: Gen (Id a)
+genId = do
+  gen <- mkStdGen <$> resize 100 arbitrary
```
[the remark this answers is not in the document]
fisx: Not sure about the semantics of quickcheck's `resize`, but I think it's less directly related to the 'Gen' value range.  Come to think of it, I think `100` is the default and this is bogus.  Would it be a problem if there were duplicates here?  I'm not sure, but I actually think that would be a good thing for the tests to run into.

fisx:
thanks for your feedback!  i processed everything but commented only where it wasn't obvious what i'm supposed to do.  new commit coming up!

libs/types-common/src/Data/Id.hs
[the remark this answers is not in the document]
neongreen: No, your "violate the law" point still stands.

services/gundeck/test/unit/MockGundeck.hs
```diff
@@ -0,0 +1,716 @@
...
+shrinkPretty shrnk (Pretty xs) = Pretty <$> shrnk xs
+
+
+deliver :: NotifQueue -> (UserId, ClientId) -> Payload -> NotifQueue
+deliver queue qkey qval = Map.alter (Just . tweak) qkey queue
+  where
+    tweak Nothing      = Set.singleton (payloadToInt qval)
+    tweak (Just qvals) = Set.insert    (payloadToInt qval) qvals
+
+payloadToInt :: Payload -> Int
```
neongreen: Hm, why `toRational (x * 100)` and not just taking the number in the payload?

Also, do we test invalid payloads somewhere? If not, it might be better to just redefine `newtype Payload = Payload Int`, which would have the benefit of immediately signalling that in mock Gundeck we use only a subset of possible payloads.

services/gundeck/test/unit/MockGundeck.hs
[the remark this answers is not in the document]
neongreen: For integrals, `resize 100 arbitrary` will return numbers in `[-100;100]`.

If we want tests to run into duplicates, it would be better to build that explicitly by doing something like

```haskell
frequency
  [ (1, pure <some ID>)
  , (9, <generate completely random ID>)
  ]
```

Or just use `choose (0, 100)` or something, this one has more immediately understandable semantics (I think).

services/gundeck/test/unit/MockGundeck.hs
```diff
@@ -0,0 +1,716 @@
...
+import qualified Data.Vector as Vector
+import qualified Network.URI as URI
+
+
+----------------------------------------------------------------------
+-- env
+
+type Payload = List1 Aeson.Object
+
+data MockEnv = MockEnv
```
neongreen: I wonder whether all the code that generates and consumes maps can be simplified by moving to a different representation:

```haskell
data ClientInfo = ClientInfo
    { _mrNativeAddress :: Maybe (Address "no-keys")
    , _mrWSReachable :: Bool
    , _mrNativeReachable :: Bool
    }

data MockEnv = MockEnv
    { _meRecipients :: Map (UserId, ClientId) ClientInfo
    }

clients :: UserId -> Lens' (Map (UserId, ClientId) a) (Map ClientId a)
clients = ...
```

This is less efficient but probably easier to work with and we won't have many clients in mock Gundeck so the inefficiency hopefully doesn't matter.

I haven't actually tried rewriting the code to use this representation, though.

neongreen: Oh, and also it makes `MockEnv` always valid (at least for now).

services/gundeck/test/unit/MockGundeck.hs
fisx: good idea, but:

```haskell
-- | We really don't care about the actual payloads anywhere in these tests, just that the right
-- ones arrive over the right connections.  So 'genPayload' is not very exhaustive, but only
-- generates small objects with one field containing a numeric value.  It would be nice to represent
-- this in the 'Payload' type, but the 'List1 Aeson.Object' structure is used in the production
-- code, so in the end it is more awkward than nice.
type Payload = List1 Aeson.Object

[...]

-- | Return the rounded number contained in the payload.  This is enough to detect message mixups,
-- but less noisy in the printed counter-examples.
payloadToInt :: Payload -> Int
```

ok?

services/gundeck/test/unit/MockGundeck.hs
fisx: - your data model means that a client can be native reachable if it doesn't have an address.  this could be fixed, though:

```haskell
data ClientInfo = ClientInfo
    { _ciNativeAddress   :: Maybe (Address "no-keys", Bool{- reachable -})
    , _ciWSReachable     :: Bool
    }
  deriving (Eq, Show)

data MockEnv = MockEnv
    { _meRecipients :: Map Recipient ClientInfo
    }
  deriving (Eq, Show)
```
- validity is still not guaranteed by it: Address can still contain a different `UserId` or `ClientId` from the one in the `meRecipients` map key.
- i'm not sure whether it would make things easier or more complicated.  `MockEnv` is read-only, so there is no problem with keeping the redundant information in sync.

but i'll give it a try.  i'm also curious how it'll go.

fisx:
I tried your change in data model, and I think it's slightly better.  Confusingly, both the size of the `MockEnv` values and the runtime of the native push tests have dropped by one order of magnitude.  I'm not sure whether this shows that the new data model is actually faster, or whether I have made a mistake, but superficially, the `MockEnv` values look fine.

Note that the test for the websockets function remains slow.  I'm not sure if that means that there is a performance issue in the bulkpush code there?

fisx:
btw i found that your suggested data structure inspires interesting errors in the code that combines disconnected user ids and client ids into recipients.  the nested map works better in some places, and is still ok in the others.

services/gundeck/test/unit/MockGundeck.hs
neongreen: > in the end it is more awkward than nice

Alright.

> rounded number

I still don't understand what's the rationale behind multiplying an always-integer number by 100, but I guess it doesn't hurt either.

fisx: > I still don't understand what's the rationale behind multiplying an always-integer number by 100, but I guess it doesn't hurt either.

i think i noticed numbers that only differed in their tenths and hundredths, and i wanted to keep more information after rounding.

services/gundeck/test/unit/MockGundeck.hs
```diff
@@ -0,0 +1,710 @@
...
+    }
+  deriving (Eq, Show)
+
+newtype MockEnv = MockEnv
+    { _meClientInfos :: Map UserId (Map ClientId ClientInfo)
+    }
+  deriving (Eq, Show)
+
+data MockState = MockState
+  { _msNonRandomGen    :: StdGen
```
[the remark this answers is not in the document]
fisx: tried it, it's *a lot* better.  commit coming up, thanks!

services/gundeck/test/unit/MockGundeck.hs
```diff
@@ -0,0 +1,710 @@
...
+makeLenses ''MockState
+
+instance Eq MockState where
+  (MockState _ w n c) == (MockState _ w' n' c') = w == w' && n == n' && c == c'
+
+instance Show MockState where
+  show (MockState _ w n c) = intercalate "\n"
+    ["", "websocket: " <> show w, "native: " <> show n, "cassandra: " <> show c, ""]
+
+instance Semigroup MockState where
```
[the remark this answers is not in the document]
fisx: When I wrote the Monoid instance, I got an error that it didn't match the SemiGroup instance (or that the latter was missing).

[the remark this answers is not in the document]
neongreen: If you're only using the instance to get a `mempty`, it might be better to move `emptyMockState` as a local declaration in `runMockGundeck`.

neongreen:
I've also pushed a (hopefully non-controversial) commit.

neongreen:
```haskell
data MockState = MockState
  {
  -- | A record of notifications that have been pushed via websockets.
    _msWSQueue         :: NotifQueue
  -- | A record of notifications that have been pushed via native push.
  , _msNativeQueue     :: NotifQueue
  -- | Non-transient notifications that are stored in the database for later
  -- delivery (e.g. because the client is offline).
  , _msCassQueue       :: NotifQueue
  }
  deriving (Eq)
```

Is my understanding (in the comments) correct?

services/gundeck/test/unit/MockGundeck.hs
```diff
@@ -0,0 +1,710 @@
...
+genRecipient :: HasCallStack => MockEnv -> Gen Recipient
+genRecipient env = do
+  uid   <- QC.elements (allUsers env)
+  route <- genRoute
+  cids  <- Set.toList <$> genSubsetOf (clientIdsOfUser env uid)
+  pure $ Recipient uid route cids
+
+-- REFACTOR: see 'Route' type about missing 'RouteNative'.
+genRoute :: HasCallStack => Gen Route
+genRoute = QC.elements [RouteAny, RouteDirect]
```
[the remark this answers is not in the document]
fisx: that both direct or native are supported.  (we never use native-only, so i don't generate that here.)

services/gundeck/test/unit/MockGundeck.hs
```diff
@@ -0,0 +1,710 @@
...
+
+genId :: Gen (Id a)
+genId = do
+  gen <- mkStdGen <$> arbitrary
+  pure . Id . fst $ random gen
+
+genClientId :: Gen ClientId
+genClientId = newClientId <$> arbitrary
+
+genProtoAddress :: Gen (UserId -> ClientId -> Address "no-keys")
```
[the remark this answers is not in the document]
fisx: right.  i was using this in a different way where that wouldn't have worked.  now it was easy to fix.

services/gundeck/test/unit/MockGundeck.hs
```diff
@@ -0,0 +1,712 @@
...
+          , cid <- cids
+          ]
+
+    reformat :: [(Recipient, Payload)] -> [((UserId, ClientId), Payload)]
+    reformat = mconcat . fmap go
+      where
+        go (Recipient uid _ cids, pay) = (\cid -> ((uid, cid), pay)) <$> cids
+
+    rcps :: [((UserId, Maybe ClientId, Bool), (Recipient, Payload))]
+    rcps = mconcat $ go <$> filter (not . (^. pushTransient)) pushes
```
neongreen: Why do we filter out transient pushes? Shouldn't they still be pushed via websockets?

services/gundeck/test/unit/MockGundeck.hs
```diff
@@ -0,0 +1,712 @@
...
+      . filter reachable
+      . reformat
+      . mconcat . fmap removeSelf
+      . mconcat . fmap insertAllClients
+      $ rcps
+      where
+        reachable :: ((UserId, ClientId), payload) -> Bool
+        reachable (ids, _) = wsReachable env ids
+
+        removeSelf :: ((UserId, Maybe ClientId, any), (Recipient, Payload)) -> [(Recipient, Payload)]
```
neongreen: It always returns exactly one value so the list and `mconcat` aren't necessary. Omission or by design?

fisx:
> ```haskell
> data MockState = MockState
>   {
>   -- | A record of notifications that have been pushed via websockets.
>     _msWSQueue         :: NotifQueue
>   -- | A record of notifications that have been pushed via native push.
>   , _msNativeQueue     :: NotifQueue
>   -- | Non-transient notifications that are stored in the database for later
>   -- delivery (e.g. because the client is offline).
>   , _msCassQueue       :: NotifQueue
>   }
>   deriving (Eq)
> ```
> 
> Is my understanding (in the comments) correct?

almost.  i didn't find them in the branch, so i'll commit them with a small correction: non-transient notifications are *always* stored in cassandra, even before we try to deliver them via websocket.

neongreen:
>  i didn't find them in the branch

Yeah, I didn't want to commit them because I wasn't sure about them, but I would've committed them after your confirmation.

services/gundeck/test/unit/MockGundeck.hs
fisx: i think that's a bug that wasn't caught because genPush didn't touch it.  you're thorough with this one!  :-)

services/gundeck/test/unit/MockGundeck.hs
fisx: probably used to make sense, but doesn't any more.  fixing!

services/gundeck/test/unit/MockGundeck.hs
```diff
@@ -227,16 +239,13 @@ validateMockEnv env = do
               throwError (show (uid, cid, adr))
 
 genSubsetOf :: forall a. (Eq a, Ord a) => [a] -> Gen (Set a)
-genSubsetOf xs = Set.fromList <$> do
-  bools :: [Bool] <- vectorOf (length xs) arbitrary
-  let reachables :: [a] = mconcat $ zipWith (\x yes -> [ x | yes ]) xs bools
-  pure reachables
+genSubsetOf = fmap Set.fromList . sublistOf
```
fisx: this actually justifies naming it `subsetOf`, i guess...

services/gundeck/test/unit/MockGundeck.hs
```diff
@@ -253,8 +262,8 @@ genClientId = newClientId <$> arbitrary
 
 genProtoAddress :: Gen (UserId -> ClientId -> Address "no-keys")
 genProtoAddress = do
-  _addrTransport <- QC.elements [minBound..]
-  arnEpId <- arbitrary
+  _addrTransport :: Transport <- QC.elements [minBound..maxBound]
```
fisx: why the `maxBound`?

neongreen: I thought it clarified the intent better (`[minBound..]` looks line an infinite list), but I grepped now and we have `[minBound..]` in other places as well so I'll revert this change.

services/gundeck/test/unit/MockGundeck.hs
```diff
@@ -0,0 +1,719 @@
...
+        insertAllClients :: (any, (Recipient, Payload))
+                         -> [(any, (Recipient, Payload))]
+        -- if the recipient client list is empty, fill in all devices of that user
+        insertAllClients (same, (Recipient uid route [], pay)) = [(same, (rcp', pay))]
+          where rcp' = Recipient uid route (clientIdsOfUser env uid)
+        -- otherwise, no special hidden meaning.
+        insertAllClients same@(_, (Recipient _ _ (_:_), _)) = [same]
+
+    expectCass :: NotifQueue
+    expectCass = foldl' (uncurry . deliver) mempty . mconcat $ go <$> pushes
```
neongreen: Is there a reason why `insertAllClients` is not used here? (If it's an omission, I'll fix it.)

fisx: i think it may be an oversight.  (i wonder if we need to improve coverage by the Arbitrary generators further to catch this?)

btw i think there is some more code duplication to get rid of, but you're probably on it.

services/gundeck/test/unit/MockGundeck.hs
fisx: ok, fair enough.  i have no strong opinion then, feel free to leave it in  :-)

(i like to leave it out because if it's an error, it will always trigger on the first run that touches this, and it's prettier because fewer bytes of source code.  more zen.)

libs/gundeck-types/src/Gundeck/Types/Push/V2.hs
```diff
@@ -177,6 +183,9 @@ instance FromJSON ApsData where
 -----------------------------------------------------------------------------
 -- Priority
 
+-- | REFACTOR: do we every ues LowPriority?  to test, (a) remove the constructor and see what goes
```
tiago-loureiro: Typo :)

services/gundeck/test/unit/Push.hs
```diff
@@ -0,0 +1,85 @@
...
+  where
+    mkNotifs :: Pretty MockEnv -> Property
+    mkNotifs (Pretty env) = forAllShrink
+      (Pretty <$> resize len (genNotifs env))
+      (shrinkPretty shrinkNotifs)
+      (webBulkPushProp env)
+
+webBulkPushProp :: MockEnv -> Pretty [(Notification, [Presence])] -> Property
+webBulkPushProp env (Pretty notifs) =
+    counterexample "^ environment, notifications\n" $
```
neongreen: QuickCheck was printing the environment and the list of notifications twice (the first time because it always prints function parameters, the second time because of `counterexample`). I couldn't convince it to stop doing that so I dropped the `counterexample`. Both were formatted the same, luckily, so the tests aren't hurt.

services/gundeck/test/unit/MockGundeck.hs
neongreen: I think I solved the mystery – `Arbitrary Aeson.Value` could be doing that. We don't have it anymore, but if fractional payloads ever creep into the tests again, `toBoundedInteger` will catch them so I feel pretty safe.

services/gundeck/test/unit/MockGundeck.hs
```diff
@@ -0,0 +1,676 @@
...
+  rcps :: Range 1 1024 (Set Recipient) <- do
+    numrcp <- choose (1, min 1024 (length alluids))
+    rcps   <- genRecipients numrcp env
+    unsafeRange . Set.fromList <$> dropSomeDevices `mapM` rcps
+  pload <- genPayload
+  inclorigin <- arbitrary
+  transient <- arbitrary
+  pure $ newPush sender rcps pload
+    -- TODO: & pushConnections .~ _
+    -- TODO: & pushOriginConnection .~ _
```
neongreen: I think at least this one should be done before it's merged, no idea about `pushConnections`.

services/gundeck/src/Gundeck/Push.hs
```diff
@@ -69,60 +75,88 @@ push (req ::: _) = do
...
     let tgts  = mkTarget <$> uniq
     unless (p^.pushTransient) $
-        Stream.add i tgts pload =<< view (options.optSettings.setNotificationTTL)
-    void . forkIO $ do
-        prs <- Web.push notif tgts (p^.pushOrigin) (p^.pushOriginConnection) (p^.pushConnections)
-        pushNative notif p =<< nativeTargets p prs
+        mpaStreamAdd i tgts pload =<< mpaNotificationTTL
+    mpaForkIO $ do
+        prs <- mpyPush notif tgts (p^.pushOrigin) (p^.pushOriginConnection) (p^.pushConnections)
+        unless (p^.pushTransient) $
```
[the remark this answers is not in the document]
fisx: thanks!

fisx:
there is still an error uncovered by the last test coverage expansion:

```
stack test . --pedantic --test --bench --no-run-benchmarks --local-bin-path=dist --fast --test-arguments=--quickcheck-replay=186970 --ghc-options=-Wwarn
```

`pushAny` correctly does not send anything; `pushAll` pushes to a device that shouldn't receive.  if i implement `connWhitelistSieve` as `id` in `mockOldSimpleWebPush`, it's the opposite.

hypothesis: something about "empty list means all devices" again.  maybe `pushAny` calls `mpyPush` with input where empty means full, but `pushAll`...  (where does `pushAll` call `mpyPush`?  how can the behavior of `mockOldSimpleWebPush` change the behavior of `pushAll`?!

libs/gundeck-types/src/Gundeck/Types/Push/V2.hs
```diff
@@ -177,6 +183,9 @@ instance FromJSON ApsData where
 -----------------------------------------------------------------------------
 -- Priority
 
+-- | REFACTOR: do we ever use LowPriority?  to test, (a) remove the constructor and see what goes
+-- wrong; (b) log use of 'LowPriority' by clients in production and watch it a few days.  if it is
```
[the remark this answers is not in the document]
neongreen: Matthias added a comment (thanks!), but ideally I'd like to move faster on this and investigate the potential removal now instead of not-now.

Given that `native_priority` seems to be unused in client code, I think it's a sane approach to deprecate it in a separate PR (just reply with 400) and see whether staging breaks. I'll ask QA about this.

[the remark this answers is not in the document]
fisx: i am planning to go through the `REFACTOR` comments soon, but for now i think this pull request is really ripe and shouldn't contain any more of it.  (not sure you are suggesting that, just saying...)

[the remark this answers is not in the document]
neongreen: Yeah, I wasn't suggesting doing it in this PR.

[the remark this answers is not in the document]
neongreen: I asked @red-avtovo and he pointed out that since notifications aren't very reliable at the moment, QA might not notice that they are even more broken. So, let's wait with removing LowPriority.

fisx:
```
TASTY_PATTERN='$NF ~ /native pushes/'  stack test . --pedantic --test --bench --no-run-benchmarks --local-bin-path=dist --fast --test-arguments=--quickcheck-replay=785105 --ghc-options=-Wwarn
```

device `e` is native reachable and ws reachable.  it gets filtered out by `pushConnections`, is not sent to the websocket, and therefore gets pushed natively.  that sounds quite insane, but both production implementations (with `--bulk-push` and without) behave that way.

is this a bug in production?

libs/gundeck-types/src/Gundeck/Types/Push/V2.hs
```diff
@@ -197,18 +206,21 @@ data Push = Push
...
       --
-      -- TODO: should be @Set (Recipient, Maybe (NonEmptySet ConnId))@, and '_pushConnections'
-      -- should go away.  Rationale: the current setup only works under the assumption that no
-      -- 'ConnId' is used by two 'Recipient's.  This is *probably* correct, but not in any contract.
-      -- Coincidentally, where are we using '_pushConnections' to limit pushes to individual
-      -- devices?  Is it possible we can remove '_pushConnections' without touching
+      -- REFACTOR: '_pushRecipients' should be @Set (Recipient, Maybe (NonEmptySet ConnId))@, and
+      -- '_pushConnections' should go away.  Rationale: the current setup only works under the
+      -- assumption that no 'ConnId' is used by two 'Recipient's.  This is *probably* correct, but
+      -- not in any contract.  Coincidentally, where are we using '_pushConnections' to limit pushes
```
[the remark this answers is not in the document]
tiago-loureiro: > I would also like to see _pushConnections go but I recall being told long long ago that it was actually needed for something.

Right, we cannot remove this without touching `pushRecipients`: let's discuss it in a subsequent PR.

libs/gundeck-types/src/Gundeck/Types/Push/V2.hs
```diff
@@ -222,7 +256,7 @@ data Push = Push
       -- REFACTOR: this make no sense any more since native push notifications have no more payload.
       -- https://github.com/wireapp/wire-server/pull/546
     , _pushNativeAps :: !(Maybe ApsData)
-      -- ^ APNs-specific metadata.
+      -- ^ APNs-specific metadata.  REFACTOR: can this be removed?
```
tiago-loureiro: Yes it can!

libs/types-common/src/Data/Id.hs
```diff
@@ -183,6 +186,12 @@ instance FromJSON ClientId where
             fail "Invalid ClientId"
         return (ClientId x)
 
+instance FromJSONKey ClientId where
```
tiago-loureiro: How about a little refactor and adding `clientIdFromText :: Text -> Parser (Id a)` to avoid duplication of code in `FromJSONKey` and `FromByteString` ?

services/gundeck/src/Gundeck/Push/Native/Types.hs
```diff
@@ -34,6 +34,9 @@ import Gundeck.Aws.Arn
 import Gundeck.Types
 
 -- | Native push address information of a device.
+--
+-- REFACTOR: the @s@ phantom type can probably go away, too!
```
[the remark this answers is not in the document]
tiago-loureiro: `Address "keys"` was removed here: https://github.com/wireapp/wire-server/pull/546

This datatype was used to distinguish between addresses that _had_ encryption keys (and thus could be symmetrically encrypted) and those that could not.

Since we have now removed this, this distinction no longer makes sense and thus can (and should) be removed and described in the comment _REFACTOR: the @s@ phantom type can probably go away, too!_

services/gundeck/src/Gundeck/Push.hs
```diff
@@ -69,101 +75,132 @@ push (req ::: _) = do
...
-        resp <- mapM (compilePushResp notifIdMap) =<< mpaBulkPush (compilePushReq <$> targets)
+        resp <- compilePushResps targets <$> mpaBulkPush (compilePushReq <$> targets)
 
         -- native push
-
-        forM_ resp $ \((notif, psh), alreadySent) -> do
-            natives <- mpaNativeTargets psh alreadySent
-            mpaPushNative notif psh natives
+        forM_ resp $ \((notif, psh), alreadySent) -> unless (psh ^. pushTransient) $
+            mpaPushNative notif psh =<< nativeTargets psh alreadySent
```
tiago-loureiro: Shouldn't this call `mpaNativeTargets` instead of `nativeTargets` ?

fisx:
@neongreen wanna have a "last" look?  check that https://github.com/wireapp/wire-server/pull/554/commits/469c78a549c04819d8c4709cac56ba8e41efbc72 makes sense and that i didn't break `compilePushResps`, and see if you want to approve?

thanks for your work on this, that was quite necessary!

services/gundeck/src/Gundeck/Push.hs
```diff
@@ -290,6 +292,9 @@ nativeTargets p pres =
     eligibleClient _ RecipientClientsAll = True
     eligibleClient a (RecipientClientsSome cs) = (a^.addrClient) `elem` cs
 
+    whitelistedOrNoWhitelist a = p^.pushConnections == mempty
```
neongreen: `null` should work too, since both `elem` and `null` come from `Foldable`.

services/gundeck/test/unit/MockGundeck.hs
```diff
@@ -494,7 +493,9 @@ handlePushNative Push{..} = do
           isOriginDevice = origin == (uid, Just cid)
           isAllowedPerOriginRules =
             not isOriginUser || (_pushNativeIncludeOrigin && not isOriginDevice)
-      when (isNative && isReachable && isAllowedPerOriginRules) $
+      -- Condition 5: push to cid iff (a) listed in pushConnections or (b) pushConnections is empty.
+      let isWhitelisted = Set.null _pushConnections || fakeConnId cid `elem` _pushConnections
```
neongreen: Seems better to have either `null+elem` or `Set.null+Set.member`, but not a mix of those.

services/gundeck/src/Gundeck/Push.hs
```diff
@@ -69,101 +75,132 @@ push (req ::: _) = do
...
-            mkNotificationTarget :: Recipient -> NotificationTarget
-            mkNotificationTarget r = target (r ^. recipientId)
-                                   & targetClients .~ r ^. recipientClients
+            mkTarget :: Recipient -> NotificationTarget
+            mkTarget r =
+              target (r^.recipientId)
+                & targetClients .~ case r^.recipientClients of
+                    RecipientClientsAll -> []
+                      -- clients are stored in cassandra as a list with a notification.  empty list
+                      -- is intepreted as "all clients" by 'Gundeck.Notification.Data.toNotif'.
```
neongreen:
```suggestion
                      -- is interpreted as "all clients" by 'Gundeck.Notification.Data.toNotif'.
```

PR #561, opened 2019-01-07
App-level websocket pongs.

### LINKED ISSUE #560: Make server respond to websocket ping messages
Webapp sends `'Wire is so much nicer with internet!'` [every now and then](https://github.com/wireapp/wire-webapp/blob/b6117ed93297feaab2ea845095d75a678161cfc4/app/script/event/WebSocketService.js#L224) to make sure it is still able to talk to the server. However because server doesn't respond to this message, in some cases it is impossible to detect if webapp is actually offline.

See this message for details: https://github.com/wireapp/wire-desktop/issues/1129#issuecomment-437965302

Please make Wire respond with some string like `'Indeed it is'` on each such ping request, so that it's possible to improve webapp.

fisx:
Fixes #560.

@marcoconti83 this needs to be ok'ed by all clients, since it generates web socket messages that need to be handled by the clients.  More complicated solutions are conceivable, like a way for the client to turn this on or off (but I'm not in favour of that).

I also wonder if there is really no other way for client to tell whether the browser has lost the websocket connection, but I am ready to believe that's the case.

tasks:

- [x] send app-level pongs
- [x] integration test


services/cannon/src/Cannon/App.hs
```diff
@@ -115,12 +115,22 @@ readLoop ws s = loop
                 send (connection ws) (pong p)
                 loop
             ControlMessage (Close _ _) -> return ()
-            _                          -> reset counter s 0 >> loop
+            _ -> do
+                reset counter s 0
+                sendAppLevelPong
```
neongreen: I feel slightly uneasy about sending a pong for every data message – perhaps we should only pong to messages that say "ping"? Though I can't see any _actual_ problems with sending a pong for every data message.

I also feel uneasy about sending and receiving payloads that are not Protobuf/JSON-encoded (or generally encoded in the same way as other payloads), but dragging Protobuf into this code looks like a mistake.

fisx: > I feel slightly uneasy about sending a pong for every data message – perhaps we should only pong to messages that say "ping"? Though I can't see any _actual_ problems with sending a pong for every data message.

the reason i chose to ignore the ping message is that it is currently a rather ad hoc string, and i think it's brittle and dangerous to depend on that.  note that we don't expect any other messages but `ping`; everything else from the client is sent via http.  (right?)

> I also feel uneasy about sending and receiving payloads that are not Protobuf/JSON-encoded (or generally encoded in the same way as other payloads), but dragging Protobuf into this code looks like a mistake.

i'm indifferent.  let's look into protobuffing it once it's clear what we want to do.  I still hope there is a way to avoid this PR entirely.

neongreen: > I still hope there is a way to avoid this PR entirely.

Agreed.

fisx:
> [17c189a](https://github.com/wireapp/wire-server/commit/17c189a9e5c48c31eaa62bc529b6c709c460f240) fails, [a8fd6ba](https://github.com/wireapp/wire-server/commit/a8fd6ba719b6b634d1f552e6ce24d9bd7cb93062) passes. needs more thinking...

could not reproduce.  perhaps it's related to https://github.com/wireapp/wire-server/pull/573#issuecomment-454954837 ?

PR #570, opened 2019-01-14
Feature/docker image consistency

### LINKED ISSUE #563: 'make docker-services' target docker-services is missing in Makefile
Instructions in the README on how to build the Wire server with Docker [specify](https://github.com/wireapp/wire-server#2-use-docker) to run `make docker-services` but that target is missing from the [Makefile](https://github.com/wireapp/wire-server/blob/develop/Makefile) there is a generic target for any [`docker-service-%`](https://github.com/wireapp/wire-server/blob/develop/Makefile#L117) but not a catch-all `docker-services` target.

```
:~/wire-server (develop)
$ make docker-services
make: *** No rule to make target `docker-services'.  Stop.
```

jschaul:
With this PR, only one way of building docker images is supported: 1) build `alpine-intermediate` (i.e. compile all the haskell code) 2) split the intermediate image with all the compiled binaries into smaller ones (e.g. `brig`), with one binary each. This is what CI does as well.

Re-compiling any docker image with updated code requires first recreating `alpine-intermediate`. 

Fixes #563


Makefile
```diff
@@ -97,25 +97,36 @@ docker-builder:
...
-	docker build -t $(DOCKER_USER)/migrations:$(DOCKER_TAG) -f build/alpine/Dockerfile.migrations .
+	docker build -t $(DOCKER_USER)/migrations:$(DOCKER_TAG) -f build/alpine/Dockerfile.migrations --build-arg intermediate=$(DOCKER_USER)/alpine-intermediate --build-arg deps=$(DOCKER_USER)/alpine-deps .
 	docker tag $(DOCKER_USER)/migrations:$(DOCKER_TAG) $(DOCKER_USER)/migrations:latest
 	if test -n "$$DOCKER_PUSH"; then docker login -u $(DOCKER_USERNAME) -p $(DOCKER_PASSWORD); docker push $(DOCKER_USER)/migrations:$(DOCKER_TAG); docker push $(DOCKER_USER)/migrations:latest; fi;
 
 .PHONY: docker-exe-%
 docker-exe-%:
-	docker build -t $(DOCKER_USER)/"$*":$(DOCKER_TAG) -f build/alpine/Dockerfile.executable --build-arg executable="$*" .
+	docker image ls | grep $(DOCKER_USER)/alpine-deps > /dev/null || (echo "'make docker-deps' required.", exit 1)
+	docker image ls | grep $(DOCKER_USER)/alpine-intermediate > /dev/null || (echo "'make docker-intermediate' required."; exit 1)
```
fisx: I assume `docker build` has no way of knowing whether it needs to run or not, so it will always run, so we can't replace these two lines with make deps in the goal.

build/alpine/Dockerfile
```diff
@@ -1,32 +0,0 @@
-# Requires docker >= 17.05 (requires support for multi-stage builds)
```
fisx: what was the purpose of this file before?  i can't find a reason why it was there in the code you changed / removed.

Makefile
```diff
@@ -97,25 +97,36 @@ docker-builder:
...
 	if test -n "$$DOCKER_PUSH"; then docker login -u $(DOCKER_USERNAME) -p $(DOCKER_PASSWORD); docker push $(DOCKER_USER)/"$*":$(DOCKER_TAG); docker push $(DOCKER_USER)/"$*":latest; fi;
 
-.PHONY: docker-service-%
-docker-service-%:
-	$(MAKE) -C "services/$*" docker
+.PHONY: docker-services
+docker-services:
+	docker image ls | grep $(DOCKER_USER)/alpine-deps > /dev/null || (echo "'make docker-deps' required.", exit 1)
+	docker image ls | grep $(DOCKER_USER)/alpine-intermediate > /dev/null || (echo "'make docker-intermediate' required."; exit 1)
+	$(MAKE) -C services/brig docker
```
fisx: isn't there some duplication between `make docker-exe-brig` and `make -C services/brig docker`?  or am i missing a difference?  if so you can either remove the redundancy or leave a comment.

Makefile
jschaul: Well, the question is about the default - should the default be "always recompile everything - including a stack clean", or not. I opted for not recompiling by default. Should there be 2 different targets (one that recompiles and one that doesn't) ? If so, naming suggestions?

fisx: nah, doing that manually is better.

Makefile
jschaul: `make -C services/brig docker` == `make docker-exe-brig docker-exe-brig-integration docker-exe-brig-schema`. I'll add a comment.

PR #569, opened 2019-01-14
Change of team permission semantics

fisx:

PR #572, opened 2019-01-16
Change of team permission semantics

fisx:
Fixes #569

~~I'm not sure what the implications to the client are, though.  The permission matrix for the `member` role has not changed, but the semantics of the individual roles has, at least on the backend.  Do clients need to be aware of that change?~~

All roles that have been around before this change remain unchanged in their permission bits (the semantics change slightly, but that doesn't affect clients currently).

Note to the reviewer: you can either read the changes [relative to #569](https://github.com/wireapp/wire-server/compare/4c12778~1...e08b03a), or [relative to develop before #569](https://github.com/wireapp/wire-server/compare/ea7a19d~1...e08b03a)

### DIFF
--- a/services/galley/test/integration/API/Teams.hs
+++ b/services/galley/test/integration/API/Teams.hs
@@ -130,7 +130,7 @@ testCreateTeamWithMembers g b c _ = do
     owner <- Util.randomUser b
     user1 <- Util.randomUser b
     user2 <- Util.randomUser b
-    let pp = Util.symmPermissions [CreateConversation, CRUDConversationMember]
+    let pp = Util.symmPermissions [CreateConversation, AddRemoveConvMember]
     let m1 = newTeamMember' pp user1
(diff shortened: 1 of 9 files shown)


services/galley/test/integration/API/Teams.hs
```diff
@@ -243,7 +243,7 @@ testAddTeamMemberInternal :: Galley -> Brig -> Cannon -> Maybe Aws.Env -> Http (
 testAddTeamMemberInternal g b c a = do
     owner <- Util.randomUser b
     tid <- Util.createTeam g "foo" owner []
-    let p1 = Util.symmPermissions [AddTeamMember] -- permissions are irrelevant on internal endpoint
+    let p1 = Util.symmPermissions [GetBilling] -- permissions are irrelevant on internal endpoint
```
jschaul: I don't understand this change.

PR #575, opened 2019-01-18
In SCIM handlers, format (almost) all errors using SCIM error schema

neongreen:


services/spar/src/Spar/Scim.hs
```diff
@@ -108,8 +107,45 @@ apiScim = hoistScim (toServant (Scim.siteServer configuration))
...
+    toSpar :: Scim.ScimHandler Spar a -> Spar a
+    toSpar = Scim.fromScimHandler
+             (throwError . SAML.CustomServant . Scim.scimToServantErr)
+
+    -- Wrap all errors into the format required by SCIM.
+    --
+    -- FIXME: this doesn't catch impure exceptions (e.g. thrown with 'error').
+    -- Let's hope that SCIM clients can handle non-SCIM-formatted errors
+    -- properly. See <https://github.com/haskell-servant/servant/issues/1022>
+    -- for why it's hard to catch impure ex
```
fisx: I wouldn't call this a `FIXME`, but rather a `NB:`.  Just don't use impure exceptions.  But either way it certainly is worth mentioning this here.

services/spar/src/Spar/Scim.hs
```diff
@@ -108,8 +107,45 @@ apiScim = hoistScim (toServant (Scim.siteServer configuration))
...
+    wrapScimErrors :: Spar a -> Spar a
+    wrapScimErrors = over (_Spar . mapped) $ \io ->
+        try @_ @SomeException io <&> \case
+            -- We caught an exception that's not a Spar exception. It should be
```
fisx:
```suggestion
            -- We caught an exception that's not a Spar exception. It is
```

services/spar/src/Spar/Scim.hs
```diff
@@ -108,8 +107,45 @@ apiScim = hoistScim (toServant (Scim.siteServer configuration))
...
+            -- We caught a 'CustomServant' exception. It should be left as-is
+            -- because by now all SCIM errors have become 'CustomServant' errors
+            -- (thanks to 'toSpar') and we don't want to wrap them *again*.
+            Right (Left (SAML.CustomServant x)) ->
+                Left (SAML.CustomServant x)
+            -- We caught some other Spar exception. It should be wrapped into
+            -- SCIM.serverError and rethrown.
+            Right (Left sparError) ->
+                Left . SAML.CustomServant . Scim.scimToServantErr $
+                Scim.serverError (cs (errBody (sparToServantErr sparError)))
```
fisx: Why not just `Left sparError`?  Then it will be converted later on in `sparToServantErr` like before.

Come to think of it, I would like to add a constructor `... | SparScimError ScimError` to `SparCustomError`, change `toSpar` as follows, and leave the conversion to `ServantErr` to the `Error` module:

```
toSpar = Scim.fromScimHandler (throwError . SparScimError)
```

We will still have this weird construct where the saml library wraps all errors, even those that have nothing to do with saml, but it keeps the errors in the app logic more strongly typed, and moves the burdon of rendering them more cleanly to the `Error` module.

fisx: (do you want me to do that?)

neongreen: I'm okay with the constructor, as long as the /scim/v2 branch uses the SCIM format for all errors (as the spec requires). We also should take care to not wrap SCIM errors twice.

fisx: agreed.  let me know if you want me to do any of this.

neongreen: I will get around to it at some point but I'm not sure when.

neongreen:
I'll return to this PR once #698 is merged.

services/spar/src/Spar/Scim.hs
```diff
@@ -79,9 +85,53 @@ apiScim :: ServerT APIScim Spar
...
+
+    -- Wrap /all/ errors into the format required by SCIM, even server exceptions that have
+    -- nothing to do with SCIM.
+    --
+    -- FIXME: this doesn't catch impure exceptions (e.g. thrown with 'error').
+    -- Let's hope that SCIM clients can handle non-SCIM-formatted errors
+    -- properly. See <https://github.com/haskell-servant/servant/issues/1022>
+    -- for why it's hard to catch impure exceptions.
+    wrapScimErrors :: Spar a -> Spar a
+    wrapScimErrors = over _Spar $ \act -> \env -> do
```
fisx:
```suggestion
    wrapScimErrors = over _Spar $ \act env -> do
```

services/spar/src/Spar/Scim.hs
```diff
@@ -79,9 +85,53 @@ apiScim :: ServerT APIScim Spar
...
+
+        case result of
+            -- We caught an exception that's not a Spar exception at all. It is wrapped into
+            -- Scim.serverError.
+            Left someException ->
+                pure $ Left . SAML.CustomError . SparScimError $
+                    Scim.serverError (cs (displayException someException))
+
+            -- We caught a 'SparScimError' exception. It is left as-is.
+            Right (Left (SAML.CustomError (SparScimError x))) ->
```
fisx:
```suggestion
            Right err@(Left (SAML.CustomError (SparScimError x))) ->
```

services/spar/src/Spar/Scim.hs
```diff
@@ -79,9 +85,53 @@ apiScim :: ServerT APIScim Spar
...
+                pure $ Left (SAML.CustomError (SparScimError x))
```
fisx:
```suggestion
                pure err
```

fisx:
sorry about the glitch! :/

PR #576, opened 2019-01-18
Make role-perm mapping more robust to change.

fisx:
[as discussed here](https://github.com/wireapp/wire-server/pull/573#discussion_r249049467)

### DIFF
--- a/libs/galley-types/test/unit/Test/Galley/Types.hs
+++ b/libs/galley-types/test/unit/Test/Galley/Types.hs
@@ -1,0 +1,24 @@
+{-# LANGUAGE OverloadedStrings   #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TypeApplications    #-}
+
+module Test.Galley.Types where
+
+import Imports
+import Control.Lens
+import Galley.Types.Teams
+import Test.Tasty
+import Test.Tasty.HUnit
+import Data.Set
+
+tests :: TestTree
+tests = testGroup "Tests"
+    [ testCase "ownwer has all permissions" $
+          rolePermissions RoleOwner @=? fullPermissions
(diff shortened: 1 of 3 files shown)


fisx:
@tiago-loureiro I'm pretty sure that's not what you had in mind, but I hope you'll like it anyway.  :-)

The reason I didn't want to touch `rolePerms` is that right now it is written down exactly the way we wrote it down in the specs, which I think is nice.

If you prefer, I will change just the `rolePerms RoleOwner` case and use `[minBound..]` there, and remove the tests.  But I would be a little sad :-).

libs/galley-types/test/unit/Test/Galley/Types.hs
```diff
@@ -0,0 +1,24 @@
...
+import Imports
+import Control.Lens
+import Galley.Types.Teams
+import Test.Tasty
+import Test.Tasty.HUnit
+import Data.Set
+
+tests :: TestTree
+tests = testGroup "Tests"
+    [ testCase "ownwer has all permissions" $
```
neongreen: Typo

libs/galley-types/test/unit/Test/Galley/Types.hs
```diff
@@ -0,0 +1,24 @@
...
+          rolePermissions RoleOwner @=? fullPermissions
+    , testCase "smaller roles (further to the left/top in the type def) are strictly more powerful" $
+          -- we may not want to maintain this property in the future when adding more roles, but for
+          -- now it's true, and it's nice to have that written down somewhere.
+          forM_ [(r1, r2) | r1 <- [minBound..], r2 <- [minBound..]] $ \(r1, r2) -> do
```
neongreen: r2 <- drop 1 [r1..]

And then you don't need ||

PR #577, opened 2019-01-21
Always carry a Role in the Invitation type.

fisx:
Without this, the default role would have to be picked twice: once in the backend, and once in team settings.  This is both extra work and poses the risk of running out of sync.

With these changes, the backend stores all future `Invitation`s in the DB with a default role that is defined near the type.  For migration, we allow for database entries with missing `Role`, and fill in the default there again.  This is marked in the code as `FUTUREWORK: remove`.

### DIFF
--- a/libs/brig-types/src/Brig/Types/Swagger.hs
+++ b/libs/brig-types/src/Brig/Types/Swagger.hs
@@ -391,10 +391,10 @@ invitation = defineModel "Invitation" $ do
     property "inviter" bytes' $
         description "User ID of the inviter"
     property "role" role $ do
-        description "User ID of the inviter"
+        description "Role of the invited user"
         optional
     property "id" bytes' $
-        description "UUID used to refer the invitation"
+        description "UUID used to refer to the invitation"
     property "email" string' $ do
         description "Email of the invitee"
         optional
--- a/libs/brig-types/src/Brig/Types/Swagger.hs
+++ b/libs/brig-types/src/Brig/Types/Swagger.hs
@@ -2,10 +2,11 @@
 
 module Brig.Types.Swagger where
 
-import Imports
 import Data.Swagger
 import Data.Swagger.Build.Api
+import Galley.Types.Teams (defaultRole)
 
+import qualified Data.Swagger.Model.Api     as Model
 import qualified Galley.Types.Swagger       as Galley
 import qualified Galley.Types.Teams.Swagger as Galley
 
@@ -375,11 +376,23 @@ invitationRequest = defineModel "InvitationRequest" $ do
         description "Locale to use for the invitation."
         optional
 
+role :: DataType
+role = Model.Prim $ Model.Primitive
+    { Model.primType     = Model.PrimString
+    , Model.defaultValue = Just defaultRole
+    , Model.enum         = Just [minBound..]
+    , Model.minVal       = Just minBound
+    , Model.maxVal       = Just maxBound
+    }
+
 invitation :: Model
 invitation = defineModel "Invitation" $ do
     description "An invitation to join Wire"
     property "inviter" bytes' $
         description "User ID of the inviter"
+    property "role" role $ do
+        description "User ID of the inviter"
+        optional
     property "id" bytes' $
         description "UUID used to refer the invitation"
     property "email" string' $ do
--- a/libs/brig-types/src/Brig/Types/Swagger.hs
+++ b/libs/brig-types/src/Brig/Types/Swagger.hs
@@ -2,6 +2,7 @@
 
 module Brig.Types.Swagger where
 
+import Imports
 import Data.Swagger
 import Data.Swagger.Build.Api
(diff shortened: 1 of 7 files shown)


libs/brig-types/src/Brig/Types/Team/Invitation.hs
```diff
@@ -49,7 +49,7 @@ instance ToJSON InvitationRequest where
 instance FromJSON Invitation where
     parseJSON = withObject "invitation" $ \o ->
         Invitation <$> o .: "team"
-                   <*> o .:? "role"
+                   <*> o .: "role"
```
jschaul: old clients won't have this field, this change is backwards-incompatible.

libs/brig-types/src/Brig/Types/Team/Invitation.hs
```diff
@@ -49,7 +49,8 @@ instance ToJSON InvitationRequest where
 instance FromJSON Invitation where
     parseJSON = withObject "invitation" $ \o ->
         Invitation <$> o .: "team"
-                   <*> o .:? "role"
+                       -- clients don't can leave the default role choice to us:
```
jschaul:
```suggestion
                       -- clients, when leaving "role" empty, can leave the default role choice to us
```

jschaul: What's the swagger for `Invitation`? Does that indicate role is optional or required?

libs/brig-types/src/Brig/Types/Swagger.hs
```diff
@@ -375,11 +376,23 @@ invitationRequest = defineModel "InvitationRequest" $ do
...
+    , Model.maxVal       = Just maxBound
+    }
+
 invitation :: Model
 invitation = defineModel "Invitation" $ do
     description "An invitation to join Wire"
     property "inviter" bytes' $
         description "User ID of the inviter"
+    property "role" role $ do
+        description "User ID of the inviter"
```
jschaul: copy paste error.

fisx: sorry.

...  third try?  (or is it 4th?)

services/brig/src/Brig/Team/API.hs
```diff
@@ -166,7 +166,8 @@ createInvitation (_ ::: uid ::: tid ::: req) = do
     body :: InvitationRequest <- parseJsonBody req
     idt  <- maybe (throwStd noIdentity) return =<< lift (fetchUserIdentity uid)
     from <- maybe (throwStd noEmail)    return (emailIdentity idt)
-    let inviteePerms = Team.rolePermissions . fromMaybe Team.RoleMember . irRole $ body
+    let inviteePerms = Team.rolePermissions inviteeRole
+        inviteeRole  = fromMaybe Team.RoleMember . irRole $ body
```
neongreen: Use `Team.defaultRole`?

services/brig/src/Brig/Team/DB.hs
```diff
@@ -162,5 +163,8 @@ countInvitations t = fromMaybe 0 . fmap runIdentity <$>
     cqlSelect :: PrepQuery R (Identity TeamId) (Identity Int64)
     cqlSelect = "SELECT count(*) FROM team_invitation WHERE team = ?"
 
+-- | FUTUREWORK: two weeks after https://github.com/wearezeta/backend-issues/issues/775 has been
```
neongreen: Is this an accident, or is the rule going to be "backend-issues can be referred-to in code but not in issue comments"?

services/brig/src/Brig/Team/API.hs
fisx: will fix!

services/brig/src/Brig/Team/DB.hs
fisx: accident.  :-)  will fix, too!

services/brig/src/Brig/Team/DB.hs
```diff
@@ -162,5 +163,8 @@ countInvitations t = fromMaybe 0 . fmap runIdentity <$>
...
+-- deployed, we won't find any `Nothing` role values any more; then it will be safe to change the
```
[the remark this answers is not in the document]
neongreen: > So Maybe Role / defaultRole should stay here.

I like this principle.

I don't like this principle applied inconsistently, however. Sometimes we leave stuff in, sometimes we deprecate things (like managed convs) but make it fairly easy for people to reenable them, sometimes we just don't communicate anything to anybody at all (like when a DB migration tool has to be run manually and we just write a tool but don't write anywhere that it has to be used).

If we decide to apply this principle consistently then I'd like everyone to explicitly agree on that because otherwise I have no idea what I'm allowed / not allowed to do and what I should / shouldn't look for when reviewing other people's PRs.

julialongtin:
additionally, do we have a place where we place things we've all agreed to?

PR #579, opened 2019-01-21
Make adding users implicit in CreateConversation permission.

fisx:
without this, collaborators can only create *empty* conversations, but they need to create 1:1 convs, too.

### DIFF
--- a/services/galley/src/Galley/API/Create.hs
+++ b/services/galley/src/Galley/API/Create.hs
@@ -84,6 +84,8 @@ createTeamGroupConv zusr zcon tinfo body = do
             let uu = filter (/= zusr) $ map (view userId) mems
             checkedConvSize uu
         else do
+            when (length mems > 1) $ do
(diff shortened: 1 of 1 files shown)


services/galley/src/Galley/API/Create.hs
```diff
@@ -84,7 +84,6 @@ createTeamGroupConv zusr zcon tinfo body = do
             let uu = filter (/= zusr) $ map (view userId) mems
             checkedConvSize uu
         else do
-            void $ permissionCheck zusr AddRemoveConvMember mems
```
tiago-loureiro: Let's change this into sth like: _if user has no AddConvMember permission, then ensure size <= 1_

fisx:
> Let's ensure no large groups are created by collaborators

it's a change in plan that clients should limit the kinds of conversations a collaborator can create, but that seems fine.  i have no strong opinion either way.

PR #580, opened 2019-01-22
make makedeb buildable for non-amd64 architectures.

julialongtin:


tools/makedeb/Makefile
```diff
@@ -2,10 +2,11 @@ LANG := en_US.UTF-8
 SHELL        := /usr/bin/env bash
 NAME         := makedeb
 VERSION      ?=
+ARCH  	     := $(shell [ -f "`which dpkg-architecture`" ]  && dpkg-architecture -qDEB_HOST_ARCH )
```
fisx: What happens if ``[ ! -f "`which dpkg-architecture`" ]``?  That probably shouldn't happen, so maybe throw a helpful error?

julialongtin: it quietly does nothing, and if you don't have dpkg-architecture, you do not have the ability to build debian packages anyways, as you are not on a debian-alike system.

fisx: > it quietly does nothing, and if you don't have dpkg-architecture, you do not have the ability to build debian packages anyways, as you are not on a debian-alike system.

then wouldn't it be better to just do this?:

```suggestion
ARCH  	     := $(shell dpkg-architecture -qDEB_HOST_ARCH )
```

i prefer "complain loudly" over "quietly do nothing" if what i clearly intend to do is not going to work.  (i'm talking about the general concept, in this case it's not that big a deal, so i already approved.)

julialongtin:
i went with 'be quiet' because the makefile allows someone to do a stack
build (in the compile rule), and a stack install (in the install rule),
without requiring you to build a .deb package. so it's possible to install
this, and install dpkg-deb, without being on debian. i've personally
installed dpkg-deb, without having dpkg-architecture, back in my slackware
days.

PR #581, opened 2019-01-22
SCIM -> Scim

neongreen:
I wanted to do it after the SCIM pull-requests are merged, but now I think it would be better to merge it before (so that I can work on SCIM PRs more comfortably).


fisx:
i'll review it when it passes ci.

PR #586, opened 2019-01-23
Use network-2.7 for more informative "connection failed" errors

neongreen:
```
service: Network.Socket.getAddrInfo: does not exist (Name does not resolve)
```

Starting from network-2.7, `getAddrInfo` also prints the host that it was trying to connect to.

### DIFF
--- a/tools/bonanza/src/Bonanza/Metrics.hs
+++ b/tools/bonanza/src/Bonanza/Metrics.hs
@@ -1,7 +1,9 @@
 {-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE RecordWildCards   #-}
 
-{-# OPTIONS_GHC -Wno-deprecations #-}  -- for Network.BSD
+-- Network.BSD got deprecated in network-2.7; this line won't be needed once we
+-- move to network-3.0 because then we can use the network-bsd package
+{-# OPTIONS_GHC -Wno-deprecations #-}
 
 module Bonanza.Metrics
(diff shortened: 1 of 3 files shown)


neongreen:
Perhaps it's better to leave it until the evening so that alpine-builder can be rebuilt promptly.

tools/bonanza/src/Bonanza/Metrics.hs
```diff
@@ -1,6 +1,8 @@
 {-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE RecordWildCards   #-}
 
+{-# OPTIONS_GHC -Wno-deprecations #-}  -- for Network.BSD
+
```
fisx: Couldn't you resolve the deprecations?

neongreen: No. `Network.BSD` was moved to `network-bsd`, which is not deprecated but is `network >= 3.0` only. The ecosystem hasn't moved to `network-3.0` yet so we can't use it. Tracking issue: https://github.com/commercialhaskell/stackage/issues/4312

neongreen: My understanding is that those deprecations are kinda "we're going to move it out", not "don't use it because better options are available".

neongreen:
By golly! Not 1️⃣ but 2️⃣ green circles.

Then I will merge this one in the evening today and rebuild alpine-builder.

PR #587, opened 2019-01-23
allow non-amd64 architecture, and work around 32 bit portability issue.

julialongtin:


libs/libzauth/libzauth-c/Makefile
```diff
@@ -1,7 +1,7 @@
 LANG := en_US.UTF-8
 SHELL   := /usr/bin/env bash
 VERSION := "3.0.0"
-ARCH    := amd64
+ARCH    := $(shell if [ -f "`which dpkg-architecture`" ]; then dpkg-architecture -qDEB_HOST_ARCH; else [ -f "`which dpkg`" ] && dpkg --print-architecture; fi )
```
fisx: That's different from a similar line in the wire-server repo.  Is there a good reason for that?

libs/libzauth/libzauth-c/src/lib.rs
```diff
@@ -130,10 +130,11 @@ pub extern fn zauth_token_type(t: &ZauthToken) -> ZauthTokenType {
...
 
-#[no_mangle]
-pub extern fn zauth_token_time(t: &ZauthToken) -> c_long {
-    t.0.timestamp
-}
+// Commented out, looks unused, and causing portability issues with ia32.
+//#[no_mangle]
+//pub extern fn zauth_token_time(t: &ZauthToken) -> c_long {
+//    t.0.timestamp
+//}
```
jschaul: while this particular function might indeed not be used by the [nginz zauth module](https://github.com/wireapp/wire-server/blob/develop/services/nginz/third_party/nginx-zauth-module/zauth_module.c), I have a question: Does the underlying issue with the timestamp remain in ia32 upon [token verification](https://github.com/wireapp/wire-server/blob/develop/libs/libzauth/libzauth/src/zauth.rs#L138), i.e. in [is_expired](https://github.com/wireapp/wire-server/blob/develop/libs/libzauth/libzauth/src/zauth.rs#L166-L169)?

Can you confirm you can compile nginz, and in demo-mode nginz correctly handles expired vs non-expired tokens under ia32? If not, commenting out this code won't help much. If this is just an FFI portability issue, but verification happens as expected under nginz, I'd prefer removing the code, rather than commenting it out.

libs/libzauth/libzauth-c/Makefile
julialongtin: To which other line are you referring? This is the second version of this construct, due to dpkg-dev not being installed on some of our docker images.

fisx: https://github.com/wireapp/wire-server/blob/7a831025b29d0b9e9d33176a48115f251b578e37/tools/makedeb/Makefile#L5

oops, same line, as it seems.  ignore me!

libs/libzauth/libzauth-c/src/lib.rs
julialongtin: I can confirm that i can compile / launch nginz, but i have no idea how to perform the test you just aluded to.

julialongtin: it looks like the code you've pointed to is using 64 bit integers, even on ia32, so should not have any problems until December 4th, 292,277,026,596 AD.

PR #590, opened 2019-01-23
break PREFIX into PREFIX_PACKAGE and PREFIX_INSTALL, to fix the path …

julialongtin:
…when building debian packages as a non-root user.


libs/libzauth/libzauth-c/Makefile
```diff
@@ -19,7 +21,7 @@ all: build
 clean:
 	cargo clean
 	rm -rf test/target
-	rm -rf deb$(PREFIX)/
```
fisx: don't we want to clean both here?

julialongtin: cleaning an install (which is what prefix_install is for) would be performing an uninstall. the install rule places results directly on the filesystem, not in a temporary location, so it does not need cleaned up after.

PR #591, opened 2019-01-23
add a readme, for how to build libzauth.

julialongtin:


libs/libzauth/README.md
```diff
@@ -0,0 +1,23 @@
...
+
+### Building / Installing
+
+```bash
+make dist
+sudo dpkg -i libzauth*.deb
+sudo ldconfig
+```
+
+## Generic
```
fisx: Isn't this the same as "compile and install natively"?

PR #592, opened 2019-01-23
Tidy up partner role.

fisx:
This changes will mess with staging only, since none of the features touched have seen master yet.  All changes have been agreed on with the other teams.

1. json representation of TeamMember will look more like the one of Invitation, i.e. instead of { invited: { by: <uuid>, at: <timestamp> }, it will contain { created_by: <uuid>, created_at: <timestamp> }.
2. collaborator will disappear from everywhere, and partner will be used instead.  (when you send collaborator, the backend will hear "partner", but when you expect to receive collaborator, you are in trouble.

### DIFF
--- a/libs/brig-types/test/unit/Test/Brig/Types/User.hs
+++ b/libs/brig-types/test/unit/Test/Brig/Types/User.hs
@@ -6,6 +6,8 @@
 {-# LANGUAGE TypeFamilies        #-}
 {-# LANGUAGE ViewPatterns        #-}
 
+{-# OPTIONS_GHC -Wno-orphans #-}
+
 module Test.Brig.Types.User where
 
 import Imports
@@ -18,6 +20,7 @@ import Data.Aeson
 import Data.Aeson.Types
 import Data.Proxy
 import Data.Typeable (typeOf)
+import Galley.Types.Teams
 import Test.Brig.Types.Arbitrary ()
 import Test.QuickCheck
 import Test.Tasty
@@ -88,6 +91,7 @@ roundtripTests =
     , run @PhoneUpdate Proxy
     , run @ReAuthUser Proxy
     , run @SelfProfile Proxy
+    , run @TeamMember Proxy
     , run @UpdateServiceWhitelist Proxy
     , run @UserHandleInfo Proxy
     , run @UserIdentity Proxy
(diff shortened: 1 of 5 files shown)


fisx:
all integration tests pass locally.

libs/brig-types/test/unit/Test/Brig/Types/User.hs
```diff
@@ -6,6 +6,8 @@
 {-# LANGUAGE TypeFamilies        #-}
 {-# LANGUAGE ViewPatterns        #-}
 
+{-# OPTIONS_GHC -Wno-orphans #-}
```
neongreen: Can you write what instance this is needed for?

libs/galley-types/src/Galley/Types/Teams.hs
```diff
@@ -499,12 +499,10 @@ instance ToJSON TeamMember where
...
 teamMemberJson :: (TeamMember -> Bool) -> TeamMember -> Value
 teamMemberJson withPerms m = object $
-    [ "user" .= _userId m ] <>
+    [ "user"        .= _userId m ] <>
     [ "permissions" .= _permissions m | withPerms m ] <>
-    [ "invited" .= (invmetaJson <$> _invitation m) ]
-  where
-    invmetaJson :: (UserId, UTCTimeMillis) -> Value
-    invmetaJson (by, at) = object [ "by" .= by, "at" .= at ]
+    [ "created_by"  .= (fst <$> _invitation m) ] <>
```
neongreen: Is this a non-breaking change?

libs/galley-types/src/Galley/Types/Teams.hs
```diff
@@ -515,13 +513,16 @@ parseTeamMember :: Value -> Parser TeamMember
 parseTeamMember = withObject "team-member" $ \o ->
     TeamMember <$> o .:  "user"
                <*> o .:  "permissions"
-               <*> (parseInv =<< (o .:? "invited"))
```
neongreen: So the "invited" field is not present anymore?

libs/galley-types/src/Galley/Types/Teams.hs
neongreen: > All changes have been agreed on with the other teams.

Oh okay.

libs/galley-types/src/Galley/Types/Teams.hs
neongreen: > All changes have been agreed on with the other teams.

Oh also okay.

neongreen:
Sure

PR #593, opened 2019-01-23
Add support for disallowing certain phone number prefixes

jschaul:
This PR is part of a cat and mouse game.

Happy to take comments about perhaps more consistent naming (`exclude`, `block`, `disallow`, `blacklist`, ... ?).

### DIFF
--- a/libs/brig-types/src/Brig/Types/Common.hs
+++ b/libs/brig-types/src/Brig/Types/Common.hs
@@ -187,6 +187,18 @@ instance FromByteString PhonePrefix where
 instance ToByteString PhonePrefix where
     builder = builder . fromPhonePrefix
 
+data ExcludedPrefix = ExcludedPrefix { phonePrefix :: PhonePrefix
+                                     , comment :: Text
+                                     } deriving (Eq, Show)
+
+instance FromJSON ExcludedPrefix where
+    parseJSON = withObject "ExcludedPrefix" $ \o -> ExcludedPrefix
+        <$> o .: "phone_prefix"
+        <*> o .: "comment"
+
+instance ToJSON ExcludedPrefix where
+    toJSON (ExcludedPrefix p c) = object ["phone_prefix" .= p, "comment" .= c]
+
 -----------------------------------------------------------------------------
 -- UserIdentity
 
--- a/libs/brig-types/src/Brig/Types/Common.hs
+++ b/libs/brig-types/src/Brig/Types/Common.hs
@@ -169,6 +169,11 @@ isValidPhonePrefix = either (const False) (const True) . parseOnly e164Prefix
   where
     e164Prefix = char '+' *> count 1 digit *> count 14 (optional digit) *> endOfInput
 
+
+phonePrefixes :: Phone -> [PhonePrefix]
+phonePrefixes p = catMaybes $ parsePhonePrefix <$> Text.inits (fromPhone p)
+
+
 instance FromJSON PhonePrefix where
     parseJSON (String s) = case parsePhonePrefix s of
         Just p  -> return p
--- a/libs/brig-types/src/Brig/Types/Common.hs
+++ b/libs/brig-types/src/Brig/Types/Common.hs
@@ -151,6 +151,37 @@ instance ToJSON PhoneBudgetTimeout where
     toJSON (PhoneBudgetTimeout t) = object [ "expires_in" .= t ]
 
 -----------------------------------------------------------------------------
+-- PhonePrefix (for excluding from SMS/calling)
+
+newtype PhonePrefix = PhonePrefix { fromPhonePrefix :: Text } deriving (Eq, Show, ToJSON)
+
+-- | Parses a phone number prefix with a mandatory leading '+'.
+parsePhonePrefix :: Text -> Maybe PhonePrefix
+parsePhonePrefix p
+    | isValidPhonePrefix p = Just $! PhonePrefix p
+    | otherwise      = Nothing
+
+-- | Checks whether a phone number prefix is valid,
+-- i.e. it is like a E.164 format phone number, but shorter
+-- (with a mandatory leading '+', followed by 1-15 digits.)
+isValidPhonePrefix :: Text -> Bool
+isValidPhonePrefix = either (const False) (const True) . parseOnly e164Prefix
+  where
+    e164Prefix = char '+' *> count 1 digit *> count 14 (optional digit) *> endOfInput
+
+instance FromJSON PhonePrefix where
+    parseJSON (String s) = case parsePhonePrefix s of
+        Just p  -> return p
+        Nothing -> fail "Invalid phone number prefix. Expected E.164 format (with only one digit or more allowed after the +)."
+    parseJSON _          = mempty
+
+instance FromByteString PhonePrefix where
+    parser = parser >>= maybe (fail "Invalid phone") return . parsePhonePrefix
+
+instance ToByteString PhonePrefix where
+    builder = builder . fromPhonePrefix
+
+-----------------------------------------------------------------------------
 -- UserIdentity
 
 -- | The private unique user identity that is used for login and
--- a/libs/brig-types/src/Brig/Types/Common.hs
+++ b/libs/brig-types/src/Brig/Types/Common.hs
@@ -165,7 +165,7 @@ parsePhonePrefix p
 -- i.e. it is like a E.164 format phone number, but shorter
 -- (with a mandatory leading '+', followed by 1-15 digits.)
 isValidPhonePrefix :: Text -> Bool
-isValidPhonePrefix = either (const False) (const True) . parseOnly e164Prefix
+isValidPhonePrefix = isRight . parseOnly e164Prefix
   where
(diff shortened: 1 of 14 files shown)


services/brig/src/Brig/API/User.hs
```diff
@@ -524,15 +529,23 @@ sendActivationCode emailOrPhone loc call = case emailOrPhone of
...
         when exists $
             throwE $ UserKeyInUse pk
         blacklisted <- lift $ Blacklist.exists pk
         when blacklisted $
             throwE (ActivationBlacklistedUserKey pk)
+
+        -- check if any prefixes of this phone number are blocked
+        prefixExcluded <- lift $ Blacklist.existsAnyPrefix (canonical)
+        when prefixExcluded $
+            throwE (ActivationBlacklistedUserKey pk)
```
jschaul: Should we differentiate the error here?

libs/brig-types/src/Brig/Types/Common.hs
```diff
@@ -150,6 +150,45 @@ instance FromJSON PhoneBudgetTimeout where
...
+phonePrefixes :: Phone -> [PhonePrefix]
+phonePrefixes p = allPrefixes (fromPhone p)
+
+-- | get all valid prefixes of a phone number or phone number prefix
+-- e.g. from +123456789 get prefixes ["+1", "+12", "+123", ..., "+123456789" ]
+allPrefixes :: Text -> [PhonePrefix]
+allPrefixes t = catMaybes $ parsePhonePrefix <$> Text.inits t
+
+instance FromJSON PhonePrefix where
+    parseJSON (String s) = case parsePhonePrefix s of
```
neongreen: `withText` is better than `mempty` because it gives a descriptive error message in case of JSON type mismatch.

services/brig/src/Brig/API/User.hs
neongreen: I think not.

services/brig/src/Brig/Data/Blacklist.hs
```diff
@@ -27,3 +40,41 @@ keySelect = "SELECT key FROM blacklist WHERE key = ?"
...
+-- Excluded phone prefixes
+
+insertPrefix :: MonadClient m => PhonePrefix -> m ()
+insertPrefix prefix = retry x5 $ write prefixInsert (params Quorum (Identity $ fromPhonePrefix prefix))
+
+deletePrefix :: MonadClient m => PhonePrefix -> m ()
+deletePrefix prefix = retry x5 $ write prefixDelete (params Quorum (Identity $ fromPhonePrefix prefix))
+
+existsPrefix :: MonadClient m => PhonePrefix -> m Bool
+existsPrefix prefix = return . isJust =<< fmap runIdentity <$>
```
neongreen: Not sure `fmap runIdentity` is needed here

neongreen: Also, `return . isJust =<<` == `isJust <$>`

libs/brig-types/src/Brig/Types/Common.hs
```diff
@@ -150,6 +150,45 @@ instance FromJSON PhoneBudgetTimeout where
...
 
+-----------------------------------------------------------------------------
+-- PhonePrefix (for excluding from SMS/calling)
+
+newtype PhonePrefix = PhonePrefix { fromPhonePrefix :: Text } deriving (Eq, Show, ToJSON)
+
+-- | Parses a phone number prefix with a mandatory leading '+'.
+parsePhonePrefix :: Text -> Maybe PhonePrefix
+parsePhonePrefix p
+    | isValidPhonePrefix p = Just $! PhonePrefix p
```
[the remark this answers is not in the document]
jschaul: I admit I copy pasted from `parsePhone` a few lines above. Removed the `!` as I don't think it's necessary.

libs/brig-types/src/Brig/Types/Common.hs
jschaul: thanks.

services/brig/src/Brig/Data/Blacklist.hs
jschaul: thanks. Actually, this function ended up not being used, so it's removed entirely.

services/brig/schema/src/V56.hs
```diff
@@ -0,0 +1,25 @@
...
+migration :: Migration
+migration = Migration 56 "Add table to exclude phone number prefixes" $ do
+    -- A table for manual excluding of phone number prefixes that abuse sms sending
+    -- and/or calling.
+    --
+    -- Operations we need to support:
+    --   * Add a new prefix of arbitrary length
+    --   * Remove a prefix
+    --   * Given a phone number, check whether it matches any existing prefix
+    void $ schema' [r|
```
neongreen: Sorry for only realizing it now, but I think it could be good to also add a "comment" field to the table (and even make it mandatory and non-empty). Otherwise in some time we might end up with a bunch of blacklisted prefixes in the database and no idea why or when they were blacklisted.

services/brig/src/Brig/API.hs
```diff
@@ -190,6 +190,15 @@ sitemap o = do
     post "/i/users/blacklist" (continue addBlacklist) $
         param "email" ||| param "phone"
 
+    get "/i/users/phone-prefix" (continue getPhonePrefixes) $
```
neongreen: The usecase is going to be "look up a certain phone to figure out why it's blocked", right? Or..?

(Whatever the usecase is, maybe there should be a comment about the intended usecase.)

services/brig/src/Brig/API/User.hs
```diff
@@ -524,15 +529,23 @@ sendActivationCode emailOrPhone loc call = case emailOrPhone of
             Just (Just uid, c) -> sendActivationEmail   ek c        uid -- User re-requesting activation
 
     Right phone -> do
-        pk <- maybe (throwE $ InvalidRecipient (userPhoneKey phone))
-                    (return . userPhoneKey)
+        -- validatePhone returns the canonical E.164 phone number format
+        canonical <- maybe (throwE $ InvalidRecipient (userPhoneKey phone))
+                    (return)
```
fisx: - brackets are not needed.
- i prefer `pure` to `return`, but the wire code base disagrees.

services/brig/src/Brig/API/User.hs
```diff
@@ -524,15 +529,23 @@ sendActivationCode emailOrPhone loc call = case emailOrPhone of
...
+        let pk = userPhoneKey canonical
         exists <- lift $ isJust <$> Data.lookupKey pk
         when exists $
             throwE $ UserKeyInUse pk
         blacklisted <- lift $ Blacklist.exists pk
         when blacklisted $
             throwE (ActivationBlacklistedUserKey pk)
+
+        -- check if any prefixes of this phone number are blocked
+        prefixExcluded <- lift $ Blacklist.existsAnyPrefix (canonical)
```
fisx: you could drop some brackets

services/brig/src/Brig/Data/Blacklist.hs
```diff
@@ -27,3 +39,33 @@ keySelect = "SELECT key FROM blacklist WHERE key = ?"
...
+    return (PhonePrefix <$> results)
+
+existsAnyPrefix :: MonadClient m => Phone -> m Bool
+existsAnyPrefix phone = do
+    let prefixes = fromPhonePrefix <$> allPrefixes (fromPhone phone)
+    results <- fmap runIdentity <$> retry x1 (query prefixSelectAll (params Quorum (Identity $ prefixes)))
+    return $ (not . null) results
+
+prefixInsert :: PrepQuery W (Identity Text) ()
+prefixInsert = "INSERT INTO excluded_phones (prefix) VALUES (?)"
```
fisx: i've seen this pattern of flipping two words in a composite to make up a new identifier, where the difference in the names has nothing to do with the difference in semantics.  didn't like it back then either, but consistency may be more important.

also, i'm not sure the two identifiers need to be top-level, as each `PrepQuery` is called exactly once (no?).

services/brig/src/Brig/API.hs
fisx: also: why is it under `/i/users`?  (it's ok for constency with blacklist, just curious about the original reason.)

services/brig/src/Brig/Data/Blacklist.hs
jschaul: I can make the CQL a sub function, [if you prefer?](https://github.com/wireapp/wire-server/pull/593/commits/69e22b6cb0b1f4339239d7342e9a1577b0ffeffa)

services/brig/src/Brig/API.hs
jschaul: the phone prefixes are related to users in some way (it blocks some phone numbers - a user's property), unlike e.g. `/i/monitoring`, which is unrelated to users.
Plural is `prefixes`, not `prefices`. Changed the paths to be plural.

services/brig/schema/src/V56.hs
jschaul: https://github.com/wireapp/wire-server/pull/593/commits/6605fa765dc3dcf8f9cab922dca91b1fcea2c942

PR #594, opened 2019-01-24
architecture independance, better use of make features, more docs.

julialongtin:


services/nginz/README.md
```diff
@@ -1,16 +1,38 @@
...
+sudo apt install libossp-uuid-dev libpcre3-dev libsodium23 runit
+```
+
+### Building / Installing
+```bash
+make dist
+dpkg -i build/nginz*.deb
+```
+
+##Generic
```
jschaul:
```suggestion
## Generic
```

services/nginz/README.md
```diff
@@ -1,16 +1,38 @@
 # NGINX build with extra modules
 
-## Compile natively
+## Compile and Install natively
```
jschaul: The current nginz debian package (similarly to the other services), as it is, is not designed to be installed anywhere else than on our own infrastructure. You can see this in the assumptions the [run script makes](https://github.com/wireapp/wire-server/blob/develop/services/nginz/deb/etc/sv/nginz/run#L8-L10), namely the existence of a `www-user`, a `get_config` executable downloading a configuration file, and `runit`.

Without changing these assumptions, providing a README update on "how to install natively" is misleading.

For development and demo, the [integration](https://github.com/wireapp/wire-server/blob/develop/services/integration.sh) and [demo](https://github.com/wireapp/wire-server/blob/develop/deploy/services-demo/demo.sh) scripts start the binaries (in this case `nginz`) with the right configuration files for the use case.

The debian packages, when developed many years ago, fitted only one use case: work on our infrastructure. While expanding this use case can be discussed, updating the README without changing the run script makes little sense.

PR #597, opened 2019-01-25
Use custom snapshots

neongreen:
This uses Stack's custom snapshot feature (<https://docs.haskellstack.org/en/stable/custom_snapshot/>). A custom snapshot extends the set of packages in the resolver and can be used from any other project (including other Wire repos). No more copying parts of stack.yaml from one project to another.

Custom snapshots are stored in `~/.stack`. This has several benefits:

* They survive `stack clean`. (Well, Git deps have to be cloned anyway because of a bug in Stack, but they won't be recompiled, only recloned, and for the biggest deps I've switched to using tarballs so that should be faster.)

* They survive branch switches (i.e. if different package versions are used in different branches, no recompiling is necessary when switching between the two).

* We no longer need to cache both `.stack` and `.stack-work` on CI.

A thing to keep in mind is that snapshots are immutable. A snapshot can be based on the previous version of the snapshot, but Stack takes a bit of time to symlink everything from the older snapshot to the newer one, so having a new snapshot every day would be inadvisable. For that reason packages that change frequently (like hscim) should remain in `stack.yaml`.

### DIFF
--- a/snapshots/wire-1.0.yaml
+++ b/snapshots/wire-1.0.yaml
@@ -1,0 +1,120 @@
+# A custom Stack snapshot (https://docs.haskellstack.org/en/stable/custom_snapshot/) used for
+# Wire code. Should be never changed once pushed to develop.
+
+resolver: lts-12.10
+name: wire-1.0
+
+packages:
+
+############################################################
+# Packages where we need specific lower/upper bounds
+############################################################
+
+- async-2.2.1
+- hinotify-0.4
+- fsnotify-0.3.0.1
+- base-prelude-1.3
+- base58-bytestring-0.1.0
+- cql-4.0.1
+- currency-codes-2.0.0.0
+- data-timeout-0.3
+- geoip2-0.3.1.0
+- mime-0.4.0.2
+- multiset-0.3.4.1
+- text-icu-translit-0.1.0.7
+- wai-middleware-gunzip-0.0.2
+- network-uri-static-0.1.1.0  # includes 'relativeReference'
+- list-t-1.0.1  # v1.0.0.1 doesn't build
+- unliftio-0.2.10  # for pooled concurrency utils in UnliftIO.Async
+- network-2.7.0.2  # to get nicer errors when connections fail
+- HaskellNet-SSL-0.3.4.1  # first version to support network-2.7
+
+- git: https://github.com/kim/hs-collectd
+  commit: 885da222be2375f78c7be36127620ed772b677c9
+
+- git: https://github.com/kim/snappy-framing
+  commit: d99f702c0086729efd6848dea8a01e5266c3a61c
+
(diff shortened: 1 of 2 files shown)


snapshots/wire-1.0.yaml
```diff
@@ -0,0 +1,118 @@
+# A custom Stack snapshot (https://docs.haskellstack.org/en/stable/custom_snapshot/) used for
+# Wire code. Should be never changed once pushed to develop.
```
fisx: what happens if it does?  how can i recover?  (trying to understand, i'm not suggesting we should ever do that.)

neongreen: Stack has two ways to refer to snapshots:

* URL
* filepath

In wire-server we refer to the snapshot by filepath. If the snapshot file changes, Stack will detect changes and treat it as a new snapshot (all snapshots are indexed by hash so if you change the file and then change it back, the only thing you should do is delete the second snapshot from your system).

In other repositories we will refer to snapshots by URL. Stack only downloads URLs once and never checks them again. So, when building code from other repositories, people would be getting inconsistent results depending on whether they built that code before or not.

neongreen:
Should be updated after the SCIM PR is merged

snapshots/wire-1.0.yaml
jschaul: Typo:

```suggestion
# Wire code. Should never be changed once pushed to develop.
```

Do I understand correctly that if we want to bump the version of a library, `snapshots/wire-1.1.yaml` (a copy of this file with one version changed) file would be created, and `stack.yaml` would from then on refer to wire-1.1?

> Should be never changed once pushed to develop.

The snapshot docs specify "custom snapshot[s] are assumed to be mutable". So is this a guideline for us only? Perhaps add a comment on how to deal with added, removed, or updated dependencies.

neongreen: > Do I understand correctly that if we want to bump the version of a library, snapshots/wire-1.1.yaml (a copy of this file with one version changed) file would be created, and stack.yaml would from then on refer to wire-1.1?

Not a copy, more like a patch. Snapshots can be based on other snapshots. So, it would be something like:

```yaml
resolver: <link to wire-1.0.yaml>

packages: [updated packages here]

# there is also an ability to remove packages but I don't remember how it's done
```

Actually if we want that to be usable in other repositories, we should use a link in our stack.yaml as well, not a filepath. I'll amend the PR.

> The snapshot docs specify "custom snapshot[s] are assumed to be mutable".

If you refer to snapshots by filepath, then yes, they can be mutable. Stack will notice changes and just create a new snapshot (leaving the old one available, too).

If you refer to snapshots by URL, as we're planning to do in other repos, then they have to be immutable because Stack is only going to download the URL once.

PR #600, opened 2019-01-28
Make default $DOCKER_TAG == "latest".

### LINKED ISSUE #599: make docker-services fails because zauth.h is missing
### $ `make docker-services`

```
Step 6/13 : COPY --from=libzauth-builder /usr/include/zauth.h /usr/include/zauth.h
COPY failed: stat /var/lib/docker/overlay2/380620e81dd76b3ea925a64c6b8f3dedf7bf8c612402cefb1e4bb8798e2d1a02/merged/usr/include/zauth.h: no such file or directory
Makefile:124: recipe for target 'docker' failed
make[1]: *** [docker] Error 1
make[1]: Leaving directory '/home/hermes/projects/wire-server/services/nginz'
Makefile:124: recipe for target 'docker-services' failed
make: *** [docker-services] Error 2
```

fisx:
Fixes #599

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,7 @@
 SHELL            := /usr/bin/env bash
 LANG             := en_US.UTF-8
 DOCKER_USER      ?= quay.io/wire
-DOCKER_TAG       ?= local
+DOCKER_TAG       ?= latest
 
 default: fast
 
--- a/build/alpine/Makefile
+++ b/build/alpine/Makefile
@@ -1,6 +1,6 @@
 LANG := en_US.UTF-8
 DOCKER_USER   ?= quay.io/wire
-DOCKER_TAG    ?= local
+DOCKER_TAG    ?= latest
 
 default: deps prebuilder builder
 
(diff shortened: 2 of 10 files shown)


fisx:
@jschaul I am not very confident about this and would like to know what you think.  was there a reason to have `local` as default?  if so, is there a simple way to get rid of that reason?

jschaul:
This doesn't solve anything. All locally built docker images get tagged with `latest` in addition to `local`. The `local` tag can be used to distinguish them from images pulled from quay.io, if needed. You can perhaps close this PR?

fisx:
> This doesn't solve anything. All locally built docker images get tagged with `latest` in addition to `local`. The `local` tag can be used to distinguish them from images pulled from quay.io, if needed. You can perhaps close this PR?

I am closing it, but reluctantly so.  I would expect `latest` to be the default, and if I don't like that I have to override it, which is easy enough.  The current behavior (before this PR) is surprising because if I pull a new latest, it is only picked up if I never have built anything local.

If that convinces you, please re-open.  If you can think of a good place to add some documentation that helps with this, let me know (even though I prefer systems not to be surprising over many over surprises being well-documented).

Anyway, closing.  #605 is definitely more helpful!  :-)

PR #602, opened 2019-01-29
Keep track of who is allowed to modify users (we vs. SCIM)

neongreen:


libs/brig-types/src/Brig/Types/Common.hs
```diff
@@ -337,3 +337,25 @@ codeParser :: String -> (String -> Maybe a) -> Parser a
...
+    | ManagedBySCIM
+    deriving (Eq, Show)
+
+instance FromJSON ManagedBy where
+    parseJSON = withText "ManagedBy" $ \case
+        "wire" -> pure ManagedByWire
+        "scim" -> pure ManagedBySCIM
+        other  -> fail $ "Invalid ManagedBy: " ++ show other
+
+instance ToJSON ManagedBy where
```
fisx: can you also add a rountrip unit test?

neongreen: Done.

neongreen:
### Changes in this PR

* New field `managed_by` for all users, with two possible options ("wire" and "scim"). Users with "scim" can't be edited from Brig, but can be edited from Spar. The `managed_by` field is exposed in the `/self` profile so that client apps can look at it and decide whether to allow changes to the profile, or not.

* New event ("managed_by was changed").

* New internal endpoint `PUT /i/users/:uid/managed-by`, intended to be used by Spar to "claim" a user who was not provisioned via SCIM but who will from now on be managed by SCIM.

* Some reshuffling around `UserEvent` (no visible changes but I would like for whoever is familiar with that code to take a look).

### Planned changes in future PRs

* This PR keeps track of `managed_by` but doesn't enforce it. To enforce it I plan to create endpoints `PUT /i/self`, `PUT /i/self/email`, `PUT /i/self/phone`, ... and use them to make changes from Spar. Public endpoints would be checking `managed_by` first.

* New error type thrown by public endpoints (`403 "read-only"`).

* New endpoint `POST /scim/user-ownership/:uid <Bool>` to convert an existing user into a SCIM user or relinquish the ownership. Relinquish. That's a nice word.

fisx:
> To enforce it I plan to create endpoints PUT /i/self, PUT /i/self/email, PUT /i/self/phone, ... and use them to make changes from Spar. Public endpoints would be checking managed_by first.

i would also check that the internal end-points only work on users managed by scim.

libs/brig-types/src/Brig/Types/Swagger.hs
```diff
@@ -150,6 +154,12 @@ user = defineModel "User" $ do
         description "Unique user handle."
         optional
 
+managedBy :: DataType
+managedBy = string $ enum
+    [ "wire"
+    , "scim"
+    ]
```
fisx: Nice, i didn't know this when i wrote the swagger `DataType` for `Role`, will fix.

Is the following better (automatically keeps in sync with changes in the type) or worse (convoluted and hard to read)?  If better, I will change this everywhere in the code-base in a separate PR.

```
dataTypeFromBoundedEnum :: forall a proxy. (Bounded a, Enum a, Aeson.ToJSON a) => proxy a -> DataType
dataTypeFromBoundedEnum _ = string . enum $ unpack . Aeson.encode <$> [(minBound :: a)..]

role :: DataType
role = dataTypeFromBoundedEnum (Proxy @Role)
```

services/spar/src/Spar/Intra/Brig.hs
```diff
@@ -177,7 +177,6 @@ setName buid name = do
     . path "/self"
     . header "Z-User" (toByteString' buid)
     . header "Z-Connection" ""
-    . expect2xx
```
fisx: oops...

libs/brig-types/src/Brig/Types/Swagger.hs
neongreen: I'd say it's better as long as `dataTypeFromBoundedEnum` has a haddock explaining that the Aeson instance is used.

I would also prefer to use AllowAmbiguousTypes and `dataTypeFromBoundedEnum @Role`, but this isn't important.

neongreen:
> i would also check that the internal end-points only work on users managed by scim.

This is reasonable in the context of this PR, but I would say that internal endpoints should try to be general-purpose tools unless there are severe reasons to do otherwise. Strong coupling between services (this is in Spar because Brig needs it, this is in Brig because Spar needs it, etc) makes it harder to reason about them in isolation, and Brig is already hard to reason about.

(Sidenote: ideally I'd have a separation between user-facing services – e.g. `Captain` and `Spar` – and everything else. User-facing services don't care about database states and whatever, internal services don't question orders coming down the chain of command. No sure if anybody else would agree with this vision though.)

fisx:
> (Sidenote: ideally I'd have a separation between user-facing services – e.g. `Captain` and `Spar` – and everything else. User-facing services don't care about database states and whatever, internal services don't question orders coming down the chain of command. No sure if anybody else would agree with this vision though.)

Sounds possibly reasonable.  I would do that after fusing spar and galley, though.

neongreen:
> I would do that after fusing spar and galley, though.

You mean Brig and Galley?

fisx:
> > I would do that after fusing spar and galley, though.
> 
> You mean Brig and Galley?

yes, sorry.  don't fuse spar!  :)

services/spar/src/Spar/Scim.hs
```diff
@@ -246,6 +246,8 @@ createValidScimUser (ValidScimUser user uref handl mbName) = do
     -- Set user handle on brig (which can't be done during user creation yet).
     -- TODO: handle errors better here?
     lift $ Intra.Brig.setHandle buid handl
+    -- Flag the user as SCIM-managed
+    lift $ Intra.Brig.setManagedBy buid ManagedBySCIM
```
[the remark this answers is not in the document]
neongreen: Yeah, initially I went that way but then reverted the change. I don't remember what was the reason for reverting it :/

I will try it out again and maybe I'll remember.

libs/brig-types/src/Brig/Types/Common.hs
```diff
@@ -337,3 +337,34 @@ codeParser :: String -> (String -> Maybe a) -> Parser a
...
     maybe (fail err) return (conv code)
+
+-----------------------------------------------------------------------------
+-- ManagedBy
+
+-- | Who controls changes to the user profile.
+data ManagedBy
+      -- | The profile can be changed in-app; user doesn't show up via SCIM at all.
+    = ManagedByWire
+      -- | The profile can only be changed via SCIM, with several exceptions:
```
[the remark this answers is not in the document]
neongreen: The profile = all user-editable attributes.

"properties" aren't a part of that because we (the backend) don't even know what is stored there. It looks like it's mostly per-client settings and therefore not something SCIM should ever be bothered with: https://github.com/wireapp/wire-webapp/blob/dev/app/script/properties/PropertiesType.js#L22-L41

I foresee no inconsistencies. The only user-visible thing we allow changing is passwords, but SCIM only works for SSO-enabled users who (IIRC) don't have passwords anyway. Also users will (should) be notified when their password is changed via SCIM, so it's not like they will have something reverted silently.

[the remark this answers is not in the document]
neongreen: (I will add an expanded version of this comment to the docs.)

services/spar/src/Spar/Scim.hs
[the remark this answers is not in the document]
neongreen: @jschaul if you're still working, can you look at this one?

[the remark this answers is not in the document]
neongreen: (I pushed a commit that does that change)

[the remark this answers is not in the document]
neongreen: Note: setManagedBy is currently unused, but will be used in the next PR.

libs/brig-types/src/Brig/Types/Common.hs
```diff
@@ -386,3 +386,44 @@ codeParser :: String -> (String -> Maybe a) -> Parser a
...
+      --      but it's questionable whether this should even count as a /change/ of a user
+      --      profile).
+      --
+      -- There are some other things that SCIM can't do yet, like setting accent IDs, but they
+      -- are not essential, unlike e.g. passwords.
+    | ManagedBySCIM
+    deriving (Eq, Show, Bounded, Enum)
+
+instance FromJSON ManagedBy where
+    parseJSON = withText "ManagedBy" $ \case
```
ChrisPenner: Is it reasonable to pull out a helper:

```haskell
typeName :: forall a. Typeable a => String
typeName = show $ typeRep @a
```

And use something like 

```haskell
withText (typeName @ManagedBy)
```
So we get a compile-time error here if the type is renamed?

I suppose this should all be covered by tests anyways; probably not worth it 🤷‍♂️

fisx: I don't think `typeName` will compile without a `Proxy` argument, but then it would work, and I am in favor of the idea.

libs/brig-types/src/Brig/Types/User.hs
```diff
@@ -246,6 +252,7 @@ data NewUser = NewUser
     , newUserLocale         :: !(Maybe Locale)
     , newUserPassword       :: !(Maybe PlainTextPassword)
     , newUserExpiresIn      :: !(Maybe ExpiresIn)
+    , newUserManagedBy      :: !(Maybe ManagedBy)
```
ChrisPenner: Just curious; do we tend to make all data strict? I presume it makes sense in most (if not all) cases; has anyone considered using a global `StrictData` pragma?

libs/brig-types/src/Brig/Types/Common.hs
ChrisPenner: @fisx We would need `AllowAmbiguousTypes`; not sure if the team has opinions on that particular extension or not 🤷‍♂️

libs/brig-types/src/Brig/Types/User.hs
neongreen: > do we tend to make all data strict

Yes.

> has anyone considered using a global `StrictData` pragma?

I recall considering it when I first got here. Then I got scared of enabling it and decided not to spend more time on figuring out whether it's safe or not. I don't have any specific concerns in mind, though.

libs/brig-types/src/Brig/Types/Common.hs
ChrisPenner: @neongreen Or am I misunderstanding the purpose of this string; does it need to match the type name exactly or is it actually just for error messages?

neongreen: > I don't think `typeName` will compile without a `Proxy` argument

It will if you enable AllowAmbiguousTypes (which you should)

libs/brig-types/src/Brig/Types/User.hs
ChrisPenner: @neongreen It would either speed things up; slow things down, or neither haha; tough to say which without trying it.

libs/brig-types/src/Brig/Types/Common.hs
neongreen: This said, the type name doesn't really matter. It only shows up in Aeson error messages and I'm not sure that deviating from this "standard" way of writing JSON instances is worth it.

libs/brig-types/src/Brig/Types/User.hs
neongreen: I'm just slightly worried that we might be relying on laziness in some place or other.

ChrisPenner: It's certainly possible; but if that's true the result will be either that the integration tests fail; or that we discover some place where we use laziness unexpectedly; If we do require it in some place we can explicitly denote it with `~` to make it lazy again;

Granted, this maybe just takes time that nobody has; just curious :)

neongreen:
@jschaul did you look at the commit that addresses "couldn't this be set already at the time a user in brig is created"? It's not small enough for me to comfortably say "yeah this code has been reviewed".

https://github.com/wireapp/wire-server/pull/602/commits/45448d62a6a81f7d7185deab9ab551bb2d605450

PR #606, opened 2019-02-01
Enable journaling locally and fix integration tests

neongreen:
The tests got broken when a commit was introduced where one of the tests used `addTeamMemberInternal` and `addTeamMemberInternal` but didn't follow them up with a corresponding `assertQueue`.

### DIFF
--- a/services/galley/galley.integration.yaml
+++ b/services/galley/galley.integration.yaml
@@ -30,14 +30,7 @@ settings:
 logLevel: Info
 logNetStrings: false
 
-### Journaling of team events
-#
-# If you want to use journaling but journal to a fake sqs service
-# such as "fake-sqs" or "localstack",
-# run e.g. `../../deploy/docker-ephemeral/run.sh`
-# and use a localhost endpoint
-#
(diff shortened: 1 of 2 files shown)


neongreen:
@fisx you should also take a look since (I think) you wrote that code.

neongreen:
(Not review, just take a look.)

jschaul:
To get more adequate CI output for pending PRs, I'll merge this PR.

fisx:
Alternatively, [galley integration test plumbing](https://github.com/wireapp/wire-server/blob/9fac5236cc5c24c8377487a6cdb1a430d2c4acc8/services/galley/test/integration/API.hs#L40-L45) could be changed to work like [in brig](https://github.com/wireapp/wire-server/blob/9fac5236cc5c24c8377487a6cdb1a430d2c4acc8/services/brig/test/integration/Util.hs#L82-L83), and cleanup would happen automatically.

But I just tried, and I feel it makes more sense to start a package `wire-test` that contains these things, so we can start working on de-duplicating some of the integration test code.

PR #609, opened 2019-02-05
[WIP] [needs to be redone] Limit AWS connections in Gundeck

neongreen:
Warning: untested. But can be reviewed already.

### DIFF
--- a/libs/extended/src/Network/HTTP/Client/Extended.hs
+++ b/libs/extended/src/Network/HTTP/Client/Extended.hs
@@ -1,0 +1,109 @@
+{-# LANGUAGE NoImplicitPrelude #-}
+{-# LANGUAGE DeriveGeneric #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+
+-- | A version of "Network.HTTP.Client" with extra utilities.
+module Network.HTTP.Client.Extended
+    ( module Network.HTTP.Client
+    -- * Connection limits
+    , addPerHostConnectionLimit
+    , PerHostConnectionLimitReached(..)
+    ) where
+
+import Imports
+import Control.Exception (throwIO)
+import Control.Lens
+import Data.ByteString.Short
+import Data.Hashable
+import Network.HTTP.Client
+import Network.HTTP.Client.Internal (Connection(..), ManagerSettings(..))
+
+import qualified STMContainers.Map as STMMap
+
+-- | Connection destination: just host and port.
+data Destination = Destination {-# UNPACK #-} !ShortByteString !Int
+    deriving (Eq, Show, Generic)
+
+instance Hashable Destination
+
+-- | Add a per-host hard connection limit. When there are more than N open connections to a
+-- host, any attempt to establish a new connection to that host will immediately fail with
+-- 'PerHostConnectionLimitReached'.
+--
+-- Existing limits in 'Manager' are more fine-grained: for instance, those limits apply to raw
+-- and TLS connections separately. The limit set by 'addHardConnectionLimit' applies to all
+-- connections to the host/port, regardless of their type.
+--
+-- Hosts are defined "whatever host is set in the 'Request'" so if there are several servers
+-- behind a load balancer, we're still going to count them as a single host.
+--
+-- /Note 2019-02-05:/ @http-client@ Github issue requesting this functionality:
+-- <https://github.com/snoyberg/http-client/issues/307>. We might try to upstream our
+-- implementation.
+addPerHostConnectionLimit :: MonadIO m => Int -> ManagerSettings -> m ManagerSettings
+addPerHostConnectionLimit limit managerSettings = liftIO $ do
+    -- Create a map counting active connections per host.
+    connCount :: STMMap.Map Destination Int <- STMMap.newIO
+
+    -- When a connection is created, increment a counter (and possibly throw an exception).
+    let onCreate :: Destination -> IO ()
+        onCreate dest@(Destination host_ port_) = do
+            success <- atomically $ do
+                current <- fromMaybe 0 <$> STMMap.lookup dest connCount
+                when (current < limit) $ STMMap.insert (current + 1) dest connCount
+                pure (current < limit)
+            unless success $ throwIO $
+                PerHostConnectionLimitReached (fromShort host_) port_
+
+    -- When a connection is destroyed, which is something 'Manager' should always do,
+    -- decrement the counter.
+    let onDestroy :: Destination -> IO ()
+        onDestroy dest = atomically $ do
+            current <- fromMaybe 0 <$> STMMap.lookup dest connCount
+            if current <= 1
+                then STMMap.delete dest connCount
+                else STMMap.insert (current - 1) dest connCount
+
+    -- Modify all connection-creating functions in the 'Manager' to perform 'onCreate' and
+    -- 'onDestroy'. Technically 'connectionClose' can be performed on an already closed
+    -- connection and we have to handle that gracefully (at least according to the
+    -- 'Connection' docs), but 'Manager' seems to close connections exactly once.
+    let managerRawConnection' = do
+            create <- managerRawConnection managerSettings
+            pure $ \addr host_ port_ -> do
+                let dest = Destination (fromString host_) port_
+                conn <- onCreate dest >> create addr host_ port_
+                pure (conn & connectionCloseL %~ (onDestroy dest >>))
+
+        managerTlsConnection' = do
+            create <- managerTlsConnection managerSettings
+            pure $ \addr host_ port_ -> do
+                let dest = Destination (fromString host_) port_
+                conn <- onCreate dest >> create addr host_ port_
+                pure (conn & connectionCloseL %~ (onDestroy dest >>))
+
+        managerTlsProxyConnection' = do
+            create <- managerTlsProxyConnection managerSettings
+            pure $ \connstr check server ha host_ port_ -> do
+                let dest = Destination (fromString host_) port_
+                conn <- onCreate dest >> create connstr check server ha host_ port_
+                pure (conn & connectionCloseL %~ (onDestroy dest >>))
+
+    pure managerSettings
+        { managerRawConnection = managerRawConnection'
+        , managerTlsConnection = managerTlsConnection'
+        , managerTlsProxyConnection = managerTlsProxyConnection'
+        }
+
+-- | A lens for 'connectionClose'.
+connectionCloseL :: Lens' Connection (IO ())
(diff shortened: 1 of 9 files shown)


libs/extended/src/Network/HTTP/Client/Extended.hs
```diff
@@ -0,0 +1,101 @@
...
+import qualified STMContainers.Map as STMMap
+
+-- | Connection destination: just host and port.
+data Destination = Destination {-# UNPACK #-} !ShortByteString !Int
+    deriving (Eq, Show, Generic)
+
+instance Hashable Destination
+
+-- | Add a per-host hard connection limit. When there are more than N open connections to a
+-- host, any new connection to that host will immediately fail with 'PerHostConnectionLimit'.
```
fisx: I assume this is per-DNS-name, so it may be a bunch of hosts behind a load balancer?  May be worth clarifying the comment.

What if it's more complicated, and we want to have a limit per all hosts satisfying a certain property?  (I assume that's not required for our use case, in that case we shouldn't implement it now.)

fisx: `PerHostConnectionLimitReached` instead of `PerHostConnectionLimit`?

services/gundeck/src/Gundeck/Options.hs
```diff
@@ -125,6 +126,11 @@ optsParser = Opts <$>
             <> value (AWSEndpoint "sns.eu-west-1.amazonaws.com" True 443)
             <> showDefault
             <> help "aws SNS endpoint")
+        <*>
+        (optional $ option auto $
+            long "aws-connection-limit"
+            <> metavar "SIZE"
+            <> help "maximum number of simultaneous connections to AWS")
```
fisx: `"default: no limit"`

services/gundeck/deb/etc/sv/gundeck/run
```diff
@@ -49,4 +55,6 @@ exec chpst -u $USER \
     --aws-arn-env=${KHAN_ENV?'unset'} \
     --aws-sqs-endpoint=${GUNDECK_AWS_SQS_ENDPOINT:-'https://sqs.eu-west-1.amazonaws.com'} \
     --aws-sns-endpoint=${GUNDECK_AWS_SNS_ENDPOINT:-'https://sns.eu-west-1.amazonaws.com'} \
+    --aws-sns-endpoint=${GUNDECK_AWS_SNS_ENDPOINT:-'https://sns.eu-west-1.amazonaws.com'} \
```
tiago-loureiro: One copy+paste too many?

libs/extended/src/Network/HTTP/Client/Extended.hs
```diff
@@ -0,0 +1,101 @@
...
+                conn <- onCreate dest >> create addr host_ port_
+                pure (conn & connectionCloseL %~ (onDestroy dest >>))
+
+        managerTlsConnection' = do
+            create <- managerTlsConnection managerSettings
+            pure $ \addr host_ port_ -> do
+                let dest = Destination (fromString host_) port_
+                conn <- onCreate dest >> create addr host_ port_
+                pure (conn & connectionCloseL %~ (onDestroy dest >>))
+
```
tiago-loureiro: Minor nitpick: this block can be refactored since it's exactly the same in `managerRawConnection'` and `managerTlsConnection'`
```
let dest = Destination (fromString host_) port_
conn <- onCreate dest >> create addr host_ port_
pure (conn & connectionCloseL %~ (onDestroy dest >>))
```

libs/extended/src/Network/HTTP/Client/Extended.hs
neongreen: If we need a more complicated rule, we can pass a function from host and port to some Hashable a.

neongreen:
> Perhaps this could be tested by

Sure, though I still don't know how to write an integration test for it (or whether we have to write one – but it'd be nice).

neongreen:
> perhaps explain what's needed to push all of this (or which part) upstream?

Nothing except that I don't think it's good enough for http-client. It's too harsh with regard to sudden connection spikes.

neongreen:
@tiago-loureiro the copy-and-paste thing has been fixed.

neongreen:
> 1. starting docker images + services on the develop branch, killing fake-sqs container, then sending a range of POST /push/v2 to gundeck -> this should lead to OOM (or high memory usage) on gundeck

I tried doing that in a variety of ways (killing fake-sqs, killing localstack, slowing down all connections), but couldn't reproduce high memory usage. Either connections fail immediately (when SQS is down), or something else timeouts first (when the whole network is slowed down).

If somebody else tries and succeeds, please tell me how you did it.

neongreen:
Me and @jschaul tested that with limit=20 Gundeck doesn't die when spammed with connections.

Unfortunately we failed to test that the Gundeck from `develop` does die.

services/gundeck/deb/etc/sv/gundeck/run
```diff
@@ -49,4 +55,5 @@ exec chpst -u $USER \
     --aws-arn-env=${KHAN_ENV?'unset'} \
     --aws-sqs-endpoint=${GUNDECK_AWS_SQS_ENDPOINT:-'https://sqs.eu-west-1.amazonaws.com'} \
     --aws-sns-endpoint=${GUNDECK_AWS_SNS_ENDPOINT:-'https://sns.eu-west-1.amazonaws.com'} \
+    ${GUNDECK_AWS_CONNECTION_LIMIT}
```
jschaul:
```suggestion
    ${GUNDECK_AWS_CONNECTION_LIMIT} \
```

neongreen:
Added a default. Also see: https://github.com/wireapp/wire-server-deploy/pull/22

tiago-loureiro:
Hmmm, I think we should give this another go wrt destroying connections in some cases... here's how/what I tested (a bit manual, I know...) but bear with me:

Use [this commit](https://github.com/wireapp/wire-server/commit/19c3d47081e441ac6c2cd8a05998cb98794755a4) which adds some more debug info and then prepare a few terminals

Terminal 1 (gundeck)
`LOG_LEVEL=Info ../../dist/gundeck -c gundeck.integration.yaml`
^-- just keep an eye on "True/False" and "Destroying"

Terminal 2 (your SQS replacement)
`while true; do sudo lsof -n -i | grep 4575; sleep 0.1; done`
^-- use this to monitor established connections

Terminal 3 - kill fake sqs using `docker kill` (check the correct pid with docker ps) and then; if you want start netcat with `nc -lk 4575` if you want but you can also just allow it to fail early.

Terminal 4 -
```
#!/usr/bin/env bash

ulimit -n 8192

for i in $(seq 1 "$1")
do
   sleep 0.01;
   curl -XPOST localhost:8086/push/tokens -H'Z-User: 0d76baf4-d8fb-4867-9ccf-62100fffc079' -H'Z-Connection: 0' -H'Content-type: application/json' -d'{"transport":"GCM","app":"test","token":"12312312312321","client":"1231231231231"}' &
done
```
^-- Just run it with `./script.sh 10`

Gundeck does not manage to recover from `PerHostConnectionLimitReached`; in fact, if you don't start `netcat`, you won't see a single `Destroying` (fails too early?). If you start `netcat` (which does not really speak HTTP) then things get even more interesting...

The final test that I ran was the following; change this on the config file:
```
-  snsEndpoint: http://localhost:4575 # https://sns.eu-west-1.amazonaws.com
+  snsEndpoint: http://httpstat.us/500?sleep=1 # http://localhost:4575 # https://sns.eu-west-1.amazonaws.com
```

This just talks to a server that returns 500 (which is what SNS does when we struggle); in such cases, we seem to be doing the right thing, i.e., the connections remain open but we still accept new requests as long as we have connections available in the pool.

Sorry for not having integration tests but this is rather tricky :)

neongreen:
> you won't see a single Destroying

Yeah, apparently the counter doesn't get decremented properly. I'm debugging this.

neongreen:
Yeah, it just doesn't run `kpDestroy` on connection timeout, for some reason. Or rather, it does but not always.

To reproduce, `nc -l 12777` and then start this and press Enter several times:

```haskell
import Imports

-- qualified imports because I wanted to see how it feels to code like that
import qualified Network.HTTP.Client.Extended as Http
import qualified Network.HTTP.Client.TLS as Http.Tls
import qualified Network.HTTP.Conduit as Http.Conduit
import qualified Conduit as Conduit
import qualified Data.Conduit.Binary as Conduit.Binary
import qualified Data.Aeson as Aeson
import qualified UnliftIO.Resource as Resource

main :: IO ()
main = do
    manager <- Http.newManager =<<
        Http.addPerHostConnectionLimit 10
             Http.Tls.tlsManagerSettings
                 { Http.managerResponseTimeout = Http.responseTimeoutMicro 15000000 }
    Resource.runResourceT $ for_ [(1::Int) ..] $ \i -> do
        _ <- getLine
        initialRequest <- Http.parseRequest "POST http://localhost:12777"
        let request = initialRequest
                { Http.requestBody = Http.RequestBodyLBS . Aeson.encode $ i
                }
        forkIO $ do
            response <- Resource.liftResourceT (Http.Conduit.http request manager)
            Resource.liftResourceT $
                print =<<
                (Http.responseBody response `Conduit.connect` Conduit.Binary.sinkLbs)
```

neongreen:
See https://github.com/snoyberg/http-client/issues/394

neongreen:
> Note that http-client relies on \[weak\] finalizers to deallocate connections, so you shouldn't expect immediate deallocation. Try to insert `performMajorGC` after each getLine, and you'll see more consistent behavior.
>
> We can make resource deallocation more reliable (probably not 100% reliable though), but it will be a feature request then \[...\]

neongreen:
@ChrisPenner I'm thinking about tackling this issue in a different way: just increment a counter (in wire-server code) whenever we make a request to SNS, and decrement it when the connection succeeds or the thread dies.

Do you know a reliable way to do something when a thread dies (no matter what happens to this thread)? Is `finally` enough or not really?

neongreen:
I don't work at Wire anymore, so somebody else will have to finish this one.

services/gundeck/src/Gundeck/Env.hs
```diff
@@ -77,7 +77,7 @@ createEnv m o = do
             . C.setResponseTimeout 10
             . C.setProtocolVersion C.V3
             $ C.defSettings
-    a <- Aws.mkEnv l o n
```
fisx: Then `n` probably shouldn't be created earlier?  Or, it if is used somewhere else, we should use the one created in `mkEnv`?

fisx:
Obsoleted by #869.

PR #611, opened 2019-02-06
Add module descriptions for all Spar modules

neongreen:
I'm very often bitten by "should X go to this module or to some other module?", so I thought spending a bit of time on adding module descriptions would be a good idea.

I would also love it if CI barked at PRs that introduce modules without documentation. Thoughts?

### DIFF
--- a/services/spar/src/Spar/API/Util.hs
+++ b/services/spar/src/Spar/API/Util.hs
@@ -6,6 +6,7 @@
 {-# LANGUAGE ScopedTypeVariables   #-}
 {-# LANGUAGE PolyKinds             #-}
 
+-- | Swagger utilities.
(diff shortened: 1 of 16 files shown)


neongreen:
(We can check that on CI by running Haddock.)

services/spar/src/Spar/API/Util.hs
```diff
@@ -6,6 +6,7 @@
 {-# LANGUAGE ScopedTypeVariables   #-}
 {-# LANGUAGE PolyKinds             #-}
 
+-- | Swagger utilities.
```
fisx: I guess that means a better place for this module would be `Spar.API.Swagger.Util`.  Oh well.

PR #613, opened 2019-02-07
Slightly update SCIM comments in regard to names, etc

neongreen:
No logic changes.

### DIFF
--- a/services/spar/src/Spar/Scim.hs
+++ b/services/spar/src/Spar/Scim.hs
@@ -167,11 +167,13 @@ validateScimUser ScimTokenInfo{stiIdP} user = do
 -- | Map the SCIM data on the spar and brig schemata, and throw errors if the SCIM data does
 -- not comply with the standard / our constraints. See also: 'ValidScimUser'.
 --
+-- Checks like "is this handle claimed already?" are not performed. Only schema checks.
+--
 -- __Mapped fields:__
 --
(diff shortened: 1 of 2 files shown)


services/spar/src/Spar/Scim.hs
```diff
@@ -167,34 +169,26 @@ validateScimUser ScimTokenInfo{stiIdP} user = do
 --
 -- __Mapped fields:__
 --
---   * @userName@ is mapped to our 'userHandle'. If there is no handle, we use 'userId',
---     because having some unique @userName@ is a SCIM requirement.
---
---   * @name@ is left empty and is never stored, even when it's sent to us via SCIM.
+--   * @userName@ is mapped to our 'userHandle'.
```
jschaul: How is it mapped to userHandle, userHandle being globally unique? Perhaps explain or link to an explanation?

neongreen: This function doesn't do such checks. It only checks the schema.

Checks are done elsewhere when the user is created / updated.

I have updated the comment.

fisx: If a handle is taken, the create request from the scim client will be answered with an error status?  What does the error look like?  Is that something that could use more haddocks?  (Not sure about any of this.)

neongreen: > Is that something that could use more haddocks?

Definitely.

PR #617, opened 2019-02-11
Allow emails as subject IDs

neongreen:


services/spar/src/Spar/Scim.hs
```diff
@@ -194,37 +195,48 @@ validateScimUser'
...
+    -- NB: We assume that checking that the user does _not_ exist has already been done before;
+    -- the hscim library check does a 'get' before a 'create'.
     pure $ ValidScimUser user uref handl mbName
 
+  where
+    -- Validate a subject ID (@externalId@). Right now we only support emails, but other
+    -- subject ID formats should be supported in the future.
+    validateSubject :: Text -> m SAML.NameID
+    validateSubject txt = do
+        unless (isJust (parseEmail txt)) $
```
fisx: I think I remember now that saml2-web-sso does not care about the `NameID` format, it just takes whatever it gets.  This would give us a very simple and possibly good-enough heuristics to support both email and unspecified: if it parses as an email, use that, otherwise, use unspecified.

I think other formats are pretty obscure.

services/spar/test-integration/Util/Core.hs
```diff
@@ -288,13 +289,20 @@ nextWireId = Id <$> liftIO UUID.nextRandom
...
+nextSubject :: MonadIO m => m NameID
+nextSubject = do
+  email <- Brig.fromEmail <$> randomEmail
+  pure $ SAML.NameID (SAML.UNameIDEmail email) Nothing Nothing Nothing
+
 nextUserRef :: MonadIO m => m SAML.UserRef
 nextUserRef = liftIO $ do
-  (UUID.toText -> tenant) <- UUID.nextRandom
-  (UUID.toText -> subject) <- UUID.nextRandom
+  tenant <- UUID.toText <$> UUID.nextRandom
```
fisx: I'm curious: is there a reason i'm missing why this is better?  (I don't care either way.)

services/spar/test-integration/Util/Scim.hs
```diff
@@ -92,7 +92,7 @@ randomScimUser = do
     pure $ Scim.User.empty
         { Scim.User.userName     = "scimuser_" <> suffix
         , Scim.User.displayName  = Just ("Scim User #" <> suffix)
-        , Scim.User.externalId   = Just ("scimuser_extid_" <> suffix)
+        , Scim.User.externalId   = Just ("scimuser_extid_" <> suffix <> "@wire.com")
```
fisx: `@example.com`.

services/spar/test-integration/Util/Scim.hs
```diff
@@ -396,7 +396,9 @@ instance IsUser Scim.User.User where
     maybeHandle = Just . Just . Handle . Scim.User.userName
     maybeName = Just . fmap Name . Scim.User.displayName
     maybeTenant = const Nothing
-    maybeSubject = Just . fmap SAML.opaqueNameID . Scim.User.externalId
+    maybeSubject =
+        let emailNameID s = SAML.NameID (SAML.UNameIDEmail s) Nothing Nothing Nothing
+        in Just . fmap emailNameID . Scim.User.externalId
```
fisx: Doesn't this give you 2 `Just`?  one from `externalId`, one Just?  I know it was like that before, so it's probably correct.  Haven't checked the context.  Please ignore.

services/spar/test/Test/Spar/ScimSpec.hs
```diff
@@ -33,7 +33,7 @@ spec = describe "toScimStoredUser'" $ do
     let usr :: Scim.User
         usr = Scim.User
           { Scim.userName = "02b35298-088f-11e9-b4a4-478635dd0d2b"
-          , Scim.externalId = Just "c1704a48-0a1e-11e9-9186-9b185fe892e8"
+          , Scim.externalId = Just "test@wire.com"
```
fisx: `@example.com`.

fisx:
I haven't checked that either, but I think the test case would not have failed if the test case would erase the weak version number before comparing (rather than blindly copying it from the message of the failing test into the test itself).

services/spar/src/Spar/Scim.hs
neongreen: > if it parses as an email, use that, otherwise, use unspecified.

Makes sense. I'm embarrassed that I didn't think of this.

services/spar/test-integration/Util/Core.hs
neongreen: * The flow is broken (right-left-middle).

* It's significantly less common. I grepped and we _never_ do that in our codebase, other than in the two places above. (I mean, we use view patterns for sure, but in all those cases the pattern is more complicated, e.g. a nested tuple).

services/spar/test-integration/Util/Scim.hs
neongreen: True, but it's supposed to be like that. I'll add comments and rewrite the class a bit.

neongreen:
> I haven't checked that either, but I think the test case would not have failed if the test case would erase the weak version number before comparing

What do you mean / what is this about?

fisx:
> > I haven't checked that either, but I think the test case would not have failed if the test case would erase the weak version number before comparing
> 
> What do you mean / what is this about?

This is about the CI.  There is a failure where the version hashes don't match, and I think that's because some other part of the values that are compared has changed.

But I may be completely off track, only saw the error and didn't read the test that failed.

services/spar/test-integration/Util/Scim.hs
```diff
@@ -85,17 +85,32 @@ registerScimToken teamid midpid = do
...
     emails <- getRandomR (0, 3) >>= \n -> replicateM n randomScimEmail
     phones <- getRandomR (0, 3) >>= \n -> replicateM n randomScimPhone
-    pure $ Scim.User.empty
-        { Scim.User.userName     = "scimuser_" <> suffix
-        , Scim.User.displayName  = Just ("Scim User #" <> suffix)
-        , Scim.User.externalId   = Just ("scimuser_extid_" <> suffix <> "@example.com")
-        , Scim.User.emails       = emails
-        , Scim.User.phoneNumbers = phones
-        }
+    (externalId, subj) <- getRandomR (0, 1::Int) <&> \case
```
fisx: add a comment here:

```
-- related, but non-trivial to re-use here: `nextSubject`
```

(and perhaps link back from there)

services/spar/test-integration/Util/Scim.hs
```diff
@@ -381,6 +396,9 @@ class IsUser u where
     maybeName :: Maybe (u -> Maybe Name)
     maybeTenant :: Maybe (u -> Maybe SAML.Issuer)
     maybeSubject :: Maybe (u -> Maybe SAML.NameID)
+    -- | Not all types have a 'SAML.NameID' but some of them have a subject ID anyway, just
+    -- not in a structured form.
+    maybeSubjectRaw :: Maybe (u -> Maybe Text)
```
fisx: would be nice to give two example types here, like `UserRef` and `ScimUser` (or whathever it's called)?

neongreen:
Done and done. Waiting for integration.

PR #618, opened 2019-02-12
Change the bot port in integration tests to something less common

neongreen:
Ideally warp-tls should crash when the port is taken, but it doesn't do that for some reason. I'll wait until I get an answer to https://github.com/yesodweb/wai/issues/730 and then I might open another PR to further improve the situation (or just check for the port in `integration.sh`).


neongreen:
We don't seem to use that part (we always pass the config).

jschaul:
Ah true. We could probably remove the environment-variable-loading part of the tests as they're not used.

PR #621, opened 2019-02-13
Test that the location in SCIM user resources is correct

neongreen:


services/spar/src/Spar/Options.hs
```diff
@@ -63,7 +63,7 @@ deriveOpts raw = do
               ("sso" : path') -> compile path'
               path'           -> compile path'
 
-            compile path = "/" <> SBS.intercalate "/" (reverse ("scim" : path))
+            compile path = "/" <> SBS.intercalate "/" (reverse ("v2" : "scim" : path))
```
fisx: Are you sure you want to reverse `path` here?  (I know, not introduced in this change, but I'm still wondering...)

Also add a test where `path` has at least two elements and is not palindromatic?

neongreen: I don't want to reverse the path.

The reason the path is reversed is that `derive` wants to use pattern-matching on the last element of the path. It's originally your code IIRC, but if you don't like it anymore I can rewrite it with `init` and `lastMay`.

neongreen: > Also add a test where `path` has at least two elements and is not palindromatic?

`scim/v2/Users` satisfies those requirements 🤔

fisx: sorry.  please ignore.

PR #624, opened 2019-02-14
Bump handlebars.js to 4.1.0.

fisx:
https://www.npmjs.com/advisories/755

thanks @ffflorian!

(not tested.  is there a way to test this before staging?)

### DIFF
--- a/services/nginz/zwagger-ui/tab.html
+++ b/services/nginz/zwagger-ui/tab.html
@@ -17,7 +17,7 @@
   <script src='swagger-ui/lib/jquery.slideto.min.js' type='text/javascript'></script>
   <script src='swagger-ui/lib/jquery.wiggle.min.js' type='text/javascript'></script>
   <script src='swagger-ui/lib/jquery.ba-bbq.min.js' type='text/javascript'></script>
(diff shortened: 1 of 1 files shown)


neongreen:
I don't see how this would work. Where would `handlebars-4.1.0.js` come from?

neongreen:
https://github.com/wireapp/wire-server/blob/develop/services/nginz/Makefile#L102-L110

Swagger files are downloaded here.

The minor version of Swagger would probably need to be updated (if they are still doing security patches for the version we're using).

fisx:
good point.  i guess we could fetch it from [here](https://github.com/wycats/handlebars.js/archive/v4.1.0.tar.gz)?

not sure about all the stuff that could go wrong there, though.

fisx:
Given we're delivering this [in production](https://prod-nginz-https.wire.com/swagger-ui/), and given I don't really know how this works, I don't feel confident enough to decide the issue does not affect us.

If anybody can, feel free to close.

fisx:
(swagger-ui doesn't have a fix yet, i just checked.)

fisx:
the bug does not affect us here.

PR #625, opened 2019-02-14
Fix: do not have newlines in log messages.

fisx:


neongreen:
```
  test-integration/Test/Spar/APISpec.hs:342:70: 
  1) Spar.API, binding existing users to sso identities, GET /sso-initiate-bind/:idp, known IdP, running session with sso user, responds with 200 and a bind cookie
       uncaught exception: ErrorCall
       no wire cookie
         error, called at test-integration/Util/Core.hs:589:27 in main:Util.Core
         loginSsoUserFirstTime, called at test-integration/Test/Spar/APISpec.hs:342:70 in main:Test.Spar.APISpec
```

services/gundeck/src/Gundeck/Push/Websocket.hs
```diff
@@ -99,7 +99,8 @@ logBadCannons (uri, (err, prcs)) = do
     forM_ prcs $ \prc ->
         Log.warn $ logPresence prc
             ~~ Log.field "created_at" (ms $ createdAt prc)
-            ~~ Log.msg (val "WebSocket presence unreachable: " +++ (toByteString $ show (uri, err, resource prc)))
+            ~~ Log.msg (val "WebSocket presence unreachable: " +++
+                        show (uri, resource prc, show err))
```
neongreen: Perhaps remove the duplicate `show`?

neongreen: Also it might be better to report the URI and resource via `Log.field`.

neongreen: Oh wait, the duplicate `show` is actually what makes things work. There should be a comment about it, then.

fisx: Ok, I will elaborate on this in a different PR.

PR #633, opened 2019-02-15
Downgrade metrics output

fisx:
The metrics dashboards expect the capture variables from the wai routes in the metrics output, so here you go.


tiago-loureiro:
LGTM, please merge once tests succeed.

PR #636, opened 2019-02-18
SAML input sanitization

fisx:
Depends on https://github.com/wireapp/saml2-web-sso/pull/47

- [x] re-align with saml2-web-sso changes;
- [x] change stack.yaml to point to saml2--web-sso master.

### DIFF
--- a/services/spar/src/Spar/Data/Instances.hs
+++ b/services/spar/src/Spar/Data/Instances.hs
@@ -30,7 +30,7 @@ import qualified Web.Scim.Class.User as Scim
 
 instance Cql SAML.XmlText where
     ctype = Tagged TextColumn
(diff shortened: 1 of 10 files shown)


services/spar/src/Spar/Data/Instances.hs
```diff
@@ -28,6 +28,13 @@ import qualified SAML2.WebSSO as SAML
 import qualified Web.Scim.Class.User as Scim
 
 
+instance Cql SAML.XmlText where
+    ctype = Tagged TextColumn
+    toCql = CqlText . SAML.escapeXmlText
```
neongreen: We don't need to escape it when writing to Cassandra.

PR #642, opened 2019-02-21
logging vs. newlines

fisx:
tasks:

- [x] do not log `\n` in spar (with test case)
- [x] get https://gitlab.com/twittner/tinylog/merge_requests/3 merged
- [x] ~~get https://gitlab.com/twittner/tinylog/merge_requests/4 merged, or move its contents to our `extended` package.~~ bump tinylog dep to https://gitlab.com/twittner/tinylog/merge_requests/4
- [x] update tinylog dependency in `stack.yaml`
- [x] do not log `\n` in any service
- [x] double-check behavior hasn't changed (except for the newline filtering).


fisx:
see https://gitlab.com/twittner/tinylog/merge_requests/3

services/spar/src/Spar/Run.hs
```diff
@@ -74,7 +76,26 @@ mkLogger opts = Log.new $ Log.defSettings
...
+  where
+    rndrRaw :: Log.Renderer
+    rndrRaw = if logNetStrings opts
+      then Log.renderNetstr
+      else Log.renderDefault
+
+-- | Many logging processors handle newlines poorly.  Instead of showing discipline and not
+-- injecting any newlines into your log messages and making sure nobody else does, you can
+-- choose to call `noNewlines` on your renderer that replaces all consecutive whitespace in
+-- the output by a single ' '.
```
fisx: either the docs or the implementation are lying.  what we want to happen is this:

```
"\t  \n" -> " "
```

perhaps call it `canonicalizeWhitespace` and implement it as `LBS.unwords . LBS.words`?

fisx:
Cargohold has no option for netstr format log messages or debug level.  I can change this in a separate PR if at least one person is in favor, but it seems like this hasn't been a problem so far?

tiago-loureiro:
> Cargohold has no option for netstr format log messages or debug level. I can change this in a separate PR if at least one person is in favor, but it seems like this hasn't been a problem so far?

Both `cargohold` and `proxy` need to be migrated to use a config file only; fine if it's done separately, yes.

stack.yaml
```diff
@@ -41,6 +41,8 @@ extra-deps:
   commit: c03d17d656ac467350c983d5f844c199e5daceea    # master (Feb 21, 2019)
 - git: https://github.com/wireapp/hscim
   commit: 42f6018812bf0f04741231b67b1f5e790ce0d489    # master (Feb 25, 2019)
+- git: https://gitlab.com/fisx/tinylog
+  commit: ee39f436247e0c7ecd64ff7c1b28f3378179abfe    # https://gitlab.com/twittner/tinylog/merge_requests/4
```
tiago-loureiro: Guess this can be changed now?

fisx: it's an independent change and (unless we make the new config fields optional, i'd have to check what the other services do there), it complicates the PR because it requires changes in our ops repos.

why not do it separately?

tiago-loureiro: I guess I misunderstood this comment

>  get https://gitlab.com/twittner/tinylog/merge_requests/4 merged, or move its contents to our extended package. bump tinylog dep to https://gitlab.com/twittner/tinylog/merge_requests/4

And reg.:

> requires changes in our ops repos.

What does that mean?

fisx: tinylog merge request 4 *may* be going in soon, but it may take a while, or never happen.  i didn't want to wait for that; and leaving the comment in stack.yaml means that it'll be easy to follow up on changes upstream, should they ever happen.  then we can either pull in the changes that don't make it upstream, or just bump the dependency again to something more official.

fisx: if you meant "change cargohold to support config of tinylog like the others": *that* would require to change other repos in order for ci / deployment to work.  i think.  unless it's all optional config fields.

fisx:
@ChrisPenner [only if this interests you] CI now seems to have difficulty signalling success back to github?

neongreen:
> Cargohold has no option for netstr format log messages or debug level. I can change this in a separate PR if at least one person is in favor

I'm in the process of changing it.

fisx:
> > Cargohold has no option for netstr format log messages or debug level. I can change this in a separate PR if at least one person is in favor
> 
> I'm in the process of changing it.

wanna review this PR first?  :)

neongreen:
[scrolls through the PR]

Apparently yes.

services/brig/src/Brig/App.hs
```diff
@@ -150,11 +150,7 @@ data Env = Env
 makeLenses ''Env
 
 mkLogger :: Opts -> IO Logger
-mkLogger opts = Log.new $ Log.defSettings
-  & Log.setLogLevel (Opt.logLevel opts)
-  & Log.setOutput Log.StdOut
-  & Log.setFormat Nothing
-  & Log.setNetStrings (Opt.logNetStrings opts)
+mkLogger opts = Log.new $ Log.simpleDefSettings (Opt.logLevel opts) (Opt.logNetStrings opts)
```
neongreen: What about also inlining `mkLogger` here and in Cannon?

services/gundeck/src/Gundeck/Env.hs
```diff
@@ -40,16 +39,9 @@ makeLenses ''Env
...
-mkLogger opts = Logger.new $ Logger.defSettings
-    & Logger.setLogLevel (opts ^. optLogLevel)
-    & Logger.setOutput Logger.StdOut
-    & Logger.setFormat Nothing
-    & Logger.setNetStrings (opts ^. optLogNetStrings)
-
 createEnv :: Metrics -> Opts -> IO Env
 createEnv m o = do
-    l <- mkLogger o
+    l <- Logger.new $ Logger.simpleDefSettings (o ^. optLogLevel) (o ^. optLogNetStrings)
```
neongreen: Or un-inlining it here.

services/spar/src/Spar/Run.hs
```diff
@@ -70,11 +70,10 @@ initCassandra opts lgr = do
 -- logger
 
 mkLogger :: Opts -> IO Logger
-mkLogger opts = Log.new $ Log.defSettings
-  & Log.setLogLevel (toLevel $ saml opts ^. SAML.cfgLogLevel)
-  & Log.setOutput Log.StdOut
-  & Log.setFormat Nothing
-  & Log.setNetStrings (logNetStrings opts)
+mkLogger opts = Log.new $ Log.simpleDefSettings level netstr
```
neongreen: The resulting logger used to have `format = Nothing` and now it will have `format = iso8601UTC`.

I have no idea whether it will break anything or not.

services/spar/test-integration/Test/LoggingSpec.hs
```diff
@@ -0,0 +1,19 @@
+module Test.LoggingSpec (spec) where
+
+import Imports
+import Control.Lens
+import Data.String.Conversions (ST)
```
neongreen: `Text` is already available from `Imports`.

services/spar/test-integration/Test/LoggingSpec.hs
```diff
@@ -0,0 +1,19 @@
...
+import Spar.App
+import System.Logger as Log
+import System.IO.Silently (capture)
+import Util
+
+
+spec :: HasCallStack => SpecWith TestEnv
+spec = describe "logging" $ do
+  it "does not log newlines" $ do
```
neongreen: Would be nice to leave a comment explaining why newlines are not logged (it's not obvious that `simpleDefSettings` does that by default).

services/spar/src/Spar/Run.hs
fisx: as a rule, you need to look at my git commit messages.  :)

neongreen: Oh, I never do because I don't expect anyone to put anything there. I'll keep in mind that there might be hints in commit messages.

neongreen: > the date format defaults to the value in `defSettings` if `Nothing`

Where does that happen? In https://gitlab.com/twittner/tinylog/blob/develop/src/System/Logger.hs#L126 I see that `Nothing` and `Just` are treated differently. `setFormat` doesn't do anything clever either: https://gitlab.com/twittner/tinylog/blob/develop/src/System/Logger/Settings.hs#L73.

fisx: https://gitlab.com/fisx/tinylog/blob/settings-auxiliaries/src/System/Logger.hs#L194

it's a bit hard to find, i had to grep the entire package for `iso8601UTC`.

fisx: (do you have a suggestion how to document this here?  i thought it's just relevant for the change, but not for the resulting code.)

ChrisPenner:
@fisx Good to know about the concourse problem 👍

services/spar/test-integration/Test/LoggingSpec.hs
fisx: https://gitlab.com/fisx/tinylog/blob/settings-auxiliaries/src/System/Logger/Settings.hs#L212-219

is that enough?  or does it have to be in this test?

services/gundeck/src/Gundeck/Env.hs
fisx: I chose "inline".

services/spar/test-integration/Test/LoggingSpec.hs
neongreen: I'd rather have it in the test. My reasoning is that if `simpleDefSettings` changes and this test starts to fail, a comment that immediately points to `simpleDefSettings` would be helpful. Otherwise it's kinda hard to figure out where exactly the newlines were _supposed_ to be stripped (in the tinylog library? in our settings? in some monad we have around? or perhaps we intercept stdout and do something to it?)

services/spar/src/Spar/Run.hs
fisx: You pointed out correctly that `renderer` is not the only place where this is used.

I added a test case that establishes that for `iso8601UTC`, there is no difference between the two cases in https://gitlab.com/twittner/tinylog/blob/develop/src/System/Logger.hs#L126

I also [opened this issue](https://gitlab.com/twittner/tinylog/issues/2).

services/spar/test-integration/Test/LoggingSpec.hs
fisx: done.  (if you still want something else: would you like to make a commit yourself?)

services/spar/src/Spar/Run.hs
neongreen: I pulled your Gitlab branch and there is a difference:

```
> logger <- new (simpleDefSettings Debug False)
> fatal logger (msg "hi")
2019-03-05T23:04:24Z, F, hi

> logger2 <- new (setFormat Nothing $ simpleDefSettings Debug False)
> fatal logger2 (msg "hi")
F, hi
```

ChrisPenner:
@fisx Concourse seems to have completed successfully; was there a particular part you were referring to as not finishing?

fisx:
> @fisx Concourse seems to have completed successfully; was there a particular part you were referring to as not finishing?

yep, everything works again.  perhaps it was just failing for lack of cpu/ram/...?  anyway, i'll ping you if it happens again.  thanks!

services/spar/src/Spar/Run.hs
fisx: hm, appearently i didn't add the test to Spec.hs :-(

services/cannon/src/Cannon/API.hs
```diff
@@ -39,18 +38,11 @@ import qualified Network.WebSockets          as Ws
...
-    & Logger.setOutput Logger.StdOut
-    & Logger.setFormat Nothing
-    & Logger.setNetStrings (o^.logNetStrings)
-
 run :: Opts -> IO ()
 run o = do
     ext <- loadExternal
     m <- metrics
-    g <- mkLogger o
+    g <- Logger.new $ simpleDefSettings (o^.logLevel) (o^.logNetStrings)
```
ChrisPenner: In light of recent talks about imports; perhaps we can use `Log.simpleDefSettings` here and `Log.new` to be consistent with other files and usages in the PR?

stack.yaml
```diff
@@ -41,6 +41,8 @@ extra-deps:
   commit: c03d17d656ac467350c983d5f844c199e5daceea    # master (Feb 21, 2019)
 - git: https://github.com/wireapp/hscim
   commit: 42f6018812bf0f04741231b67b1f5e790ce0d489    # master (Feb 25, 2019)
+- git: https://gitlab.com/fisx/tinylog
+  commit: ef3a8f12233adce75f05f0c556ec69ea5ab9a008    # https://gitlab.com/twittner/tinylog/merge_requests/4
```
ChrisPenner: Reminder to switch this to master or whatever once it's merged :)

neongreen:
Maybe shouldn't be approved if it's still WIP

services/cannon/src/Cannon/API.hs
fisx: i made the mistake of starting with this, so now this PR will be very, very noisy...  But it's a fair point.  :)

fisx:
I *think* this should be pretty close.  I left a few `TODOs` in there that I want to handle separately soon; I will finish the logger import refactoring elsewhere; and I should cleanup the tinylog upstream dependency a little, but tinylog maintainenance is not real-time, so, separately.

I will take another look if I've missed anything, but it's ready for review.  @neongreen if you want to speed things up...  :-)

fisx:
@ChrisPenner do you have any idea why the linter still can't find the .linting directory?  It's there in my local commit.

fisx:

PR #645, opened 2019-02-22
Rich profiles

neongreen:
## Implementation

* [x] Types
* [x] Public GET rich info endpoint
* [x] Private PUT rich info endpoint
* [x] Support extra fields in hscim
* [x] Add a rich info limit
* [x] Allow setting rich info via SCIM
* [x] Prohibit duplicate fields
* [x] Delete fields with values ""

## Documentation

* [x] Reference documentation

## Tests

* [x] Roundtrip for types
* [x] There is a default empty rich info
* [x] Team members can see other team members' rich info
* [x] Non-team-members don't have rich info
* [x] Guests can't see rich info
* [x] Members of other teams can't see rich info
* [x] Writing a smaller list erases fields that are not present in it
* [x] Empty fields ("") are deleted
* [x] SCIM is actually doing the setting
* [x] SCIM should also be doing the querying
* [x] Duplicate fields are forbidden
* [x] There is a rich info limit

## Questions

* [x] Should `userDeleted` users be handled separately? Tiago's answer: no, as long as DB invariants aren't broken it's fine.

## In a separate PR

* [ ] Delete rich info from the database when the user is deleted (and test)
* [ ] Don't update rich info when it's unchanged (and test)
* [ ] Events (and test)
* [ ] Internal documentation

## The estimation game for this PR 🆕 🌟 

Estimated: 4 days.

So far: 5.5 days (Artyom, Chris, Matthias).

### DIFF
--- a/docs/reference/user/rich-info.md
+++ b/docs/reference/user/rich-info.md
@@ -100,4 +100,4 @@ Rich info can not be pulled from Wire and will not show up in SCIM queries.
 
 ## Limitations {#RefRichInfoLimitations}
 
-* JSON-encoded rich info cannot exceed 5000 characters in length. There are no limitations on the number of fields, or the length of field names or values.
+* The whole of user-submitted information (field names and values) cannot exceed 5000 characters in length. There are no limitations on the number of fields, or the length of field names or values.
--- a/docs/reference/user/rich-info.md
+++ b/docs/reference/user/rich-info.md
@@ -102,4 +102,6 @@ Rich info can not be pulled from Wire and will not show up in SCIM queries.
 
 ## Limitations {#RefRichInfoLimitations}
 
-* The whole of user-submitted information (field names and values) cannot exceed 5000 characters in length. There are no limitations on the number of fields, or the length of field names or values.
+* The whole of user-submitted information (field names and values) cannot exceed 5000 characters in length. There are no limitations on the number of fields, or the maximum of field names or values.
+
+* Field values can not be empty (`""`). If they are empty, the corresponding field will be removed.
--- a/docs/reference/user/rich-info.md
+++ b/docs/reference/user/rich-info.md
@@ -10,7 +10,7 @@ This page describes a part of the user profile called "Rich info". The correspon
 
 For every team user we can store a list of key-value pairs that are displayed in the user profile. This is similar to "custom profile fields" in Slack and other enterprise messengers.
 
-Different users can have different sets of fields; there is no team-wide schema for fields. All fields are strings.
+Different users can have different sets of fields; there is no team-wide schema for fields. All field values are strings. Fields are passed as an ordered list, and the order information is preserved when displaying fields in client apps.
 
 Only team members and partners can see the user's rich info.
 
@@ -38,7 +38,7 @@ Only team members and partners can see the user's rich info.
 
 If the requesting user is not allowed to see rich info, error code 403 is returned with the `"insufficient-permissions"` error label.
 
-If the rich info is missing, an empty field list is returned:
+Otherwise, if the rich info is missing, an empty field list is returned:
 
 ```json
 {
@@ -49,7 +49,7 @@ If the rich info is missing, an empty field list is returned:
 
 ### Setting rich info {#RefRichInfoPut}
 
-Not implemented yet.
+**Not implemented yet.** Currently the only way to set rich info is via SCIM.
 
 ### Events {#RefRichInfoEvents}
 
@@ -70,7 +70,7 @@ Connected users who are not members of user's team will not receive an event (no
 
 ## SCIM support {#RefRichInfoScim}
 
-Rich info can be pushed to Wire by setting the `"richInfo"` field belonging to the `"urn:wire:scim:schemas:profile:1.0"` extension. Both `PUT /scim/v2/Users/:id` and `PUT /scim/v2/Users/:id` are supported.
+Rich info can be pushed to Wire by setting the `"richInfo"` field belonging to the `"urn:wire:scim:schemas:profile:1.0"` extension. Both `PUT /scim/v2/Users/:id` and `POST /scim/v2/Users/:id` can contain rich info. Here is an example for `PUT`:
 
 ```javascript
 PUT /scim/v2/Users/:id
--- a/docs/reference/user/rich-info.md
+++ b/docs/reference/user/rich-info.md
@@ -1,0 +1,103 @@
+# Rich info {#RefRichInfo}
+
+_Author: Artyom Kazak_
+
+---
+
+This page describes a part of the user profile called "Rich info". The corresponding feature is called "Rich profiles".
+
+## Summary {#RefRichInfoSummary}
+
+For every team user we can store a list of key-value pairs that are displayed in the user profile. This is similar to "custom profile fields" in Slack and other enterprise messengers.
+
+Different users can have different sets of fields; there is no team-wide schema for fields. All fields are strings.
+
+Only team members and partners can see the user's rich info.
+
+## API {#RefRichInfoApi}
+
+### Querying rich info {#RefRichInfoGet}
+
+`GET /users/:user/rich-info`. Sample output:
+
+```json
+{
+    "fields": [
+        {
+            "type": "Department",
+            "value": "Sales & Marketing"
+        },
+        {
+            "type": "Favorite color",
+            "value": "Blue"
+        }
+    ]
+}
+```
+
+If the requesting user is not allowed to see rich info, error code 403 is returned with the `"insufficient-permissions"` error label.
+
+If the rich info is missing, an empty field list is returned:
+
+```json
+{
+    "fields": []
+}
+```
+
+### Setting rich info {#RefRichInfoPut}
+
+Not implemented yet.
+
+### Events {#RefRichInfoEvents}
+
+**Not implemented yet.**
+
+When user's rich info changes, the backend sends out an event to all team members:
+
+```json
+{
+    "type": "user.rich-info-update",
+    "user": {
+        "id": "<user ID>"
+    }
+}
+```
+
+Connected users who are not members of user's team will not receive an event (nor can they query user's rich info by other means).
+
+## SCIM support {#RefRichInfoScim}
+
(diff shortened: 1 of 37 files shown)


docs/reference/user/rich-info.md
```diff
@@ -0,0 +1,103 @@
...
+
+---
+
+This page describes a part of the user profile called "Rich info". The corresponding feature is called "Rich profiles".
+
+## Summary {#RefRichInfoSummary}
+
+For every team user we can store a list of key-value pairs that are displayed in the user profile. This is similar to "custom profile fields" in Slack and other enterprise messengers.
+
+Different users can have different sets of fields; there is no team-wide schema for fields. All fields are strings.
```
fisx: add `fields are passed as an ordered list, and the order information is used for display in the client.`

or something to that end.

docs/reference/user/rich-info.md
```diff
@@ -0,0 +1,103 @@
...
+            "type": "Favorite color",
+            "value": "Blue"
+        }
+    ]
+}
+```
+
+If the requesting user is not allowed to see rich info, error code 403 is returned with the `"insufficient-permissions"` error label.
+
+If the rich info is missing, an empty field list is returned:
```
fisx: "Otherwise, if the rich info is missing, ..."

docs/reference/user/rich-info.md
```diff
@@ -0,0 +1,103 @@
...
+        "id": "<user ID>"
+    }
+}
+```
+
+Connected users who are not members of user's team will not receive an event (nor can they query user's rich info by other means).
+
+## SCIM support {#RefRichInfoScim}
+
+Rich info can be pushed to Wire by setting the `"richInfo"` field belonging to the `"urn:wire:scim:schemas:profile:1.0"` extension. Both `PUT /scim/v2/Users/:id` and `PUT /scim/v2/Users/:id` are supported.
```
fisx: > Both `GET /scim/v2/Users/:id` and `PUT /scim/v2/Users/:id` are supported.

?

fisx: (what about delete?)

libs/brig-types/src/Brig/Types/User.hs
```diff
@@ -235,6 +235,45 @@ instance FromJSON SelfProfile where
...
+instance ToJSON RichInfo where
+    toJSON u = object
+        [ "fields" .= richInfoFields u
+        , "version" .= (0 :: Int)
+        ]
+
+instance FromJSON RichInfo where
+    parseJSON = withObject "RichInfo" $ \o -> do
+        o .:? "version" >>= \case
+            Nothing -> RichInfo <$> o .: "fields"
```
fisx: I would (a) make the `version` field mandatory, or (b) do not introduce it at all now, but with the second version, and then make `Nothing` the first version.  Having two names (Nothing and Just 0) for one version seems confusing to me.

libs/brig-types/src/Brig/Types/User.hs
```diff
@@ -235,6 +235,45 @@ instance FromJSON SelfProfile where
...
+
+data RichField = RichField
+    { richFieldType  :: !Text
+    , richFieldValue :: !Text
+    }
+    deriving (Eq, Show)
+
+instance ToJSON RichField where
+    toJSON u = object
+        [ "type" .= richFieldType u
```
fisx: `"name"` instead of `"type"`?  or `"key"`?  i like both better than `"type"`.

libs/brig-types/src/Brig/Types/User.hs
```diff
@@ -527,6 +534,13 @@ instance FromJSON ManagedByUpdate where
 instance ToJSON ManagedByUpdate where
     toJSON m = object ["managed_by" .= mbuManagedBy m]
 
+instance FromJSON RichInfoUpdate where
+    parseJSON = withObject "rich-info-update" $ \o ->
```
fisx: The description should be either the Haskell type or some English phrase; the dashes are odd.

services/brig/src/Brig/API.hs
```diff
@@ -1081,6 +1102,18 @@ getUserClient (user ::: cid ::: _) = lift $ do
         Just c  -> json c
         Nothing -> setStatus status404 empty
 
+getRichInfo :: UserId ::: UserId ::: JSON -> Handler Response
```
fisx: Are you going to use the references into the docs anywhere?

services/brig/src/Brig/API/User.hs
```diff
@@ -328,6 +330,15 @@ changeHandle uid conn hdl = do
             throwE ChangeHandleExists
         lift $ Intra.onUserEvent uid (Just conn) (handleUpdated uid hdl)
 
+----------------------------------------------------------------------------
+-- Change Rich Info
+
+changeRichInfo :: UserId -> ConnId -> RichInfoUpdate -> AppIO ()
+changeRichInfo uid _conn (RichInfoUpdate ri) = do
+    Data.updateRichInfo uid ri
+    -- FUTUREWORK: send an event
```
fisx: reference the docs here?

services/brig/src/Brig/Data/User.hs
```diff
@@ -313,6 +313,9 @@ lookupAuth u = fmap f <$> retry x1 (query1 authSelect (params Quorum (Identity u
   where
     f (pw, st) = (pw, fromMaybe Active st)
 
+-- | Return users with given IDs.
+--
+-- Skips nonexistent users. /Does not/ skip users who have been deleted.
```
fisx: I would argue that's surprising enough to justify giving this function the name `lookupUsersIncludeDeleted`.  This would also force us to go around and make sure all callers seem to know what they're doing.

fisx:
> * Mention that behavior with duplicate field names is undefined

Why?  The badness of undefined behavior is the sum of badnesses of all bad behaviors.  Why not just allow it?  It's straight-forward enough given it's different list elements.

libs/brig-types/src/Brig/Types/User.hs
neongreen: I also like "name" better, but having "type" in SCIM gives us more hope for compatibility because `{"type": ..., "value": ...}` is how all other SCIM stuff is formatted, and it's quite possible that some provisioning agent would support `"type"` but not `"name"`. Given this, having `"type"` in SCIM but `"name"` in Brig would be unnecessarily confusing.

fisx: Ok, agreed, but that would be good to document.  Just a short sentence like "'type' is better than the laternative 'key' or 'name' because SCIM uses it in analogous places."

neongreen:
> Why not just allow it?

* Saying that it's allowed requires actually supporting it in the future, which can turn out to be a pain, or prevent us from changing the database schema, or whatever.

* Saying that it's prohibited requires doing a check in the handler.

I guess that I'll go ahead and write this check (it's not hard) because otherwise somebody _will_ send us bad data anyway and we'll have to deal with it somehow.

fisx:
I'll add these tests listed in the description above:

- Guests can't see rich info
- Members of other teams can't see rich info
- Empty fields are deleted

fisx:
> I guess that I'll go ahead and write this check (it's not hard) because otherwise somebody _will_ send us bad data anyway and we'll have to deal with it somehow.

yes, please!  i think the only good excuse for "undefined behavior" is that validating input is computationally expensive, and may have been valid in 1960 when ANSI C was drafted, but it is not today.

docs/reference/user/rich-info.md
neongreen: Will do.

docs/reference/user/rich-info.md
neongreen: It should've been PUT and POST (Dejan got confused over whether POST was supported). DELETE is irrelevant here. I'll clarify the comment.

libs/brig-types/src/Brig/Types/User.hs
neongreen: I'll make it mandatory then.

services/brig/src/Brig/API.hs
neongreen: I will add references to internal docs (which have not been written yet). Internal docs will be a superset of reference docs.

neongreen:
> Empty fields are deleted

Just to clarify, it means that fields with value `""` should be deleted and not present in the JSON blob. The reason for this behavior is that ~always~ sometimes it might be impossible to configure ~Okta~ a provisioning agent in such a way that it doesn't send a field when that field is not present.

neongreen:
Also I have to note that before testing this behavior it should actually be implemented 🙂 I forgot that it wasn't.

libs/brig-types/src/Brig/Types/User.hs
neongreen: All other update types (that existed before me) use hyphens. I'm not sure why.

neongreen:
I remembered why I didn't want rich info to show up in SCIM:

> > By duplication I meant this blob only, which can grow and shrink, etc. and you will store the same blob on 2 different DB's
>
> What about this: we won't return it in SCIM and it will be write-only for SCIM. Then Spar doesn't need to store it.

On the other hand, if it doesn't show up in SCIM then "do a GET and then a PUT" becomes more fragile, so let's store it in `scim_user` after all (consistency trumps duplication).

services/spar/src/Spar/Scim/Types.hs
```diff
@@ -27,23 +27,79 @@ import Imports
...
+instance ToJSON ScimUserExtra where
+  toJSON v = object
+    [ userExtraURN .= _sueRichInfo v
+    ]
+
+-- | Parse 'RichInfo', trying several formats in a row. We have to know how to parse different
+-- formats, because not all provisioning agents can send us information in the canonical
+-- @ToJSON RichInfo@ format.
+--
+-- FUTUREWORK: allow strings as well
```
fisx: allow strings for what?  i don't know what this line means.

fisx:
```
----LOGS FROM spar-68f9668f9f-dbbw7:
service: no or bad config file: AesonException "Error in $: key \"richInfoLimit\" not present"


----LOGS FROM brig-858b9d8f69-kmjzv:
service: user error (AesonException "Error in $.optSettings: key \"setRichInfoLimit\" not present")
```

Is there a reason the two services have different names for the same field?

I forget: where do the integration tests pick up their config files when run on concourse?  I thought it was in `services/*/*.integration.yaml`, but those files have these fields.

PR #649, opened 2019-03-04
Updated emails

tiago-loureiro:
The previous version of the emails would stretch the logos... if you want to have a better look, here it is: https://wireapp.github.io/wire-emails/dist (although not too useful to check the _actual_ fix...)


tiago-loureiro:
> Impossible to read the diff

I know :/

PR #650, opened 2019-03-05
Add spar to the arch diagram.

fisx:


fisx:
(is there any place where this is used and we need to check the new width works out?)

ChrisPenner:
BTW when changing images it's nice to have the image in the PR description so we don't have to hunt down the new version 😄 
Here it is for others looking:

diagram

I thought there was one other spot where this was used; but I can't remember it off the top of my head. I think it was just in a github readme though so it should be fine 👍

neongreen:
> BTW when changing images it's nice to have the image in the PR description so we don't have to hunt down the new version

You can go to the diff and click the rich view buttom

neongreen:
> I think it was just in a github readme though so it should be fine

The Github readme just links to this diagram (unless you're talking about some other Github readme)

fisx:
concourse is having k8s issues.  i don't really feel like running it again for this one.  :)

tiago-loureiro:
Can we perhaps make the arrow not cross-over but go around gundeck/cannon (pedantic arrow deisgn kicking in :)) ? Also, from a quick glance, I have the feeling that spar talks only to galley, I think it could be made more clear if it goes around it (sort of opposite to nginz)

fisx:
> Can we perhaps make the arrow not cross-over but go around gundeck/cannon (pedantic arrow deisgn kicking in :)) ? Also, from a quick glance, I have the feeling that spar talks only to galley, I think it could be made more clear if it goes around it (sort of opposite to nginz)

[spar talks to both brig and galley](https://github.com/wireapp/wire-server/tree/develop/services/spar/src/Spar/Intra).  i think this means that brig and galley form a fully connected subgraph.  :-)

I prefer the arrow like this, but if that doesn't impress you ask me one more time and I'll try to make a new version that you like!

jschaul:
Minor nitpicks:

* SCIM peers are not a dependency for Spar: the communication is initiated from the SCIM peer to spar, not the other way around (or did I misunderstand?)
* the new arrow is not straight :crying_cat_face:

PR #652, opened 2019-03-05
Fix: empty objects `{}` are valid TeamMemberDeleteData.

fisx:
This is a production issue, but the web team agreed to provide a work-around and submit `{ "password": null }` rather than `{}`.


fisx:
Thanks @bennyn, @Yserz!

PR #659, opened 2019-03-13
Docs: using scim with curl.

fisx:


docs/reference/user/provisioning/scim-via-curl.md
```diff
@@ -0,0 +1,172 @@
+# Using the SCIM API with curl {#RefScimViaCurl}
```
neongreen: I would recommend changing the path to "docs/reference/provisioning/scim-via-curl.md" because the `reference/user/provisioning` hierarchy seems needlessly deep.

(NB: I intended "user/" to be "parts of the user concept", not "everything related to users". But then I actually expect us to change the hierarchy several times, anyway.)

docs/reference/user/provisioning/scim-via-curl.md
```diff
@@ -0,0 +1,172 @@
...
+If you want to dive into the backend code, start [reading here in our
+backend](https://github.com/wireapp/wire-server/blob/develop/services/spar/src/Spar/Scim.hs)
+and [our hsim library](https://github.com/wireapp/hscim).
+
+
+## Creating an SCIM token
+
+First, we need a little shell environment:
+
+```
```
neongreen:
```bash

File: docs/reference/user/provisioning/scim-via-curl.md [hunk 3]
```diff
@@ -0,0 +1,172 @@
...
+export WIRE_BACKEND=prod-nginz-https.wire.com
```
neongreen: Perhaps change it to `https://prod-nginz-https.wire.com` so that "https" would not be hardcoded into bash snippets. (Somebody might want to try all that locally and then it'd be "http".)

docs/reference/user/provisioning/scim-via-curl.md
```diff
@@ -0,0 +1,172 @@
...
+  https://$WIRE_BACKEND/self
+```
+
+Now you are ready to create a SCIM token:
+
+```
+export SCIM_TOKEN_FULL=$(curl -X POST \
+  --header "Authorization: Bearer $BEARER" \
+  --header 'Content-Type: application/json;charset=utf-8' \
+  -d '{ "description": "scim-test-1", "password": "'"$WIRE_PASSWD"'" }' \
```
neongreen: > "scim-test-1"

Maybe use $(date) as a part of it?

docs/reference/user/provisioning/scim-via-curl.md
```diff
@@ -0,0 +1,172 @@
...
+  "schemas"     : ["urn:ietf:params:scim:schemas:core:2.0:User"],
+  "externalId"  : "f8c4ffde-4592-11e9-8600-afe11dc7d07b",
+  "userName"    : "nick",
+  "displayName" : "The Nick"
+}'
+```
+
+We also support custom fields that are used in rich profiles in this
+form [see
+{#RefRichInfo}](https://github.com/wireapp/wire-server/blob/develop/docs/reference/user/rich-info.md):
```
neongreen: I think the link should be something like `(../rich-info.md)`, otherwise it won't work when publishing to Github Pages.

* https://github.blog/2013-01-31-relative-links-in-markup-files/
* https://github.blog/2016-12-05-relative-links-for-github-pages/

docs/reference/user/provisioning/scim-via-curl.md
```diff
@@ -0,0 +1,172 @@
...
+
+```
+export SCIM_USER='{
+  "schemas"     : ["urn:ietf:params:scim:schemas:core:2.0:User"],
```
neongreen:
```suggestion
  "schemas"     : ["urn:ietf:params:scim:schemas:core:2.0:User", "urn:wire:scim:schemas:profile:1.0"],
```

docs/reference/user/provisioning/scim-via-curl.md
```diff
@@ -0,0 +1,172 @@
...
+admin credentials.  The way this works behind the scenes in your
+browser or cell phone, and in plain sight if you want to use curl, is
+you need to get a wire token.
+
+```
+export BEARER=$(curl -X POST \
+  --header 'Content-Type: application/json' \
+  --header 'Accept: application/json' \
+  -d '{"email":"'"$WIRE_ADMIN"'","password":"'"$WIRE_PASSWD"'"}' \
+  https://$WIRE_BACKEND/login'?persist=false' | jq -r .access_token)
```
neongreen: > `login'?persist=false'`

How does this work?

docs/reference/user/provisioning/scim-via-curl.md
```diff
@@ -0,0 +1,172 @@
...
+```
+curl -X GET \
+  --header "Authorization: Bearer $SCIM_TOKEN" \
+  --header 'Content-Type: application/json;charset=utf-8' \
+  https://$WIRE_BACKEND/scim/v2/Users/
+```
+
+### update user
+
+```
```
neongreen: It'd be nice to mention that PUT requests can not be partial.

fisx:
particularly good feedback today, thanks!

docs/reference/user/provisioning/scim-via-curl.md
neongreen: I meant the single quotes

fisx: they keep the `?` from making bash try to expand this into a filename with one (or optionally none?  not sure about the precise semantics.) character intsead of the `?`.

fisx:
> If you tested it, I have no objections

I did, thanks!

PR #671, opened 2019-03-21
Remove some unused instances

neongreen:


neongreen:
(No, I was just going through the code and it looked like they were unused)

PR #681, opened 2019-03-27
Tweak arch diagram.

fisx:
Fixes https://github.com/wireapp/wire-server/pull/650#issuecomment-476726549

ping @jschaul


jschaul:
(Perhaps we can discuss when physically in the same place? describing images through text is a little hard)

fisx:
I think the diagram was good enough as it was, and this isn't a productive use of our time.  If you disagree feel free to take over.

PR #685, opened 2019-03-27
Use URLs for snapshots; switch fork repos

neongreen:
Now wire-1.2.yaml contains a link to wire-1.1.yaml, and not just a relative path.

Without this PR we can't use wire-server snapshots in other repos, because Stack is not smart enough to fetch snapshots based on relative paths.

From this point on, the resolver/snapshot files *actually* have to be immutable.

This PR will cause another "copying precompiled package" cascade. @fisx sorry!

### DIFF
--- a/snapshots/README.md
+++ b/snapshots/README.md
@@ -1,0 +1,23 @@
+This directory contains [custom Stack snapshots][custom] used for Wire code.
+
+[custom]: https://docs.haskellstack.org/en/stable/custom_snapshot/
+
+Snapshot definitions should never be changed (once committed to `develop`), because in other
+repositories we refer to snapshot definitions by URL.
+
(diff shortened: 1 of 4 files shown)


neongreen:
@ChrisPenner I added a readme.

neongreen:
This won't/can't pass the CI.

neongreen:
Since this PR breaks snapshot caching anyway, I've also updated snapshot files to point at @wireapp forks instead of mine and Tiago's forks. The only repo left to transfer is https://github.com/tiago-loureiro/haskell-multihash.

tiago-loureiro:
> Since this PR breaks snapshot caching anyway, I've also updated snapshot files to point at @wireapp forks instead of mine and Tiago's forks. The only repo left to transfer is https://github.com/tiago-loureiro/haskell-multihash.

Done

PR #686, opened 2019-03-27
Reexport System.Logger from System.Logger.Extended

neongreen:


neongreen:
Note to self: don't merge before https://github.com/wireapp/wire-server/pull/654

services/spar/src/Spar/Run.hs
```diff
@@ -39,7 +39,6 @@ import qualified Network.Wai.Middleware.Prometheus as Promth
 import qualified Network.Wai.Utilities.Server as WU
 import qualified SAML2.WebSSO as SAML
 import qualified Spar.Data as Data
-import qualified System.Logger as Log
 import qualified System.Logger.Extended as Log
```
fisx: What about re-exporting `Extended` as well?  (I have no opinion.)

neongreen: Re-exporting `Extended` from where?

fisx: From `System.Logger.Extended`, like `System.Logger`.

PR #687, opened 2019-03-27
Cassandra cleanup

julialongtin:
without explicit imports.


libs/cassandra-util/src/Cassandra/Exec.hs
```diff
@@ -1,58 +1,29 @@
...
 
-import Imports
-import Cassandra.CQL
+import Imports hiding (init)
+import Cassandra.CQL (R, Consistency)
 import Control.Monad.Catch
 import Data.Conduit
-import Database.CQL.IO
+
+-- | Things we just import and re-export.
```
neongreen:
```suggestion
-- Things we just import and re-export.
```

libs/cassandra-util/src/Cassandra/Exec.hs
```diff
@@ -1,58 +1,29 @@
...
+import Database.CQL.IO as C (MonadClient, Client, ClientState, BatchM, PrepQuery, Page(hasMore, result, nextPage), adjustResponseTimeout, adjustSendTimeout, adjustConsistency, batch, setSerialConsistency, setConsistency, setType, addPrepQuery, addQuery, queryString, prepared, schema, write, query1, query, emptyPage, shutdown, runClient, retry, localState, liftClient, paginate, init)
+-- | We only use these locally.
```
neongreen:
```suggestion
-- We only use these locally.
```

PR #692, opened 2019-03-28
Fix prometheus metrics.

fisx:
In #654, we started counting errors into prometheus via `/i/metrics` directly.  This doesn't work yet (thanks @jschaul for reporting):

x

```
http_request_duration_seconds_count{handler="prometheus",method="GET",status_code="200"} 20
# HELP net.errors count status >= 500 responses
# TYPE net.errors counter
net.errors 0.0
```

14f8f6ea2808574fc911e025c3396ffbb738caa5 disables counting altogether, which reverts to the behavior before #654 (at least regarding error counts).

Spar constructs the prometheus counter [like this](https://github.com/wireapp/wire-server/blob/b9a6af6f05f3ce297b6685f8fcb6a035d0bc5777/services/spar/src/Spar/Run.hs#L89), and the label looks exactly like [this one](https://github.com/wireapp/wire-server/blob/b9a6af6f05f3ce297b6685f8fcb6a035d0bc5777/libs/wai-utilities/src/Network/Wai/Utilities/Server.hs#L306).

I suspect there is some transformation of the label going on in `metrics-core` or somewhere in our tooling that I need to replicate in constructing the counter.

### DIFF
--- a/libs/wai-utilities/src/Network/Wai/Utilities/Server.hs
+++ b/libs/wai-utilities/src/Network/Wai/Utilities/Server.hs
@@ -13,7 +13,6 @@ module Network.Wai.Utilities.Server
       -- * Middlewares
     , measureRequests
     , catchErrors
-    , OnErrorMetrics
     , heavyDebugLogging
 
       -- * Utilities
@@ -55,7 +54,6 @@ import qualified Data.Text.Lazy.Encoding     as LT
 import qualified Network.Wai.Predicate       as P
 import qualified Network.Wai.Routing.Route   as Route
 import qualified Network.Wai.Utilities.Error as Error
-import qualified Prometheus                  as Prm
 import qualified System.Logger               as Log
 import qualified System.Posix.Signals        as Sig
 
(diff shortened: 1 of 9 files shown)


fisx:
metrics-core:

```
$ curl http://localhost:8082/i/monitoring | json_p
{
   [...]
   "net" : {
      "errors" : 0,
      [...]
      }
   }
}
```

so i guess i need to run `/libs/metrics-collectd` over this and see what it spits out, and that's what i need to construct instead of "net.errors".

fisx:
- 2: I've just copied the behavior of the old metrics logic to prometheus, so if it is wrong now, it was wrong before.  I will remove 'net.errors' everywhere, please take a look if that's what you want.
- 1: Solved!  :)

fisx:
@jschaul as discussed: i reverted the over-enthusiastic commit that removed everything, and tried again, this time only removing the `net_errors` counter in prometheus, leaving the old metrics code intact.  good?

PR #695, opened 2019-03-28
Move the connections doc, add haddocks

neongreen:


neongreen:
Not very sure that Message is actually unused by all clients

tiago-loureiro:
> Not very sure that Message is actually unused by all clients

Some clients actually do send it but we have hidden it anyway in the UI for a while since it works as a nice source of spam.

We can (and should) remove this at some point, perhaps adding a TODO in this PR?

PR #696, opened 2019-03-28
attempt to fix branding for outgoing SMSs.

julialongtin:


tiago-loureiro:
> I notice that renderTextWithBranding nor the rest of the renderText* family have any unit tests; I think these should be pretty simple to write since they're all pure functions and would help ensure that our template keys and branding replacement stay in sync. They would be a welcome addition to this PR 😄

👍 Totally agree that it makes sense, but let's do it in a separate PR though; given that this is currently affecting `prod`, it would be good to roll it out sooner rather than later.

Code looks good to me too and I tried this out _manually_ too locally using my own phone number.

PR #698, opened 2019-04-01
Fix: user deletion via scim

fisx:
The new tests fail thusly:

```
Failures:

  test-integration/Test/Spar/Scim/UserSpec.hs:543:22: 
  1) Spar.Scim.User, DELETE /Users/:id, should delete user from brig, spar.scim_user, spar.user
       expected: (Nothing,Nothing,Nothing)
        but got: (Just (User {userId = 3eb69525-b292-44a7-bba3-b33af69733a2, userIdentity = Just (SSOIdentity (UserSSOId {userSSOIdTenant = "<Issuer xmlns:samlp=\"urn:oasis:names:tc:SAML:2.0:protocol\" xmlns:samla=\"urn:oasis:names:tc:SAML:2.0:assertion\" xmlns:samlm=\"urn:oasis:names:tc:SAML:2.0:metadata\" xmlns:ds=\"http://www.w3.org/2000/09/xmldsig#\" xmlns=\"urn:oasis:names:tc:SAML:2.0:assertion\">https://issuer.net/_802b24ec-0542-4ff5-bd82-b688478b4272</Issuer>", userSSOIdSubject = "<NameID xmlns:samlp=\"urn:oasis:names:tc:SAML:2.0:protocol\" xmlns:samla=\"urn:oasis:names:tc:SAML:2.0:assertion\" xmlns:samlm=\"urn:oasis:names:tc:SAML:2.0:metadata\" xmlns:ds=\"http://www.w3.org/2000/09/xmldsig#\" xmlns=\"urn:oasis:names:tc:SAML:2.0:assertion\">scimuser_extid_8974660</NameID>"}) Nothing Nothing), userName = Name {fromName = "Scim User #8974660"}, userPict = Pict {fromPict = []}, userAssets = [], userAccentId = ColourId {fromColourId = 0}, userDeleted = False, userLocale = en, userService = Nothing, userHandle = Just (Handle {fromHandle = "scimuser_8974660"}), userExpire = Nothing, userTeam = Just 189f6570-3ef1-4a91-b3e1-a57e71f4453e, userManagedBy = ManagedByScim}),Nothing,Nothing)

  To rerun use: --match "/Spar.Scim.User/DELETE /Users/:id/should delete user from brig, spar.scim_user, spar.user/"

  test-integration/Test/Spar/Scim/UserSpec.hs:568:13: 
  2) Spar.Scim.User, DELETE /Users/:id, should free externalId and everything else in the scim user for re-use
       uncaught exception: ErrorCall
       Assertions failed:
        1: 201 =/= 409

       Response was:

       Response {responseStatus = Status {statusCode = 409, statusMessage = "bad-upstream"}, responseVersion = HTTP/1.1, responseHeaders = [("Transfer-Encoding","chunked"),("Date","Mon, 01 Apr 2019 11:00:55 GMT"),("Server","Warp/3.2.25"),("Cache-Control","no-cache, no-store"),("Pragma","no-cache")], responseBody = Just "{\"code\":409,\"message\":\"set handle failed\",\"label\":\"bad-upstream\"}", responseCookieJar = CJ {expose = []}, responseClose' = ResponseClose}
         error, called at src/Bilge/Assert.hs:69:30 in bilge-0.22.0-6BuG2eVEmcLDQI4T1WJLQ7:Bilge.Assert
         <!!, called at src/Bilge/Assert.hs:86:19 in bilge-0.22.0-6BuG2eVEmcLDQI4T1WJLQ7:Bilge.Assert
         !!!, called at test-integration/Test/Spar/Scim/UserSpec.hs:568:13 in main:Test.Spar.Scim.UserSpec

  To rerun use: --match "/Spar.Scim.User/DELETE /Users/:id/should free externalId and everything else in the scim user for re-use/"

Randomized with seed 437206805

Finished in 35.6825 seconds
153 examples, 2 failures, 31 pending
```

### DIFF
--- a/services/brig/src/Brig/Unique.hs
+++ b/services/brig/src/Brig/Unique.hs
@@ -2,6 +2,7 @@
 -- to contention, i.e. where strong guarantees on uniqueness are desired.
 module Brig.Unique
     ( withClaim
+    , deleteClaim
     , lookupClaims
 
       -- * Re-exports
@@ -49,6 +50,18 @@ withClaim u v t io = do
 
     cql :: PrepQuery W (Int32, C.Set (Id a), Text) ()
     cql = "UPDATE unique_claims USING TTL ? SET claims = claims + ? WHERE value = ?"
+
+deleteClaim :: MonadClient m
+    => Id a
+    -> Text
+    -> Timeout
+    -> m ()
+deleteClaim u v t = do
+    let ttl = max minTtl (fromIntegral (t #> Second))
+    retry x5 $ write cql $ params Quorum (ttl * 2, C.Set [u], v)
+  where
+    cql :: PrepQuery W (Int32, C.Set (Id a), Text) ()
+    cql = "UPDATE unique_claims USING TTL ? SET claims = claims - ? WHERE value = ?"
 
 -- | Lookup the current claims on a value.
 lookupClaims :: MonadClient m => Text -> m [Id a]
--- a/services/brig/src/Brig/Unique.hs
+++ b/services/brig/src/Brig/Unique.hs
@@ -52,9 +52,12 @@ withClaim u v t io = do
     cql = "UPDATE unique_claims USING TTL ? SET claims = claims + ? WHERE value = ?"
 
 deleteClaim :: MonadClient m
-    => Id a
-    -> Text
-    -> Timeout
+    => Id a     -- ^ The 'Id' associated with the claim.
+    -> Text     -- ^ The value on which to acquire the claim.
+    -> Timeout  -- ^ The minimum timeout (i.e. duration) of the rest of the claim.  (Each
+                --   claim can have more than one claimer (even though this is a feature we
+                --   never use), so removing a claim is an update operation on the database.
+                --   Therefore, we reset the TTL the same way we reset it in 'withClaim'.)
     -> m ()
 deleteClaim u v t = do
     let ttl = max minTtl (fromIntegral (t #> Second))
--- a/services/brig/src/Brig/User/Handle.hs
+++ b/services/brig/src/Brig/User/Handle.hs
@@ -32,13 +32,16 @@ claimHandle u h = do
                     -- Update profile
                     User.updateHandle (userId u) h
                     -- Free old handle (if it changed)
-                    for_ (mfilter (/= h) (userHandle u))
-                        freeHandle
+                    for_ (mfilter (/= h) (userHandle u)) $
+                        freeHandle u
             return (isJust claimed)
 
 -- | Free a 'Handle', making it available to be claimed again.
-freeHandle :: Handle -> AppIO ()
-freeHandle h = retry x5 $ write handleDelete (params Quorum (Identity h))
+freeHandle :: User -> Handle -> AppIO ()
+freeHandle u h = do
+    retry x5 $ write handleDelete (params Quorum (Identity h))
+    let key = "@" <> fromHandle h
+    deleteClaim (userId u) key (30 # Minute)
 
 -- | Lookup the current owner of a 'Handle'.
 lookupHandle :: Handle -> AppIO (Maybe UserId)
--- a/services/brig/test/integration/API/User/Handles.hs
+++ b/services/brig/test/integration/API/User/Handles.hs
@@ -85,18 +85,20 @@ testHandleUpdate brig cannon = do
     Search.assertCan'tFind brig uid2 uid hdl
     Search.assertCanFind   brig uid2 uid hdl2
 
-    -- Other users can immediately claim the old handle (the claim is removed with the old
-    -- handle).
+    -- Other users can immediately claim the old handle (the claim of the old handle is
+    -- removed).
     put (brig . path "/self/handle" . contentJson . zUser uid2 . zConn "c" . body update) !!! do
         const 200 === statusCode
 
     -- The old handle can be claimed again immediately by the user who previously
     -- owned it (since the claim is either still active but his own, or expired).
+    -- make sure 'hdl' is not used by 'uid2' already.
     hdl3 <- randomHandle
     let update3 = RequestBodyLBS . encode $ HandleUpdate hdl3
     put (brig . path "/self/handle" . contentJson . zUser uid2 . zConn "c" . body update3) !!! do
         const 200 === statusCode
-    put (brig . path "/self/handle" . contentJson . zUser uid . zConn "c" . body update) !!!
+    -- now 'uid2' takes 'hld' back.
+    put (brig . path "/self/handle" . contentJson . zUser uid2 . zConn "c" . body update) !!!
         const 200 === statusCode
 
 testHandleRace :: Brig -> Http ()
--- a/services/brig/test/integration/API/User/Handles.hs
+++ b/services/brig/test/integration/API/User/Handles.hs
@@ -85,14 +85,17 @@ testHandleUpdate brig cannon = do
     Search.assertCan'tFind brig uid2 uid hdl
     Search.assertCanFind   brig uid2 uid hdl2
 
-    -- Other users cannot immediately claim the old handle since the previous claim
-    -- is still active.
+    -- Other users can immediately claim the old handle (the claim is removed with the old
+    -- handle).
     put (brig . path "/self/handle" . contentJson . zUser uid2 . zConn "c" . body update) !!! do
-        const 409 === statusCode
-        const (Just "handle-exists") === fmap Error.label . decodeBody
+        const 200 === statusCode
 
     -- The old handle can be claimed again immediately by the user who previously
     -- owned it (since the claim is either still active but his own, or expired).
+    hdl3 <- randomHandle
+    let update3 = RequestBodyLBS . encode $ HandleUpdate hdl3
+    put (brig . path "/self/handle" . contentJson . zUser uid2 . zConn "c" . body update3) !!! do
+        const 200 === statusCode
     put (brig . path "/self/handle" . contentJson . zUser uid . zConn "c" . body update) !!!
         const 200 === statusCode
 
--- a/services/spar/src/Spar/Intra/Brig.hs
+++ b/services/spar/src/Spar/Intra/Brig.hs
@@ -273,9 +273,9 @@ deleteBrigUser buid = do
   if
     | sCode < 300 -> pure ()
     | inRange (400, 499) sCode
-      -> throwSpar . SparBrigErrorWith (responseStatus resp) $ "failed to delete user"
-    | otherwise -> throwSpar . SparBrigError . cs
-      $ "delete user failed with status " <> show sCode
+      -> throwSpar $ SparBrigErrorWith (responseStatus resp) "failed to delete user"
+    | otherwise
+      -> throwSpar $ SparBrigError ("delete user failed with status " <> cs (show sCode))
 
 -- | Check that a user id exists on brig and has a team id.
 isTeamUser :: (HasCallStack, MonadSparToBrig m) => UserId -> m Bool
--- a/services/spar/src/Spar/Intra/Brig.hs
+++ b/services/spar/src/Spar/Intra/Brig.hs
@@ -12,6 +12,7 @@ module Spar.Intra.Brig
   , setBrigUserHandle
   , setBrigUserManagedBy
   , setBrigUserRichInfo
+  , checkHandle
   , bindBrigUser
   , deleteBrigUser
   , createBrigUser
@@ -253,6 +254,21 @@ setBrigUserRichInfo buid richInfo = do
        -> throwSpar . SparBrigErrorWith (responseStatus resp) $ "set richInfo failed"
      | otherwise
        -> throwSpar . SparBrigError . cs $ "set richInfo failed with status " <> show sCode
+
+checkHandle :: (HasCallStack, MonadSparToBrig m) => Handle -> UserId -> m ()
+checkHandle hnd buid = do
+  resp <- call
+    $ method HEAD
+    . paths ["users", "handles", toByteString' hnd]
+    . header "Z-User" (toByteString' buid)
+    . header "Z-Connection" ""
+  let sCode = statusCode resp
+  if | sCode < 300
+       -> pure ()
+     | inRange (400, 499) sCode
+       -> throwSpar . SparBrigErrorWith (responseStatus resp) $ "check handle failed"
+     | otherwise
+       -> throwSpar . SparBrigError . cs $ "check handle failed with status " <> show sCode
 
 -- | This works under the assumption that the user must exist on brig.  If it does not, brig
 -- responds with 404 and this function returns 'False'.
--- a/services/spar/src/Spar/Intra/Brig.hs
+++ b/services/spar/src/Spar/Intra/Brig.hs
@@ -255,6 +255,8 @@ setBrigUserRichInfo buid richInfo = do
      | otherwise
        -> throwSpar . SparBrigError . cs $ "set richInfo failed with status " <> show sCode
 
+-- | At the time of writing this, @HEAD /users/handles/:uid@ does not use the 'UserId' for
+-- anything but authorization.
 checkHandleAvailable :: (HasCallStack, MonadSparToBrig m) => Handle -> UserId -> m Bool
 checkHandleAvailable hnd buid = do
   resp <- call
--- a/services/spar/src/Spar/Intra/Brig.hs
+++ b/services/spar/src/Spar/Intra/Brig.hs
@@ -263,9 +263,9 @@ checkHandle hnd buid = do
     . header "Z-User" (toByteString' buid)
     . header "Z-Connection" ""
   let sCode = statusCode resp
-  if | sCode < 300
+  if | sCode == 404
        -> pure ()
-     | inRange (400, 499) sCode
+     | sCode < 500
        -> throwSpar . SparBrigErrorWith (responseStatus resp) $ "check handle failed"
      | otherwise
        -> throwSpar . SparBrigError . cs $ "check handle failed with status " <> show sCode
--- a/services/spar/src/Spar/Intra/Brig.hs
+++ b/services/spar/src/Spar/Intra/Brig.hs
@@ -12,7 +12,7 @@ module Spar.Intra.Brig
   , setBrigUserHandle
   , setBrigUserManagedBy
   , setBrigUserRichInfo
-  , checkHandle
+  , checkHandleAvailable
   , bindBrigUser
   , deleteBrigUser
   , createBrigUser
@@ -255,16 +255,18 @@ setBrigUserRichInfo buid richInfo = do
      | otherwise
        -> throwSpar . SparBrigError . cs $ "set richInfo failed with status " <> show sCode
 
-checkHandle :: (HasCallStack, MonadSparToBrig m) => Handle -> UserId -> m ()
-checkHandle hnd buid = do
+checkHandleAvailable :: (HasCallStack, MonadSparToBrig m) => Handle -> UserId -> m Bool
+checkHandleAvailable hnd buid = do
   resp <- call
     $ method HEAD
     . paths ["users", "handles", toByteString' hnd]
     . header "Z-User" (toByteString' buid)
     . header "Z-Connection" ""
   let sCode = statusCode resp
-  if | sCode == 404
-       -> pure ()
+  if | sCode == 200  -- handle exists
+       -> pure False
+     | sCode == 404  -- handle not found
+       -> pure True
      | sCode < 500
        -> throwSpar . SparBrigErrorWith (responseStatus resp) $ "check handle failed"
      | otherwise
--- a/services/spar/src/Spar/Scim/Types.hs
+++ b/services/spar/src/Spar/Scim/Types.hs
@@ -38,8 +38,11 @@ import Spar.Types
 import qualified Data.HashMap.Strict              as HM
 import qualified Data.Text                        as T
 import qualified SAML2.WebSSO                     as SAML
-import qualified Web.Scim.Schema.User             as Scim.User
+import qualified Web.Scim.Class.Auth              as Scim.Auth
+import qualified Web.Scim.Class.Group             as Scim.Group
+import qualified Web.Scim.Class.User              as Scim.User
 import qualified Web.Scim.Schema.Schema           as Scim
+import qualified Web.Scim.Schema.User             as Scim.User
 import qualified Web.Scim.Server                  as Scim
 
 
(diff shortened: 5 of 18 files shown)


neongreen:
I'm going to look at this now.

services/spar/src/Spar/Scim/User.hs
```diff
@@ -266,7 +234,7 @@ validateScimUser' idp richInfoLimit user = do
 -- requirement.)
 createValidScimUser
   :: forall m. (m ~ Scim.ScimHandler Spar)
-  => ValidScimUser -> m (Scim.StoredUser ScimUserExtra)
+  => ValidScimUser -> m (Scim.StoredUser SparTag)
 createValidScimUser (ValidScimUser user uref handl mbName richInfo) = do
 
     -- FUTUREWORK: The @hscim@ library checks that the handle is not taken before 'create' is
```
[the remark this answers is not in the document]
neongreen: The next comment ("FUTUREWORK: it's annoying that we have duplicate checks (handles, rich info, etc are validated both by Spar and by Brig)") is also outdated.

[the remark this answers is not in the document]
neongreen: When this PR is merged, please add "SCIM users are written to the database before even checking things like 'handle is not claimed'" to the list of SCIM problems.

services/spar/src/Spar/Scim/User.hs
```diff
@@ -315,24 +283,24 @@ updateValidScimUser tokinfo uidText newScimUser = do
...
           <- lift $ updScimStoredUser (newScimUser ^. vsuUser) oldScimStoredUser
 
         -- update 'SAML.UserRef'
-        let uref = newScimUser ^. vsuSAMLUserRef
-        lift . wrapMonadClient $ Data.insertSAMLUser uref uid  -- on spar
-        bindok <- lift $ Intra.Brig.bindBrigUser uid uref  -- on brig
+        let newuref = newScimUser ^. vsuSAMLUserRef
+        lift . wrapMonadClient $ Data.insertSAMLUser newuref uid  -- on spar
+        -- TODO: compute UserRef from oldScimStoredUser and call Data.deleteSAMLUser on that
+        -- (exercise to the reader: what goes wrong if we don't?  it has nothin to do with the
```
neongreen: ~Life~ Reading code is hard enough as it is, let's not have exercises for the reader.

services/spar/src/Spar/Scim/User.hs
```diff
@@ -426,29 +427,36 @@ parseUid uidText = maybe err pure $ readMaybe (Text.unpack uidText)
 -- requirements of strong ETags ("same resources have the same version").
 calculateVersion
   :: Text               -- ^ User ID
```
neongreen: Let's have a `UserId` here.

services/spar/test-integration/Spec.hs
```diff
@@ -29,7 +29,7 @@ main :: IO ()
 main = do
   (wireArgs, hspecArgs) <- partitionArgs <$> getArgs
   env <- withArgs wireArgs mkEnvFromOptions
-  withArgs hspecArgs . hspec . beforeAll (pure env) . afterAll destroyEnv $ mkspec
+  withArgs (hspecArgs <> ["-m", "###"]) . hspec . beforeAll (pure env) . afterAll destroyEnv $ mkspec
```
neongreen: Looks like you can use [`focus`](https://hackage.haskell.org/package/hspec-2.7.1/docs/Test-Hspec.html#v:focus) instead.

services/spar/src/Spar/Scim/User.hs
```diff
@@ -377,43 +345,76 @@ toScimStoredUser' (SAML.Time now) baseuri (idToText -> uid) usr =
...
   where
     meta' = meta
       { Scim.lastModified = moddate
-      , Scim.version = calculateVersion scimuid usr
+      , Scim.version = calculateVersion (idToText scimuid) usr
       }
 
+
+deleteScimUser
+  :: ScimTokenInfo -> UserId -> ExceptT Scim.ScimError Spar ()
```
neongreen: Elsewhere we have `Scim.ScimHandler Spar`.

services/spar/test-integration/Test/Spar/DataSpec.hs
```diff
@@ -118,6 +117,19 @@ spec = do
...
+          uref <- nextUserRef
+          uid  <- nextWireId
+          do
+            ()   <- runSparCass $ Data.insertSAMLUser uref uid
+            muid <- runSparCass (Data.getSAMLUser uref) `eventually` isJust
+            liftIO $ muid `shouldBe` Just uid
+          do
+            ()   <- runSparCass $ Data.deleteSAMLUser uref
+            muid <- runSparCass (Data.getSAMLUser uref) `eventually` isNothing
+            liftIO $ muid `shouldBe` Nothing
```
neongreen: `shouldBe` after `eventually` seems to contribute nothing in this case. (Unlike `` `shouldBe` Just uid ``.)

services/spar/test-integration/Test/Spar/Scim/UserSpec.hs
```diff
@@ -519,7 +522,29 @@ specDeleteUser = do
...
+                !!! const 204 === statusCode
+
+            brigUser :: Maybe User
+              <- eventually (runSpar $ Intra.getBrigUser uid) isNothing
+            samlUser :: Maybe UserId
+              <- eventually (getUserIdViaRef' uref) isNothing
+            scimUser :: Maybe (ScimC.User.StoredUser SparTag)
+              <- eventually (getScimUser uid) isNothing
+
+            liftIO $ (brigUser, samlUser, scimUser)
```
neongreen: Useless `shouldBe` again

services/spar/test-integration/Test/Spar/Scim/UserSpec.hs
```diff
@@ -529,11 +554,20 @@ specDeleteUser = do
             -- Expect first call to succeed
             deleteUser_ (Just tok) (Just uid) spar
                 !!! const 204 === statusCode
-            -- The second call may return either of 204 or 404 depending on whether Brig has
-            -- finished deletion. This assertion is here to document that this is currently
-            -- the expected behaviour
+            -- Subsequent calls will return 404 eventually
+            eventually (deleteUser_ (Just tok) (Just uid) spar) ((== 404) . statusCode)
+                !!! const 404 === statusCode
```
neongreen: Useless `!!!`

services/spar/test-integration/Util/Core.hs
```diff
@@ -205,6 +209,17 @@ pendingWith :: (HasCallStack, MonadIO m) => String -> m ()
 pendingWith = liftIO . Test.Hspec.pendingWith
 
 
+-- | Run a probe a couple of times, until a "good" value materializes or until patience runs
```
neongreen: 15 sounds like more than a couple

```suggestion
-- | Run a probe several times, until a "good" value materializes or until patience runs
```

services/spar/test-integration/Util/Core.hs
```diff
@@ -205,6 +209,17 @@ pendingWith :: (HasCallStack, MonadIO m) => String -> m ()
...
+-- out.
+eventually :: TestSpar a -> (a -> Bool) -> TestSpar a
+eventually action good = do
+    env <- ask
+    liftIO $ retrying
+        (exponentialBackoff 50 <> limitRetries 15)
```
neongreen: Nitpick: 50µs is rather small for anything IO-related.

services/spar/test-integration/Util/Core.hs
```diff
@@ -750,44 +765,24 @@ getSsoidViaSelf uid = maybe (error "not found") pure =<< getSsoidViaSelf' uid
...
-    (\_ -> pure . isNothing)
-    (\_ -> runClient (env ^. teCql) $ Data.getSAMLUser uref)
+  eventually (view teCql >>= \cql -> runClient cql $ Data.getSAMLUser uref) isNothing
+
+-- | FUTUREWORK: arguably this function should move to Util.Scim, but it also is related to
+-- the other lookups above into the various user tables in the various cassandras.  we should
+-- probably clean this up a little, and also pick better names for everything.
+getScimUser :: HasCallStack => UserId -> TestSpar (Maybe (ScimC.User.StoredUser SparTag))
+getScimUser uid = do
+  eventually (view teCql >>= \cql -> runClient cql $ Data.getScimUser uid) isNothing
```
neongreen: Why is `Nothing` considered good here?

services/spar/test-integration/Test/Spar/DataSpec.hs
neongreen: Oh wait, so `eventually` can return even when `good` has not been achieved?

That's unexpected and I think somebody will bump into this later on.

neongreen:
Tests in this PR make sense.

I'm curious as to why they actually fail 🙂

services/spar/src/Spar/Scim/Types.hs
```diff
@@ -56,6 +59,59 @@ userExtraURN = "urn:wire:scim:schemas:profile:1.0"
...
+
+instance Scim.Group.GroupTypes SparTag where
+  type GroupId SparTag = ()
+
+instance Scim.Auth.AuthTypes SparTag where
+  type AuthData SparTag = ScimToken
+  type AuthInfo SparTag = ScimTokenInfo
+
+
+-- | Wrapper to work around complications with type synonym family application in instances.
```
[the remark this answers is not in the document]
fisx: they are, in the tests.

services/spar/src/Spar/Scim/User.hs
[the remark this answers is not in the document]
fisx: > When this PR is merged, please add "SCIM users are written to the database before even checking things like 'handle is not claimed'" to the list of SCIM problems.

I would rather fix this here.  What other things like "handle is not claimed" exist?  Could you make a reasonably complete list?

services/spar/test-integration/Spec.hs
fisx: (appearently not in 2.5.5, though.  and i like my solution, too.  :))

services/spar/test-integration/Test/Spar/DataSpec.hs
fisx: Rename it to `aFewTimes`, and made haddocks more explicit.

services/spar/test-integration/Util/Core.hs
fisx: i set it to 1000.

services/spar/test-integration/Util/Core.hs
fisx: I'll improve the docs again.  This seems to be a very confusion helper.  :)

fisx: actually, it was just wrong.  should be fixed now?

neongreen:
> I have a hunch that if you claim a handle and then delete the user, the claim still lingers for 30 minutes. This hunch is based on the fact that I haven't found any code anywhere that deletes anything from the unique_claims table.
>
> If it's so, can I change it so that it deletes the claim when the handle is freed? @tiago-loureiro

fisx:
> > I have a hunch that if you claim a handle and then delete the user, the claim still lingers for 30 minutes. This hunch is based on the fact that I haven't found any code anywhere that deletes anything from the unique_claims table.
> > If it's so, can I change it so that it deletes the claim when the handle is freed? @tiago-loureiro

further support for your hunch comes from the fact that the test case fails [here](https://github.com/wireapp/wire-server/blob/aaa5b5f68a21577e55f365b6b5c3b7988122f3d5/services/spar/src/Spar/Scim/User.hs#L259), and i can't think of any reason why it should fail other than that brig keeps the handle locked after deletion.

half-locked, that is: [checkHandle](https://github.com/wireapp/wire-server/blob/aaa5b5f68a21577e55f365b6b5c3b7988122f3d5/services/spar/src/Spar/Scim/User.hs#L462) happily reports that the handle is avaiable, but [claiming the handle](https://github.com/wireapp/wire-server/blob/aaa5b5f68a21577e55f365b6b5c3b7988122f3d5/services/spar/src/Spar/Scim/User.hs#L259) fails a few microseconds later.  [here is the test case again](https://github.com/wireapp/wire-server/blob/aaa5b5f68a21577e55f365b6b5c3b7988122f3d5/services/spar/test-integration/Test/Spar/Scim/UserSpec.hs#L561-L570).

i will take a shot at fixing brig now.

neongreen:
If it's not WIP anymore I will re-review it now.

services/spar/src/Spar/Scim/User.hs
```diff
@@ -315,29 +283,36 @@ updateValidScimUser tokinfo uidText newScimUser = do
...
       else do
-        newScimStoredUser :: Scim.StoredUser ScimUserExtra
+        newScimStoredUser :: Scim.StoredUser SparTag
           <- lift $ updScimStoredUser (newScimUser ^. vsuUser) oldScimStoredUser
 
-        -- update 'SAML.UserRef'
-        let uref = newScimUser ^. vsuSAMLUserRef
-        lift . wrapMonadClient $ Data.insertSAMLUser uref uid  -- on spar
-        bindok <- lift $ Intra.Brig.bindBrigUser uid uref  -- on brig
+        -- update 'SAML.UserRef' on spar
```
neongreen:
```suggestion
        -- update 'SAML.UserRef' on spar and delete the old 'SAML.UserRef'
```

services/spar/src/Spar/Scim/User.hs
```diff
@@ -315,29 +283,36 @@ updateValidScimUser tokinfo uidText newScimUser = do
...
+        let newuref = newScimUser ^. vsuSAMLUserRef
+        lift . wrapMonadClient $ Data.insertSAMLUser newuref uid
+        midp :: Maybe IdP <- maybe (pure Nothing) (lift . wrapMonadClient . Data.getIdPConfig) stiIdP
+        forM_ midp $ \idp -> do
+          let eid = Scim.externalId . Scim.value . Scim.thing $ oldScimStoredUser
+          olduref <- mkUserRef idp eid
+          lift . wrapMonadClient $ Data.deleteSAMLUser olduref
```
neongreen: If `olduref == newuref`, we're going to insert it and immediately delete it, which looks like a bug. If I'm not mistaken and it's indeed a bug, let's add a test for it.

services/spar/src/Spar/Scim/User.hs
```diff
@@ -315,29 +283,36 @@ updateValidScimUser tokinfo uidText newScimUser = do
...
+
+        -- update 'SAML.UserRef' on spar
```
neongreen:
```suggestion
        -- update 'SAML.UserRef' on brig
```

services/spar/src/Spar/Scim/User.hs
```diff
@@ -425,30 +426,51 @@ parseUid uidText = maybe err pure $ readMaybe (Text.unpack uidText)
...
+-}
+assertUserRefUnused :: SAML.UserRef -> Scim.ScimHandler Spar ()
+assertUserRefUnused userRef = do
+  mExistingUserId <- lift $ wrapMonadClient (Data.getSAMLUser userRef)
+  unless (isNothing mExistingUserId) $
+    throwError Scim.conflict {Scim.detail = Just "externalId is already taken"}
 
 {-|
-Check that the UserRef is not taken; or that it's taken by the given user id.
+Check that the UserRef is not taken *by another user*.
```
neongreen: I'm worried that it's easy to take "another user" to mean "the passed UserId".

```suggestion
Check that the 'UserRef' is not taken by /any/ user other than the passed 'UserId'
(it is also acceptable if it is not taken by anybody).
```

services/spar/src/Spar/Scim/User.hs
```diff
@@ -425,30 +426,51 @@ parseUid uidText = maybe err pure $ readMaybe (Text.unpack uidText)
...
-    Nothing -> return ()
-    -- A user exists; verify that it's the same user before updating
-    Just existingUserId  ->
-      unless (existingUserId == wireUserId) $
-        throwError Scim.conflict {Scim.detail = Just "externalId is already taken"}
+  unless (mExistingUserId `elem` [Nothing, Just wireUserId]) $ do
+    throwError Scim.conflict {Scim.detail = Just "externalId does not match UserId"}
+
+assertHandleUnused :: Handle -> UserId -> Scim.ScimHandler Spar ()
+assertHandleUnused = assertHandleUnused' "externalId is already taken"
```
neongreen:
```suggestion
assertHandleUnused = assertHandleUnused' "userName is already taken"
```

services/spar/src/Spar/Scim/User.hs
```diff
@@ -425,30 +426,51 @@ parseUid uidText = maybe err pure $ readMaybe (Text.unpack uidText)
...
+assertHandleUnused' :: Text -> Handle -> UserId -> Scim.ScimHandler Spar ()
+assertHandleUnused' msg hndl uid = lift (Brig.checkHandleAvailable hndl uid) >>= \case
+  True  -> pure ()
+  False -> throwError Scim.conflict {Scim.detail = Just msg}
+
+assertHandleNotUsedElsewhere :: Handle -> UserId -> Scim.ScimHandler Spar ()
+assertHandleNotUsedElsewhere hndl uid = do
+  musr <- lift $ Brig.getBrigUser uid
+  unless ((userHandle =<< musr) == Just hndl) $
+    assertHandleUnused' "externalId does not match UserId" hndl uid
```
neongreen:
```suggestion
    assertHandleUnused' "userName does not match UserId" hndl uid
```

services/spar/test-integration/Util/Core.hs
```diff
@@ -205,6 +209,17 @@ pendingWith :: (HasCallStack, MonadIO m) => String -> m ()
 pendingWith = liftIO . Test.Hspec.pendingWith
 
 
+-- | Run a probe several times, until a "good" value materializes or until patience runs out.
+-- The result may be good or not, depending on whether we run out of time.
```
neongreen: Tbh this function is still unintuitive enough for me that I feel it would be worth it to be more explicit. Here's my attempt:

```suggestion
-- If all retries were unsuccessful, 'aFewTimes' will return the last obtained value, even
-- if it does not satisfy the predicate.
```

services/brig/src/Brig/User/Handle.hs
```diff
@@ -32,13 +32,16 @@ claimHandle u h = do
                     -- Update profile
                     User.updateHandle (userId u) h
                     -- Free old handle (if it changed)
-                    for_ (mfilter (/= h) (userHandle u))
-                        freeHandle
+                    for_ (mfilter (/= h) (userHandle u)) $
+                        freeHandle u
             return (isJust claimed)
 
 -- | Free a 'Handle', making it available to be claimed again.
```
neongreen: So the comment was lying :tsk:

services/spar/test-integration/Test/Spar/Scim/UserSpec.hs
```diff
@@ -216,6 +221,54 @@ testRichInfo = do
     liftIO $ scimUserId scimStoredUser' `shouldBe` scimUserId scimStoredUser
     probeUser (scimUserId scimStoredUser) richInfo'
 
+-- | Create a user implicitly via saml login; remove it via brig leaving a dangling entry in
+-- @spar.user@; create it via scim.  This should to work despite the dangling database entry.
```
neongreen:
```suggestion
-- @spar.user@; create it via scim.  This should work despite the dangling database entry.
```

services/spar/test-integration/Test/Spar/Scim/UserSpec.hs
```diff
@@ -55,7 +57,7 @@ specCreateUser = describe "POST /Users" $ do
         testCreateSameExternalIds
     it "provides a correct location in the 'meta' field" $ testLocation
     it "handles rich info correctly (this also tests put, get)" $ testRichInfo
-    it "gives created user a valid 'SAML.UserRef' for SSO" $ pending
+    it "gives created user a valid 'SAML.UserRef' for SSO" $ testScimCreateVsUserRef
```
neongreen: The SCIM todo list should be updated with this test.

services/spar/test-integration/Test/Spar/Scim/UserSpec.hs
```diff
@@ -219,6 +221,54 @@ testRichInfo = do
...
+testScimCreateVsUserRef = do
+    (tok, (_ownerid, _teamid, idp)) <- registerIdPAndScimToken
+    (usr, uname) :: (Scim.User.User SparTag, SAML.UnqualifiedNameID)
+        <- randomScimUserWithSubject
+    let uref   = SAML.UserRef tenant subj
+        subj   = either (error . show) id $ SAML.mkNameID uname Nothing Nothing Nothing
+        tenant = idp ^. SAML.idpMetadata . SAML.edIssuer
+
+    uid <- createViaSaml idp uref
+    samlUserShouldSatisfy 'a' uref isJust
```
neongreen: Is there a reason for a char instead of `HasCallStack` on the internal function?

services/spar/test-integration/Test/Spar/Scim/UserSpec.hs
```diff
@@ -216,6 +221,54 @@ testRichInfo = do
...
+
+    deleteViaBrig uid
+    samlUserShouldSatisfy 'b' uref isJust  -- brig doesn't talk to spar right now when users
+                                           -- are deleted there.  we need to work around this
+                                           -- fact for now.  (if the test fails here, this may
+                                           -- mean that you fixed the behavior and can
+                                           -- simplify this test.)
```
neongreen:
```suggestion
                                           -- change this to isNothing.)
```

services/spar/src/Spar/Scim/User.hs
```diff
@@ -442,7 +442,7 @@ to a single `externalId`.
 -}
 assertUserRefUnused :: SAML.UserRef -> Scim.ScimHandler Spar ()
 assertUserRefUnused userRef = do
-  mExistingUserId <- lift $ wrapMonadClient (Data.getSAMLUser userRef)
+  mExistingUserId <- lift $ getUser userRef
```
neongreen: `getUser` only reports conflicts in the same team.

This might work for now if we have a rule that each IdP can only be "used" by one team, but it seems fragile and I don't think anyone will remember to fix it if we lift that rule somehow.

(To be honest I'm not even 100% sure we have that rule, off the top of my head.)

fisx: if we add the same idp to two teams, implicit user creation breaks immediately: we won't know any more where to create the user.  and i don't expect implicit user creation via saml to go away ever.

more importantly: why do you think cross-team conflicts are not reported?y  i think they are.

but if you know of a good place where i can add a comment that idps can only live in one team, i'd like to do that.

services/brig/src/Brig/Unique.hs
```diff
@@ -50,6 +51,18 @@ withClaim u v t io = do
     cql :: PrepQuery W (Int32, C.Set (Id a), Text) ()
     cql = "UPDATE unique_claims USING TTL ? SET claims = claims + ? WHERE value = ?"
 
+deleteClaim :: MonadClient m
+    => Id a
+    -> Text
+    -> Timeout
```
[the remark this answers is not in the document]
fisx:
```haskell
-- | Each claim can have more than one claimer (even though this is a feature we never use),
-- so removing a claim is an update operation on the database.  Therefore, we reset the TTL
-- the same way we reset it in 'withClaim'.
```

services/brig/test/integration/API/User/Handles.hs
```diff
@@ -85,14 +85,17 @@ testHandleUpdate brig cannon = do
...
-    -- Other users cannot immediately claim the old handle since the previous claim
-    -- is still active.
+    -- Other users can immediately claim the old handle (the claim is removed with the old
+    -- handle).
     put (brig . path "/self/handle" . contentJson . zUser uid2 . zConn "c" . body update) !!! do
-        const 409 === statusCode
-        const (Just "handle-exists") === fmap Error.label . decodeBody
+        const 200 === statusCode
 
     -- The old handle can be claimed again immediately by the user who previously
```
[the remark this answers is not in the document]
fisx: This last test was actually meaningless the way it was put.  I have comments now and a fix now.

services/spar/src/Spar/Intra/Brig.hs
```diff
@@ -254,6 +255,23 @@ setBrigUserRichInfo buid richInfo = do
      | otherwise
        -> throwSpar . SparBrigError . cs $ "set richInfo failed with status " <> show sCode
 
+checkHandleAvailable :: (HasCallStack, MonadSparToBrig m) => Handle -> UserId -> m Bool
```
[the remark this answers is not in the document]
fisx: Arguably this should be in the documentation of the brig end-point, but ok:

```haskell
-- | At the time of writing this, @HEAD /users/handles/:uid@ does not use the 'UserId' for
-- anything but authorization.
```

services/spar/src/Spar/Scim/User.hs
```diff
@@ -261,35 +203,61 @@ validateScimUser' idp richInfoLimit user = do
...
+                unspec = SAML.mkUNameIDUnspecified txt
+            pure . either (const unspec) id $ eEmail
+        case SAML.mkNameID unameId Nothing Nothing Nothing of
+            Right nameId -> pure nameId
+            Left err -> throwError $ Scim.ScimError
+                mempty
+                (Scim.Status 400)
+                Nothing
+                (Just $ "Can't construct a subject ID from externalId: " <> Text.pack err)
+                -- This cannot happen at the time of writing this comment, but there may be
```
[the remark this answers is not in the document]
fisx: `mkNameID` does not throw errors for any of the `UnqualifiedNameID` constructors we use.  But this comment seems to me more confusing than helpful now, so I removed it.

services/spar/src/Spar/Scim/User.hs
fisx: yes, seems to me it's a bug.  i'll add a test case, thanks!

services/spar/test-integration/Test/Spar/Scim/UserSpec.hs
fisx: Couldn't get `HasCallStack` to do the trick.  I never got to the code line that I was interested in.  (Wanna try?)

fisx: (I'll resolve this since I don't think it's important enough, but feel free to re-open.)

services/spar/src/Spar/Scim/User.hs
fisx: hum, the enhanced test from 42edf8b already passes without the subsequent fix.  both prod code and test seem fine to me *now*, but it puzzles me that the old prod code doen't make the test fail.  i will take another look tomorrow.

fisx: the test was still wrong.  with 0e04f99, the bug you found can be reproduced, and the tests pass with the fix.  all good now, i think!

fisx:
@tiago-loureiro  could you take a look at the (very few) changes in brig please?

An alternative to what I did would be to change the test in spar to claim everything from the old user *but* the handle.  But unless we can think of a reason for brig's old behavior I'd rather not.  Note that [some pre-existing code](https://github.com/wireapp/wire-server/pull/698#discussion_r272664559) was also mislead by it.

services/brig/src/Brig/Unique.hs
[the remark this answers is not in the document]
neongreen: Ok, I have to say that I still don't know how exactly it works, but you seem to do 🙂

services/spar/test-integration/Test/Spar/Scim/UserSpec.hs
neongreen: > (Wanna try?)

Nah.

services/spar/src/Spar/Scim/User.hs
neongreen: > why do you think cross-team conflicts are not reported

Oh, this is interesting. Even though the `getUser` method is defined in `Spar.Scim.User`, it's not in scope – instead `getUser` from `Spar.App` is used.

I'm marking this as resolved.

PR #706, opened 2019-04-08
Do not autoprovision saml users if SCIM exists.

fisx:


fisx:
ready for review!

neongreen:
I'll look at this one now.

services/spar/src/Spar/App.hs
```diff
@@ -159,20 +162,39 @@ getUser uref = do
...
+
+-- | Like 'autoprovisionSamlUser', but for an already existing 'UserId'.
+autoprovisionSamlUserWithId :: UserId -> SAML.UserRef -> Maybe Name -> ManagedBy -> Spar ()
+autoprovisionSamlUserWithId buid suid mbName managedBy = do
+  teamid <- (^. idpExtraInfo) <$> getIdPConfigByIssuer (suid ^. uidTenant)
+  scimtoks <- wrapMonadClient $ Data.getScimTokens teamid
+  if null scimtoks
+    then createSamlUserWithId buid suid mbName managedBy
+    else throwError . SAML.Forbidden $
+            "bad credentials (note that your team has uses SCIM, " <>
```
neongreen: "bad credentials" might mislead someone to think that they actually have wrong credentials 🤔

I'd say this is a clear-cut case of "bad request".

fisx: I disagree.  it's arguably an error in the IdP setup because they send people our way that they should have created first, but it's a perfectly valid request for all we know.  Just a user we haven't seen before.

Does this make sense?  If not feel free to rename it in a separate PR.

PR #707, opened 2019-04-08
Fix HasCallStack usage

neongreen:
I couldn't resist.

Hspec only gives us the first entry of the call stack -- for some weird reason -- so having a HasCallStack constraint on the test itself breaks reporting. The 'Char' trick is still useful, just not in this case.


fisx:
thanks!

PR #713, opened 2019-04-11
Makefile cleanups.

julialongtin:


Makefile
```diff
@@ -50,6 +51,7 @@ i:
 	$(MAKE) -C services/brig i
 	$(MAKE) -C services/gundeck i
 	$(MAKE) -C services/spar i
+	$(MAKE) -C libs/cassandra-util i
```
fisx: hum, this is not a service.  why do you want to build it here?

julialongtin:
sorry, this snuck in.

services/restund/Makefile
```diff
@@ -18,6 +18,8 @@ WIRE_VERSION    := 6
 EXTRA_MODULES   := zrest drain
 
 DOCKER          ?= true
+# FIXME: use quay.io/wire
```
ChrisPenner: ☝️

julialongtin: The other Makefiles use quay.io/wire . This Makefile is an outlier. I wanted to make it clear something was strange, while making a minor improvement by allowing us to change the username, and test.

PR #725, opened 2019-04-23
Update nginx to latest stable

jschaul:


jschaul:
Actually, perhaps we should follow the upstream changes from these commits for the docker build:

- [this](https://github.com/nginxinc/docker-nginx/commit/0501f0b73fd7fbfb139d1e98a7fd3b63426b16a1#diff-b9fc03e94d38e24c8cf233e6e4aee610) 
- [this](https://github.com/nginxinc/docker-nginx/commit/9a052e07b2c283df9960375ee40be50c5c462a7e#diff-b9fc03e94d38e24c8cf233e6e4aee610)
- [update alpine](https://github.com/nginxinc/docker-nginx/commit/2364fdc54af554d28ef95b7be381677d10987986#diff-b9fc03e94d38e24c8cf233e6e4aee610)

jschaul:
Updated this PR, can be re-reviewed.

I tried to incorporate the changes from the upstream docker image building; however that does not work to compile the `nginx-zauth-module` and `nginx-module-vts` modules that are needed. Even if in theory they can be dynamically loaded (which also requires us to change config files), they need to still be compiled and available on the filesystem as .so files. The [vts module appears unmaintained](https://github.com/vozlt/nginx-module-vts) and doesn't provide details on usage with nginx 1.16 and dynamic loading.
Switching to dynamic modules is a harder task than I anticipated, and should be done on a separate PR.

services/nginz/Dockerfile
```diff
@@ -8,7 +8,7 @@ RUN cd /src/libzauth/libzauth-c \
     && make install
 
 # Nginz container
-FROM alpine:3.11
+FROM alpine:3.10
```
arianvp: What does "not supported" mean? Perhaps leave a comment for the next person who feels like bumping this number? What went wrong?

services/nginz/Dockerfile
```diff
@@ -1,5 +1,5 @@
 # Requires docker >= 17.05 (requires support for multi-stage builds)
-FROM alpine:3.10 as libzauth-builder
+FROM alpine:3.11 as libzauth-builder
```
arianvp: Nevermind the change got reverted here again. I was reading this from commit to commit. Ignore previous comment :)

PR #734, opened 2019-04-26
[WIP] How SAML user errors are displayed.

fisx:
We currently have some discrepancies between how the backend reports login errors in SAML and what the client expects, plus at least one minor glitch in the backend.

This is an attempt to resolve these issues.


fisx:
related parts of frontend code:

- https://github.com/wireapp/wire-web-config-default/blob/master/wire-webapp/content/translation/en-US.json#L31
- https://github.com/wireapp/wire-webapp/blob/dev/src/script/auth/module/action/BackendError.ts#L92
- https://github.com/wireapp/wire-webapp/blob/dev/src/script/auth/page/SingleSignOn.tsx#L99

fisx:
see also: https://github.com/wireapp/wire-webapp/pull/6542

PR #735, opened 2019-04-26
Fix 'stack --profile'

neongreen:


neongreen:
If you try to build wire-server with profiling before this PR, it will bark
at auto-all. It's been deprecated and renamed a long time ago.

PR #749, opened 2019-05-08
Cannon analytics (1)

fisx:
Hopefully provide more details in log files on cannon's mysterious 5xx metrics peaks.

PR #750, opened 2019-05-08
Cannon analytics (2)

fisx:
Second episode after the popular pilot #749

If some application (or inner middleware) responds with status >=500 rather than throwing an exception, turn that into an actual exception so `catchErrors` can do its thing.

### DIFF
--- a/libs/wai-utilities/src/Network/Wai/Utilities/Server.hs
+++ b/libs/wai-utilities/src/Network/Wai/Utilities/Server.hs
@@ -15,6 +15,7 @@ module Network.Wai.Utilities.Server
     , catchErrors
     , OnErrorMetrics
     , heavyDebugLogging
+    , rethrow5xx
     , lazyResponseBody
 
       -- * Utilities
@@ -185,7 +186,7 @@ measureRequests m rtree = withPathTemplate rtree $ \p ->
 -- See 'catchErrors'.
 catchErrors :: Logger -> OnErrorMetrics -> Middleware
 catchErrors l m app req k =
-    app req k `catch` errorResponse
+    rethrow5xx app req k `catch` errorResponse
   where
     errorResponse :: SomeException -> IO ResponseReceived
     errorResponse ex = do
@@ -271,6 +272,19 @@ emitLByteString lbs = do
     tvar <- newTVarIO (cs lbs)
     -- | Emit the bytestring on the first read, then always return "" on subsequent reads
     return . atomically $ swapTVar tvar mempty
+
+-- | Run the 'Applicatino'; check the response status; if >=500, throw a 'Wai.Error' with
+-- label @"server-error"@ and the body as the error message.
+rethrow5xx :: Middleware
+rethrow5xx app req k = app req k'
+  where
+    k' resp = do
+        let st = responseStatus resp
+        if statusCode st < 500
+            then k resp
+            else do
+                rsbody :: LText <- liftIO $ cs <$> lazyResponseBody resp
+                throwM $ Wai.Error st "server-error" rsbody
 
 -- | This flushes the response!  If you want to keep using the response, you need to construct
 -- a new one with a fresh body stream.
--- a/libs/wai-utilities/src/Network/Wai/Utilities/Server.hs
+++ b/libs/wai-utilities/src/Network/Wai/Utilities/Server.hs
@@ -50,14 +50,15 @@ import Network.Wai.Utilities.Response
 import System.Logger.Class hiding (Settings, Error, format)
 import System.Posix.Signals (installHandler, sigINT, sigTERM)
 
-import qualified Network.Wai.Utilities.Error as Wai
 import qualified Data.ByteString             as BS
 import qualified Data.ByteString.Char8       as C
 import qualified Data.ByteString.Lazy        as LBS
 import qualified Data.Text.Lazy.Encoding     as LT
+import qualified Network.Wai.Internal        as WaiInt
 import qualified Network.Wai.Predicate       as P
 import qualified Network.Wai.Routing.Route   as Route
 import qualified Network.Wai.Utilities.Error as Error
+import qualified Network.Wai.Utilities.Error as Wai
 import qualified Prometheus                  as Prm
 import qualified System.Logger               as Log
 import qualified System.Posix.Signals        as Sig
@@ -186,7 +187,7 @@ measureRequests m rtree = withPathTemplate rtree $ \p ->
 -- See 'catchErrors'.
 catchErrors :: Logger -> OnErrorMetrics -> Middleware
 catchErrors l m app req k =
-    rethrow5xx app req k `catch` errorResponse
+    rethrow5xx l app req k `catch` errorResponse
   where
     errorResponse :: SomeException -> IO ResponseReceived
     errorResponse ex = do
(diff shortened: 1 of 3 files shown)


fisx:
If I go back to 0f3e8a184cee9430c8a8cc9412a274306fdb806b and change [this line](https://github.com/wireapp/wire-server/blob/0f3e8a184cee9430c8a8cc9412a274306fdb806b/services/cannon/src/Cannon/API.hs#L191) thusly:

```haskell
        Just rs -> traceShow (responseStatus rs) $ return rs
```

I get the error as well:

```
Status {statusCode = 500, statusMessage = "Internal Server Error"}
```

Which supports the following **hypothesis**: `wai-websockets` is responding with 500 in [`websocketsApp`](https://hoogle.zinfra.io/file/root/.stack/snapshots/x86_64-linux/custom-wire-1.1-G6NVtOVxqIB2/8.4.3/doc/wai-websockets-3.0.1.2/Network-Wai-Handler-WebSockets.html#v:websocketsApp), and always has.  So far, this has not bothered anybody, and the old metrics code has not caught this because it could only see exceptions thrown inside the `Application`, not 500 responses returned as an ordinary value.  Our clients always throw away the response, because they have a working websocket and are happy.

I have looked for code that uses `wai-websockets` correctly and looked at `servant-websockets` for a little, but that's out of maintenance and needs some love before it will compile again.

I **propose** to throw away the response in [this line](https://github.com/wireapp/wire-server/blob/0f3e8a184cee9430c8a8cc9412a274306fdb806b/services/cannon/src/Cannon/API.hs#L191) and construct one with status 2xx; leave a comment, and forget about this issue.

fisx:
Nope, the real story is subtly but importantly different: `ResponseRaw` carries two values: an action that can be called to construct a `Response`, and a pure backup `Response` (not sure under what circumstances that is used, perhaps if the action throws an exception?).  Since `responseStatus` is pure and it makes no sense for it to call the action, it just goes to the backup response and pulls the value from that.  In the `/await` handler, this is a 500 from `wai-websockets` that is meant to only emerge if the websocket upgrade fails badly.

```haskell
> responseStatus $ ResponseRaw undefined 
    (ResponseBuilder (Status 500 "blargh") undefined undefined)
Status {statusCode = 500, statusMessage = "blargh"}
```

Clearly, this is not a good default behaviour, but it's hard to say what is.

We need to fix this in two places now: the prometheus middleware and our own `catchAll`.  Both should just ignore raw requests until we can think of something better.

More commits coming up!

fisx:
https://github.com/yesodweb/wai/issues/265#issuecomment-46533699  (thanks @jschaul)

fisx:
or, much closer to home: https://github.com/wireapp/wire-server/blob/a6f16fa19a5d1488a76b5ec3728f828a77faa942/libs/metrics-wai/src/Data/Metrics/Middleware.hs#L46-L52

fisx:
@jschaul the 5xx reported by the prometheus middleware are harmless false positives (the middleware looks in the wrong place, and the actual response is what it should be).  I will fix that in a separate PR.

This here is ready for re-review.

limengyu1990:
[cannon] E, request=9547e08224e6cf973e856144d9d6bb2c, Error {code = Status {statusCode = 500, statusMessage = "Internal Server Error"}, label = "server-error", message = "The web application attempted to send a WebSockets response, but WebSockets are not supported by your WAI handler."}

fisx:
> [cannon] E, request=9547e08224e6cf973e856144d9d6bb2c, Error {code = Status {statusCode = 500, statusMessage = "Internal Server Error"}, label = "server-error", message = "The web application attempted to send a WebSockets response, but WebSockets are not supported by your WAI handler."}

yes, this problem has been fixed in #751.  i just merged this into the branch here, could you try again?

fisx:
I think instead of constructing and throwing an `Wai.Error`, we can call `onError` from within `rethrow5xx` directly (which would suggest a name change of `rethrow5xx`).  I'll think about that as soon as I get back to this.

fisx:
> I think instead of constructing and throwing an `Wai.Error`, we can call `onError` from within `rethrow5xx` directly (which would suggest a name change of `rethrow5xx`). I'll think about that as soon as I get back to this.

changed my mind on this: logging the errors requires us to empty the streamed response body, and instead of constructing a new response we may as well throw the error.

I am quite confident this is a sound change by now, and perhaps we can get this merged soonish so I can stop thinking about it and keeping it up to date?

It'll improve logging and metrics!  :)

services/cannon/src/Cannon/App.hs
```diff
@@ -138,7 +137,7 @@ rejectOnError p x = do
         NotSupported              -> rejectRequest p (f "protocol not supported" "N/A")
         MalformedRequest _ m      -> rejectRequest p (f "malformed-request" (Text.pack m))
         OtherHandshakeException m -> rejectRequest p (f "other-error" (Text.pack m))
-        _                         -> throwM x
+        _                         -> pure ()
```
tiago-loureiro: So we now just ignore other errors instead of rethrowing the `HandshakeException`?

services/cannon/src/Cannon/App.hs
```diff
@@ -65,11 +65,10 @@ continue l ws clock k = liftIO $ do
     case result of
         (Left  (Left x)) ->
             let text = client (key2bytes k) . msg (val "read: " +++ show x) in
-            case fromException x of
-                Just ConnectionClosed -> Logger.debug l text
-                _                     -> Logger.warn  l text
-        (Right (Left x)) -> Logger.warn l $
-            client (key2bytes k) . msg (val "write: " +++ show x)
+            Logger.debug l text
```
tiago-loureiro: How come no more warning but only at the debug level?

services/cannon/src/Cannon/App.hs
fisx: no, this is not a behavioral change; the `throwM` is in the line below.

libs/wai-utilities/src/Network/Wai/Utilities/Server.hs
```diff
@@ -271,6 +274,38 @@ emitLByteString lbs = do
     -- | Emit the bytestring on the first read, then always return "" on subsequent reads
     return . atomically $ swapTVar tvar mempty
 
+-- | Run the 'Applicatino'; check the response status; if >=500, throw a 'Wai.Error' with
```
tiago-loureiro: `Applicatino` 🤓

services/cannon/src/Cannon/App.hs
fisx: there were a lot of warnings about clients closing web sockets untidily.  i think joe and i agreed that they are not very interesting, and just hanging up on a websocket is both legit and unlikely to go away no matter what our moral position on them is.

libs/wai-utilities/src/Network/Wai/Utilities/Server.hs
fisx: heh.  let's make a bunch of PRs and fix that spelling error in wai/warp!  :-)

services/cannon/src/Cannon/App.hs
tiago-loureiro: Ah right; missed it 👍

fisx:
quick test:

```
diff --git a/services/brig/src/Brig/API.hs b/services/brig/src/Brig/API.hs
index 5d5fe493..4908c8dd 100644
--- a/services/brig/src/Brig/API.hs
+++ b/services/brig/src/Brig/API.hs
@@ -67,7 +67,7 @@ sitemap o = do

     -- Internal ---------------------------------------------------------------

-    get  "/i/status" (continue $ const $ return empty) true
+    get  "/i/status" (continue $ const $ return $ responseLBS status500 [] mempty) true
     head "/i/status" (continue $ const $ return empty) true

     get "/i/monitoring" (continue $ const $ view metrics >>= fmap json . render) $
```

Then run `/services/start-services-only.sh`.  The script itself calls this end-point, and you can see that it gets logged before doing anything else:

```
[...]
[brig] E, request=N/A, Error {code = Status {statusCode = 500, statusMessage = "Internal Server Error"}, label = "server-error", message = ""}
[...]
```

The prometheus middleware also sees it:

```
curl http://localhost:8082/i/metrics | grep 500
[...]
 http_request_duration_seconds_bucket{handler="/i/status",method="GET",status_code="500",le="0.01"} 42
[...]
```

PR #754, opened 2019-05-10
Fix prometheus mutation.

fisx:
it turns out cloning functions that are marked NOINLINE and contain a top-level call to unsafePerformIO is changing behavior of your code.  D'OH!


fisx:
Tested this change and the duplicate entry is gone:

```
$ curl http://localhost:8082/i/metrics | grep -i help | sort
# HELP gc_bytes_allocated_total Bytes allocated since the start of the server
# HELP gc_bytes_used_current Number of bytes in active use as of the last GC
# HELP gc_bytes_used_max Maximum amount of memory living on the heap after the last major GC
# HELP gc_seconds_cpu CPU time spent on last GC
# HELP gc_seconds_wall Wall clock time spent on last GC
# HELP http_request_duration_seconds The HTTP request latencies in seconds.
# HELP net_connections description not provided
# HELP net_errors description not provided
# HELP net_requests description not provided
# HELP net_resources_i_status_get_status_200 description not provided
# HELP net_resources_i_status_get_time description not provided
```

PR #757, opened 2019-05-13
[WIP] Backoffice frontend build docker

tiago-loureiro:


fisx:
@tiago-loureiro elsewhere:

> Also we need to make sure that this works on k8s too.

i'm not sure about that.  if it works with the make rules `run-develop` and `run-demo` that I am about to add, it's a coherent, self-contained improvement.  Making it work in other setups can be done in a separate PR, once we need it?

fisx:
@tiago-loureiro could you review it, then I'll approve and merge.  github wants that.

tools/backoffice-frontend/var/www/swagger-ui/lib/call-trace.js
```diff
@@ -0,0 +1,66 @@
+function findSession() {
```
tiago-loureiro: This is no longer used and can be removed. Is this referenced anywhere? If so, we can remove it.

fisx:
new plan: stern should serve the frontend itself from a new end-point.  this will make most of this PR obsolete (no need to a separate nginx image); the stern image from quay.io can be used instead of the composite docker image constructed here, and developers can just run the stern executable to play with the backoffice.

This will also simplify hegemony terminator (external readers: that's internal ops stuff for the wire cloud setup, sorry): we don't need to serve a frontend from there, and we don't need to keep the frontend there in sync with the one everywhere else.

Question: does it make sense to use [servant-swagger-ui](https://hackage.haskell.org/package/servant-swagger-ui) even if we don't use servant for anything else?  @arianvp there is no package swagger-ui based on wai, or at least i couldnt' find one.

tools/backoffice-frontend/var/www/swagger-ui/lib/call-trace.js
fisx: (let's see if we can remove the whole thing, though...)

PR #760, opened 2019-05-16
Add upstream references in stack.yaml deps (wai-middleware-prometheus).

fisx:


stack.yaml
```diff
@@ -38,8 +38,8 @@ packages:
 - tools/db/auto-whitelist
 
 extra-deps:
-- git: https://github.com/wireapp/prometheus-haskell
-  commit: 9bcd5423e0e33a6db8b1e69b50d74e8b47daa737
+- git: https://github.com/wireapp/prometheus-haskell  # https://github.com/fimad/prometheus-haskell/pull/44
+  commit: 9bcd5423e0e33a6db8b1e69b50d74e8b47daa737    # (related: https://github.com/fimad/prometheus-haskell/pull/45)
```
jschaul: why not use the latest commit `a804ecdec6a7d704fd1aefafda92fdb0b8fe1d9c` where PR 44 is merged?

fisx: because the tip of that branch is younger than this change, and there is no benefit to touching it again, not even in terms of clarity, and this is just a temporary change until https://github.com/fimad/prometheus-haskell/pull/45 gets merged and then we'll have to touch this again anyway, and i was trying to save time.  :)

should i change it anyway?

jschaul: This PR was, and still is confusing to me, as the comment references a PR, which, at the hash of that commit (9bcd...) , isn't merged yet (pr 44). 

>  i was trying to save time. :)

If this is "temporary" anyway, and there is no actual change (just a confusing comment which points to code that has not been merged yet, so it's incorrect documentation in a way), how about you open a new PR when the stuff you want to have merged is actually merged?

fisx: it shouldn't be confusing, but i also don't want us to keep track of our deps in my head.  i've tried again, is it less confusing now?

(i don't want to change the commit hash because it is completely unnecessary and it means more rebuilding.)

fisx:
thanks for being so stubborn!  i was wining about it, but it was very valid criticism.  :-)

PR #761, opened 2019-05-20
Feature/legalhold tokens

jschaul:
(see also note in https://github.com/wireapp/wire-server/commit/2483ac124d9d6bfa6dc31d64b5bcad088e91128b)

Adds new token types to:

* [x] zauth
* [x] libzauth
* [x] nginx-auth-module

token generation and storage in brig

* [x] implement legal hold user/access token login `/i/legalhold/login`
*  [x] implement legal hold token refresh `/legalhold/access`
*  [x] throw error in the edge case where `/i/legalhold/login` is used to create a session legal hold token (not a possible code path in the current implementation but possible type wise - see `TODO` `undefined` in current changeset
*  [x] integration test in legal hold tests: legal hold login leads to a `lu` cookie and a `la` access token.
*  [x] integration tests: legal hold login cookie can be refreshed and revoked (many of the same tests as done [here](https://github.com/wireapp/wire-server/blob/develop/services/brig/test/integration/API/User/Auth.hs)
* [x] make use of `/i/legalhold-login` (instead of current /i/sso-login)

end-to-end test using ACL, nginz route configuration
*  [x] adjust `deploy/services-demo/conf/nginz/zauth_acl.txt` (as well as the other 2 configuration places) to allow the routes needed by the legal hold service
* [x] manual or integration test to check the change in `nginx-auth-module` is working (So far there are no tests testing nginz - except the [smoketests](https://github.com/wireapp/wire-server/tree/develop/tools/api-simulations)) : 1) creating a token by talking to brig directly under `/i/legalhold/login` and 2) making a request to nginz using that token on a route allowed by the `zauth_acl` and making sure nginz is allowing a valid legalhold token through.

* [x] Pick one of two options to deal with token refresh endpoint:
1. ~~keep `/access` and `/legalhold/access` separate (in which case `/legalhold/access` needs to be mapped to `brig` in the nginx configurations (in all 3 places)~~
2. merge `/access` and `/legalhold/access` endpoints (in which case nginz config doesn't need to be adjusted - the nginz config change in this PR can be undone), instead the types on that endpoint need to probably be `Either UserToken LegalHoldUserToken` (see e.g. https://github.com/wireapp/wire-server/blob/develop/services/brig/src/Brig/API.hs#L1216-L1217 an example Either endpoint).

* [x] fix galley tests with the new requirement that legalhold users = team users
* [x] logout & cookie cleanup test

Finally, after merging and deploying, ensure legal hold service can reach all its necessary routes (make sure all nginx configs are adjusted and possibly `/access` to changed to `/legalhold/access` in the legal hold service (depending on the choice in the task above)


libs/zauth/src/Data/ZAuth/Token.hs
```diff
@@ -61,7 +71,16 @@ import Data.ByteString.Lazy.Char8 (split, break, drop)
 import Data.UUID
 import Sodium.Crypto.Sign (Signature (..))
 
-data Type = A | U | B | P deriving (Eq, Show)
+-- | A = Access (Used as short-lived token for Users)
+-- U = User (Used as a cookie for Users to refresh access tokens)
+-- B = Bot
+-- P = Provider
+-- LA = LegalHold Access (Used as short-lived token for LegalHold Service)
+-- LU = LegalHold User (Used as a cookie for LegalHold Service to refresh access tokens)
```
fisx: now i'm wondering if `LD` (legalhold device) isn't a better name.  less symmetrical, but more accurate.  no strong opinion, though.

fisx: you could also move the list from this comment into the declaration:

```haskell
data Type
    = A  --^ Access (...)
    | B  --^ ...
...
```

jschaul: > you could also move ...

thanks!

> now i'm wondering if LD (legalhold device) ...

I prefer the symmetrical name, I think.

services/brig/src/Brig/User/API/Auth.hs
```diff
@@ -204,7 +239,9 @@ rmCookies (uid ::: req) = do
     Auth.revokeAccess uid pw ids lls !>> authError
     return empty
 
-renew :: JSON ::: Maybe ZAuth.UserToken ::: Maybe ZAuth.AccessToken -> Handler Response
+-- FUTUREWORK: let renew take an Either (Token User, Token Access), (Token LegalHoldUser, Token LegalHoldAccess)
+-- Then remove /legalhold/access endpoint.
+renew :: ZAuth.TokenPair u a => JSON ::: Maybe (ZAuth.Token u) ::: Maybe (ZAuth.Token a) -> Handler Response
```
tiago-loureiro: 👌

deploy/services-demo/conf/nginz/zauth_acl.txt
```diff
@@ -4,10 +4,13 @@ a (blacklist (path "/provider")
              (path "/bot/**")
              (path "/i/**"))
 
-u (whitelist (path "/access"))
```
fisx: why?

libs/libzauth/libzauth-c/Cargo.lock
```diff
@@ -1,3 +1,5 @@
+# This file is automatically @generated by Cargo.
+# It is not intended for manual editing.
```
fisx: should it be under vc, though?

libs/zauth/src/Data/ZAuth/Creation.hs
```diff
@@ -96,6 +99,22 @@ accessToken1 dur usr = do
...
+    d <- expiry dur
+    newToken d LU Nothing (mkLegalHoldUser usr rnd)
+
+legalHoldAccessToken :: Integer -> UUID -> Word64 -> Create (Token LegalHoldAccess)
+legalHoldAccessToken dur usr con = do
+    d <- expiry dur
+    newToken d LA Nothing (mkLegalHoldAccess usr con)
+
+legalHoldAccessToken1 :: Integer -> UUID -> Create (Token LegalHoldAccess)
+legalHoldAccessToken1 dur usr = do
```
fisx: what's the idea here?  (i know, same as `accessToken1` above, but could you add a haddock line in both places even though it's not related?  :))

libs/zauth/src/Data/ZAuth/Token.hs
```diff
@@ -61,7 +71,16 @@ import Data.ByteString.Lazy.Char8 (split, break, drop)
...
+data Type = A -- ^ Access (Used as short-lived token for Users)
+          | U -- ^ User (Used as a cookie for Users to refresh access tokens)
+          | B -- ^ Bot
+          | P -- ^ Provider
+          | LA -- ^ LegalHold Access (Used as short-lived token for LegalHold Service)
+          | LU -- ^ LegalHold User (Used as a cookie for LegalHold Service to refresh access tokens)
+          deriving (Eq, Show)
+
+-- | Tag: Tokens for Users with no tag are refreshable themselves and called "UserToken"
+-- Tokens for Users with the tag 'S' are non-refreshable themselves and called "SessionToken"
```
fisx: I would really like to rename this:

```haskell
data SessionTag = SessionTag
```

it's fine to leave the serialization intact, but at least the haskell code might as well readable.

(but not in this PR i guess; if you agree you can leave a `FUTUREWORK` comment.)

libs/zauth/src/Data/ZAuth/Token.hs
```diff
@@ -163,33 +204,41 @@ mkBot = Bot
...
 
-readHeader :: Properties -> Maybe Header
-readHeader p = Header
+readHeader :: Type -> Properties -> Maybe Header
+readHeader t p = Header
     <$> (lookup "v" p >>= fromByteString')
     <*> (lookup "k" p >>= fromByteString')
     <*> (lookup "d" p >>= fromByteString')
-    <*> (lookup "t" p >>= readType)
+    <*> (lookup "t" p >>= readType >>= \t' -> if t == t' then Just t' else Nothing)
```
fisx: nit-pick:

```haskell
    <*> (lookup "t" p >>= readType t)
[...]
    readType A "a" = Just A
    readType U "u" = Just U
    readType B "b" = Just B
    readType P "p" = Just P
    readType LA "la" = Just LA
    readType LU "lu" = Just LU
    readType _  _ = Nothing
```

fisx: I also think that `Header` should be this:

```haskell
data Header (t :: Type) = Header
    { _version   :: !Int
    , _key       :: !Int
    , _time      :: !Integer
    , _tag       :: Maybe Tag
    } deriving (Eq, Show)
```

This is also not part of the feature, but I'm tempted to request that this is included anyway.  (Out of curiosity: did `readHeader` start taking a type argument because it didn't work any more without?  Or is that just an extra check that we *can* do, but neglected to do so far?)

libs/zauth/test/Arbitraries.hs
```diff
@@ -11,17 +11,23 @@ import Sodium.Crypto.Sign
 import Test.Tasty.QuickCheck
 
 instance Arbitrary (Token Access) where
-    arbitrary = mkToken <$> arbitrary <*> arbitrary <*> arbitrary
+    arbitrary = mkToken <$> arbitrary <*> (mkHeader <$> arbitrary <*> arbitrary <*> arbitrary <*> pure A <*> arbitrary) <*> arbitrary
 
 instance Arbitrary (Token User) where
-    arbitrary = mkToken <$> arbitrary <*> arbitrary <*> arbitrary
+    arbitrary = mkToken <$> arbitrary <*> (mkHeader <$> arbitrary <*> arbitrary <*> arbitrary <*> pure U <*> arbitrary) <*> arbitrary
```
fisx:
```suggestion
    arbitrary = mkToken <$> arbitrary <*> ((typ .~ U) <$> arbitrary) <*> arbitrary
```

libs/zauth/test/Arbitraries.hs
```diff
@@ -11,17 +11,23 @@ import Sodium.Crypto.Sign
...
 
 instance Arbitrary (Token Bot) where
     arbitrary = mkToken <$> arbitrary <*> arbitrary <*> arbitrary
 
 instance Arbitrary (Token Provider) where
     arbitrary = mkToken <$> arbitrary <*> arbitrary <*> arbitrary
 
+instance Arbitrary (Token LegalHoldAccess) where
+    arbitrary = mkToken <$> arbitrary <*> (mkHeader <$> arbitrary <*> arbitrary <*> arbitrary <*> pure LA <*> arbitrary) <*> arbitrary
```
fisx:
```suggestion
    arbitrary = mkToken <$> arbitrary <*> ((typ .~ LA) <$> arbitrary) <*> arbitrary
```

libs/zauth/test/Arbitraries.hs
```diff
@@ -11,17 +11,23 @@ import Sodium.Crypto.Sign
...
+
+instance Arbitrary (Token LegalHoldUser) where
+    arbitrary = mkToken <$> arbitrary <*> (mkHeader <$> arbitrary <*> arbitrary <*> arbitrary <*> pure LU <*> arbitrary) <*> arbitrary
```
fisx:
```suggestion
    arbitrary = mkToken <$> arbitrary <*> ((typ .~ LU) <$> arbitrary) <*> arbitrary
```

libs/zauth/test/Arbitraries.hs
```diff
@@ -42,14 +48,20 @@ instance Arbitrary Bot where
...
+
 instance Arbitrary ByteString where
     arbitrary = fromString <$> arbitrary `suchThat` (not . any (== '.'))
 
 instance Arbitrary Signature where
     arbitrary = Signature <$> arbitrary
 
 instance Arbitrary Type where
-    arbitrary = elements [A, U]
+    arbitrary = elements [A, U, LA, LU]
```
fisx:
```suggestion
    arbitrary = elements [minBound..]
```

(That requires deriving Bounded, Enum in `Type`, and it's also not the same.  But why are some constructors missing?  Would that break anything?  Or fix anything?)

services/integration.sh
```diff
@@ -102,10 +103,27 @@ run cannon "2" ${orange}
...
+    prefix=$([ -w /usr/local ] && echo /usr/local || echo "${HOME}/.wire-dev")
+    (cd ${NGINZ_WORK_DIR} && LD_LIBRARY_PATH=$LD_LIBRARY_PATH:${prefix}/lib/ ${TOP_LEVEL}/dist/nginx -p ${NGINZ_WORK_DIR} -c ${NGINZ_WORK_DIR}/conf/nginz/nginx.conf -g 'daemon off;' || kill_all) \
+        | sed -e "s/^/$(tput setaf ${colour})[nginz] /" -e "s/$/$(tput sgr0)/" &
+}
+
+NGINZ_PORT=""
+if [[ $INTEGRATION_USE_NGINZ -eq 1 ]]; then
+    NGINZ_PORT=8080
+    # Note: for integration tests involving nginz,
+    # nginz and brig must share the same zauth public/private keys
```
fisx: is there an easy way to check this, and turn this comment into an error if it's violated?

libs/zauth/test/ZAuth.hs
```diff
@@ -26,6 +26,10 @@ tests = do
         [ testGroup "Parsing"
             [ testProperty "decode . encode == id [access]" testDecEncAccessToken
             , testProperty "decode . encode == id [user]" testDecEncUserToken
+            , testProperty "decode . encode == id [legalhold access]" testDecEncLegalHoldAccessToken
+            , testProperty "decode . encode == id [legalhold user]" testDecEncLegalHoldUserToken
+            , testProperty "decode as User . encode as LegalHoldUser == Nothing " testUserIsNotLegalHoldUser
```
fisx:
```suggestion
            , testProperty "decode as User . encode as LegalHoldUser == Nothing" testUserIsNotLegalHoldUser
```

libs/zauth/test/ZAuth.hs
```diff
@@ -26,6 +26,10 @@ tests = do
...
+            , testProperty "decode as LegalHoldUser . encode as User == Nothing " testUserIsNotLegalHoldUser'
```
fisx:
```suggestion
            , testProperty "decode as LegalHoldUser . encode as User == Nothing" testUserIsNotLegalHoldUser'
```

services/brig/src/Brig/API/Types.hs
```diff
@@ -106,6 +106,16 @@ data PasswordResetError
     | InvalidPasswordResetCode
     | ResetPasswordMustDiffer
 
+data LegalHoldLoginError
+    = LegalHoldLoginNoBindingTeam
+    | LegalHoldLoginLegalHoldNotEnabled
+    -- same as LoginError
```
fisx: what about this then?:

```haskell
data LegalHoldLoginError
    = LegalHoldLoginNoBindingTeam
    | LegalHoldLoginLegalHoldNotEnabled 
    | LegalHoldLoginError LoginError
```

should it ever change, it'll be easy to switch back to your implementation and then mutate it at will.  but if it doesn't change it's much easier to read.

services/brig/src/Brig/API/Error.hs
```diff
@@ -401,3 +415,7 @@ can'tAddLegalHoldClient =
     Wai.Error status400
               "client-error"
               "LegalHold clients cannot be added manually. LegalHold must be enabled on this user by an admin"
+
+legalHoldNotEnabled :: Wai.Error
+legalHoldNotEnabled = Wai.Error status403 "legalhold-not-enabled" "LegalHold must be enabled and configured on the team first."
```
fisx:
```suggestion
legalHoldNotEnabled = Wai.Error status403 "legalhold-not-enabled" "LegalHold must be enabled and configured on the team first"
```

services/brig/src/Brig/IO/Intra.hs
```diff
@@ -663,6 +665,16 @@ getTeamName tid = do
     req = paths ["i", "teams", toByteString' tid, "name"]
         . expect2xx
 
+getTeamLegalHoldStatus :: UserId -> TeamId -> AppIO ViewLegalHoldService
+getTeamLegalHoldStatus u tid = do
```
fisx: I think this works, but you could also hit `get "/i/teams/:tid/features/legalhold"` and avoid the fake authentication.  No strong opinion from me here.

services/brig/src/Brig/ZAuth.hs
```diff
@@ -90,52 +103,68 @@ runZAuth :: MonadIO m => Env -> ZAuth a -> m a
...
-    (UserTokenTimeout (60 * 60 * 24 * 28))    -- 28 days
-    (SessionTokenTimeout (60 * 60 * 24))      -- 1 day
-    (AccessTokenTimeout 900)                  -- 15 minutes
-    (ProviderTokenTimeout (60 * 60 * 24 * 7)) -- 7 days
+    (UserTokenTimeout (60 * 60 * 24 * 28))          -- 28 days
+    (SessionTokenTimeout (60 * 60 * 24))            -- 1 day
+    (AccessTokenTimeout 900)                        -- 15 minutes
+    (ProviderTokenTimeout (60 * 60 * 24 * 7))       -- 7 days
+    (LegalHoldUserTokenTimeout (60 * 60 * 24 * 56)) -- 56 days
+    (LegalHoldAccessTokenTimeout (60 * 15 ))        -- 15 minutes
```
fisx:
```suggestion
    (LegalHoldAccessTokenTimeout (60 * 15))        -- 15 minutes
```

fisx: (anyway i think @jschaul you said yourself a while ago that we shouldn't have defaults in the code, but enforce the existence of an exhaustive options file.  the more i think about it, the more i agree.  make a `FUTUREWORK` here about this?)

services/brig/src/Brig/ZAuth.hs
```diff
@@ -90,52 +103,68 @@ runZAuth :: MonadIO m => Env -> ZAuth a -> m a
...
-type AccessToken   = Token Access
-type UserToken     = Token User
-type ProviderToken = Token Provider
-type BotToken      = Token Bot
+type AccessToken          = Token Access
+type UserToken            = Token User
+type ProviderToken        = Token Provider
+type BotToken             = Token Bot
+type LegalHoldUserToken   = Token LegalHoldUser
+type LegalHoldAccessToken = Token LegalHoldAccess
```
fisx: All these type aliases don't seem to be any shorter than their definitions, only more opaque.  `FUTUREWORK` to remove them and inline the defs?

services/brig/src/Brig/ZAuth.hs
```diff
@@ -210,21 +296,58 @@ newProviderToken pid = liftZAuth $ do
         let ProviderTokenTimeout ttl = z^.settings.providerTokenTimeout
         in ZC.providerToken ttl (toUUID pid)
 
+mkLegalHoldUserToken :: MonadZAuth m => UserId -> Word32 -> UTCTime -> m LegalHoldUserToken
+mkLegalHoldUserToken u r t = liftZAuth $ do
+    z <- ask
+    liftIO $ ZC.runCreate (z^.private) (z^.settings.keyIndex) $
+       ZC.newToken (utcTimeToPOSIXSeconds t) LU Nothing (mkLegalHoldUser (toUUID u) r)
```
fisx: The only difference to `mkUserToken'` is the `LU` instead fo the `U`.  I feel there should be a good way to combine the two implementations.  If you:

```haskell
 mkLegalHoldUserToken :: forall (t :: Token) m. MonadZAuth m 
       => UserId -> Word32 -> UTCTime -> m (Token t)
```

And then somehow construct a value from the type-level `LU` or `U`, resp., you have one function that can be called in place of either of these.  Same (probably) for the rest of the class methods.

In fact, here is what I would do:

```haskell
class TokenType {- or some better name -} (a :: Type) where
  tokenTypeVal :: Type  -- i think this will work if you turn on `AllowAmbiguousTypes`; if not use a `Proxy`.

instance TokenType U where
  tokenTypeVal _ = U

instance TokenType LU where
  tokenTypeVal _ = LU
```

and then, instead of as a class method, you can implement `mkUserToken` thusly:

```haskell
mkUserToken :: forall (t :: Type) m. (TokenType t, MonadZAuth m) => UserId -> Word32 -> UTCTime -> m (Token t)
mkUserToken u r t = liftZAuth $ do
    z <- ask
    liftIO $ ZC.runCreate (z^.private) (z^.settings.keyIndex) $
       ZC.newToken (utcTimeToPOSIXSeconds t) (tokenTypeVal @t) Nothing (mkLegalHoldUser (toUUID u) r)
```

services/brig/test/integration/API/Team/Util.hs
```diff
@@ -156,6 +157,14 @@ getTeams u galley =
 newTeam :: Team.BindingNewTeam
 newTeam = Team.BindingNewTeam $ Team.newNewTeam (unsafeRange "teamName") (unsafeRange "defaultIcon")
 
+putEnabled :: HasCallStack => TeamId -> LegalHoldStatus -> Galley -> Http ()
+putEnabled tid enabled g = do
```
fisx:
```suggestion
putLegalHoldEnabled tid enabled g = do
```

deploy/services-demo/conf/nginz/zauth_acl.txt
fisx: (hm, ok, somehow this must have to do with the fact that we're using nginz in integration tests now.  perhaps it's easier if you explain this to me in person / on the phone.)

(i would also like to talk about the rust code, since i'm not very familiar with that code base.  i'll see if you have time tomorrow.)

jschaul: This is actually unrelated to integration tests.

whitelisting /access for user tokens has no effect, as that route is not zauth-authenticated. This line has been in the code for a few years without effect, so we might as well remove it. I noticed this having no effect due to a new integration test that also talks to nginz.

libs/libzauth/libzauth-c/Cargo.lock
jschaul: It helps to have .lock files under version control to always get the same thing when compiling today or in 6 months. 
If we were working on rust code a lot, it could make sense to not put it under version control, to notice when new versions of the rust compiler or some libraries become available and something becomes incompatible.

libs/zauth/src/Data/ZAuth/Token.hs
jschaul: > is that just an extra check that we can do, but neglected to do so far?

Exactly, the type header was never checked, leading to issues, since LegalHold and regular tokens are the same *except* for their type header.

services/brig/src/Brig/ZAuth.hs
jschaul: That's not what I said, please don't misquote me :). There can be defaults in the code as long as those defaults can be overridden via configuration files. The discussion related to a hardcoded setting "by default SSO is disabled and must be enabled per-team via backoffice". And there was no way to change this behaviour via configuration - but now there is, thanks to your work on that other PR). This part of the code is different, as any default value here can easily be overridden in configuration, if, for some installation, expiry times should behave differently. In fact, we make use of that in two separate places at the moment.

services/brig/src/Brig/ZAuth.hs
jschaul: I suppose it saves parentheses. And reads more naturally. Not sure I agree to remove them, but I'm not opposed if you want to remove them in some future PR.

services/brig/src/Brig/ZAuth.hs
jschaul: Thanks! I'll try that.

services/integration.sh
jschaul: If it's violated, tests will fail with an "Unauthorized" 401 from nginz. What do you mean exactly with "turn into an error" ? Normally, nginz and brig don't run on the same machine, they run on different VMs or on different pods, so I'm not sure there can be a good way to detect this precise condition. Any ideas?

libs/zauth/src/Data/ZAuth/Creation.hs
jschaul: Good idea. Done.

libs/zauth/src/Data/ZAuth/Token.hs
jschaul: > nit-pick:
> 
> ```haskell
>     <*> (lookup "t" p >>= readType t)
> [...]
>     readType A "a" = Just A
>     readType U "u" = Just U
>     readType B "b" = Just B
>     readType P "p" = Just P
>     readType LA "la" = Just LA
>     readType LU "lu" = Just LU
>     readType _  _ = Nothing
> ```

Thanks, that's clearer!

services/integration.sh
fisx: hm, good point.

i suppose you could write a test that does nothing but try to get a 200 from nginz the easiest way possible, and if it gets a 401, the error has "make sure keys match, or something" in it.

You decide whether it's worth it.

services/brig/src/Brig/User/API/Auth.hs
```diff
@@ -185,11 +191,24 @@ ssoLogin (req ::: persist ::: _) = do
...
 logout (_ ::: Nothing ::: Nothing) = throwStd authMissingCookieAndToken
 logout (_ ::: Nothing ::: Just _ ) = throwStd authMissingCookie
 logout (_ ::: Just _  ::: Nothing) = throwStd authMissingToken
-logout (_ ::: Just ut ::: Just at) = do
+logout (_ ::: Just (Left _) ::: Just (Right _)) = throwStd authTokenMismatch
+logout (_ ::: Just (Right _) ::: Just (Left _)) = throwStd authTokenMismatch
+logout (_ ::: Just (Left ut) ::: Just (Left at)) = do
+    Auth.logout ut at !>> zauthError
+    return empty
+logout (_ ::: Just (Right ut) ::: Just (Right at)) = do
```
fisx:
```suggestion
logout (_ ::: Just (either id id -> ut) ::: Just (either id id -> at)) = do
```

And then you could save yourself writing the morally redundant `Left` case.  But arguably your code is way more readable.

services/brig/src/Brig/User/API/Auth.hs
```diff
@@ -204,25 +223,33 @@ rmCookies (uid ::: req) = do
...
-renew (_ ::: Just ut ::: at) = do
-    a <- Auth.renewAccess ut at !>> zauthError
-    tokenResponse a
+renew (_ ::: Just userToken ::: accessToken) = do
+    case (userToken, accessToken) of
+         (Left ut, Just (Left at)) -> (Auth.renewAccess ut (Just at) !>> zauthError) >>= tokenResponse
+         (Left ut, Nothing) -> Auth.renewAccess @ZAuth.User @ZAuth.Access ut Nothing !>> zauthError >>= tokenResponse
+         (Right lut, Just (Right lat)) -> Auth.renewAccess @ZAuth.LegalHoldUser @ZAuth.LegalHoldAccess lut (Just lat) !>> zauthError >>= tokenResponse
+         (Right lut, Nothing) -> Auth.renewAccess @ZAuth.LegalHoldUser @ZAuth.LegalHoldAccess lut Nothing !>> zauthError >>= tokenResponse
+         (_, _) -> throwStd authTokenMismatch
```
fisx: What happens if we change `TokenPair u a` to `TokenPair u u` everywhere?  I suppose if that worked, it would have been done, but at least here it would nicely collapse 4 cases into two, because we wouldn't have to dynamically pattern-match into the second part of the tuple to find the type.

(I'm not sure that makes sense.  Should I try myself?)

services/brig/src/Brig/User/API/Auth.hs
fisx: nope, sorry, that wouldn't type-check.  i think there may be a solution down that rabbit-hole, but on second thoughts i'm against it.  :)

services/brig/src/Brig/User/Auth.hs
```diff
@@ -222,4 +228,32 @@ ssoLogin (SsoLogin uid label) typ = do
...
             AuthEphemeral          -> throwE LoginEphemeral
             AuthInvalidUser        -> throwE LoginFailed
-    newAccess uid typ label
+    newAccess @ZAuth.User @ZAuth.Access uid typ label
+
+-- | Log in as a LegalHold service, getting LegalHoldUser/Access Tokens.
+legalHoldLogin :: LegalHoldLogin -> CookieType -> ExceptT LegalHoldLoginError AppIO (Access ZAuth.LegalHoldUser)
+legalHoldLogin (LegalHoldLogin uid label) typ = do
+    Data.reauthenticate uid Nothing `catchE` \case
+        ReAuthMissingPassword -> pure ()
```
fisx: This looks very wrong to me.  If it's not, it could probably use a comment.  I would have expected something like this:

```suggestion
        ReAuthMissingPassword -> throwE e
```

services/brig/src/Brig/User/Auth.hs
```diff
@@ -222,4 +228,32 @@ ssoLogin (SsoLogin uid label) typ = do
...
+        ReAuthError e -> case e of
+            AuthInvalidCredentials -> pure ()
```
fisx: same here.

services/brig/src/Brig/User/Auth.hs
```diff
@@ -222,4 +228,32 @@ ssoLogin (SsoLogin uid label) typ = do
...
+    -- legalhold login is only possible if
+    -- * the user is a team user
+    -- * and the team has legalhold enabled
+    mteam <- lift $ Data.lookupUserTeam uid
+    case mteam of
+         Nothing -> throwE LegalHoldLoginNoBindingTeam
+         Just tid -> assertLegalHoldEnabled uid tid
+    -- create access token and cookie
+    newAccess @ZAuth.LegalHoldUser @ZAuth.LegalHoldAccess uid typ label
+        `catchE` \e -> throwE $ LegalHoldLoginError e
```
fisx:
```suggestion
        `catchE` (throwE . LegalHoldLoginError)
```

(only if you like it, too.)

services/brig/src/Brig/API.hs
```diff
@@ -59,6 +59,8 @@ import qualified Brig.Provider.API             as Provider
 import qualified Brig.Team.API                 as Team
 import qualified Brig.Team.Email               as Team
 import qualified Brig.TURN.API                 as TURN
+--import qualified System.Logger.Class           as Log
```
fisx: remove this line?

services/brig/src/Brig/User/API/Auth.hs
jschaul: But TokenPair u u doesn't make much sense - The acceptable inputs are (User, Access) and (LegalHoldUser, LegalHoldAccess), but they are two different types, not twice the same.

services/brig/src/Brig/User/Auth.hs
jschaul: Good point, thanks, I forgot about this. Galley will then have to pass on the password to brig.

services/brig/test/integration/API/User/Auth.hs
```diff
@@ -279,6 +345,92 @@ testLimitRetries (Just conf) brig = do
...
+testTeamUserLegalHoldLogin brig galley = do
+    -- create team user Alice
+    (alice, tid) <- createUserWithTeam brig galley
+    now <- liftIO getCurrentTime
+    -- fail if legalhold isn't activated yet for this user
+    legalHoldLogin brig (LegalHoldLogin alice (Just defPassword) Nothing) PersistentCookie !!! do
+        const 403 === statusCode
+
+    putLegalHoldEnabled tid LegalHoldEnabled galley -- enable it for this team
+    _rs <- legalHoldLogin brig (LegalHoldLogin alice (Just defPassword) Nothing) PersistentCookie
```
fisx: Fail again here at the moment LH is enabled for team, but not for member?

services/brig/test/integration/API/User/Auth.hs
```diff
@@ -76,12 +98,56 @@ tests conf m z b = testGroup "auth"
...
+    u <- randomUser b
+    let Just email = userEmail u
+    -- Login with email
+    rs <- login b (defEmailLogin email) PersistentCookie
+        <!! const 200 === statusCode
+    let c = decodeCookie rs
+        t = decodeToken rs
+    -- ensure nginz allows refresh at /access
+    _rs <- post (n . path "/access" . cookie c . header "Authorization" ("Bearer " <> (toByteString' t))) <!! do
+        const 200 === statusCode
```
fisx: I think this is the line that the comment in 114 refers to.  You could move the comment here (not tested):

```suggestion
        assertEqual "/access (requires that private key on nginz and public key on brig match)"
            (const 200) statusCode
```

services/brig/test/integration/API/User/Auth.hs
```diff
@@ -52,12 +57,29 @@ tests conf m z b = testGroup "auth"
...
+                , test m "failure-suspended" (testSuspendedLegalHoldLogin b g)
+                , test m "failure-no-user" (testNoUserLegalHoldLogin b)
+                , test m "failure-wrong-password" (testWrongPasswordLegalHoldLogin b g)
+                , test m "always-persistent-cookie" (testLegalHoldSessionCookie b g)
+                , test m "legalhold-logout" (testLegalHoldLogout b g)
+                ]
+            , testGroup "nginz"
+                [ test m "nginz-login" (testNginz b n)
+                , test m "nginz-legalhold-login" (testNginzLegalHold b g n)
+                ]
```
fisx: Are [these](https://github.com/wireapp/wire-server/blob/dbf96139ac4d19db854f72abd98306e5679f30a2/services/galley/test/integration/API/Teams/LegalHold.hs#L102-L111) covered now?  If so, please remove that TODO.  If not, please fix!  :)

services/brig/test/integration/API/User/Auth.hs
jschaul: That should be tested as part of galley's legal hold tests (if it's not, it should get added there, but it's not related to this PR)

services/brig/test/integration/API/User/Auth.hs
jschaul: The Bilge.assert and the HUnit.assert don't work well together, so the above doesn't compile, but I can do this:

```haskell
_rs <- get (n . path "/clients" . header "Authorization" ("Bearer " <> (toByteString' t)))
liftIO $ assertEqual "Ensure nginz is started. Ensure nginz and brig share the same private/public zauth keys. Ensure ACL file is correct." 200 (statusCode _rs)
```

fisx: Oops, I thought there was an `assertEqual` in hunit and one in bilge.  You could you https://hoogle.zinfra.io/file/wire-server/.stack-work/install/x86_64-linux/custom-wire-1.1-G6NVtOVxqIB2/8.4.3/doc/bilge-0.22.0/Bilge-Assert.html#v:assertTrue, though?

services/brig/test/integration/API/User/Auth.hs
jschaul: The first one, yes. I don't know about the state of the other two, but they're not related to this PR directly.

services/brig/src/Brig/ZAuth.hs
jschaul: I left a FUTUREWORK for this, as it's not as easy as I had hoped.

services/brig/src/Brig/IO/Intra.hs
jschaul: Thanks!
Actually, this was a source of a bug: one galley integration test creates a user with no permissions, but here, I expect the call from brig to galley to succeed, which it doesn't when a user doesn't have permissions to view these settings. 
I changed this now to make use of the `/i/` endpoint.

services/brig/src/Brig/IO/Intra.hs
```diff
@@ -666,14 +666,12 @@ getTeamName tid = do
     req = paths ["i", "teams", toByteString' tid, "name"]
         . expect2xx
 
-getTeamLegalHoldStatus :: UserId -> TeamId -> AppIO ViewLegalHoldService
-getTeamLegalHoldStatus u tid = do
+getTeamLegalHoldStatus :: TeamId -> AppIO LegalHoldTeamConfig
```
fisx: :+1:

services/galley/test/integration/API/Teams/LegalHold.hs
```diff
@@ -460,16 +460,11 @@ testRemoveLegalHoldFromTeam = do
...
         -- enable legalhold for member
-        do  requestLegalHoldDevice owner member tid !!! testResponse 201 Nothing
-            liftIO $ putStrLn "---2a"
-            liftIO $ threadDelay 1000000
-            -- TODO: the line below fails with a 500 somewhere. Why?
-            approveLegalHoldDevice (Just defPassword) member member tid !!! testResponse 200 Nothing
-            liftIO $ putStrLn "---2b"
-            UserLegalHoldStatusResponse userStatus _ _ <- getUserStatusTyped member tid
-            liftIO $ assertEqual "After approval user legalhold status should be Enabled"
+        do requestLegalHoldDevice owner member tid !!! testResponse 201 Nothing
```
fisx: yeay, 3-space-indentation!  :-)

services/brig/test/integration/API/User/Auth.hs
```diff
@@ -671,19 +863,24 @@ listCookiesWithLabel b u l = do
...
 assertSanePersistentCookie ck = do
     assertBool "type" (cookie_persistent ck)
     assertBool "http-only" (cookie_http_only ck)
     assertBool "expiry" (cookie_expiry_time ck > cookie_creation_time ck)
     assertBool "domain" (cookie_domain ck /= "")
     assertBool "path" (cookie_path ck /= "")
+    let mtoken = (fromByteString (cookie_value ck)) :: Maybe (ZAuth.Token u)
+    assertBool "cookie token exists" (isJust mtoken)
+    let (Just token) = mtoken
+    assertBool "cookie token has correct type" ((token^.ZAuth.header.ZAuth.typ) == ZAuth.zauthType @u)
```
fisx: very nit-picky stuff, leading to a question that maybe genuinely interesting in the end:

- I feel strongly we should require operators be surrounded spaces in our coding style.  This is a good case to motivate this rule: the double-use of `.` for module qualification and function composition is very hard to read.
- The error message from pattern match failure in line 875 is just as good as the one from the assertion in 874, so this code can be simplified.
- I personally like type information best where a name is bound (ie., introduced).
- Too many parens.  :)

So this could be rewritten:

```haskell
    let Just (token :: ZAuth.Token u) = fromByteString (cookie_value ck)
        tokentype = ZAuth.zauthType @u
    assertBool "cookie token has correct type" $ token ^. ZAuth.header . ZAuth.typ == tokentype
```

...  and then I still have one more question: Can `token` really be parsed if it fails the subsequent assertion?  Should it?

services/galley/src/Galley/Intra/Client.hs
```diff
@@ -52,16 +53,16 @@ notifyClientsAboutLegalHoldRequest requesterUid targetUid lastPrekey' = do
                 . json (LegalHoldClientRequest requesterUid lastPrekey')
                 . expect2xx
 
-getLegalHoldAuthToken :: UserId -> Galley OpaqueAuthToken
-getLegalHoldAuthToken uid = do
+getLegalHoldAuthToken :: UserId -> Maybe PlainTextPassword -> Galley OpaqueAuthToken
```
fisx: I am right that this is because some team members may have a password, and of course we want that password to be provided as an extra safety hook if a user approves a LH device, and we just forgot about that in the earlier part of the implementation?

PR #763, opened 2019-05-27
Upgrade CQL protocol version to V4

tiago-loureiro:
We're running cassandra 3.11+ across the board so this operation should be safe. Relevant changes [from V3](https://github.com/apache/cassandra/blob/trunk/doc/native_protocol_v4.spec#L1171-L1187)


tiago-loureiro:
Let's keep a close eye on this one and make sure nothing goes funky on `staging` or any other env.

PR #770, opened 2019-06-02
Fix flaky unit test.

### LINKED ISSUE #764: Tests fail due to commitBuffer: invalid argument (invalid character)
While building this occurred. Distro is ArchLinux:
```
Test suite failure for package cannon-0.31.0
    cannon-tests:  exited with: ExitFailure 1
[.. full log ..]
    benchmarking Cannon.Dict/slices 1
    time                 31.03 ms   (20.21 ms .. 43.63 ms)
                         0.656 Rcannon-tests: <stdout>: commitBuffer: invalid argument (invalid character)

Test suite failure for package types-common-0.16.0
    tests:  exited with: ExitFailure 1
[.. full log ..]
    Tests
      Properties
        Ascii
          validate (toText x) == Right x:                             OK (0.02s)
            +++ OK, passed 100 tests.
          unsafeFromByteString (toByteString x) == x:                 OK (0.01s)
            +++ OK, passed 100 tests.
          validate (toText x <> "tests: <stdout>: commitBuffer: invalid argument (invalid character)
```

fisx:
Fixes #764 

This PR restricts input to [`Ascii.unsafeFromByteString`](https://github.com/wireapp/wire-server/blob/f03d8f701bfb3458e81dd446f4f772eee01875e8/libs/types-common/src/Data/Text/Ascii.hs#L293-L297) to valid input in the test.  It is safe to not test invalid input, since the function name suggests caution, and we're actually only ever calling it on valid input:

```sh
$ git grep -Hn unsafeFromByteString
libs/types-common/src/Data/Text/Ascii.hs:52:    , unsafeFromByteString
libs/types-common/src/Data/Text/Ascii.hs:196:encodeBase64 = unsafeFromByteString . B64.encode
libs/types-common/src/Data/Text/Ascii.hs:236:encodeBase64Url = unsafeFromByteString . B64Url.encode
libs/types-common/src/Data/Text/Ascii.hs:267:encodeBase16 = unsafeFromByteString . B16.encode
libs/types-common/src/Data/Text/Ascii.hs:296:unsafeFromByteString :: AsciiChars c => ByteString -> AsciiText c
libs/types-common/src/Data/Text/Ascii.hs:297:unsafeFromByteString = AsciiText . decodeLatin1
libs/types-common/test/Test/Properties.hs:33:        , testProperty "unsafeFromByteString (toByteString x) == x" $
libs/types-common/test/Test/Properties.hs:34:            \(encodeBase64 -> a) -> Ascii.unsafeFromByteString (toByteString' a) == a
libs/types-common/test/Test/Properties.hs:35:               -- (unsafeFromByteString occasionally fails on Ascii strings)
```

### DIFF
--- a/libs/types-common/test/Test/Properties.hs
+++ b/libs/types-common/test/Test/Properties.hs
@@ -31,7 +31,8 @@ tests = testGroup "Properties"
         [ testProperty "validate (toText x) == Right x" $
             \(a :: Ascii) -> Ascii.validate (Ascii.toText a) == Right a
         , testProperty "unsafeFromByteString (toByteString x) == x" $
-            \(a :: Ascii) -> Ascii.unsafeFromByteString (toByteString' a) == a
+            \(encodeBase64 -> a) -> Ascii.unsafeFromByteString (toByteString' a) == a
+               -- (unsafeFromByteString occasionally fails on Ascii strings)
         , testProperty "validate (toText x <> \"𝄞\") /= Right x" $
             \(a :: Ascii) -> Ascii.validate (Ascii.toText a <> "𝄞") /= Right a
         ]


fisx:
The problem is that the test case is failing because it's doing a thing that is supposed to fail.  :-)

The `Arbitrary` instance for `Ascii` is correct, the problem is that the valid inputs of the function form a subset of the type.

I checked the places where `unsafeFromByteString` is called, and it always is locally guaranteed to get valid input (see the grep output above).  So I came up with this quick & easy fix.

PR #775, opened 2019-06-10
Documentation changes for TURN setup.

raman-techindustan:
Made changes for TURN/STUN server.

### DIFF
--- a/mailboxes.json
+++ b/mailboxes.json
@@ -1,0 +1,7 @@
+[
+    { "host": "imap.bar.com"
+    , "user": "foo@bar.com"
+    , "pass": "secret"
+    , "conn": 1
+    }
+]
--- a/services/restund/README.md
+++ b/services/restund/README.md
@@ -127,3 +127,17 @@ Example rkt command:
     --user=restund \
     --group=restund
 ```
+
+In case You have set up restund without docker, you just need to make some of these changes:
+
+Put your private IP of the server in place of: `{{ ansible_default_ipv4.address }}`. And replace restund listen ports with `3478`, for both UDP and TCP.
+
+You may comment these out in case you don't want to use:
+```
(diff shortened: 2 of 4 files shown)


mailboxes.json
```diff
@@ -0,0 +1,7 @@
+[
```
tiago-loureiro: This file is already present at: https://github.com/wireapp/wire-server/blob/develop/tools/api-simulations/mailboxes.example.json - perhaps you can update the description added above to point to that file instead?

services/restund/README.md
```diff
@@ -127,3 +127,17 @@ Example rkt command:
     --user=restund \
     --group=restund
 ```
+
+In case You have set up restund without docker, you just need to make some of these changes:
+
```
tiago-loureiro: Please note that the `restund` config file is valid for both running `restund` with `docker` or without

services/restund/README.md
```diff
@@ -127,3 +127,17 @@ Example rkt command:
...
+Put your private IP of the server in place of: `{{ ansible_default_ipv4.address }}`. And replace restund listen ports with `3478`, for both UDP and TCP.
```
tiago-loureiro: Note that private IP may be slightly misleading;

`turn_relay_addr` is the address that the TURN server will bind to
`turn_public_addr` may be the same or not; it is the address "over which other TURN servers should reach this TURN server over UDP". For a better description, check [this PR](https://github.com/wireapp/wire-server-deploy/pull/109)

R4m4n: UDP for the server can be defined in the `udp_listen`. And `turn_public_addr` can be commented out if not needed.

R4m4n:
@tiago-loureiro,
Is it good to go?

PR #785, opened 2019-06-17
Log this message only at debug level

tiago-loureiro:
This line accounts for > 90% of our log lines and does not really provide any interesting info besides the fact that these users should probably publish prekeys more often.


tiago-loureiro:
There are multiple reasons why this _may_ happen: a device that is no longer used, for instance, and not deleted from your account, will _never_ push more prekeys.

I think in general this might be an interesting piece of information (revealing bad strategy for prekey refresh or even knowing that we have a lot of "ghost" devices) but at info level it seems a bit too noisy to me as it keeps spamming our logs.

PR #787, opened 2019-06-18
Expose legalhold status and check if it is really disabled

tiago-loureiro:
Exposes the GET LH status so that admins can view this information. In addition, it _actually_ checks for LH disabled at the DB level so that we can manually disable LH using the internal endpoint.


tiago-loureiro:
I will merge this as-is given there won't probably be reviews this week. Please feel free to review it later! 🙇

PR #789, opened 2019-06-19
restund: making restund build work on alpine

z-dule:


arianvp:
Restund build has been moved to github.com/wireapp/restund/

PR #780, opened 2019-06-12
[WIP] Add helper script for parsing access.log files.

fisx:
tasks:

- [ ] document sample usage
- [ ] remove https://github.com/wireapp/docker-squid4/blob/7e556c79c1c90040785e84a1a14262f1983ed516/parse-accesslog.hs and reference this here.

PR #792, opened 2019-06-21
[WIP] team settings extra

fisx:
Related to #780, but can be independently merged.

also related:
- https://github.com/wireapp/wire-server/blob/more-scripts/deploy/services-demo/create_test_team_admin.sh
- https://github.com/wireapp/wire-server/blob/more-scripts/deploy/services-demo/create_test_team_member.sh

(these generate test users, though, and what we want here is a way to invite proper users in bulk.)

### DIFF
--- a/stack.yaml
+++ b/stack.yaml
@@ -36,6 +36,7 @@ packages:
 - tools/stern
 - tools/db/service-backfill
 - tools/db/auto-whitelist
+- tools/team-settings-extra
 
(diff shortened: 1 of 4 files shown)


fisx:
Obsoleted by #801.

This approach here is fun because it uses `servant-client` rather than bash and curl, but it's not done, and it would not add any functionality to the product on top of the shell scripts that we already have.

PR #797, opened 2019-07-08
Feature/legalhold fixes

tiago-loureiro:
In this PR:
 - [x] - _It looks like the "approve" endpoint for legal hold does not return a "Content-Type" in its response, so Firefox thinks that it is XML._
 - [x] - ~~_the teams/team_id/legalhold/user_id response does not contain the requester field_~~ removed from the event too
 - [x] - _I register legal hold service and send the pending request to a user. This user sees the popup on the client but waits now. In this moment, i update the legal hold service settings via backend to a new public key. Then i accept the pending legal hold request by the user. Result: It is activated. Should the backend not throw an error because of the wrong public key?_
 - [x] - _If i use the backend call to deactivate LH for a user on a user that was never under LH and never in pending state. The backend sends out events to the clients and the clients show system messages._

### DIFF
--- a/services/galley/src/Galley/API/LegalHold.hs
+++ b/services/galley/src/Galley/API/LegalHold.hs
@@ -25,7 +25,7 @@ import qualified Galley.Data                  as Data
 import qualified Galley.Data.LegalHold        as LegalHoldData
 import qualified Galley.Intra.Client as Client
 import qualified Galley.External.LegalHoldService as LHService
-import qualified System.Logger as Logger
+import qualified System.Logger.Class as Log
 
 assertLegalHoldEnabled :: TeamId -> Galley ()
 assertLegalHoldEnabled tid = unlessM (isLegalHoldEnabled tid) $ throwM legalHoldNotEnabled
@@ -126,7 +126,6 @@ getUserStatus (_zusr ::: tid ::: uid ::: _) = do
     -- assertLegalHoldEnabled tid
     mTeamMember <- Data.teamMember tid uid
     teamMember <- maybe (throwM teamMemberNotFound) pure mTeamMember
-    lg <- view applog
     statusResponse <- case (view legalHoldStatus teamMember) of
         UserLegalHoldDisabled ->
             pure $ UserLegalHoldStatusResponse UserLegalHoldDisabled Nothing Nothing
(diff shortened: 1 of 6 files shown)


services/galley/src/Galley/API/LegalHold.hs
```diff
@@ -205,7 +205,6 @@ approveDevice (zusr ::: tid ::: uid ::: connId ::: req ::: _) = do
     LHService.confirmLegalHold clientId tid uid legalHoldAuthToken
     LegalHoldData.setUserLegalHoldStatus tid uid UserLegalHoldEnabled
     -- send event at this point
-
```
fisx: Is comment in the line above a `TODO`?

services/galley/src/Galley/API/LegalHold.hs
```diff
@@ -222,11 +221,22 @@ disableForUser
...
+            Just UserLegalHoldPending -> True
+            _                         -> False
+
+    disableLH = do
+        DisableLegalHoldForUserRequest mPassword <- fromJsonBody req
+        ensureReAuthorised zusr mPassword
+        Client.removeLegalHoldClientFromUser uid
+        LHService.removeLegalHold tid uid
+        LegalHoldData.setUserLegalHoldStatus tid uid UserLegalHoldDisabled
+        -- send event at this point
```
fisx: TODO?

services/galley/src/Galley/API/LegalHold.hs
```diff
@@ -222,11 +221,22 @@ disableForUser
...
+        then disableLH >> pure empty
+        else pure noContent
+  where
+    -- If not enabled nor pending, then it's disabled
+    userLHNotDisabled mems = do
+        let target = findTeamMember uid mems
+        case fmap (view legalHoldStatus) target of
+            Just UserLegalHoldEnabled -> True
+            Just UserLegalHoldPending -> True
+            _                         -> False
```
fisx: enumerate all cases so we will catch changes in the data type in the future?  (i think it's only two more, but i haven't checked.)

services/galley/test/integration/API/Teams/LegalHold.hs
```diff
@@ -71,7 +72,7 @@ tests s = testGroup "Teams LegalHold API"
     , test s "DELETE /teams/{tid}/legalhold/{uid}" testDisableLegalHoldForUser
 
       -- legal hold settings
-    , test s "POST /teams/{tid}/legalhold/settings" testCreateLegalHoldTeamSettings
+    , test s "XXX POST /teams/{tid}/legalhold/settings" testCreateLegalHoldTeamSettings
```
fisx: is this intentional?  what does it mean?

services/galley/src/Galley/API/LegalHold.hs
tiago-loureiro: Yes, this is a TODO - I will fix that. We currently send events at a different point in time, we should improve that.

services/galley/src/Galley/API/LegalHold.hs
tiago-loureiro: 👍

services/galley/test/integration/API/Teams/LegalHold.hs
tiago-loureiro: X-Rated 🙈

PR #799, opened 2019-07-08
Cleanup

fisx:
stumbled over this while reviewing https://github.com/wireapp/wire-server/pull/797

### DIFF
--- a/tools/stern/src/Stern/API.hs
+++ b/tools/stern/src/Stern/API.hs
@@ -44,7 +44,7 @@ import System.Logger.Class hiding ((.=), name, Error, trace)
 import Util.Options
 
 import qualified Data.Metrics.Middleware      as Metrics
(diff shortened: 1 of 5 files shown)


tools/stern/src/Stern/API.hs
```diff
@@ -469,6 +469,7 @@ isUserKeyBlacklisted emailOrPhone = do
         then response status200 "The given user key IS blacklisted"
         else response status404 "The given user key is NOT blacklisted"
   where
+    -- (we can't use 'Network.Wai.Utilities.Response.json' as it doesn't use a status code.)
```
tiago-loureiro: You could replace it with:
```
return . setStatus st . json $ object ["status" .= (reason :: Text)]
```

fisx: m|

(thanks!)

PR #800, opened 2019-07-09
LH: disable by default; tests.

fisx:
better tests, plus:

- make LH disabled by default in all teams.
- do not allow to create LH devices via `POST /clients` any more.
- always test that LH is enabled first when performing LH operations.


fisx:
can (does not have to) be reviewed.  but i'll add a few more small fixes in separate commits before i want to merge.

fisx:
(We're not done, but i would like to get this reviewed and merged at this stage, so I removed the WIP.)

fisx:
(ah.  ci only runs on PRs based on develop i guess...)

PR #808, opened 2019-07-23
Canonicalize capture names

fisx:
Since some servant- and prometheus-related refactorings earlier this year (I think), capturing metrics only worked under the assumption that all capture variables names (the path segments like `:uid`) were had the same name when in the same node in the routing tree.

This PR adds tests that enforces this assumption.  I will try to fix the unnecessary assumption independently, but even though not technically necessary, I think it's really quite unnecessary to allow for different names for the same thing.

If we have different capture variables for the same node in the routing table, we get errors like this in the future;

```
FAIL
    src/Data/Metrics/Test.hs:17:
    inconcistent sitemap
    expected: []
     but got: [bad routing tables: the prefix "/teams" contains these variables with resp. numbers of routes under them: [(":tid",8),(":id",2)]]
[...]
```

This happens in all services; and has been fixed for existing routing tables so the tests pass now.

Consequently, some alerts and graphana dashboards will not work, since they have the old paths hard-coded in their rules.  Consider fixing this before deploying this PR (either in staging or in production).

### DIFF
--- a/docs/reference/user/rich-info.md
+++ b/docs/reference/user/rich-info.md
@@ -18,7 +18,7 @@ Only team members and partners can see the user's rich info.
 
 ### Querying rich info {#RefRichInfoGet}
 
-`GET /users/:user/rich-info`. Sample output:
+`GET /users/:uid/rich-info`. Sample output:
 
 ```json
 {
--- a/libs/metrics-wai/package.yaml
+++ b/libs/metrics-wai/package.yaml
@@ -12,14 +12,16 @@ dependencies:
 - base ==4.*
 - bytestring >=0.10
 - clock >=0.6
+- containers
 - exceptions >=0.6
 - http-types >=0.8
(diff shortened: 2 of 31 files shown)


fisx:
https://github.com/wireapp/wire-server/pull/808/commits/cb0a3ab9811c9d9b307c3041d1df801489636639 is optional: I think it simplifies the code in two ways:

- it introduces a more specific type than `Either` that makes things more readable.
- The `Either` turns into a `Maybe`, and the capture variables are not even tracked any more.  This eliminates a small (but provably non-empty :)) set of possible errors.

If anybody objects to this (it will require slightly more work to get the dashboards and alerts back on their feed), I'm happy to revert it.

fisx:
ah: the capture variables are used in the handlers, and we need to change them there, too.

fisx:

PR #809, opened 2019-07-25
Fix package deps

fisx:
Develop contains some changes in `package.yaml` files that make some executables depend on packages they don't need, which can trigger dynamic linker errors in our ~~docker~~ ops setup.

This PR reverts these changes and adds some more conservative ones to get the tests added in #808 to build again.


fisx:
unit and integration tests pass locally.

PR #813, opened 2019-08-04
New feature flags backend

tiago-loureiro:
This PR intends to add an extra value to enable/disable SSO per team: current default is enabled. In addition, it moves all "flags" under a single CF since it's easier to manage and helps keeping all these "flags" in one place. Feel free to keep it as-is if you find that isolation is preferable.

TODO:
 - [x] Add integration tests for the new feature/endpoint
 - [x] Make spar check this endpoint
 - [x] Add endpoint in stern to check/set this value
 - ~~[ ] There's a TODO in the code, have a look at that too: separate PR~~

### DIFF
--- a/services/galley/src/Galley/API.hs
+++ b/services/galley/src/Galley/API.hs
@@ -921,20 +921,20 @@ sitemap = do
     -- possible internally. Viewing the status should be allowed
     -- for any admin
 
-    get "/i/teams/:tid/legalhold" (continue Teams.getLegalholdStatusInternal) $
+    get "/i/teams/:tid/features/legalhold" (continue Teams.getLegalholdStatusInternal) $
         capture "tid"
         .&. accept "application" "json"
 
-    put "/i/teams/:tid/legalhold" (continue Teams.setLegalholdStatusInternal) $
+    put "/i/teams/:tid/features/legalhold" (continue Teams.setLegalholdStatusInternal) $
         capture "tid"
         .&. jsonRequest @LegalHoldTeamConfig
         .&. accept "application" "json"
 
-    get "/i/teams/:tid/sso" (continue Teams.getSSOStatusInternal) $
+    get "/i/teams/:tid/features/sso" (continue Teams.getSSOStatusInternal) $
         capture "tid"
         .&. accept "application" "json"
 
-    put "/i/teams/:tid/sso" (continue Teams.setSSOStatusInternal) $
+    put "/i/teams/:tid/features/sso" (continue Teams.setSSOStatusInternal) $
         capture "tid"
         .&. jsonRequest @SSOTeamConfig
         .&. accept "application" "json"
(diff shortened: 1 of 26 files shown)


fisx:
> [...] it moves all "flags" under a single CF since it's easier to manage and helps [...]

What's a "CF"?

fisx:
> > [...] it moves all "flags" under a single CF since it's easier to manage and helps [...]
> 
> What's a "CF"?

ah.  column family.  very straight-forward.  :-)

fisx:
>     * [ ]  There's a TODO in the code, have a look at that too.

this is about what should happen when sso is disabled.  spar needs to stop authenticating users against registered idps, but it shouldn't have to check on every authentication, just on idp registration.  options:

1. galley should notify spar if a team is disabled.  spar should keep the idp as registered, but with a local `enabled` flag that keeps it from using it for authentication.
2. galley should refuse to disable sso in teams that have registered idps.

2. is more vulnerable to race conditions (delete idp -> disable sso -> create a new one *just* fast enough to go before the last step -> now you have a disabled team with a working idp).  1. is probably a bit more work, and trickier to deploy with the required migrations.

i think i'll aim for 1.

services/galley/src/Galley/API.hs
```diff
@@ -904,23 +906,49 @@ sitemap = do
...
+    put "/i/teams/:tid/features/sso" (continue Teams.setSSOStatusInternal) $
+        capture "tid"
+        .&. jsonRequest @SSOTeamConfig
+        .&. accept "application" "json"
+
+    -- End of team features
 
     get "/i/test/clients" (continue getClients)
         zauthUserId
+    -- TODO: What is this endpoint? Is this used anywhere?
```
jschaul: Grep or ag helps finding usage, it's used here: https://github.com/wireapp/wire-server/blob/develop/services/brig/test/integration/API/User/Client.hs#L319
Maybe you could add a comment and remove the TODO.

services/galley/test/integration/API/Teams.hs
```diff
@@ -146,6 +150,27 @@ testCreateTeamWithMembers = do
...
+
+    let check :: HasCallStack => String -> SSOStatus -> TestM ()
+        check msg enabledness = do
+          SSOTeamConfig status <- jsonBody <$> (getSSOEnabledInternal tid <!! testResponse 200 Nothing)
+          liftIO $ assertEqual msg enabledness status
+
+    check "Teams should start with SSO disabled" SSODisabled
+
+    putSSOEnabledInternal tid SSOEnabled
+    check "Calling 'putEnabled True' should enable SSO" SSOEnabled
```
jschaul: I like this, very clear.

services/galley/src/Galley/API.hs
fisx: :+1:

(will do that in #817)

services/galley/test/integration/API/Teams.hs
fisx: thanks :)

PR #816, opened 2019-08-06
Cleanup logging

arianvp:
We want to make sure that we were using the `MonadLogger` typeclass everywhere, as this will make it easy to consistently inject metadata into the logs throughout the project. 

Turns out,  most places where already fine.

I made this PR by using `sed` and then making stuff compile. So maybe some places we could have left untouched. We should go over the diff and decide what things do not need to be changed, such that the diff becomes a bit smaller

### DIFF
--- a/libs/api-client/src/Network/Wire/Client/HTTP.hs
+++ b/libs/api-client/src/Network/Wire/Client/HTTP.hs
@@ -25,7 +25,7 @@ import Network.Wire.Client.Monad
 import qualified Data.ByteString.Lazy as Lazy
 import qualified Data.Text.Encoding   as T
 import qualified Network.HTTP.Client  as Rq
-import qualified System.Logger.Class        as Log
+import qualified System.Logger.Class  as Log
 
 data Error = Error
     { code    :: Int
--- a/libs/api-client/src/Network/Wire/Client/HTTP.hs
+++ b/libs/api-client/src/Network/Wire/Client/HTTP.hs
@@ -25,7 +25,7 @@ import Network.Wire.Client.Monad
 import qualified Data.ByteString.Lazy as Lazy
 import qualified Data.Text.Encoding   as T
 import qualified Network.HTTP.Client  as Rq
-import qualified System.Logger        as Log
+import qualified System.Logger.Class        as Log
 
 data Error = Error
     { code    :: Int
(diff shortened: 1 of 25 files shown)


libs/api-client/src/Network/Wire/Client/HTTP.hs
```diff
@@ -25,7 +25,7 @@ import Network.Wire.Client.Monad
 import qualified Data.ByteString.Lazy as Lazy
 import qualified Data.Text.Encoding   as T
 import qualified Network.HTTP.Client  as Rq
-import qualified System.Logger        as Log
+import qualified System.Logger.Class        as Log
```
fisx:
```suggestion
import qualified System.Logger.Class  as Log
```

services/brig/index/src/Main.hs
```diff
@@ -5,7 +5,7 @@ import           Eval
 import           Options
 import           Options.Applicative
 import           System.Exit
-import qualified System.Logger       as Log
+import qualified System.Logger.Class       as Log
```
fisx:
```suggestion
import qualified System.Logger.Class as Log
```

services/cannon/src/Cannon/App.hs
```diff
@@ -17,7 +17,8 @@ import Network.WebSockets hiding (Request, Response, requestHeaders)
 import System.Logger.Class hiding (Error, close)
 
 import qualified Data.Text.Lazy as Text
-import qualified System.Logger  as Logger
+-- TODO(arianvp): Make this Classy. we can utilise UnliftIO here
```
fisx: Is this TODO obsolete?

services/galley/src/Galley/Run.hs
```diff
@@ -17,7 +17,7 @@ import qualified Control.Concurrent.Async           as Async
 import qualified Data.Metrics.Middleware            as M
 import qualified Network.Wai.Middleware.Gunzip      as GZip
 import qualified Network.Wai.Middleware.Gzip        as GZip
-import qualified System.Logger                      as Log
+import qualified System.Logger.Class                      as Log
```
fisx:
```suggestion
import qualified System.Logger.Class                 as Log
```

services/galley/test/integration/Main.hs
```diff
@@ -31,7 +31,7 @@ import Util.Test
 import qualified API
 import qualified API.SQS                as SQS
 import qualified Data.ByteString.Char8  as BS
-import qualified System.Logger   as Logger
+import qualified System.Logger.Class   as Logger
```
fisx:
```suggestion
import qualified System.Logger.Class    as Logger
```

this is...  fun :-)

services/spar/src/Spar/App.hs
```diff
@@ -72,12 +73,15 @@ instance HasCreateUUID Spar where
 instance HasLogger Spar where
   -- FUTUREWORK: optionally use 'field' to index user or idp ids for easier logfile processing.
   logger (toLevel -> lv) mg = do
+    let mg'    = Log.msg . condenseLogMsg . cs $ mg
+    log lv mg'
```
fisx: This is correct, but while you're at it, could you:

```haskell
  logger lv = log (toLevel lv) . Log.msg . cs
```

`condenseLogMsg` should be deleted.  That's from an older version of tinylog that didn't handle newlines in log messages.

PR #819, opened 2019-08-07
Make develop great again!

fisx:
Fixes #816


fisx:
@arianvp if(f!) CI is green and you agree with everything, you can merge yourself.

PR #823, opened 2019-08-08
Block legalhold on stern / backoffice.

fisx:
Revert this commit if once LH tokens are fully implemented and ready for production use.


fisx:
(the problem with making ci passing mandatory is that signalling between concourse and github doesn't always work...)

PR #825, opened 2019-08-09
Feature Flags in galley options.

fisx:
SSO is likely to become an enterprise feature in the future, but for now it is in free beta, and everybody can use it without contacting us first.

This has been agreed with @tiago-loureiro and @marcoconti83.


jschaul:
Motivation for this change is unclear, please document.

fisx:
hum, can somebody see the failing test case in concourse?  i can't.  i'll run this locally later.

jschaul:
Thanks for the explanation.

If SSO should be disabled or enabled in different deployment environments, and changes from one day to the next (as it has done this week here), then the default should be in configuration, not in code. Can you add that to the configuration for galley instead?

jschaul:
> hum, can somebody see the failing test case in concourse? i can't. i'll run this locally later.

patience when opening the log helps. For your convenience:

```	
      GET, PUT [/i]?/teams/{tid}/legalhold:                            FAIL (0.21s)	
        test/integration/API/Teams/LegalHold.hs:504:	
        Teams should start with LegalHold disabled	
        expected: LegalHoldEnabled	
         but got: LegalHoldDisabled
```

fisx:
>     1. Add SSO default to config, not code.

I'd rather not.  @tiago-loureiro could you be the tie-breaker?

>     2. You meant to change SSO, not Legalhold

Oops.  I was particularly tired today, sorry.  Thanks.

fisx:
tests are not passing yet locally, but i don't understand why.  i'll look into it later.

fisx:
there are at least these reason why we should make it a config variable:

- we're very unlikely to change it more than once in the future.  it's in beta now (which i wasn't aware of, hence the back-and-forth), but once it's not, it'll be in `default: off` for good.
- turning it on requires to re-run the migration script (see 7c08142f78380eccb3f7f82302f7cfdc65199b0b, so good that you raised this issue, thanks!)
- it feels inconsistent to give some feature flags config varialbes, but not others.  and while there may be on-prem use cases for having legalhold enabled by default, i don't thinkg we want to support out of the box.

jschaul:
Opening the concourse page and waiting ~30-60 seconds loads for me. Does it not load for you?

```
      FAIL
08:05:56
	
        Exception: Assertions failed:
08:05:56
	
         2: Right (SSOTeamConfig {ssoTeamConfigStatus = SSOEnabled}) =/= Right (SSOTeamConfig {ssoTeamConfigStatus = SSODisabled})
08:05:56
	
        
08:05:56
	
        Response was:
08:05:56
	
        
08:05:56
	
        Response {responseStatus = Status {statusCode = 200, statusMessage = "OK"}, responseVersion = HTTP/1.1, responseHeaders = [("Transfer-Encoding","chunked"),("Date","Sun, 11 Aug 2019 06:01:55 GMT"),("Server","Warp/3.2.25"),("Content-Encoding","gzip"),("Content-Type","application/json")], responseBody = Just "{\"status\":\"enabled\"}", responseCookieJar = CJ {expose = []}, responseClose' = ResponseClose}
08:05:56
	
08:05:56
	
          error, called at src/Bilge/Assert.hs:69:30 in bilge-0.22.0-8ufUi02ZiYj1pTUQT3VQuV:Bilge.Assert
08:05:56
	
          <!!, called at src/Bilge/Assert.hs:86:19 in bilge-0.22.0-8ufUi02ZiYj1pTUQT3VQuV:Bilge.Assert
08:05:56
	
          !!!, called at test/integration/API/Teams.hs:1161:27 in main:API.Teams
08:05:56
	
          getSSO, called at test/integration/API/Teams.hs:1175:9 in main:API.Teams
```

> turning it on requires to re-run the migration script

Instead of having to run a migration **again** at some point in the future, what prevents spar from calling the internal endpoint (to enable the feature flag explicitly) whenever an idp is created (and removing that flag whenever all idps for a team are deleted)? That would prevent needing to run a migration in the future. Not only we need to run migrations, all self-hosters using SSO will need to. If we can avoid needing these migrations, let's please try to avoid them.

Main reason why to have a config flag for default use of SSO:

- different on-premise installations will follow a different logic (they either want idp creation to be just possible, or just not possible at all, or to be disabled by default but enabled manually). As soon as whatever the wish of an on-premise installation A is different to the wish of an on-premise installation B (or either of them are different than the then-default for wire.com), there will be a problem that cannot be solved by changing that line in the code - configuration will need to be introduced. 
Following suit with "better do things all the way to the end right away instead of leaving something partially-done that has to be touched again, having to spend much more time in the future", my opinion is that a configuration flag should be introduced as part of this PR.

tiago-loureiro:
I think it makes sense to have that extra config flag - we can always have it defaulted to "something" but being able to override it without code changes sounds like a good idea to me.

fisx:
Ok, I'm going to implement the config flag.  Just for the record: this means that if somebody has it enabled by default, and then decides to disable it by default, they need to run the migration, or the system will enter an inconsistent state.  I'll try to mention this everywhere.

fisx:
(re. ci: no, that part of the output was swallowed even after the page was done loading and rendering)

fisx:
the integration tests work for both all feature flag settings now.  all features should be turned on on the ci, but for the future, they should be run with different feature flags enabled (see https://github.com/wearezeta/backend-issues/issues/869)

services/galley/src/Galley/API/Teams.hs
```diff
@@ -511,10 +511,13 @@ getLegalholdStatus (uid ::: tid ::: ct) = do
 -- | Get legal SSO status for a team.
```
jschaul:
```suggestion
-- | Get SSO status for a team.
```

fisx:
```
service: user error (AesonException "Error in $.settings: When parsing the record Settings of type Galley.Options.Settings the key featureFlags was not present.")
```

sigh.  i'll fix this.

fisx:
some tests expect legalhold to be enabled.  my fault, will fix.

fisx:
WARNING! `assertLegalHoldEnabled` from https://github.com/wireapp/wire-server/pull/761/files

services/galley/src/Galley/API/Teams.hs
```diff
@@ -512,8 +512,8 @@ getLegalholdStatus (uid ::: tid ::: ct) = do
 getSSOStatusInternal :: TeamId ::: JSON -> Galley Response
 getSSOStatusInternal (tid ::: _) = do
     defConfig <- do
-        FeatureFlags flags <- view (options . optSettings . setFeatureFlags)
-        pure $ if FeatureSSO `elem` flags
+        yes <- view (options . optSettings . featureEnabled FeatureSSO)
+        pure $ if yes
```
jschaul: `if yes` reads a bit like `if true`, but it's sometimes true and sometimes not, so it's not a good variable name. Suggestion: Rename `yes` to `featureSSO`.

services/galley/src/Galley/API/Teams.hs
```diff
@@ -540,8 +540,8 @@ getLegalholdStatusInternal (tid ::: _) = do
 -- | Enable or disable legal hold for a team.
 setLegalholdStatusInternal :: TeamId ::: JsonRequest LegalHoldTeamConfig ::: JSON -> Galley Response
 setLegalholdStatusInternal (tid ::: req ::: _) = do
-    do  FeatureFlags flags <- view (options . optSettings . setFeatureFlags)
-        unless (FeatureLegalHold `elem` flags) $ throwM legalHoldFeatureFlagNotEnabled
+    do  yes <- view (options . optSettings . featureEnabled FeatureLegalHold)
```
jschaul: same here - rename to `featureLegalHold` ?

services/galley/src/Galley/Options.hs
```diff
@@ -21,12 +21,19 @@ data Settings = Settings
...
     -- | URI prefix for conversations with access mode @code@
     , _setConversationCodeURI   :: !HttpsUrl
-    , _setFeatureFlags          :: !FeatureFlags
+    , _setFeatureFlags          :: !(Maybe FeatureFlags)
     } deriving (Show, Generic)
 
 deriveFromJSON toOptionFieldName ''Settings
 makeLenses ''Settings
 
+featureEnabled :: FeatureFlag -> Getter Settings Bool
```
jschaul: :+1:

fisx:
my fault again:

```
Failures:
  test-integration/Test/Spar/APISpec.hs:655:11: 
  1) Spar.API, POST /identity-providers, sso disabled for team, responds with 403 forbidden
      predicate failed on: Response {responseStatus = Status {statusCode = 201, statusMessage = "Created"}, responseVersion = HTTP/1.1, responseHeaders = [("Transfer-Encoding","chunked"),("Date","Thu, 29 Aug 2019 18:35:05 GMT"),("Server","Warp/3.2.25"),("Cache-Control","no-cache, no-store"),("Pragma","no-cache"),("Content-Type","application/json;charset=utf-8")], responseBody = Just "{\"id\":\"126e70ee-a374-44e8-b4c1-8e2adf16597c\",\"metadata\":{\"issuer\":\"https://issuer.net/_6b7d8905-2f7f-4549-ae62-597905e55158\",\"requestURI\":\"https://requri.net/\",\"certAuthnResponse\":[\"<ds:KeyInfo xmlns:ds=\\\"http://www.w3.org/2000/09/xmldsig#\\\"><ds:X509Data><ds:X509Certificate>MIIB/zCCASigAwIBAgIOEyBZjWrTHqmgBPAVkAUwDQYJKoZIhvcNAQELBQAwADAeFw0xODA4MTYxMjQzNTdaFw0zODA4MTExMjQzNTdaMAAwgd0wDQYJKoZIhvcNAQEBBQADgcsAMIHHAoHBAMHb4Ne1z2cQD1TXcVmYBy0Q1EnmQl5IncCfC6/eGrp0qpa5sqaQPlRtvS3UEczpAgf9ml+kL6aK56xEBH2Zv/mlkvBEbxASxVha3LhcIg9TNAg0vm2KJBG1pZvHx8OIKhpDCfabkSJF+MxXvtTrp0JTRfQr2BHkegZNX3hCaF5JGyGIMBinTRwEi5duDfNUsJoG5MwNq/hrd7pLdjOWgs4CLlNV6L+3rvhhYt+e0QUeh9QrZFUfhXxezlfYfP36WQIBETANBgkqhkiG9w0BAQsFAAOBwQBtZqvROSfV1znZws9h6M749g1HRpm3vub3RKAZOWfqP2Qag2ML+BjAqEIH1SAaQSZlFbKRsKM2Bp/QpG5ByshwrxoS9ausPNynulMA7dEPvWOExfqYO9Vj/0ejxwAmilseKrVfv333yvcgVRNRqP/LMxqe/8Hw3Ax+Ul83usIZLQ5m4sW9/IUVwlDLk31ddIkPVpx2USKL9eVDXjVhIl7itgJxPyG0wc0I9Ad/ZWy/Dbbilwz1tHcZSZsxSdNFW+k=</ds:X509Certificate></ds:X509Data></ds:KeyInfo>\"]},\"extraInfo\":\"6fdd09a2-7007-4cb4-99e5-65e7f0a4c2aa\"}", responseCookieJar = CJ {expose = []}, responseClose' = ResponseClose}
```

PR #826, opened 2019-08-09
Misc Cleanup

fisx:
Please consult the commit messages for more details.

### DIFF
--- a/libs/api-bot/package.yaml
+++ b/libs/api-bot/package.yaml
@@ -1,4 +1,4 @@
-defaults: 
+defaults:
   local: ../../package-defaults.yaml
 name: api-bot
 version: '0.4.2'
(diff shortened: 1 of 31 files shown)


fisx:
> All commits other than `Remove 'exposed-modules' sections from all package.yaml files. ` are fine to me.
> 
> Though generally I understood that it should be up to the library author to decide what is internal and what should be exposed. Exposing everything globally can inadvertently create weird dependencies between modules in the future (e.g. a part of brig depending on an internal module of galley, rather than proper refactoring moving shared code to a shared library), and doesn't restrict people from using functions that their author intended to keep private to that library for good reason.
> 
> To me, the extra effort of adding a line (which ghc clearly tells you to do when you forget) when creating a new module isn't very high, and the risk of misuse of internal functions inadvertently is much higher.
> 
> Can you take that commit out again please? (Or get someone else's opinion)

can you give an (even hypothetical) example where an internal module would create such a problem?  i have only ever had problems with modules that were internal, and that i wanted to use for very good reason, and i never could imagine anything going wrong on changes in the library that wouldn't be caught by the type checker.

but i'm not very imaginative, so please enlighten me.  :)

any other opinions?

if this question doesn't change your mind, i'll move this discussion to a separate PR.  (it's not something i want to fight over for hours, though.)

arianvp:
I dont have a very strong opinion about the exposed-module  stuff. I think I am slightly more in favor for listing them explicitly. So perhaps we should  move that part in a separate PR and bikeshed about that a bit more

PR #828, opened 2019-08-14
Add more debug logging

arianvp:
This adds debug logs for operations that directly affect a user which are useful to have.  when running in info log level, these actions aren't logged.

This is part 1 of a a series of pull requests.
In part 2 i'm planning to add a structured json log format (next to netstr and plan). Which should ease integration of log aggregators like fluent-bit.

I thought it's a good idea to get this part reviewed already as the second part is clearly standalone.


services/galley/src/Galley/API/LegalHold.hs
```diff
@@ -42,6 +42,9 @@ createSettings (zusr ::: tid ::: req ::: _) = do
     assertLegalHoldEnabled tid
 
     membs <- Data.teamMembers tid
+    Log.debug $ Log.field "targets" (toByteString (show membs))
+              . Log.msg (Log.val "LegalHold.createSettings")
```
tiago-loureiro: I think using the `show` instance for `TeamMember` will create a log of noise:
```
data TeamMember = TeamMember
    { _userId          :: UserId
    , _permissions     :: Permissions
    , _invitation      :: Maybe (UserId, UTCTimeMillis)
    , _legalHoldStatus :: UserLegalHoldStatus
    } deriving (Eq, Ord, Show, Generic)
```
we should log the affected user id(s), comma separated.

Worth using a separate function altogether and move the logging details there? (i.e., the log level/field, etc.)

services/galley/src/Galley/API/LegalHold.hs
```diff
@@ -164,6 +174,8 @@ approveDevice
     -> Galley Response
 approveDevice (zusr ::: tid ::: uid ::: connId ::: req ::: _) = do
     assertLegalHoldEnabled tid
+    Log.debug $ Log.field "target" (toByteString uid)
```
tiago-loureiro: For consistency sake I'd simply use `target` here too (even though it's just one)

services/galley/src/Galley/API/Teams.hs
```diff
@@ -112,6 +113,8 @@ createNonBindingTeam (zusr::: zcon ::: req ::: _) = do
     let zothers = map (view userId) others
     ensureUnboundUsers (zusr : zothers)
     ensureConnected zusr zothers
+    Log.debug $ Log.field "targets" (toByteString (show zothers))
+              . Log.msg (Log.val "createNonBindingTeam")
```
tiago-loureiro: I liked how you scoped `Legalhold...`. How about using `Teams.createNonBindingTeam` here too?

arianvp: I agree but I am too lazy to do this manually, but I have convinced GHC to do it for us instead.  (By stealing some code from the super nice co-log library): here is a small protoype:

```haskell
{-# LANGUAGE ConstraintKinds #-}

import Data.Monoid
import GHC.Stack
import Prelude hiding (log)

data Level = Debug | Info | Warn | Error  deriving Show
showSourceLoc :: CallStack -> String
showSourceLoc cs = showCallStack
  where
    showCallStack :: String
    showCallStack = case getCallStack cs of
        []                             -> "<unknown loc>"
        [(name, loc)]                  -> "lol" -- showLoc name loc
        (_, loc) : (callerName, _) : _ -> showLoc callerName loc
    showLoc :: String -> SrcLoc -> String
    showLoc name SrcLoc{..} =
        srcLocModule <> "." <>  name <> "#" <>  (show srcLocStartLine)
        
class Monad m => MonadLogger m where
    logMsg :: CallStack -> Level -> String -> m ()
    
instance MonadLogger IO where
    logMsg callStack level msg = do
      putStrLn $ "[" ++ show level ++ "] " ++ showSourceLoc callStack ++ " " ++ msg
type MonadLogger' m = (MonadLogger m, HasCallStack)

log :: MonadLogger' m => Level -> String -> m ()
log = withFrozenCallStack (logMsg callStack)

smelt :: MonadLogger' m => m ()
smelt = do
  log Error "NOOO"
  log Warn "Maybe"
  kaas

kaas :: MonadLogger' m => m ()
kaas = do
  log Info "I like cheese"
 
main :: IO ()
main = do
  kaas
  smelt

```

```
[Info] Main.kaas#40 I like cheese
[Error] Main.smelt#34 NOOO
[Warn] Main.smelt#35 Maybe
[Info] Main.kaas#40 I like cheese
```

services/galley/src/Galley/API/LegalHold.hs
arianvp: oh yeh that was an oversight.  I should be logging the userIds here not the TeamMember

services/galley/src/Galley/API/LegalHold.hs
fisx: you probably mean `targets`?  plural?

tiago-loureiro: > you probably mean targets? plural?

yes

services/galley/src/Galley/API/Teams.hs
arianvp: I ended up doing this manually . still think this is a cute thing to have though; ill open an issue for it

services/galley/src/Galley/API/LegalHold.hs
```diff
@@ -71,8 +75,12 @@ getSettings (zusr ::: tid ::: _) = do
 removeSettings :: UserId ::: TeamId ::: JsonRequest RemoveLegalHoldSettingsRequest ::: JSON -> Galley Response
 removeSettings (zusr ::: tid ::: req ::: _) = do
     assertLegalHoldEnabled tid
-
     membs <- Data.teamMembers tid
+    let zothers = map (view userId) membs
+    -- TODO user id's of members
```
tiago-loureiro: What is this `TODO` about, can be removed - right?

services/galley/src/Galley/API/LegalHold.hs
```diff
@@ -42,6 +42,10 @@ createSettings (zusr ::: tid ::: req ::: _) = do
     assertLegalHoldEnabled tid
 
     membs <- Data.teamMembers tid
+    let zothers = map (view userId) membs
+    Log.debug $ Log.field "targets" (toByteString (show zothers))
```
fisx:
```suggestion
    Log.debug $ Log.field "targets" (show $ toByteString <$> zothers)
```

fisx: (you're still using show on the `Id` type, which is fine for debugging, but the `ToByteString` instance is more concise.)

(speaking of which: why debug?  are customers going to want to use this?  then it should at least be `Info`.)

tiago-loureiro: Don't do `show . toByteString`, `toByteString` is sufficient.

libs/extended/src/System/Logger/Extended.hs
```diff
@@ -18,12 +18,19 @@ import qualified Data.ByteString.Lazy.Char8 as L
 import qualified Data.ByteString.Lazy.Builder as B
 import qualified System.Logger.Class as LC
 
+-- TODO(arianvp): Get rid of boolean blindness
```
arianvp: These TODOs are removed in the next PR btw

PR #830, opened 2019-08-15
Limit login retries

fisx:
Implements https://github.com/wearezeta/backend-issues/issues/841#issuecomment-518033685 .

Open questions:

1. I use `forM_` instead of `pooledForConcurrentlyN_` [here](https://github.com/wireapp/wire-server/blob/33a66e5732404bc8de91d9329dd3347d8d5be622/services/brig/test/integration/API/User/Auth.hs#L246).  The latter will fail because `Bridge.Budget` is not concurrency-proof: if two threads / services request a budget coin concurrently, only one of them will decrement the budget value in cassandra.  We could use [counters](https://cassandra.apache.org/doc/latest/cql/types.html#counters) instead of `int` in the [schema](https://github.com/wireapp/wire-server/blob/33a66e5732404bc8de91d9329dd3347d8d5be622/services/brig/schema/src/V37.hs), but it is unclear what that will mean for the other use case.  (cases?  i think there is only one.)
2. Should I move [LoginRetryOpts](https://github.com/wireapp/wire-server/blob/33a66e5732404bc8de91d9329dd3347d8d5be622/services/brig/src/Brig/Options.hs#L174) to [Settings](https://github.com/wireapp/wire-server/blob/33a66e5732404bc8de91d9329dd3347d8d5be622/services/brig/src/Brig/Options.hs#L274)?  It's not clear to me what the difference is, except perhaps that `Settings` can be changed at run-time, in which case `LoginRetryOpts` shouldn't go there (I think).

### DIFF
--- a/services/brig/brig.integration.yaml
+++ b/services/brig/brig.integration.yaml
@@ -104,10 +104,6 @@ emailSMS:
     tCreatorWelcomeUrl: http://127.0.0.1:8080/creator-welcome-website
     tMemberWelcomeUrl: http://127.0.0.1:8080/member-welcome-website
 
-loginRetry:
-  loginRetryTimeout: 5
-  loginRetryLimit: 5
-
 zauth:
   privateKeys: test/resources/zauth/privkeys.txt
   publicKeys: test/resources/zauth/pubkeys.txt
@@ -141,6 +137,9 @@ optSettings:
   setUserCookieThrottle:
     stdDev: 5
     retryAfter: 1
+  limitFailedLogins:
+    timeout: 5  # seconds.  if you reach the limit, how long do you have to wait to try again.
+    retryLimit: 5  # how many times can you have a failed login in that timeframe.
   setRichInfoLimit: 5000  # should be in sync with Spar
   setDefaultLocale: en
   setMaxTeamSize: 32
--- a/services/brig/brig.integration.yaml
+++ b/services/brig/brig.integration.yaml
@@ -104,6 +104,10 @@ emailSMS:
     tCreatorWelcomeUrl: http://127.0.0.1:8080/creator-welcome-website
     tMemberWelcomeUrl: http://127.0.0.1:8080/member-welcome-website
 
+loginRetry:
+  loginRetryTimeout: 5
+  loginRetryLimit: 5
+
 zauth:
   privateKeys: test/resources/zauth/privkeys.txt
   publicKeys: test/resources/zauth/pubkeys.txt
--- a/services/brig/brig.integration.yaml
+++ b/services/brig/brig.integration.yaml
@@ -137,7 +137,7 @@ optSettings:
   setUserCookieThrottle:
     stdDev: 5
     retryAfter: 1
-  limitFailedLogins:
+  setLimitFailedLogins:
     timeout: 5  # seconds.  if you reach the limit, how long do you have to wait to try again.
     retryLimit: 5  # how many times can you have a failed login in that timeframe.
   setRichInfoLimit: 5000  # should be in sync with Spar
--- a/services/brig/src/Brig/User/Auth.hs
+++ b/services/brig/src/Brig/User/Auth.hs
@@ -104,7 +104,7 @@ withRetryLimit
     -> UserId
     -> ExceptT LoginError AppIO ()
 withRetryLimit action uid = do
-    mLimitFailedLogins <- view (settings . to Opt.limitFailedLogins)
+    mLimitFailedLogins <- view (settings . to Opt.setLimitFailedLogins)
     forM_ mLimitFailedLogins $ \opts -> do
         let bkey = BudgetKey ("login#" <> idToText uid)
             budget = Budget
--- a/services/brig/src/Brig/User/Auth.hs
+++ b/services/brig/src/Brig/User/Auth.hs
@@ -17,7 +17,7 @@ module Brig.User.Auth
 
 import Imports
 
-import Control.Lens (view)
+import Control.Lens (view, to)
 import Brig.App
 import Brig.API.Types
 import Brig.Budget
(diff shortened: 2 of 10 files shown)


fisx:
(I'll make a PR to wire-server-deploy soon, but since the flag is optional, that can be done independently.)

fisx:
> (I'll make a PR to wire-server-deploy soon, but since the flag is optional, that can be done independently.)

heh :) I guess i missed one after all:

```
          Exception: test/integration/API/User/Auth.hs:235:9-44: Irrefutable pattern failed for pattern Just opts
```

services/brig/brig.integration.yaml
```diff
@@ -104,6 +104,10 @@ emailSMS:
     tCreatorWelcomeUrl: http://127.0.0.1:8080/creator-welcome-website
     tMemberWelcomeUrl: http://127.0.0.1:8080/member-welcome-website
 
+loginRetry:
+  loginRetryTimeout: 5
+  loginRetryLimit: 5
+
```
tiago-loureiro: Can we move this inside `optSettings` and named them something in the lines of:

```
optSettings:
  limitFailedLogins:
    timeout: 300 # if you reach the limit, how long do you have to wait to try again (let's use seconds)
    retryLimit: 5 # how many times can you have a failed login in that timeframe
```

services/brig/src/Brig/User/Auth.hs
```diff
@@ -71,19 +75,50 @@ lookupLoginCode phone = Data.lookupKey (userPhoneKey phone) >>= \case
...
+
+checkRetryLimit :: UserId -> ExceptT LoginError AppIO ()
+checkRetryLimit = withRetryLimit checkBudget
+
+withRetryLimit
+    :: (BudgetKey -> Budget -> ExceptT LoginError AppIO (Budgeted ()))
+    -> UserId
+    -> ExceptT LoginError AppIO ()
+withRetryLimit action uid = do
+    () <- traceShow 'a' $ pure ()
```
tiago-loureiro: This can be removed now?

services/brig/src/Brig/User/Auth.hs
```diff
@@ -71,19 +75,50 @@ lookupLoginCode phone = Data.lookupKey (userPhoneKey phone) >>= \case
...
+    mLoginRetryOpts <- view loginRetry
+    () <- traceShow mLoginRetryOpts $ pure ()
```
tiago-loureiro: This can be removed now?

services/brig/brig.integration.yaml
```diff
@@ -137,6 +137,9 @@ optSettings:
   setUserCookieThrottle:
     stdDev: 5
     retryAfter: 1
+  limitFailedLogins:
```
tiago-loureiro: sry, can u make it `limitFailedLogins` -> `setLimitFailedLogins` for consistency's sake? ☺️

fisx:
(this will pass CI once the resp. PRs there have been deployed.)

services/brig/src/Brig/User/Auth.hs
```diff
@@ -71,19 +75,46 @@ lookupLoginCode phone = Data.lookupKey (userPhoneKey phone) >>= \case
...
         AuthSuspended          -> throwE LoginSuspended
         AuthEphemeral          -> throwE LoginEphemeral
-        AuthInvalidCredentials -> throwE LoginFailed
-        AuthInvalidUser        -> throwE LoginFailed
+        AuthInvalidCredentials -> loginFailed uid
+        AuthInvalidUser        -> loginFailed uid
     newAccess uid typ label
 login (SmsLogin phone code label) typ = do
     uid <- resolveLoginId (LoginByPhone phone)
+    checkRetryLimit uid
```
tiago-loureiro: We could also remove this check, no? It's anyways checked when calling `loginFailed`, right?

services/brig/test/integration/API/User/Auth.hs
```diff
@@ -227,7 +230,51 @@ testThrottleLogins conf b = do
...
+    -- throttling should stop and login should work again
+    do  let Just retryAfterSecs = fromByteString =<< getHeader "Retry-After" resp
+            retryTimeout = Opts.Timeout $ fromIntegral retryAfterSecs
+        liftIO $ do
+            assertEqual "throttle delay (1)" retryTimeout (Opts.timeout opts)
+            threadDelay (1000000 * (retryAfterSecs - 1))
+
+    -- fail again later into the block time window
+    rsp <- login brig (defEmailLogin email) SessionCookie <!! const 403 === statusCode
+    do  let Just retryAfterSecs = fromByteString =<< getHeader "Retry-After" rsp
```
tiago-loureiro: In principle, `retryAfterSecs` has a value of ~1 here, right? Won't `retryAfterSecs - 2` risk being negative? What will happen with `threadDelay` if given a negative value? My empirical tests show that `threadDelay` with a negative value == no-op so I guess this is fine given that you assert in the line above.

services/brig/src/Brig/User/Auth.hs
fisx: Not sure I understand, but let me try: if you only do the check in `loginFailed`, then after having exhausted your retry limit you can still successfully login.  You just need to eventually guess the password correctly.  So the check here is necessary.

Does that qualify as an answer?  Does it resolve anything?

services/brig/test/integration/API/User/Auth.hs
fisx: does https://github.com/wireapp/wire-server/pull/830/commits/eec445408881a39bef043c089f91eb2afcca2d7d answer your question?

(it's also a lot more correct, so thanks for the question.  :))

fisx: does https://github.com/wireapp/wire-server/pull/830/commits/eec445408881a39bef043c089f91eb2afcca2d7d answer your question?

(it's also a lot more correct, so thanks for the question.  :))

services/brig/src/Brig/User/Auth.hs
tiago-loureiro: My point was: do we even need to call this `checkRetryLimit uid` here at all? In fact, is that function needed at all?

Calling `loginFailed` will call `decrRetryLimit` which will in turn reduce the budget in case of a failed login, right? And when that happens, we'll anyway throw `LoginBlocked`. Similarly as to how we do not explicitly check for the budget with SMS/Call

fisx: summary of phone call: yes, both `checkRetryLimit` and `loginFailed` *in principle* might throw out-of-budget, but the latter is only ever called right after the former, so it won't.  the reason we need both is that `loginFailed` is only called in case of budget exhaustion or bad password, and we need to at least check for budget exhaustion in order for that to fail.  :-)

i welcome any ideas how to model this better!

PR #831, opened 2019-08-15
Suspend inactive users

fisx:
Implements https://github.com/wearezeta/backend-issues/issues/841#issuecomment-516414296 .


services/brig/src/Brig/Options.hs
```diff
@@ -318,3 +329,13 @@ instance FromJSON Opts
 
 Lens.makeLensesFor [("optSettings", "optionSettings")] ''Opts
 Lens.makeLensesFor [("setEmailVisibility", "emailVisibility")] ''Settings
+
+-- | TODO: this should check cookie *age*, not *renew age*...  but where is that?
+validateOpts :: MonadThrow m => Opts -> m ()
+validateOpts opts = do
```
tiago-loureiro: That option is named `userTokenTimeout`

fisx:
Not sure this will pass; if it does, I'm not sure it always will in the future; and with >30s it's certainly not a fast test.  :(

services/brig/test/integration/API/User/Auth.hs
```diff
@@ -439,6 +442,57 @@ testNewSessionCookie config b = do
         const 200     === statusCode
         const Nothing === getHeader "Set-Cookie"
 
+testSuspendInactiveUsers :: HasCallStack => Maybe Opts.Opts -> Brig -> Http ()
```
fisx: question for the reviewer: this is currently taking >30s, and i'm not sure it'll work reliably on different machines and timing contexts.  is there anything we could do about this?

(for once, i'm tempted to change the config back from 29s to maybe 15s, and if concourse makes it pass twice we'll merge it and hope for the best.)

services/brig/src/Brig/User/Auth.hs
```diff
@@ -143,8 +148,17 @@ revokeAccess u pw cc ll = do
 --------------------------------------------------------------------------------
 -- Internal
 
+catchSuspendInactiveUser :: UserId -> e -> ExceptT e AppIO ()
+catchSuspendInactiveUser uid errval = do
+  mustsuspend <- lift $ mustSuspendInactiveUser uid
+  Log.debug . Log.msg $ show (uid, mustsuspend)
```
tiago-loureiro: Can we instead use the `user` field here to log the uuid

services/brig/src/Brig/User/Auth/Cookie.hs
```diff
@@ -110,6 +110,30 @@ renewCookie old = do
...
+        let suspendHere :: UTCTime
+            suspendHere = addUTCTime (-suspendAge) now
+
+            youngEnough :: Cookie () -> Bool
+            youngEnough = (>= suspendHere) . cookieCreated
+
+        ckies <- listCookies uid []
+        let mustSuspend
+              | null ckies            = False
+              | all youngEnough ckies = False
```
tiago-loureiro: isn't it `any` here?

services/brig/src/Brig/User/Auth/Cookie.hs
```diff
@@ -110,6 +110,30 @@ renewCookie old = do
     DB.insertCookie u old' (Just (DB.TTL (fromIntegral ttl)))
     return new
 
+-- | Whether a user has not renewed any of her cookies for longer than
+-- 'suspendCookiesOlderThanSecs'.  Call this always before 'newCookie', 'nextCookie',
+-- 'newCookieLimited' if there is a chance that the user should be suspended (we don't do it
+-- implicitly because of cyclical dependencies).
+mustSuspendInactiveUser :: UserId -> AppIO Bool
+mustSuspendInactiveUser uid = view (settings . to setSuspendInactiveUsers) >>= \case
+    Nothing -> pure True
```
tiago-loureiro: Without anything set on the config the users _always_ get suspended?

fisx: yes.

services/brig/src/Brig/User/Auth/Cookie.hs
fisx: yes.

services/brig/src/Brig/User/Auth.hs
fisx: yes.

jschaul:
There should be a feature flag for this (like described in the issue), and this behaviour should be disabled by default, and only enabled on installations where required.
 
The problem with enabling this by default is that there is no good UI to deal with suspended accounts - people will just be faced with a screen where they are unable to login, without further information (I think), or perhaps a link to customer support (in the best case). This creates extra, unnecessary work for customer support to re-activate accounts, in cases where users stop using Wire for a few months and try to log in again later. Users need to contact support, get an answer (which can take some time).

One alternative would be to only have this behaviour for team/business users, not personal users.

fisx:
> There should be a feature flag for this (like described in the issue), and this behaviour should be disabled by default, and only enabled on installations where required.

[agreed, and that's what i did](https://github.com/wireapp/wire-server/blob/15be317b6e3dd60db3bb2101edaac19989db78f4/services/brig/brig.integration.yaml#L143-L144).

Perhaps we need to make it somehow appearent that this part of the config is optional?  I am reluctant to add another flag that makes this flag be ignored, though.  I think we should solve it with better comments somehow.  What do you think @jschaul?

jschaul:
> > There should be a feature flag for this (like described in the issue), and this behaviour should be disabled by default, and only enabled on installations where required.
> 
> [agreed, and that's what i did](https://github.com/wireapp/wire-server/blob/15be317b6e3dd60db3bb2101edaac19989db78f4/services/brig/brig.integration.yaml#L143-L144).
> 

Ah, I didn't realize that. Sorry!


> Perhaps we need to make it somehow appearent that this part of the config is optional? I am reluctant to add another flag that makes this flag be ignored, though. I think we should solve it with better comments somehow. What do you think @jschaul?

It's currently quite hard to understand brig configuration though, with some specific parts that can be optional, and if unset the behaviour changes. The featureFlag list with true/false is clearer to read and understand, but you're right, adding another flag to enable/disable can feel a little like duplication. This is a more general problem for all sorts of behaviour that can be turned on or off via configuration. Since we have configuration files in multiple places, it's easy to forget one place to add a configuration, or to add a commented-out piece of configuration with a comment on what it does to comment it in. 
Perhaps we need to have a reference configuration file with all the possible options shown with comments? Not sure how to best solve that.

For this PR, could you add a comment to the brig integration config file?

fisx:
> For this PR, could you add a comment to the brig integration config file?

sure!

fisx:
```
Gundeck
  API tests
    Push
      Register a user:                                             2019-08-29T13:52:40Z, D, Connecting to 10.233.67.169:9042
2019-08-29T13:52:40Z, I, Known hosts: [datacenter1:rack1:10.233.67.169:9042]
2019-08-29T13:52:40Z, I, New control connection: datacenter1:rack1:10.233.67.169:9042#11
OK (1.01s)
      Delete a user:                                               2019-08-29T13:52:42Z, D, Connection established: datacente
r1:rack1:10.233.67.169:9042#15
OK (1.07s)
      Replace presence:                                            2019-08-29T13:52:42Z, D, Connection established: datacente
r1:rack1:10.233.67.169:9042#16
OK (1.01s)
      Remove stale presence:                                       FAIL
        Exception: ESC[93mAssertions failed:
        ESC[m 1: ESC[91m0 =/= 1ESC[m
        
        Response was:
        
        Response {responseStatus = Status {statusCode = 200, statusMessage = "OK"}, responseVersion = HTTP/1.1, responseHeaders = [("Transfer-Encoding","chunked"),("Date","Thu, 29 Aug 2019 13:52:54 GMT"),("Server","Warp/3.2.25"),("Content-Encoding","gzip"),("Content-Type","application/json")], responseBody = Just "[{\"device_id\":\"4039383081\",\"client_id\":null,\"resource\":\"http://cannon-0.cannon.test-usyvq8y5sogr:8080/i/push/baf7e8a0-d05f-41ae-adb7-37e60d9a2a7f/4039383081\",\"created_at\":1567086764651,\"user_id\":\"baf7e8a0-d05f-41ae-adb7-37e60d9a2a7f\"}]", responseCookieJar = CJ {expose = []}, responseClose' = ResponseClose}
          error, called at src/Bilge/Assert.hs:69:30 in bilge-0.22.0-8ufUi02ZiYj1pTUQT3VQuV:Bilge.Assert
          <!!, called at src/Bilge/Assert.hs:86:19 in bilge-0.22.0-8ufUi02ZiYj1pTUQT3VQuV:Bilge.Assert
          !!!, called at test/integration/API.hs:832:5 in main:API
          ensurePresent, called at test/integration/API.hs:163:5 in main:API
```

can't reproduce locally.  i think that's a flaky test.

PR #832, opened 2019-08-16
Remove `exposed-modules` sections from all package.yaml files.

fisx:
(with these sections, we need to manually add new modules to expose them; without, everything gets exposed by default.)

cherry-picked from https://github.com/wireapp/wire-server/pull/826 because controversial.

### DIFF
--- a/libs/api-bot/package.yaml
+++ b/libs/api-bot/package.yaml
@@ -1,4 +1,4 @@
-defaults: 
+defaults:
   local: ../../package-defaults.yaml
 name: api-bot
 version: '0.4.2'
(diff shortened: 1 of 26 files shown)


arianvp:
I bikeshedded about it a bit more and I am in favor of this.

fisx:
I also went through all the cabal files on `develop` one more time.  There are generally few modules that are not exposed, and the policy on what to expose is clearly inconsistent.  So I'm still convinced this is a good idea, and making identifiers internal (eg., because a type should only be ever constructed by a smart constructor, and never pattern-matched against) should be the module boundary, not the package boundary.

PR #833, opened 2019-08-20
SSO permissions

fisx:
galley has an end-point only used by spar for deciding team ownership.  so far this end-point has not acknowledged owners if they didn't have an email address.

this constraint is relaxed here, the code for membership status is cleaned up, and tests are added.

### DIFF
--- a/services/brig/src/Brig/Team/Util.hs
+++ b/services/brig/src/Brig/Team/Util.hs
@@ -11,19 +11,34 @@ import Control.Error
 import qualified Brig.IO.Intra as Intra
 import qualified Data.Set as Set
 
-data TeamOwnershipStatus = IsOnlyTeamOwner | IsOneOfManyTeamOwners | IsNotTeamOwner | NoTeamOwnersAreLeft
+-- | Every team must have at least one owner with an email address for billing and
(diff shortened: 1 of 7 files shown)


fisx:
The CI test failure was caused by a flaky test introduced in https://github.com/wireapp/wire-server/pull/830

I'll fix this elsewhere.

services/brig/src/Brig/Team/Util.hs
```diff
@@ -11,19 +11,34 @@ import Control.Error
...
+-- | Every team must have at least one owner with an email address for billing and
+-- administration.  'TeamOwnershipStatus' distinguishes all the relevant cases.
+data TeamOwnershipStatus
+  = IsOnlyTeamOwnerWithEmail
+  | IsOneOfManyTeamOwnersWithEmail
+  | IsTeamOwnerWithoutEmail
+  | IsNotTeamOwner
   deriving (Eq, Show, Bounded, Enum)
 
 -- | A team owner is a team member with full permissions *and* an email address.
```
tiago-loureiro: Can you adjust this comment then please?

fisx: oops, thanks.  i removed the comment, as it doesn't really belong here.

PR #834, opened 2019-08-20
Fix flaky integration test.

fisx:
promised [here](https://github.com/wireapp/wire-server/pull/833#issuecomment-523064347)


tiago-loureiro:
Tried locally and stuck in the `limit-retry` test, don't merge as-is

fisx:
> Looks good now, just needs to get the conflicts resolved.

done.

> Did you double check the implementation difference in the `Timeout` newtype introduced [here](https://github.com/wireapp/wire-server/pull/830) ?

what do you mean?  i just switched from `DiffTime` to `NominalDiffTime` in order to be able to do arithmetics against `UTCTime`.  the only noteworthy effect of this *should be* that we lose or win a leap second every now and then.  are you saying this has something to do with my stupid mistake earlier in this PR?

PR #836, opened 2019-08-23
Json logging

arianvp:


arianvp:
Self-review:

 Use of this type is discouraged. Note the following equivalence:

 > Data.Monoid.Last x === Maybe (Data.Semigroup.Last x)

 In addition to being equivalent in the structural sense, the two
 also have 'Monoid' instances that behave the same. This type will
 be marked deprecated in GHC 8.8, and removed in GHC 8.10.
 Users are advised to use the variant from "Data.Semigroup" and wrap
 it in 'Maybe'.

libs/imports/src/Orphans.hs
```diff
@@ -6,10 +6,3 @@
 -- | Orphan instances for non-Wire-specific types and classes.
 module Orphans () where
```
fisx: please decomission the orphanage.  it's abandoned.

arianvp:
@fisx  Example logs:

```json
[galley] {"request":"N/A","targets":"2a18c226-eb2f-4037-8d74-dac55a04417c","msgs":["D","LegalHold.approveDevice"]}
[galley] {"request":"N/A","targets":"[\"2a18c226-eb2f-4037-8d74-dac55a04417c\",\"65a325d3-5ec9-4669-97be-7605b76894db\"]","msgs":["D","LegalHold.removeSettings"]}
[galley] {"code":"403","label":"no-team-member","request":"N/A","msgs":["D","\"Requesting user is not a team member.\""]}
[galley] {"request":"N/A","targets":"[\"2a18c226-eb2f-4037-8d74-dac55a04417c\",\"65a325d3-5ec9-4669-97be-7605b76894db\"]","msgs":["D","LegalHold.removeSettings"]}
[galley] {"code":"403","label":"operation-denied","request":"N/A","msgs":["D","\"Insufficient permissions (missing ChangeLegalHoldTeamSettings)\""]}
[galley] {"request":"N/A","targets":"[\"2a18c226-eb2f-4037-8d74-dac55a04417c\",\"65a325d3-5ec9-4669-97be-7605b76894db\"]","msgs":["D","LegalHold.removeSettings"]}
[galley] {"code":"403","label":"access-denied","request":"N/A","msgs":["D","\"This operation requires reauthentication\""]}
[galley] {"request":"N/A","targets":"[\"2a18c226-eb2f-4037-8d74-dac55a04417c\",\"65a325d3-5ec9-4669-97be-7605b76894db\"]","msgs":["D","LegalHold.removeSettings"]}
[galley] {"request":"N/A","targets":"[\"2a18c226-eb2f-4037-8d74-dac55a04417c\",\"65a325d3-5ec9-4669-97be-7605b76894db\"]","msgs":["D","LegalHold.removeSettings'"]}
[galley] {"request":"N/A","targets":"[\"2a18c226-eb2f-4037-8d74-dac55a04417c\",\"65a325d3-5ec9-4669-97be-7605b76894db\"]","msgs":["D","LegalHold.removeSettings"]}
[galley] {"request":"N/A","targets":"[\"2a18c226-eb2f-4037-8d74-dac55a04417c\",\"65a325d3-5ec9-4669-97be-7605b76894db\"]","msgs":["D","LegalHold.removeSettings'"]}
[galley] {"code":"404","label":"no-team","request":"N/A","msgs":["D","\"team not found\""]}
[galley] {"code":"404","label":"no-team","request":"N/A","msgs":["D","\"team not found\""]}
[galley] {"request":"N/A","targets":"e5fc30a4-a24b-424b-a827-4ba1ee566882","msgs":["D","Teams.addTeamMemberInternal"]}
[galley] {"request":"N/A","targets":"[\"992eb7f9-2c28-4172-97b0-b44b9e5a4119\",\"e5fc30a4-a24b-424b-a827-4ba1ee566882\"]","msgs":["D","LegalHold.createSettings"]}
[galley] {"request":"N/A","targets":"e5fc30a4-a24b-424b-a827-4ba1ee566882","msgs":["D","LegalHold.requestDevice"]}
```

tiago-loureiro:
`[galley] {"request":"N/A","targets":"[\"2a18c226-eb2f-4037-8d74-dac55a04417c\",\"65a325d3-5ec9-4669-97be-7605b76894db\"]","msgs":["D","LegalHold.removeSettings"]}`

The `\"` is due to the `show . toByteString` mentioned [here](https://github.com/wireapp/wire-server/pull/828#discussion_r318054713)

Also, I see that `"D"` shows up as "msgs", that should be a diff. field named `"level"`

arianvp:
> Also, I see that "D" shows up as "msgs", that should be a diff. field named "level"

This I cannot change I think. `tinylog` internally sets the loglevel not as a `field` but just appends it to the `Msg`  I think.  I'm not sure how to not make it do that?  I'm actually _ignoring_ the `logLevel` argument in my `jsonRenderer` function https://github.com/wireapp/wire-server/pull/836/files#diff-79a662d6500cd50cb5717cf50673a5c8R62

fisx:
> `[galley] {"request":"N/A","targets":"[\"2a18c226-eb2f-4037-8d74-dac55a04417c\",\"65a325d3-5ec9-4669-97be-7605b76894db\"]","msgs":["D","LegalHold.removeSettings"]}`

the `targets` value could be an array instead of a string.  (not sure that makes sense; only do it if you like the idea, and it's easy to change.)

> [debug level should have its own field]

there are ways to parse this out of the messages list again, but i think it's not worth the effort and the risk of complications with tinylog updates.

libs/extended/src/System/Logger/Extended.hs
```diff
@@ -1,32 +1,90 @@
...
+import qualified Data.ByteString.Lazy.Char8 as L
 import qualified System.Logger.Class as LC
 
--- TODO(arianvp): Get rid of boolean blindness
--- TODO(arianvp): Add JSON log format. This will make our lives a lot easier
---    This will add a dependency on aeson for this package,
---    but it already transitively depended on it through imports.
---    Interestingly, the only place where imports uses Aeson
---    is in the Orphans module which defines Aeson Orphans instances for Log.Level.
---    So while we're at it, we should probably move those orphans here.
```
fisx: this means that now we can remove aeson from `/libs/imports/package.yaml`?

libs/extended/src/System/Logger/Extended.hs
```diff
@@ -1,32 +1,90 @@
...
-mkLogger :: Log.Level -> Bool -> IO Log.Logger
-mkLogger lvl netstr = Log.new
+
+deriving instance Generic LC.Level
+instance FromJSON LC.Level
+instance ToJSON LC.Level
```
fisx: i would move these to `System.Logger.Extended.Orphans`, or `System.Logger.Orphans`.  (This way we won't accidentally add more orphans because there is no warning about it.)  [ignore if you disagree]

libs/extended/src/System/Logger/Extended.hs
```diff
@@ -1,32 +1,90 @@
...
+
+-- | The log formats supported
+data LogFormat = JSON | Plain | Netstring 
+  deriving stock (Eq, Show, Generic)
+  deriving anyclass (ToJSON, FromJSON) -- TODO write this instance manually?
```
fisx: Yes, please.

```haskell
instance ToJSON LogFormat where
  toJSON = toJSON . show

instance FromJSON LogFormat where
  parseJSON (String "JSON") = pure JSON
  parseJSON (String "Plain") = pure Plain
  parseJSON (String "Netstring") = pure Netstring
  parseJSON _bad = fail "LogFormat"
```

Don't forget to add a quickcheck roundtrip test.  we collect most of them in `/libs/brig-types/test/.../Common.hs` for historical reasons, but you could also start a test module here if you want to be thorough.  :-)

libs/extended/src/System/Logger/Extended.hs
```diff
@@ -1,32 +1,90 @@
...
+
+-- | We use this as an intermediate structure to ease the implementation of the
+-- ToJSON instance but we could just inline everything. I think this has
+-- negligible impact and makes the code a bit more readable. Let me know
+data Element' = Element' [(Builder, Builder)] [Builder]
+
+instance ToJSON Element' where
+  toJSON = error "Use toEncoding"
```
fisx: https://hoogle.zinfra.io/file/root/.stack/snapshots/x86_64-linux/custom-wire-1.1-G6NVtOVxqIB2/8.4.3/doc/aeson-1.3.1.1/Data-Aeson.html#t:ToJSON

> Instances in general must specify toJSON and should (but don't need to) specify toEncoding.

What is the reason for implementing `toEncoding`, and not `doJSON`?

libs/extended/src/System/Logger/Extended.hs
```diff
@@ -1,32 +1,90 @@
...
+      msgsToSeries  = pair "msgs" . list (text . cs . eval)
+
+collect :: [Element] -> Element'
+collect = foldr go (Element' [] []) 
+  where
+    go :: Element -> Element' -> Element'
+    go (Bytes b) (Element' f m) =
+      Element' f (b : m)
+    go (Field k v) (Element' f m) = 
+      Element' ((k,v) : f) m
```
fisx:
```suggestion
      Element' ((k, v) : f) m
```

libs/extended/src/System/Logger/Extended.hs
```diff
@@ -1,32 +1,90 @@
...
+netStringsToLogFormat False = Plain
+
+-- | Creates a logger given a log format Also takes an useNetstrings argument
+-- which is there because we cannot immediatelly deprecate the old interface.
+-- Old configs only provide the useNetstrings argument and not the logFormat
+-- argument, and in that case implement the old behaviour of either enabling
+-- plain text logging or netstring logging.  If both arguments are set,
+-- logFormat takes presedence over useNetstrings
+--
+-- TODO: Once we get rid of the useNetstrings in our config files, we can
```
fisx:
```suggestion
-- FUTUREWORK: Once we get rid of the useNetstrings in our config files, we can
```

libs/extended/src/System/Logger/Extended.hs
```diff
@@ -1,32 +1,90 @@
...
+-- change the type of mkLogger to  mkLogger :: Log.Level -> LogFormat -> IO
+-- Log.Logger
+mkLogger :: Log.Level -> Maybe (Last Bool) -> Maybe (Last LogFormat) -> IO  Log.Logger
+mkLogger lvl useNetstrings logFormat = do
+  mkLogger'' lvl $
+    case (fmap netStringsToLogFormat <$> useNetstrings) <> logFormat of
+      Just x -> getLast x
+      Nothing -> Plain
+    
+mkLogger'' :: Log.Level -> LogFormat -> IO Log.Logger
```
fisx: Call this `mkLoggerNew`, add a comment that it's a version of `mkLogger` that does not support the deprecated `useNetstrings` flag; and mention it in the haddocs of `mkLogger` instead of repeading the simplified type signature.

libs/extended/src/System/Logger/Extended.hs
```diff
@@ -40,15 +98,16 @@ mkLogger lvl netstr = Log.new
 --
 --   * use 'canonicalizeWhitespace'.
 --
-simpleSettings :: Log.Level -> Bool -> Log.Settings
+simpleSettings :: Log.Level -> LogFormat -> Log.Settings
 simpleSettings lvl netstr
```
fisx:
```suggestion
simpleSettings lvl logformat
```

libs/extended/src/System/Logger/Extended.hs
```diff
@@ -40,15 +98,16 @@ mkLogger lvl netstr = Log.new
...
   = Log.setLogLevel lvl
   . Log.setRenderer (canonicalizeWhitespace rndr)
   $ Log.defSettings
   where
     rndr = case netstr of
```
fisx:
```suggestion
    rndr = case logformat of
```

services/galley/galley.integration.yaml
```diff
@@ -27,8 +27,8 @@ settings:
   intraListing: false
   conversationCodeURI: https://app.wire.com/join/
 
-logLevel: Info
-logNetStrings: false
+logLevel: Debug
+logFormat: JSON
```
fisx: i see that you wanted to test this, but it's a bit odd to do have gally log in json, and everybody else log the old way.  Could you revert this change?

libs/extended/src/System/Logger/Extended.hs
fisx: (i may have broken indentation here; please double-check.)

libs/extended/src/System/Logger/Extended.hs
arianvp: I decided to use the `Generic` deriving because we did the same for `LogLevel` already and wanted to keep the code similar to the code I was already adding :P

libs/extended/src/System/Logger/Extended.hs
arianvp: Yeh I could, but this `Element'` datatype is private to this module, and we only call `toEncoding` on it and not `toJSON`  . My original plan was to get rid of the intermediate `Element'` datatype completely so that no intermediate structure is allocated at all and it would be streaming bytestrings all the way down.

I'll just get rid of the `ToJSON` instance and rename `toEncoding` to `toElementEncoding`

libs/extended/src/System/Logger/Extended.hs
fisx: That raises the question why you added the `TODO`...  :-)

What you say makes sense.  I'm now in favor of removing the `TODO` and leaving the generic instances in.

arianvp:
> the targets value could be an array instead of a string. (not sure that makes sense; only do it if you like the idea, and it's easy to change.)

it's not easy to change.  `type Field = (Builder, Builder)`  Tinylog makes too many assumptions about the logformat for us to do something more sensible here

arianvp:
>Also, I see that "D" shows up as "msgs", that should be a diff. field named "level"

@tiago-loureiro  do you want me to add an extra field called `level` eventhough I can not get rid of the `D` in the log messages? (I really still don't understand why this `D` is showing up there in the first place. I'm very confused about it)

fisx:
> > the targets value could be an array instead of a string. (not sure that makes sense; only do it if you like the idea, and it's easy to change.)
> 
> it's not easy to change. `type Field = (Builder, Builder)` Tinylog makes too many assumptions about the logformat for us to do something more sensible here

you could parse it back from the string, i guess...  :-)  (could you?)

fisx:
> > Also, I see that "D" shows up as "msgs", that should be a diff. field named "level"
> 
> @tiago-loureiro do you want me to add an extra field called `level` eventhough I can not get rid of the `D` in the log messages? (I really still don't understand why this `D` is showing up there in the first place. I'm very confused about it)

it's the call to `lmsg` in [this line](https://gitlab.com/twittner/tinylog/blob/7e980a4fc2c3cf419e3361e4fd8c3123202f0310/src/System/Logger.hs#L207).

I say we leave it in and add an extra field for debug level as tiago suggested, just in case it'll make analytics easier.

libs/extended/src/System/Logger/Extended.hs
```diff
@@ -37,29 +37,25 @@ data LogFormat = JSON | Plain | Netstring
 -- | We use this as an intermediate structure to ease the implementation of the
 -- ToJSON instance but we could just inline everything. I think this has
 -- negligible impact and makes the code a bit more readable. Let me know
-data Element' = Element' [(Builder, Builder)] [Builder]
+data Element' = Element' Series [Builder]
```
fisx: just curious: did you check that it's actually faster?  i always expect these things to be fused away by ghc, but i guess i'm expecting too much...

arianvp:
> you could parse it back from the string, i guess... :-) (could you?)
There is an `unsafeToEncoding` function that turns any bytestring into an `Encoding`.   We could use that I guess but we have to be sure that what we put in is valid JSON otherwise bad things happen =)

libs/extended/src/System/Logger/Extended.hs
arianvp: no did not check.

PR #838, opened 2019-08-24
Remove Debug.Trace from Imports.

fisx:
Debug.Trace should not be used in production code, but it's easy to miss a trace statement in a PR.  It's much easier to see that `Debug.Trace` is still imported somewhere.  Also, not providing it implicitly makes it clearer that it shouldn't be used in production code.


fisx:
ready for review!

PR #839, opened 2019-08-24
Cleanup integration tests

fisx:
Do not touch working copy in integration tests.  Use temp files instead of files that are under version control.


fisx:
(once this is merged, we should change config files all over our integration test infrastructure.)

fisx:
nope, this is not at all working yet.  the turn config files are also needed by brig, so we can't generate them under `/tmp/<some-hash>/....` in the integration tests; that's too late.

we could have them in the place where they are now, but make sure the integration tests keep a copy and write it over the changed file once the test is over (whether passed or failed).

this is probably related to the question of dynamic config reloading, but i would really like to find a quick and simple solution.  it's been bothering me every day as long as i've worked here...  :)

services/brig/test/integration/Main.hs
```diff
@@ -48,6 +50,20 @@ data Config = Config
 
 instance FromJSON Config
 
+lookupTempDir :: IO FilePath
+lookupTempDir = fromMaybe "/tmp" . getFirst . mconcat . map First
```
fisx: https://github.com/wireapp/wire-server/pull/836#issuecomment-524376542

fisx:
ready for review!

services/brig/test/integration/API/TURN.hs
```diff
@@ -165,14 +163,29 @@ toTurnURI s h p t = turnURI s ip p t
...
-    tmpDir <- getTemporaryDirectory
-    tmpPathFile <- writeTempFile tmpDir "-turn.tmp" newConf
-    copyFile tmpPathFile cfgDest
+
+type TurnUpdater = String -> IO ()
+
+withTurnFile :: FilePath -> (TurnUpdater -> Http ()) -> Http ()
+withTurnFile cfgDest action = do
+    tmpdirRoot <- lookupTempDir
+    Temp.withTempDirectory tmpdirRoot "wire.temp" $ \tempdir -> do
```
tiago-loureiro: I am unable to compile things right now and check this :) 
Question: does it make sense to `tmpdirRoot <- lookupTempDir` -> wouldn't it be easier to simply use `withSystemTempDirectory`?

fisx: yes, i didn't think of that.  will change, thanks!

fisx:
ready for review.

Issue #821, opened 2019-08-08
Makefile glitch


fisx:
if you do

```
make -C services/galley/ integration
```

it will write to `services/galley/dist`, but it should write to `dist`.  Somewhere we probably need to make the path relative to the repo root, not to the working directory.

mheinzel:
Is this fixed by https://github.com/wireapp/wire-server/pull/904?

fisx:
> Is this fixed by #904?

yes, thanks!

PR #840, opened 2019-08-27
[WIP] Cleanup Makefiles

fisx:
- install executables into project-root/dist right away
- do not copy it to project root only *sometimes* (fixes https://github.com/wireapp/wire-server/issues/821)
- extract common stack arguments into makefile-local shell variable
- pass $WIRE_STACK_OPTIONS everywhere where applicable
- keep $WIRE_STACK_OPTIONS intact in haddock goals
- `compile` goal in brig, spar: drop the unexpected `--fast` (everybody else doesn't have it).
- make sure calls to stack in sub-Makefiles do not build the entire project (add the `.` argument to build).

after merging this and if you like tidy working copies, you can run:

```shell
cd `stack path --project-root`
mv .stack-work .stack-work-backup
find . -name dist -exec rm -rf {} \;
find . -name .stack-work -exec rm -rf {} \;
mv .stack-work-backup .stack-work
```

after that, you shouldn't see any of those folders in the sub-directories any more.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,8 @@
 SHELL            := /usr/bin/env bash
+# STACK_OPTIONS are the options used for most stack calls in this
+# Makefile; WIRE_STACK_OPTIONS is the place for the user to override
+# things.
+STACK_OPTIONS := --pedantic --test --bench --no-run-benchmarks --local-bin-path=$(shell stack path --project-root)/dist
 LANG             := en_US.UTF-8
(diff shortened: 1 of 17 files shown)


fisx:
lots of conceptual mistakes.  i guess i need to understand better what i'm doing there...

fisx:
i will never get around to doing this, so i will close it now.

PR #847, opened 2019-09-06
Cleanup: incoherent functions for response body parsing.

fisx:
The first commit gives a good idea of what I'm doing here.

### DIFF
--- a/libs/api-client/src/Network/Wire/Client/API/Auth.hs
+++ b/libs/api-client/src/Network/Wire/Client/API/Auth.hs
@@ -15,6 +15,7 @@ import Bilge
 import Brig.Types.User.Auth as Auth hiding (Cookie, user)
 import Control.Monad.Catch (MonadMask)
 import Data.List.NonEmpty
+import Data.Text (pack)
 import Data.Time (getCurrentTime)
 import Network.HTTP.Client (generateCookie)
 import Network.HTTP.Types.Method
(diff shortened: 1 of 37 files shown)


fisx:
Ready for review!

fisx:
(The CI error should be fixed with this: https://github.com/wireapp/wire-server/pull/848)

fisx:
> In the usage, it seems you replaced all existing instances of decodeBody with either `responseJsonMaybe` or `responseJsonThrow ErrorCall`.  [...]

`Maybe` is an instance of `MonadThrow`, so you can do this:

```haskell
decodeBody :: MonadThrow m => m ()
decodeBody = undefined

alsoDecodeBody :: Maybe ()
alsoDecodeBody = decodeBody
```

This can be very confusing if you expect `throw` to actually do something exception-y, but instead it'll just discard the exception value and return `Nothing`.

Before this PR, there was just one `decodeBody` function in this region of the test suite, and it was used polymorphically for both throwing exceptions and returning `Nothing`.  I also found this confusing, and when I realized it I decided I'd better specialize the `Nothing` part.

> Can you add a helper for `responseJsonThrow ErrorCall` or adapt the responseJsonThrow one to take `ErrorCall` as an argument by default (to allow for less typing when using this, especially in tests)

will do, thanks for insisting.  i think I thought that would be a good idea, but then i was too lazy.  :)

jschaul:
> `Maybe` is an instance of `MonadThrow`, so you can do this:
> 
> ```haskell
> decodeBody :: MonadThrow m => m ()
> decodeBody = undefined
> 
> alsoDecodeBody :: Maybe ()
> alsoDecodeBody = decodeBody
> ```
> 
> This can be very confusing if you expect `throw` to actually do something exception-y, but instead it'll just discard the exception value and return `Nothing`.
> 
> Before this PR, there was just one `decodeBody` function in this region of the test suite, and it was used polymorphically for both throwing exceptions and returning `Nothing`. I also found this confusing, and when I realized it I decided I'd better specialize the `Nothing` part.

:bulb: Aha! Thanks for the explanation, I had been wondering why `decodeBody` sometimes gave a Maybe and sometimes threw an exception. I think the refactored Code is much clearer now. Thanks!

PR #850, opened 2019-09-11
Change feature flags from boolean to custom enum types.

fisx:
Different feature flags have subtly different semantics, and `Bool` is not a good type to represent everything we want to say.  Here is a better way.

Related: 
https://github.com/wireapp/wire-server/pull/845#issuecomment-529586511
https://github.com/wearezeta/backend-issues/issues/877#issuecomment-529573103

### DIFF
--- a/services/galley/galley.integration.yaml
+++ b/services/galley/galley.integration.yaml
@@ -37,13 +37,11 @@ settings:
     # able to register new idps.
     # disabled for integration tests (the ones who need it on will
     # turn it on themselves).
-    sso: false
+    sso: disabled-by-default
 
     # Legal Hold: this decides whether customer support / backoffice
-    # is allowed to turn the feature on for individual teams.  the
-    # default for new teams is always "false", no matter what the
-    # feature flag is set to.
-    legalhold: true
+    # is allowed to turn the feature on for individual teams.
+    legalhold: disabled-by-default
 
 logLevel: Info
 logNetStrings: false
--- a/services/galley/src/Galley/API/Teams.hs
+++ b/services/galley/src/Galley/API/Teams.hs
@@ -533,10 +533,10 @@ getLegalholdStatus (uid ::: tid ::: ct) = do
 getSSOStatusInternal :: TeamId ::: JSON -> Galley Response
 getSSOStatusInternal (tid ::: _) = do
     defConfig <- do
-        featureSSO <- view (options . optSettings . featureEnabled FeatureSSO)
-        pure $ if featureSSO
-            then SSOTeamConfig SSOEnabled
-            else SSOTeamConfig SSODisabled
+        featureSSO <- view (options . optSettings . setFeatureFlags . flagSSO)
+        pure . SSOTeamConfig $ case featureSSO of
+            FeatureSSOEnabledByDefault  -> SSOEnabled
+            FeatureSSODisabledByDefault -> SSODisabled
     ssoTeamConfig <- SSOData.getSSOTeamConfig tid
     pure . json . fromMaybe defConfig $ ssoTeamConfig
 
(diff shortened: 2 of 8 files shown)


tiago-loureiro:
Can you make a PR on https://github.com/wireapp/wire-server-deploy/ to get the CI tests going?

services/galley/galley.integration.yaml
```diff
@@ -37,13 +37,11 @@ settings:
     # able to register new idps.
     # disabled for integration tests (the ones who need it on will
     # turn it on themselves).
-    sso: false
+    sso: disabled-by-default
```
jschaul: Same comments apply here than to https://github.com/wireapp/wire-server-deploy/pull/144:

* specify in comments the allowed options
* update the comment above.

services/galley/src/Galley/API/Teams.hs
```diff
@@ -553,21 +553,25 @@ setSSOStatusInternal (tid ::: req ::: _) = do
...
-    legalholdEnabled <- view (options . optSettings . featureEnabled FeatureLegalHold)
-    if legalholdEnabled
-      then do
+    featureLegalHold <- view (options . optSettings . setFeatureFlags . flagLegalHold)
+    case featureLegalHold of
+      FeatureLegalHoldDisabledByDefault -> do
         legalHoldTeamConfig <- LegalHoldData.getLegalHoldTeamConfig tid
         pure . json . fromMaybe disabledConfig $ legalHoldTeamConfig
-      else do
+      FeatureLegalHoldDisabledPermantently -> do
```
jschaul: typo: `FeatureLegalHoldDisabledPermantently` ->  `FeatureLegalHoldDisabledPermanently`

fisx:
> Please document the semantics of what happens when, for an existing state in a given database, an operator changes a featur flag from:
> 
> legalhold: disabledPermanently to disabledByDefault (I guess now the feature can be used)
> legalhold: disabledByDefault to disabledpermanently (do existing users/devices get deleted from the database? Does it only disable the functionality for new users?)
> 
> sso: disabledByDefault -> enabledByDefault: what happens?
> sso: enabledByDefault -> disabledByDefault: what happens to previous teams?

For legalhold, this is something we discussed in https://github.com/wireapp/wire-server/pull/761, but decided we do that in a different PR.  It belongs no more here than there.  Short answer: don't do it, it's complicated.

For SSO, the interesting case is enabled -> disabled by default.  This will result in all IdPs that are already registered to continue to work, and all new IdP registrations on teams that are still on the default setting to fail.  Where would you like me to document this?

tools/stern/src/Main.hs
```diff
@@ -8,4 +8,4 @@ main :: IO ()
 main = getOptions desc Nothing defaultPath >>= start
   where
     desc = "Stern - Backoffice Service"
-    defaultPath = "/etc/wire/stern/conf/stern.yaml"
+    defaultPath = "./stern.example.yaml"
```
jschaul: please undo this change, as it breaks the defaults on kubernetes-deployed environments, and it's inconsistent with regards to the other services.

fisx: oops  :-|=|  that wasn't supposed to go in.  sorry!

fisx:
I will get back to [this](https://github.com/wireapp/wire-server/blob/515a2c83a0bb431b4565c1c7b0142cacf07f4bc6/services/galley/galley.integration.yaml#L29) tomorrow.  I vaguely remember that we had an ad-hoc referencing scheme there, but I'll have to find out again how that works.  (And perhaps document it?  :))

docs/reference/team/config-options.md
```diff
@@ -0,0 +1,50 @@
+# Legal hold
+
+Fragment.
+
+This page is about the yaml files that determine the configuration of
+the Wire backend services.
+
+
+## Feature flags
```
jschaul: Thanks for the docs!

PR #852, opened 2019-09-11
Fix swagger.

fisx:
Fixes https://github.com/wearezeta/backend-issues/issues/885


fisx:
ah:

```
"2019-09-12T06:16:31.56923 |srv=galley|galley: user error (AesonException \"Error in $.settings.featureFlags.sso: expected Bool, encountered String\")", 
```

yes, shouldn't have merged this before #850...

PR #853, opened 2019-09-11
Fix metrics.

fisx:
@tiago writes:

> Start cannon and curl localhost:8083/i/metrics
> 
> Notice there is no net_websocket_clients on the output?
> 
> Now try curl localhost:8083/i/monitoring and again curl localhost:8083/i/metrics.
> 
> Look at what happens! :) Anyway... just wanted to share this... so, culprit
> 
> https://github.com/wireapp/wire-server/blob/develop/services/cannon/src/Cannon/API.hs#L72-L77
> https://github.com/wireapp/wire-server/blob/develop/services/galley/src/Galley/API.hs#L1007-L1010
> 
> Can we please move those to a different place? Calling /i/monitoring should only render metrics and not do some extra logic.

### DIFF
--- a/services/cannon/src/Cannon/API.hs
+++ b/services/cannon/src/Cannon/API.hs
@@ -70,11 +70,15 @@ sitemap = do
     head "/i/status" (continue (const $ return empty)) true
 
 monitoring :: Media "application" "json" -> Cannon Response
-monitoring = const $ do
+monitoring _ = do
+    refreshMetrics
+    json <$> (Metrics.render =<< monitor)
+
+refreshMetrics :: Cannon ()
+refreshMetrics = do
     m <- monitor
     s <- D.size =<< clients
     gaugeSet (fromIntegral s) (path "net.websocket.clients") m
-    json <$> Metrics.render m
 
 docs :: Media "application" "json" ::: Text -> Cannon Response
 docs (_ ::: url) = do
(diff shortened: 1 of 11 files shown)


services/cannon/src/Cannon/API.hs
```diff
@@ -70,11 +69,8 @@ sitemap = do
     head "/i/status" (continue (const $ return empty)) true
 
 monitoring :: Media "application" "json" -> Cannon Response
-monitoring = const $ do
-    m <- monitor
-    s <- D.size =<< clients
-    gaugeSet (fromIntegral s) (path "net.websocket.clients") m
-    json <$> Metrics.render m
+monitoring _ = do
```
tiago-loureiro: The `do` can be removed? :P

fisx:
I also just noticed that stern has `/i/monitoring`, but not '/i/metrics`.  Should I fix that, too?  (I'd like to do it in a separate PR, though.)

jschaul:
> I also just noticed that stern has `/i/monitoring`, but not '/i/metrics`. Should I fix that, too? (I'd like to do it in a separate PR, though.)

Yes, please :) (separate PR sounds good)

services/cannon/src/Cannon/API.hs
fisx: i was tempted to argue against that, but fine...  :)

fisx:
One more problem: `waiPrometheusMiddleware` creates its own monitoring handle, which doesn't get updated by `refreshMetrics`.  One more commit coming up!

fisx:
never mind my last commit.  we're using the same state in prometheus-client and our own metrics-core, metrics-wai.  i currently don't fully understand why, but it's anyway unrelated to this PR...

fisx:
ci failure: flaky gundeck test again, unrelated.

fisx:
again...

PR #856, opened 2019-09-13
Add debug logging for endpoints that don't have a user but affect a user

arianvp:
E.g.  this is useful to see if someone is actively trying to reset
someone's password.

### DIFF
--- a/services/brig/src/Brig/API/User.hs
+++ b/services/brig/src/Brig/API/User.hs
@@ -155,7 +155,7 @@ createUser new@NewUser{..} = do
     -- Create account
     (account, pw) <- lift $ newAccount new { newUserIdentity = ident } (Team.inInvitation . fst <$> teamInvitation) tid
     let uid = userId (accountUser account)
-    Log.debug $ field "user" (toByteString uid) . msg (val "User.createUser")
+    Log.debug $ field "user" (toByteString uid) . field "action" (Log.val "User.createUser")
     Log.info $ field "user" (toByteString uid) . msg (val "Creating user")
     activatedTeam <- lift $ do
         Data.insertAccount account Nothing pw False searchable
@@ -514,7 +514,7 @@ preverify tgt code = do
 onActivated :: ActivationEvent -> AppIO (UserId, Maybe UserIdentity, Bool)
 onActivated (AccountActivated account) = do
     let uid = userId (accountUser account)
-    Log.debug $ field "user" (toByteString uid) . msg (val "User.onActivated")
+    Log.debug $ field "user" (toByteString uid) . field  "action" (Log.val "User.onActivated")
     Log.info $ field "user" (toByteString uid) . msg (val "User activated")
     Intra.onUserEvent uid Nothing $ UserActivated account
     return (uid, userIdentity (accountUser account), True)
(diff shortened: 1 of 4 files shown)


services/brig/src/Brig/API/User.hs
```diff
@@ -155,7 +155,7 @@ createUser new@NewUser{..} = do
     -- Create account
     (account, pw) <- lift $ newAccount new { newUserIdentity = ident } (Team.inInvitation . fst <$> teamInvitation) tid
     let uid = userId (accountUser account)
-
+    Log.debug $ field "user" (toByteString uid) . msg (val "User.createUser")
     Log.info $ field "user" (toByteString uid) . msg (val "Creating user")
```
tiago-loureiro: Can we just move `User.createUser` to an `action` field?

services/brig/src/Brig/API/User.hs
```diff
@@ -514,6 +514,7 @@ preverify tgt code = do
 onActivated :: ActivationEvent -> AppIO (UserId, Maybe UserIdentity, Bool)
 onActivated (AccountActivated account) = do
     let uid = userId (accountUser account)
+    Log.debug $ field "user" (toByteString uid) . msg (val "User.onActivated")
```
tiago-loureiro: Same as above

services/brig/src/Brig/User/Auth.hs
```diff
@@ -67,6 +67,7 @@ sendLoginCode phone call force = do
     case user of
         Nothing -> throwE $ SendLoginInvalidPhone phone
         Just  u -> do
+            Log.debug $ field "user" (toByteString u) . msg (val "User.sendLoginCode")
```
tiago-loureiro: I think this is not necessary given it's already covered by `login` ?

services/brig/src/Brig/User/Auth.hs
```diff
@@ -81,11 +82,14 @@ sendLoginCode phone call force = do
 lookupLoginCode :: Phone -> AppIO (Maybe PendingLoginCode)
 lookupLoginCode phone = Data.lookupKey (userPhoneKey phone) >>= \case
     Nothing -> return Nothing
-    Just  u -> Data.lookupLoginCode u
+    Just  u -> do
+      Log.debug $ field "user" (toByteString u) . msg (val "User.lookupLoginCode")
+      Data.lookupLoginCode u
```
tiago-loureiro: Don't think this is needed, also covered on the `login` endpoint

services/brig/src/Brig/API/User.hs
arianvp: Do you want to do that for the other ones as well? From the previous PR

services/brig/src/Brig/User/Auth.hs
arianvp: No this is part of the `/login/send` endpoint instead of the `/login` endpoint if I understood correctly

services/brig/src/Brig/User/Auth.hs
arianvp: this one is only used in an internal endpoint `/i/users/login-code` so maybe we dont need it

services/brig/src/Brig/API/User.hs
tiago-loureiro: Yes, would be great 🙇

PR #861, opened 2019-09-18
[WIP] Stern: new servant swagger implementation including UI

fisx:
This PR adds under new paths a re-implementation of stern to the old code.  To try it out, pull this branch, build, run stern, and open http://localhost:8091/servant/api-docs.

Here is what it currently looks like when using http://hackage.haskell.org/package/servant-swagger-ui.  Alternatives: http://hackage.haskell.org/package/servant-swagger-ui-redoc, http://hackage.haskell.org/package/servant-swagger-ui-jensoleg.  (And obviously we could make our own from them in the future to improve/customize things further.)

2019-09-18-120850_2048x1152_scrot
2019-09-18-120905_2048x1152_scrot
2019-09-18-120950_2048x1152_scrot
2019-09-18-121037_2048x1152_scrot


fisx:
**status:** this is not ready for review, but it won't go stale easily, because most of the tricky work is in a new module.  it'll still be easy enough to merge this in 3 months.

if you want to make progress with a self-contained stern docker image before that, you can fall back to https://github.com/wireapp/wire-server/pull/757

jschaul:
> but it won't go stale easily

Hm, more than a year has passed. That qualifies as stale, I think. I also think this PR is too big and trying to achieve too much at once, ending up not achieving anything (by not being worked on anymore).

Suggestion: break this work up into smaller pieces.

PR #863, opened 2019-09-18
Sneak up on flaky test.

fisx:
The recoverAll was not needed because the test itself is already running inside `retryWhile`.

### DIFF
--- a/services/gundeck/test/integration/API.hs
+++ b/services/gundeck/test/integration/API.hs
@@ -7,7 +7,7 @@ import Bilge.Assert
 import Control.Arrow ((&&&))
 import Control.Concurrent.Async       (Async, async, wait, concurrently_, forConcurrently_)
 import Control.Lens                   ((.~), (^.), (^?), view, (<&>), _2, (%~))
-import Control.Retry                  (retrying, constantDelay, limitRetries)
+import Control.Retry                  (retrying, recoverAll, constantDelay, limitRetries)
 import Data.Aeson              hiding (json)
 import Data.Aeson.Lens
 import Data.ByteString.Conversion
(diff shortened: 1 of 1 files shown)


fisx:
I have no idea why this is happening so often now.  The test and all the related code are quite old.

services/gundeck/test/integration/API.hs
```diff
@@ -159,9 +159,9 @@ removeStalePresence = do
     w <- wsRun ca uid con (wsCloser m)
     wsAssertPresences uid 1
     liftIO $ void $ putMVar m () >> wait w
+    threadDelay 200000  -- sometimes the last line fails with @1 =/= 0@.  does this line help?
```
fisx:
```suggestion
    threadDelay 10000000  -- sometimes the last line fails with @0 =/= 1@.  does this line help?
```

fisx:
I've set the delay to 10 seconds now.  If that fixes it, I will move the push into the retry-loop instead of the delay so that it only takes as long as is needs to.

services/gundeck/test/integration/API.hs
```diff
@@ -159,9 +159,9 @@ removeStalePresence = do
     w <- wsRun ca uid con (wsCloser m)
     wsAssertPresences uid 1
     liftIO $ void $ putMVar m () >> wait w
+    threadDelay 10000000  -- sometimes the last line fails with @0 =/= 1@.  does this line help?
```
tiago-loureiro: Seems to have helped, yes

services/gundeck/test/integration/API.hs
```diff
@@ -7,7 +7,7 @@ import Bilge.Assert
 import Control.Arrow ((&&&))
 import Control.Concurrent.Async       (Async, async, wait, concurrently_, forConcurrently_)
 import Control.Lens                   ((.~), (^.), (^?), view, (<&>), _2, (%~))
-import Control.Retry                  (retrying, recoverAll, constantDelay, limitRetries)
+import Control.Retry                  (retrying, constantDelay, limitRetries)
```
fisx:
```suggestion
import Control.Retry                  (retrying, recoverAll, constantDelay, limitRetries)
```

PR #867, opened 2019-09-30
Make property size configurable

arianvp:


fisx:
> Problem now is that they're mandatory. We should have a better story for introducing new config options. This is not very ergnomic. I'm not sure 'where' in the codebase to set the default. I could make a newtype wrapper with a custom FromJSON instance but this is conflating encoding with business logic, which I do not like.

Arguably, we do this in so many places that you should just go ahead and do that, since it's easy.

I agree that we should have a better way to do this, but it's probably something to do separately from the issue at hand.

services/brig/src/Brig/API.hs
```diff
@@ -954,17 +954,16 @@ sitemap o = do
...
         throwStd propertyKeyTooLarge
     lbs <- Lazy.take (maxValueLen + 1) <$> liftIO (lazyRequestBody (fromJsonRequest req))
     unless (Lazy.length lbs <= maxValueLen) $
         throwStd propertyValueTooLarge
     val <- hoistEither $ fmapL (StdError . badRequest . pack) (eitherDecode lbs)
     API.setProperty u c k val !>> propDataError
     return empty
-  where
-    maxKeyLen   = 256
-    maxValueLen = 512
```
fisx: Can you leave these lines in and rename them do `default...`?  Instead of inlining the default values above?

services/brig/src/Brig/App.hs
```diff
@@ -34,7 +34,6 @@ module Brig.App
     , turnEnv
     , turnEnvV2
     , internalEvents
-
```
fisx: can go away.

services/galley/src/Galley/App.hs
```diff
@@ -123,6 +123,7 @@ instance MonadHttp Galley where
         httpManager <- view manager
         liftIO $ withResponse req httpManager handler
 
+
```
fisx: can go

services/brig/src/Brig/API.hs
```diff
@@ -954,17 +954,16 @@ sitemap o = do
 
 setProperty :: UserId ::: ConnId ::: PropertyKey ::: JsonRequest PropertyValue -> Handler Response
 setProperty (u ::: c ::: k ::: req) = do
-    unless (Text.compareLength (Ascii.toText (propertyKeyName k)) maxKeyLen <= EQ) $
+    maxKeyLen <- fromMaybe 256 <$> view (settings . propertyMaxKeyLen)
+    maxValueLen <- fromMaybe 512 <$> view (settings . propertyMaxValueLen)
+    unless (Text.compareLength (Ascii.toText (propertyKeyName k)) (fromIntegral maxKeyLen) <= EQ) $
```
fisx: the `fromIntegral` shouldn't be necessary.  we should just store the types expected here in `Settings`.

arianvp:
LGTM if tests passes

tiago-loureiro:
> Problem now is that they're mandatory. We should have a better story for introducing new config options. This is not very ergnomic. I'm not sure 'where' in the codebase to set the default. I could make a newtype wrapper with a custom FromJSON instance but this is conflating encoding with business logic, which I do not like.

This was/is actually not the case, right? Seems like not setting anything right now is fine.

fisx:
> This was/is actually not the case, right? Seems like not setting anything right now is fine.

Correct.  Arguably we don't want that either, but that's how it works now.  I think comment was on an earlier commit on this branch.

PR #869, opened 2019-10-01
Make gundeck handle AWS outages better.

fisx:
Fixes https://github.com/zinfra/backend-issues/issues/647

This is downwards compatible in the sense that if you don't change the config, the behavior of the gundeck before and after this PR won't be distinguishable in any way.

If we add a budget to the configuration, `push1'` (formerly `push1`) will only be called *sometimes*, which means there is a chance push notifications will be lost.  On the bright side, this is a robust way to limit the number of parallel runs of `push1`, which makes it a net win, since we won't choke on SNS outages any more.

Tasks:

- [x] write integration test for `ThreadBudget`
- [x] resolve TODO in integration test.

FUTUREWORK: there are two branches with interesting, but buggy code for testing the lower bound of running threads with quickcheck-state-machine:

- https://github.com/wireapp/wire-server/blob/fisx/gundeck-handling-sns-outages-backup/services/gundeck/test/unit/ThreadBudget.hs
- https://github.com/wireapp/wire-server/blob/fisx/gundeck-handling-sns-outages-backup2/services/gundeck/test/unit/ThreadBudget.hs

### DIFF
--- a/services/gundeck/gundeck.integration.yaml
+++ b/services/gundeck/gundeck.integration.yaml
@@ -24,6 +24,9 @@ settings:
   httpPoolSize: 1024
   notificationTTL: 24192200
   bulkPush: true
+  maxConcurrentNativePushes:
+    hard: 30
+    soft: 10
 
 logLevel: Info
 logNetStrings: false
--- a/services/gundeck/gundeck.integration.yaml
+++ b/services/gundeck/gundeck.integration.yaml
@@ -25,8 +25,8 @@ settings:
   notificationTTL: 24192200
   bulkPush: true
   maxConcurrentNativePushes:
-    hard: 30
-    soft: 10
+    hard: 30  # more than this number of threads will not be allowed
+    soft: 10  # more than this number of threads will be warned about
 
 logLevel: Info
 logNetStrings: false
--- a/services/gundeck/src/Gundeck/Options.hs
+++ b/services/gundeck/src/Gundeck/Options.hs
@@ -36,12 +36,20 @@ data Settings = Settings
     -- | Use this option to group push notifications and send them in bulk to Cannon, instead
     -- of in individual requests
     , _setBulkPush        :: !Bool
-    -- | Maximum number of concurrent threads to 'Gundeck.Push.Native.push1'.
-    , _setMaxNativePushPerMinute :: !(Maybe Int)
+    -- | Maximum number of concurrent threads calling SNS.
+    , _setMaxConcurrentNativePushes :: !(Maybe MaxConcurrentNativePushes)
+    } deriving (Show, Generic)
+
+data MaxConcurrentNativePushes = MaxConcurrentNativePushes
+    { _mcnpHard :: Int
+    , _mcnpSoft :: Int
     } deriving (Show, Generic)
 
 deriveFromJSON toOptionFieldName ''Settings
 makeLenses ''Settings
+
+deriveFromJSON toOptionFieldName ''MaxConcurrentNativePushes
+makeLenses ''MaxConcurrentNativePushes
 
 data Opts = Opts
     { _optGundeck   :: !Endpoint       -- ^ Hostname and port to bind to
--- a/services/gundeck/src/Gundeck/Options.hs
+++ b/services/gundeck/src/Gundeck/Options.hs
@@ -41,8 +41,8 @@ data Settings = Settings
     } deriving (Show, Generic)
 
 data MaxConcurrentNativePushes = MaxConcurrentNativePushes
-    { _limitHard :: Int  -- ^ more than this number of threads will not be allowed
-    , _limitSoft :: Int  -- ^ more than this number of threads will be warned about
+    { _limitHard :: !(Maybe Int)  -- ^ more than this number of threads will not be allowed
+    , _limitSoft :: !(Maybe Int)  -- ^ more than this number of threads will be warned about
     } deriving (Show, Generic)
 
 deriveFromJSON toOptionFieldName ''Settings
--- a/services/gundeck/src/Gundeck/Options.hs
+++ b/services/gundeck/src/Gundeck/Options.hs
@@ -41,8 +41,8 @@ data Settings = Settings
     } deriving (Show, Generic)
 
 data MaxConcurrentNativePushes = MaxConcurrentNativePushes
-    { _limitHard :: Int
-    , _limitSoft :: Int
+    { _limitHard :: Int  -- ^ more than this number of threads will not be allowed
+    , _limitSoft :: Int  -- ^ more than this number of threads will be warned about
     } deriving (Show, Generic)
 
 deriveFromJSON toOptionFieldName ''Settings
--- a/services/gundeck/src/Gundeck/Options.hs
+++ b/services/gundeck/src/Gundeck/Options.hs
@@ -41,8 +41,8 @@ data Settings = Settings
     } deriving (Show, Generic)
 
 data MaxConcurrentNativePushes = MaxConcurrentNativePushes
-    { _mcnpHard :: Int
-    , _mcnpSoft :: Int
+    { _limitHard :: Int
+    , _limitSoft :: Int
     } deriving (Show, Generic)
 
 deriveFromJSON toOptionFieldName ''Settings
--- a/services/gundeck/src/Gundeck/Push.hs
+++ b/services/gundeck/src/Gundeck/Push.hs
@@ -334,7 +334,7 @@ nativeTargets p pres =
 
 
 addToken :: UserId ::: ConnId ::: JsonRequest PushToken ::: JSON -> Gundeck Response
-addToken (uid ::: cid ::: req ::: _) = mpaRunWithBudget snsTimeout $ do
+addToken (uid ::: cid ::: req ::: _) = mpaRunWithBudget snsThreadBudgetReached $ do
     new <- fromJsonBody req
     (cur, old) <- foldl' (matching new) (Nothing, []) <$> Data.lookup uid Data.Quorum
     Log.info $ "user"  .= UUID.toASCIIBytes (toUUID uid)
@@ -447,8 +447,8 @@ invalidToken :: Response
 invalidToken = json (Error status400 "invalid-token" "Invalid push token")
              & setStatus status404
 
-snsTimeout :: Response
-snsTimeout = json (Error status400 "sns-timoeut" "Too many concurrent calls to SNS; is SNS down?")
+snsThreadBudgetReached :: Response
+snsThreadBudgetReached = json (Error status400 "sns-thread-budget-reached" "Too many concurrent calls to SNS; is SNS down?")
            & setStatus status413
 
 tokenTooLong :: Response
--- a/services/gundeck/src/Gundeck/Push.hs
+++ b/services/gundeck/src/Gundeck/Push.hs
@@ -150,7 +150,7 @@ pushAny' p = do
     mpaForkIO $ do
         alreadySent <- mpyPush notif tgts (p^.pushOrigin) (p^.pushOriginConnection) (p^.pushConnections)
         unless (p^.pushTransient) $
-            mpaPushNative notif p =<< nativeTargets p alreadySent
+            mpaPushNative notif p =<< nativeTargets p (nativeTargetsRecipients p) alreadySent
   where
     mkTarget :: Recipient -> NotificationTarget
     mkTarget r =
@@ -191,7 +191,7 @@ pushAll pushes = do
         forM_ resp $ \((notif, psh), alreadySent) ->
             unless (psh ^. pushTransient)
                 $ mpaRunWithBudget ()
-                    $ mpaPushNative notif psh =<< nativeTargets psh alreadySent
+                    $ mpaPushNative notif psh =<< nativeTargets psh (nativeTargetsRecipients psh) alreadySent
 
 
 -- REFACTOR: @[Presence]@ here should be @newtype WebSockedDelivered = WebSockedDelivered [Presence]@
@@ -266,20 +266,27 @@ pushNative notif p rcps = do
     let prio = p^.pushNativePriority
     Native.push (Native.NativePush (ntfId notif) prio Nothing) rcps
 
--- | TODO: 'nativeTargets' calls cassandra once for each 'Recipient' of the 'Push'.  Instead,
--- it should be called once with @[Push]@ for every call to 'pushAll', and that call should
--- only call cassandra once in total, yielding all addresses of all recipients of all pushes.
-nativeTargets :: forall m. (MonadNativeTargets m, MonadMapAsync m) => Push -> [Presence] -> m [Address]
-nativeTargets psh alreadySent =
-    mntgtMapAsync addresses rcps' >>= fmap concat . mapM check
+-- | Compute list of 'Recipient's from a 'Push' that may be interested in a native push.  More
+-- filtering in 'nativeTargets'.
+nativeTargetsRecipients :: Push -> [Recipient]
+nativeTargetsRecipients psh = filter routeNative (toList (fromRange (psh ^. pushRecipients)))
   where
-    rcps' :: [Recipient]
-    rcps' = filter routeNative (toList (fromRange (psh ^. pushRecipients)))
-
-    -- Interested in native pushes?
     routeNative u = u ^. recipientRoute /= RouteDirect
                  && (u ^. recipientId /= psh ^. pushOrigin || psh ^. pushNativeIncludeOrigin)
 
+-- | TODO: 'nativeTargets' calls cassandra once for each 'Recipient' of the 'Push'.  Instead,
+-- it should be called once with @[Push]@ for every call to 'pushAll', and that call should
+-- only call cassandra once in total, yielding all addresses of all recipients of all pushes.
+--
+-- FUTUREWORK: we may want to turn 'mntgtMapAsync' into an ordinary `mapM`: it's cassandra
+-- access, so it'll be fast either way given the size of the input, and synchronous calls
+-- impose a much lower risk of choking when system load peaks.
+nativeTargets
+  :: forall m. (MonadNativeTargets m, MonadMapAsync m)
+  => Push -> [Recipient] -> [Presence] -> m [Address]
+nativeTargets psh rcps' alreadySent =
+    mntgtMapAsync addresses rcps' >>= fmap concat . mapM check
+  where
     addresses :: Recipient -> m [Address]
     addresses u = do
         addrs <- mntgtLookupAddresses (u ^. recipientId)
--- a/services/gundeck/src/Gundeck/Push.hs
+++ b/services/gundeck/src/Gundeck/Push.hs
@@ -324,8 +324,16 @@ nativeTargets p pres =
     check (Left  e) = mntgtLogErr e >> return []
     check (Right r) = return r
 
+
+-- | Thin wrapper around 'runWithBudget''.
+respondWithBudget :: Gundeck Response -> Gundeck Response
+respondWithBudget action = do
+  view threadBudgetState >>= \case
+    Nothing  -> action
+    Just tbs -> runWithBudget' tbs snsTimeout action
+
 addToken :: UserId ::: ConnId ::: JsonRequest PushToken ::: JSON -> Gundeck Response
-addToken (uid ::: cid ::: req ::: _) = do
+addToken (uid ::: cid ::: req ::: _) = respondWithBudget $ do
     new <- fromJsonBody req
     (cur, old) <- foldl' (matching new) (Nothing, []) <$> Data.lookup uid Data.Quorum
     Log.info $ "user"  .= UUID.toASCIIBytes (toUUID uid)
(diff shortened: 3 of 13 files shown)


services/gundeck/gundeck.integration.yaml
```diff
@@ -24,6 +24,9 @@ settings:
   httpPoolSize: 1024
   notificationTTL: 24192200
   bulkPush: true
+  nativePushBudget:  # 20 push notifications every 20 seconds.
+    budgetTimeout: 20
```
arianvp: Can we come up with a more scientific number? Can you add a comment that is a bit more descriptive? Is this the push budget per user? Or for the entire server? For the entire sever it seems a bit low no?

services/gundeck/src/Gundeck/Push/Native.hs
```diff
@@ -30,13 +31,30 @@ import qualified Gundeck.Notification.Data as Stream
 import qualified Gundeck.Push.Data         as Data
 import qualified System.Logger.Class       as Log
 
-push :: NativePush -> [Address] -> Gundeck [Result]
+push :: NativePush -> [Address] -> Gundeck [()]
 push _    [] = return []
 push m   [a] = pure <$> push1 m a
 push m addrs = mapConcurrently (push1 m) addrs
```
arianvp: I don't think you're actually solving the root cause here. We're still starting all the requests concurrently.  I think the budgeting should happen _outside_ of the `mapConcurrently`. not _inside_

libs/types-common/src/Control/Budget.hs
```diff
@@ -1,6 +1,6 @@
 {-# LANGUAGE GeneralizedNewtypeDeriving #-}
 
-module Brig.Budget
```
arianvp: We can't move this outside of brig as it makes assumptions about the brig schema

arianvp:
@fisx and I decided that we want to do rate-limit per instance, not per service.    Having multiple gundecks  look at the same budget in cassandra will not work due to the functions involving budgets not being transactional. 

We probably just want to keep some rate-limit data in-process of each gundeck process.   We looked at the https://hackage.haskell.org/package/token-bucket-0.1.0.1/docs/Control-Concurrent-TokenBucket.html library for this which looks promising.

Also,  we don't want to rate limit _all_ requests to SNS. If they can take the traffic we should just give it. We should instead only back off /  drop notifications if SNS is giving fatal errors.

We should discuss this a bit more tomorrow.

services/gundeck/src/Gundeck/Push/Native.hs
fisx: no, we want to limit the number of concurrent calls to push1.

however, you are correct in that this doesn't solve the issue:

1. Budget works across instances, but if `withBudget` is called twice concurrently, both callers with get the same budget unit to spend.  This may not be a problem when doing sms budgets per user, but it will be if we have many 1k (10k?  more?) calls per second.
2. It doesn't distinguish between the long-running calls that time out because SNS is not responding (not even with a "connection refused") and the calls that return successfully after a few milliseconds.  The former need to be limited much more aggressively than the latter.

fisx:
> Also, we don't want to rate limit _all_ requests to SNS. If they can take the traffic we should just give it. We should instead only back off / drop notifications if SNS is giving fatal errors.

true, and good point.  i think what we want to do is limite the number of concurrently active calls to `push1`, not the number of calls no matter how long they will run.  so we need to somehow release the budget token only once `push1` returns *or* times out.

> We should discuss this a bit more tomorrow.

you won't be in, but sure.  :)

fisx:
see also: https://web.archive.org/web/20120808023806/http://ksdlck.com/post/17418037348/rate-limiting-at-webscale-lazy-leaky-buckets

services/gundeck/gundeck.integration.yaml
fisx: the new approach (if you look at the overall changes now) solves this: it keeps track of the number of `push1` calls that are currently still being processed in some other thread, and makes sure that number does not exceed a certain threshold (well, at least heuristically, which should be good enough).

fisx:
note to self:

```
  thread budgets
    works with mock actions:    FAIL (0.50s)
      test/unit/ThreadBudget.hs:36:
      expected: []
       but got: [Unknown]
```

fisx:
still fails non-deterministically:

```
  thread budgets
    works with mock actions:    FAIL (1.30s)
      test/unit/ThreadBudget.hs:85:
      unexpected log data: [NoBudget,NoBudget]
```

fisx:
```
Main
  thread budgets
    unit test:           OK (1.30s)
    qc stm (sequential): 
Model Nothing

   == Init (NumberOfThreads 16) ==> InitResponse (Reference (Concrete Opaque)) [ 0 ]
FAIL (0.06s)
      *** Failed! Falsifiable (after 6 tests and 3 shrinks):
      Commands
        { unCommands =
            [ Command (Init (NumberOfThreads 16)) (fromList [ Var 0 ])
            , Command
                (Run
                   (Reference (Symbolic (Var 0)))
                   [ ( NumberOfThreads 17 , MilliSeconds 1 ) ]
                   (NumberOfThreads 0)
                   (MilliSeconds 0))
                (fromList [])
            ]
        }
      *** Exception running callback: 
      makeOperations: impossible.
        error, called at src/Test/StateMachine/Types/History.hs:87:20 in quickcheck-state-machine-0.4.2-H7amjJjlQp1EpEtdEtCbey:Test.StateMachine.Types.History
      Use --quickcheck-replay=141937 to reproduce.

1 out of 2 tests failed (1.37s)
```

giving up for today...  :-)

fisx:
```
Progress 1/2: gundeckMain
  thread budgets
    unit test:           OK (1.31s)
    qc stm (sequential): FAIL (0.02s)
      *** Failed! Exception: 'HUnitFailure (Just (SrcLoc {srcLocPackage = "main", srcLocModule = "ThreadBudget", srcLocFile = "test/unit/ThreadBudget.hs", srcLocStartLine = 249, srcLocStartCol = 3, srcLocEndLine = 250, srcLocEndCol = 115})) "wrong number of over-budget calls: (0,0,0,5)"' (after 5 tests and 2 shrinks):
      Commands
        { unCommands =
            [ Command (Init (NumberOfThreads 5)) (fromList [ Var 0 ])
            , Command
                (Run
                   (Reference (Symbolic (Var 0)))
                   []
                   (NumberOfThreads 0)
                   (MilliSeconds 0))
                (fromList [])
            ]
        }
      Use --quickcheck-replay=796259 to reproduce.
```

ok, **now** i'm **really** giving up.  maybe.  the error message is helpful now, and it's probably something stupid in the tests themseves that makes this fail.  monday!

services/gundeck/test/unit/ThreadBudget.hs
```diff
@@ -0,0 +1,414 @@
...
+
+import qualified System.Logger.Class as LC
+import qualified Test.StateMachine.Types as STM
+import qualified Test.StateMachine.Types.Rank2 as Rank2
+
+
+----------------------------------------------------------------------
+-- helpers
+
+newtype NumberOfThreads = NumberOfThreads { fromNumberOfThreads :: Int }
```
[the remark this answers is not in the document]
fisx: (no!)

fisx:
ready for review!

fisx:
See also:

- https://github.com/zinfra/cailleach/pull/259
- https://github.com/zinfra/hegemony/pull/1457

services/gundeck/src/Gundeck/ThreadBudget.hs
```diff
@@ -0,0 +1,212 @@
...
+
+-- | Store all handles for cleanup in 'watchThreadBudgetState'.
+type BudgetMap = SizedHashMap UUID (Maybe (Async ()))
+
+-- generated with @makeLenses@, but we only want one of the two lenses that generates.
+threadBudgetLimits ::
+  Lens' ThreadBudgetState MaxConcurrentNativePushes
+threadBudgetLimits f_a1A3A (ThreadBudgetState x1_a1A3B x2_a1A3C)
+  = (fmap (\ y1_a1A3D -> (ThreadBudgetState y1_a1A3D) x2_a1A3C))
+      (f_a1A3A x1_a1A3B)
```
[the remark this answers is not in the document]
fisx: i'd rather not, since this is generated code (as indicated in the comment).  is it ok if i don't?

services/gundeck/src/Gundeck/ThreadBudget.hs
```diff
@@ -0,0 +1,212 @@
...
+cancelAllThreads :: ThreadBudgetState -> IO ()
+cancelAllThreads (ThreadBudgetState _ ref) = readIORef ref
+  >>= mapM_ cancel . catMaybes . SHM.elems
+
+showDebugHandles :: BudgetMap -> String
+showDebugHandles = show . SHM.size
+
+
+mkThreadBudgetState :: HasCallStack => MaxConcurrentNativePushes -> IO ThreadBudgetState
+mkThreadBudgetState limits = if limits ^. mcnpHard < limits ^. mcnpSoft
```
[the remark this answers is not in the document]
fisx: Hum, we currently allow `0` or `-1` meaning infinite.  This raises the point though that we should check that integers are positive, so I'll do at least that.

Additionally, I could allow both limits to be optional individually, and change the semantics so that soft limit always logs, and hard limit always blocks.  I would use `null` instead fo `0`, though.  I don't think this change is worth the effort, so please let me know if you disagree, then i'll do it anyway (not *that* much trouble).

libs/types-common/src/Data/SizedHashMap.hs
```diff
@@ -0,0 +1,54 @@
+module Data.SizedHashMap
```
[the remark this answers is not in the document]
fisx: we talked about this.  since this is a very thin layer around `HashMap`, i think it's not worth it.  @arianvp do you want to break the tie?

fisx:
> Note: at the moment, only messages on native push are restricted. There are other calls to SNS, namely token registration. Albeit at much lower throughput rate, perhaps those calls should also be restricted? (fine to add a FUTUREWORK where those calls are made)

:+1: 
 
> Overall, this implementation operates at a thread level. For native pushes, this is fine, as today we open a separate thread to do native pushes. More generally, however, we can't easily make use of this implementation elsewhere, where connections to an external service pile up, and time out on the same thread that the http request comes in from.

I don't understand.  If they don't open in different threads, how can they pile up?  Do you mean the `forkIO` happens without a surrounding `withThreadBudget`?  Do you have an example in the code where what you means happens?

> Could you add a line of motivational comment why you chose to operate at the thread level, as opposed to the http connection pool level?

https://github.com/wireapp/wire-server/blob/e8a8a42fa4c7e494fb7d9aa8a91dbab8d625689a/services/gundeck/src/Gundeck/ThreadBudget.hs#L12-L14

i guess this is a little vague, but the bottom line is that we'd have to fork http-client for the foreseeable future, since snoyman doesn't appear to be interested in this problem, and I estimated that it'd be significantly harder and not significantly more beneficial to get such a fix released than it would be to wrap something around the application logic.

I suppose I will add this thought to the haddocks there.  thanks for the question.  :)

jschaul:
> > Overall, this implementation operates at a thread level. For native pushes, this is fine, as today we open a separate thread to do native pushes. More generally, however, we can't easily make use of this implementation elsewhere, where connections to an external service pile up, and time out on the same thread that the http request comes in from.
> 
> I don't understand. If they don't open in different threads, how can they pile up? Do you mean the `forkIO` happens without a surrounding `withThreadBudget`? Do you have an example in the code where what you means happens?

SNS is down. I make 1000s of requests to `POST /push/token` within a few seconds. All [requests to SNS](https://github.com/wireapp/wire-server/blob/develop/services/gundeck/src/Gundeck/Push.hs#L314-L389) to add tokens stall and eventually time out: gundeck has thousands of threads and thousands of open http connections. Same problem here. No `forkIO` anywhere. Can't easily be wrapped in the threadbudget logic.

> 
> > Could you add a line of motivational comment why you chose to operate at the thread level, as opposed to the http connection pool level?
> 
> https://github.com/wireapp/wire-server/blob/e8a8a42fa4c7e494fb7d9aa8a91dbab8d625689a/services/gundeck/src/Gundeck/ThreadBudget.hs#L12-L14
> 
> i guess this is a little vague, but the bottom line is that we'd have to fork http-client for the foreseeable future, since snoyman doesn't appear to be interested in this problem, 

Snoyman clearly writes he's not opposed to rate limiting in the library, he just says he's not going to implement it himself. There's a difference!

> and I estimated that it'd be significantly harder and not significantly more beneficial to get such a fix released than it would be to wrap something around the application logic.

I understand that making a contribution to an external library takes more time. Still, it would provide a more general solution. Documenting this "I'll do some workaround to a problematic library" would be a good first step.

services/gundeck/src/Gundeck/ThreadBudget.hs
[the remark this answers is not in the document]
jschaul: Making `hard` (or both) optional sounds good!

libs/types-common/src/Data/SizedHashMap.hs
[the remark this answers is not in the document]
fisx: i changed my mind about this and will add some unit tests.

services/gundeck/src/Gundeck/ThreadBudget.hs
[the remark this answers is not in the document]
jschaul: Why is it important to not have the second lens from `ThreadBudgetState` , i.e. for `threadBudgetRunning` ? (Please add a comment why that would be harmful)

It's no different from hand-writing your Lens though, is it?

But once it's in the source code, it's no longer "generated" automatically in any way, right? So then, nicer naming sounds like it could make the code less ugly to look at.
I.e. I don't understand how to replace `...` in the sentence `this code has to remain with ugly variables, because...`.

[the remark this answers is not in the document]
fisx: heh, i just noticed something: there must not be a lens for the limits, since we don't want to allow changing it during runtime of a thread budget state.  it should be a getter, or just a function, which means that i should probably just write it by hand.

this doesn't resolve our bike shedding issue about making generated code prettier by rewriting it, but it's evidence that we're not bike shedding too much, since you put the finger on a real issue.  thanks!  :-)

services/gundeck/src/Gundeck/Options.hs
```diff
@@ -36,11 +36,21 @@ data Settings = Settings
     -- | Use this option to group push notifications and send them in bulk to Cannon, instead
     -- of in individual requests
     , _setBulkPush        :: !Bool
+    -- | Maximum number of concurrent threads calling SNS.
+    , _setMaxConcurrentNativePushes :: !(Maybe MaxConcurrentNativePushes)
+    } deriving (Show, Generic)
+
+data MaxConcurrentNativePushes = MaxConcurrentNativePushes
+    { _mcnpHard :: Int
+    , _mcnpSoft :: Int
```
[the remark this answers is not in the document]
fisx: (it'll be `_set*Limit` instead, because the json plumbing eats the everything up to the first capital letter.)

fisx:
> SNS is down. I make 1000s of requests to POST /push/token within a few seconds. All requests to SNS to add tokens stall and eventually time out: gundeck has thousands of threads and thousands of open http connections. Same problem here. No forkIO anywhere. Can't easily be wrapped in the threadbudget logic.

Ah, sorry for being thick, got it now.  I want to do this:

```haskell
addToken :: UserId ::: ConnId ::: JsonRequest PushToken ::: JSON -> Gundeck Response
addToken (uid ::: cid ::: req ::: _) = mpaRunWithBudget $ do
    ...
```

...  but it's slightly more complicated.  I think I can write a more flexible variant of `runWithBudget` though that throw an error rather than returning a `()` in case of budget exhaustion.  Then this should fix it, no?

fisx:
> I understand that making a contribution to an external library takes more time. Still, it would provide a more general solution. Documenting this "I'll do some workaround to a problematic library" would be a good first step.

ba105205cf101da7f22b3d4bd85e2e69718fbabd ?

fisx:
> > SNS is down. I make 1000s of requests to POST /push/token within a few seconds. All requests to SNS to add tokens stall and eventually time out: gundeck has thousands of threads and thousands of open http connections. Same problem here. No forkIO anywhere. Can't easily be wrapped in the threadbudget logic.
> 
> Ah, sorry for being thick, got it now. I want to do this:
> 
> ```haskell
> addToken :: UserId ::: ConnId ::: JsonRequest PushToken ::: JSON -> Gundeck Response
> addToken (uid ::: cid ::: req ::: _) = mpaRunWithBudget $ do
>     ...
> ```
> 
> ... but it's slightly more complicated. I think I can write a more flexible variant of `runWithBudget` though that throw an error rather than returning a `()` in case of budget exhaustion. Then this should fix it, no?

how about 5a9861d6cdde6efe149c238ef890d90682e4373b ?

fisx:
CI passed here: https://bridge-ie-concourse.zinfra.io/teams/main/pipelines/wire-server-pr/jobs/wire-server-integration/builds/905

services/gundeck/src/Gundeck/Push.hs
```diff
@@ -172,8 +188,10 @@ pushAll pushes = do
         resp <- compilePushResps targets <$> mpaBulkPush (compilePushReq <$> targets)
 
         -- native push
-        forM_ resp $ \((notif, psh), alreadySent) -> unless (psh ^. pushTransient) $
-            mpaPushNative notif psh =<< nativeTargets psh alreadySent
+        forM_ resp $ \((notif, psh), alreadySent) ->
+            unless (psh ^. pushTransient)
+                $ mpaRunWithBudget ()
+                    $ mpaPushNative notif psh =<< nativeTargets psh alreadySent
```
jschaul: Blocker: 
The budget call above increments the budget count by `1` each time this line is called. However, inside Native.push, there is a `mapConcurrently` on the amount of targets. If I send a message in a group chat with 300 participants each having 2 phones, budget is increased by 1 but 600 connections to SNS (worst case, if no one of those participants has the app and a websocket connection open) are opened at the same time. 

So setting a hard limit to X can mean vastly different amounts of connections to SNS are opened, depending on conversation size. This makes it hard to heuristically set a `hard` limit as setting it to `10` (random example number) means we would disallow 10 people in a 1-on-1 to send a message in the same milisecond (which we would not want as that's totally fine resource wise), but at the same time we do not want to allow 10 people sending a message to 600 devices in the same milisecond (as 6000 connections to SNS is probably too much).

fisx: I was expecting this to be fine-grained enough, but we'll have to do experiments to find out who's right.

We could easily push down the calls to `runWithBudget`, but that would make it harder to make sure we haven't missed anything.

Alternatively, we could make `runWithBudget` take another argument that says by number the budget should be decreased, and compute that number purely from the context (`((notif, psh), alreadySent)` in this case.

Which one should we try next?

fisx: My call would be to keep it as coarse-grained as it is now and see if that's enough.  It maybe, given that there is a wide enough gap between the budget needed for normal operation and what accumulates during outages.  But the extra parameter to `runWithBudget` doesn't sound difficult to implement either.

services/gundeck/src/Gundeck/ThreadBudget.hs
```diff
@@ -0,0 +1,224 @@
...
+  >>= mapM_ cancel . catMaybes . SHM.elems
+
+showDebugHandles :: BudgetMap -> String
+showDebugHandles = show . SHM.size
+
+
+mkThreadBudgetState :: HasCallStack => MaxConcurrentNativePushes -> IO ThreadBudgetState
+mkThreadBudgetState limits = if limits ^. limitHard < limits ^. limitSoft
+  then throwM . ErrorCall $
+         "setMaxConcurrentNativePushes: hard limit < soft limit: " <> show limits
```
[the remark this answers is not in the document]
fisx: thanks, i'll drop that check.  fix coming up!

services/gundeck/src/Gundeck/Push.hs
fisx: I've implemented something.  You can be the judge of whether it's still clear enough to be useful.  I made an effort to make the individual commits easier to read than the whole change.

fisx:
can be merged by reviewer

remember this is very unlikely to cause any production issues as long as we don't turn it on.

Tasks for when this is on staging:

- [ ] load-test.
- [ ] see if there are any issues with lots of threads picking tokens within a very short time window.
- [ ] FUTUREWORK: take another look at http-client.  also at cql*, which has the same problem.
- [ ] FUTUREWORK: we should write the lower-bounds checks to make sure actions are actually executed usually.  (branches `fisx/gundeck-handling-sns-outages-backup`, `fisx/gundeck-handling-sns-outages-backup2`)

jschaul:
Ideas for futurework:

- [ ] - review metrics - at the moment, counters are exposed only once every 10 seconds, which may mean we miss big spikes at metrics level (though there should be logs shown)
- [ ] - does it make sense to have some retry logic to even out 5 seconds of little traffic, followed by a spike of 1000 requests, followed by 5 seconds of little traffic?

services/gundeck/src/Gundeck/Push.hs
```diff
@@ -172,8 +188,15 @@ pushAll pushes = do
         resp <- compilePushResps targets <$> mpaBulkPush (compilePushReq <$> targets)
 
         -- native push
-        forM_ resp $ \((notif, psh), alreadySent) -> unless (psh ^. pushTransient) $
-            mpaPushNative notif psh =<< nativeTargets psh alreadySent
+        forM_ resp $ \((notif :: Notification, psh :: Push), alreadySent :: [Presence]) -> do
```
tiago-loureiro: At this point, `alreadySent == number of devices of all users in a conversation, right?`?

fisx: No, `alreadySent` contains what went out over websocket and doesn't need to be pushed natively.  I think conversations aren't really a thing here, we just have the `Notification` and `Push` to determine who should hear from this.

PR #870, opened 2019-10-07
Update docs.

fisx:


docs/reference/provisioning/scim-via-curl.md
```diff
@@ -8,6 +8,11 @@ This page shows you how to communicate with the wire backend through
 the [SCIM API](http://www.simplecloud.info/) by example.  All examples
 are [curl](https://curl.haxx.se/) (in bash syntax).
 
+We support setting the handle and user name in wire (the thing with
```
jschaul: Hm, perhaps you can clarify a little: What do you mean with `in wire`? Via curl? There is an API endpoint? Via the graphical desktop/mobile clients? 
Instead of `the thing with @`, perhaps clarify a bit more, e.g.: A user *handle* is globally unique and usually visually shown with a `@`. Example: `@wireuser123`. A *user name* has no restrictions and is shown in the conversation flow.
I'm not sure the average reader of this page easily understand what you mean with `rich-info` and `group provisioning`. Could you perhaps define these concepts? (Would a mini-glossary make sense?)

fisx: The intended audience is people who develop against our backend, and the lack of context in this paragraph matches the rest of the document.

I can't prioritize this PR now.  But I was passing on the link and need to add this bit of information on top, and I would like to have it contained in the document next time that happens.

Does that make sense?

fisx: note to self: next time mark PR as "doesn't have to be good, just has to be no worse than the status before".  this gives relevant context to the reviewer.  (sorry @jschaul :))

PR #872, opened 2019-10-09
Store raw idp metadata with typed details in c*

fisx:
- Change`IdPMetadataInfo` to contain the raw text from the xml body.
- Introduce a new content type tag for servant that can feed the raw text to `IdPMetadataInfo` (the `SAML.XML` tag we used before comes with its own parser instances that do not allow for that).
- Store this raw text in a new table with a new pair of cassandra helper functions (the `SAML.storeIdPConfig` function that could have been used for that has its type fixed by the `saml-web-sso` library as well; I think somewhere in the library we want to be able to call the database functions, so the library needs to decide the type).

Why?  Will give us more info for trouble shooting!

### DIFF
--- a/services/spar/src/Spar/API/Types.hs
+++ b/services/spar/src/Spar/API/Types.hs
@@ -131,9 +131,16 @@ type APIAuthResp
 
 type APIIDP
      = Header "Z-User" UserId :> IdpGet
+  :<|> Header "Z-User" UserId :> IdpGetRaw
   :<|> Header "Z-User" UserId :> IdpGetAll
   :<|> Header "Z-User" UserId :> IdpCreate
   :<|> Header "Z-User" UserId :> IdpDelete
+
+-- | This is a bit of a weird special case: we record the xml we get from the team admin as
+-- verbatim as possible for trouble shooting, so we can't return it here with content type
+-- 'XML' (that would trigger an attempt to encode an XML document, but we have a 'Text').
+-- we'll just send it as 'PlainText', indicating that it's not parsed.
+type IdpGetRaw  = Capture "id" SAML.IdPId :> "raw" :> Get '[PlainText] Text
 
 type IdpGet     = Capture "id" SAML.IdPId :> Get '[JSON] IdP
 type IdpGetAll  = Get '[JSON] IdPList
--- a/services/spar/src/Spar/API/Types.hs
+++ b/services/spar/src/Spar/API/Types.hs
@@ -136,11 +136,7 @@ type APIIDP
   :<|> Header "Z-User" UserId :> IdpCreate
   :<|> Header "Z-User" UserId :> IdpDelete
 
--- | This is a bit of a weird special case: we record the xml we get from the team admin as
--- verbatim as possible for trouble shooting, so we can't return it here with content type
--- 'XML' (that would trigger an attempt to encode an XML document, but we have a 'Text').
--- we'll just send it as 'PlainText', indicating that it's not parsed.
-type IdpGetRaw  = Capture "id" SAML.IdPId :> "raw" :> Get '[PlainText] Text
+type IdpGetRaw  = Capture "id" SAML.IdPId :> "raw" :> Get '[RawXML] Text
 
 type IdpGet     = Capture "id" SAML.IdPId :> Get '[JSON] IdP
 type IdpGetAll  = Get '[JSON] IdPList
--- a/services/spar/src/Spar/API/Types.hs
+++ b/services/spar/src/Spar/API/Types.hs
@@ -136,7 +136,7 @@ type APIIDP
   :<|> Header "Z-User" UserId :> IdpCreate
   :<|> Header "Z-User" UserId :> IdpDelete
 
-type IdpGetRaw  = Capture "id" SAML.IdPId :> "raw" :> Get '[RawXML] Text
+type IdpGetRaw  = Capture "id" SAML.IdPId :> "raw" :> Get '[RawXML] RawIdPMetadata
 
 type IdpGet     = Capture "id" SAML.IdPId :> Get '[JSON] IdP
 type IdpGetAll  = Get '[JSON] IdPList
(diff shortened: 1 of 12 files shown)


fisx:
ping @arianvp

services/spar/src/Spar/Types.hs
```diff
@@ -67,24 +70,27 @@ deriveJSON deriveJSONOptions ''IdPList
...
+-- trouble shooting, but @SAML.XML@ only gives us access to the xml tree, not the raw text.
+-- 'RawXML' helps with that.
+data RawXML
+
+instance Accept RawXML where
+  contentType Proxy = "application" // "xml"
 
-instance SAML.HasXML IdPMetadataInfo where
-  parse = fmap IdPMetadataValue . SAML.parse
+instance MimeUnrender RawXML IdPMetadataInfo where
```
arianvp: cool. this seems like an elegant solution.

services/spar/test-integration/Test/Spar/APISpec.hs
```diff
@@ -712,7 +714,11 @@ specCRUDIdentityProvider = do
           metadata <- makeTestIdPMetadata
           idp <- call $ callIdpCreate (env ^. teSpar) (Just owner) metadata
           idp' <- call $ callIdpGet (env ^. teSpar) (Just owner) (idp ^. idpId)
-          liftIO $ idp `shouldBe` idp'
+          rawmeta <- runSparCass $ Data.getIdPRawMetadata (idp ^. idpId)
```
arianvp: Shouldn't you call the API here instead of the database? The database is an implementation detail considering this is the `APISpec`

arianvp: Or don't we have an API call to retrieve  `IdPMetadataValue`

services/spar/test/Arbitrary.hs
```diff
@@ -46,7 +46,7 @@ instance Arbitrary NoContent where
   arbitrary = pure NoContent
 
 instance Arbitrary IdPMetadataInfo where
-  arbitrary = IdPMetadataValue <$> arbitrary
+  arbitrary = IdPMetadataValue mempty <$> arbitrary
```
arianvp:
```suggestion
  arbitrary = do
    xml <- arbitrary
    pure $ IdPMetadataValue (encode xml) xml
```

How about this?

services/spar/test-integration/Test/Spar/APISpec.hs
fisx: nope, we don't, and that's intentional.  we need this for trouble-shooting production issues, and if we do we have access to the database directly, so the database schema is not an implementation detail here, but the interface.  :)

fisx: (there are other examples in the integration tests where we (usually me) talk to the database directly.  this is usually to make sure the internal state is inconsistent even when it is not exposed directly, and tricking the interface into giving indirect evidence of this consistency is anywhere between awkward and infeasible.

i'm happy to have a fight over this over beer.  :)

services/spar/test/Arbitrary.hs
fisx: good idea!  and it *almost* type-checked.  :)

one more commit coming up...

services/spar/test-integration/Test/Spar/APISpec.hs
arianvp: I'm convinced. Just think it will be useful to have an endpoint in the future.  CS will not be poking into the cassandra database manually when debugging issues with customers. Given that _we_ are currently the customer support for SSO it is fine, but this will be a blocker in the future when we delegate this to others.

arianvp: We fought over coffee instead.

fisx:
just thought of one more thing: when deleting an IdP, we should probably also delete the metadata.  i'll take care of that.

fisx:
ping @arianvp

services/spar/src/Spar/API/Types.hs
```diff
@@ -131,13 +131,20 @@ type APIAuthResp
...
+  :<|> Header "Z-User" UserId :> IdpGetRaw
   :<|> Header "Z-User" UserId :> IdpGetAll
   :<|> Header "Z-User" UserId :> IdpCreate
   :<|> Header "Z-User" UserId :> IdpDelete
 
+-- | This is a bit of a weird special case: we record the xml we get from the team admin as
+-- verbatim as possible for trouble shooting, so we can't return it here with content type
+-- 'XML' (that would trigger an attempt to encode an XML document, but we have a 'Text').
+-- we'll just send it as 'PlainText', indicating that it's not parsed.
+type IdpGetRaw  = Capture "id" SAML.IdPId :> "raw" :> Get '[PlainText] Text
```
arianvp: Don't we have the `RawXML` datatype for that though?

services/spar/src/Spar/API/Types.hs
```diff
@@ -136,11 +136,7 @@ type APIIDP
   :<|> Header "Z-User" UserId :> IdpCreate
   :<|> Header "Z-User" UserId :> IdpDelete
 
--- | This is a bit of a weird special case: we record the xml we get from the team admin as
--- verbatim as possible for trouble shooting, so we can't return it here with content type
--- 'XML' (that would trigger an attempt to encode an XML document, but we have a 'Text').
--- we'll just send it as 'PlainText', indicating that it's not parsed.
-type IdpGetRaw  = Capture "id" SAML.IdPId :> "raw" :> Get '[PlainText] Text
+type IdpGetRaw  = Capture "id" SAML.IdPId :> "raw" :> Get '[RawXML] Text
```
arianvp: Maybe a newtype `newtype Metadata = Metadata Text`

arianvp: This has the added benefit that you dont have an Orphan Instance for `Text`

fisx:
ci passed locally

PR #877, opened 2019-10-16
Tweak invite script

fisx:
Nothing big, part of it a bit opinionated, but I'm happy to remove anything that's controversial.

The most important change is probably that the output now contains the failed email addresses in the same lines as the errors, which makes it easier to generate a list of failed accounts with grep.

### DIFF
--- a/deploy/services-demo/create_team_members.sh
+++ b/deploy/services-demo/create_team_members.sh
@@ -73,8 +73,7 @@ do
 
     if ( ( echo "$INVITATION_ID" | grep -q '"code"' ) &&
          ( echo "$INVITATION_ID" | grep -q '"label"' ) ) ; then
-      echo "Got an error, aborting: $INVITATION_ID"
-      exit 1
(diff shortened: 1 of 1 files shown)


deploy/services-demo/create_team_members.sh
```diff
@@ -2,10 +2,10 @@
 
 set -e
 
-ADMIN_UUID="a09e9521-e14e-4285-ad71-47caa97f4a16"
-TEAM_UUID="9e57a378-0dca-468f-9661-7872f5f1c910"
-BRIG_HOST="http://localhost:8082"
-CSV_FILE="myfile.csv"
+ADMIN_UUID="n/a"
+TEAM_UUID="n/a"
+BRIG_HOST="http://localhost:8080"
```
tiago-loureiro: Note that the `8082` default was due to the fact that this is generally the port used by `brig` in integration tests and when [starting services](https://github.com/wireapp/wire-server/blob/develop/services/start-services-only.sh)

PR #878, opened 2019-10-16
Support HEAD requests for `/sso/initiate-bind`

fisx:


fisx:

PR #883, opened 2019-10-21
Change SCIM documentation to actually use SAML Ids

arianvp:
UUIDs are a bit confusing here. It is more common that your SAML provider is configured to have email addresses isntead

### DIFF
--- a/docs/reference/provisioning/scim-via-curl.md
+++ b/docs/reference/provisioning/scim-via-curl.md
@@ -97,10 +97,19 @@ export SCIM_USER='{
   "displayName" : "The Nick"
 }'
 ```
-The `externalId` is used to construct a 'SAML.UserRef'. If it looks like an email
-address, the constructed 'SAML.UserRef' will have `nameid-format:emailAddress`,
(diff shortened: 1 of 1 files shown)


docs/reference/provisioning/scim-via-curl.md
```diff
@@ -92,19 +92,25 @@ A minimal definition of a user looks like this:
...
   "userName"    : "nick",
   "displayName" : "The Nick"
 }'
 ```
+The `externalId` is used to construct a 'SAML.UserRef'. If it looks like an email
+address, the constructed 'SAML.UserRef' will have `nameid-format:emailAddress`,
+otherwise the format will be `unspecified`.
+*It is important to configure your SAML provider to use `nameid-format:emailAddress` or `unspecified`. Other
+nameid formats are not supported at this moment*.
+
```
fisx:
```suggestion
See also: https://github.com/wireapp/wire-server/blob/c507ed64a7d4f0af2bffe2f9c3eb4b5f89a477c0/services/spar/src/Spar/Scim/User.hs#L149-L158
```

PR #891, opened 2019-10-24
benchmarking for O(n) function, remove expensive call from atomicModifyIORef

fisx:
Just a little experiment to extablish whether we really want to call the more expensive `budgetSpent`, and not just `bsize`, for the metrics.


services/gundeck/src/Gundeck/ThreadBudget.hs
```diff
@@ -23,10 +23,14 @@ module Gundeck.ThreadBudget
   , runWithBudget'
   , watchThreadBudgetState
 
-  -- * for testing
+  -- * for testing: FUTUREWORK: factor out into a library?
```
fisx: I suggest we move this module to `Gundeck.ThreadBudget.Internal`, and make `Gundeck.ThreadBudget` a very simple module that only re-exports what we need for non-testing purposes.  Perhaps you can do that in #892?

PR #892, opened 2019-10-24
emit counter metric on soft/hard limit breached

jschaul:
(I wonder actually whether `runWithBudget` could not be within the Gundeck Monad, so as not to need to pass `metrics` around, but maybe that's another discussion. There is no MonadMetrics, is there?)

This PR can be (adapted if needed) and merged by the reviewer.


services/gundeck/src/Gundeck/ThreadBudget.hs
```diff
@@ -123,22 +123,22 @@ unregister ref key
...
   key <- liftIO nextRandom
   (`finally` unregister ref key) $ do
     oldsize <- allocate ref key spent
 
-    warnNoBudget (maybe False (oldsize >=) (limits ^. limitSoft))
-                 (maybe False (oldsize >=) (limits ^. limitHard))
-                 oldsize
+    let softLimitBreached = (maybe False (oldsize >=) (limits ^. limitSoft))
+        hardLimitBreached = (maybe False (oldsize >=) (limits ^. limitHard))
+    warnNoBudget softLimitBreached hardLimitBreached oldsize
```
fisx: this is better.  thanks!

fisx:
> (I wonder actually whether `runWithBudget` could not be within the Gundeck Monad, so as not to need to pass `metrics` around, but maybe that's another discussion. There is no MonadMetrics, is there?)

I wouldn't bother to change anything now, but if I would, it'd be something like this:

```haskell
data LimitKind = Hard | Soft

class MonadThreadBudgetReport m where
    reportLimitBreached :: MaxConcurrentNativePushes -> LimitKind -> Int -> m ()
```

That would make it straight-forward to make this a stand-alone package (except maybe for the Options stuff).

Also note https://github.com/wireapp/wire-server/pull/892/commits/3d6229edcf0a3a4074dbd06497bada76534a27fd

PR #893, opened 2019-10-28
Change order in which user data is created in spar

fisx:
before: `[ scim -> ] saml -> brig`
now: `brig -> saml [ -> scim ]`

Rationale: creation of users via scim (or implicitly via saml) is
idempotent; dealing with zombie users in brig is easier than handling
all errors correctly that can happen in brig during creating a user.


tiago-loureiro:
@arianvp we can close this now, right?

PR #900, opened 2019-11-04
throttle galley->gundeck on team deletion

jschaul:
Even though #897 already reduces the amount of events sent on team deletions, the function logic on team deletions is unbounded in the number of events sent in parallel: Every non-team member of every team conversation gets an event for each conversation that is deleted (N=unbounded), and every team member gets an event that their team has been deleted (M=MAX_TEAM_SIZE).
This PR thus reduces the events sent in parallel from M+N=unbounded at once to `N=MAX_CONV_SIZE - 1` and `M=128` during team deletion, irrespective of team size.

Note that these constraints apply only to team deletions, not other events.

Also adds a delay when processing user deletions.

### DIFF
--- a/services/brig/src/Brig/InternalEvent/Process.hs
+++ b/services/brig/src/Brig/InternalEvent/Process.hs
@@ -23,6 +23,10 @@ onEvent n = handleTimeout $ case n of
         Log.info $ msg (val "Processing user delete event")
                 ~~ field "user" (toByteString uid)
         API.lookupAccount uid >>= mapM_ API.deleteAccount
+        -- As user deletions are expensive resource-wise in the context of
+        -- bulk user deletions (e.g. during team deletions),
+        -- wait 50ms before processing the next event
+        liftIO $ threadDelay 50000
     DeleteService pid sid -> do
         Log.info $ msg (val "Processing service delete event")
                 ~~ field "provider" (toByteString pid)
--- a/services/brig/src/Brig/InternalEvent/Process.hs
+++ b/services/brig/src/Brig/InternalEvent/Process.hs
@@ -4,7 +4,9 @@ module Brig.InternalEvent.Process
 
 import Imports
 import Brig.App
+import Brig.Options (setDeleteThrottleMillis, defDeleteThrottleMillis)
 import Brig.InternalEvent.Types
+import Control.Lens (view)
 import Control.Monad.Catch
 import Data.ByteString.Conversion
 import System.Logger.Class (field, msg, (~~), val)
@@ -25,8 +27,9 @@ onEvent n = handleTimeout $ case n of
         API.lookupAccount uid >>= mapM_ API.deleteAccount
         -- As user deletions are expensive resource-wise in the context of
         -- bulk user deletions (e.g. during team deletions),
-        -- wait 50ms before processing the next event
-        liftIO $ threadDelay 50000
+        -- wait 'delay' ms before processing the next event
+        delay <- fromMaybe defDeleteThrottleMillis . setDeleteThrottleMillis <$> view settings
+        liftIO $ threadDelay (1000 * delay)
     DeleteService pid sid -> do
         Log.info $ msg (val "Processing service delete event")
                 ~~ field "provider" (toByteString pid)
--- a/services/galley/src/Galley/API/Teams.hs
+++ b/services/galley/src/Galley/API/Teams.hs
@@ -61,6 +61,7 @@ import Network.Wai.Predicate hiding (setStatus, result, or)
 import Network.Wai.Utilities
 import UnliftIO (mapConcurrently)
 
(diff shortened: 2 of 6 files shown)


services/brig/src/Brig/InternalEvent/Process.hs
```diff
@@ -23,6 +23,10 @@ onEvent n = handleTimeout $ case n of
         Log.info $ msg (val "Processing user delete event")
                 ~~ field "user" (toByteString uid)
         API.lookupAccount uid >>= mapM_ API.deleteAccount
+        -- As user deletions are expensive resource-wise in the context of
+        -- bulk user deletions (e.g. during team deletions),
+        -- wait 50ms before processing the next event
+        liftIO $ threadDelay 50000
```
q3k: Make this configurable by a flag.

Shouldn't this be rate limited at a higher level than within an event handler?

services/galley/src/Galley/API/Teams.hs
```diff
@@ -196,7 +197,14 @@ uncheckedDeleteTeam zusr zcon tid = do
         (ue, be) <- foldrM (pushConvDeleteEvents now membs) ([],[]) convs
         let e = newEvent TeamDelete tid now
         let r = list1 (userRecipient zusr) (membersToRecipients (Just zusr) membs)
-        pushSome ((newPush1 zusr (TeamEvent e) r & pushConn .~ zcon) : ue)
+        -- To avoid DoS on gundeck, send team deletion events in chunks
+        let chunks = List.chunksOf 128 (toList r)
```
q3k: Make this configurable by flag.

services/brig/src/Brig/InternalEvent/Process.hs
jschaul: User deletions get pushed on an SQS queue. This event handler then pulls those events from a queue and processes them one by one. Queuing is used here mainly because user deletions entail a lot of deletions in different places. Without queuing, if while a user deletion is ongoing, a process crashes, for example, the user deletion would only be half-way done. This way, with events, upon a crash, the event will be retried (which is fine, deletion requests are idempotent in their end state). We want to ensure everything gets deleted if a user deletes their account.

What do you mean with "at a higher level" ?

services/galley/src/Galley/API/Teams.hs
jschaul: I'm not against making this configurable, but I don't think it's very worthwhile. What is the exact benefit you see? Or are you thinking in a logic of "one shall not put magic numbers anywhere into a code base" ?

services/brig/src/Brig/InternalEvent/Process.hs
tiago-loureiro: I think this can even be a little "slower", e.g., `100ms`.

q3k: I was hoping whatever event library we use could have a per-endpoint rate limiting system. If that's not the case, then I guess this is okay for now.

services/galley/src/Galley/API/Teams.hs
q3k: More of a logic of 'anything that can be useful to tune at runtime by SREs should be a flag'. We're already discussing about 50 vs 100ms, and I also think this should be adjustable based on performance and testing. Not having to recompile the code to be able to tune things is nice.

services/galley/src/Galley/API/Teams.hs
```diff
@@ -196,7 +197,14 @@ uncheckedDeleteTeam zusr zcon tid = do
...
+        forM_ chunks $ \chunk -> case chunk of
+            [] -> return ()
+            x:xs -> push1 (newPush1 zusr (TeamEvent e) (list1 x xs) & pushConn .~ zcon)
+        -- To avoid DoS on gundeck, send conversation deletion events slowly
+        forM_ ue $ \event ->
```
tiago-loureiro: How about sleeping _a bit_ also here (maybe even 10ms would help) before `push1 event` ? These can be a lot of events (basically unbound as mentioned) so it could help too.

tiago-loureiro:
Other than the comments, 👍

services/galley/src/Galley/API/Teams.hs
jschaul: These can't be unbounded, each `event` here can be as long as the amount of participants of one single conversation. Or am I reading the code wrong?
It's onbounded when sending all of them in a `pushSome ue`, which is what it did before.

tiago-loureiro: Right, I just meant the `ue` is a list of infinite size, right? The size is basically nr of conversations where non team members are present, right? So if we have 1k events, we will do 1k requests to gundeck, right? Or am _I_ reading it wrong? :)

PR #904, opened 2019-11-11
add local dist folders in case stack confuses root folder with local …

jschaul:
…directory

Stack version 1 and 2 seem to behave differently with respect to their current working directory, leading to executables being written to e.g. `wire-server/services/brig/dist/brig`, instead of into `wire-server/dist/brig`.

This is an attempt to work around the problem.


arianvp:
Will this work with our `makedeb` machinery, which wants to put the `deb` in the local `dist` instead of the toplevel `dist` ?

jschaul:
> Will this work with our makedeb machinery, which wants to put the deb in the local dist instead of the toplevel dist ?

Yes, it will place the `.deb` file into the local dist folder, which is then the root dist folder. I just tried a `VERSION=0.0.1 make dist` - it works for me.

arianvp:
LGTM then.

This is rather a pragmatic fix, which is fine, but it's still kinda confusing that there are 2 `dist` folders at play here. (Though they're now the same with a symlink).  E.g. the `.deb`s were always installed to the local `dist` folder AFAIK and I don't really know if there was a good reason for that.  (not sure what the context is; or if CI scripts for the deb files depend on locations etc)


We should probably at some point clean up the makefiles a bit to just use the same path everywhere. Also I think most of the makefiles are nearly identical for each of the services; might be worth to have just one makefile with some parameters instead.  (definitely future work).

jschaul:
> not sure what the context is; or if CI scripts for the deb files depend on locations etc)

historical, no good reason. All the services (such as `brig` and `galley`) used to be in their own github repository, and had their own dist folder. The deb file machinery is still from that time (before 2017) and was just never changed.

Agreed to try and use the same dist path everywhere. Consolidating makefiles might not be trivial, but if you know of a good way to do that, sure.

PR #906, opened 2019-11-13
metrics-wai: cleanup/refactor prometheus middleware

jschaul:


stack.yaml
```diff
@@ -42,18 +42,12 @@ extra-deps:
 - servant-swagger-1.1.6
 - swagger2-2.3
 - git: https://github.com/wireapp/prometheus-haskell
```
tiago-loureiro: Since `wireapp` and `fimap` are now in sync, how about we pin to `fimap` instead?

stack.yaml
```diff
@@ -42,18 +42,12 @@ extra-deps:
...
-  # version including
-  # https://github.com/fimad/prometheus-haskell/pull/44 has been
-  # merged (any time after Tue 21 May 2019 03:42:48 PM CEST).
-  #
-  # an even better fix is in
-  # https://github.com/fimad/prometheus-haskell/pull/45; see comment
-  # in /libs/metrics-wai/src/Data/Metrics/Middleware/Prometheus.hs at
-  # `instrumentHandlerValue`.
+  # version 0.1.2.0 or later is available on hackage
+  # (required: https://github.com/fimad/prometheus-haskell/pull/45)
```
tiago-loureiro: Small nitpick: what we're looking for is the [wai-middleware-prometheus package](http://hackage.haskell.org/package/wai-middleware-prometheus), which may very well not be the version stated here (since that is used for `prometheus-proc`).

Basically any bump/publish by the author of the `wai-middleware-prometheus` should be good to go.

PR #909, opened 2019-11-14
services/brig: pave way for local whitelist

q3k:
Currently whitelisting for account activation is performed via an
external service, if configured.

We would like to move logic from the external service into Wire itself.
As such, this change brings in configuration format changes that will
allow to select either external or internal whitelisting.

For now, internal whitelisting only supports a statically configured
list of allowed email domains. In the future, it will gain extra
features like in-database storage of allowed emails.

### DIFF
--- a/docs/reference/user/activation.md
+++ b/docs/reference/user/activation.md
@@ -129,7 +129,22 @@ X-Zeta-Code: 123456
 
 ## Phone/email whitelist {#RefActivationWhitelist}
 
-The backend can be configured to only allow specific phone numbers or email addresses to register. The following options have to be set in `brig.yaml`:
+The backend can be configured to only allow specific phone numbers or email addresses to register.
+
+There are two whitelisting methods:
+
+ - contact an external whitelisting service by HTTP
+ - use built-in whitelisting
+
+Both are configured by settings options in `brig.yaml`. If both are configured simultaneously, internal whitelisting is ignored.
+
+If an email address or phone number are rejected by the whitelist, `POST /activate/send` or `POST /register` will return `403 Forbidden`:
+
+### External whitelisting
+
+External whitelisting is **deprecated** and will be removed in a future version of Wire-Server.
+
+The following options have to be set in `brig.yaml`:
 
 ```yaml
 optSettings:
@@ -141,7 +156,20 @@ optSettings:
 
 When those options are present, the backend will do a GET request at `<whitelistUrl>?email=...` or `<whitelistUrl>?mobile=...` for every activation request it receives. It will expect either status code 200 ("everything good") or 404 ("provided email/phone is not on the whitelist").
 
-If an email address or phone number are rejected by the whitelist, `POST /activate/send` or `POST /register` will return `403 Forbidden`:
+### Internal whitelisting
+
+Internal whitelisting currently only supports filtering by email domain. Activations/registrations will only be allowed from emails that are directly specified in a static list of domains.
+
+The following options have to be set in `brig.yaml`:
+
+```yaml
+optSettings:
+  setWhitelist:
+    whitelistDomains:
+     - example.com
+     - contractor.com
+```
+
 
 ```json
 {
--- a/services/brig/src/Brig/Whitelist.hs
+++ b/services/brig/src/Brig/Whitelist.hs
@@ -44,34 +44,34 @@ deriveJSON toOptionFieldName 'internalWhitelistDomains
 -- consults a statically configured whitelist of allowed email domains..
 verifyInternal :: InternalWhitelist -> Either Email Phone -> Bool
 verifyInternal (InternalWhitelist domains) key =
-  let allowFromDomains (Left e) = (emailDomain e) `elem` domains
-      allowFromDomains _        = False
-  in allowFromDomains key
+    let allowFromDomains (Left e) = (emailDomain e) `elem` domains
+        allowFromDomains _        = False
+    in allowFromDomains key
 
 -- | Do a request to the whitelist service and verify that the provided email/phone address is
 -- whitelisted.
 verifyService :: (MonadIO m, MonadMask m, MonadHttp m) => Whitelist -> Either Email Phone -> m Bool
 verifyService (Whitelist url user pass) key = if isKnownDomain key
-  then return True
-  else recovering x3 httpHandlers . const $ do
-     rq <- parseRequest $ T.unpack url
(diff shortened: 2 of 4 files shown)


q3k:
Reviewers: how do I write tests for this? We seem to have no unit tests in this area of Wire Server?

services/brig/src/Brig/Whitelist.hs
```diff
@@ -1,44 +1,68 @@
...
 instance FromJSON Whitelist
 
+data InternalWhitelist = InternalWhitelist
+    { internalWhitelistDomains :: ![Text]} deriving (Show, Generic)
+
+deriveJSON toOptionFieldName 'internalWhitelistDomains
+
+-- | Check internal whitelist for given email/phone number. Currently this
+-- consults a statically configured whitelist of allowed email domains..
+verifyInternal :: (MonadIO m, MonadMask m, MonadHttp m) => InternalWhitelist -> Either Email Phone -> m Bool
```
jschaul: Why not simply make this function signature `verifyInternal :: InternalWhitelist -> Either Email Phone -> Bool` ? You're not using monadic effects here.

This would also make unit testing this function easy.

services/brig/src/Brig/Whitelist.hs
```diff
@@ -1,44 +1,68 @@
...
-    , whitelistPass :: !Text     -- ^ Password
+    , whitelistUser :: !Text     -- ^ Service Username (basic auth)
+    , whitelistPass :: !Text     -- ^ Service Password (basic auth)
+
     } deriving (Show, Generic)
 
 instance FromJSON Whitelist
 
+data InternalWhitelist = InternalWhitelist
+    { internalWhitelistDomains :: ![Text]} deriving (Show, Generic)
```
jschaul: Use newtype:

```
newtype InternalWhitelist = InternalWhitelist [Text]
```

services/brig/src/Brig/Whitelist.hs
q3k: To keep the API symmetrical between `verifyInternal` and `verifyService`. In addition, `verifyInternal` will need to be in IO (or something) once it starts checking cassandra for specific whitelisting.

q3k:
Yes, this is the first PR of a series to introduce more and more whitelisting functionality internally.

For testing, I'll probably try to integrate this into integration testing.

services/brig/src/Brig/Whitelist.hs
jschaul: Not a good reason (You can always `pure Whitelist.verifyInternal b key` in Handler.hs.), and could be refactored later if the need arises. Also, it will never need MonadHttp, nor MonadMask, even if it talks to cassandra (it would need `MonadClient` at that point). I would generally recommend to keep function signatures as lean as you can.

services/brig/src/Brig/Whitelist.hs
```diff
@@ -1,44 +1,68 @@
...
+import Data.Aeson.TH (deriveJSON)
 import Bilge.IO
 import Bilge.Request
 import Bilge.Response
 import Bilge.Retry
 import Brig.Types
 import Control.Monad.Catch (MonadMask, throwM)
 import Control.Retry
-import Data.Text
+import qualified Data.Text as T
```
jschaul: Import lists are generally (across the codebase) split into two lists, one with `import ...` and one with `import qualified ... as ...`, separated by a blank line. See for example [here](https://github.com/wireapp/wire-server/blob/develop/services/brig/src/Brig/API.hs#L33-L40)

services/brig/src/Brig/Whitelist.hs
q3k: But that changes the config format from

```
setInternalWhitelist:
    whitelistDomains: [...]
```

into 

```
setInternalWhitelist: [...]
```

right?

services/brig/src/Brig/Whitelist.hs
q3k: Done.

services/brig/src/Brig/Whitelist.hs
q3k: Done.

services/brig/src/Brig/Whitelist.hs
jschaul: Correct. You could possibly rename the field in Options.hs to `setWhitelistDomains` (or something). This suggestion (using `newtype`) would remove one level of nesting. If you think you'll need other whitelist-related configurations other than the domains (I can't currently think of any), and you prefer deeply nested configs rather than configs where settings share a common prefix in their name, then feel free to leave this as-is.

services/brig/src/Brig/Whitelist.hs
```diff
@@ -1,44 +1,69 @@
...
+
+deriveJSON toOptionFieldName 'internalWhitelistDomains
+
+-- | Check internal whitelist for given email/phone number. Currently this
+-- consults a statically configured whitelist of allowed email domains..
+verifyInternal :: InternalWhitelist -> Either Email Phone -> Bool
+verifyInternal (InternalWhitelist domains) key =
+  let allowFromDomains (Left e) = (emailDomain e) `elem` domains
+      allowFromDomains _        = False
+  in allowFromDomains key
```
jschaul: so far, brig uses 4-space indentation everywhere (just like galley/cannon/proxy/cargohold - some code in `spar` has been 2-space indented, but at least so far it's consistent within the same file). We can of course have a debate on this, but generally I'd suggest keeping the indentation level of files as they are when modifying them (and not selectively indent some functions differently - that makes working with those files later much harder)

q3k: Done.

docs/reference/user/activation.md
```diff
@@ -141,7 +156,20 @@ optSettings:
...
-If an email address or phone number are rejected by the whitelist, `POST /activate/send` or `POST /register` will return `403 Forbidden`:
+### Internal whitelisting
+
+Internal whitelisting currently only supports filtering by email domain. Activations/registrations will only be allowed from emails that are directly specified in a static list of domains.
+
+The following options have to be set in `brig.yaml`:
+
+```yaml
+optSettings:
+  setWhitelist:
```
jschaul: This needs to be `setInternalWhitelist`, I believe.

tiago-loureiro:
I will close this PR as all the code here is also part of https://github.com/wireapp/wire-server/pull/982. The discussions / arguments raised are still valuable though

PR #915, opened 2019-11-25
Remove unnecessary fanout team events

tiago-loureiro:
Currently, `team.conversation-create` and `team.conversation-delete` events are sent to _all_ team members that are _not_ part of the conversation.

While this may theoretically be useful for conversation discovery in case of teams (e.g., you may be interested in joining these conversations) there's currently no UI nor a use case for it so we're better off avoiding unnecessary fan outs. In case this does become relevant again, one could simply undo (redo?) this PR.

### DIFF
--- a/services/brig/test/integration/API/Provider.hs
+++ b/services/brig/test/integration/API/Provider.hs
@@ -643,7 +643,7 @@ testDeleteConvBotTeam config db brig galley cannon = withTestService config db b
         -- 200 response on success
         Team.deleteTeamConv galley tid cid uid2
         -- Events for the users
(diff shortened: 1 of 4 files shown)


fisx:
can we close these, then?  (not sure, just checking)

- https://github.com/zinfra/backend-issues/issues/747

it looks like we've spent quite a bit of time on this topic in the past:

- https://github.com/zinfra/backend-issues/issues/677
- https://github.com/zinfra/backend-issues/issues/607
- https://github.com/zinfra/backend-issues/issues/573
- https://github.com/zinfra/backend-issues/issues/554

PR #921, opened 2019-12-09
Handle search within team

tiago-loureiro:
This solves the problem for on premise deployments where it is not desirable have team members connecting outside their team

### DIFF
--- a/services/brig/test/integration/API/Search.hs
+++ b/services/brig/test/integration/API/Search.hs
@@ -41,7 +41,6 @@ testOptInOut brig = do
     refreshIndex brig
     assertSearchable "opted in" brig uid1 True
     assertCanFind brig uid2 uid1 h1
-    -- Even if users opt in, team users in other teams should not be found
 
 testSearchByName :: Brig -> Http ()
 testSearchByName brig = do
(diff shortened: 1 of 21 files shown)


tiago-loureiro:
> This only touches `/users/handles/<handle>` but it doesn't change the logic of `/search/contacts?q=<prefix>`, meaning users still can connect to users outside their team (team users if they set their searchable status to true, and to users not part of a team).

Right, although there's no UI for this (setting status to true) and typically admins ensure that no non-team accounts can be created. However, your point is still valid, I will touch the other endpoint as well, thanks @jschaul !

> Please also link to an issue describing the problem that this PR attempts to solve.

Sure.

tiago-loureiro:
I forgot to comment again, ready for re-review.

services/brig/test/integration/API/Search.hs
```diff
@@ -41,7 +41,7 @@ testOptInOut brig = do
     refreshIndex brig
     assertSearchable "opted in" brig uid1 True
     assertCanFind brig uid2 uid1 h1
-
+    -- Even if users opt in, team users in other teams should not be found
```
jschaul: Did you mean to write a few more test lines here? Or what does this comment relate to? (maybe move to the test case that does this / delete this line)

tiago-loureiro: Ah please ignore this, it was moved to `testNonSearchableAcrossTeam`

PR #922, opened 2019-12-11
brig: fix integration tests after #867

q3k:
Seems to have been accidentally broken in #867 / 870b10b80a3dceaae07da21bbed29fc1d025f627 .

Log:

    Brig API Integration
      user
        property
          put /properties/:key - 403:                                FAIL
            Exception: Assertions failed:
             1: 403 =/= 200
             2: Just "property-key-too-large" =/= Nothing

### DIFF
--- a/services/brig/test/integration/API/User/Property.hs
+++ b/services/brig/test/integration/API/User/Property.hs
@@ -111,12 +111,12 @@ testPropertyLimits (Just opts) brig = do
         maxValueLen = fromIntegral $ fromMaybe defMaxValueLen . setPropertyMaxValueLen $ optSettings opts
 
     -- Maximum key length
-    setProperty brig (userId u) (C.replicate (maxKeyLen+1) 'x') (String "y") !!! do
+    setProperty brig (userId u) (C.replicate (maxKeyLen + 1) 'x') (String "y") !!! do
         const 403 === statusCode
         const (Just "property-key-too-large") === fmap Error.label . responseJsonMaybe
 
     -- Maximum value length
-    setProperty brig (userId u) "foo" (String (T.replicate (maxValueLen+1) "x")) !!! do
+    setProperty brig (userId u) "foo" (String (T.replicate (maxValueLen + 1) "x")) !!! do
         const 403 === statusCode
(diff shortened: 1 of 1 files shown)


arianvp:
Interesting. Why did this pass on CI and on my machine as well? 

because we do not have a configuration that tests this behaviour by default?

q3k:
No idea! This was breaking tests ran by `make i` in wire-server. What tests did you run on your machine?

services/brig/test/integration/API/User/Property.hs
```diff
@@ -102,16 +102,21 @@ testClearProperties brig = do
...
+testPropertyLimits Nothing _ = error "no config!"
+testPropertyLimits (Just opts) brig = do
     u <- randomUser brig
+
+    let maxKeyLen = fromIntegral $ fromMaybe defMaxKeyLen . setPropertyMaxKeyLen $ optSettings opts
+        maxValueLen = fromIntegral $ fromMaybe defMaxValueLen . setPropertyMaxValueLen $ optSettings opts
+
     -- Maximum key length
-    setProperty brig (userId u) (C.replicate 257 'x') (String "y") !!! do
+    setProperty brig (userId u) (C.replicate (maxKeyLen+1) 'x') (String "y") !!! do
```
jschaul:
```suggestion
    setProperty brig (userId u) (C.replicate (maxKeyLen + 1) 'x') (String "y") !!! do
```

services/brig/test/integration/API/User/Property.hs
```diff
@@ -102,16 +102,21 @@ testClearProperties brig = do
...
         const 403 === statusCode
         const (Just "property-key-too-large") === fmap Error.label . responseJsonMaybe
 
     -- Maximum value length
-    setProperty brig (userId u) "foo" (String (T.replicate 513 "x")) !!! do
+    setProperty brig (userId u) "foo" (String (T.replicate (maxValueLen+1) "x")) !!! do
```
jschaul:
```suggestion
    setProperty brig (userId u) "foo" (String (T.replicate (maxValueLen + 1) "x")) !!! do
```

PR #926, opened 2019-12-18
SCIM Fixes Phase 1 + 2

arianvp:
- [x] Implement PATCH
- [x] Add tests for patch (Almost done; except for `remove`)
- [x] Fix performance issues with `getUsers`
- [x] Allow SCIM to discover existing brig users (`getUsers` should synthesize SCIM records)
      - For now; only brig users that are SSO users. to be sure.
- [x] Add tests for existing brig user discovery
- [x] Either move `hscim` to `wire-server` or update `package.yaml` to point to correct version

### DIFF
--- a/services/spar/src/Spar/Intra/Brig.hs
+++ b/services/spar/src/Spar/Intra/Brig.hs
@@ -162,8 +162,11 @@ getBrigUser buid = do
 -- | Get a list of users; returns a shorter list if some 'UserId's come up empty (no errors).
 --
 -- TODO: implement an internal end-point on brig that makes this possible with one request.
+-- TODO(arianvp): This endpoint exists!
 getBrigUsers :: (HasCallStack, MonadSparToBrig m) => [UserId] -> m [User]
 getBrigUsers = fmap catMaybes . mapM getBrigUser
+
+
 
 -- | Get a user; returns 'Nothing' if the user was not found.
 --
--- a/services/spar/src/Spar/Intra/Brig.hs
+++ b/services/spar/src/Spar/Intra/Brig.hs
@@ -269,7 +269,7 @@ setBrigUserRichInfo buid richInfo = do
      | otherwise
        -> throwSpar . SparBrigError . cs $ "set richInfo failed with status " <> show sCode
 
-getBrigUserRichInfo :: (HasCallStack, MonadSparToBrig m) => UserId -> m RichInfo
+getBrigUserRichInfo :: (HasCallStack, MonadSparToBrig m) => UserId -> m (Maybe RichInfo)
 getBrigUserRichInfo buid = do
   resp <- call
     $ method GET
@@ -278,7 +278,8 @@ getBrigUserRichInfo buid = do
     . header "Z-Connection" ""
   case statusCode resp of
     200 -> do
-      parseResponse @RichInfo resp
+      Just <$> parseResponse @RichInfo resp
+    404   -> pure Nothing
     _   -> throwSpar (SparBrigErrorWith (responseStatus resp) "Could not retrieve rich info")
     
 
--- a/services/spar/src/Spar/Intra/Brig.hs
+++ b/services/spar/src/Spar/Intra/Brig.hs
@@ -175,7 +175,6 @@ getBrigUsers :: (HasCallStack, MonadSparToBrig m) => [UserId] -> m [User]
 getBrigUsers = fmap catMaybes . mapM getBrigUser
 
 
-
 -- | Get a user; returns 'Nothing' if the user was not found.
 --
 -- TODO: currently this is not used, but it might be useful later when/if
@@ -269,7 +268,8 @@ setBrigUserRichInfo buid richInfo = do
      | otherwise
        -> throwSpar . SparBrigError . cs $ "set richInfo failed with status " <> show sCode
 
-getBrigUserRichInfo :: (HasCallStack, MonadSparToBrig m) => UserId -> m (Maybe RichInfo)
+-- TODO: We should add an internal endpoint for this instead
+getBrigUserRichInfo :: (HasCallStack, MonadSparToBrig m) => UserId -> m RichInfo
 getBrigUserRichInfo buid = do
   resp <- call
     $ method GET
@@ -277,9 +277,7 @@ getBrigUserRichInfo buid = do
     . header "Z-User" (toByteString' buid)
     . header "Z-Connection" ""
   case statusCode resp of
-    200 -> do
-      Just <$> parseResponse @RichInfo resp
-    404   -> pure Nothing
+    200 -> parseResponse @RichInfo resp
     _   -> throwSpar (SparBrigErrorWith (responseStatus resp) "Could not retrieve rich info")
     
 
--- a/services/spar/src/Spar/Intra/Brig.hs
+++ b/services/spar/src/Spar/Intra/Brig.hs
@@ -4,10 +4,12 @@
 module Spar.Intra.Brig
   ( toUserSSOId
   , fromUserSSOId
+  , toExternalId
   , getBrigUser
   , getBrigUserTeam
   , getBrigUsers
   , getBrigUserByHandle
+  , getBrigUserRichInfo
   , setBrigUserName
   , setBrigUserHandle
   , setBrigUserManagedBy
@@ -62,6 +64,12 @@ fromUserSSOId (UserSSOId (cs -> tenant) (cs -> subject)) =
     (Left msg, _)      -> throwError msg
     (_, Left msg)      -> throwError msg
 
+-- | Converts a brig User SSO Id into an external id
+toExternalId :: MonadError SparError m => UserSSOId -> m Text
+toExternalId ssoid = do
+  uref <- either (throwSpar . SparCouldNotParseBrigResponse . cs) pure $ fromUserSSOId ssoid
+  let subj = uref ^. SAML.uidSubject
+  pure $ SAML.nameIDToST subj
 
 parseResponse :: (FromJSON a, MonadError SparError m) => Response (Maybe LBS) -> m a
 parseResponse resp = do
@@ -260,6 +268,19 @@ setBrigUserRichInfo buid richInfo = do
        -> throwSpar . SparBrigErrorWith (responseStatus resp) $ "set richInfo failed"
      | otherwise
        -> throwSpar . SparBrigError . cs $ "set richInfo failed with status " <> show sCode
+
+getBrigUserRichInfo :: (HasCallStack, MonadSparToBrig m) => UserId -> m RichInfo
+getBrigUserRichInfo buid = do
+  resp <- call
+    $ method GET
+    . paths [ "users", toByteString' buid, "rich-info" ]
+    . header "Z-User" (toByteString' buid)
+    . header "Z-Connection" ""
+  case statusCode resp of
+    200 -> do
+      parseResponse @RichInfo resp
+    _   -> throwSpar (SparBrigErrorWith (responseStatus resp) "Could not retrieve rich info")
+    
 
 -- | At the time of writing this, @HEAD /users/handles/:uid@ does not use the 'UserId' for
 -- anything but authorization.
--- a/services/spar/src/Spar/Scim/User.hs
+++ b/services/spar/src/Spar/Scim/User.hs
@@ -70,7 +70,16 @@ instance Scim.UserDB SparTag Spar where
            -> Maybe Scim.Filter
            -> Scim.ScimHandler Spar (Scim.ListResponse (Scim.StoredUser SparTag))
   getUsers ScimTokenInfo{stiTeam} mbFilter = do
+    -- TODO(arianvp): Rewrite this procedure completely. It already doesn't
+    -- work for customers using SCIM. We should special-case the filter, and do
+    -- smart stuff, Especially when the filter provided filters on a primary
+    -- key, we can do a quick database lookup instead.  and for now we only
+    -- support filters on primary keys anyway
+    --
+    -- DECISION: We will not support the full getUsers. It is not used by Okta
+    -- nor Azure. If the user does not provide a Filter, we will error out
     members <- lift $ getTeamMembers stiTeam
+    -- TODO(arianvp): FIXME FIXME: getBrigUsers does O(n) getBrigUser calls
     brigusers :: [User]
       <- filter (not . userDeleted) <$>
          lift (Intra.Brig.getBrigUsers ((^. Galley.userId) <$> members))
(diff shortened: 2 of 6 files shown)


arianvp:
See https://github.com/wireapp/hscim/pull/35 for bulk of the implementation of PATCH

services/spar/test-integration/Test/Spar/Scim/UserSpec.hs
```diff
@@ -463,13 +466,31 @@ specUpdateUser = describe "PUT /Users/:id" $ do
         it "fails to update user with 404" testUserUpdateFailsWithNotFoundIfOutsideTeam
     context "scim_user has no entry with this id" $ do
         it "fails" $ pending
+    it "does not update if nothing changed" $ testSameUpdateNoChange
     context "brig user is updated" $ do
+        -- TODO(arianvp): This will be fixed by
```
tiago-loureiro: A note here; we have a field in `brig` named: `managed_by`, and one of the values is `scim`. Ideally, users that are managed by `scim` should not be able to update fields since it makes sense for the IdP to be the source of truth.

services/spar/test-integration/Util/Scim.hs
```diff
@@ -72,6 +73,11 @@ registerScimToken teamid midpid = do
 randomScimUser :: MonadRandom m => m (Scim.User.User SparTag)
 randomScimUser = fst <$> randomScimUserWithSubject
 
+randomUserName :: MonadRandom m => m Text
+randomUserName = do
```
tiago-loureiro: This isn't really used anywhere is it?

services/spar/test-integration/Test/Spar/Scim/UserSpec.hs
```diff
@@ -622,6 +643,141 @@ testBrigSideIsUpdated = do
     brigUser      <- maybe (error "no brig user") pure =<< runSpar (Intra.getBrigUser userid)
     brigUser `userShouldMatch` validScimUser
 
+----------------------------------------------------------------------------
+-- Patcing users
```
tiago-loureiro:
```suggestion
-- Patching users
```

services/spar/test-integration/Util/Scim.hs
arianvp: Nope not anymore; I now use the existing randomUser and extract the name instead

services/spar/test-integration/Test/Spar/Scim/UserSpec.hs
```diff
@@ -622,6 +643,141 @@ testBrigSideIsUpdated = do
...
+            liftIO $ storedUser `shouldBe` storedUser'
+
+        -- NOTE: Remove at the moment actually never works! As all the fields
+        -- we support are required in our book
+        it "userName cannot be removed according to scim" $ do
+            (tok, _) <- registerIdPAndScimToken
+            user <- randomScimUser
+            storedUser <- createUser tok user
+            let userid = scimUserId storedUser
+            _ <- patchUser tok userid $ PatchOp.PatchOp
```
tiago-loureiro: `patchUser` asserts on a 200 but this returns a 400, no?

services/spar/test-integration/Test/Spar/Scim/UserSpec.hs
```diff
@@ -622,6 +643,141 @@ testBrigSideIsUpdated = do
...
+                [ removeAttrib "userName" ]
+            pure ()
+        it "displayName cannot be removed in spar (though possible in scim). Diplayname is required in Wire" $ do
+            (tok, _) <- registerIdPAndScimToken
+            user <- randomScimUser
+            storedUser <- createUser tok user
+            let userid = scimUserId storedUser
+            _ <- patchUser tok userid $ PatchOp.PatchOp
```
tiago-loureiro: This seems to currently return a 200, right? According to [the spec](https://tools.ietf.org/html/rfc7644#section-3.12) should this not return a 400?

services/spar/test-integration/Test/Spar/Scim/UserSpec.hs
```diff
@@ -622,6 +643,141 @@ testBrigSideIsUpdated = do
...
+                [ removeAttrib "displayName" ]
+            pure ()
+        it "externalId cannot be removed in spar (though possible in scim)" $ do
+            (tok, _) <- registerIdPAndScimToken
+            user <- randomScimUser
+            storedUser <- createUser tok user
+            let userid = scimUserId storedUser
+            _ <- patchUser tok userid $ PatchOp.PatchOp
```
tiago-loureiro: Same as above I think?

tiago-loureiro:
OK, I think some of my comments were answered with: https://github.com/wireapp/wire-server/commit/59f8eaa523b2bcc748882af03e51e942bba5e9d6 :)

arianvp:
I added the other tests (testing for `409` case and `404` case). Just need to fix the displayName test now ...

services/spar/src/Spar/Scim/User.hs
```diff
@@ -65,51 +65,54 @@ import qualified Web.Scim.Schema.User             as Scim.User (schemas)
...
+  -- can currently look for users based on handle or SSO ID. 
+  --
+  -- This endpoint will look for users in Brig, and then synthesize the
+  -- corresponding SCIM record.
+  -- look for SSO users in brig; and then return the corresponding SCIM record.
+  -- getUsers itself will not do any mutations in the SCIM database; as it
+  -- feels weird to me (arian) that a GET causes side-effects like that.
+  -- Instead, 'updateUser' will when updating a user see if the user was
+  -- already in the SCIM Table; and if not insert it on demand, just like
+  -- 'POST'.
```
arianvp: Actually I changed my mind here.

`getUsers` should insert the synthesized record into the scim database; and `updateUsers` should only set `mangedBy = SCIM` on the user.

Reason?  the SCIM record have a `version` and a `modifiedAt` field that need to be correct. The only way to do that is to set them during creation time (I can't come up with a `modifiedAt` out of thing air and then have it the same the next time `updateUser` is called unless I store it).

services/spar/src/Spar/Scim/User.hs
```diff
@@ -65,27 +65,54 @@ import qualified Web.Scim.Schema.User             as Scim.User (schemas)
...
-          Just filter_ ->
-            let user' = Scim.value (Scim.thing user)
-            in case Scim.filterUser filter_ user' of
-                 Right res -> pure res
-                 Left err  -> throwError $
-                   Scim.badRequest Scim.InvalidFilter (Just err)
-    -- FUTUREWORK: once bigger teams arrive, we should have pagination here.
-    Scim.fromList <$> filterM check scimusers
+  getUsers _ Nothing = do
+    -- TODO(arianvp): This shouldn't be a 501 according to joe
```
fisx: yes.  the rule: "if it should wake up admins at night even if it happens a lot, don't make it 500."

services/spar/src/Spar/Scim/User.hs
```diff
@@ -289,6 +307,12 @@ updateValidScimUser
   :: forall m. (m ~ Scim.ScimHandler Spar)
   => ScimTokenInfo -> UserId -> ValidScimUser -> m (Scim.StoredUser SparTag)
 updateValidScimUser tokinfo@ScimTokenInfo{stiIdP} uid newScimUser = do
+    -- TODO(arianvp):
+    --   getUsers currently synthesizes SCIM users out of thin air, but doesn't store
+    --   them in the SCIM table. Maybe it should. However one thing it for sure 
+    --   does not do is set the ManagedByScim flag, because get should be read-only
+    --   So PUT (which might be issued after a GET /Users) must set ManagedByScim
+    --   if this was not yet the case for this user
```
fisx: see above: it's weird, but get should both set the managed-by flag and write the scim data to spar.

services/spar/src/Spar/Scim/User.hs
arianvp: We deviate from the spec here though.  SCIM _requires_ a 501 because the SCIM authors are ... odd

arianvp: But maybe lets ignore their advice

services/spar/test-integration/Test/Spar/Scim/UserSpec.hs
```diff
@@ -416,6 +508,8 @@ testGetNoDeletedUsers = do
     -- Try to find the user
     getUser_ (Just tok) userid (env ^. teSpar)
         !!! const 404 === statusCode
+
+    -- TODO(arianvp): What does this mean; @fisx ??
```
fisx: see above?

services/spar/src/Spar/Intra/Brig.hs
```diff
@@ -258,6 +269,20 @@ setBrigUserRichInfo buid richInfo = do
...
+getBrigUserRichInfo buid = do
+  resp <- call
+    $ method GET
+    . paths [ "users", toByteString' buid, "rich-info" ]
+    . header "Z-User" (toByteString' buid)
+    . header "Z-Connection" ""
+  case statusCode resp of
+    200 -> do
+      Just <$> parseResponse @RichInfo resp
+    404   -> pure Nothing
```
tiago-loureiro: I think `brig` should never return a `404` -> unless the user is not found which would be some kind of DB inconsistency? If there's no rich info what is returned is a 200 with:

```
{"version":0,"fields":[]}
```

services/spar/src/Spar/Scim/User.hs
```diff
@@ -61,31 +64,101 @@ import qualified Web.Scim.Schema.ResourceType     as Scim
...
+    -- The code is rather problematic as it does O(n) HTTP requests to Brig (as
+    -- brig does not support getting users in bulk). Given that Azure would
+    -- send a /Users request _per user_ this would yield in O(n^2) requests,
+    -- which blows up extremely quickly for large teams.
+    --
+    -- NOTE: unless you supply a filter, it will only list
+    -- SCIM users, not all users in a team. That was the original behaviour
+    -- and we didn't feel like changing it for this as it's going to be removed
+    -- soon anyway. Please scream loudly in the review if you disagree with this,
+    -- but it would complicate this code a bit, instead of leaving it as is.
```
tiago-loureiro: Scream loudly in agreement :)

services/spar/src/Spar/Scim/User.hs
```diff
@@ -61,31 +64,101 @@ import qualified Web.Scim.Schema.ResourceType     as Scim
...
+    pure $ Scim.fromList scimusers
+
+  getUsers ScimTokenInfo{stiTeam, stiIdP} (Just filter') = do
+    idp <- stiIdP ?? Scim.serverError "No IdP configured for the provisioning token"
+    idpConfig <- (wrapMonadClient . Data.getIdPConfig $ idp)  !? Scim.serverError "No IdP configured for the provisioning token"
+    case filter' of
+      Scim.FilterAttrCompare (Scim.AttrPath schema attrName _subAttr) Scim.OpEq (Scim.ValString val)
+        | Scim.isUserSchema schema -> do
+            let
+              createOrGetScimUser :: BrigTypes.User -> MaybeT (Scim.ScimHandler Spar) (Scim.StoredUser SparTag)
```
tiago-loureiro: Could we move this to a separate function or a where clause? I find it hard to read this `case` statement which is really about checking the `attribute`?

services/spar/src/Spar/Scim/User.hs
```diff
@@ -501,6 +564,20 @@ assertHandleNotUsedElsewhere hndl uid = do
   unless ((userHandle =<< musr) == Just hndl) $
     assertHandleUnused' "userName does not match UserId" hndl uid
 
+-- | Fails if the handle isn't set
```
tiago-loureiro: How can this file given that `Handle` is not a `Maybe` on the `NeededInfo` datatype?

services/spar/src/Spar/Scim/User.hs
tiago-loureiro: Also: wouldn't it make this function simpler also if we don't use `MaybeT` but rather `Spar` directly?

arianvp: We need MaybeT for the short-circuit behaviour instead of returning an error. It's only used in this specific function. E.g. lists should never return 404 but an empty list instead

services/spar/src/Spar/Intra/Brig.hs
arianvp: Code doesn't agree with you. It returns a 404 if the user isn't found from what I could see

arianvp:
Self review: I forgot to set the managed_by flag :)  that's one extra line to add

services/spar/src/Spar/Scim/User.hs
arianvp: Because the comment is old. Thanks

services/spar/src/Spar/Intra/Brig.hs
tiago-loureiro: Yes; my point was more in the lines of: how did u get a userId on the SCIM to a user that does not exist? It exists on the SCIM table but not on brig? This sounds like a very inconsistent state, don't u agree?

services/spar/test-integration/Test/Spar/Scim/UserSpec.hs
```diff
@@ -355,13 +411,35 @@ testListNoDeletedUsers = do
     (tok, _) <- registerIdPAndScimToken
     storedUser <- createUser tok user
     let userid = scimUserId storedUser
-    -- Delete the user (TODO: do it via SCIM)
+    -- Delete the user (TODO: do it via SCIM) TODO(arianvp): Ask @fisx what he meant with this comment???
```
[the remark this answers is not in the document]
arianvp: I'm still not sure I understand. don't we have that function already? Are we testing something specific here. e.g. do we want to maintain the possibility to delete users from brig's perspective?

services/spar/src/Spar/Intra/Brig.hs
arianvp: What if the brig user is deleted from the brig side inbetween? This is a usecase we apparently support; given we have other tests that assert this behaviour (see `deleteUserOnBrig` calls in the test-suite).   However we use the `deleted` flag on `brigUser` to decide if the user is present or not.   I don't actually know what the `deleted` flag means (does it mean it is 'marked' for deletion; or do we never delete user data?) If it's the first; then I think this function is correct as is; if it is the second then erroring out might be a sane thing to do; though I still think it doesn't harm us to silently ignore this?





(grml; i miss transactions)

jschaul: I don't have enough context to answer all your questions, but I can answer this one:

>  I don't actually know what the deleted flag means (does it mean it is 'marked' for deletion; or do we never delete user data?)

Upon deletion of a user in brig, all user data is deleted (except the user's UUID in the `user` table), and a so-called `tombstone` is inserted: https://github.com/wireapp/wire-server/blob/develop/services/brig/src/Brig/API/User.hs#L788-L789 - i.e. the only thing remaining in the database is the UUID and a flag "deleted", to differentiate a now-deleted user from a never-existed user.

arianvp:
@tiago-loureiro could you please have another quick look? I have answered all remarks; left some dangling ( The one about  richInfo returning a `Maybe` when the user is not present; and the one about deleting users that I discussed with @fisx . they both seem related; and I just need more info from @fisx to understand what the correct thing to do is. but I think it is harmless to merge as is; shouldn't be a blocker in my opinion).

PR #927, opened 2019-12-19
The adding member is not always an admin

tiago-loureiro:
This _actually_ fixes https://github.com/wireapp/wire-server/pull/924 - the fact that tests did not break before is a sign that nothing got changed.

Background: previously it was assumed that the user _adding_ others to a conversation was a `wire_admin` - this is not always true: when adding to a conversation, a user adds itself. Also, it's conceivable that in the future there will be more roles able to add users to conversations so this PR also takes that into account making this more forwards compatible.

### DIFF
--- a/services/galley/src/Galley/API/Update.hs
+++ b/services/galley/src/Galley/API/Update.hs
@@ -330,7 +330,7 @@ joinConversation zusr zcon cnv access = do
     -- NOTE: When joining conversations, all users become members
     -- as this is our desired behavior for these types of conversations
     -- where there is no way to control who joins, etc.
-    addToConversation (botsAndUsers (Data.convMembers conv)) (zusr, roleNameWireMember) zcon (newUsers, roleNameWireMember) conv
+    addToConversation (botsAndUsers (Data.convMembers conv)) (zusr, roleNameWireMember) zcon ((, roleNameWireMember) <$> newUsers) conv
 
 addMembers :: UserId ::: ConnId ::: ConvId ::: JsonRequest Invite -> Galley Response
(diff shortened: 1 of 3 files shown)


services/galley/src/Galley/API/Update.hs
```diff
@@ -587,13 +587,13 @@ rmBot (zusr ::: zcon ::: req) = do
 -------------------------------------------------------------------------------
 -- Helpers
 
-addToConversation :: ([BotMember], [Member]) -> UserId -> ConnId -> [UserId] -> Data.Conversation -> RoleName -> Galley Response
-addToConversation _              _   _    [] _ _    = return $ empty & setStatus status204
-addToConversation (bots, others) usr conn xs c role = do
+addToConversation :: ([BotMember], [Member]) -> (UserId, RoleName) -> ConnId -> ([UserId], RoleName) -> Data.Conversation -> Galley Response
```
jschaul: Shouldn't this be:

```suggestion
addToConversation :: ([BotMember], [Member]) -> (UserId, RoleName) -> ConnId -> ([UserId, RoleName]) -> Data.Conversation -> Galley Response
```
? There's no requirement to add all users to a conversation with the same conversation role, or is there?

tiago-loureiro: Yes, this actually makes the code a lot cleaner https://github.com/wireapp/wire-server/pull/927/commits/1b4768d763f52092ec97981c2787f9a62828ce1d

To be precise, it allows me to fully remove the section:
```
    userRole u	
       | u == orig = origRole	
       | otherwise = othersRole
```
and have the users to be added to the conversation fully qualified, thanks 👍

PR #931, opened 2020-01-03
SCIM top level extra attrs

fisx:


fisx:
Hm, I thought I'd opened this as a draft PR...  Sorry about the noise.

arianvp:
@fisx Note that in order for this to be merged we either have to:

1. Implement PATCH for these extra attributes as well 
2. OR: Go back to the original behaviour I had in hscim (which you removed) that ignores attributes that aren't supported and implement PATCH for extra attributes later.

We have to choose one of these two options; otherwise customers will get their accounts quarantined by Azure because of too many errors.

I think option 1 is probably the right course of action.   Customers for now can not set the extra attributes through azure anyway; so we should do it well immediately;

arianvp:
I think we could introduce a typeclass to allow us to 'extend' the patch behaviour for the `extra` fields: https://github.com/wireapp/hscim/issues/36

libs/brig-types/src/Brig/Types/User.hs
```diff
@@ -274,25 +276,67 @@ instance ToJSON SelfProfile where
 ----------------------------------------------------------------------------
 -- Rich info
 
-data RichInfo = RichInfo
+data RichInfo
+  = RichInfoV0
     { richInfoFields :: ![RichField]  -- ^ An ordered list of fields
```
arianvp: Now `richInfoInlinedFields` is a partial function. Not sure if I like that

arianvp: How about we default the new field as empty;

akshaymankar:
Removed `[WIP]` tag to test in pipeline

akshaymankar:
I think this is ready now.

@fisx Github won't let me ask you for a review, but it would be great if you did.

libs/brig-types/src/Brig/Types/Test/Arbitrary.hs
```diff
@@ -313,13 +313,17 @@ instance Arbitrary UserProfile where
         <*> arbitrary
 
 instance Arbitrary RichField where
-    arbitrary = RichField <$> arbitrary <*> arbitrary
+    arbitrary = RichField <$> arbitrary <*> (arbitrary `suchThat` (/= ""))
```
fisx: mention `normalizeRichInfo` here, and where it is used, so we know what goes wrong if this restriction is missing here.

libs/brig-types/src/Brig/Types/Test/Arbitrary.hs
```diff
@@ -313,13 +313,17 @@ instance Arbitrary UserProfile where
...
 
 instance Arbitrary RichInfo where
     arbitrary = do
-        richInfoFields <- nubOn richFieldType <$> arbitrary
+        richInfoAssocList <- nubOn richFieldType <$> arbitrary
```
fisx: would be slightly nicer if this nubbing went into a general-purpose smart constructor (even though i don't think we'd use that anywhere else).

libs/brig-types/src/Brig/Types/User.hs
```diff
@@ -273,34 +279,98 @@ instance ToJSON SelfProfile where
...
+
+instance FromJSON RichInfoAssocList where
+  parseJSON v =
+    RichInfoAssocList <$> withObject "RichInfoAssocList" richInfoAssocListFromObject v
+
+instance ToJSON RichInfoAssocList where
+  toJSON (RichInfoAssocList l) = object [ "fields" .= l
+                                        , "version" .= (0 :: Int)
+                                        ]
+-- TODO: Make richFieldType @CI Text@
```
fisx: why not just do it?

libs/brig-types/src/Brig/Types/User.hs
```diff
@@ -601,8 +677,9 @@ instance FromJSON RichInfoUpdate where
     parseJSON = withObject "rich-info-update" $ \o ->
         RichInfoUpdate <$> o .: "rich_info"
 
+-- TODO(akshay): Maybe we need to support outputing assoc lists as well as hashmaps
```
fisx: why

libs/brig-types/src/Brig/Types/User.hs
```diff
@@ -273,34 +279,98 @@ instance ToJSON SelfProfile where
...
+                fields <- richInfoAssocListFromObject richinfoObj
+                pure fields
+              Array fields -> parseJSON (Array fields)
+              v -> Aeson.typeMismatch "Object" v
+          Just v -> Aeson.typeMismatch "Object" v
+
+      hmMapKeys :: (Eq k2, Hashable k2) => (k1 -> k2) -> HashMap k1 v -> HashMap k2 v
+      hmMapKeys f = HashMap.fromList . (map (\(k,v) -> (f k, v))) . HashMap.toList
+
+      mLookup :: (MonadFail m, Show k, Eq k, Hashable k) => k -> HashMap k v -> m v
```
fisx:
```suggestion
      hmLookup :: (MonadFail m, Show k, Eq k, Hashable k) => k -> HashMap k v -> m v
```

(for consistency)

services/brig/src/Brig/API.hs
```diff
@@ -1076,7 +1076,7 @@ getRichInfo (self ::: user ::: _) = do
         (Just t1, Just t2) | t1 == t2 -> pure ()
         _ -> throwStd insufficientTeamPermissions
     -- Query rich info
-    json . fromMaybe emptyRichInfo <$> lift (API.lookupRichInfo user)
+    json . fromMaybe (RichInfoAssocList []) <$> lift (API.lookupRichInfo user)
```
fisx: we really should split these functions up into a typed core and a wrapper that only does the json conversion.  yeay servant!  :)

services/spar/src/Spar/Scim/User.hs
```diff
@@ -235,13 +235,15 @@ validateScimUser' idp richInfoLimit user = do
...
-            { Scim.status = Scim.Status 413 }
+        let errorIfTooBig s name =
+              when (s > richInfoLimit) $ throwError $
+              (Scim.badRequest Scim.InvalidValue
+                (Just . cs $
+                  cs name <> " exceeds the limit: max " <> show richInfoLimit <>
+                  " characters, but got " <> show s))
+              { Scim.status = Scim.Status 413 }
+        errorIfTooBig (richInfoAssocListSize $ richInfoAssocList richInfo) richInfoAssocListURN
+        errorIfTooBig (richInfoMapSize richInfo) richInfoMapURN
```
fisx: I think this doubles the effective rich info limit.  Why don't you just add the two sizes?  But I don't think it's a problem, we put some pretty ad-hoc number here anyway.

libs/brig-types/src/Brig/Types/Test/Arbitrary.hs
```diff
@@ -35,11 +35,11 @@ import Data.Currency
 import Data.IP
 import Data.Json.Util (UTCTimeMillis (..), toUTCTimeMillis)
 import Data.LanguageCodes
-import Data.List.Extra (nubOn)
 import Data.Misc
 import Data.PEM (pemParseBS)
 import Data.Proxy
 import Data.Range
+import Data.List.Extra (nubOn)
```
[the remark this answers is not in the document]
akshaymankar: I think I removed it and the added it again and didn't care where it was. I am going to revert this. But we really need to agree upon a formatting tool so we can auto-format these things.

libs/brig-types/src/Brig/Types/User.hs
akshaymankar: Old comment that I forgot to remove :(

libs/brig-types/src/Brig/Types/User.hs
akshaymankar: By `mLookup` I meant monadic lookup, `hmLookup` is too close to `HM.lookup` which is lookup that returns a `Maybe`. I am going to rename it to `lookupOrFail`.

services/spar/src/Spar/Scim/User.hs
akshaymankar: I thought eventually people would be moving stuff from old rich info to new one, but we'll continue storing both. So I didn't want to break stuff just because now there is more data and the limit is same.

libs/brig-types/src/Brig/Types/User.hs
akshaymankar: I thought it would be too big, but it was not :)

libs/brig-types/src/Brig/Types/Test/Arbitrary.hs
akshaymankar: I thought we are going to delete the assoc lists in future, so it doesn't seem like it is worth the effort.

libs/brig-types/src/Brig/Types/User.hs
```diff
@@ -273,36 +279,100 @@ instance ToJSON SelfProfile where
...
+toRichInfoAssocList (RichInfo mp al) = RichInfoAssocList . nubrf $ toal mp <> al
+  where
+    nubrf :: [RichField] -> [RichField]
+    nubrf = nubBy ((==) `on` \(RichField k _) -> CI.mk k)
+
+    toal :: Map (CI Text) Text -> [RichField]
+    toal = map (\(k, v) -> RichField k v) . Map.toAscList
+
+richInfoMapURN, richInfoAssocListURN :: Text
+richInfoMapURN = "urn:ietf:params:scim:schemas:extension:wire:1.0:User"
```
arianvp: There was a TODO whether `2.0` was needed in the string. It is not. so this is fine :)

libs/brig-types/src/Brig/Types/User.hs
```diff
@@ -273,36 +279,100 @@ instance ToJSON SelfProfile where
...
+
+      lookupOrFail :: (MonadFail m, Show k, Eq k, Hashable k) => k -> HashMap k v -> m v
+      lookupOrFail key theMap = case HM.lookup key theMap of
+                                  Nothing -> fail $ "key '" ++ show key ++ "' not found"
+                                  Just v -> return v
+
+richInfoAssocListFromObject :: Object -> Aeson.Parser [RichField]
+richInfoAssocListFromObject richinfoObj = do
+  version :: Int <- richinfoObj .: "version"
+  when (version /= 0) $ fail $ "unknown version: " <> show version
```
arianvp: I'm still a bit confused what the rationale of the `version` field was. @fisx could you enlighten us? Do we need to bump it?

libs/brig-types/src/Brig/Types/User.hs
```diff
@@ -313,37 +383,46 @@ instance ToJSON RichField where
...
 --
+-- This works on @[RichField]@ so it can be reused by @RichInfoAssocList@
+--
 -- NB: we could just calculate the length of JSON-encoded payload, but it is fragile because
 -- if our JSON encoding changes, existing payloads might become unacceptable.
-richInfoSize :: RichInfo -> Int
-richInfoSize (RichInfo fields) =
-    sum [Text.length t + Text.length v | RichField t v <- fields]
+richInfoAssocListSize :: [RichField] -> Int
+richInfoAssocListSize fields = sum [Text.length (CI.original t) + Text.length v | RichField t v <- fields]
```
arianvp: Personally I prefer `map` (or in this case  `foldMap`) over list comprehensions. But it's not a hill I will die on

services/spar/test/Test/Spar/ScimSpec.hs
```diff
@@ -85,3 +102,124 @@ spec = describe "toScimStoredUser'" $ do
 
     Scim.meta result `shouldBe` meta
     Scim.value (Scim.thing result) `shouldBe` usr
+
+  it "roundtrips" . property $ do
+      \(sue :: ScimUserExtra) ->
+        eitherDecode' (encode sue) `shouldBe` Right sue
+
+  describe "ScimUserExtra" $ do
+    describe "Patchable" $ do
```
arianvp: Maybe a test that shows attributes are case-insensitive? Though I hope that test is already in the `CI` datatype in the library we depend on...

libs/brig-types/src/Brig/Types/User.hs
akshaymankar: I have no real preference in this case, I will merge it without any change for now. Maybe next time we touch this code, we can refactor :)

services/spar/test/Test/Spar/ScimSpec.hs
akshaymankar: Done :)

PR #941, opened 2020-01-13
Run hscim azure tests

fisx:
~~blocked by #958, which is fine.  this is not urgent.~~

This was a TODO in a fix from last week, but it's actually quite simple to actually do it.

Also, removed some trailing whitespace.

### DIFF
--- a/services/spar/test-integration/Spec.hs
+++ b/services/spar/test-integration/Spec.hs
@@ -18,6 +18,7 @@ import Data.Text (pack)
 import Imports
 import Servant.API (toHeader)
 import Spar.Run (mkApp)
+import Spar.Scim.Types
 import System.Environment (withArgs)
 import System.Random (randomRIO)
 import Test.Hspec
@@ -62,7 +63,7 @@ mkspec = do
 mkspec' :: TestEnv -> Spec
 mkspec' env = do
   describe "hscim acceptance tests" $
-    microsoftAzure AcceptanceConfig {..}
+    microsoftAzure @SparTag AcceptanceConfig {..}
   where
     scimAppAndConfig = do
       (app, _) <- mkApp (env ^. teOpts)
@@ -71,3 +72,4 @@ mkspec' env = do
           scimPathPrefix = "/scim/v2"
       pure (app, queryConfig)
     genUserName = pack <$> replicateM 9 (randomRIO ('a', 'z'))
+    responsesFullyKnown = False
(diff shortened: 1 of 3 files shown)


fisx:
interesting...  @arianvp perhaps we can take a look at these together?  looks like hscim doesn't like the way spar works.  but who is right and who is wrong?

arianvp:
All the endpoints return `404`. This smells like the path `hscim` expects is different from the one `spar` provides. perhaps the tests should take the base url as an argument?

fisx:
one last test still failing:

```
Failures:

  src/Web/Scim/Test/Acceptance.hs:77:7: 
  1) hscim acceptance tests User Operations
       status mismatch:
         expected: 201
         but got:  404

```

Don't we have to register the idp somehow?  What's the scim token when running the hscim acceptance tests on spar anyway?

fisx:
```
  src/Web/Scim/Test/Acceptance.hs:115:7: 
  1) hscim acceptance tests User Operations
       status mismatch:
         expected: 200
         but got:  400
         details:  SResponse {simpleStatus = Status {statusCode = 400, statusMessage = "Bad Request"}, simpleHeaders = [("Cache-Control","no-cache, no-store"),("Pragma","no-cache")], simpleBody = ""}
```

something wrong with parsing the request in this test case.  which is odd, because the parsing post and get request has worked at this point.

fisx:
```
Failures:

  test-integration/Test/Spar/APISpec.hs:712:17:                                      
  1) Spar.API, delete corner cases, re-create previously deleted, dangling users     
       "<?xml version=\"1.0\" encoding=\"UTF-8\"?><!DOCTYPE html PUBLIC \"-//W3C//DTD XHTML 1.1//EN\" \"http://www.w3.org/TR/xhtml11/DTD/xhtml11.dtd\"><html xml:lang=\"en\" xmlns=\"http://www.w3.org/1999/xhtml\"><head>  <title>wire:sso:error:bad-upstream</title>   <script type=\"text/javascript\">       const receiverOrigin = '*';       window.opener.postMessage({\"payload\":{\"errors\":[\"sso-login failed (race condition?)\"],\"label\":\"forbidden\"},\"type\":\"AUTH_ERROR\"}, receiverOrigin);   </script></head></html>" does not contain "<title>wire:sso:success</title>"                         

  To rerun use: --match "/Spar.API/delete corner cases/re-create previously deleted, dangling users/"
```

not sure this is a flake?  doesn't look like it has anything to do with what i'm doing here...  will investigate!

fisx:
it was a flake.

tiago-loureiro:
> it was a flake.

Caused by what?

fisx:
> > it was a flake.
> 
> Caused by what?

I don't know, but it's unrelated and i'll take care of it elsewhere.  https://bridge-ie-concourse.zinfra.io/builds/32113  https://github.com/zinfra/backend-issues/issues/1200

PR #944, opened 2020-01-15
Serve /config.json from (custom) backends to orient clients.

fisx:


mheinzel:
A different approach was taken, this is obsolete.

PR #946, opened 2020-01-20
Implement a 'domain -> custom backend' mapping.

fisx:
Fixes https://github.com/zinfra/backend-issues/issues/1098

### DIFF
--- a/libs/galley-types/src/Galley/Types.hs
+++ b/libs/galley-types/src/Galley/Types.hs
@@ -503,7 +503,6 @@ mkConversationCode k v (HttpsUrl prefix) = ConversationCode
 data CustomBackend = CustomBackend
     { backendConfigJsonUrl :: !HttpsUrl
     , backendWebappWelcomeUrl :: !HttpsUrl
-    , backendBlockCloudUsers :: !Bool
     } deriving (Eq, Show)
 
 newtype EmailDomain = EmailDomain
@@ -1144,7 +1143,6 @@ instance ToJSON CustomBackend where
     toJSON j = object
         $ "config_json_url"    .= backendConfigJsonUrl j
         # "webapp_welcome_url" .= backendWebappWelcomeUrl j
-        # "block_cloud_users"  .= backendBlockCloudUsers j
         # []
 
 instance FromJSON CustomBackend where
@@ -1152,4 +1150,3 @@ instance FromJSON CustomBackend where
         CustomBackend
             <$> o .: "config_json_url"
             <*> o .: "webapp_welcome_url"
-            <*> o .: "block_cloud_users"
--- a/libs/galley-types/src/Galley/Types.hs
+++ b/libs/galley-types/src/Galley/Types.hs
@@ -44,6 +44,9 @@ module Galley.Types
     , ConversationReceiptModeUpdate (..)
     , ConversationMessageTimerUpdate (..)
     , ConvType                  (..)
+    , CustomBackend             (..)
+    , EmailDomain               (emailDomainText)
+    , mkEmailDomain
     , Invite                    (..)
     , NewConv                   (..)
     , NewConvManaged            (..)
@@ -64,6 +67,7 @@ import Imports
 import Control.Lens ((.~))
 import Data.Aeson
 import Data.Aeson.Types (Parser)
+import Data.Bifunctor (bimap)
 import Data.ByteString.Conversion
 import Data.Misc
 import Data.Time
@@ -76,10 +80,12 @@ import Galley.Types.Conversations.Roles
 import Gundeck.Types.Push (Priority)
 import URI.ByteString
 
+import qualified Data.Attoparsec.ByteString as Atto
 import qualified Data.Code           as Code
 import qualified Data.HashMap.Strict as HashMap
 import qualified Data.Map.Strict     as Map
 import qualified Data.Text.Encoding  as T
+import qualified Text.Email.Validate as Email
 
 -- Conversations ------------------------------------------------------------
 
@@ -492,6 +498,31 @@ mkConversationCode k v (HttpsUrl prefix) = ConversationCode
   where
     q = [("key", toByteString' k), ("code", toByteString' v)]
     link = prefix & (queryL . queryPairsL) .~ q
+
+
+data CustomBackend = CustomBackend
+    { backendConfigJsonUrl :: !HttpsUrl
+    , backendWebappWelcomeUrl :: !HttpsUrl
+    , backendBlockCloudUsers :: !Bool
+    } deriving (Eq, Show)
+
+newtype EmailDomain = EmailDomain
+    { emailDomainText :: Text
+    } deriving (Eq, Generic, Show)
+
+mkEmailDomain :: ByteString -> Either String EmailDomain
+mkEmailDomain = bimap show EmailDomain . T.decodeUtf8' <=< validateDomain
+  where
+    -- this is a slightly hacky way of validating an email domain,
+    -- but Text.Email.Validate doesn't expose the parser for the domain.
+    validateDomain = fmap Email.domainPart . Email.validate . ("local-part@" <>)
+
+instance FromByteString EmailDomain where
+    parser = do
+        bs <- Atto.takeByteString
+        case mkEmailDomain bs of
+            Left err -> fail ("Failed parsing ByteString as EmailDomain: " <> err)
+            Right domain -> pure domain
 
 -- Instances ----------------------------------------------------------------
 
@@ -1108,3 +1139,17 @@ instance FromJSON ConversationCode where
         ConversationCode <$> o .: "key"
             <*> o .:  "code"
             <*> o .:? "uri"
+
+instance ToJSON CustomBackend where
+    toJSON j = object
+        $ "config_json_url"    .= backendConfigJsonUrl j
+        # "webapp_welcome_url" .= backendWebappWelcomeUrl j
+        # "block_cloud_users"  .= backendBlockCloudUsers j
+        # []
+
+instance FromJSON CustomBackend where
+    parseJSON = withObject "CustomBackend" $ \o ->
+        CustomBackend
+            <$> o .: "config_json_url"
+            <*> o .: "webapp_welcome_url"
+            <*> o .: "block_cloud_users"
--- a/services/galley/schema/src/V37.hs
+++ b/services/galley/schema/src/V37.hs
@@ -1,0 +1,17 @@
+module V37 (migration) where
+
+import Imports
+import Cassandra.Schema
+import Text.RawString.QQ
+
+migration :: Migration
+migration = Migration 37 "Create table `custom_backend`" $ do
+    schema' [r|
+        CREATE TABLE custom_backend (
+            domain             text,
+            config_json_url    blob,
+            webapp_welcome_url blob,
+            block_cloud_users  boolean,
+            PRIMARY KEY (domain)
+        );
+        |]
--- a/services/galley/schema/src/V37.hs
+++ b/services/galley/schema/src/V37.hs
@@ -11,7 +11,6 @@ migration = Migration 37 "Create table `custom_backend`" $ do
             domain             text,
             config_json_url    blob,
             webapp_welcome_url blob,
-            block_cloud_users  boolean,
             PRIMARY KEY (domain)
         );
         |]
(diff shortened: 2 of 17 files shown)


services/galley/schema/src/V37.hs
```diff
@@ -0,0 +1,19 @@
...
+
+migration :: Migration
+migration = Migration 37 "Create table `custom-instances`" $ do
+    schema' [r|
+        CREATE TABLE custom-instances (
+            domain             text,  -- email domain of the custom backend owner.
+            config_json_url    text,  -- url under which this custom backend's
+                                      -- config.json can be got.
+            webapp_welcome_url text,  -- needed for webapp due to CORS trouble
+            block_cloud_users  bool,  -- iff NULL or False, new cloud users
```
jschaul: Let's make this an integer / custom CQL type / custom Haskell type for futureproofing, and maybe rename to `cloud_users`. `cd services/galley && make db-reset` helps to re-create cassandra schemas locally.

libs/galley-types/src/Galley/Types.hs
```diff
@@ -1108,3 +1129,18 @@ instance FromJSON ConversationCode where
         ConversationCode <$> o .: "key"
             <*> o .:  "code"
             <*> o .:? "uri"
+
+-- TODO generic instance?
```
fisx: it's [here](https://hackage.haskell.org/package/aeson-1.4.6.0/docs/Data-Aeson-TH.html#v:deriveJSON), but @arianvp just convinced me we shouldn't do it: better write manual instances and round-trip tests, then we'll get more feedback from the type checker on what to keep in mind when we change the type.  also it already works.  (just need to add an `Arbitrary` instance [here](https://github.com/wireapp/wire-server/blob/1409d5882e6d1f12728e14fff52961d79ae95526/libs/brig-types/src/Brig/Types/Test/Arbitrary.hs) and a line for the test [here](https://github.com/wireapp/wire-server/blob/1409d5882e6d1f12728e14fff52961d79ae95526/libs/brig-types/test/unit/Test/Brig/Types/Common.hs#L59).)

([note that galley-types is tested in brig-types, for complicated reasons](https://github.com/wireapp/wire-server/blob/1409d5882e6d1f12728e14fff52961d79ae95526/libs/brig-types/test/unit/Test/Brig/Types/Common.hs#L7-L8))

services/galley/schema/src/V37.hs
```diff
@@ -0,0 +1,19 @@
...
+                                      -- with emails in this domain can be created.
+            PRIMARY KEY (domain)
+        );
+        |]
```
fisx: @jschaul any opinions on types, names, anything?  should we add any tuning details?

services/galley/src/Galley/API.hs
```diff
@@ -925,6 +926,17 @@ sitemap = do
         returns (ref Model.ssoTeamConfig)
         response 200 "SSO status" end
 
+    get "/custom-instance/by-domain/:domain" (continue CustomInstance.getCustomInstanceByDomainUntyped) $
+        capture "domain"
+        .&. accept "application" "json"
+
+    document "GET" "getCustomInstanceByDomain" $ do
+        summary "TODO"
+        parameter Path "domain" bytes' $
```
fisx:
```suggestion
        parameter Path "domain" text' $
```

services/galley/schema/src/V37.hs
fisx: Thanks, will do!

(I'm not sure we'll need it in this case, and there is this is also sent as a boolean to the clients, so it'll be hard to change anyway.  But yes, custom booleans are never a bad thing.)

@mheinzel what do you think?  here is some pseudo-code:

```haskell
data CustomInstanceAllowCloudUser = CloudUserAllowed | CloudUserBlocked
  deriving (Eq, Ord, Bounded, Enum, Show, Generic)

instance Cql CustomInstanceAllowCloudUser where
    ctype = Tagged IntColumn

    fromCql (CqlInt i) = case i of
        0 -> return CloudUserAllowed
        1 -> return CloudUserBlocked
        n -> fail $ "unexpected relation: " ++ show n
    fromCql _ = fail "custom instance allow cloud user: int expected"

    toCql CloudUserAllowed = CqlInt 0
    toCql CloudUserBlocked = CqlInt 1
```

...  and then we will certainly need to keep the custom json instances...

services/galley/schema/src/V37.hs
fisx: https://github.com/wireapp/wire-server/pull/946#discussion_r369003743

this means that the `boolean` here should be an `int`.

stack-deps.nix
```diff
@@ -28,7 +28,7 @@ let
         rev = "4067ad96b125942545dbdec8c1a89f1e1b65d013";
         sha256 = "1i9dlhw0xk1viglyhail9fb36v1awrypps8jmhrkz8k1bhx98ci3";
       };
-      cargoSha256 = "1373rpy0fi3cpacv06x1cv4cv0brwdri2680ymdkq8w44syp20ym";
+      cargoSha256 = "0m85c49hvvxxv7jdipfcaydy4n8iw4h6myzv63v7qc0fxnp1vfm8";
```
fisx: should this land from https://github.com/wireapp/wire-server/pull/949?  (no strong opinion)

services/galley/src/Galley/API.hs
```diff
@@ -1059,6 +1071,10 @@ sitemap = do
         .&. opt zauthConnId
         .&. jsonRequest @RemoveBot
 
+    put "/i/custom-instance/by-domain/:domain" (continue CustomInstance.internalPutCustomInstanceByDomain) $
+        capture "domain"
+        .&. jsonRequest @CustomInstance
+
```
fisx: `delete` is missing.

services/galley/src/Galley/Data/CustomInstance.hs
```diff
@@ -0,0 +1,23 @@
...
+
+import Imports
+import Cassandra
+import Galley.Data.Instances ()
+import Galley.Types
+
+import qualified Galley.Data.Queries  as Cql
+
+getCustomInstance :: MonadClient m => EmailDomain -> m (Maybe CustomInstance)
+getCustomInstance domain = fmap toCustomInstance <$> do
```
fisx: I guess you could avoid the `toCustomInstance` declaration altogether:

```suggestion
getCustomInstance domain = fmap CustomInstance <$> do
```

(no strong opinion.)

services/galley/src/Galley/Data/CustomInstance.hs
```diff
@@ -0,0 +1,23 @@
...
+    retry x1 $ query1 Cql.selectCustomInstance (params Quorum (Identity domain))
+  where
+    toCustomInstance (customInstanceConfigJsonUrl, customInstanceWebappWelcomeUrl, customInstanceBlockCloudUsers) =
+        CustomInstance{..}
+
+setCustomInstance :: MonadClient m => EmailDomain -> CustomInstance -> m ()
+setCustomInstance domain CustomInstance{..} = do
+    retry x5 $ write Cql.updateCustomInstance (params Quorum (customInstanceConfigJsonUrl, customInstanceWebappWelcomeUrl, customInstanceBlockCloudUsers, domain))
```
fisx: Why is `domain` the last argument here?  In the schema it looks like it should be the first?

(ah, i see, it just has to be consistent with `Cql.updateCustomInstance`.  still, would be nice to have it consistent all over.  anyway, never mind.)

services/galley/src/Galley/Data/Instances.hs
```diff
@@ -121,3 +123,9 @@ instance Cql SSOStatus where
 
     toCql SSODisabled = CqlInt 0
     toCql SSOEnabled = CqlInt 1
+
+instance Cql EmailDomain where
+    ctype = Tagged TextColumn
+    toCql = CqlText . emailDomainText
+    fromCql (CqlText txt) = either fail pure (mkEmailDomain (encodeUtf8 txt))
```
fisx:
```suggestion
    fromCql (CqlText txt) = either fail pure . mkEmailDomain $ encodeUtf8 txt
```

(I have strong opinions on this, but they are inconsistent and irrelevant.  :))

services/galley/schema/src/V37.hs
mheinzel: If we want to do future-proofing this way, it would seem only consequential to me also expose it to the clients that way. If we have a JSON instance along the lines of
```
CloudUserAllowed -> "false"
CloudUserBlocked -> "true"
```
and then need to add `CloudUserAllowedOnTuesdays`, what do we send the clients? Will the possible values become `true, false, "on_tuesdays"`?
I'd rather go with `"allowed", "blocked"`, so it's natural to add `"on_tuesdays"`.

mheinzel: So, is it conceivable that there will be a third value in the future (not sure what that would be)? If so, I'd suggest to also reflect that in the JSON schema and let the client teams know about it.

libs/galley-types/src/Galley/Types.hs
mheinzel: I agree with sticking to manual instances for now. Otherwise it's too easy to break the JSON structure without noticing.

services/galley/schema/src/V37.hs
fisx: @jschaul that makes me think that we should stick with the boolean.  i don't see how the question whether users should be blocked can have any other answer but "yes" or "no", and anything we want to do beyond that would just go to a different attribute in json and a different column in cassandra?

generally i think this trick of always using ints for booleans in case a third value pops up works much better if we don't expose the same structure to clients.

stack-deps.nix
mheinzel: I'll wait for #949 to be merged, then we don't need this here anymore.

services/galley/src/Galley/Data/CustomInstance.hs
mheinzel: It's not that easy, since we get a tuple and need to turn it into a record. This could be done generically with a fancy typeclass, but I think there are better solutions for the underlying problem. Happy to discuss this.

mheinzel:
> Fixes [zinfra/backend-issues#1100](https://github.com/zinfra/backend-issues/issues/1100)

I think this is actually [zinfra/backend-issues#1098](https://github.com/zinfra/backend-issues/issues/1098)

services/galley/src/Galley/API.hs
mheinzel: I actually reverted this suggestion as `text'` doesn't exist.

services/galley/src/Galley/API.hs
mheinzel: fixed

services/galley/src/Galley/API.hs
fisx: sorry, i meant this: https://hackage.haskell.org/package/swagger-0.3.0/docs/Data-Swagger-Build-Api.html#v:string-39-

services/galley/schema/src/V37.hs
```diff
@@ -0,0 +1,16 @@
...
+import Imports
+import Cassandra.Schema
+import Text.RawString.QQ
+
+migration :: Migration
+migration = Migration 37 "Create table `custom-instances`" $ do
+    schema' [r|
+        CREATE TABLE custom_instances (
+            domain             text,
+            config_json_url    blob,
```
fisx: why blob, not text?  text is arbitrarily long and utf8, which should be enough for urls.  blob is less restrictive, so reading it back may go wronger.

i have to admit:

```
$ git grep -Hn ur[li] | grep schema/src
services/brig/schema/src/V35.hs:15:            , url      blob
services/brig/schema/src/V35.hs:43:            , base_url     blob
services/galley/schema/src/V20.hs:62:            base_url     blob,
services/galley/schema/src/V31.hs:12:            base_url     blob,
[...]
services/spar/schema/src/V0.hs:41:            , request_uri   text
```

the only other url column that has *not* blob type is one that i wrote.  so i guess the question is: why am i wrong?

libs/galley-types/src/Galley/Types.hs
```diff
@@ -493,6 +499,30 @@ mkConversationCode k v (HttpsUrl prefix) = ConversationCode
     q = [("key", toByteString' k), ("code", toByteString' v)]
     link = prefix & (queryL . queryPairsL) .~ q
 
+
+data CustomBackend = CustomBackend
+    { instanceConfigJsonUrl :: !HttpsUrl
+    , instanceWebappWelcomeUrl :: !HttpsUrl
```
jschaul: rename `instance` -> `backend` ?

mheinzel: good catch, fixed!

PR #948, opened 2020-01-21
Fix more SCIM bugs

fisx:
not really bugs, but rather good examples of malevolent code obfuscation.  please read commit-by-commit!

### DIFF
--- a/services/spar/src/Spar/Scim/User.hs
+++ b/services/spar/src/Spar/Scim/User.hs
@@ -371,11 +371,21 @@ updateValidScimUser tokinfo uid newScimUser = do
             -- this can only happen if user is found in spar.scim_user, but missing on brig.
             -- (internal error?  race condition?)
 
-        -- TODO: rich info and/or user handle may not have changed.  in that case don't write
-        -- it.
-        maybe (pure ()) (lift . Intra.Brig.setBrigUserName uid) $ newScimUser ^. vsuName
-        lift . Intra.Brig.setBrigUserHandle uid $ newScimUser ^. vsuHandle
-        lift . Intra.Brig.setBrigUserRichInfo uid $ newScimUser ^. vsuRichInfo
+        oldScimUser :: ValidScimUser
+          <- validateScimUser tokinfo . Scim.value . Scim.thing $ oldScimStoredUser
+              -- the old scim user from our db is already validated, but this also recovers
+              -- the extra details not stored in the DB that we need here.
+
+        lift $ do
+            case newScimUser ^. vsuName of
+                Just nm | oldScimUser ^. vsuName /= Just nm -> Intra.Brig.setBrigUserName uid nm
+                _ -> pure ()
+
+            when (oldScimUser ^. vsuHandle /= newScimUser ^. vsuHandle) $
+                Intra.Brig.setBrigUserHandle uid $ newScimUser ^. vsuHandle
+
+            when (oldScimUser ^. vsuRichInfo /= newScimUser ^. vsuRichInfo) $
+                Intra.Brig.setBrigUserRichInfo uid $ newScimUser ^. vsuRichInfo
 
         -- store new user value to scim_user table (spar). (this must happen last, so in case
         -- of crash the client can repeat the operation and it won't be considered a noop.)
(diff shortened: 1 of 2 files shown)


services/spar/src/Spar/Scim/User.hs
```diff
@@ -222,7 +219,9 @@ validateScimUser'
   -> m ValidScimUser
 validateScimUser' idp richInfoLimit user = do
     uref :: SAML.UserRef <- mkUserRef idp (Scim.externalId user)
-    handl <- validateHandle (Scim.userName user)
+    handl <- validateHandle . Text.toLower . Scim.userName $ user
+        -- FUTUREWORK: 'Scim.userName' should be case insensitive; then the toLower here would
```
arianvp: I'm not sure exactly what this comment means. why is `toLower` brittle?

services/spar/src/Spar/Scim/User.hs
```diff
@@ -300,7 +300,6 @@ createValidScimUser (ValidScimUser user uref handl mbName richInfo) = do
     -- FUTUREWORK(arianvp): Get rid of manual lifting. Needs to be SCIM instances for ExceptT
     -- This is the pain and the price you pay for the horribleness called MTL
     storedUser <- lift $ toScimStoredUser buid user
-    idpConfig <-  lift $ SAML.getIdPConfigByIssuer (uref ^. SAML.uidTenant)
```
arianvp: I'm wondering if this changes behaviour? Are we sure this is the same `idpConfig` ? can there be cases where these are distinct? why not?

services/spar/src/Spar/Scim/User.hs
```diff
@@ -324,7 +323,7 @@ createValidScimUser (ValidScimUser user uref handl mbName richInfo) = do
 updateValidScimUser
   :: forall m. (m ~ Scim.ScimHandler Spar)
   => ScimTokenInfo -> UserId -> ValidScimUser -> m (Scim.StoredUser SparTag)
-updateValidScimUser tokinfo@ScimTokenInfo{stiIdP} uid newScimUser = do
+updateValidScimUser tokinfo uid newScimUser = do
```
arianvp: I can't leave comments in places that are not edited by you because Github sucks. Open this file and scroll to line 341:


Does `Scim.getUser` need all of `tokinfo` to work or just a part of `tokinfo`?  Reason I ask is because it's nice if this function only takes what it needs, instead of a `tokinfo` which also contains an `idp` which might lead us to confuse it with the idp in `newScimUser` or `oldScimUser`

arianvp:
This seems to have broken a (unrelated?) test on concourse. Is this flakey or some regression?

```
      team-only:                                                        FAIL (0.68s)

        test/integration/API/Provider.hs:1662:

        event type

        expected: ConvAccessUpdate

         but got: MemberLeave

```

services/spar/src/Spar/Scim/User.hs
fisx: we need to make sure it's always called, so it would be easier if there was no need to call it at all.  that would result in code that is less rigid and easy to break.

does this explain it better?  can you suggest a comment that makes this clearer, but is not too long?

services/spar/src/Spar/Scim/User.hs
fisx: I think except for yet another race condition (during the handling of this http request, the idp changed or went away, so the first lookup found something different from the second lookup) I don't think this is an issue.  I'll double-check tomorrow.

services/spar/src/Spar/Scim/User.hs
fisx: It's used as an argument for a function from hscim, so I can't think of a way to take the redundant idp-id out of the argument.  You make a good point, though.  If you can think of anything I'd like to hear it.

services/spar/src/Spar/Scim/User.hs
fisx: b4a4dfca61545cee2ed983e5b02721d12ac352bd  good?

fisx:
b4a4dfc good?

fisx:
(ja, appearently i-tests are flaky.)

PR #949, opened 2020-01-21
Update `cargoSha256` for cryptobox-c in stack-deps.nix

akshaymankar:


arianvp:
@fisx all very valid questions! This is because the function we're using in Nix is a "fixed-output derivation" this means   "arbitrary I/O is allowed as long as you promise to be reproducible. The promise is kept by checking that the end result always has the same hash". This means that cargo gets internet access and can download packages itself; bypassing all of nix's sandboxing and pinning mechanism. This hack works if cargo is indeed reproducible, but it apparently is not!

We should use a less-problematic way of building cargo dependencies in our nix derivation.

We should move to https://github.com/nix-community/carnix to get more fine-grained and reproducible pinning. (Pins each cargo package individually; and lets nix download it instead of giving cargo network access). However that is something for a different issue.

PR #950, opened 2020-01-23
Remove verifcation code from email subject line.

fisx:


fisx:
https://github.com/wireapp/wire-emails/pull/222

fisx:
(I think this *could* be merged, but I just broke wire-emails on `1.0.65`, and would like to wait for `1.0.66` to get in the changes this was originally about.  i will re-request review when i'm done.)

fisx:
re . 39a993a: there were some language files not supported by [wire-emails](https://github.com/wireapp/wire-emails/blob/master/bin/translate.js#L24-L31).  i just removed them, and [it seems there is always a default](https://github.com/wireapp/wire-server/blob/f4a535420aa8ab12a0ebf7d581b5ced0c056b318/services/brig/src/Brig/Template.hs#L45)

fisx:
(please review commit-by-commit, but note the first is reverted in the second.)

jschaul:
> some language files not supported by wire-emails

Hm, how come the set of languages has decreased with time?

fisx:
> > some language files not supported by wire-emails
> 
> Hm, how come the set of languages has decreased with time?

great question.  i don't know why, but [this is where](https://github.com/wireapp/wire-emails/commit/41f5e57421bec84e610adfd6e4d505bf72cd081f).

I wonder if we should build something that keeps track of the list of supported languages?

PR #952, opened 2020-01-27
A quick hack for printing cassandra schemas on dev machines.

fisx:
missing: integration test that fails if this gets out of sync.

I'm not particularly convinced of the approach, but *something* like this would be nice: https://github.com/zinfra/backend-issues/issues/1086


fisx:
also missing: the schema goes to the txt file twice, no idea why.

fisx:
> > the schema goes to the txt file twice, no idea why.
> 
> Fixed in the small commit I took the liberty to add to your branch.

cool, thanks!

(why does the diff look so odd, though?  did you make sure cassandra was fully migrated to the latest versions?)

jschaul:
> (why does the diff look so odd, though? 

don't know. I suppose the order of the tables is random?

> did you make sure cassandra was fully migrated to the latest versions?)

yes! At least I ran all the db-reset stuff on this branch.

PR #953, opened 2020-01-29
Allow up to 256 characters as handle, dots and dashes too

tiago-loureiro:
This fits more globally into making users unique and searchable in an enterprise world. 64 characters should be enough for most use cases; I did _not_ run any integration tests, so I will let the CI do that job.

Adding extra `-` and `.` was done so that it supports, more directly, conversions of emails to handles. `@` and `+` should _NOT_ be used so that "phone numbers", "emails", and "handles" remain disjoint sets.

### DIFF
--- a/libs/brig-types/src/Brig/Types/Common.hs
+++ b/libs/brig-types/src/Brig/Types/Common.hs
@@ -47,14 +47,17 @@ isValidHandle :: Text -> Bool
 isValidHandle t = either (const False) (const True)
                 $ parseOnly handle t
   where
-    handle = count 2 (satisfy charsFirst)
-          *> count 62 (optional (satisfy charsSecond))
+    handle = count 2 (satisfy chars)
+          *> count 254 (optional (satisfy chars))
           *> endOfInput
     -- NOTE: Ensure that characters such as `@` and `+` should _NOT_
     -- be used so that "phone numbers", "emails", and "handles" remain
     -- disjoint sets.
-    charsFirst = inClass "a-z0-9_"
-    charsSecond = inClass "a-z0-9_.-"
+    -- The rationale behind max size here relates to the max length of
+    -- an email address as defined here:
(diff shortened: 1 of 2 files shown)


libs/brig-types/src/Brig/Types/Common.hs
```diff
@@ -47,10 +47,14 @@ isValidHandle :: Text -> Bool
...
-          *> count 19 (optional (satisfy chars))
+    handle = count 2 (satisfy charsFirst)
+          *> count 62 (optional (satisfy charsSecond))
           *> endOfInput
-    chars  = inClass "a-z0-9_"
+    -- NOTE: Ensure that characters such as `@` and `+` should _NOT_
+    -- be used so that "phone numbers", "emails", and "handles" remain
+    -- disjoint sets.
+    charsFirst = inClass "a-z0-9_"
+    charsSecond = inClass "a-z0-9_-."
```
jschaul: What's the rationale to allow `-` and `.`, but not as a first and a second character? "John Smith" could now be `jo.smith` but not `j.smith` with this change.

What's the rationale for exactly 64 characters, is it arbitrary? Please add a comment what it relates to (even to say it's arbitrary - so future readers of the code don't try to interpret non-existent meaning).

tiago-loureiro: > What's the rationale to allow - and ., but not as a first and a second character? "John Smith" could now be jo.smith but not j.smith with this change.

Previous self thought this made sense... I'd like to avoid weird usernames such as `..` but we already allow `__` so we might as well simplify the parser so I reverted that change.

> What's the rationale for exactly 64 characters, is it arbitrary? Please add a comment what it relates to (even to say it's arbitrary - so future readers of the code don't try to interpret non-existent meaning).

I changed my mind slightly and added a comment at https://github.com/wireapp/wire-server/pull/953/commits/92d37cfddf67d85025826351001a21700188e94f

Thanks for the feedback 🙇

PR #957, opened 2020-02-03
Whitespace

fisx:
this should be the rest of it...


fisx:
(at least in `*.hs`; i did it with `find` and `perl`.)

PR #958, opened 2020-02-04
[WIP] Upgrade to GHC 8.6.5 (LTS 14.12)

akshaymankar:
This is marked WIP until we merge the CI pull request: https://github.com/wireapp/wire-server-private/pull/161.

After we merge the PR in wire-server-private, we should merge this PR. After that we'll have to make sure we trigger these jobs before we can trigger more PR and/or staging jobs:
- [ ] docker-images/alpine-prebuilder
- [ ] docker-images/alpine-builder
- [ ] docker-images/alpine-deps
- [ ] docker-images/ubuntu14-ghc-8
- [ ] docker-images/wire-server-deps-ubuntu
- [ ] docker-images/wire-server-cache-ubuntu

These pipelines are order specific, we need to make sure we figure that out and trigger them in the order which ensure they are built properly.

### DIFF
--- a/libs/types-common-journal/package.yaml
+++ b/libs/types-common-journal/package.yaml
@@ -14,6 +14,7 @@ extra-source-files:
 - proto/UserEvents.proto
 ghc-options:
 - -fno-warn-redundant-constraints
+- -ddump-splices
 dependencies:
 - base ==4.*
 - bytestring
@@ -33,7 +34,8 @@ library:
   - Proto.TeamEvents_Fields
   - Proto.UserEvents
   - Proto.UserEvents_Fields
-# build-type: Custom
-# verbatim: |
 custom-setup:
-  dependencies: [base, Cabal, proto-lens-setup]
+  dependencies:
+  - base
+  - Cabal
+  - proto-lens-setup
--- a/libs/types-common-journal/package.yaml
+++ b/libs/types-common-journal/package.yaml
@@ -14,7 +14,6 @@ extra-source-files:
 - proto/UserEvents.proto
 ghc-options:
 - -fno-warn-redundant-constraints
-- -ddump-splices
 dependencies:
 - base ==4.*
 - bytestring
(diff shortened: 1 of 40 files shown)


snapshots/wire-2.0.yaml
```diff
@@ -0,0 +1,101 @@
...
+
+# Dropped from upstream snapshot
+- bloodhound-0.16.0.0
+- template-0.2.0.10
+- HaskellNet-0.5.1
+- HaskellNet-SSL-0.3.4.1
+- snappy-0.2.0.2
+- smtp-mail-0.2.0.0
+- stm-containers-1.1.0.4
+- wai-route-0.4.0 # Not latest as latst one breaks wai-routing
```
fisx: This one should have gone to its own section here, no?

services/galley/src/Galley/Data/Types.hs
```diff
@@ -113,5 +113,5 @@ generate cnv s t = do
 
 mkKey :: MonadIO m => ConvId -> m Key
 mkKey cnv = do
-    Just sha256 <- liftIO $ getDigestByName "SHA256"
+    sha256 <- liftIO $ fromJust <$> getDigestByName "SHA256"
```
fisx: This change makes error messages a lot worse, at least historically.  What's the motivation to avoid the incomplete pattern instead?  Does that have something to do with ghc upgrade and `MonadFail`?

libs/types-common-journal/package.yaml
```diff
@@ -14,11 +14,12 @@ extra-source-files:
 - proto/UserEvents.proto
 ghc-options:
 - -fno-warn-redundant-constraints
+- -ddump-splices
```
fisx:
```suggestion

```

services/galley/src/Galley/Data/Types.hs
akshaymankar: Yes, adding `MonadFail` here was a more _hairy_ change. But since this should never error, we don't have to worry about bad error messsages.

snapshots/wire-2.0.yaml
akshaymankar: Done

PR #940, opened 2020-01-13
Remove collectd metrics (finally!)

jschaul:
Removes:

* `/i/monitoring` endpoint from all services
* `metrics-collectd` library (was unused anyway)
* json/collectd output from `metrics-core`
* collectd from bonanza
* the `measured` middleware on all services (this is now done using prometheus). Most grafana dashboards should be updated by now.

Rewrites the logic of one integration test.

PR #959, opened 2020-02-04
Release 2020 01 30

tiago-loureiro:
# 2020-01-08

## API changes (relevant client developers)

- Allow up to 256 characters as handle, dots and dashes too (#953)
  - All handles related endpoints, namely:
    - POST "/users/handles"
    - HEAD "/users/handles/:handle"
    - GET "/users/handles/:handle"
  - now accept this new format of handles
- Refuse to delete non-empty IdPs (412 precondition failed) (#875)
  - DELETE "identity-providers/:idp" will now return 412 if there are users provisioned with that IDP
- Linear onboarding feature: Provide information about custom backends (#946)
  - New public endpoint:
    - GET "/custom-backend/by-domain/:domain"
  - New interal endpoints:
    - PUT "/i/custom-backend/by-domain/:domain"
    - DELETE "/i/custom-backend/by-domain/:domain"

## Bug fixes

- Make sure that someone is SSO user before setting ManagedBy (#947)
- Misc SCIM bugfixes (#948)

## Internal changes

- Fix complexity issue in cassandra query. (#942)
- Remove collectd metrics (finally!) (#940)
- Update `cargoSha256` for cryptobox-c in stack-deps.nix (#949)


CHANGELOG.md
```diff
@@ -1,5 +1,24 @@
 # 2020-01-08
 
+## API changes (relevant for self-hosters and client developers)
+
+- Allow up to 256 characters as handle, dots and dashes too (#953)
+- Refuse to delete non-empty IdPs (412 precondition failed) (#875)
+- Make sure that someone is SSO user before setting ManagedBy (#947)
+- Linear onboarding feature: Provide information about custom backends (#946)
```
jschaul: Listing the new API endpoints in the changelog would be nice.

CHANGELOG.md
```diff
@@ -1,5 +1,24 @@
 # 2020-01-08
 
+## API changes (relevant for self-hosters and client developers)
```
jschaul: How about splitting this section? As far as I can see, there is no change relevant for self-hosters, as no new configuration is introduced.

CHANGELOG.md
```diff
@@ -1,5 +1,24 @@
...
+
+- Allow up to 256 characters as handle, dots and dashes too (#953)
+- Refuse to delete non-empty IdPs (412 precondition failed) (#875)
+- Make sure that someone is SSO user before setting ManagedBy (#947)
```
jschaul: these two look like bug fixes / internal changes to me.

tiago-loureiro: I guess we can debate this, my thinking here was: executing the same request against our API now returns a different result. This was not something that was there by design, we _changed_ the design. WDYT?

CHANGELOG.md
tiago-loureiro: I still keep 2 sections around but I find the release log confusing, happy to rethink it. Keeping a section "relevant for self hosters" only that depends on config changes sounds good.

CHANGELOG.md
tiago-loureiro: Good idea 👍

tiago-loureiro:
In general I'd like to clarify what those sections mean because I find them a bit confusing. API changes are also relevant for self hosters IMHO, because they may change the behavior of clients.

jschaul:
> In general I'd like to clarify what those sections mean because I find them a bit confusing. API changes are also relevant for self hosters IMHO, because they may change the behavior of clients.

Sure, we should define this a bit better. I read this as "for people installing/upgrading a wire-server installation, take note of this change, as you may wish to adapt your installation/upgrade scripts and configuration". Nothing prevents an operator of an installation to also read the section on API changes :smile:

CHANGELOG.md
jschaul: Sure, leave them there, but maybe write which endpoints now return different results. `/endpoint1 now behaves differently (#PRnumber)` is more informational for client developers looking at the changelog I think.

tiago-loureiro:
I adjusted the CHANGELOG. Some things may be overkill on some release notes (e.g., listing affected endpoints might be too noisy if, e.g., `/users` endpoint is affected)

CHANGELOG.md
```diff
@@ -1,5 +1,35 @@
...
+    - GET "/users/handles/:handle"
+  - now accept this new format of handles
+- Refuse to delete non-empty IdPs (412 precondition failed) (#875)
+  - DELETE "identity-providers/:idp" will now return 412 if there are users provisioned with that IDP
+- Linear onboarding feature: Provide information about custom backends (#946)
+  - New public endpoint:
+    - GET "/custom-backend/by-domain/:domain"
+  - New interal endpoints:
+    - PUT "/i/custom-backend/by-domain/:domain"
+    - DELETE "/i/custom-backend/by-domain/:domain"
```
jschaul:
```suggestion
```

There is only a PUT endpoint, no delete endpoint. See https://github.com/wireapp/wire-server/pull/960/files#diff-1682ecde5b570fb15dfaa9ccbeb59a5cR75

tiago-loureiro: Code says otherwise :)

develop:

https://github.com/wireapp/wire-server/blob/develop/services/galley/src/Galley/API.hs#L1078

this PR:
https://github.com/wireapp/wire-server/blob/1795e7d6ea15c41f136b0fa24daaddbe33d39603/services/galley/src/Galley/API.hs#L1078

fisx: https://github.com/wireapp/wire-server/pull/960/files#diff-1682ecde5b570fb15dfaa9ccbeb59a5cR75 is about config.json, https://github.com/wireapp/wire-server/blob/develop/services/galley/src/Galley/API.hs#L1078 is about the domain-based redirect *to* that config.json.

so you're both right.  :)

PR #961, opened 2020-02-05
Add missing options to services-demo config files

mheinzel:
With the current config files, brig and galley refused to start with errors like

```
brig: user error (AesonException "Error in $.zauth.authSettings: key \"legalHoldAccessTokenTimeout\" not present")
```

Not sure if the given values are the right ones for the demo, though. Is there something we should change?


deploy/services-demo/conf/brig.demo.yaml
```diff
@@ -86,6 +86,7 @@ zauth:
     providerTokenTimeout: 604800 # 7 days
     legalHoldUserTokenTimeout: 4838400    # 56 days
     legalHoldSessionTokenTimeout: 604800  # 7 days
+    legalHoldAccessTokenTimeout: 30
```
jschaul:
```suggestion
    legalHoldAccessTokenTimeout: 900 # 15 minutes
```

PR #962, opened 2020-02-05
SRV resolution

jschaul:
* a new shared library for federation-related functions
* implementation of SRV lookup with ordering.

Real integration testing requires creating real DNS records (unless there is a way to mock these?), so I have not included those.

### DIFF
--- a/libs/federation-util/src/Network/Federation/Util/DNS.hs
+++ b/libs/federation-util/src/Network/Federation/Util/DNS.hs
@@ -7,7 +7,8 @@ import Network.DNS
 import Network.Federation.Util.Internal
 
 -- | Looks up a SRV record given a domain, returning A(AAA) records with their
--- ports (ordered by priority and weight according to RFC 2782).
+-- ports (ordered by priority and weight according to RFC 2782). Connection
+-- attempts should be made to the returned result list in order.
 --
 -- Example:
 --
--- a/libs/federation-util/src/Network/Federation/Util/DNS.hs
+++ b/libs/federation-util/src/Network/Federation/Util/DNS.hs
@@ -22,4 +22,4 @@ srvLookup :: Text -> ResolvSeed -> IO (Maybe [(Domain, Word16)])
 srvLookup = srvLookup' srvDefaultPrefix
 
 srvDefaultPrefix :: Text
-srvDefaultPrefix = "_wire" -- TODO: _wire-server? _xmpps?
+srvDefaultPrefix = "_wire-server"
--- a/libs/federation-util/src/Network/Federation/Util/DNS.hs
+++ b/libs/federation-util/src/Network/Federation/Util/DNS.hs
@@ -1,0 +1,25 @@
+module Network.Federation.Util.DNS
+    ( srvLookup
+    ) where
+
+import Imports
+import Network.DNS
+import Network.Federation.Util.Internal
+
+-- | Looks up a SRV record given a domain, returning A(AAA) records with their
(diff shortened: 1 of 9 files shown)


libs/federation-util/src/Network/Federation/Util/Internal.hs
```diff
@@ -0,0 +1,68 @@
...
+-- Given a prefix (e.g. _wire) and a domain (e.g. wire.com),
+-- provides a list of A(AAA) names and port numbers upon a successful
+-- DNS-SRV request, or `Nothing' if the DNS-SRV request failed.
+-- Modified version inspired from http://hackage.haskell.org/package/pontarius-xmpp
+srvLookup' :: Text -> Text -> ResolvSeed -> IO (Maybe [(Domain, Word16)])
+srvLookup' prefix realm resolvSeed = withResolver resolvSeed $ \resolver -> do
+    srvResult <- lookupSRV resolver $ encodeUtf8 $ prefix <> "._tcp." <> realm <> "."
+    case srvResult of
+        Right [] -> return Nothing
+        Right [(_, _, _, ".")] -> return Nothing -- explicit "not available"
```
tiago-loureiro: > Right [(_, _, _, ".")] -> return Nothing -- explicit "not available"

This is intended to match SRV records such as:

```
0 10 443 .
```

right? Can we add a test for that? Also, what happens if someone sets sth like:
```
0 10 443 .
0 10 443 bla.com.
```

libs/federation-util/src/Network/Federation/Util/DNS.hs
```diff
@@ -0,0 +1,25 @@
...
+-- >
+-- > main :: IO ()
+-- > main = do
+-- >   rs <- makeResolvSeed defaultResolvConf
+-- >   x <- srvLookup "staging.zinfra.io" rs
+srvLookup :: Text -> ResolvSeed -> IO (Maybe [(Domain, Word16)])
+srvLookup = srvLookup' srvDefaultPrefix
+
+srvDefaultPrefix :: Text
+srvDefaultPrefix = "_wire" -- TODO: _wire-server? _xmpps?
```
tiago-loureiro: `_wire-server` sounds 👍

libs/federation-util/src/Network/Federation/Util/Internal.hs
```diff
@@ -0,0 +1,68 @@
...
+        Right srvResult' -> do
+            -- Get [(Domain, PortNumber)] of SRV request, if any.
+            -- Sorts the records based on the priority value.
+            orderedSrvResult <- orderSrvResult srvResult'
+            return $ Just $ fmap (\(_, _, port, domain) -> (domain, port)) orderedSrvResult
+        -- The service is not available at this domain.
+        Left _ -> return Nothing
+
+-- FUTUREWORK: maybe improve sorting algorithm here?
```
tiago-loureiro: In terms of performance or logic wise?

libs/federation-util/src/Network/Federation/Util/Internal.hs
jschaul: This case just deals with the case explicitly mentioned in [the SRV RFC](https://tools.ietf.org/html/rfc2782):

> If there is precisely one SRV RR, and its Target is "." (the root domain), abort.

If someone sets multiple records, one of them being the root domain, well, then just the normal logic applies and a server may not get contacted. We can't prevent people from misconfiguring their DNS.

jschaul:
I added more tests. @tiago-loureiro can you take another look if you're happy with the coverage?

PR #967, opened 2020-02-07
Refactor: separate http parsing / generation from app logic.

fisx:
### problem

API handlers carry very little type information, which makes the code hard to read.

### solution

- split up every handler function into a wrapper that takes care of json parsing and rendering, and a core that contains the app logic.
- the core takes and returns a typed values.
- the core is in the `AppIO` monad wherever feasible, and the wrapper is in `Handler` (as before).
- the core carries the name of the original, un-split function; the wrapper has a `H` appended to the original name.

### rationale

This change is easy to do on only parts of the code base, will make the code more readable in the short run, and is hard to get wrong (errors should all be caught by the compiler immediately).

It is helpful for the transition to servant, since the core functions are closer (and ideally identical) to the handler functions we will need there.

The name choices make sure that the original names remain after the transition to servant, and the `H` parts will go away.

### DIFF
--- a/services/brig/src/Brig/API.hs
+++ b/services/brig/src/Brig/API.hs
@@ -1270,15 +1270,18 @@ listUsers self = \case
     byIds uids = lift $ API.lookupProfiles self uids
 
 listActivatedAccountsH :: JSON ::: Either (List UserId) (List Handle) -> Handler Response
-listActivatedAccountsH (_ ::: qry) = case qry of
+listActivatedAccountsH (_ ::: qry) = do
+    json <$> lift (listActivatedAccounts qry)
+
+listActivatedAccounts :: Either (List UserId) (List Handle) -> AppIO [UserAccount]
+listActivatedAccounts = \case
     Left  us -> byIds (fromList us)
     Right hs -> do
-        us <- mapM (lift . API.lookupHandle) (fromList hs)
+        us <- mapM (API.lookupHandle) (fromList hs)
         byIds (catMaybes us)
   where
-    byIds uids = json
-               . filter (isJust . userIdentity . accountUser)
-              <$> (lift $ API.lookupAccounts uids)
+    byIds uids = filter (isJust . userIdentity . accountUser)
+             <$> API.lookupAccounts uids
 
 listAccountsByIdentityH :: JSON ::: Either Email Phone -> Handler Response
 listAccountsByIdentityH (_ ::: emailOrPhone) = lift $ json
@@ -1286,17 +1289,31 @@ listAccountsByIdentityH (_ ::: emailOrPhone) = lift $ json
 
 getActivationCodeH :: JSON ::: Either Email Phone -> Handler Response
 getActivationCodeH (_ ::: emailOrPhone) = do
+    json <$> getActivationCode emailOrPhone
+
+getActivationCode :: Either Email Phone -> Handler GetActivationCodeResp
+getActivationCode emailOrPhone = do
     apair <- lift $ API.lookupActivationCode emailOrPhone
-    maybe (throwStd activationKeyNotFound) (return . found) apair
-  where
-    found (k, c) = json $ object [ "key" .= k, "code" .= c ]
+    maybe (throwStd activationKeyNotFound) (return . GetActivationCodeResp) apair
+
+data GetActivationCodeResp = GetActivationCodeResp (ActivationKey, ActivationCode)
+
+instance ToJSON GetActivationCodeResp where
+    toJSON (GetActivationCodeResp (k, c)) = object [ "key" .= k, "code" .= c ]
 
 getPasswordResetCodeH :: JSON ::: Either Email Phone -> Handler Response
 getPasswordResetCodeH (_ ::: emailOrPhone) = do
+    json <$> getPasswordResetCode emailOrPhone
+
+getPasswordResetCode :: Either Email Phone -> Handler GetPasswordResetCodeResp
+getPasswordResetCode emailOrPhone = do
     apair <- lift $ API.lookupPasswordResetCode emailOrPhone
-    maybe (throwStd invalidPwResetKey) (return . found) apair
-  where
-    found (k, c) = json $ object [ "key" .= k, "code" .= c ]
+    maybe (throwStd invalidPwResetKey) (return . GetPasswordResetCodeResp) apair
+
+data GetPasswordResetCodeResp = GetPasswordResetCodeResp (PasswordResetKey, PasswordResetCode)
+
+instance ToJSON GetPasswordResetCodeResp where
+    toJSON (GetPasswordResetCodeResp (k, c)) = object [ "key" .= k, "code" .= c ]
 
 updateUserH :: UserId ::: ConnId ::: JsonRequest UserUpdate -> Handler Response
 updateUserH (uid ::: conn ::: req) = do
@@ -1314,8 +1331,13 @@ getAccountStatusH :: JSON ::: UserId -> Handler Response
 getAccountStatusH (_ ::: usr) = do
     status <- lift $ API.lookupStatus usr
     return $ case status of
-        Just s  -> json $ object ["status" .= s]
+        Just s  -> json $ AccountStatusResp s
         Nothing -> setStatus status404 empty
+
+data AccountStatusResp = AccountStatusResp AccountStatus
+
+instance ToJSON AccountStatusResp where
+    toJSON (AccountStatusResp s) = object ["status" .= s]
 
 changePhoneH :: UserId ::: ConnId ::: JsonRequest PhoneUpdate -> Handler Response
 changePhoneH (u ::: _ ::: req) = do
--- a/services/brig/src/Brig/API.hs
+++ b/services/brig/src/Brig/API.hs
@@ -1303,12 +1303,11 @@ instance ToJSON GetActivationCodeResp where
 
 getPasswordResetCodeH :: JSON ::: Either Email Phone -> Handler Response
 getPasswordResetCodeH (_ ::: emailOrPhone) = do
-    json <$> getPasswordResetCode emailOrPhone
+    maybe (throwStd invalidPwResetKey) (pure . json) =<< lift (getPasswordResetCode emailOrPhone)
 
-getPasswordResetCode :: Either Email Phone -> Handler GetPasswordResetCodeResp
+getPasswordResetCode :: Either Email Phone -> AppIO (Maybe GetPasswordResetCodeResp)
 getPasswordResetCode emailOrPhone = do
-    apair <- lift $ API.lookupPasswordResetCode emailOrPhone
-    maybe (throwStd invalidPwResetKey) (return . GetPasswordResetCodeResp) apair
+    GetPasswordResetCodeResp <$$> API.lookupPasswordResetCode emailOrPhone
 
 data GetPasswordResetCodeResp = GetPasswordResetCodeResp (PasswordResetKey, PasswordResetCode)
 
--- a/services/brig/src/Brig/API.hs
+++ b/services/brig/src/Brig/API.hs
@@ -24,7 +24,7 @@ import Data.Misc (IpAddr (..))
 import Data.Range
 import Data.Text.Encoding (decodeLatin1)
 import Data.Text.Lazy (pack)
-import Galley.Types (UserClients (..))
+import Galley.Types (UserClients (..), UserClientMap(..))
 import Network.HTTP.Types.Status
 import Network.Wai (Response, lazyRequestBody)
 import Network.Wai.Predicate hiding (setStatus, result)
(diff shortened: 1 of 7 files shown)


fisx:
There are actually a few handlers that already are just the wrappers, with a clean call into a single application logic function.  I left those unchanged in most cases to avoid boring boilerplate.

services/brig/src/Brig/Provider/API.hs
```diff
@@ -790,69 +856,93 @@ removeBotH (zusr ::: zcon ::: cid ::: bid) = do
...
 
 botGetUserClientsH :: UserId -> Handler Response
-botGetUserClientsH uid =
-    json <$> lift (fmap pubClient <$> User.lookupClients uid)
+botGetUserClientsH uid = do
+    json <$> botGetUserClients uid
+
+botGetUserClients :: UserId -> Handler [PubClient]
+botGetUserClients uid = do
+    lift (pubClient <$$> User.lookupClients uid)
```
mheinzel: The `lift` could be pulled out of this function.

services/brig/src/Brig/API.hs
```diff
@@ -1488,56 +1484,73 @@ updateRichInfo (uid ::: _ ::: req) = do
...
+            ActivationPass                -> ActivationResponseNothing204
   where
-    respond (Just ident) first = setStatus status200 $ json (ActivationResponse ident first)
-    respond Nothing      _     = setStatus status200 empty
+    respond (Just ident) first = ActivationResponseJust $ ActivationResponse ident first
+    respond Nothing      _     = ActivationResponseNothing200
+
+data ActivationResponseWithStatus
+    = ActivationResponseJust ActivationResponse
+    | ActivationResponseNothing200
```
mheinzel: What do you think about not putting status codes here, but more descriptive words instead? But I'm not sure I understand what the different cases exactly mean and why they have those specific status codes.

services/brig/src/Brig/API.hs
```diff
@@ -950,16 +950,25 @@ sitemap o = do
 
 setPropertyH :: UserId ::: ConnId ::: PropertyKey ::: JsonRequest PropertyValue -> Handler Response
 setPropertyH (u ::: c ::: k ::: req) = do
+    empty <$ (setProperty u c k =<< lazyParsePropertyValue (lazyRequestBody (fromJsonRequest req)))
+
+setProperty :: UserId -> ConnId -> PropertyKey -> PropertyValue -> Handler ()
+setProperty u c k propval = do
     maxKeyLen <- fromMaybe defMaxKeyLen <$> view (settings . propertyMaxKeyLen)
-    maxValueLen <- fromMaybe defMaxValueLen <$> view (settings . propertyMaxValueLen)
     unless (Text.compareLength (Ascii.toText (propertyKeyName k)) (fromIntegral maxKeyLen) <= EQ) $
```
mheinzel: The difference here is that we already try to parse the full thing and only afterwards check that it's not too long. Without this change, the check happens first. Maybe we should leave this as it was for now and discuss it next week?

services/brig/src/Brig/API.hs
```diff
@@ -1200,42 +1200,63 @@ createUserNoVerify uData = do
...
         Just n  -> json $ object ["name" .= n]
         Nothing -> setStatus status404 empty
 
 listUsersH :: JSON ::: UserId ::: Either (List UserId) (List Handle) -> Handler Response
-listUsersH (_ ::: self ::: qry) = case qry of
-    Left  us -> toResponse =<< byIds (fromList us)
+listUsersH (_ ::: self ::: qry) =
+    toResponse <$> listUsers self qry
+  where
+    toResponse = \case
```
mheinzel: Not necessary to change, just wanted to note that this could maybe be done more nicely by using non-empty lists, e.g. `maybe (setStatus status404 empty) json . nonEmpty <$> listUsers self qry`.

services/brig/src/Brig/API.hs
```diff
@@ -1270,33 +1270,50 @@ listUsers self = \case
...
 getPasswordResetCodeH (_ ::: emailOrPhone) = do
+    json <$> getPasswordResetCode emailOrPhone
+
+getPasswordResetCode :: Either Email Phone -> Handler GetPasswordResetCodeResp
+getPasswordResetCode emailOrPhone = do
     apair <- lift $ API.lookupPasswordResetCode emailOrPhone
-    maybe (throwStd invalidPwResetKey) (return . found) apair
-  where
-    found (k, c) = json $ object [ "key" .= k, "code" .= c ]
+    maybe (throwStd invalidPwResetKey) (return . GetPasswordResetCodeResp) apair
```
mheinzel: We could return a `Maybe GetPasswordResetCodeResp` here, then we could run this in `AppIO`.

services/brig/src/Brig/API.hs
```diff
@@ -1547,20 +1547,18 @@ canBeDeletedH (uid ::: tid) = do
     onlyOwner <- lift (Team.teamOwnershipStatus uid tid)
     case onlyOwner of
```
mheinzel: I would consider this case distinction logic, so it could live in a separate `canBeDeleted` function.

services/brig/src/Brig/User/API/Auth.hs
```diff
@@ -214,40 +214,52 @@ legalHoldLogin l = do
...
-    cs <- lift $ Auth.listCookies u (maybe [] fromList ll)
-    return . json $ CookieList cs
+listCookiesH (u ::: ll ::: _) = json <$> lift (listCookies u ll)
+
+listCookies :: UserId -> Maybe (List CookieLabel) -> AppIO CookieList
+listCookies u ll = do
+    CookieList <$> Auth.listCookies u (maybe [] fromList ll)
 
 rmCookiesH :: UserId ::: JsonRequest RemoveCookies -> Handler Response
 rmCookiesH (uid ::: req) = do
```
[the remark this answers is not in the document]
fisx: I was wondering about that, too, but sometimes having them there saves me one compilation round when I add a second line below without looking here.

services/brig/src/Brig/Team/API.hs
```diff
@@ -133,29 +133,45 @@ routes = do
...
+getInvitationCode t r = do
     code <- lift $ DB.lookupInvitationCode t r
-    maybe (throwStd invalidInvitationCode) (return . found) code
+    maybe (throwStd invalidInvitationCode) (return . FoundInvitationCode) code
+
+data FoundInvitationCode = FoundInvitationCode InvitationCode
+  deriving (Eq, Show, Generic)
+
+instance ToJSON FoundInvitationCode where
+    toJSON c = object [ "code" .= c ]
```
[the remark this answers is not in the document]
fisx: yes, good catch, thanks!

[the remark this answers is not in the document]
fisx: (How annoying that this doesn't get caught by the ghc termination checker... :)

services/brig/src/Brig/Provider/API.hs
```diff
@@ -495,20 +517,28 @@ addServiceH (pid ::: req) = do
...
 
 getServiceH :: ProviderId ::: ServiceId -> Handler Response
 getServiceH (pid ::: sid) = do
-    s <- DB.lookupService pid sid >>= maybeServiceNotFound
-    return (json s)
+    json <$> getService pid sid
+
+getService :: ProviderId -> ServiceId -> Handler Service
+getService pid sid = do
+    DB.lookupService pid sid >>= maybeServiceNotFound
```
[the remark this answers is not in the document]
fisx: > if you leave it for now, we can do it later.

agreed.  i think this would have kept me from finishing even brig, and i still need to get the others done.  effects are a much bigger topic.

services/brig/src/Brig/Team/API.hs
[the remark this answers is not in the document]
fisx: 2f65206320da82073579ad677cd0d816ed7fbcf6

services/brig/src/Brig/Provider/API.hs
```diff
@@ -618,63 +655,85 @@ deleteAccountH (pid ::: req) = do
...
+getServiceTagList () = return (ServiceTagList allTags)
   where
     allTags = [(minBound :: ServiceTag) ..]
 
 updateServiceWhitelistH :: UserId ::: ConnId ::: TeamId ::: JsonRequest UpdateServiceWhitelist -> Handler Response
 updateServiceWhitelistH (uid ::: con ::: tid ::: req) = do
-    upd :: UpdateServiceWhitelist <- parseJsonBody req
+    flip setStatus empty <$> (updateServiceWhitelist uid con tid =<< parseJsonBody req)
+
+updateServiceWhitelist :: UserId -> ConnId -> TeamId -> UpdateServiceWhitelist -> Handler Status
```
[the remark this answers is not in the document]
fisx: 7017464ef187e0f732c7d6f07359f5be74892e63, 097e188a

services/brig/src/Brig/API.hs
fisx: I don't think that's the case.  There are two independent checks: one on the key (which comes from a path segment), and one on the body length.  The check on the body length is sensitive because it needs to happen before parsing, but I didn't think that flipping the order of the two independent checks mattered, so I flipped.

Yes, let's look at it again together on monday.  It's not hard to change back, or make a more conservative change.

services/brig/src/Brig/API.hs
fisx: still, good idea.  *arguably* nicer, yes, but then we need to worry about the serialization of non-empty lists being compatible with the one for empty list.

services/brig/src/Brig/API.hs
fisx: 351e9d4c 

(not sure which is better; now there are some implicit assumptions in both cases about the *reasons* for the boolean.  but i think either one is good enough.)

services/brig/src/Brig/API.hs
fisx: efa3ffaf

services/brig/src/Brig/API.hs
fisx: 097e188a

services/brig/src/Brig/Provider/API.hs
fisx: 117118cb69c5791346aa01a36a7547a9424016e2

services/brig/src/Brig/API.hs
fisx: 2a0419f7e660c5e8557f7ac1a61fc0d71437bc2b

I think now it's the same, except for more names explaining what's happening.

Before:

```haskell
setProperty :: UserId ::: ConnId ::: PropertyKey ::: JsonRequest PropertyValue -> Handler Response
setProperty (u ::: c ::: k ::: req) = do
    maxKeyLen <- fromMaybe defMaxKeyLen <$> view (settings . propertyMaxKeyLen)
    maxValueLen <- fromMaybe defMaxValueLen <$> view (settings . propertyMaxValueLen)
    unless (Text.compareLength (Ascii.toText (propertyKeyName k)) (fromIntegral maxKeyLen) <= EQ) $
        throwStd propertyKeyTooLarge
    lbs <- Lazy.take (maxValueLen + 1) <$> liftIO (lazyRequestBody (fromJsonRequest req))
    unless (Lazy.length lbs <= maxValueLen) $
        throwStd propertyValueTooLarge
    val <- hoistEither $ fmapL (StdError . badRequest . pack) (eitherDecode lbs)
    API.setProperty u c k val !>> propDataError
    return empty
```

After:

```haskell
setPropertyH :: UserId ::: ConnId ::: PropertyKey ::: JsonRequest PropertyValue -> Handler Response
setPropertyH (u ::: c ::: k ::: req) = do
    propkey <- safeParsePropertyKey k
    propval <- safeParsePropertyValue (lazyRequestBody (fromJsonRequest req))
    empty <$ setProperty u c propkey propval

setProperty :: UserId -> ConnId -> PropertyKey -> PropertyValue -> Handler ()
setProperty u c propkey propval = do
    API.setProperty u c propkey propval !>> propDataError

safeParsePropertyKey :: PropertyKey -> Handler PropertyKey
safeParsePropertyKey k = do
    maxKeyLen <- fromMaybe defMaxKeyLen <$> view (settings . propertyMaxKeyLen)
    unless (Text.compareLength (Ascii.toText (propertyKeyName k)) (fromIntegral maxKeyLen) <= EQ) $
        throwStd propertyKeyTooLarge
    pure k

-- | Parse a 'PropertyValue' from a bytestring.  This is different from 'FromJSON' in that
-- checks the byte size of the input, and fails *without consuming all of it* if that size
-- exceeds the settings.
safeParsePropertyValue :: IO Lazy.ByteString -> Handler PropertyValue
safeParsePropertyValue lreqbody = do
    maxValueLen <- fromMaybe defMaxValueLen <$> view (settings . propertyMaxValueLen)
    lbs <- Lazy.take (maxValueLen + 1) <$> liftIO lreqbody
    unless (Lazy.length lbs <= maxValueLen) $
        throwStd propertyValueTooLarge
    hoistEither $ fmapL (StdError . badRequest . pack) (eitherDecode lbs)
```

mheinzel: Ah, yes, you are right. And I also like the updated version of the code better. :+1:

fisx:
Integration tests passed locally.

PR #968, opened 2020-02-10
[WIP] build nginz on nixos

fisx:
This is a convolution of horrible hacks.  I want to get it to work on my machine first, then discuss (probably once @arianvp is back) how to do it properly.

Current status fails here:

```
[...]
 + gcc version: 7.4.0 (GCC) 
checking for gcc -pipe switch ... found
checking for --with-ld-opt="-l/home/mf/src/wire-server/services/nginz/libzauth/lib/libzauth.so" ... not found
./configure: error: the invalid value in --with-ld-opt="-l/home/mf/src/wire-server/services/nginz/libzauth/lib/libzauth.so"
```

This looks to me like `configure` confuses the extra linker options string with the name of an executable?  But I'm probably wrong...

### DIFF
--- a/libs/libzauth/Makefile
+++ b/libs/libzauth/Makefile
@@ -1,7 +1,9 @@
 LANG := en_US.UTF-8
 dist:
 	$(MAKE) -C libzauth-c $@
-	cp libzauth-c/target/release/*.deb .
+
+install:
+	$(MAKE) -C libzauth-c $@
 
 clean:
 	$(MAKE) -C libzauth-c $@
--- a/services/nginz/README.md
+++ b/services/nginz/README.md
@@ -73,3 +73,15 @@ If you are using macOS and you used `brew` to install openssl, the `Makefile` al
 ## How to run it
 
 Have a look at our demo config in [services demo](../../deploy/services-demo/conf/nginz)
+
+
+
+
+## build on nixos
+
+be in this directory
+nix-shell ../../shell.nix
+. ./env.sh
+make -C ../../libs/libzauth install
+make
+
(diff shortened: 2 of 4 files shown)


fisx:
- https://duckduckgo.com/?q=checking+for+--with-ld-opt&ia=web
- https://serverfault.com/questions/869794/nginx-with-cc-opt-and-with-ld-opt-configure-options/905448

will look into it tomorrow.

fisx:
Second commit does it for me, I'm able to run integration tests locally now!

Now the questions is how do you do this properly.  @arianvp?

arianvp:
This is how you would build `nginx` with `libzauth` support idiomatically: 

https://github.com/wireapp/wire-server/commit/a60c79296bd7f812b02ed7c821581d7694171b42

fisx:
Should i pull that commit into this PR, or do you want to?  Either way I think we should have this.

How have you done this on your machine until now?

Thanks!

arianvp:
> Should i pull that commit into this PR, or do you want to? Either way I think we should have this.

I'll do it. the commit isn't fully standalone at the moment.
Also note; it currently takes whatever `nginx` version is in `nixpkgs` and also whatever version of the `modules` that we depend on. I'm not sure if this is a problem or not?  I like it that it's more upstream. Are these modules stable enough for this to be ok; or should I pin the nix packages to the same versions our `Makefile`s and `Dockerfile`s are using?

> How have you done this on your machine until now?
The components I touched in the recent months didn't need nginz for the integration tests

fisx:
> [...] should I pin the nix packages to the same versions our `Makefile`s and `Dockerfile`s are using?

Pinning is probably better?  We're getting emails on new releases, so bumping by hand seems safer than letting it happen without notice.  Hard to tell.

fisx:
https://github.com/wireapp/wire-server/pull/988 has solved this.

PR #969, opened 2020-02-10
Limit max nr of devices

tiago-loureiro:
- Max number of devices is now configurable in options.
- Running tests against `Application`s of services with modified options has been fixed to allow for requests with content in the body.

### DIFF
--- a/libs/bilge/src/Bilge/IO.hs
+++ b/libs/bilge/src/Bilge/IO.hs
@@ -54,7 +54,6 @@ import Control.Monad.Catch
 import Control.Monad.Fail (MonadFail)
 import Control.Monad.Trans.Control
 import Network.HTTP.Client as Client hiding (method, httpLbs)
-import Network.Socket (SockAddr(..), tupleToHostAddress)
 import qualified Network.HTTP.Client as Client (method)
 import qualified Data.ByteString.Lazy as LB
 
@@ -136,7 +135,7 @@ instance MonadHttp Wai.Session where
           , Wai.httpVersion = Client.requestVersion req
           , Wai.requestHeaders = Client.requestHeaders req
           , Wai.isSecure = Client.secure req
-          , Wai.remoteHost = SockAddrInet 8080 (tupleToHostAddress (127, 0, 0, 1))  -- (this is fabricated, but leaving this undefined doesn't work.)
+          , Wai.remoteHost = error "no remote host"
           , Wai.requestHeaderHost = lookupHeader "HOST" req
           , Wai.requestHeaderRange = lookupHeader "RANGE" req
           , Wai.requestHeaderReferer = lookupHeader "REFERER" req
--- a/libs/bilge/src/Bilge/IO.hs
+++ b/libs/bilge/src/Bilge/IO.hs
@@ -120,7 +120,9 @@ trivialBodyReader bodyBytes = do
 
 instance MonadHttp Wai.Session where
   handleRequestWithCont req cont = do
-      wResponse :: Wai.SResponse <- Wai.request wRequest
+      reqBody <- liftIO $ getHttpClientRequestBody (Client.requestBody req)
+      -- `srequest` sets the requestBody for us
+      wResponse :: Wai.SResponse <- Wai.srequest (Wai.SRequest wRequest reqBody)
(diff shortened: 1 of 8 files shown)


fisx:
better guess: 1268f21a  (how did this ever work?!)

i will temporarily remove the `[WIP]` in the PR name so we'll know tomorrow morning if this helps.

tiago-loureiro:
> better guess: [1268f21](https://github.com/wireapp/wire-server/commit/1268f21a14a4272f60a536bbde8b6fcdbda70afe) (how did this ever work?!)
> 
> i will temporarily remove the `[WIP]` in the PR name so we'll know tomorrow morning if this helps.

No, that's not it... but I do think you are on the right track. Looking [here](https://github.com/wireapp/wire-server/blob/develop/libs/bilge/src/Bilge/IO.hs#L131-L140) there is indeed no request body being set which explains the behaviour we have been seeing.

mheinzel:
Not sure what to do with the `remoteHost` change, but I just pushed a fix for the empty request body.

We used wai-extra's [request](https://www.stackage.org/haddock/lts-14.12/wai-extra-3.0.28/Network-Wai-Test.html#v:request), which for some reason overrides the requestBody with an empty one. I switched to [srequest](https://www.stackage.org/haddock/lts-14.12/wai-extra-3.0.28/Network-Wai-Test.html#v:srequest), which overrides it with a given ByteString and the problematic test passes now.

fisx:
Cool, thanks!

I would leave the remote host field as it is now.  More likely to work in future corner cases.

libs/bilge/src/Bilge/IO.hs
```diff
@@ -133,7 +136,7 @@ instance MonadHttp Wai.Session where
           , Wai.httpVersion = Client.requestVersion req
           , Wai.requestHeaders = Client.requestHeaders req
           , Wai.isSecure = Client.secure req
-          , Wai.remoteHost = error "no remote host"
+          , Wai.remoteHost = SockAddrInet 8080 (tupleToHostAddress (127, 0, 0, 1))  -- (this is fabricated, but leaving this undefined doesn't work.)
```
tiago-loureiro: Is this actually needed then?

services/brig/test/integration/API/User/Client.hs
```diff
@@ -163,7 +163,7 @@ testTooManyClients opts brig = do
 
     -- There is only one temporary client, adding a new one
     -- replaces the previous one.
-    forM_ [0..(9 :: Int)] $ \i ->
+    forM_ [0..(3 :: Int)] $ \i ->
```
mheinzel: Not 100% sure about this, since 9 explicitly exceeds the default max permitted clients. So if for some reason temporary clients get treated as permanent, this test would (if I understand correctly) fail with 9, but succeed with 3.

libs/bilge/src/Bilge/IO.hs
mheinzel: I don't think so. All tests passed for me without this change.

services/brig/test/integration/API/User/Client.hs
fisx: good point.  this should be pulled into the block with the changed settings.  or just rolled back.

libs/bilge/src/Bilge/IO.hs
fisx: I just opted for leaving it in to allow for future corner cases, but I don't have a strong case and I don't insist.

mheinzel: I think in this case where we call the `Application` directly via `Wai.Session`, we don't need a remote host. There is no actual networking going on.
If for some reason this code starts doing network requests, it might even be better to find out because of an error instead of silently assuming an address. :man_shrugging:

tiago-loureiro: Yes, this was exactly @ChrisPenner point [back in the days](https://github.com/wireapp/wire-server/pull/724#discussion_r278115085), and @fisx agreed ;)

mheinzel: I'll revert this commit and then wait for a last CI run to succeed.

services/brig/test/integration/API/User/Client.hs
mheinzel: I'll pull it into the settings block, we should be explicit about the configuration values we expect.

PR #970, opened 2020-02-10
Ignore flaky cookie limit test

akshaymankar:
The test hardly ever passes on K8s, specially after GHC 8.6.5 upgrade. This PR marks it ignored. I have created another issue to track it: https://github.com/zinfra/backend-issues/issues/1150


akshaymankar:
As this never fails locally, we decided to only skip it in K8s. PR for that is here: https://github.com/wireapp/wire-server-deploy/pull/184

PR #971, opened 2020-02-11
Symlink local dist folders in tools to the global one

akshaymankar:
Similar to 3063265319e8301a9769941a6f83aab3c9e56dc5.

This was done to make it work with both stack 1 and 2. Not sure if we need to support stack 1 anymore. But for now, this seems fine for consistency.

Unfortunately the PR pipeline won't be able to verify this, as we don't compile tools in the PR pipelines. 

To verify this before merge I have created another pipeline here: https://bridge-ie-concourse.zinfra.io/teams/main/pipelines/stern-test. We can use that to know if this works as intended.


fisx:
(I don't think we need to test this, just merge!)

PR #973, opened 2020-02-11
Better docs on brig integration yaml

tiago-loureiro:
Small doc follow up to https://github.com/wireapp/wire-server/pull/969


fisx:
seem uncontroversial enough, i'll just merge this.

PR #974, opened 2020-02-13
Formatting with ormolu

jschaul:
https://github.com/zinfra/backend-issues/issues/1163

I used latest ormolu from master, compiled using this stack file: 

```yml
cat ormolu/stack.yaml
resolver: lts-14.3
packages:
  - '.'
extra-deps:
- ghc-lib-parser-8.8.2.20200205@sha256:343f889f7b29f5ec07cf0d18d2a53f250fa5c002b6468a6a05b385d0191b8d34,8408
```

and the following ugly bash script which excludes CPP files (requires `ag`):

```bash
#!/usr/bin/env bash

CPP=( $(ag -G .hs -l CPP) )

containsElement () {
  local e match="$1"
  shift
  for e; do [[ "$e" == "$match" ]] && return 0; done
  return 1
}

for f in $(find . | grep .hs$ | grep -v .stack); do
    if ! containsElement ${f:2} "${CPP[@]}" ; then
        ormolu --ghc-opt -XTypeApplications --ghc-opt -XBangPatterns --mode inplace "$f"
        git add "$f"
    fi
done
```


fisx:
Oooooh, exciting!  :)

For the bash script, I prefer the following alternative: https://github.com/wireapp/saml2-web-sso/blob/7a16875b381fc7bce6624defaf38bfd780bdf14e/brittany.sh

Also, I would like to move the CPP code into modules that contain as little code a possible, but that can be done independently.

fisx:
ormolu (as generated here):

```haskell
testUpdateEvents :: Brig -> Galley -> Cannon -> Http ()
testUpdateEvents brig galley cannon = do
  (alice, tid) <- createUserWithTeam brig galley
  inviteeEmail <- randomEmail
  -- invite and register Bob
  let invite = InvitationRequest inviteeEmail (Name "Bob") Nothing Nothing
  inv <- responseJsonError =<< postInvitation brig tid alice invite
  Just inviteeCode <- getInvitationCode brig tid (inInvitation inv)
  rsp2 <-
    post
      ( brig . path "/register"
          . contentJson
          . body (accept inviteeEmail inviteeCode)
      )
      <!! const 201 === statusCode
  let Just bob = userId <$> responseJsonMaybe rsp2
  -- ensure Alice and Bob are not connected
  void $ getConnection brig bob alice <!! const 404 === statusCode
  void $ getConnection brig alice bob <!! const 404 === statusCode
  -- Alice updates her profile
  let newColId = Just 5
      newAssets = Just [ImageAsset "abc" (Just AssetComplete)]
      newName = Just $ Name "Alice in Wonderland"
      newPic = Nothing -- Legacy
      userUpdate = UserUpdate newName newPic newAssets newColId
      update = RequestBodyLBS . encode $ userUpdate
  -- Update profile & receive notification
  WS.bracketRN cannon [alice, bob] $ \[aliceWS, bobWS] -> do
    put (brig . path "/self" . contentJson . zUser alice . zConn "c" . body update)
      !!! const 200 === statusCode
    liftIO $ mapConcurrently_ (\ws -> assertUpdateNotification ws alice userUpdate) [aliceWS, bobWS]
```

(old version of) floskell:

```haskell
testUpdateEvents :: Brig -> Galley -> Cannon -> Http ()
testUpdateEvents brig galley cannon =
  do
    (alice, tid) <- createUserWithTeam brig galley
    inviteeEmail <- randomEmail
    -- invite and register Bob
    let invite = InvitationRequest inviteeEmail (Name "Bob") Nothing Nothing
    inv <- responseJsonError =<< postInvitation brig tid alice invite
    Just inviteeCode <- getInvitationCode brig tid (inInvitation inv)
    rsp2 <- post (brig.path "/register".contentJson.body (accept inviteeEmail inviteeCode))
      <!! const 201
      === statusCode
    let Just bob = userId <$> responseJsonMaybe rsp2
    -- ensure Alice and Bob are not connected
    void $ getConnection brig bob alice <!! const 404 === statusCode
    void $ getConnection brig alice bob <!! const 404 === statusCode
    -- Alice updates her profile
    let newColId   = Just 5
        newAssets  = Just [ImageAsset "abc" (Just AssetComplete)]
        newName    = Just $ Name "Alice in Wonderland"
        newPic     = Nothing -- Legacy
        userUpdate = UserUpdate newName newPic newAssets newColId
        update     = RequestBodyLBS.encode $ userUpdate
    -- Update profile & receive notification
    WS.bracketRN cannon [alice, bob]
      $ \[aliceWS, bobWS] -> do
        put (brig.path "/self".contentJson.zUser alice.zConn "c".body update) !!! const 200
          === statusCode
        liftIO
          $ mapConcurrently_ (\ws -> assertUpdateNotification ws alice userUpdate)
                             [aliceWS, bobWS]
```

can't make up my mind...  :)

fisx:
the script in e1fec9f assumes that CPP is only ever enabled in the module itself.  is that something we can enforce?  i vote for yes.

i hope to have time to finish this over the week-end.

tools/ormolu.sh
```diff
@@ -0,0 +1,31 @@
...
+set -e
+
+# TODO: change to stack project root directory.
+
+
+if [ "`git status -s src/ test/ | grep -v \?\?`" != "" ]; then
+    echo "working copy not clean."
+    if [ "$1" == "-f" ]; then
+        echo "running with -f.  THIS WILL DESTROY YOUR UNCOMMITTED CHANGES."
+        git reset --hard HEAD
```
mheinzel: Do we really need to reset to HEAD? If I really want to run this on my uncommitted changes via `-f`, I would expect them to get formatted (which is kind of unsafe, since I don't have a versioned copy of them, but usually fine), but not just dropped completely. Not sure what the advantage of that is.

mheinzel:
Did you check that this is idempotent by running the command twice? I know it's one of the design goals of ormolu, but just to make sure we don't end up with weird diffs in random places later.

tools/ormolu.sh
fisx: fair enough, i changed that back.

fisx:
> Did you check that this is idempotent by running the command twice? I know it's one of the design goals of ormolu, but just to make sure we don't end up with weird diffs in random places later.

there is a flag for that; i enabled it.

(i think that may mean that we'll need to deal with ormolu bugs every now and then, where idempotency doesn't work.)

mheinzel:
Just as a note, on my machine the `tools/ormolu.sh` command takes almost 5 minutes, which is not ideal, but tolerable with the right workflow.

fisx:
I personally am happy with the script now, but if we agree on it we will have to move all CPP code to minimal modules whose names match `/CPP$/`.  We should probably make that change and get it merged separately before merging the first ormolu run.

We should also have a concourse pipeline running the script with `-c`.

fisx:
> Just as a note, on my machine the `tools/ormolu.sh` command takes almost 5 minutes, which is not ideal, but tolerable with the right workflow.

yes, as written in the `-h` output, this is not supposed to be used as a git commit hook, or an editor save hook, but mostly for a pipeline making sure all formatting has been applied.  one module usually takes slightly less than a second on my machine.  still slightly irritating for a save-hook, but i'm ready to try to get used to it.

jschaul:
> move all CPP code to minimal modules whose names match /CPP$/

I think `git grep -L "LANGUAGE CPP"` will work fine; it seems unnecessary to rename modules/files for this purpose. I'm not opposed to reducing the size of files with CPP, to increase the amount of lines that can be formatted, but I'm not sure how worthwhile and useful of a task that is (i.e. if it takes more than a few hours I think we should rather work on something more productive). I think files/modules should be named according to what they do, not according to "contains / does not contain a CPP pragma".

In my opinion this PR (formatting all non-CPP files) and the work on reducing unformatted lines by refactoring some code around could be two different PRs and be merged independently. What do you think?

A pipeline can only enter into effect *after* this PR is merged, not before, so it is not blocking this PR.

tools/ormolu.sh
```diff
@@ -65,7 +65,7 @@ FAILURES=0
 for hsfile in $(git grep -L "LANGUAGE CPP" | grep '\.hs$'); do
     echo -n "$hsfile..."
     FAILED=0
-    stack exec -- ormolu --mode $ARG_ORMOLU_MODE --check-idempotency $LANGUAGE_EXTS "$hsfile" || FAILED=1
+    ormolu --mode $ARG_ORMOLU_MODE --check-idempotency $LANGUAGE_EXTS "$hsfile" || FAILED=1
```
fisx: wow.  good catch with the stack start time, thanks!

tools/ormolu.sh
```diff
@@ -2,7 +2,7 @@
 
 cd "$( dirname "${BASH_SOURCE[0]}" )/.."
 
-stack exec -- which ormolu >/dev/null || ( echo "please run stack install ormolu."; exit 1 )
+which ormolu >/dev/null || ( echo "please run 'stack install ormolu' and ensure ormolu is on your PATH."; exit 1 )
```
fisx: not tested:

```suggestion
ORMOLU_VERSION=0.0.3.1
ormolu -v >/dev/null 2>&1 | grep -q $ORMOLU_VERSION || ( echo "please install ormolu $ORMOLU_VERSION (eg., run 'stack install ormolu' and ensure ormolu is on your PATH.)"; exit 1 )
```

fisx:
> > move all CPP code to minimal modules whose names match /CPP$/
> 
> I think `git grep -L "LANGUAGE CPP"` will work fine; it seems unnecessary to rename modules/files for this purpose. I'm not opposed to reducing the size of files with CPP, to increase the amount of lines that can be formatted, but I'm not sure how worthwhile and useful of a task that is (i.e. if it takes more than a few hours I think we should rather work on something more productive). I think files/modules should be named according to what they do [...]

I can live with your change, but I prefer the filename-based solution because it will fail if somebody doesn't know (or doesn't care about) the convention of keeping CPP in minimal modules, which I think is a good thing independently of the formatting story. 

> In my opinion this PR (formatting all non-CPP files) and the work on reducing unformatted lines by refactoring some code around could be two different PRs and be merged independently. What do you think?

yes, i just wanted to do it the other way round.  but you're right, it does not matter at all.

> A pipeline can only enter into effect _after_ this PR is merged, not before, so it is not blocking this PR.

:+1:

tools/ormolu.sh
mheinzel: This means we have to bump the ormolu version in this file every time we bump it in our stack.yaml? If there's a way to avoid that, that would be nice. Otherwise, we should at least leave a comment in the stack.yaml.

fisx: good idea: d6c1ef12e47911709b1e37157c95e974ee5bd8b6

fisx: (sorry about the perl dep, if anybody wants to translate this to sed, i'll start a translation table for next time.  :))

fisx: oops: 87f8cc2fdc0d501d10be79db6476affe64e17470

jschaul: :+1: for the ormolu version checking (and adding to stack.yml)

PR #977, opened 2020-02-17
debug spar prod

fisx:
We need to collect more information from `/sso/finalize-login` for cases where customers complain about errors.  Currently, customers can only show us a json blob containing some info about the error, but not the original request, which would be required for reproducing it.

Depends on https://github.com/wireapp/saml2-web-sso/pull/57

Couple things to decide:
1. we could not log this with level `Info`, but make the json blob that customers get really big.  gives us the information in cases where customers are able to cut&paste, but logs will have less potentially sensitive data.
2.  we could add an end-point to spar for whitelisting teams that we log.
3.  we could just log the data and decide it's not sensitive.  (it is only logged in case of real errors, not "login rejected".)

answer: 1.


arianvp:
Could we fix https://github.com/zinfra/backend-issues/issues/1081 while we're at it?

PR #978, opened 2020-02-18
Remove unnecessary LANGUAGE CPP pragmas

mheinzel:
Those files were not actually using any CPP. Now they get formatted, too!

(14 files remaining with actual CPP usage)


fisx:
```
brig-types                       > /wire-server/libs/brig-types/src/Brig/Types/Team/LegalHold.hs:227:12: error:
brig-types                       >     Pattern syntax in expression context: fromIdObject@ClientId
brig-types                       >     Type application syntax requires a space before '@'
brig-types                       >     |
brig-types                       > 227 |       <*> (fromIdObject@ClientId <$$> (o .:? "client"))
```

i think you did an ormolu run without the language extensions turned on.  This happened to me, too.

mheinzel:
Yes, just saw it locally, too. It happened before I added language extensions to my ormolu editor integration. A bit annoying that re-running it with the correct settings doesn't fix it, but I'm fixing it manually now.

PR #980, opened 2020-02-18
Licenses

jschaul:
Adds automation to generate license headers in haskell/rust source files. See https://github.com/zinfra/backend-issues/issues/1154 for details.

A separate PR will be made that applies this change to all files if this is approved.

See also #981

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -48,11 +48,10 @@ formatf:
 formatc:
 	./tools/ormolu.sh -c
 
+# For any Haskell or Rust file that doesn't mention AGPL yet, add a license header.
+# It's your own reponsibility to keep ormolu happy.
 .PHONY: add-license
 add-license:
-	# For any Haskell or Rust file that doesn't mention AGPL yet,
-	#   add a license header.
-	# It's your own reponsibility to keep ormolu happy.
 	for file in $$(git grep -L "GNU Affero General Public License" | grep '\.hs$$\|\.hsc$$\|\.rs$$'); do \
 		echo "Adding license to $${file}."; \
 		licensure -i $${file}; \
--- a/Makefile
+++ b/Makefile
@@ -36,19 +36,29 @@ haddock-shallow:
 # formats all Haskell files (which don't contain CPP)
 .PHONY: format
 format:
-	licensure -p
 	./tools/ormolu.sh
 
 # formats all Haskell files even if local changes are not committed to git
 .PHONY: formatf
 formatf:
-	licensure -p
 	./tools/ormolu.sh -f
 
 # checks that all Haskell files are formatted; fail if a `make format` run is needed.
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -36,11 +36,13 @@ haddock-shallow:
 # formats all Haskell files (which don't contain CPP)
 .PHONY: format
 format:
+	licensure -p
 	./tools/ormolu.sh
 
 # formats all Haskell files even if local changes are not committed to git
 .PHONY: formatf
 formatf:
+	licensure -p
```
fisx: If we do this, we should also add a CI check that running licensure is idempotent.

docs/developer/dependencies.md
```diff
@@ -95,6 +95,23 @@ sudo apt install rustc cargo -y
...
+
+## Formatting Haskell files
+
+You need `ormolu` on your PATH, get it with `stack install ormolu`
+
+## Generating license headers
+
+We use [`licensure`](https://github.com/chasinglogic/licensure).
+
+As of 2020-02-18, you need the latest version:
```
fisx:
```suggestion
To get it:
```

docs/developer/dependencies.md
```diff
@@ -95,6 +95,23 @@ sudo apt install rustc cargo -y
...
+
+```
+git clone https://github.com/chasinglogic/licensure
+cd licensure
+cargo install --path .
```
fisx:
```suggestion
git checkout 96d7fad9201e19e5193bf7435ab5fdf4a6ad4685  # master as of 2020-02-18
cargo install --path .
```

Makefile
mheinzel: It actually isn't idempotent, see my review.

jschaul:
@mheinzel have another look, I added a makefile target that ensures headers are not added multiple times. It's not as good as idempotence, of course, but perhaps "good enough" to allow someone creating a new file to add a license header without going copy-pasting.

Makefile
```diff
@@ -48,6 +48,18 @@ formatf:
...
+add-license:
+	# For any Haskell or Rust file that doesn't mention AGPL yet,
+	#   add a license header.
+	# It's your own reponsibility to keep ormolu happy.
+	for file in $$(git grep -L "GNU Affero General Public License" | grep '\.hs$$\|\.rs$$'); do \
+		echo "Adding license to $${file}."; \
+		licensure -i $${file}; \
+	done;
+	@echo ""
+	@echo "you most probably want to run 'make formatf' now to keep ormolu happy"
```
jschaul, line 61: Of course this task could directly run ormolu, but I think it's preferrable to leave people the option to check their changes before having ormolu wreak havoc (potentially). So I prefer to separate these steps out and just do an echo here.

PR #981, opened 2020-02-18
three example files 

jschaul:
Related to https://github.com/wireapp/wire-server/pull/980 and https://github.com/zinfra/backend-issues/issues/1154

* 2 with a regular header (Haskell, rust);
* one manually-edited more complicated licensed file (which includes modified third-party code)

If this is approved, either on this PR, or a subsequent one, all other files can be committed with an extra header.


jschaul:
Obsoleted with #1045

PR #988, opened 2020-02-21
Compile nginx with libzauth  using nix

arianvp:
This allows you to use nginz on Nix and NixOS machines

Noteworthy:

both vts and moreheaders are packaged in nixpkgs, so no need
to fetch those git submodules...   I don't think it's a very big problem. We can pin them but it seems to be working fine for now...

I also use niv for version management of nixpkgs. instead of hacky own
thing.

To build this with the newly set up binary cache, try this:
```
$ nix-env -iA cachix -f https://cachix.org/api/v1/install
$ cachix use wire-server
$ nix-build ./nix -A nginz
```


services/integration.sh
```diff
@@ -105,9 +105,19 @@ run spar "" ${orange}
 
 function run_nginz() {
     colour=$1
-    prefix=$([ -w /usr/local ] && echo /usr/local || echo "${HOME}/.wire-dev")
-    (cd ${NGINZ_WORK_DIR} && LD_LIBRARY_PATH=$LD_LIBRARY_PATH:${prefix}/lib/ ${TOP_LEVEL}/dist/nginx -p ${NGINZ_WORK_DIR} -c ${NGINZ_WORK_DIR}/conf/nginz/nginx.conf -g 'daemon off;' || kill_all) \
-        | sed -e "s/^/$(tput setaf ${colour})[nginz] /" -e "s/$/$(tput sgr0)/" &
+
+    # For nix we dont need LD_LIBRARY_PATH; we link agaiinterpolationnst libzauth directly.
```
arianvp: typo

fisx:
I probably didn't set [this variable](https://github.com/wireapp/wire-server/blob/07789ae6170936e9df877389d748a2ae04f9fa07/services/integration.sh#L124).  Will investigate later.

PR #989, opened 2020-02-21
Refactor: separate HTTP handlers from app logic (galley)

mheinzel:
This has already been done for brig in #967, now for galley.

### DIFF
--- a/services/galley/src/Galley/API/CustomBackend.hs
+++ b/services/galley/src/Galley/API/CustomBackend.hs
@@ -40,6 +40,5 @@ internalPutCustomBackendByDomainH (domain ::: req) = do
 
 internalDeleteCustomBackendByDomainH :: EmailDomain ::: JSON -> Galley Response
 internalDeleteCustomBackendByDomainH (domain ::: _) = do
-  -- simple enough to not need a separate function
   Data.deleteCustomBackend domain
   pure (empty & setStatus status200)
--- a/services/galley/src/Galley/API/CustomBackend.hs
+++ b/services/galley/src/Galley/API/CustomBackend.hs
@@ -1,7 +1,7 @@
 module Galley.API.CustomBackend
-  ( getCustomBackendByDomain,
-    internalPutCustomBackendByDomain,
-    internalDeleteCustomBackendByDomain,
+  ( getCustomBackendByDomainH,
+    internalPutCustomBackendByDomainH,
+    internalDeleteCustomBackendByDomainH,
   )
 where
 
@@ -19,21 +19,27 @@ import Network.Wai.Utilities
 
 -- PUBLIC ---------------------------------------------------------------------
 
-getCustomBackendByDomain :: EmailDomain ::: JSON -> Galley Response
-getCustomBackendByDomain (domain ::: _) =
+getCustomBackendByDomainH :: EmailDomain ::: JSON -> Galley Response
+getCustomBackendByDomainH (domain ::: _) =
+  json <$> getCustomBackendByDomain domain
+
+getCustomBackendByDomain :: EmailDomain -> Galley CustomBackend
+getCustomBackendByDomain domain =
   Data.getCustomBackend domain >>= \case
     Nothing -> throwM (customBackendNotFound domain)
-    Just customBackend -> pure (json customBackend)
+    Just customBackend -> pure customBackend
 
 -- INTERNAL -------------------------------------------------------------------
 
-internalPutCustomBackendByDomain :: EmailDomain ::: JsonRequest CustomBackend -> Galley Response
-internalPutCustomBackendByDomain (domain ::: req) = do
+internalPutCustomBackendByDomainH :: EmailDomain ::: JsonRequest CustomBackend -> Galley Response
+internalPutCustomBackendByDomainH (domain ::: req) = do
   customBackend <- fromJsonBody req
+  -- simple enough to not need a separate function
   Data.setCustomBackend domain customBackend
   pure (empty & setStatus status201)
 
-internalDeleteCustomBackendByDomain :: EmailDomain ::: JSON -> Galley Response
-internalDeleteCustomBackendByDomain (domain ::: _) = do
+internalDeleteCustomBackendByDomainH :: EmailDomain ::: JSON -> Galley Response
+internalDeleteCustomBackendByDomainH (domain ::: _) = do
+  -- simple enough to not need a separate function
   Data.deleteCustomBackend domain
   pure (empty & setStatus status200)
--- a/services/galley/src/Galley/API/LegalHold.hs
+++ b/services/galley/src/Galley/API/LegalHold.hs
@@ -36,8 +36,13 @@ isLegalHoldEnabled tid = do
     Just LegalHoldDisabled -> False
     Nothing -> False
 
-createSettings :: UserId ::: TeamId ::: JsonRequest NewLegalHoldService ::: JSON -> Galley Response
-createSettings (zusr ::: tid ::: req ::: _) = do
+createSettingsH :: UserId ::: TeamId ::: JsonRequest NewLegalHoldService ::: JSON -> Galley Response
+createSettingsH (zusr ::: tid ::: req ::: _) = do
+  newService :: NewLegalHoldService <- fromJsonBody req
+  setStatus status201 . json <$> createSettings zusr tid newService
+
+createSettings :: UserId -> TeamId -> NewLegalHoldService -> Galley ViewLegalHoldService
+createSettings zusr tid newService = do
   assertLegalHoldEnabled tid
   membs <- Data.teamMembers tid
   let zothers = map (view userId) membs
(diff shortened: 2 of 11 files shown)


mheinzel:
There are integration test failures because of status code mismatches, I'll update this PR once it's fixed.

mheinzel:
Passing now.

I don't like the fact that we have endpoints that only in some cases require a JSON body (as well as potentially masking invalid-payload errors by permission errors), but I can't change that. At least using `OptionalJsonRequest` seems to be the proper way of specifying this behaviour.

services/galley/src/Galley/API/LegalHold.hs
```diff
@@ -131,8 +156,18 @@ getUserStatus (_zusr ::: tid ::: uid ::: _) = do
...
 -- | Request to provision a device on the legal hold service for a user
-requestDevice :: UserId ::: TeamId ::: UserId ::: JSON -> Galley Response
-requestDevice (zusr ::: tid ::: uid ::: _) = do
+requestDeviceH :: UserId ::: TeamId ::: UserId ::: JSON -> Galley Response
+requestDeviceH (zusr ::: tid ::: uid ::: _) = do
+  requestDevice zusr tid uid <&> \case
+    RequestDeviceSuccess -> empty & setStatus status201
+    RequestDeviceAlreadyPending -> empty & setStatus status204
+
+data RequestDeviceResult
```
arianvp: If it's a response I think it should be in `galley-types` no?

mheinzel:
> At least using OptionalJsonRequest seems to be the proper way of specifying this behaviour.

To clarify: The current approach DOES slightly change behaviour if the request body both
- contains something, but not a valid `TeamDeleteData`/`TeamMemberDeleteData` JSON structure
- is not looked at by the endpoint in that specific branch of execution (i.e. non-binding or non-existing team or not having necessary permission)

For non-binding teams, we should check what clients send. Do they ever make a request knowing that this is a non-binding team, so they leave out the body or maybe send an empty object? (especially with non-binding teams being a legacy thing)

The other cases seems like a client bug to me. I would rather find out that there is invalid JSON sent than getting an permission error. Also, if this happened, it would most likely also happen in the case of having proper permission, so I don't think this is hiding somewhere in the the current clients.

That being said, maybe we don't want to change behaviour to avoid the risk of breaking something? I'm happy to adjust this to delay throwing "invalid-request" until the body is actually used.

services/galley/src/Galley/API/LegalHold.hs
fisx: I also wonder if we can't just return `userLHStatus` below, and not introduce a new type in the first place?  That would shift exception handling around, not sure what I think of that either.

I say if in doubt, don't change this at all!  :)

services/galley/src/Galley/API/LegalHold.hs
```diff
@@ -204,10 +222,10 @@ approveDevice (zusr ::: tid ::: uid ::: connId ::: req ::: _) = do
   legalHoldAuthToken <- Client.getLegalHoldAuthToken uid mPassword
   LHService.confirmLegalHold clientId tid uid legalHoldAuthToken
   LegalHoldData.setUserLegalHoldStatus tid uid UserLegalHoldEnabled
-  -- TODO: send event at this point (see also:
-  -- https://github.com/wireapp/wire-server/pull/802#pullrequestreview-262280386)
-  pure empty
   where
+    -- TODO: send event at this point (see also:
```
fisx: This comment should go one line up, to where we're returning from this function.

services/galley/src/Galley/API/LegalHold.hs
mheinzel: Makes sense, there also some more types like this here and in brig.

Ideally, there would be some kind of `galley-api` package where we put those types and routes and also the mapping to status codes you see above. `galley-types` is kind of like that, but it doesn't seem like the status code mapping fits there are the moment.

services/galley/src/Galley/API/LegalHold.hs
mheinzel: ah, seems like ormolu pulls this into the `where`, so I have to put it two lines up to keep it from doing that. Or I add a `pure ()` below. :D

services/galley/src/Galley/API/Teams.hs
```diff
@@ -391,10 +407,20 @@ updateTeamMember (zusr ::: zcon ::: tid ::: req ::: _) = do
   -- push to all members (user is privileged)
   let pushPriv = newPush zusr (TeamEvent ePriv) $ privilegedRecipients
   for_ pushPriv $ \p -> push1 $ p & pushConn .~ Just zcon
-  pure empty
 
-deleteTeamMember :: UserId ::: ConnId ::: TeamId ::: UserId ::: Request ::: Maybe JSON ::: JSON -> Galley Response
-deleteTeamMember (zusr ::: zcon ::: tid ::: remove ::: req ::: _ ::: _) = do
+deleteTeamMemberH :: UserId ::: ConnId ::: TeamId ::: UserId ::: Request ::: Maybe JSON ::: JSON -> Galley Response
+deleteTeamMemberH (zusr ::: zcon ::: tid ::: remove ::: req ::: _ ::: _) = do
+  body <- fromJsonBody (JsonRequest req) -- TODO: can we demand it here?
```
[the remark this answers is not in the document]
mheinzel: Then let's do that.

services/galley/src/Galley/API/LegalHold.hs
mheinzel: So should we leave it for now?

services/galley/src/Galley/API/Update.hs
```diff
@@ -505,22 +529,28 @@ newMessage usr con cnv msg now (m, c, t) ~(toBots, toUsers) =
...
 
-updateConversationName :: UserId ::: ConnId ::: ConvId ::: JsonRequest ConversationRename -> Galley Response
-updateConversationName (zusr ::: zcon ::: cnv ::: req) = do
-  body <- fromJsonBody req
+updateConversationNameH :: UserId ::: ConnId ::: ConvId ::: JsonRequest ConversationRename -> Galley Response
+updateConversationNameH (zusr ::: zcon ::: cnv ::: req) = do
+  convRename <- fromJsonBody req
+  updateConversationName zusr zcon cnv convRename
+
+updateConversationName :: UserId -> ConnId -> ConvId -> ConversationRename -> Galley Response
```
fisx:
```suggestion
updateConversationName :: UserId -> ConnId -> ConvId -> ConversationRename -> Galley ()
```

services/galley/src/Galley/API/Update.hs
```diff
@@ -295,61 +305,66 @@ addCode (usr ::: zcon ::: cnv) = do
   mCode <- Data.lookupCode key ReusableCode
   case mCode of
     Nothing -> do
-      c <- generate cnv ReusableCode (Timeout 3600 * 24 * 365) -- one year TODO: configurable
-      Data.insertCode c
+      code <- generate cnv ReusableCode (Timeout 3600 * 24 * 365) -- one year TODO: configurable
```
fisx: thanks!  :)

services/galley/src/Galley/API/Update.hs
```diff
@@ -578,6 +600,8 @@ newMessage usr con cnv msg now (m, c, t) ~(toBots, toUsers) =
                   . set pushTransient (newOtrTransient msg)
            in (toBots, p : toUsers)
 
+--
```
fisx: ?

services/galley/src/Galley/API/Update.hs
```diff
@@ -279,6 +280,8 @@ updateConversationMessageTimer usr zcon cnv timerUpdate@(ConversationMessageTime
       pushEvent timerEvent users bots zcon
       pure timerEvent
 
+--
```
fisx: ?

services/galley/src/Galley/API/Update.hs
```diff
@@ -649,32 +653,38 @@ isTyping zusr zcon cnv typingData = do
         & pushRoute .~ RouteDirect
         & pushTransient .~ True
 
-addService :: JsonRequest Service -> Galley Response
-addService req = do
+addServiceH :: JsonRequest Service -> Galley Response
+addServiceH req = do
+  -- simple enough to not need a separate function
```
fisx:
```suggestion
```

(we do that in other places, and we don't want to comment on it every time.)

services/galley/src/Galley/API/Update.hs
```diff
@@ -649,32 +653,38 @@ isTyping zusr zcon cnv typingData = do
...
   Data.insertService =<< fromJsonBody req
   return empty
 
-rmService :: JsonRequest ServiceRef -> Galley Response
-rmService req = do
+rmServiceH :: JsonRequest ServiceRef -> Galley Response
+rmServiceH req = do
+  -- simple enough to not need a separate function
```
fisx:
```suggestion
```

services/galley/src/Galley/API/CustomBackend.hs
```diff
@@ -19,21 +19,27 @@ import Network.Wai.Utilities
...
   customBackend <- fromJsonBody req
+  -- simple enough to not need a separate function
   Data.setCustomBackend domain customBackend
   pure (empty & setStatus status201)
 
-internalDeleteCustomBackendByDomain :: EmailDomain ::: JSON -> Galley Response
-internalDeleteCustomBackendByDomain (domain ::: _) = do
+internalDeleteCustomBackendByDomainH :: EmailDomain ::: JSON -> Galley Response
+internalDeleteCustomBackendByDomainH (domain ::: _) = do
+  -- simple enough to not need a separate function
```
fisx:
```suggestion
```

PR #990, opened 2020-02-24
Cleanup handlers in gundeck, cannon

fisx:
https://github.com/zinfra/backend-issues/issues/1185

all in all this was definitely worth it, but i'm not sure i got 5207159 right, and it would be nice to have some explanation for the status code function.

### DIFF
--- a/services/gundeck/src/Gundeck/API.hs
+++ b/services/gundeck/src/Gundeck/API.hs
@@ -209,6 +209,8 @@ pushH (req ::: _) = do
 -- returned but the status code is set to 404.
 -- FUTUREWORK: We should change this behaviour as it's extremely confusing. We
 -- could kindly reject with a 400, and not event hit the database at all.
+-- This was introduced in
+-- https://github.com/zinfra/orlop/pull/30/commits/a358dfc1cb225c92066ea79db28c8824531ae231
 --
 -- If the 'since' parameter is present, and a notification 'since' is actually
 -- found in the database, this returns all the notifications since 'since'
--- a/services/gundeck/src/Gundeck/API.hs
+++ b/services/gundeck/src/Gundeck/API.hs
@@ -209,7 +209,9 @@ paginateH (_ ::: uid ::: sinceRaw ::: clt ::: size) = do
     isV1UUID :: UUID -> Maybe UUID
     isV1UUID u = if UUID.version u == 1 then Just u else Nothing
     --
-    -- the motivation of updStatus is far from clear to me, but this is how it's always been done.
+    -- FUTUREWORK: updStatus sets the status code to 404 in even if it returns notifications.
+    -- even if this is not a mistake, it certainly should be changed into something less
+    -- surprising.  (sync with clients before we change this!)
     updStatus :: Bool -> Response -> Response
     updStatus True = setStatus status404
     updStatus False = case since of
--- a/services/gundeck/src/Gundeck/API.hs
+++ b/services/gundeck/src/Gundeck/API.hs
@@ -195,6 +195,36 @@ pushH (req ::: _) = do
   ps <- fromJsonBody (JsonRequest req)
   empty <$ Push.push ps
 
+-- | Returns a list of notifications for given 'uid'
+--
+--
+-- Takes an optional parameter 'since' which is a V1 UUID, (which includes a
+-- timestamp).
+--
+-- If the parameter 'since' is omitted, all notifications of the user are
+-- returned. This is not recommended. (TODO: Ask client teams if they ever use
+-- this)
+--
+-- If the parameter 'since' fails to parse, all notifications of the user are
+-- returned but the status code is set to 404.
+-- FUTUREWORK: We should change this behaviour as it's extremely confusing. We
+-- could kindly reject with a 400, and not event hit the database at all.
+--
+-- If the 'since' parameter is present, and a notification 'since' is actually
+-- found in the database, this returns all the notifications since 'since'
+-- (exclusive of 'since' itself) and returns a status code 200.
+--
+-- If the 'since' parameter is present, and a notification 'since' is not found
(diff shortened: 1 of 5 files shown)


services/gundeck/src/Gundeck/API.hs
```diff
@@ -131,7 +137,94 @@ sitemap = do
...
+    since :: Maybe (Maybe NotificationId)
+    since = parseUUID <$> sinceRaw
+    --
+    parseUUID :: ByteString -> Maybe NotificationId
+    parseUUID = UUID.fromASCIIBytes >=> isV1UUID >=> return . Id
+    --
+    isV1UUID :: UUID -> Maybe UUID
+    isV1UUID u = if UUID.version u == 1 then Just u else Nothing
+    --
+    -- the motivation of updStatus is far from clear to me, but this is how it's always been done.
```
arianvp: I have the same! I was confused about this. I was just reading the docs and figuring out why   we return `404` but have notifications to serve.

services/gundeck/src/Gundeck/API.hs
```diff
@@ -131,7 +137,94 @@ sitemap = do
...
+  ps <- fromJsonBody (JsonRequest req)
+  empty <$ Push.push ps
+
+paginateH :: JSON ::: UserId ::: Maybe ByteString ::: Maybe ClientId ::: Range 100 10000 Int32 -> Gundeck Response
+paginateH (_ ::: uid ::: sinceRaw ::: clt ::: size) = do
+  Notification.PaginateResult gap page <- Notification.paginate uid (join since) clt size
+  pure . updStatus gap . json $ page
+  where
+    since :: Maybe (Maybe NotificationId)
+    since = parseUUID <$> sinceRaw
```
arianvp: how about :

```
since :: Maybe NotificationId
since = parseUUID =<< sinceRaw

```

arianvp: and think we can just inline  it in the call to `Notification.paginate` then.

join and fmap is the same as bind

services/gundeck/src/Gundeck/API.hs
fisx: my theory is that the client wanted something else, so we want to say "too bad", but at the same time we want to deliver something that may still be better than nothing.  would be nice if the author had left a comment about what's going on, though.

services/gundeck/src/Gundeck/API.hs
fisx: nope, won't work, because of the bizarre semantics of this end-point.  there is two uses of `since`; one joins, and the other checks for a parse error.  they do different things.

services/gundeck/src/Gundeck/API.hs
mheinzel: So there is one case that kind of makes sense to me: When `resultGap` is true (which means "a start ID ('since') has been given which could not be found", usually when a client missed notifications by being offline for too long), we return the `QueuedNotificationList`, but set status 404 to tell the client that something is missing. It's now up to the client to just use those notifications or force a slow sync. This could be a flag in the JSON response instead of a status code, but I see how it's used.

But the other case that confuses me more is that if the `since` query parameter is not a valid V1 UUID, we still fetch notifications and give a 404. So `since` could be any string at all and the result would not be distinguishable from having requested a valid UUID that was just too old. Do we do this for some kind of backwards compatibility reason? Why would that happen, so we can't assume this to be an invalid request?

Before changing this, we should have a discussion with someone who knows the history of this (@tiago-loureiro) or talk to the client teams.

mheinzel:
The Presence API was not touched? Except for this and the open question about fetching notifications, this looks good to me.

services/gundeck/src/Gundeck/API.hs
```diff
@@ -131,7 +137,96 @@ sitemap = do
...
+    parseUUID :: ByteString -> Maybe NotificationId
+    parseUUID = UUID.fromASCIIBytes >=> isV1UUID >=> return . Id
+    --
+    isV1UUID :: UUID -> Maybe UUID
+    isV1UUID u = if UUID.version u == 1 then Just u else Nothing
+    --
+    -- FUTUREWORK: updStatus sets the status code to 404 in even if it returns notifications.
+    -- even if this is not a mistake, it certainly should be changed into something less
+    -- surprising.  (sync with clients before we change this!)
+    updStatus :: Bool -> Response -> Response
```
arianvp: This is business logic, and business logic with `Bool -> Response -> Response` is very blind.  Can we bubble this out of the turning things into `json` ?

services/gundeck/src/Gundeck/API.hs
arianvp: I disagree with your reading of this code.    Both lead to a 404. There's just some confusing code duplication that hides this fact

services/gundeck/src/Gundeck/API.hs
mheinzel: I had a similar comment, but apparently linked to that particular commit, so it doesn't show up in the files view.

> I'm not sure about this part, since now you have logic here, in the wrapper function. The typed paginate function should look at `since` and make decisions.

If we can change the behaviour to throw a 400 when `since` doesn't parse as V1 UUID, IMO that can be handled in the wrapper, but not the distinction between `since` being provided or not.

services/gundeck/src/Gundeck/API.hs
arianvp: I have added to the documentation this weird behaviour.

Given we have an explicit test that asserts this, lets not change it for now. but I agree it's awkward https://github.com/zinfra/orlop/pull/30/commits/a358dfc1cb225c92066ea79db28c8824531ae231


I have a proposal to make this API a bit more elegant 

What if we _include_ the `since` notification in the response list? i.e. it's an inclusive bound instead of an exclusive bound. Then clients
can distinguish by the notification being there and not being there by ... checking that it's not there.

PR #993, opened 2020-02-26
Add remote identifier mapping to galley conversation-related tables

arianvp:
We will need this for federation

### DIFF
--- a/services/galley/schema/src/V38AddRemoteIdentifiers.hs
+++ b/services/galley/schema/src/V38AddRemoteIdentifiers.hs
@@ -6,24 +6,23 @@ import Text.RawString.QQ
 
 addRemoteIdentifiers :: Migration
 addRemoteIdentifiers = Migration 38 "Add remote identifiers to conversation related tables" $ do
-  -- User table is used to answer the question: What users are part of a
+  -- The user table answers the question:  What conversations am I a part of?
+  -- In the context of federation one now needs to know: Where is this
+  -- conversation located?
+  schema'
+    [r|
+      ALTER TABLE user ADD (
+        remote_conv        uuid,
+        remote_conv_domain text
+      );
+    |]
+  -- member table is used to answer the question: What users are part of a
   -- conversation?  In the context of federation it is now interesting to know:
   -- Are those users remote or local?
   schema'
     [r|
-      ALTER TABLE user ADD (
-        remote_user uuid,
-        remote_user_domain text
-      );
-    |]
-
-  -- The member table answers the question:  What conversations am I a member
-  -- of?  In the context of federation one now needs to know: Where is this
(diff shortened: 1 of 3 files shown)


services/galley/schema/src/V38AddRemoteIdentifiers.hs
```diff
@@ -0,0 +1,29 @@
...
+import Text.RawString.QQ
+
+addRemoteIdentifiers :: Migration
+addRemoteIdentifiers = Migration 38 "Add remote identifiers to conversation related tables" $ do
+  -- User table is used to answer the question: What users are part of a
+  -- conversation?  In the context of federation it is now interesting to know:
+  -- Are those users remote or local?
+  schema'
+    [r|
+      ALTER TABLE user ADD (
```
mheinzel: This table mostly gives us a list of conversations a user is in. I don't think we need to store that for remote users. We just need a list of users for a conversation.

services/galley/schema/src/V38AddRemoteIdentifiers.hs
```diff
@@ -0,0 +1,29 @@
...
+        remote_user_domain text
+      );
+    |]
+
+  -- The member table answers the question:  What conversations am I a member
+  -- of?  In the context of federation one now needs to know: Where is this
+  -- conversation located?
+  schema'
+    [r|
+      ALTER TABLE member ADD (
```
mheinzel: So those new fields stay `null` for all local users? And the existing fields will stay `null` for all remote users? And when we look up an opaque user ID, we query this table and look at which fields are `null` to determine whether the user is local or remote?

services/galley/schema/src/V38AddRemoteIdentifiers.hs
mheinzel: I think what you rather want might be to list the remote conversations a local user is in? So remote_conv instead of remote_user? Did you mix up the changes in the two tables?

services/galley/schema/src/V38AddRemoteIdentifiers.hs
mheinzel: Similarly to the other table, maybe you want remote_user to link remote users to a local conversation?

And then user uuid is null for remote users and remote_user* is null for local users?

services/galley/schema/src/V38AddRemoteIdentifiers.hs
mheinzel: Reading your comments properly, I really think you mixed up the two tables.

The `member` table maps a conversation to its members/users.
The `user` table maps a user to its conversations.

mheinzel:
This became part of https://github.com/wireapp/wire-server/pull/1070

PR #997, opened 2020-02-27
Move and create federation-related types

mheinzel:
As a preparation to start working on the endpoints, we want to make these types available.


libs/types-common/src/Data/Id.hs
```diff
@@ -47,14 +47,34 @@ data T
...
+
 type InvitationId = Id I
 
+-- | A local user ID
 type UserId = Id U
 
+-- | A UUID local to this backend, for which we know a mapping to a
+-- remote qualified user ID exists.
+-- These IDs should never leak to other backends or their clients.
+type MappedUserId = Id (Mp U)
```
mheinzel: I'm not sure this type will actually show up in our code, but I'd rather have it here just in case.

libs/types-common/src/Data/Id.hs
```diff
@@ -47,14 +47,34 @@ data T
 
 data STo
 
+data Mp a
```
arianvp: Personally not a fan of these abbreviated names, even if this was already the case originally.

libs/types-common/test/Test/Properties.hs
```diff
@@ -178,6 +186,17 @@ tests =
 roundtrip :: (EncodeWire a, DecodeWire a) => Tag' -> a -> Either String a
 roundtrip (Tag' t) = runGet (getWireField >>= decodeWire) . runPut . encodeWire t
 
+jsonRoundtrip ::
+  forall a.
+  (Arbitrary a, Typeable a, ToJSON a, FromJSON a, Eq a, Show a) =>
+  TestTree
+jsonRoundtrip = testProperty msg trip
+  where
+    msg = show $ typeOf (undefined :: a)
```
arianvp: You can do
```
Data.Typeable.typeRep (Proxy a)
```
or
```
Type.Reflection.typeRep @a 
```

these days

libs/types-common/src/Data/Id.hs
mheinzel: Absolutely agree. I will make this clearer.

libs/types-common/test/Test/Properties.hs
mheinzel: done

libs/types-common/src/Data/Qualified.hs
```diff
@@ -0,0 +1,73 @@
...
+        _qDomain :: Domain
+      }
+  deriving (Eq, Show, Generic)
+
+renderQualified :: (a -> Text) -> Qualified a -> Text
+renderQualified renderLocal (Qualified localPart domain) =
+  renderLocal localPart <> "@" <> domainText domain
+
+-- | The string to parse must contain exactly one @"@"@ to separate local part from domain.
+mkQualified :: (Text -> Either String a) -> Text -> Either String (Qualified a)
```
arianvp: add a TODO that we should clean this up

PR #1001, opened 2020-03-04
Update types of some galley endpoints to be federation-aware

mheinzel:
This should not change any behaviour, but shows the places where we need to make changes to make this work.

### DIFF
--- a/libs/galley-types/src/Galley/Types.hs
+++ b/libs/galley-types/src/Galley/Types.hs
@@ -313,10 +313,10 @@ data OtrFilterMissing
     OtrReportAllMissing
   | -- | Complain only about missing
     --      recipients who are /not/ on this list
-    OtrIgnoreMissing (Set UserId) -- TODO (mheinzel)
+    OtrIgnoreMissing (Set OpaqueUserId)
   | -- | Complain only about missing
     --      recipients who /are/ on this list
-    OtrReportMissing (Set UserId) -- TODO (mheinzel)
+    OtrReportMissing (Set OpaqueUserId)
 
 data NewOtrMessage
   = NewOtrMessage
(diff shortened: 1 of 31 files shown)


libs/galley-types/src/Galley/Types.hs
```diff
@@ -313,10 +313,10 @@ data OtrFilterMissing
     OtrReportAllMissing
   | -- | Complain only about missing
     --      recipients who are /not/ on this list
-    OtrIgnoreMissing (Set UserId)
+    OtrIgnoreMissing (Set UserId) -- TODO (mheinzel)
```
fisx: TODO what?

```suggestion
    OtrIgnoreMissing (Set UserId) -- TODO (mheinzel): use 'OpaqueUserId'
```

services/galley/src/Galley/API/Query.hs
```diff
@@ -89,6 +89,7 @@ getConversations zusr range size =
       | Data.isConvDeleted c = Data.deleteConversation (Data.convId c) >> pure False
       | otherwise = pure True
 
+-- TODO(mheinzel): in minimal API, but not essential
```
fisx: so, it should be removed from minimal api?  i'm not sure that's what this says.  maybe i don't have enough context to review this, in that case please disregard.

mheinzel: I removed the comment in fd86f75, was just a note to myself saying that this is part of the "minimal API" [1], so we will want to do it soonish, but it's not an endpoint we need to get a basic federation integration test working.

[1] the API implemented in AVES (https://aves.services.zinfra.io/swagger#!/), which is a good reference point, as it pretty much corresponds to what you need to make our clients work for basic use cases (but without teams).

libs/galley-types/src/Galley/Types.hs
mheinzel: You must be looking at an old version of the code, this TODO has been done in `4ee3cdd`.

PR #1004, opened 2020-03-05
stack snapshot 3.0.

fisx:
This does not change any dependencies from 2.2.

I broke 2.* a while ago by editing the files, and thus learned
that even editing typos in comments is a bad idea.  the easiest
way out is to abandon the old snapshot and clone it into a new
one with the same content.


akshaymankar:
Don't forget to trigger all the cache pipelines when this is merged.

fisx:
> Please don't forget to trigger all the cache pipelines when this is merged.

I don't think that's needed, since the dependencies of 2.2 and 3.0 are the same, and stack appears to be smart enough to figure that out.

PR #1005, opened 2020-03-06
Remove autoconnect functionality; deprecate end-point.

fisx:
https://github.com/zinfra/backend-issues/issues/557#issuecomment-595668101


services/brig/src/Brig/API.hs
```diff
@@ -1511,10 +1508,9 @@ verifyDeleteUserH (r ::: _) = do
   API.verifyDeleteUser body !>> deleteUserError
   return (setStatus status200 empty)
 
-onboardingH :: JSON ::: UserId ::: JsonRequest AddressBook -> Handler Response
-onboardingH (_ ::: uid ::: r) = do
-  ab <- parseJsonBody r
-  json <$> API.onboarding uid ab !>> connError
+-- | DEPRECATED / does nothing.
+onboardingH :: JSON ::: UserId ::: JsonRequest Value -> Handler Response
```
mheinzel: I think we still want to return `{"results":[],"auto-connects":[]}`, so clients don't choke on the response. 
So we could either leave the `MatchingResult` type and do `pure $ MatchingResult [] []` here, or just produce a JSON Value with the right fields.

Maybe this structure should also still be documented in Swagger.

services/brig/src/Brig/API.hs
```diff
@@ -813,12 +813,9 @@ sitemap o = do
...
     accept "application" "json"
       .&. header "Z-User"
-      .&. jsonRequest @AddressBook
+      .&. jsonRequest @Value
   document "POST" "onboardingV3" $ do
-    Doc.summary "Upload contacts and invoke matching. Returns the list of Matches"
-    Doc.body (Doc.ref Doc.addressBook) $ Doc.description "Address book"
-    Doc.returns (Doc.ref Doc.onboardingMatches)
-    Doc.response 200 "Matches" Doc.end
+    Doc.summary "[DEPRECATED]  the end-point does nothing.  (it used to upload contacts and invoke matching.)"
```
mheinzel: You can even mark the endpoint as deprecated in Swagger, look at how we do it for `deprecatedCompletePasswordResetH`.

mheinzel: (you can also move the endpoint to the "Deprecated" section)

fisx:
Thanks!  I am against keeping the swagger docs for the body type, but if you feel strongly about it let me know and I'll do it.

PR #1008, opened 2020-03-06
federation: resolve opaque IDs at the edges of galley

mheinzel:
This saves us some unnecessary ID resolution lookups.

Probably easiest to review each commit individually.

### DIFF
--- a/services/galley/src/Galley/API/Util.hs
+++ b/services/galley/src/Galley/API/Util.hs
@@ -252,6 +252,8 @@ getConversationAndCheckMembershipWithError ex zusr = \case
       throwM ex
     return c
 
+-- FUTUREWORK(federation): implement function to resolve IDs in batch
+
 -- | this exists as a shim to find and mark places where we need to handle 'OpaqueUserId's.
 resolveOpaqueUserId :: OpaqueUserId -> Galley (MappedOrLocalId Id.U)
 resolveOpaqueUserId (Id opaque) =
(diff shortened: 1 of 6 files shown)


services/galley/src/Galley/API/Util.hs
```diff
@@ -57,29 +57,30 @@ ensureConnectedOrSameTeam u uids = do
   sameTeamUids <- forM uTeams $ \team ->
     fmap (view userId) <$> Data.teamMembersLimited team uids
   -- Do not check connections for users that are on the same team
-  ensureConnected u (makeIdOpaque <$> uids \\ join sameTeamUids)
+  ensureConnectedToLocals u (uids \\ join sameTeamUids)
```
arianvp: with 'locals' we here mean the "local version of the remote id" right?

mheinzel: No, `ensureConnectedToLocals` only accepts local user IDs. It checks the connections of users on the same backend ("locals" or maybe "local users"?), just by looking at the database.

For now it is sufficient to call this function here, since `ensureConnectedOrSameTeam` is only called with local IDs, but maybe it's cleare to use `ensureConnected u (Local <$> _)` to show that in the future we will have to consider remote users, too? I could also bubble up the type by already making `ensureConnectedOrSameTeam` accept MappedOrLocalId and converting at the call sites. What do you think?

arianvp: Add a note to future us is probably sufficient for now

PR #1011, opened 2020-03-11
types-common: write unit tests for Data.Qualified

mheinzel:
Should have been part of #997 already.

### DIFF
--- a/libs/types-common/test/Test/Qualified.hs
+++ b/libs/types-common/test/Test/Qualified.hs
@@ -22,29 +22,15 @@ tests =
     "Qualified"
     [ testGroup
         "serialization"
-        [ testCase "render foo@bar.com" $
-            assertEqual
-              "foo@bar.com"
-              "foo@bar.com"
-              ( renderQualifiedHandle
-                  ( Qualified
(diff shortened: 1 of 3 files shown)


libs/types-common/test/Test/Qualified.hs
```diff
@@ -0,0 +1,54 @@
...
+
+jsonRoundtrip ::
+  forall a.
+  (Arbitrary a, Typeable a, ToJSON a, FromJSON a, Eq a, Show a) =>
+  TestTree
+jsonRoundtrip = testProperty msg trip
+  where
+    msg = "jsonRoundTrip @(" <> show (typeRep @a) <> ")"
+    trip (v :: a) =
+      counterexample (show $ toJSON v) $
```
fisx:
```suggestion
      counterexample (show v) $
```

(I suspect you have a reason for not doing that, but I'd like to know.

mheinzel: If the test fails, the `===` already causes both sides to be printed using `Show`, which is exactly that `v`. This additionally shows us what the intermediate JSON value looks like.

PR #1013, opened 2020-03-11
Update types of some brig endpoints to be federation-aware

mheinzel:
One thing I'm still thinking about (and left some TODOs for):

In Galley when encounter a remote ID where we can't handle it yet, we throw a HTTP 501 "federation-not-implemented". Here, it's not as easy since we throw errors into different sum types for different routes (ClientError, UserError, ConnectionError), which we would all have to extend with a case for federation. I'm currently leaning towards just doing that (instead of just treating remote IDs as non-existent ones).

### DIFF
--- a/libs/types-common/src/Data/Domain.hs
+++ b/libs/types-common/src/Data/Domain.hs
@@ -13,16 +13,17 @@ import Imports
 import Test.QuickCheck (Arbitrary (arbitrary))
 import qualified Test.QuickCheck as QC
 
--- | Following [RFC-1035](https://www.ietf.org/rfc/rfc1035.txt), Section 2.3.1,
+-- | A Fully Qualified Domain Name (FQDN).
+--
+-- Following [RFC-1035](https://www.ietf.org/rfc/rfc1035.txt), Section 2.3.1,
 -- except for:
 -- * not allowing a space @" "@
--- * TODO accepting digits as the first letter of labels (except for the last label)
--- * TODO requiring at least two labels
--- * not only must labels be 63 characters or less, the whole domain must be 253 characters or less
+-- * accepting digits as first letter of labels (except for the last label, not allowing IPs)
+-- * requiring at least two labels
+-- * not only must labels be 63 chars or less, the whole domain must be 253 chars or less
 --
--- TODO: rename to Fqdn
---
--- <domain> ::= <label> | <domain> "." <label>
+-- <domain> ::= <label> "." <tld> | <label> "." <domain>
+-- <tld> ::= <letter> [ [ <ldh-str> ] <let-dig> ]
 -- <label> ::= <let-dig> [ [ <ldh-str> ] <let-dig> ]
 -- <ldh-str> ::= <let-dig-hyp> | <let-dig-hyp> <ldh-str>
 -- <let-dig-hyp> ::= <let-dig> | "-"
@@ -48,9 +49,13 @@ instance FromByteString Domain where
 domainParser :: Atto.Parser Domain
 domainParser = do
   parts <- domainLabel `Atto.sepBy1` Atto.char '.'
+  when (length parts < 2) $
+    fail "Invalid domain name: cannot be dotless domain"
+  when (isDigit (BS.Char8.head (last parts))) $
+    fail "Invalid domain name: last label cannot start with digit"
   let bs = BS.intercalate "." parts
   when (BS.length bs > 253) $
-    fail "Invalid domain: too long"
+    fail "Invalid domain name: too long"
   case Text.E.decodeUtf8' bs of
     Left err -> fail $ "Invalid UTF-8 in Domain: " <> show err
     Right txt -> pure . Domain $ Text.toCaseFold txt
@@ -87,13 +92,17 @@ newtype DomainText
 instance Arbitrary DomainText where
   arbitrary = DomainText <$> domain
     where
-      -- <domain> ::= <label> | <domain> "." <label>
+      -- <domain> ::= <label> "." <tld> | <label> "." <domain>
       domain =
         QC.oneof
-          [ label,
-            conc [domain, pure ".", label]
+          [ conc [label, pure ".", tldLabel],
+            conc [label, pure ".", domain]
           ]
           `QC.suchThat` ((<= 255) . Text.length)
+      -- <tld> ::= <letter> [ [ <ldh-str> ] <let-dig> ]
+      tldLabel =
+        conc [letter, opt (conc [opt ldhStr, letDig])]
+          `QC.suchThat` ((<= 63) . Text.length)
       -- <label> ::= <let-dig> [ [ <ldh-str> ] <let-dig> ]
       label =
         conc [letDig, opt (conc [opt ldhStr, letDig])]
@@ -112,11 +121,15 @@ instance Arbitrary DomainText where
             (5, letDig) -- fewer hyphens
           ]
       -- <let-dig> ::= <letter> | <digit>
-      -- <letter> ::= any one of the 52 alphabetic characters A through Z in
-      -- upper case and a through z in lower case
-      -- <digit> ::= any one of the ten digits 0 through 9
       letDig =
-        Text.singleton <$> QC.elements (['a' .. 'z'] <> ['A' .. 'Z'] <> ['0' .. '9'])
+        QC.oneof [letter, digit]
+      -- <letter> ::= any one of the 52 alphabetic characters A through Z in
+      letter =
+        Text.singleton <$> QC.elements (['a' .. 'z'] <> ['A' .. 'Z'])
+      -- <digit> ::= any one of the ten digits 0 through 9
+      digit =
+        Text.singleton <$> QC.elements ['0' .. '9']
+      -- upper case and a through z in lower case
       -- helpers
       conc :: [QC.Gen Text] -> QC.Gen Text
       conc = fmap Text.concat . sequenceA
--- a/libs/types-common/src/Data/Domain.hs
+++ b/libs/types-common/src/Data/Domain.hs
@@ -9,9 +9,10 @@ import qualified Data.ByteString.Char8 as BS.Char8
 import Data.ByteString.Conversion
 import qualified Data.Text as Text
 import qualified Data.Text.Encoding as Text.E
-import Imports
+import Imports hiding (isAlphaNum)
 import Test.QuickCheck (Arbitrary (arbitrary))
 import qualified Test.QuickCheck as QC
+import Util.Attoparsec (takeUpToWhile)
 
 -- | A Fully Qualified Domain Name (FQDN).
 --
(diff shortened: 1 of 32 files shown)


services/brig/src/Brig/API.hs
```diff
@@ -1133,30 +1157,33 @@ getUserNameH (_ ::: self) = do
     Just n -> json $ object ["name" .= n]
     Nothing -> setStatus status404 empty
 
-listUsersH :: JSON ::: UserId ::: Either (List UserId) (List Handle) -> Handler Response
+listUsersH :: JSON ::: UserId ::: Either (List OpaqueUserId) (List Handle) -> Handler Response
```
arianvp: Should the handles be qualified here?

mheinzel: Optionally qualified, right? Good catch.

I don't think we are using this anywhere right now (we are definitely not in web) and I'm not sure what the use case is, but I'll fix it.

libs/types-common/src/Data/IdMapping.hs
```diff
@@ -17,6 +17,11 @@ opaqueIdFromMappedOrLocal = \case
   Local localId -> makeIdOpaque localId
   Mapped IdMapping {idMappingLocal} -> makeMappedIdOpaque idMappingLocal
 
+partitionMappedOrLocalIds :: Foldable f => f (MappedOrLocalId a) -> ([Id a], [IdMapping a])
+partitionMappedOrLocalIds = foldMap $ \case
+  Mapped mapping -> (mempty, [mapping])
```
arianvp: neat

libs/types-common/src/Data/Qualified.hs
```diff
@@ -4,18 +4,54 @@ module Data.Qualified where
...
+
+qualified :: Qualified a -> OptionallyQualified a
+qualified = Right
+
+{-
+TODO: do it properly
+Either is a hack, conceptually we want this:
+
+data OptionallyQualified a
+  = OptionallyQualified
```
arianvp: Do you still want to resolve htis before merging?

services/brig/src/Brig/API/Client.hs
```diff
@@ -101,17 +124,33 @@ rmClient u con clt pw =
...
       lift $ execDelete u (Just con) client
 
-claimPrekey :: UserId -> ClientId -> AppIO (Maybe ClientPrekey)
-claimPrekey u c = do
+claimPrekey :: MappedOrLocalId Id.U -> ClientId -> AppIO (Maybe ClientPrekey)
+claimPrekey u c = case u of
+  Local localUser ->
+    claimLocalPrekey localUser c
+  Mapped _ ->
+    -- TODO: error?
```
arianvp: no not an error right? we contact the other backend and ask it to claim a prekey for us no?

services/brig/src/Brig/API/Client.hs
```diff
@@ -101,17 +124,33 @@ rmClient u con clt pw =
...
     pk@(Just _) -> return pk
 
-claimPrekeyBundle :: UserId -> AppIO PrekeyBundle
-claimPrekeyBundle u = do
+claimPrekeyBundle :: MappedOrLocalId Id.U -> AppIO PrekeyBundle
+claimPrekeyBundle = \case
+  Local localUser ->
+    claimLocalPrekeyBundle localUser
+  Mapped IdMapping {idMappingLocal} ->
+    -- TODO: error?
```
arianvp: No, same as above

libs/types-common/src/Data/Qualified.hs
mheinzel: Yes, that's the main things that's missing.

services/brig/src/Brig/API/Client.hs
mheinzel: Only until we implement the logic. In Galley, we fail with some `federation-not-implemented` in these cases (a remote ID being used, but the function not being prepared for it).

libs/types-common/src/Data/Qualified.hs
mheinzel: Implemented now.

services/brig/src/Brig/API/Client.hs
mheinzel: I removed all of these now. Still not 100% happy with it, but seems reasonable, since it will just behave the same as if a local user was not found.

mheinzel:
I rebases and implemented a parser for `OptionallyQualified Handle` and IDs.

To do that, I also rewrote the parsers for handles, IDs and domains, to make them composable (which they aren't if they always try to consume the whole input). Then I based all other validation and smart constructors on these Attoparsec parsers. I added some roundtrip tests, but it would still be nice if someone could have a look at these parsers specifically.

libs/types-common/src/Data/Domain.hs
```diff
@@ -43,12 +74,52 @@ instance FromJSON Domain where
...
+      -- <domain> ::= <label> | <domain> "." <label>
+      domain =
+        QC.oneof
+          [ label,
+            conc [domain, pure ".", label]
+          ]
+          `QC.suchThat` ((<= 255) . Text.length)
+      -- <label> ::= <let-dig> [ [ <ldh-str> ] <let-dig> ]
+      label =
+        conc [letDig, opt (conc [opt ldhStr, letDig])]
```
arianvp: We had this cool parser combinator framework at UU that would  generate a parser, pretty printer, and quickcheck generator for each  AST that you defined.  I miss it now.  (As this is just the same as the parser code, but in a different applicative :P; wonder if we can abuse that fact somehow (Free applicative?) )

Ohwell not that important; just an observation

mheinzel: I remember seeing something like that and yes, it's nice.

The parser doesn't adhere to the grammar as closely as the generator here, so testing them against each other gives me some confidence. Although I only know the parser is not too restrictive, but it could still be too permissive. :thinking:

libs/types-common/src/Data/Domain.hs
```diff
@@ -20,20 +37,34 @@ domainText :: Domain -> Text
...
 instance FromByteString Domain where
-  parser = do
-    bs <- Atto.takeByteString
-    case mkDomain =<< first show (Text.E.decodeUtf8' bs) of
-      Left err -> fail ("Failed parsing ByteString as Domain: " <> err)
-      Right domain -> pure domain
+  parser = domainParser
+
+domainParser :: Atto.Parser Domain
+domainParser = do
```
fisx: heh, you couldn't help yourself.  :)

(i assume this is because the domain validation in the email package wasn't as thorough as this one.)

libs/types-common/src/Data/Id.hs
```diff
@@ -112,20 +114,30 @@ newtype Id a
...
 
 instance Read (Id a) where
   readsPrec n = map (\(a, x) -> (Id a, x)) . readsPrec n
 
 instance FromByteString (Id a) where
   parser = do
-    x <- takeByteString
-    case fromASCIIBytes x of
+    x <-
+      -- we only want the matching ByteString
```
fisx: lacking context here.  do you mean the matching *part* of the bytestring?  if so that sounds odd to me.

also, one-letter variable name?  (sorry :))

libs/types-common/test/Test/Domain.hs
```diff
@@ -0,0 +1,27 @@
...
+tests :: TestTree
+tests =
+  testGroup
+    "Domain"
+    [ testGroup "serialization" testDomainSerialization
+    ]
+
+testDomainSerialization :: [TestTree]
+testDomainSerialization =
+  [ testProperty "Arbitrary DomainText generates valid domains" $
```
fisx: It's good to phrase it like this, and not "mkDomain works for *all* inputs".  :)

fisx: You could run `mkDomain` on some invalid (and valid) domains given as string literals, if you really wanted to.

libs/types-common/src/Data/Domain.hs
mheinzel: Yes, it allows some extra things like comments (`(comment)example.com`) and bracketed IP addresses (`[1.1.1.1.]`).

libs/types-common/test/Test/Domain.hs
mheinzel: I'll do that, similar to the tests for handles.

libs/types-common/src/Data/Id.hs
mheinzel: Yes, only the matching part. This is because if we compose the parser with e.g. the domain parser to parse a qualified ID like `61a73a52-e526-4892-82a9-3d638d77629f@example.com`, the parser for the UUID should not consume that whole string, but leave the `@example.com` for the next parser.

I'll add that explanation.

libs/types-common/src/Data/Handle.hs
```diff
@@ -23,37 +31,43 @@ newtype Handle
...
-    parseOnly handle t
+isValidHandle = isRight . parseHandleEither
+
+parseHandleEither :: Text -> Either String Handle
+parseHandleEither = Atto.parseOnly (handleParser <* Atto.endOfInput) . Text.E.encodeUtf8
+
+handleParser :: Atto.Parser Handle
+handleParser = do
+  bs <- matching (Atto.count 2 handleChar *> Atto.skipMany handleChar)
+  when (BS.length bs > 256) $
```
arianvp: Why not use count here like before?

libs/types-common/src/Data/Id.hs
```diff
@@ -112,20 +114,30 @@ newtype Id a
...
+        void $ Atto.count 8 hexDigit <* Atto.char '-'
+        void $ Atto.count 4 hexDigit <* Atto.char '-'
+        void $ Atto.count 4 hexDigit <* Atto.char '-'
+        void $ Atto.count 4 hexDigit <* Atto.char '-'
+        void $ Atto.count 12 hexDigit
+    case UUID.fromASCIIBytes x of
       Nothing -> fail "Invalid UUID"
       Just ui -> return (Id ui)
+    where
+      matching = fmap fst . Atto.match
```
arianvp: TIL Atto.match;  I did some nasty things in hscim instead.

libs/types-common/src/Data/Handle.hs
mheinzel: No fundamental reason. `count n (optional _)` seems like it would always spin `n` times, even if there is no valid handleChar anymore. In addition, it will allocate a bunch of unnecessary `Maybe`s. Maybe the error message is also clearer? On the other hand, `skipMany` will not stop after 256 chars, so not sure which one is better. :man_shrugging:

libs/types-common/src/Data/Domain.hs
mheinzel: We now also decided to stick to FQDNs, requiring at least two labels, so I'll update the parser again.

services/brig/src/Brig/API.hs
mheinzel: done

libs/types-common/src/Data/Domain.hs
mheinzel: done.

libs/types-common/test/Test/Domain.hs
mheinzel: added the tests, done.

mheinzel:
All feedback has been incorporated and we require multiple domain labels now (not allowing dotless domains).

I'm rebasing now. Anyone wants to take another look in the meantime?

mheinzel:
I rebase and needed some changes since the cabal files have now been checked in. If CI passed, I'll merge.

PR #1015, opened 2020-03-12
federation-aware joinConversationById

mheinzel:
A bit hacky, we'll do it properly soon.


mheinzel:
This PR is getting bigger and bigger, still with lots of TODOs. It also makes some nontrivial changes to existing code, mainly making some IDs in galley `Event`s opaque and serving our galley webapp via servant by wrapping the existing one.

Maybe we can split the PR a bit? I don't want to keep rebasing this for days and then merge a 2000+ lines change.

mheinzel:
I already merged some of these changes as part of smaller PRs. The rest will also follow in multiple self-contained PRs.

PR #1019, opened 2020-03-17
Federation intergration tests

arianvp:
Simply starts up some services as duplicate (Should eventually do that for all of them) one on `127.0.0.1` and one on `127.0.0.2`

I'm not sure if hacking it in the `integration.sh` script like now is very scalable, but it was the quickest thing that I got working.

### DIFF
--- a/services/brig/brig.integration2.yaml
+++ b/services/brig/brig.integration2.yaml
@@ -1,5 +1,5 @@
 brig:
-  host: 127.0.0.1
+  host: 127.0.0.2
   port: 8082
 
 cassandra:
@@ -13,15 +13,15 @@ elasticsearch:
   index: directory_test
 
 cargohold:
-  host: 127.0.0.1
+  host: 127.0.0.2
   port: 8084
 
 galley:
-  host: 127.0.0.1
+  host: 127.0.0.2
   port: 8085
 
 gundeck:
-  host: 127.0.0.1
+  host: 127.0.0.2
   port: 8086
 
 # You can set up local SQS/Dynamo running e.g. `../../deploy/dockerephemeral/run.sh`
@@ -85,11 +85,11 @@ emailSMS:
       forgot: https://wire.com/forgot/
       support: https://support.wire.com/
   user:
-    activationUrl: http://127.0.0.1:8080/activate?key=${key}&code=${code}
-    smsActivationUrl: http://127.0.0.1:8080/v/${code}
-    passwordResetUrl: http://127.0.0.1:8080/password-reset/${key}?code=${code}
-    invitationUrl: http://127.0.0.1:8080/register?invitation_code=${code}
-    deletionUrl: http://127.0.0.1:8080/users/delete?key=${key}&code=${code}
+    activationUrl: http://127.0.0.2:8080/activate?key=${key}&code=${code}
+    smsActivationUrl: http://127.0.0.2:8080/v/${code}
+    passwordResetUrl: http://127.0.0.2:8080/password-reset/${key}?code=${code}
+    invitationUrl: http://127.0.0.2:8080/register?invitation_code=${code}
+    deletionUrl: http://127.0.0.2:8080/users/delete?key=${key}&code=${code}
 
   provider:
     homeUrl: https://provider.localhost/
@@ -131,7 +131,7 @@ optSettings:
   setNexmo: test/resources/nexmo-credentials.yaml
   # setStomp: test/resources/stomp-credentials.yaml
   setUserMaxConnections: 16
-  setCookieDomain: 127.0.0.1
+  setCookieDomain: 127.0.0.2
   setCookieInsecure: true
   setUserCookieRenewAge: 2
   setUserCookieLimit: 5
--- a/services/galley/test/integration/Main.hs
+++ b/services/galley/test/integration/Main.hs
@@ -78,8 +78,8 @@ main = withOpenSSL $ runTests go
       let c = mkRequest (cannon iConf)
       let federatedBrig1' = mkRequest (federatedBrig1 iConf)
       let federatedBrig2' = mkRequest (federatedBrig2 iConf)
-      let federatedBackend1' = mkRequest (federatedBackend1 iConf)
-      let federatedBackend2' = mkRequest (federatedBackend2 iConf)
+      let federator1' = mkRequest (federator1 iConf)
+      let federator2' = mkRequest (federator2 iConf)
       let q = queueName gConf
       let e = endpoint gConf
       let convMaxSize = maxSize gConf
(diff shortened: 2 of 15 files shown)


services/galley/test/integration/Main.hs
```diff
@@ -72,31 +68,30 @@ main = withOpenSSL $ runTests go
               (pathsConsistencyCheck . treeToPaths . compile $ Galley.API.sitemap),
           API.tests setup
         ]
+    getOpts :: FilePath -> FilePath -> IO TestSetup
     getOpts gFile iFile = do
       m <- newManager tlsManagerSettings {managerResponseTimeout = responseTimeoutMicro 300000000}
-      let local p = Endpoint {_epHost = "127.0.0.1", _epPort = p}
-      gConf <- handleParseError =<< decodeFileEither gFile
-      iConf <- handleParseError =<< decodeFileEither iFile
-      -- FUTUREWORK: we don't support process env setup any more, so both gconf and iConf
```
arianvp: I took the oppurtunity to solve this FUTUREWORK section.  I'm not sure  if there is a ticket open for "unifying the integration test suites into a single library". I don't fully understand what that would mean :)

arianvp:
`make i` in `galley` now just hangs forever. I don't understand what I broke that caused this to happen

arianvp:
Added some extra debug logs. seems to get stuck in the `all_services_up` check.

arianvp:
Aha `federator` doesn't respond to `/i/status` yet and hence it is hanging. got it :)

services/galley/test/integration/TestSetup.hs
```diff
@@ -35,13 +39,20 @@ type BrigR = Request -> Request
...
 data IntegrationConfig
   = IntegrationConfig
       -- internal endpoints
       { galley :: Endpoint,
         brig :: Endpoint,
         cannon :: Endpoint,
-        provider :: LegalHoldConfig
+        provider :: LegalHoldConfig,
+        federatedBrig1 :: Endpoint, -- An escape hatch into brig on the other side. Dirty. to create users!
+        federatedBrig2 :: Endpoint, -- An escape hatch into brig on the other side. Dirty. to create users!
```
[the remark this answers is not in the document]
mheinzel: Also, I'd probably move these next to existing `brig` field, but not sure.

services/galley/test/integration/what-to-test
```diff
@@ -0,0 +1,7 @@
+For federation, what do we want to test?
+
+Lets start with conversations:
+
+1. conversation list might have remote conversations
```
mheinzel: Is the plan to create these conversation in the DB directly or do we want to go through `joinConversationById`?

services/integration.yaml
```diff
@@ -3,6 +3,24 @@ brig:
...
+federatedBrig1:
+  host: 127.0.0.2
+  port: 8082
+
+federatedBrig2:
+  host: 127.0.0.3
+  port: 8082
+
+federator1:
+  host: 127.0.02
```
mheinzel: Typo.

AIso, is there no federator on `127.0.0.1`? We would need it, if the main instance is supposed to participate in the federated tests.

arianvp: Good catch and good catch! :)

services/integration.yaml
```diff
@@ -3,6 +3,24 @@ brig:
...
+  port: 8097
+
+federator2:
+  host: 127.0.03
```
arianvp: same typo

jschaul:
> Aha `federator` doesn't respond to `/i/status` yet and hence it is hanging. got it :)

Perhaps that endpoint could be added to federator also? The business logic isn't too hard :)

mheinzel:
> Perhaps that endpoint could be added to federator also? The business logic isn't too hard :)

It's easy to add and we should do it.

(still, for our normal integration tests, we probably shouldn't start it and and thus not wait for it)

services/brig/brig.integration2.yaml
```diff
@@ -0,0 +1,164 @@
+brig:
```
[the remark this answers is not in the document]
arianvp: Fully agree with your observation to not have these multiple config files.  Though I have a slightly different suggestion to solve this. How about we use either dhall or jsonnet to template this out?

I'll add it to the TODO list for tomorrow to get rid of this abundance of configs

arianvp:
> Instead of making brig's and galley's and gundeck's integration tests all federation aware by default (..sic..), how about moving this test code into the federator service? Since some of the code is currently commented with

Hmm I'm not sure if I agree with this.   As far as I understand now `federator` will do two very specific things:  Route outgoing requests to the right backend, and route incoming federation requests to the right service. But most of the business logic of deciding whether a user is local or global will have to live in the respective services that know about the domain mapping.   This is the path that we have taken with  https://github.com/wireapp/wire-server/pull/1001 and https://github.com/wireapp/wire-server/pull/1013

I think it is important to test these things and have those tests live near the business logic
that implements it.

There will definitely be tests that make sense from the perspective of the federator.  "Hey I have a message, can I forward it to the respective service without it erroring out" .
But anything more detailed like  "I received a federated message; did I receive it on the websocket"  or "I am sending a message to a conversation through the galley API, but the UUID might actually be not on our current backend, but must be resolved to a qualified UUID" is a decision that is fully made in galley at the moment and hence that integration test should _not_ live in federator I think.

Given that we're not changing the client APIs,  Galley is still fulfilling the contract of conversations to the clients, and hence the decision to decide whether a conversation ID is local _or_ must be forwarded to `federator` because it is remote is in galley.

services/brig/brig.integration2.yaml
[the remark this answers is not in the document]
arianvp: though maybe a simple template suffices here. I just do not want us to have 20 different ways for templating YAML

services/brig/brig.integration2.yaml
```diff
@@ -0,0 +1,164 @@
...
+    host: 127.0.0.1
+    port: 9042
+  keyspace: brig_test
+
+elasticsearch:
+  url: http://127.0.0.1:9200
+  index: directory_test
+
+cargohold:
+  host: 127.0.0.2
```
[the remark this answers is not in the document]
arianvp: Because they really should only talk to the services in their own federation. They're just not up and running yet (need to add more lines to `intergration.sh` for that)

jschaul:
> test location...

I would like to keep the possibility to test certain things without needing multiple backends running locally. I.e. I want to continue to be able to run the existing integration tests without requiring more RAM/CPU. The same way that unit tests in federation-util should not have a dependency on the internet, testing the business logic of cookies and access tokens should not need federatedBrig2 to be started.

Adding the federation tests to galley/brig/gundeck may make this usecase difficult to support, so I thought grouping them inside the federator (or, if you prefer, in an entirely different location that is centralized) might make sense. If you can come up with a solution to support the usecase of testing everything that is not federation-specific and thus does not need more than one backend up to run with a makefile target, then I don't mind so much where the code lives.
 
Both galley and brig do some business logic regarding teams. There are also tests in galley and tests in brig that test this functionality. To me, sprinkling this across two services makes it more confusing than if all the team-related tests were in one place. With federation, testing one conceptual thing (does local/global UUID mapping work properly?) might be much more confusing if it's sprinkled across three or four different services, than if all these tests live in one place. Happy to discuss this more and get convinced otherwise :)

PR #1024, opened 2020-03-23
Add hie.yaml

akshaymankar:
`hie.yaml` is useful when using [haskell-ide-engine](https://github.com/haskell/haskell-ide-engine). This file needs to be regenerated whenever a new package, library or executable is added to the project. I have added a hacky script to do so, I have also added it as a make target. To regenerate the file run this:
```
make -B hie.yaml
```


mheinzel:
What would be the best way to avoid this going out of sync? Or do we rely on the HIE users to update it when they notice something is wrong?

akshaymankar:
For now, I think it makes sense to rely on HIE users for now. As I think it is just me and I don't want to force others to commit this file.  Or we can gitignore the file and the HIE users can regenerate it as required.

akshaymankar:
I think it makes sense to not commit the file, I have git-ignored it.

Makefile
```diff
@@ -181,3 +181,6 @@ db-reset:
 
 libzauth:
 	$(MAKE) -C libs/libzauth install
+
+hie.yaml:
```
mheinzel: I think this will refuse to re-generate the file once it's already there. Maybe it could be a PHONY target (instead of listing all relevant files as dependencies)?

PR #1025, opened 2020-03-23
Add comments about which endpoints send which events to clients

mheinzel:
Most of the comments were just used to collect the information and will be removed again.
I'm planning to just leave the ones on the endpoints themselves (could even make it part of the Swagger doc there), still finalizing the format.

I will also leave the comments with the corresponding handler functions on `Intra` calls.

### DIFF
--- a/services/brig/src/Brig/API.hs
+++ b/services/brig/src/Brig/API.hs
@@ -205,7 +205,8 @@ sitemap o = do
       .&. query "base_url"
   ---
 
-  -- If the user is expired, it will be removed asynchronously.
+  -- If the user is ephemeral and expired, it will be removed asynchronously,
+  -- see 'Brig.API.User.userGC'.
   -- This leads to the following events being sent to clients:
   -- - UserDeleted event to contacts of the user
   -- - MemberLeave event to members for all conversations the user was in (via galley)
@@ -220,7 +221,8 @@ sitemap o = do
     Doc.errorResponse userNotFound
   ---
 
-  -- If the user is expired, it will be removed asynchronously.
+  -- If the user is ephemeral and expired, it will be removed asynchronously,
+  -- see 'Brig.API.User.userGC'.
   -- This leads to the following events being sent to clients:
   -- - UserDeleted event to contacts of the user
   -- - MemberLeave event to members for all conversations the user was in (via galley)
@@ -270,7 +272,8 @@ sitemap o = do
     Doc.errorResponse handleNotFound
   ---
 
-  -- If the user is expired, it will be removed asynchronously.
+  -- If the user is ephemeral and expired, it will be removed asynchronously,
+  -- see 'Brig.API.User.userGC'.
   -- This leads to the following events being sent to clients:
   -- - UserDeleted event to contacts of the user
   -- - MemberLeave event to members for all conversations the user was in (via galley)
--- a/services/brig/src/Brig/API.hs
+++ b/services/brig/src/Brig/API.hs
@@ -83,7 +83,7 @@ sitemap o = do
   -- if others didn't join a connect conversation with self before:
   --   via galley: ConvConnect EdConnect event to self
   --   via galley: MemberJoin EdMembersJoin event to you and other
-  post "/i/users/:uid/auto-connect" (continue (autoConnectH N E)) $
+  post "/i/users/:uid/auto-connect" (continue autoConnectH) $
     accept "application" "json"
       .&. capture "uid"
       .&. opt (header "Z-Connection")
@@ -91,7 +91,7 @@ sitemap o = do
   -- UserActivated event to self, if user is team user
   -- UserActivated event to self, if user has SSO ID
   -- UserIdentityUpdated event to self, if email or phone get activated (should always happen)
-  post "/i/users" (continue (createUserNoVerifyH E)) $
+  post "/i/users" (continue createUserNoVerifyH) $
     accept "application" "json"
       .&. jsonRequest @NewUser
   put "/i/self/email" (continue changeSelfEmailNoSendH) $
@@ -100,7 +100,7 @@ sitemap o = do
   -- when internal event is handled asynchronously:
   --   UserDeleted event to contacts
   --   via galley: MemberLeave EdMembersLeave event to members for all conversations the user was in
-  delete "/i/users/:uid" (continue (deleteUserNoVerifyH N E)) $
+  delete "/i/users/:uid" (continue deleteUserNoVerifyH) $
     capture "uid"
   get "/i/users/connections-status" (continue deprecatedGetConnectionsStatusH) $
     query "users"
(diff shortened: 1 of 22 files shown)


mheinzel:
This is ready from my side. I didn't put the information into the swagger docs, as we use Servant based ones in some places and it seems like there is some extra work to add notes to an endpoint there (plus those would have been even further away from the code).

I also did not add information for the teams API in this PR, as we don't need it for federation yet and I want to get started on other tasks first. The starting point then would be to go to `5bae45f7f9815e752d8c17f938514406d405f9b6`, look for `FUTUREWORK` comments in `services/galley/src/Galley/API/Teams.hs` and propagate the information from there.

services/brig/src/Brig/API.hs
```diff
@@ -72,17 +72,31 @@ sitemap o = do
...
+  -- This endpoint can lead to the following events being sent to clients:
+  -- - UserActivated event to created user, if it is a team invitation or user has an SSO ID
+  -- - UserIdentityUpdated event to created user, if email or phone get activated
   post "/i/users" (continue createUserNoVerifyH) $
     accept "application" "json"
       .&. jsonRequest @NewUser
   put "/i/self/email" (continue changeSelfEmailNoSendH) $
     header "Z-User"
       .&. jsonRequest @EmailUpdate
+  -- This endpoint will asynchronously lead to the following events being sent to clients:
```
arianvp: What does asynchronously mean here? What if that fails half way through?

services/brig/src/Brig/API.hs
```diff
@@ -153,15 +169,24 @@ sitemap o = do
   post "/i/clients" (continue internalListClientsH) $
     accept "application" "json"
       .&. jsonRequest @UserSet
+  -- This endpoint can lead to the following events being sent to clients:
+  -- - ClientAdded event to the user
+  -- - ClientRemoved event to the user, if removing old clients due to max number of clients
+  -- - UserLegalHoldEnabled event to contacts of the user, if client type is legalhold
```
arianvp: So; other users are not informed by th clients added and removed, correct?

I wonder if it's harmful if we do do that?  Then we do not need this lazy discovery mechanism when sending

services/brig/src/Brig/API.hs
```diff
@@ -188,6 +217,10 @@ sitemap o = do
     Doc.errorResponse userNotFound
   ---
 
+  -- If the user is expired, it will be removed asynchronously.
+  -- This leads to the following events being sent to clients:
+  -- - UserDeleted event to contacts of the user
```
arianvp: This feels ... odd to me

services/brig/src/Brig/IO/Intra.hs
```diff
@@ -483,6 +484,7 @@ createSelfConv u = do
         . zUser u
         . expect2xx
 
+-- | Calls 'Galley.API.createConnectConversationH'.
```
arianvp: Thanks for this. these things having slightly different naming conventions is realllyyy irritating me though :D

services/galley/src/Galley/API.hs
```diff
@@ -318,6 +331,8 @@ sitemap = do
       .&> zauthBotId
       .&. zauthConvId
       .&. accept "application" "json"
+  -- This endpoint can lead to the following events being sent to clients:
```
arianvp: What does clients mean here?

services/galley/src/Galley/API.hs
```diff
@@ -293,17 +296,27 @@ sitemap = do
       .&. capture "tid"
       .&. capture "uid"
       .&. accept "application" "json"
+  -- This endpoint can lead to the following events being sent to clients:
```
arianvp: to whose clients?

services/galley/src/Galley/API.hs
```diff
@@ -392,6 +407,8 @@ sitemap = do
       description "Max. number of conversations to return"
   ---
 
+  -- This endpoint can lead to the following events being sent to clients:
```
arianvp: when you say clients, dont you mean users? im a bit confused by this terminology

services/brig/src/Brig/API.hs
mheinzel: An "internal event" is put into a queue, usually SQS (see `Brig.Queue`). 

I could either clarify that by pointing to `Brig.Queue` or just remove the information about his happening asynchronously, as it's not directly relevant to the events being sent.

What do you prefer?

services/brig/src/Brig/API.hs
mheinzel: Probably not massively harmful, but could be a performance issue. It would require an event to be sent to every user in every conversation we are part of. By doing it lazily, we spread out that cost and it also only needs to be payed by people actually sending a message (which will also get all added clients at once). Seems like this is often a good tradeoff, especially when there are just a few clients sending messages, but many passive ones.

Also, we need to provide a way for querying this information anyways, e.g. for new clients, after missing messages, or just due to unfortunate timing.

services/brig/src/Brig/API.hs
mheinzel: It is, but clients can create users with an `expires_at` field set (not sure in which case they do). I think we don't actively delete them through some kind of scheduled deletion or regular cleanup job, but just lazily when someone asks for them.

I will make clear that this is only about these ephemeral users and reference `Brig.API.User.userGC`.

arianvp: I think this is for guest accounts only.  

I wonder if we can get away with just using the TTL functionality of  Cassandra. But there's probably a good reason not to do that; I remember TIago telling me about this a few months ago

services/galley/src/Galley/API.hs
mheinzel: To all clients belonging to the people listed in the lines below, in this case the "contacts of the user" mentioned in the request (i.e. `:uid`). Would it be clearer to write "LegalHoldClientRequested event to contacts of the user the client is requested for"?

I use a similar wording everywhere, talking about events being sent to clients, to make clear that clients will receive them directly, as opposed to some server (think for example other backends with federation). Do you have any suggestions for better wording? Maybe "sent to all clients of the specified users:"?

services/galley/src/Galley/API.hs
mheinzel: Similar to above comment, client devices of the recipients (users).

services/galley/src/Galley/API.hs
mheinzel: I'll just out the "to clients" part.

services/brig/src/Brig/API.hs
mheinzel: I'll leave out the note about it being asynchronous.

mheinzel: done

services/galley/src/Galley/API.hs
mheinzel: done

mheinzel:
@arianvp I rebased it and make the adjustments we discussed. Could you take another look?

PR #1027, opened 2020-03-24
Reproducibility

arianvp:
By comitting stack.yaml.lock our build plan is fully reproducible. We've
had several cases where people could not reproduce a build on develop
before either due to cabal revisions or due to people accidentally
touching snapshot files.  stack.yaml.lock is a fixed build plan with
hashes derived from your snapshots and stack.yaml

See
https://github.com/commercialhaskell/stack/blob/master/doc/lock_files.md
for more info

We also commit the cabal files. Not committing them also reduces
reproducibility and the Stack people admit this now as well
https://tech.fpcomplete.com/blog/storing-generated-cabal-files In the
next stack release generating them will be the default and will also not be added to `.gitignore` in new projects


arianvp:
CI now seems to be failing because it has cached wrong versions of dependencies, and stack reproduces a different build from scratch than what was cached before. How do I retrigger build of this cache? Can we do this in an automated way? (Maybe `stack.yaml.lock` changes should be the main trigger for this somehow)

arianvp:
Interesting, I can reproduce this locally. If I build inside docker using:

```
FROM fpco/stack-build:lts-14.12 AS dependencies
COPY . /src/
WORKDIR /src
RUN stack build --system-ghc
```

It fails with
```
Cloning 1a1b313092beb685a9bb15685c83a3162c1e220f from https://github.com/wireapp/saml2-web-sso
Mismatched package metadata for Repo from https://github.com/wireapp/saml2-web-sso, commit 1a1b313092beb685a9bb15685c83a3162c1e220f
Found: saml2-web-sso-0.18 with cabal file 1bee1f06e66d196e1965bad98763f938d525cb663436fd61a959830b6ce136d2,8017 and tree f13e1fa7929bb8ee84eb178355e56066d730a590fca3f9a31e18617aaae71622,4476
Expected: name == saml2-web-sso, version == 0.18, tree == f13e1fa7929bb8ee84eb178355e56066d730a590fca3f9a31e18617aaae71622,4476, cabal file == d65ad9f5d25e5ac40afeb6de695b9fc8fc9e1220c04ac59f506dcfad23d62524,8017
```

too.  For some reason, the cabal file gives a different based on whether I'm inside docker or not? This smells like a stack bug to me :/

arianvp:
Fix is here: @fisx please have a look https://github.com/wireapp/saml2-web-sso/pull/60

libs/api-bot/api-bot.cabal
```diff
@@ -0,0 +1,89 @@
+cabal-version: 1.12
+
+-- This file has been generated from package.yaml by hpack version 0.32.0.
```
akshaymankar: This line will make stack not re-generate cabal files if it was compiled with older hpack. We all have to agree on the version of hpack that stack was compiled with. The latest one is 0.33. Do we want that instead?

arianvp:
Note to self : also send pr to `hsaml2`

fisx:
> Makes sense to me. (we still need to point to the merged saml2-web-sso PR of course)
> 
> > This makes me wonder why we use hpack at all though
> 
> It reduces some duplication, e.g. if we want to add a default language extension, I'd rather touch one `package-defaults.yaml` than ~40 Cabal files (assuming we don't want them to drift apart). Or if I add a new module, I don't want to be required to manually add it to the cabal file as an exposed module. I admit that these are all relatively minor things, though.

those were the reasons back then, yes.  now cabal allows for global sections of dependencies that go for all targets.  i also saw a beta-version of a little tool that injects exposed-modules into cabal files on twitter the other day, but i can't find it any more.  :/

libs/api-bot/api-bot.cabal
fisx: that would be a good reason to get rid of hpack, because whatever we do, it'll always be pain.  :(

akshaymankar:
>  i also saw a beta-version of a little tool that injects exposed-modules into cabal files on twitter the other day, but i can't find it any more. :/

This one? [autopack](https://github.com/kowainik/autopack)

arianvp:
I took the liberty to fix https://github.com/zinfra/backend-issues/issues/1246 while were at it. This greatly simplifies how we define external dependencies. We do not have to make distinctions between 'slow moving' and 'fast moving' dependencies, and do not have to worry about accidentally editing snapshot files anymore (As we do not need them anymore).  Please carefully read the commit messages when you review this.   I'm 99.9% sure I interpreted  Stack's changes to caching in 2.0 correctly and that this is the sane thing to do, but I might have missed something...

stack.yaml
```diff
@@ -50,13 +60,105 @@ extra-deps:
...
+- text-icu-translit-0.1.0.7
+- wai-middleware-gunzip-0.0.2
+- cql-io-tinylog-0.1.0
+- invertible-hxt-0.1
+- network-uri-static-0.1.2.1
+- base58-bytestring-0.1.0
+- stompl-0.5.0
+- pattern-trie-0.1.0
+
+# Not latest as latst one breaks wai-routing
```
mheinzel:
```suggestion
# Not latest as latest one breaks wai-routing
```

arianvp: it at least seems to work on CI; which is promising!  "Using precompiled package" messages everywhere

arianvp:
it at least seems to work on CI; which is promising! "Using precompiled package" messages everywhere

mheinzel:
Yes, works very nicely for me, too! :tada:

arianvp:
So can I merge @mheinzel ?

PR #1030, opened 2020-03-26
Bump to lts-14.27

arianvp:
Depends on https://github.com/wireapp/hscim/pull/44

I was wondering if bumping would work without too many changes; and it seems it did :)

### DIFF
--- a/stack.yaml
+++ b/stack.yaml
@@ -50,10 +50,7 @@ nix:
 allow-newer: true
 
 extra-deps:
-- swagger2-2.5
-# Needed for swagger2-2.5
-- optics-core-0.2@sha256:6966f4f8cc9163b63d87dcca2d2617684b4ac8a80c5e50c69e9f3adf4dcdf0e9,4409
-- indexed-profunctors-0.1@sha256:ddf618d0d4c58319c1e735e746bc69a1021f13b6f475dc9614b80af03432e6d4,1016
+- swagger2-2.4
 - git: https://github.com/fimad/prometheus-haskell
   commit: 2e3282e5fb27ba8d989c271a0a989823fad7ec43
   subdirs:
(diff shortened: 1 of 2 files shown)


arianvp:
We seem to hardcode LTS in CI; do we have to? it's a bit annoying we have to do this in two places

https://github.com/wireapp/wire-server-private/blob/d4594cf03c1756381639d86e513ab883c8587450/ci/dockerfiles/ubuntu14-ghc-8/Dockerfile#L7

stack.yaml
```diff
@@ -50,8 +50,8 @@ nix:
 allow-newer: true
 
 extra-deps:
-- servant-swagger-1.1.7.1@rev:2
-- swagger2-2.5
+# - servant-swagger-1.1.7.1@rev:2
+# - swagger2-2.5
```
mheinzel: what happened with swagger?

arianvp: revision two was added to `14.27` https://www.stackage.org/lts-14.27/package/servant-swagger-1.1.7.1

`swagger2-2.5` I accidentally dropped. Good catch. However it still compiles. What was released in `2.5` that we need? 

I'll just add it back for now

arianvp: @fisx was there a specific reason why we were on 2.5 ?   Dropping it to 2.4 worked fine (no dependency conflicts or compiler errors)

arianvp: The `swagger2-2.5` release on hackage does not even compile for me.  I'm confused.  Given @fisx is maintainer of this package maybe he knows what is up here? https://github.com/wireapp/wire-server/pull/1030/commits/86ed4053cce9038eeb8fdca450ca889ace49f209

fisx:
> We seem to hardcode LTS in CI; do we have to? it's a bit annoying we have to do this in two places
> 
> https://github.com/wireapp/wire-server-private/blob/d4594cf03c1756381639d86e513ab883c8587450/ci/dockerfiles/ubuntu14-ghc-8/Dockerfile#L7

Yes, that looks redundant, but I don't think it's necessary that the resolver there and the one of wire-server are the same.  In the docker image it is just used to build some haskell-related executables, no?

stack.yaml
fisx: [swagger2-2.5 is nice, but not important](https://github.com/GetShopTV/swagger2/releases/tag/v2.5).  I'll see if i can fix the issue upstream, then we can just wait for the next release.

PR #1034, opened 2020-03-31
Better defaults

jschaul:
especially for the folders feature: https://wire.com/en/blog/folders-feature/

These precise numbers were introduced also here:
https://github.com/wireapp/wire-server-deploy/pull/173


jschaul:
(CI failure was an unrelated OOM)

PR #1037, opened 2020-03-31
Add comments/docs about hie.yaml

jschaul:


jschaul:
It's WIP as I wanted to try using hie on my setup first, which I've not gotten round to properly configuring yet. Then that may or may not lead to another link / setup instructions. I will get back to this PR...

jschaul:
This is now ready.

jschaul:
I added some bash checks to ensure the right executable.

PR #1040, opened 2020-04-02
[WIP][DO NOT MERGE]Add temporary tool to check sanity of user data

akshaymankar:
This PR exists solely so this temporary code gets shown to the team before I run it in production.

Recommended way to run it, make sure it is run in an empty directory:

```
CASSANDRA_IP=<IP Address of one of the nodes>
./brig-schema-check $CASSANDRA_IP 9042 | split --unbuffered --verbose -l 1000 -d --suffix-length=10 --filter 'tee >(gzip > $FILE.gz)'
```

This will output a lot of lines saying `Scanning <number>`, but will also store all the output in gzipped files. To see if any errors exists run this command:

```
zgrep null *.gz
```

As a bonus, last line of the output shows total number of users. 
It takes about 15 minutes in staging with less than 15 million users.

### DIFF
--- a/services/brig/brig.cabal
+++ b/services/brig/brig.cabal
@@ -4,7 +4,7 @@ cabal-version: 2.0
 --
 -- see: https://github.com/sol/hpack
 --
--- hash: 2d75605e990ced662e68a90ee59831680ff15d186c31a1e37c95f7783cb96b1b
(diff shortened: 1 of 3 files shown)


akshaymankar:
This tool has served its purpose.

PR #1041, opened 2020-04-02
Federation: Link related issue IDs

mheinzel:
There's also a small change where in two endpoints the `Z-User` header was marked as opaque user ID.


mheinzel:
No review yet and I think the changes are pretty uncontroversial, so I'll just merge now.

PR #1043, opened 2020-04-04
Use restund 0.4.14

tiago-loureiro:
Driven by https://github.com/wireapp/restund/pull/3


tiago-loureiro:
I will take the liberty of merging this and trying it out on our `staging` env.

PR #1045, opened 2020-04-06
Add license headers

mheinzel:
Also fixes issues with ormolu in a few places.


jschaul:
What precise commands/tools did you use to generate these headers? Perhaps we should wait with this PR until the tool used is idempotent?

mheinzel:
I used licensure with the updated configuration from https://github.com/wireapp/wire-server/pull/980 to run `licensure -p && ./tools/ormolu.sh -f`.

Headroom is currently still missing Rust support and it seems to remove existing module-level comments in some cases, but otherwise we have the same result on branch [licenses-headroom](https://github.com/wireapp/wire-server/tree/licenses-headroom).

We can wait until the tools are ready, but it's hard to predict if and when they will be. The burden of manually adding headers to new files seems manageable to me, so I'm leaning towards already merging this even before we can run the script repeatedly.

vaclavsvejcar:
Hi, author of Headroom here, noticed you miss Rust support, are there any other languages you'd like to add? Basically I initially added only languages I mostly know (and use), but I'll add support for more languages if needed. Already created ticket for adding Rust support (https://github.com/vaclavsvejcar/headroom/issues/34) so feel free to discuss anything related to that with me. Also I have no clue how license headers in Rust usually look like, so if you have any sample, I'd be happy if you could share it :)

mheinzel:
Hi, thanks for getting in touch!

Similarly to C, Rust has both line comments with `//` and block comments with `/* ... */`. There is [one example in this PR](https://github.com/wireapp/wire-server/pull/1045/files#diff-2e5bc1c67b1f469b76c963f3a06b6397) as well.

At the moment, Haskell and Rust are the only languages we need to be supported. However, I left some potentially useful ideas about language support in the issue you linked.

As of PR https://github.com/wireapp/wire-server/pull/980, we went with licensure (plus some bash to prevent issues similar to https://github.com/vaclavsvejcar/headroom/issues/28). However, given that headroom 0.2.0 seems to prevent that problem out of the box now and most of our codebase and other tooling is Haskell-based anyways, I will give headroom another try once there is support for Rust.

PR #1046, opened 2020-04-06
Count team members using es

akshaymankar:
Fixes https://github.com/zinfra/backend-issues/issues/1207


akshaymankar:
Whoops, didn't expect spar to break. Fixing it now.

services/galley/src/Galley/API/Teams.hs
```diff
@@ -182,8 +183,13 @@ updateTeamStatus tid (TeamStatusUpdate newStatus cur) = do
...
   where
     journal Suspended _ = Journal.teamSuspend tid
-    journal Active c = Data.teamMembersUnsafeForLargeTeams tid >>= \mems ->
-      Journal.teamActivate tid mems c =<< Data.teamCreationTime tid
+    journal Active c = do
+      mems <- Data.teamMembersUnsafeForLargeTeams tid
+      teamCreationTime <- Data.teamCreationTime tid
+      -- TODO(Akshay): Figure out if this is required to be live data
+      -- a lot of tests assert this by using `createBindingTeam`
+      -- (TeamSize size) <- BrigTeam.getSize tid
```
fisx, line 191: I'm not sure how to read this.  Line 191 has nothing to do with 190, no?  Should this be removed?

```suggestion
```

services/galley/src/Galley/API/Teams.hs
```diff
@@ -475,8 +484,16 @@ deleteTeamMember zusr zcon tid remove mBody = do
...
       body <- mBody & ifNothing (invalidPayload "missing request body")
       ensureReAuthorised zusr (body ^. tmdAuthPassword)
+      (TeamSize sizeBeforeDelete) <- BrigTeam.getSize tid
+      -- TeamSize is 'Natural' and subtracting from  0 is an error
+      -- TeamSize could be reported as 0 if team members are added and removed very quickly,
+      -- which happens in tests
+      let sizeAfterDelete =
+            if sizeBeforeDelete == 0
+              then 0
+              else sizeBeforeDelete - 1
```
fisx, line 494: hum, this does not inspire a lot of trust...  but i have no idea how to make this better.  :(

so if the same thing happens between 1809 and 1810 instead of 0 and 1, the result will be off by one?

services/galley/src/Galley/Intra/Journal.hs
```diff
@@ -69,12 +70,11 @@ journalEvent typ tid dat tim = view aEnv >>= \mEnv -> for_ mEnv $ \e -> do
 ----------------------------------------------------------------------------
 -- utils
 
-evData :: [TeamMember] -> Maybe Currency.Alpha -> TeamEvent'EventData
-evData mems cur =
+evData :: Natural -> [TeamMember] -> Maybe Currency.Alpha -> TeamEvent'EventData
```
fisx, line 73: not sure if i should have understood this from reading this far in the PR, but what's this member list?  hopefully not the entire team?  would that be worth some haddocks here, or somewhere more appropriate?

akshaymankar: It is the entire team unfortunately, but it is only required to find whom to send the bill to. It is a task in our TODO list, so we'll take care of it later.

services/galley/src/Galley/API/Teams.hs
akshaymankar: I had something in my mind and forgot to act on it. Thankfully this confusing comment reminded me. I am not sure if member count is always 1 when a team is activated. I wanted to dig into that and it totally skipped my mind. I will look into this now.

services/galley/src/Galley/API/Teams.hs
akshaymankar: I don't understand what you mean by 1809 and 1810. Can you please explain?

services/galley/src/Galley/API/Teams.hs
akshaymankar: Better now?

services/brig/test/integration/API/Team.hs
```diff
@@ -435,8 +435,10 @@ testInvitationMutuallyExclusive brig = do
 testInvitationTooManyMembers :: Brig -> Galley -> TeamSizeLimit -> Http ()
 testInvitationTooManyMembers brig galley (TeamSizeLimit limit) = do
   (creator, tid) <- createUserWithTeam brig galley
-  pooledForConcurrentlyN_ 16 [1 .. limit -1] $ \_ ->
-    createTeamMember brig galley creator tid Team.fullPermissions
+  SearchUtil.refreshIndex brig
+  pooledForConcurrentlyN_ 16 [1 .. limit -1] $ \_ -> do
+    void $ createTeamMember brig galley creator tid Team.fullPermissions
+    SearchUtil.refreshIndex brig
```
akshaymankar, line 441: Why put it in the loop?

services/galley/src/Galley/API/Teams.hs
fisx: i mean that if you run into this race condition with number higher than 1 and 0, this will just silently get the count wrong by 1.

Does this accumulate?  So every time we enter the count becomes less accurate?

akshaymankar: Yes, this can be off by more than 1, if a lot of people join/leave the team within a very short time (seconds), we are bound to get this number wrong.

PR #1048, opened 2020-04-09
List team members by UserId

fisx:
see https://github.com/zinfra/backend-issues/issues/1197

### DIFF
--- a/services/galley/src/Galley/API.hs
+++ b/services/galley/src/Galley/API.hs
@@ -162,13 +162,13 @@ sitemap = do
     errorResponse Error.notATeamMember
   --
 
-  get "/teams/:tid/members-by-ids" (continue Teams.bulkGetTeamMembersH) $
+  post "/teams/:tid/get-members-by-ids-using-post" (continue Teams.bulkGetTeamMembersH) $
     zauthUserId
       .&. capture "tid"
       .&. def (unsafeRange hardTruncationLimit) (query "maxResults")
       .&. jsonRequest @UserIdList
       .&. accept "application" "json"
-  document "GET" "bulkGetTeamMembers" $ do
+  document "POST" "bulkGetTeamMembers" $ do
     summary
       "Get team members by user id list\n\
       \NOTE: `has_more` field in response body is always `false`."
--- a/services/galley/src/Galley/API.hs
+++ b/services/galley/src/Galley/API.hs
@@ -162,13 +162,13 @@ sitemap = do
     errorResponse Error.notATeamMember
   --
 
-  post "/teams/:tid/members" (continue Teams.bulkGetTeamMembersH) $
+  get "/teams/:tid/members-by-ids" (continue Teams.bulkGetTeamMembersH) $
     zauthUserId
       .&. capture "tid"
       .&. def (unsafeRange hardTruncationLimit) (query "maxResults")
       .&. jsonRequest @UserIdList
       .&. accept "application" "json"
-  document "POST" "bulkGetTeamMembers" $ do
+  document "GET" "bulkGetTeamMembers" $ do
     summary
       "Get team members by user id list\n\
       \NOTE: `has_more` field in response body is always `false`."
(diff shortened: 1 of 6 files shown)


services/galley/src/Galley/API.hs
```diff
@@ -162,6 +162,28 @@ sitemap = do
...
+  post "/teams/:tid/get-members-by-ids-using-post" (continue Teams.bulkGetTeamMembersH) $
+    zauthUserId
+      .&. capture "tid"
+      .&. def (unsafeRange hardTruncationLimit) (query "maxResults")
+      .&. jsonRequest @UserIdList
+      .&. accept "application" "json"
+  document "POST" "bulkGetTeamMembers" $ do
+    summary
+      "Get team members by user id list\n\
+      \NOTE: `has_more` field in response body is always `false`."
```
mheinzel, line 174:
```suggestion
    summary "Get team members by user id list"
    notes "The `has_more` field in the response body is always `false`."
```

services/galley/src/Galley/API/Teams.hs
```diff
@@ -342,6 +344,25 @@ getTeamMembers zusr tid maxResults = do
...
+  pure . json $ teamMemberListJson withPerms memberList
+
+-- | like 'getTeamMembers', but with an explicit list of users we are to return.
+bulkGetTeamMembers :: UserId -> TeamId -> Range 1 HardTruncationLimit Int32 -> [UserId] -> Galley (TeamMemberList, TeamMember -> Bool)
+bulkGetTeamMembers zusr tid maxResults uids = do
+  unless (length uids <= fromIntegral (fromRange maxResults)) $
+    throwM bulkGetMemberLimitExceeded
+  mems <- Data.teamMembersLimited tid uids
+  Data.teamMember tid zusr >>= \case
+    Nothing -> throwM notATeamMember
```
mheinzel, line 360: Also applies to the existing `getTeamMembers` above, but why do we fetch the result before checking that `zusr` is a member of the team? Should the query not be moved into the `Just m` branch?

services/galley/test/integration/API/Teams.hs
```diff
@@ -210,6 +214,34 @@ testListTeamMembersTruncated = do
       "member list does not indicate that there are more members"
       (listFromServer ^. teamMemberListHasMore)
 
+testListTeamMembersDefaultLimitByIds :: TestM ()
+testListTeamMembersDefaultLimitByIds = do
+  (owner, tid, [member1, member2]) <- Util.createBindingTeamWithNMembers 2
+  check owner tid [owner, member1, member2]
+  check owner tid [member1, member2]
+  check owner tid [member1]
+  check owner tid [] -- a bit silly, but hey.
```
mheinzel, line 223: How should the endpoint behave when you also ask for a non-existing user ID. Or *only* for non-existing ones? Or if your `maxResults`  is above the truncation limit?
We could test for some of these cases, too. (imho, mostly the first one case would be useful, maybe also the second)

services/galley/test/integration/API/Teams.hs
```diff
@@ -210,6 +214,34 @@ testListTeamMembersTruncated = do
...
+  where
+    check :: UserId -> TeamId -> [UserId] -> TestM ()
```
mheinzel, line 225:
```suggestion
    check :: HasCallStack => UserId -> TeamId -> [UserId] -> TestM ()
```

Ah, another small thing.

services/galley/src/Galley/API/Teams.hs
fisx: I agree.  26a4bb6530e4a2dedea21144e012f0ee1bb51ac2

services/galley/test/integration/API/Teams.hs
fisx: good idea.  31627edc4f5ebf9eaafc33a2c98d558ce4f1f2a6

I had some truncation errors in there, too, but then I realized that's tested in the next case, so I removed them.

PR #1050, opened 2020-04-14
Use fork of bloodhound to support ES 5.2

akshaymankar:
Fixes https://github.com/zinfra/backend-issues/issues/1299

Branch on bloodhound fork: https://github.com/wireapp/bloodhound/tree/feature/count-with-ES-5.2

I don't plan on sending a PR for this as ES 5.2 is very old and we should just upgrade from it. There is already an issue for that: https://github.com/zinfra/backend-issues/issues/1161

### DIFF
--- a/stack.yaml
+++ b/stack.yaml
@@ -91,7 +91,7 @@ extra-deps:
   commit: 0a5142cd3ba48116ff059c041348b817fb7bdb25
 
 - git: https://github.com/wireapp/bloodhound
-  commit: 88261015a6ea8b10c0c73dedfb5f2ba7f2e36e81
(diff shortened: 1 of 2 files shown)


stack.yaml
```diff
@@ -90,8 +90,8 @@ extra-deps:
 - git: https://github.com/wireapp/hspec-wai
   commit: 0a5142cd3ba48116ff059c041348b817fb7bdb25
 
-- git: https://github.com/bitemyapp/bloodhound
-  commit: c6233c493b1a7c3df8099872bbc1f66c5f25d95f
+- git: https://github.com/wireapp/bloodhound
+  commit: 88261015a6ea8b10c0c73dedfb5f2ba7f2e36e81
```
fisx, line 94:
```suggestion
  commit: 88261015a6ea8b10c0c73dedfb5f2ba7f2e36e81  # https://github.com/wireapp/bloodhound/tree/feature/count-with-ES-5.2
```

PR #1051, opened 2020-04-14
Test that custom backend domains are case-insensitive

mheinzel:
They get normalized to lowercase during parsing.

### DIFF
--- a/services/galley/test/integration/API/CustomBackend.hs
+++ b/services/galley/test/integration/API/CustomBackend.hs
@@ -38,7 +38,8 @@ tests s =
     [ test s "GET by-domain (404)" getByDomainNotFound,
       test s "GET by-domain (400)" getByDomainInvalidDomain,
       test s "PUT, GET by-domain (200)" getByDomainFound,
-      test s "PUT, DELETE, GET by-domain (404)" getByDomainDeleted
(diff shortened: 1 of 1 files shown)


services/galley/test/integration/API/CustomBackend.hs
```diff
@@ -87,3 +88,25 @@ getByDomainDeleted = do
...
+    !!! const 201 === statusCode
+  get (galley . path "/custom-backend/by-domain/aB.aB") !!! do
+    const 200 === statusCode
+    const jsonBody === responseJsonUnsafe
+  delete (galley . path "/i/custom-backend/by-domain/AB.ab")
+    !!! const 200 === statusCode
+  -- not just the matching is done in a case-insensitive way,
+  -- but we actually store the normalized domain.
+  get (galley . path "/custom-backend/by-domain/Ab.Ab") !!! do
+    const 404 === statusCode
```
fisx, line 111:
```suggestion
    const 200 === statusCode
```

As I understand it, this is exactly the problem.  The client does a lookup by domain, and shouldn't have to normalize the lookup, but interpret the string case-insensitively.

I haven't checked the context, though.

mheinzel: This is a 404 because we deleted the entry just before. There is similar call with 200 a few lines above.

PR #1052, opened 2020-04-14
Fix: searching for name/handle prefix as a team member

akshaymankar:
Fixes https://github.com/zinfra/backend-issues/issues/1286


akshaymankar:
# Deployment Steps

1. Update Index Settings
1. Update Mappings
1. Update all documents using [_update_by_query](https://www.elastic.co/guide/en/elasticsearch/reference/5.3/docs-update-by-query.html)
1. Upgrade Brig

akshaymankar:
>     1. Update Index Settings

Since we are adding an analyzer, this is [not possible without closing the index](https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-update-settings.html#update-settings-analysis). Closing an index prevents all reads and writes on the index. This would mean downtime for everything that relies on the index. With count relying on index, it actually means a lot. So, we cannot do just update index settings even if they are called updatable index settings.

I am now looking into zero downtime ways of doing it. Will update soon.

services/brig/test/integration/API/Search.hs
```diff
@@ -483,27 +481,16 @@ runBH opts =
...
-            "type": "text"
-          },
-          "team": {
-            "type": "keyword"
-          }
-        }
-      }
-    }
-  |]
+  Aeson.object
```
fisx, line 484: I assume you've ruled out that `rm -rf ~/.stack && find / -name '*stack*' -exec rm -rf {} \;` would have helped?

akshaymankar: Yeah, I removed all of those files and rebuilt it, it still didn't work.

fisx: 1. What if you change it back now, after it works?  will it keep working?  :-)P
2. Did I see somebody mess with aeson versions on some branch?  Is that something worth trying?

akshaymankar: >     1. What if you change it back now, after it works?  will it keep working?  :-)P

Definitely worth trying, next compile is going to be this, hopefully this uncovers some mysteries.

>     2. Did I see somebody mess with aeson versions on some branch?  Is that something worth trying?

I tried several things (including compiling several times using `git bisect`) before I came to the conclusion that bumping aeson was a problem, somehow after taking back the aeson bump I could compile it. But it failed to compile from scratch. So, maybe some leftover state made it compile, I am not sure.

akshaymankar: When I replace the QQ expression with `undefined`, it compiles. Then if I add it back, it keeps compiling.

akshaymankar: Once it compiles, it keeps compiling even after `stack clean`. Only removing all the `.stack-work` directories make it fail again. I think this definitely needs to be reported to stack or ghc. @fisx Can you help me shrink it down so we don't have to make them compile.

Steps to reproduce this error:

```bash
docker run -it quay.io/akshay_wire/alpine-wire-builder bash
git clone https://github.com/wireapp/wire-server
cd wire-server
git checkout akshaymankar/cursed-alpine-qq-issue
stack build brig # See this fail
vi services/brig/test/integration/API/Search.hs # Replace the qq at the bottom with undefined
stack build brig # Passes
vi services/brig/test/integration/API/Search.hs # Add the qq back, or just git checkout -f
stack build brig # Passes - #TooCoolForReproducibility
```

services/brig/index/src/Brig/Index/Options.hs
```diff
@@ -75,10 +81,20 @@ data CassandraSettings = CassandraSettings
   }
   deriving (Show)
 
+data ReindexFromAnotherIndexSettings = ReindexFromAnotherIndexSettings
+  { _reindexEsServer :: URIRef Absolute,
+    _reindexSrcIndex :: ES.IndexName,
+    _reindexDestIndex :: ES.IndexName,
+    _reindexTimeout :: Int
```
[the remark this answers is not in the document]
akshaymankar: Changed the name, to `_reindexTimoutSeconds`.

PR #1054, opened 2020-04-15
Fix swagger.

fisx:


mheinzel:
I pushed some other swagger fixes for things I noticed when reviewing this PR.

For the OTR messages, interestingly, the full query parameter descriptions were there for `postProtoOtrMessage` and `postOtrBroadcast`, but not the other versions (Proto broadcast and non-Proto conversation message).

PR #1057, opened 2020-04-16
HotFix: Tolerate nulls in DB while re-creating ES index

akshaymankar:
Fixes https://github.com/zinfra/backend-issues/issues/1308

This is required to go to production before we can push changes from #1046, hence the hotfix.


jschaul:
Merging to have artifacts built.

PR #1058, opened 2020-04-16
Host galley endpoints via Servant.Raw

mheinzel:
This wraps the existing galley `Application` into a Servant API.
Extracted from #1015, as I wanted people to be able to review and discuss it separately.

### DIFF
--- a/services/galley/src/Galley/Run.hs
+++ b/services/galley/src/Galley/Run.hs
@@ -36,10 +36,14 @@ import Galley.App
 import qualified Galley.Data as Data
 import Galley.Options (Opts, optGalley)
 import Imports
-import Network.Wai (Middleware)
+import Network.Wai (Application, Middleware)
 import qualified Network.Wai.Middleware.Gunzip as GZip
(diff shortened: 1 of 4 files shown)


services/galley/src/Galley/Run.hs
```diff
@@ -60,14 +64,24 @@ run o = do
...
   let rtree = compile sitemap
+      app :: Application
       app r k = runGalley e r (route rtree r k)
+      -- the servant API wraps the one defined using wai-routing
+      servantApp :: Application
+      servantApp r =
+        Servant.serve
+          -- we don't host any Servant endpoints yet, but will add some for the
+          -- federation API, replacing the empty API.
+          (Proxy @(Servant.EmptyAPI :<|> Servant.Raw))
```
fisx, line 75: Why not this?

```suggestion
          (Proxy @Servant.Raw)
```

I don't suggest it's better, just curious.

mheinzel: This is where I will put `Federation.Api` soon, so I wanted to already put as much of the structure in place. No deeper reason.

PR #1059, opened 2020-04-16
Move swagger model to a better place; index it.

fisx:
Fixes https://github.com/zinfra/backend-issues/issues/1304


fisx:
> Nice.
> 
> Btw, did you check if there are any other models that somebody might have forgotten to include? Wouldn't be surprised if we found more.

I have not, partly because I would really prefer to start migrating this to servant-swagger instead.  I'll keep fixing them as somebody stumbles on them.

PR #1064, opened 2020-04-17
Make some IDs in galley events opaque

mheinzel:
Still a few things to think about.

Generally, I got the feeling now that I really want to keep the usage of opaque IDs strictly to the API boundaries. This includes storing optional `IdMapping`s in database tables alongside the local ones instead of sticking to storing a single (opaque) ID, but we have to think of the implications regarding migrations and data consistency.


mheinzel:
This is outdated now and has been subsumed by other PRs.

PR #1067, opened 2020-04-20
Some random cleanups and refactorings

mheinzel:
The accumulated over time on a branch.

### DIFF
--- a/libs/wai-utilities/src/Network/Wai/Utilities/Request.hs
+++ b/libs/wai-utilities/src/Network/Wai/Utilities/Request.hs
@@ -1,4 +1,5 @@
 {-# LANGUAGE OverloadedStrings #-}
+{-# OPTIONS_GHC -Wno-orphans #-}
 
 -- This file is part of the Wire Server implementation.
 --
@@ -17,8 +18,7 @@
 -- You should have received a copy of the GNU Affero General Public License along
 -- with this program. If not, see <https://www.gnu.org/licenses/>.
 
-{-# OPTIONS -Wno-orphans #-}
--- for "instance HasRequest Request" :(
+-- orphan instance for "instance HasRequest Request" :(
 
 module Network.Wai.Utilities.Request where
 
--- a/services/brig/src/Brig/API.hs
+++ b/services/brig/src/Brig/API.hs
@@ -76,11 +76,11 @@ import Network.HTTP.Types.Status
 import Network.Wai (Response, lazyRequestBody)
 import Network.Wai.Predicate hiding (result, setStatus)
 import Network.Wai.Routing
-import Network.Wai.Utilities
-import qualified Network.Wai.Utilities as Utilities
+import Network.Wai.Utilities as Utilities
 import Network.Wai.Utilities.Response (json)
 import Network.Wai.Utilities.Swagger (document, mkSwaggerApi)
 import qualified Network.Wai.Utilities.Swagger as Doc
+import Network.Wai.Utilities.ZAuth (zauthConnId, zauthUserId)
 
 ---------------------------------------------------------------------------
 -- Sitemap
@@ -102,7 +102,7 @@ sitemap o = do
   post "/i/users/:uid/auto-connect" (continue autoConnectH) $
     accept "application" "json"
       .&. capture "uid"
-      .&. opt (header "Z-Connection")
+      .&. opt zauthConnId
       .&. jsonRequest @UserSet
   -- This endpoint can lead to the following events being sent:
   -- - UserActivated event to created user, if it is a team invitation or user has an SSO ID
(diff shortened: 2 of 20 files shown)


services/brig/src/Brig/API.hs
```diff
@@ -111,7 +111,7 @@ sitemap o = do
     accept "application" "json"
       .&. jsonRequest @NewUser
   put "/i/self/email" (continue changeSelfEmailNoSendH) $
-    header "Z-User"
+    zauthUserId
```
mheinzel, line 114: This slightly changes behaviour, as it masks the exact error message:

```
zheader = fmap (result (Fail . const accessDenied) Okay) . header
```

We already do the same in Galley, so it's probably a desirable change.

services/brig/src/Brig/User/API/Auth.hs
```diff
@@ -260,13 +262,18 @@ renew ::
...
   Maybe (Either ZAuth.AccessToken ZAuth.LegalHoldAccessToken) ->
   Handler (Either (Auth.Access ZAuth.User) (Auth.Access ZAuth.LegalHoldUser))
-renew (Just (Left ut)) (Just (Left at)) = Left <$> Auth.renewAccess ut (Just at) !>> zauthError
-renew (Just (Left ut)) Nothing = Left <$> Auth.renewAccess @ZAuth.User @ZAuth.Access ut Nothing !>> zauthError
-renew (Just (Right lut)) (Just (Right lat)) = Right <$> Auth.renewAccess @ZAuth.LegalHoldUser @ZAuth.LegalHoldAccess lut (Just lat) !>> zauthError
-renew (Just (Right lut)) Nothing = Right <$> Auth.renewAccess @ZAuth.LegalHoldUser @ZAuth.LegalHoldAccess lut Nothing !>> zauthError
-renew (Just (Left _)) (Just (Right _)) = throwStd authTokenMismatch
-renew (Just (Right _)) (Just (Left _)) = throwStd authTokenMismatch
-renew Nothing _ = throwStd authMissingCookie
+renew = \case
```
mheinzel, line 265: Not sure if this really makes the code clearer, but I found it hard to follow the pattern matching and it was a nice challenge on a Friday afternoon. :)

services/spar/src/Spar/Scim.hs
```diff
@@ -1,8 +1,8 @@
 {-# LANGUAGE DuplicateRecordFields #-}
 {-# LANGUAGE OverloadedLists #-}
 {-# LANGUAGE RecordWildCards #-}
-{-# OPTIONS_GHC -Wno-missing-methods
-    -Wno-orphans #-}
+{-# OPTIONS_GHC -Wno-missing-methods #-}
```
fisx, line 4: I wonder how that got in...?  :)

(I would have removed this and at least implemented the methods as `undefined`, but I guess that's not a great difference in terms of runtime behavior.)

libs/wai-utilities/src/Network/Wai/Utilities/Request.hs
```diff
@@ -17,8 +18,7 @@
 -- You should have received a copy of the GNU Affero General Public License along
 -- with this program. If not, see <https://www.gnu.org/licenses/>.
 
-{-# OPTIONS -Wno-orphans #-}
--- for "instance HasRequest Request" :(
+-- orphan instance for "instance HasRequest Request" :(
```
fisx, line 21: This should also be moved up.

services/brig/src/Brig/User/API/Auth.hs
fisx: Perhaps you can write 2-3 lines about what this does in English?  Should be a nice challenge, too.  :)P

libs/wai-utilities/src/Network/Wai/Utilities/Request.hs
mheinzel: It can't, with current Ormolu. It will move all comments below the `OPTIONS_GHC` and `LANGUAGE` pragmas.
This was changed in Ormolu recently, but is not released yet: https://github.com/tweag/ormolu/blob/master/CHANGELOG.md

services/spar/src/Spar/Scim.hs
mheinzel: And why is the instance needed, but not defining any of its methods doesn't break? How has this been in prod for two years? :grimacing: 

Looking into it now.

services/brig/src/Brig/User/API/Auth.hs
mheinzel: I added some comments now.

But my this change didn't make it easier to understand the code, maybe I should leave it at the old version, which was verbose, but more explicit, and just add comments to that?

services/spar/src/Spar/Scim.hs
mheinzel: So it seems like we host the full Scim `Site`, including the `GroupSite`, but the implementation for that goes through the `GroupDB` class, which would fail any time one of these endpoints is called.

I just made a commit to only host the `UserSite` and `ConfigSite` by copying a bit of code from `Web.Scim.Server`. This means people accessing group endpoints would get a proper 404 without hitting undefined methods. How does that sound to you?

About to push that in a moment.

services/brig/src/Brig/User/API/Auth.hs
fisx: no, please keep it, now that it's not quite as late in the day and there is english, it's pretty!

PR #1063, opened 2020-04-17
Tests for otr end-points.

fisx:
This can be reviewed, but the important thing is that the tests pass, not that it gets merged.  I am planning to add a few more unit / qc tests soon.

Fixes https://github.com/zinfra/backend-issues/issues/1248

PR #1068, opened 2020-04-21
delete non-empty IdPs

fisx:
https://github.com/zinfra/backend-issues/issues/1295

~~This is based on https://github.com/wireapp/wire-server/pull/1065, I can rebase this when that is merged.~~

### DIFF
--- a/services/spar/src/Spar/API.hs
+++ b/services/spar/src/Spar/API.hs
@@ -228,6 +228,14 @@ idpGetAll zusr = withDebugLog "idpGetAll" (const Nothing) $ do
   _idplProviders <- wrapMonadClientWithEnv $ Data.getIdPConfigsByTeam teamid
   pure IdPList {..}
 
+-- | Delete empty IdPs, or if @"purge=true"@ in the HTTP query, delete all users
+-- *synchronously* on brig and spar, and the IdP once it's empty.
+--
+-- The @"purge"@ query parameter is as a quick work-around until we have something better.  It
+-- may very well time out, but it processes the users under the 'IdP' in chunks of 2000, so no
+-- matter what the team size, it shouldn't choke any servers, just the client (which is
+-- probably curl running locally on one of the spar instances).
+-- https://github.com/zinfra/backend-issues/issues/1314
 idpDelete :: Maybe UserId -> SAML.IdPId -> Maybe Bool -> Spar NoContent
 idpDelete zusr idpid (fromMaybe False -> purge) = withDebugLog "idpDelete" (const Nothing) $ do
   idp <- SAML.getIdPConfig idpid
--- a/services/spar/src/Spar/API.hs
+++ b/services/spar/src/Spar/API.hs
@@ -228,15 +228,27 @@ idpGetAll zusr = withDebugLog "idpGetAll" (const Nothing) $ do
   _idplProviders <- wrapMonadClientWithEnv $ Data.getIdPConfigsByTeam teamid
   pure IdPList {..}
 
-idpDelete :: Maybe UserId -> SAML.IdPId -> Spar NoContent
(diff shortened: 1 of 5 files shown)


fisx:
nope, not yet working, fix coming up.

fisx:
@arianvp you have the most / freshest context, perhaps you can take a look at this after https://github.com/wireapp/wire-server/pull/1065?

fisx:
tests have passed locally.  i'll ping ci when https://github.com/wireapp/wire-server/pull/1063 is green again.

services/spar/src/Spar/API.hs
```diff
@@ -228,15 +228,27 @@ idpGetAll zusr = withDebugLog "idpGetAll" (const Nothing) $ do
...
       team = idp ^. SAML.idpExtraInfo . wiTeam
-  -- fail if idp is not empty
+  -- if idp is not empty: fail or purge
   idpIsEmpty <- wrapMonadClient $ isNothing <$> Data.getSAMLAnyUserByIssuer issuer
-  unless idpIsEmpty $ throwSpar SparIdPHasBoundUsers
+  when (not idpIsEmpty) $ do
+    if fromMaybe False purge
+      then do
+        let go :: Spar ()
+            go = do
```
arianvp, line 243: can we do the same without all the nesting?  at least the purge we can do with viewpattern

services/spar/src/Spar/API.hs
```diff
@@ -228,15 +228,27 @@ idpGetAll zusr = withDebugLog "idpGetAll" (const Nothing) $ do
   _idplProviders <- wrapMonadClientWithEnv $ Data.getIdPConfigsByTeam teamid
   pure IdPList {..}
 
-idpDelete :: Maybe UserId -> SAML.IdPId -> Spar NoContent
-idpDelete zusr idpid = withDebugLog "idpDelete" (const Nothing) $ do
+idpDelete :: Maybe UserId -> SAML.IdPId -> Maybe Bool -> Spar NoContent
+idpDelete zusr idpid purge = withDebugLog "idpDelete" (const Nothing) $ do
```
arianvp, line 232: The problem is this will never work for large teams. it will always fail halfway through and timeout.  We need something better for this.

services/spar/test-integration/Util/Core.hs
```diff
@@ -287,6 +289,24 @@ aFewTimes action good = do
       (\_ -> pure . not . good)
       (\_ -> action `runReaderT` env)
 
+-- | Duplicate of 'Spar.Intra.Brig.getBrigUser'.
+getUserBrig :: HasCallStack => UserId -> TestSpar (Maybe User)
+getUserBrig uid = do
+  env <- ask
+  let req =
+        (env ^. teBrig) . path "/self"
```
arianvp, line 297: Prefer if we stop calling "user-facing" endpoints from service to service. We should instead only use internal endpoints between services

services/spar/src/Spar/API.hs
fisx: a4fd2705beef216bfc0b412ad6be45ec10ea6a79

services/spar/src/Spar/API.hs
fisx: 496b0c4c504996e36adf0e701f608089f2f056f1

services/spar/test-integration/Util/Core.hs
fisx: ah, i agreed with you when you wrote this, but i just realized that luckily this is not service-to-server, but integration-test-to-service.  i suppose that's legit?

fisx: ah.  right.  so, i'm happy to take on the task of refactoring this properly, over our entire code base.  but that's a separate concern, and the helper in the tests can stay!  :)

PR #1075, opened 2020-04-24
Allow galley to start but issue a warning

tiago-loureiro:
The original logic was flawed.

### DIFF
--- a/services/galley/src/Galley/App.hs
+++ b/services/galley/src/Galley/App.hs
@@ -153,8 +153,8 @@ validateOptions l o = do
           ( msg .
               val $
                 "Your journaling events for teams larger than " <> toByteString' optFanoutLimit <>
-                "may have some admin user ids missing. \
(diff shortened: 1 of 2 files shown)


services/galley/src/Galley/App.hs
```diff
@@ -148,22 +148,18 @@ validateOptions l o = do
...
           l
-          ( msg $
-              val
-                "Your journaling events may have some admin user ids missing. \
-                \This is fine for testing purposes but NOT for production use!!"
+          ( msg .
+              val $
+                "Your journaling events for teams larger than " <> toByteString' optFanoutLimit <>
+                " may have some admin user ids missing. \
+                \ This is fine for testing purposes but NOT for production use!!"
```
jschaul, line 157: Sorry, why do we merge " not for production " code? Shouldn't this misconfiguration prevent galley from starting up?

PR #1077, opened 2020-04-25
Monitor access to DynamoDB

tiago-loureiro:
In sequence of recent issues with DynamoDB, keeping better track of requests to DynamoDB may help to bring some more insights to the topic.

With this PR we log when too many attempts to get a lock fail (5, do we think this may be too verbose?) and we track 3 different metrics:
 * `optimistic_lock_grab_attempt_failed`: attempt to get a lock failed, will be retried
 * `optimistic_lock_failed`: failed all (10) attempts to grab a lock
 * `provisioned_throughput_exceeded`: whenever we exceed the table throughput, we should [backoff](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/Programming.Errors.html#Programming.Errors.RetryAndBackoff). The logic seems correct but it's still worthwhile tracking when we exceed such thresholds.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -137,7 +137,7 @@ docker-builder:
 .PHONY: docker-intermediate
 docker-intermediate:
 	# `docker-intermediate` needs to be built whenever code changes - this essentially runs `stack clean && stack install` on the whole repo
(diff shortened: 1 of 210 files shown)


tiago-loureiro:
> > 5, do we think this may be too verbose?
> 
> I don't think so. If 5 attempts fail, I would like to know.

Yes, but we'll know with metrics already. Just thinking whether this may cause too many logs in cases where we are overloaded.

mheinzel:
Ah, in that case it seems redundant.

tiago-loureiro:
Removed the logging of warning to rely on metrics.

PR #1079, opened 2020-04-27
Make make docker-intermediate command work again

arianvp:
This allows you to build docker images locally again.

It is still super slow; which gets annoying quickly.

Perhaps we can improve this by bind-mounting `.stack-work` into the container. This should be safe
if `stack.yaml` has `docker.enable: true`  we could set `docker.image: alpine-prebuilder` on the `stack.yaml` for that

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -137,7 +137,7 @@ docker-builder:
 .PHONY: docker-intermediate
 docker-intermediate:
 	# `docker-intermediate` needs to be built whenever code changes - this essentially runs `stack clean && stack install` on the whole repo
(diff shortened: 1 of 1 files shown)


akshaymankar:
> Perhaps we can improve this by bind-mounting .stack-work into the container. This should be safe
if stack.yaml has docker.enable: true we could set docker.image: alpine-prebuilder on the stack.yaml for that

I don't think we can bind mount .stack-work while running `docker build`, unless I am missing something.

arianvp:
Yeh i'll mess around with it a bit. but not in this PR. There might be some changes in Stack 2.0 that make this harder or easier

PR #1080, opened 2020-04-28
Allow team deletion on stern

tiago-loureiro:
Adds an internal endpoint allowing binding teams of 1 member to be unconditionally deleted.

Fixes: https://github.com/zinfra/backend-issues/issues/1387

### DIFF
--- a/tools/stern/src/Stern/API.hs
+++ b/tools/stern/src/Stern/API.hs
@@ -503,7 +503,6 @@ deleteTeam (givenTid ::: email) = do
   where
     handleNoUser = ifNothing (Error status404 "no-user" "No such user with that email")
     handleNoTeam = ifNothing (Error status404 "no-binding-team" "No such binding team")
-
     wrongMemberCount = Error status403 "wrong-member-count" "Only teams with 1 user can be deleted"
     bindingTeamMismatch = Error status404 "binding-team-mismatch" "Binding team mismatch"
 
--- a/tools/stern/src/Stern/API.hs
+++ b/tools/stern/src/Stern/API.hs
@@ -232,7 +232,7 @@ sitemap = do
       .&. query "email"
   document "DELETE" "deleteTeam" $ do
     summary "Delete a team (irrevocable!) You can only delete teams with 1 user!"
-    Doc.notes "Email the userId's (to prevent copy/paste mistakes)"
+    Doc.notes "The email address of the user must be provided to prevent copy/paste mistakes"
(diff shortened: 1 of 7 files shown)


tools/stern/src/Stern/API.hs
```diff
@@ -627,7 +660,7 @@ groupByStatus conns =
     byStatus s = length . filter ((==) s . ucStatus)
 
 ifNothing :: Error -> Maybe a -> Handler a
-ifNothing e = maybe (throwM e) return
+ifNothing e = maybe (throwE e) return
```
mheinzel, line 663: So this throws errors in `ExceptT` instead of `IO` now? Good.

tools/stern/src/Stern/API.hs
```diff
@@ -228,6 +227,22 @@ sitemap = do
       Doc.optional
     Doc.response 200 "Account deleted" Doc.end
     Doc.response 400 "Bad request" (Doc.model Doc.errorModel)
+  delete "/teams/:tid" (continue deleteTeam) $
+    capture "tid"
+      .&. query "email"
+  document "DELETE" "deleteTeam" $ do
+    summary "Delete a team (irrevocable!) You can only delete teams with 1 user!"
+    Doc.notes "Email the userId's (to prevent copy/paste mistakes)"
```
mheinzel, line 235: Not quite sure what this means. It seems like you need to provide the email address of the only team member, but what does it have to do with a "userId"?

tiago-loureiro:
```suggestion
    Doc.notes "Email of the user (to prevent copy/paste mistakes)"
```

tiago-loureiro: > Not quite sure what this means. It seems like you need to provide the email address of the only team member, but what does it have to do with a "userId"?

it was meant to be user, thanks

tools/stern/src/Stern/API.hs
tiago-loureiro: Yes, I actually realised that `throwM` didn't quite work. It was throwing 5xx.

tools/stern/src/Stern/API.hs
mheinzel: Then I'd either put it into the "email" parameter's description or turn it into a full sentence like "The email address of the user must be provided to prevent copy/paste mistakes".

mheinzel: I just pushed that to the branch, hope that was okay. Once CI passed, we can merge (in my opinion).

PR #1081, opened 2020-04-28
Index billing team members

akshaymankar:
Fixes https://github.com/zinfra/backend-issues/issues/1296


services/galley/src/Galley/API/Teams.hs
```diff
@@ -535,9 +565,8 @@ deleteTeamMember zusr zcon tid remove mBody = do
     tm <- maybe (throwM teamMemberNotFound) pure targetMember
     unless (canDeleteMember dm tm) $ throwM accessDenied
   team <- tdTeam <$> (Data.team tid >>= ifNothing teamNotFound)
-  removeMembership <- Data.teamMember tid remove
   mems <- Data.teamMembersForFanout tid
-  if team ^. teamBinding == Binding && isJust removeMembership
+  if team ^. teamBinding == Binding && isJust targetMember
```
tiago-loureiro, line 569: So we were doing this twice? Good catch

akshaymankar: Yep, we were doing this twice.

akshaymankar:
This PR is now reviewable, things which are still pending before this can be merged:
- [ ] Make a tool to backfill data in production
- [ ] Add the new flag to helm charts
- [ ] Add the new flag to production config
- [ ] Integrate the new tool with helm charts

services/galley/src/Galley/API/Teams.hs
```diff
@@ -172,6 +173,8 @@ createBindingTeam zusr tid (BindingNewTeam body) = do
   let owner = newTeamMember zusr fullPermissions Nothing
   team <- Data.createTeam (Just tid) zusr (body ^. newTeamName) (body ^. newTeamIcon) (body ^. newTeamIconKey) Binding
   finishCreateTeam team owner [] Nothing
+  Data.addBillingTeamMember tid (owner ^. userId)
```
tiago-loureiro, line 176: Can't we instead do this inside the data model when adding members so we can do this transactionally?

services/galley/src/Galley/API/Teams.hs
```diff
@@ -453,9 +454,14 @@ uncheckedAddTeamMember tid nmem = do
...
   (TeamSize sizeBeforeAdd) <- addTeamMemberInternal tid Nothing Nothing nmem mems
-  -- FUTUREWORK: This takes a list of members for fanout, but it should really
-  --             use the hardlimit instead. We want to avoid double lookups though
-  Journal.teamUpdate tid (sizeBeforeAdd + 1) $ newTeamMemberList ((nmem ^. ntmNewTeamMember) : mems ^. teamMembers) (mems ^. teamMemberListType)
+  let new = nmem ^. ntmNewTeamMember
+
+  -- This has to only be done for binding teams, it is assumed here as this function is
+  -- supposed to be called only for binding teams
+  when (hasPermission new SetBilling) $
+    Data.addBillingTeamMember tid (new ^. userId)
```
tiago-loureiro, line 462: Same comment as above; when adding/removing/updating team members, we should update this table inside the data module

services/galley/src/Galley/API/Teams.hs
```diff
@@ -471,42 +477,64 @@ updateTeamMember zusr zcon tid targetMember = do
...
-  -- When journaling is enabled, we are guaranteed that teamMembersForFanout returns all users
+
+  -- Manage billing member data
+  when (team ^. teamBinding == Binding) $ do
+    -- add new billing member, if applicable
+    let isNewBillingMember =
+          not (isBillingMember previousMember)
+            && isBillingMember targetMember
+    when isNewBillingMember $
+      Data.addBillingTeamMember tid targetId
```
tiago-loureiro, line 510: I think this logic would be better suited inside the `updateTeamMember` function

libs/galley-types/src/Galley/Types/Teams.hs
```diff
@@ -362,7 +364,8 @@ newtype TeamCreationTime = TeamCreationTime
 
 data FeatureFlags = FeatureFlags
   { _flagSSO :: !FeatureSSO,
-    _flagLegalHold :: !FeatureLegalHold
+    _flagLegalHold :: !FeatureLegalHold,
+    _flagIndexedBillingTeamMembers :: !FeatureIndexedBillingTeamMembers
```
tiago-loureiro, line 368: This is really only required for an easier migration, right? Wondering if it would be easier to not make it a feature flag (it does not sound like sth we want to turn on/off) and also give it the default value of false, wdyt?

akshaymankar: > This is really only required for an easier migration, right? 

This is correct.

> Wondering if it would be easier to not make it a feature flag (it does not sound like sth we want to turn on/off) and also give it the default value of false, wdyt?

I am not sure what you mean by not feature flag? If you mean make it just a normal setting, I think it is the same effort. Giving it a default value might make our life easier with changes to our deployment setup and helm charts. But I've already made them.
Semantically, I agree it is a bit awkward, once it is turned on I don't see any reason anyone would want to turn it off. Let me know if you think it sits better with normal settings, it shouldn't be too much effort to move it.

services/galley/src/Galley/API/Teams.hs
akshaymankar: We don't want to do this for NonBindingTeams, so we'll have to put a check there. I am also not sure how we decide what goes in `API.Teams` and what goes in `Galley.Data`. Can you please elaborate on the reasoning?

tiago-loureiro: I think it's fine to do it for `BindingTeams` as members of those teams also have the `SetBilling` permission

libs/galley-types/src/Galley/Types/Teams.hs
tiago-loureiro: Yes, I think that because it's an option to help with migration, we should not make it a feature flag. And a default would help. Running that backfill can be in the docs but I don't see a good use to keep this around.

Also: we should have a way to check if the migration was done/needed to prevent someone from playing around with this option and putting galley in a bad state.

services/galley/src/Galley/API/Teams.hs
akshaymankar: I moved it there, but I also removed the checks. I found that Cassandra doesn't complain if we insert a primary key twice, it just updates rest of the columns. In this table we only have the two primary key columns. I am not sure if this has any significant performance penalty, I couldn't find any answers online. Let me know if it is a bad idea.

libs/galley-types/src/Galley/Types/Teams.hs
akshaymankar: Playing around with this option doesn't put galley in bad state. We write to the new table anyways. And checking if the migration was done would be fairly expensive unless we want to store a boolean in the database (we will have to seed it for new installations). Maybe it is ok for people to mess with this option, it would only mean billing could be broken for large teams, but data in C* will always be consistent. Also, we should probably remove this option in the next release (or the one after that) anyways

akshaymankar:
> When we call `updateTeamMember` we always have the previous permissions, right? Can we check that instead of always deleting?

I just asked the question about it. I will add it. I wasn't sure if there was any problem with that or not. For my education, can you please tell me if it is really bad in C* terms or is it just that we should reduce a query if we can?

services/galley/src/Galley/API/Teams.hs
jschaul: > I moved it there, but I also removed the checks. I found that Cassandra doesn't complain if we insert a primary key twice, it just updates rest of the columns. In this table we only have the two primary key columns. I am not sure if this has any significant performance penalty, I couldn't find any answers online. Let me know if it is a bad idea.

(I don't have much context, just replying to the comment)

Writing to cassandra using the same primary key is, as such, fine - however, if those requests happen close together time-wise, you may end up in a situation where the (INSERT key, value1, null, null) wins the race over the (INSERT key, value1, value2, value3), i.e. there can be race conditions in theory. At least I think this can be the case. Maybe I'm wrong. Running tests locally (with a bunch of inserts to local cassandra) can sometimes show this behaviour.

tiago-loureiro:
> I just asked the question about it. I will add it. I wasn't sure if there was any problem with that or not. For my education, can you please tell me if it is really bad in C* terms or is it just that we should reduce a query if we can?

I was mostly concerned about many tombstones but @jschaul concern about tombstones is also a good point. Here's a contrived example of what I am thinking about: https://gist.github.com/tiago-loureiro/65db0c0fe8db66579609159ea1b97f86

Here's what I tried:
```
INSERT ... ('bla',0);
flush tables to disk
DELETE ... ('bla',0);
flush tables to disk
DELETE ... ('bla',0);
flush tables to disk
```

In there you can see that there was a row in the first sstable and now we have 3 sstables. Whenever we want to read the value, we need to go through all these records which can add up over time. The particular use case may not be as problematic but, if we can avoid it, then all the better. And I think that, in all our use cases, we don't even need an extra DB lookup, do we?

libs/galley-types/src/Galley/Types/Teams.hs
tiago-loureiro: >Playing around with this option doesn't put galley in bad state.

Ah, totally right - thanks for the reminder.

services/galley/src/Galley/API/Teams.hs
akshaymankar: This makes a lot of sense, thanks for the input, I have added the checks to make sure we insert/delete only when required.

libs/galley-types/src/Galley/Types/Teams.hs
tiago-loureiro: To wrap up this topic: can we move this out of flags: the semantics of these flags is something like: they are something you turn on/off for specific teams, etc.

Let's move it then outside of the flags and give it a default plz?

akshaymankar: I have moved it.

libs/brig-types/test/unit/Test/Brig/Types/Common.hs
```diff
@@ -86,13 +86,27 @@ instance Show TeamMemberDeleteData where
   show a = "(TeamMemberDeleteData " <> show (a ^. tmdAuthPassword) <> ")"
 
 instance Arbitrary SSOStatus where
-  arbitrary = Test.Tasty.QuickCheck.elements [minBound ..]
```
tiago-loureiro, line 89: This will conflict with https://github.com/wireapp/wire-server/pull/1086 - @fisx can we reach a consensus here? :)

akshaymankar: Somehow I missed [this comment](https://github.com/wireapp/wire-server/pull/1086/files#r419934139) from @fisx yesterday. I think we can have this discussion in that PR. In all this PR doesn't really add/remove anything from galley/brig-types. So I will revert my refactoring

services/galley/test/integration/API/Teams.hs
```diff
@@ -1216,14 +1233,147 @@ testTeamAddRemoveMemberAboveThresholdNoEvents = do
           ((/= Galley.Types.Teams.Intra.Deleted) . Galley.Types.Teams.Intra.tdStatus)
           (getTeamInternal tid)
 
+testBillingInLargeTeam :: TestM ()
+testBillingInLargeTeam = do
+  (firstOwner, team) <- Util.createBindingTeam
+  refreshIndex
+  o <- view tsGConf
+  g <- view tsGalley
```
fisx, line 1241:
```suggestion
  opts <- view tsGConf
  galley <- view tsGalley
```

services/galley/test/integration/API/Util.hs
```diff
@@ -1349,3 +1350,29 @@ withSettingsOverrides :: MonadIO m => Opts.Opts -> WaiTest.Session a -> m a
 withSettingsOverrides opts action = liftIO $ do
   (galleyApp, _) <- Run.mkApp opts
   WaiTest.runSession action galleyApp
+
+waitForMemberDeletion :: UserId -> TeamId -> UserId -> TestM ()
+waitForMemberDeletion zusr tid uid = do
+  maybeTimedOut <- timeout 2000000 $ loop
+  case maybeTimedOut of
+    Nothing -> liftIO $ assertFailure "Timed out waiting for member deletion"
+    _ -> pure ()
```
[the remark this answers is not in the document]
akshaymankar: I applied it slightly differently so it looks nice without redundant do and ormolu :)

services/galley/test/integration/API/Teams.hs
akshaymankar: Had to rename at few more places, but it is done :)

PR #1082, opened 2020-04-28
don't poll from SQS as often

jschaul:
Relates to https://github.com/zinfra/backend-issues/issues/891

We currently poll SQS as fast as possible for three queues: two in brig and one in gundeck. For local development / integration tests, this creates a lot of unnecessary logs and high CPU usage for the fake-sqs container. On my machine, I get about 65 polling requests to fake SQS per second. This PR takes that down to 3 requests per second.

I'd like to 
* get a CI run though to see whether tests fail or not (whether any tests are time-critical when it comes to queing).
* get a review here to see whether the approach makes any sense: can we afford to be slower at querying SQS? I think we can, as neither the internal nor the email feedback or push notification feedback events need to be processed in the same split second. But I'd like to get a second opinion on this.

This PR is WIP, as the times are hardcoded to one second at the moment. I plan to make this threaddelay configurable. But what would be a good default for production (and for local/testing?)?

### DIFF
--- a/services/brig/src/Brig/AWS.hs
+++ b/services/brig/src/Brig/AWS.hs
@@ -173,10 +173,10 @@ instance Exception Error
 -- SQS
 
 listen :: (FromJSON a, Show a) => Text -> (a -> IO ()) -> Amazon ()
-listen url callback = do
-  forever $ handleAny unexpectedError $ do
-    msgs <- view rmrsMessages <$> send receive
-    void $ mapConcurrently onMessage msgs
+listen url callback = forever $ handleAny unexpectedError $ do
+  msgs <- view rmrsMessages <$> send receive
+  void $ mapConcurrently onMessage msgs
+  threadDelay 1000000
   where
     receive =
       SQS.receiveMessage url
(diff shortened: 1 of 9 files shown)


services/brig/src/Brig/AWS.hs
```diff
@@ -173,10 +173,10 @@ instance Exception Error
...
 
 listen :: (FromJSON a, Show a) => Text -> (a -> IO ()) -> Amazon ()
-listen url callback = do
-  forever $ handleAny unexpectedError $ do
-    msgs <- view rmrsMessages <$> send receive
-    void $ mapConcurrently onMessage msgs
+listen url callback = forever $ handleAny unexpectedError $ do
+  msgs <- view rmrsMessages <$> send receive
+  void $ mapConcurrently onMessage msgs
+  threadDelay 1000000
```
tiago-loureiro, line 179: How about sleeping only when nothing was consumed?

```suggestion
  when (null msgs) $
    threadDelay 1000000
```

tiago-loureiro:
One thing to point out is that this should not be happening with a real SQS queue. Looking at [this value](https://github.com/wireapp/wire-server/blob/develop/services/brig/src/Brig/AWS.hs#L183), the call should take 20 seconds if there is nothing there (if there is, it returns earlier).

> can we afford to be slower at querying SQS?

I think we can, yes. But when using a real queue, this is not needed - I think a sleep is not harmful here, but how about we sleep only if nothing was consumed, (i.e.: suggested inline). And on that note, we could default to 100ms? I guess the bottom line is that I think that we don't need realtime consumption there.

Sidenote: seems like the tests took just as long as they did before so that did not make a difference.

jschaul:
The throttle amount is now configurable, the context is put into a comment. Note that some of our environments do run with fake-sqs at the moment, where lots of unnecessary polling does occur. I set the default to 500ms and the value for local development to 1 second (where locally tests execute fine for me).
This is now ready for review.

services/brig/src/Brig/AWS.hs
jschaul: Makes sense, thanks!

PR #1083, opened 2020-04-29
Refactor: Split API modules into public/internal

mheinzel:
services not quite following the pattern:

1. **cargohold**: ~All endpoints are public here, so I'm not sure it makes sense to create a `Public` submodule. Maybe it does, for consistency's sake?~ done

2. **brig**: The endpoints here are not in a single `API` module, but there are some small ones as well. I didn't want to split each of these into multiple modules, so I just split them into `routesPublic` and `routesInternal`.


services/gundeck/src/Gundeck/API.hs
```diff
@@ -20,259 +20,19 @@ module Gundeck.API
...
-  -- Docs ------------------------------------------------------------------
-
-  get "/push/api-docs" (continue docsH) $
-    query "base_url" .&. accept "application" "json"
-  -- Status & Monitoring ---------------------------------------------------
-
+  Public.sitemap
+  Public.apiDocs
+  Internal.sitemap
   head "/i/status" (continue $ const (return empty)) true
```
jschaul, line 37: Why did you leave the `/i/status` out from the internal endpoints?

mheinzel: They seemed very generic to me, in that every service has some common non-domain-related endpoints. Seems to me like they should be separated from the domain endpoints and could even be abstracted out for re-use (although in this case there's no big benefit for just 2 trivial endpoints).

mheinzel: Also, in some services we don't have an `Internal` module, so we would have to create another module just for the status endpoint or it will be inconsistent.

mheinzel: Okay, I'll move them to the other internal endpoints.

jschaul: Thanks!

PR #1084, opened 2020-04-30
Manage license headers with headroom instead of licensure

mheinzel:
For the most parts, it's a drop in replacement, but it has some advantages.

It's a Haskell tool, so it can be installed (at exactly the version we want) through Stack, the same way as Ormolu.

Also, it knows that we want our Haskell headers after the pragmas (in line with Ormolu's style).
This way:
- Ormolu will already be happy after adding a header
- Headroom is idempotent even in conjunction with Ormolu, making it unnecessary to detect existing headers with grep
- Headroom will correctly update headers if they differ from the template (e.g. updating the year)

Thanks a lot to @vaclavsvejcar for adding the features required to make this work!


mheinzel:
Yes, we can add it right now by running Headroom and then checking if any files changed. Or even by checking for `0 modified` in the output. :stuck_out_tongue: 

There also is [an issue for a built-in command](https://github.com/vaclavsvejcar/headroom/issues/45), similar to Ormolu's `--mode check`. If that gets implemented, it would be less hacky.

akshaymankar:
> Yes, we can add it right now by running Headroom and then checking if any files changed. Or even by checking for `0 modified` in the output. stuck_out_tongue
> 
> There also is [an issue for a built-in command](https://github.com/vaclavsvejcar/headroom/issues/45), similar to Ormolu's `--mode check`. If that gets implemented, it would be less hacky.

Let's do it then :)

vaclavsvejcar:
Glad to hear Headroom works for you 👍  Feel free to create new issue if you'll have ideas for new features. I also expect that the [mentioned built-in command](https://github.com/vaclavsvejcar/headroom/issues/45) could be part of one of the next releases.

mheinzel:
Then I'll add a CI check once the command is implemented in Headroom.

PR #1086, opened 2020-05-04
Custom search visibility limit name search

tiago-loureiro:
Fixes https://github.com/zinfra/backend-issues/issues/1400

Requires confirmation but if the specified behaviour is final, the PR is good to go as is. The code is mostly boilerplate, except when doing search. It adds:

- [x] New search config option per team. Currently, the only other option than the standard is to prevent name search outside the ~~name~~ team
- [x] Possibility to toggle custom search visibility per team
- [x] Possibility to set search visibility at the team level, if enabled


libs/brig-types/src/Brig/Types/Search.hs
```diff
@@ -46,6 +46,7 @@ data TeamSearchInfo
   = -- | When searching user is not part of a team.
     NoTeam
   | -- | When searching user is part of a team and 'Brig.Options.setSearchSameTeamOnly' is True
+    --   OR the search user belongs to a team with SearchVisibilityNoNameOutsideTeam
```
fisx, line 49:
```suggestion
    --   OR the searching user belongs to a team with SearchVisibilityNoNameOutsideTeam
```

docs/reference/config-options.md
```diff
@@ -48,3 +49,10 @@ IMPORTANT: If you switch this back to `disabled-permanently` from
 `disabled-by-default`, LegalHold devices may still be active in teams
 that have created them while it was allowed.  This may change in the
 future.
+
+### Custom Search Visibility
```
jschaul, line 53: I don't understand what "customer search visibility" means, what are the effects when this is enabled or disabled? Who can see what when searching if this is enabled or disabled? Could you please describe this? (also, naming wise, "custom search visibility" doesn't give much idea on what this is about).

docs/reference/config-options.md
```diff
@@ -48,3 +49,10 @@ IMPORTANT: If you switch this back to `disabled-permanently` from
...
+
+This sets the default setting for all teams, and can be overridden by
+customer support / backoffice. [Allowed
+values](https://github.com/wireapp/wire-server/blob/custom-search-visibility-limit-name-search/libs/galley-types/src/Galley/Types/Teams.hs#L382-L385):
+`disabled-by-default`, `enabled-by-default`.
```
jschaul, line 58: CustomSearchVisibilityType has clearer semantics - can you use clearer semantics also at the config level? If we add another setting with different logic, then "disabled" and "enabled" doesn't make sense anymore.

tiago-loureiro: This is only for enabling/disabling the feature, not to set any value which is why it follows the logic behind the other feature flags. I.e., you may turn this feature on/off using the backoffice but the actual value itself is to be set independently (a bit like SSO and Legahold with settings).

docs/reference/config-options.md
tiago-loureiro: Hmmm, I guess it's "clear" that this is confusing. This is only meant to allow the team to set a specific value, i.e., one of [these](https://github.com/wireapp/wire-server/blob/custom-search-visibility-limit-name-search/libs/galley-types/src/Galley/Types/Teams/SearchVisibility.hs#L28).

services/brig/src/Brig/API/Public.hs
```diff
@@ -1030,25 +1030,22 @@ getUserDisplayNameH (_ ::: self) = do
     Just n -> json $ object ["name" .= n]
     Nothing -> setStatus status404 empty
 
-listUsersH :: JSON ::: UserId ::: Either (List OpaqueUserId) (List (OptionallyQualified Handle)) -> Handler Response
+listUsersH :: JSON ::: UserId ::: Either (List OpaqueUserId) (Range 1 4 (List (OptionallyQualified Handle))) -> Handler Response
```
tiago-loureiro, line 1033: Motivation for this change is that the check for handles is rather expensive - and clients do not use this. Checking with the different client teams I got the information that only 1 handle is ever looked up.

docs/reference/config-options.md
tiago-loureiro: Would it be more clear if named something like: `allowChangingSearchVisibility`?

services/brig/src/Brig/API/Public.hs
fisx: b7f9e02c88ec81066c4d5af1f629b6eccc00f723

services/galley/schema/src/Main.hs
```diff
@@ -65,9 +66,10 @@ main = do
       V34.migration,
       V35.migration, -- This should safe to do now
       V36.migration,
-      V37.migration
+      V37.migration,
       -- When adding migrations here, don't forget to update
       -- 'schemaVersion' in Galley.Data
+      V39.migration
```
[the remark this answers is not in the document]
fisx: f4badb6

services/galley/src/Galley/API/Teams.hs
```diff
@@ -650,6 +659,20 @@ deleteTeamConversation zusr zcon tid cid = do
   -- bot user can only be in a single conversation
   Data.removeTeamConv tid cid
 
+-- | Modify and get visibility type for a team
```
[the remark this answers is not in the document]
fisx:
```suggestion
```

I agree, but the comment also doesn't add anything to the name of the function; it just makes room for inconsistencies and increases noise.

libs/galley-types/src/Galley/Types/Teams.hs
```diff
@@ -552,7 +575,8 @@ hiddenPermissionsFromPermissions =
           (roleHiddenPerms RoleMember <>) $
             Set.fromList
               [ ChangeLegalHoldTeamSettings,
-                ChangeLegalHoldUserSettings
+                ChangeLegalHoldUserSettings,
+                ChangeCustomSearchVisibility
```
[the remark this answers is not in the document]
tiago-loureiro: This is for admins only indeed, this is the permission set for admins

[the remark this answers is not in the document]
fisx: Right.  I failed to understand my own code...  :)

fisx:
I didn't expect this:

```
      feature flags:                                                                             W, Your journaling events for teams larger than 18 may have some admin user ids missing.  This is fine for testing purposes but NOT for production use!!
W, Your journaling events for teams larger than 18 may have some admin user ids missing.  This is fine for testing purposes but NOT for production use!!
FAIL
        Exception: Assertions failed:
         2: Right (CustomSearchVisibilityTeamConfig {customSearchVisibilityTeamConfigStatus = CustomSearchVisibilityDisabled}) =/= Right (CustomSearchVisibilityTeamConfig {customSearchVisibilityTeamConfigStatus = CustomSearchVisibilityEnabled})

        Response was:

        Response {responseStatus = Status {statusCode = 200, statusMessage = "OK"}, responseVersion = HTTP/1.1, responseHeaders = [("Content-Type","application/json")], responseBody = Just "{\"status\":\"disabled\"}", responseCookieJar = CJ {expose = []}, responseClose' = ResponseClose}                                          
          error, called at src/Bilge/Assert.hs:88:28 in bilge-0.22.0-5aPm7hcwMbW8hs9PsPQhoJ:Bilge.Assert     
          <!!, called at src/Bilge/Assert.hs:105:19 in bilge-0.22.0-5aPm7hcwMbW8hs9PsPQhoJ:Bilge.Assert      
          !!!, called at test/integration/API/Teams.hs:1811:44 in main:API.Teams                             
          getCustomSearchVisibility, called at test/integration/API/Teams.hs:1831:5 in main:API.Teams        
    Per-conversation message timer
```

Looks like the new galley option dose not get the consideration it deserves?

I also want to rename a few things and improve the docs, but otherwise :+1: 

We also have too much boilerplate in our code!  But that's a story for a different day.

tiago-loureiro:
> Looks like the new galley option dose not get the consideration it deserves?

What do you mean? :)

fisx:
> Looks like the new galley option dose not get the consideration it deserves?

No, sorry, it was my mistake.  I broke the tests when i tried to make them more thorough.

fisx:
integration tests pass locally; the ci error is because the galley config file hasn't been updated yet.

docs/reference/config-options.md
```diff
@@ -48,3 +49,14 @@ IMPORTANT: If you switch this back to `disabled-permanently` from
 `disabled-by-default`, LegalHold devices may still be active in teams
 that have created them while it was allowed.  This may change in the
 future.
+
+### Team Search Visibility
+
+Is a team allowed to change its team search visibility settings?  If a
+team is enabled, it can be configured so other teams are completely
+invisible for search.
```
tiago-loureiro, line 57:
```suggestion
Is a team allowed to change its team search visibility settings? If enabled
for the team, it can be configured so that non team users do not show up in search.
```

services/galley/schema/src/Main.hs
```diff
@@ -63,9 +64,10 @@ main = do
       V32.migration,
       V33.migration,
       V34.migration,
-      V35.migration, -- This should safe to do now
+      V35.migration,
       V36.migration,
-      V37.migration
+      V37.migration,
+      V39.migration
```
tiago-loureiro, line 70: Note that here I used V39 as I'd like https://github.com/wireapp/wire-server/pull/1081 to get merged first and then can readjust this in an easier way.

services/galley/src/Galley/API/Public.hs
```diff
@@ -407,6 +433,26 @@ sitemap = do
...
+    zauthUserId
+      .&. capture "tid"
+      .&. accept "application" "json"
+  document "GET" "getTeamSearchVisibilityEnabled" $ do
+    summary "Shows whether Custom Search Visibility feature is enabled for team"
+    parameter Path "tid" bytes' $
+      description "Team ID"
+    returns (ref Model.teamSearchVisibilityEnabled)
+    response 200 "Search Visibility status" end
+  get "/custom-backend/by-domain/:domain" (continue CustomBackend.getCustomBackendByDomainH) $
```
tiago-loureiro, line 446: Not sure how this ended up here, it's not new in this PR?

services/galley/src/Galley/API/Public.hs
```diff
@@ -407,6 +433,26 @@ sitemap = do
...
+    capture "domain"
+      .&. accept "application" "json"
+  document "GET" "getCustomBackendByDomain" $ do
+    summary "Shows information about custom backends related to a given email domain"
+    parameter Path "domain" string' $
+      description "URL-encoded email domain"
+    returns (ref Model.customBackend)
+    response 200 "Custom backend" end
```
tiago-loureiro, line 454:
```suggestion
```

libs/galley-types/src/Galley/Types/Teams/SearchVisibility.hs
```diff
@@ -0,0 +1,98 @@
...
+
+instance FromJSON TeamSearchVisibilityView where
+  parseJSON = withObject "TeamSearchVisibilityView" $ \o ->
+    TeamSearchVisibilityView <$> o .: "search_visibility"
+
+-- Status of the feature
+
+-- | Is the feature enabled for a given team?  See also 'FeatureTeamSearchVisibility',
+-- 'TeamSearchVisibility'.
+data TeamSearchVisibilityEnabled = TeamSearchVisibilityDisabled | TeamSearchVisibilityEnabled
```
tiago-loureiro, line 77:
```suggestion
data TeamSearchVisibilityAvailable = TeamSearchVisibilityDisabled | TeamSearchVisibilityEnabled
```

Maybe the suggestion isn't the best but I find having the type name as the same as one of the constructors... how do you like the suggestion?

tiago-loureiro:
@fisx - other than the 2 inline suggestions, looks good 👍

services/galley/src/Galley/API/Public.hs
fisx: I was wondering myself, but should have wondered harder.  fixed here: e8786e7

libs/galley-types/src/Galley/Types/Teams/SearchVisibility.hs
fisx: 5a153e214f7e9f7153c1d21ae1b218f2c53a5c3e

I can't say I like it more, but I don't like it much less either.  You decide if you want to revert it.  :)

libs/galley-types/src/Galley/Types/Teams.hs
```diff
@@ -376,17 +379,25 @@ data FeatureLegalHold
   | FeatureLegalHoldDisabledByDefault
   deriving (Eq, Ord, Show, Enum, Bounded, Generic)
 
+data FeatureCustomSearchVisibility
+  = FeatureCustomSearchVisibilityEnabledByDefault
+  | FeatureCustomSearchVisibilityDisabledByDefault
+  deriving (Eq, Ord, Show, Enum, Bounded, Generic)
```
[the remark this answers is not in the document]
akshaymankar: Looks like that commit went away before I got a chance to look at it. I think we shouldn't keep doing bad things while waiting for us to move to servant. This kind of stuff doesn't have to have its own issue, if it does, it will probably never get prioritized. Instead we can just incrementally refactor it out of our code.
I am not sure how servant will solve this, but I am also not very hopeful that we'll get time to move to servant so soon.

> Can we open an issue on that and let it slip through for now?

If this is time sensitive, I guess it is not too much debt to be taking on.

fisx:
Before we remove the `[WIP]`, we should update concourse so that CI has a chance of passing.

All tests pass locally.

libs/galley-types/src/Galley/Types/Teams.hs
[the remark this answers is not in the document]
fisx: https://github.com/zinfra/backend-issues/issues/1407

I think this is worth its own issue just for the discussion.

> If this is time sensitive, I guess it is not too much debt to be taking on.

I would appreciate that.  :)

But if you make it go away here I won't fight you.

tiago-loureiro:
> Before we remove the `[WIP]`, we should update concourse so that CI has a chance of passing.
> 
> All tests pass locally.

Are we happy with the new config name?

docs/reference/config-options.md
akshaymankar: Why do we need this feature flag?

fisx:
> Are we happy with the new config name?

I am if nobody else objects!  :)

docs/reference/config-options.md
fisx: > Why do we need this feature flag?

I suppose we could just go with making this configurable in every team, and not have a flag that turns this off i `galley.yaml`.  But I think there was a reason for this.  @tiago-loureiro ?

PR #1090, opened 2020-05-06
Refactor: Create wire-api package for types used in the public API

mheinzel:
See https://github.com/zinfra/backend-issues/issues/1416.

This allows a clearer separation between the public interface and internal data structures. This is becoming more important with the upcoming federation changes, which will change internal representations while keeping a backwards-compatible API.

Things coming with this PR:

- copied all types used in the interface/signature of public endpoints into a new `wire-api` package (mostly client API, but also some bot/provider stuff)
- split long modules into smaller, more organized ones with a clearer structure
- explicit export lists
- Swagger models and JSON instances are next to the respective type definitions, making it easier to keep them in sync

I do not intend to change any definitions, just move them around (+ some formatting). This is supposed to be a pure refactoring.

Steps needed before this can be merged:

- [x] make `wire-api` tests compile
- [x] check in `wire-api.cabal` file
- [x] remove stuff from `*-types` that I copied over, still (re-)exporting it (to avoid diffs in import sites)
- [x] fix one TODO about which typeclass instance deriving strategy to use
- [x] rebase/merge develop, resolving conflicts

Cleanup I would like to do if there is time:
- [x] remove unused language extensions (introduced by copy-paste)
- [x] remove unnecessary package dependencies I introduced
- [x] add StrictData and remove all strictness annotations in types
- [x] make use of file "separators" consistent (lines made out of dashes)


fisx:
nice!

haven't read much of your code yet, but i already have two opinions:

- do we even need `types-common`, `*-types` packages anymore?  note that the services themselves are libraries, so for testing we can easily import them and all their types.  for production code of service A, we shouldn't need anything from service B other than the interface types, no?  (this is assuming that you want to keep internal interfaces also in wire-api.)

- I'm wondering about Cql instances.  I think the cleanest approach would be to wrap them in newtypes in the resp. service packages (brig the library etc.), and keep the instances where they are needed without leaking the cql package dependency anywhere.  If that is too much boilerplate, we can keep them with the types, and later decide to make the package dependency optional with a cabal flag.

fisx:
(i think making a package `wire-api-internal` that depends on `wire-api` would also work.)

mheinzel:
> (this is assuming that you want to keep internal interfaces also in wire-api.)

> (i think making a package wire-api-internal that depends on wire-api would also work.)

Right now, I don't want to change more than I already do, so I want to keep imports of `*.Types` intact. Otherwise, I'd pretty much have to touch every single file of the services.

Also, conceptually there are two different kinds of internal endpoints:
1. endpoints for admin/debugging tasks, to be called from outside the cluster
2. endpoints for intra-cluster communication between services

The first one might make sense to expose, either in `wire-api` or a separate `wire-api-internal`. Not sure if we want to do this soon, as I don't have a clear picture how all of these endpoints are used. If we change these types (they are not in the public API after all), how annoying will it be for the people/tools using them?

The second kind is more of an implementation detail of how we split our services. I think it should be in a per-service API definition that gets used by other services for internal calls (in the future using a generated Servant client).
 
> I think the cleanest approach would be to wrap them in newtypes in the resp. service packages (brig the library etc.), and keep the instances where they are needed without leaking the cql package dependency anywhere.

I think we want them to live in the services, yes. If newtypes are too much boilerplate, we can also consider orphan instances. :shrug:

arianvp:
> I think the cleanest approach would be to wrap them in newtypes in the resp. service packages (brig the library etc.), and keep the instances where they are needed without leaking the cql package dependency anywhere.

You do not want to couple your API types to your Cql datatypes because that means changing your API forces you to change your database. These things should be decoupled.

Either a newtype or  preferably separate datatype sounds like the cleanest approach to me. I would avoid orphan instances

fisx:
>  add StrictData and remove all strictness annotations in types

That i would have liked to see in a separate PR, but it's certainly a good change.

libs/galley-types/test/unit/Test/Galley/Types.hs
```diff
@@ -28,6 +28,7 @@ import Imports
 import Test.Tasty
 import Test.Tasty.HUnit
 
+-- wire-api: we still need this
```
fisx, line 31: I don't understand this comment.  Do you mean this?:

```suggestion
-- TODO: move this to wire-api
```

libs/wire-api/src/Wire/API/Arbitrary.hs
```diff
@@ -0,0 +1,558 @@
...
+import Wire.API.User.Activation
+import Wire.API.User.Auth
+import Wire.API.User.Client.Prekey
+import Wire.API.User.Handle
+import Wire.API.User.Identity
+import Wire.API.User.Password
+import Wire.API.User.Profile
+import Wire.API.User.RichInfo
+
+newtype Octet = Octet {octet :: Word16}
```
fisx, line 73: I've seen this when working on `brig-types`, but is this even used anywhere?

```
57 (0) develop 13:08:39 mf@lag:~/src/wire-server
$ git grep -Hn Octet
libs/brig-types/src/Brig/Types/Test/Arbitrary.hs:73:newtype Octet = Octet {octet :: Word16}
libs/brig-types/src/Brig/Types/Test/Arbitrary.hs:76:instance Arbitrary Octet where
libs/brig-types/src/Brig/Types/Test/Arbitrary.hs:77:  arbitrary = Octet <$> arbitrary `suchThat` (< 256)
tools/bonanza/test/unit/Test/Bonanza/Arbitrary.hs:176:newtype Octet = Octet {octet :: Word16}
tools/bonanza/test/unit/Test/Bonanza/Arbitrary.hs:179:instance Arbitrary Octet where
```

services/brig/src/Brig/API/Public.hs
```diff
@@ -665,6 +665,7 @@ sitemap o = do
       \number is not whitelisted, a 403 error is returned."
     Doc.body (Doc.ref Doc.newUser) $
       Doc.description "JSON body"
+    -- FUTUREWORK: I think this should be 'Doc.self' instead of 'user'
```
fisx, line 668: If so, I think that's a bug and should not be forgotten in a `FUTUREWORK`.

libs/wire-api/test/unit/Test/Wire/API/Roundtrip.hs
```diff
@@ -0,0 +1,36 @@
...
+import Imports
+import Test.Tasty (TestTree)
+import Test.Tasty.QuickCheck ((===), Arbitrary, counterexample, testProperty)
+import Type.Reflection (typeRep)
+
+testRoundTrip ::
+  forall a.
+  (Arbitrary a, Typeable a, ToJSON a, FromJSON a, Eq a, Show a) =>
+  TestTree
+testRoundTrip = testProperty msg trip
```
fisx, line 31: Where is this called?

services/galley/src/Galley/API/Public.hs
```diff
@@ -1008,6 +1010,8 @@ docs (_ ::: url) = do
   let apidoc = encode $ mkSwaggerApi (decodeLatin1 url) models sitemap
   pure $ responseLBS status200 [jsonContent] apidoc
 
+-- FUTUREWORK: /teams/api-docs does not get queried by zwagger-ui
```
fisx, line 1013: Is that the one generated by servant-swagger?  In that case I think it's added to the swagger ui differently.  (This line is stil useful, just trying to add insight.)

libs/galley-types/test/unit/Test/Galley/Types.hs
mheinzel: ah, forgot to remove this one. It was just a reminder for myself that this test needs to stay where it is.

services/brig/src/Brig/API/Public.hs
mheinzel: We have a lot of Swagger that's not up to date (or not served in swagger-ui at all). This one is relatively harmless: it just doesn't tell you about some fields it will return (`user` is a subset of `self`).

The real solution here is to make the Swagger generation less manual, e.g. with Servant.

libs/wire-api/test/unit/Test/Wire/API/Roundtrip.hs
mheinzel: Nowhere yet. I'll move the roundtrip tests over in another PR and remove this definition for now.

services/galley/src/Galley/API/Public.hs
mheinzel: Yes, that's the one and I can't find it on swagger-ui.

I think it just needs to be added [here](https://github.com/wireapp/wire-server/blob/develop/services/nginz/zwagger-ui/api-docs/resources.json), but let's do that as part of a general Swagger cleanup.

libs/wire-api/src/Wire/API/Arbitrary.hs
mheinzel: It was used in the instance below, but I just inlined it.

mheinzel:
Good questions.

>     1. So what is the purpose of `types-common` in contrast to `wire-api`?  Why aren't they just one package?

Right now, as we are depending on `types-common`, it needs to be considered part of the public API. In the future, I would like to get rid of some dependencies and move the rest into `wire-api`, but I think that's still up for debate.

>     2. And what is the purpose of `*-types` packages in the future?  Can they be resolved into `types-common` or `wire-api`?  If not, how are they conceptually separate?

At the moment their purpose is a bit unclear and they mostly exist in their current form to avoid changing too much code at once.

For the future: Types used in the public API belong in `wire-api` and completely service-internal types could be moved into the services, so what remains would be types used for internal and inter-service endpoints. Maybe a name `galley-api` etc. would be more fitting then.

services/galley/src/Galley/API/Public.hs
fisx: servant-swagger generates swagger-2.0, and the rest of our swagger is 1.2.  therefore, our swagger-ui consists of two pages, and what goes where is decided based on how it is generated.  the idea was to move everything to the second page and drop the first quickly, but, you know.

fisx:
> > 1. And what is the purpose of `*-types` packages in the future?  Can they be resolved into `types-common` or `wire-api`?  If not, how are they conceptually separate?
> 
> At the moment their purpose is a bit unclear and they mostly exist in their current form to avoid changing too much code at once.
> 
> For the future: Types used in the public API belong in `wire-api` and completely service-internal types could be moved into the services, so what remains would be types used for internal and inter-service endpoints. Maybe a name `galley-api` etc. would be more fitting then.

I still prefer to add interal apis to wire-api in `Intra`-modules (it's a bit different from the `Internal` modules found in many libraries).

Otherwise :+1:

PR #1095, opened 2020-05-11
More tests for OTR messages using protobuf

tiago-loureiro:
No new functionality in the PR. Just more tests that I found useful to add when checking the logic behind `report_missing` and how all that works

### DIFF
--- a/services/galley/test/integration/API.hs
+++ b/services/galley/test/integration/API.hs
@@ -326,6 +326,7 @@ postCryptoMessage5 = do
   conv <- decodeConvId <$> postConv alice [bob, eve] (Just "gossip") [] Nothing Nothing
   -- Missing eve
   let m = [(bob, bc, "hello bob")]
+  let m' = otrRecipients [(bob, [(bc, encodeCiphertext "hello bob")])]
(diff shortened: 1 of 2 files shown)


tiago-loureiro:
> Can this somehow be tested here?: https://github.com/wireapp/wire-server/blob/eede068cddb68b55cd1827a98f17b385c96d7ce9/services/galley/test/unit/Test/Galley/API.hs

This test only checks that the parameters passed on to `postOtrMessage` are correct. I.e., I want to make sure that certain assumptions are actually correct and currently we need to access the database for that - note that those response codes depend on specific states for those users' clients

> We're not really testing the effect of the calls to postOtrMessage', so we? So the only thing that is tested here is that sending these queries is not an error?

We are; getting a 201 or 412 will depend on the state of the DB (users in a conv / users in a team / devices of a user).

fisx:
> > We're not really testing the effect of the calls to postOtrMessage', so we? So the only thing that is tested here is that sending these queries is not an error?
> 
> We are; getting a 201 or 412 will depend on the state of the DB (users in a conv / users in a team / devices of a user).

ok, then.  it could use a comment, or not.  otherwise good to go!

PR #1096, opened 2020-05-12
Add tool to migrate data for galley

akshaymankar:
Fixes https://github.com/zinfra/backend-issues/issues/1422
One this is merged and we start creating docker images for each build, we can integrate it with helm charts using post-install hooks.

### DIFF
--- a/services/galley/Makefile
+++ b/services/galley/Makefile
@@ -10,6 +10,7 @@ EXE_SCHEMA    := ../../dist/$(NAME)-schema
 DEB           := $(NAME)_$(VERSION)+$(BUILD)_amd64.deb
 DEB_IT        := $(NAME)-integration_$(VERSION)+$(BUILD)_amd64.deb
 DEB_SCHEMA    := $(NAME)-schema_$(VERSION)+$(BUILD)_amd64.deb
+DEB_MIGRATE_DATA := $(NAME)-migrate-data_$(VERSION)+$(BUILD)_amd64.deb
 EXECUTABLES   := $(NAME) $(NAME)-integration $(NAME)-schema
 DOCKER_USER    ?= quay.io/wire
 DOCKER_TAG     ?= local
@@ -45,7 +46,7 @@ compile:
 	stack build . --pedantic --test --bench --no-run-benchmarks --no-copy-bins
 
 .PHONY: dist
-dist: guard-VERSION install $(DEB) $(DEB_IT) $(DEB_SCHEMA) .metadata
+dist: guard-VERSION install $(DEB) $(DEB_IT) $(DEB_SCHEMA) $(DEB_MIGRATE_DATA) .metadata
 
 $(DEB):
 	makedeb --name=$(NAME) \
@@ -67,6 +68,14 @@ $(DEB_SCHEMA):
 	makedeb --name=$(NAME)-schema \
 	 --version=$(VERSION) \
 	 --debian-dir=deb-schema \
+	 --build=$(BUILD) \
+	 --architecture=amd64 \
+	 --output-dir=dist
+
+$(DEB_MIGRATE_DATA):
+	makedeb --name=$(NAME)-migrate-data \
+	 --version=$(VERSION) \
+	 --debian-dir=deb-migrate-data \
 	 --build=$(BUILD) \
 	 --architecture=amd64 \
 	 --output-dir=dist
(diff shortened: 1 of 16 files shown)


services/galley/migrate-data/src/V1_BackfillBillingTeamMembers.hs
```diff
@@ -0,0 +1,83 @@
...
+
+createBillingTeamMembers :: MonadClient m => [(TeamId, UserId)] -> m ()
+createBillingTeamMembers pairs =
+  retry x5 $ batch $ do
+    setType BatchLogged
+    setConsistency Quorum
+    mapM_ (addPrepQuery cql) pairs
+  where
+    cql :: PrepQuery W (TeamId, UserId) ()
+    cql = "INSERT INTO billing_team_member (team, user) values (?, ?)"
```
jschaul, line 79: This is an anti-pattern of the use of batch statements as team Ids will mostly be different. Don't do this. Also read https://docs.datastax.com/en/cql-oss/3.3/cql/cql_using/useBatchBadExample.html (and more generally https://docs.datastax.com/en/cql-oss/3.3/cql/cql_using/useBatchTOC.html)

tiago-loureiro:
Should we also create a galley-migrations debian package https://github.com/wireapp/wire-server/tree/develop/services/galley to be able to run this on non-k8s envs?

services/galley/migrate-data/src/V1_BackfillBillingTeamMembers.hs
akshaymankar: I didn't know this, I watch out for it in future, thanks! I fixed it by not chunking the write at all, I did it only because I copied it from other places where we backfill.

akshaymankar:
@tiago-loureiro Good point, I have added packaging for deb too.

services/galley/Makefile
```diff
@@ -1,18 +1,19 @@
...
+BUILD_NUMBER     ?= 0
+BUILD_LABEL      ?= local
+BUILD            := $(BUILD_NUMBER)$(shell [ "${BUILD_LABEL}" == "" ] && echo "" || echo ".${BUILD_LABEL}")
+EXE_IT           := ../../dist/$(NAME)-integration
+EXE_SCHEMA       := ../../dist/$(NAME)-schema
+DEB              := $(NAME)_$(VERSION)+$(BUILD)_amd64.deb
+DEB_IT           := $(NAME)-integration_$(VERSION)+$(BUILD)_amd64.deb
+DEB_SCHEMA       := $(NAME)-schema_$(VERSION)+$(BUILD)_amd64.deb
+DEB_MIGRATE_DATA := $(NAME)-migrate-data_$(VERSION)+$(BUILD)_amd64.deb
+EXECUTABLES      := $(NAME) $(NAME)-integration $(NAME)-schema
```
tiago-loureiro, line 14: Do we want to add the new executable here?

tiago-loureiro:
How about a Makefile target for convenience?
```
EXE_MIGRATE_DATA       := ../../dist/$(NAME)-migrate-data

[...]

.PHONY: db-migrate-data
db-migrate-data: fast
	$(EXE_MIGRATE_DATA) --keyspace $(NAME)_test
```

tiago-loureiro:
Other than the 2 comments 👍

akshaymankar:
> How about a Makefile target for convenience?
> 
> ```
> EXE_MIGRATE_DATA       := ../../dist/$(NAME)-migrate-data
> 
> [...]
> 
> .PHONY: db-migrate-data
> db-migrate-data: fast
> 	$(EXE_MIGRATE_DATA) --keyspace $(NAME)_test
> ```

Do we really want to do this locally?

services/galley/Makefile
akshaymankar: I had no idea we have another way of creating docker images. I've added it :+1:

tiago-loureiro:
> > How about a Makefile target for convenience?
> > ```
> > EXE_MIGRATE_DATA       := ../../dist/$(NAME)-migrate-data
> > 
> > [...]
> > 
> > .PHONY: db-migrate-data
> > db-migrate-data: fast
> > 	$(EXE_MIGRATE_DATA) --keyspace $(NAME)_test
> > ```
> 
> Do we really want to do this locally?

As a way to quickly test it, why not? The same way we do `db-reset` or `db-migrate` ?

PR #1098, opened 2020-05-13
Move and add wire-api JSON roundtrip tests

mheinzel:
- [x] move Arbitrary instances next to type definitions
- [x] write more Arbitrary instances, often generically derived
- [x] copy existing roundtrip tests for types in wire-api
- [x] add missing roundtrip tests
- [x] fix some failing ones now, leave some for later (FUTUREWORK)
- [x] remove copied tests from other packages
- [x] rebase


libs/wire-api/src/Wire/API/Call/TURN.hs
```diff
@@ -109,13 +112,19 @@ modelRtcConfiguration = Doc.defineModel "RTCConfiguration" $ do
     Doc.description "Number of seconds after which the configuration should be refreshed (advisory)"
 
 instance ToJSON RTCConfiguration where
-  toEncoding (RTCConfiguration srvs ttl) =
-    pairs ("ice_servers" .= srvs <> "ttl" .= ttl)
+  toJSON (RTCConfiguration srvs ttl) =
```
mheinzel, line 115: There was a potential bug waiting here: The [documentation of ToJSON](https://www.stackage.org/haddock/lts-15.13/aeson-1.4.7.1/Data-Aeson.html#t:ToJSON) says:

> Instances in general must specify `toJSON` and should (but don't need to) specify `toEncoding`.

This is annoyingly not enforced by the compiler and leads to a broken `toJSON` implementation that violates the roundtrip property.

In prod, everything works because the handlers call `toEncoding` directly, but an innocent refactoring could break this.

fisx: So...  Write more QC tests for all instances whether `toEncoding` and `toJSON` are in sync?  (That sounds like a bit of a waste...)

mheinzel: Normal roundtrip tests guarantee that `toJSON` is implemented correctly and the default implementation of `toEncoding` uses that. But yeah, if both are implemented manually, we don't check whether they are in sync.

We could, but I'm not sure it's worth it. Then we should probably just not implement `toEncoding`. I don't think the performance benefit is significant enough to justify the additional code (and risk). Some day, the instances will be derived generically anyways, with a correct and efficient `toEncoding`.

mheinzel: The only place where we still use `toEncoding` is the instance for `List1`, and it seems to be correct there.

mheinzel:
The unit tests failed because a test timed out. The `Arbitrary` instance for `Map`s seems to create pretty big values and when running a lot of builds and tests in parallel on a slowish machine the duration increases from ~3s to over a minute. :(

I just fixed this one particular test by scaling down the `size` parameter, so we can see whether everything else is fine. Will come up with a better solution soon.

libs/types-common/src/Data/Misc.hs
```diff
@@ -251,6 +268,9 @@ instance Cql HttpsUrl where
   fromCql (CqlBlob t) = runParser parser (toStrict t)
   fromCql _ = fail "HttpsUrl: Expected CqlBlob"
 
+instance Arbitrary HttpsUrl where
+  arbitrary = pure $ HttpsUrl [URI.QQ.uri|https://example.com|]
```
fisx, line 272: I wrote this, I know, but why not:

```suggestion
  arbitrary = do
    host <- (<> ".") . ("host" <>) . pack . show . abs <$> (arbitrary @Int)
    pure 
      . HttpsUrl
      . (authorityL . _Just . authorityHostL . hostBSL %~ (host <>))
      $ [URI.QQ.uri|https://example.com|]
```

libs/types-common/src/Data/Misc.hs
```diff
@@ -277,12 +298,19 @@ instance Cql (Fingerprint a) where
...
+        "\138\140\183\EM\226#\129\EOTl\161\183\246\DLE\161\142\220\239&\171\241h|\\GF\172\180O\129\DC1!\159"
+
 --------------------------------------------------------------------------------
 -- Password
 
 newtype PlainTextPassword = PlainTextPassword
   {fromPlainTextPassword :: Text}
-  deriving (Eq, ToJSON, Generic)
+  deriving stock (Eq, Generic)
+  deriving newtype (ToJSON)
```
fisx, line 313:
```suggestion
  deriving newtype (ToJSON)  -- FUTUREWORK: this should never be serialized
```

libs/wire-api/src/Wire/API/Call/TURN.hs
fisx: This looks like one of those conversations that we'll come back to after something has gone wrong, and think: "yes, we should have acted on that thought."  :)

But I'm not sure how.  Write an hlint rule that disallows manual implementations of `toEncoding`?

mheinzel: For that we'd need to run hlint first of all. I don't think there is a quick and easy solution to this.

libs/types-common/src/Data/Code.hs
```diff
@@ -35,14 +35,17 @@ import Data.Scientific (toBoundedInteger)
 import Data.Text.Ascii
 import Data.Time.Clock
 import Imports
+import Test.QuickCheck (Arbitrary (arbitrary))
 
 -- | A scoped identifier for a 'Value' with an associated 'Timeout'.
 newtype Key = Key {asciiKey :: Range 20 20 AsciiBase64Url}
-  deriving (Eq, Show, FromJSON, ToJSON, FromByteString, ToByteString)
+  deriving (Eq, Show)
+  deriving newtype (FromJSON, ToJSON, FromByteString, ToByteString, Arbitrary)
```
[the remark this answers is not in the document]
mheinzel: But it doesn't change the instance. Which deriving strategy do you think it used before? It will only use anyclass if `LANGUAGE DeriveAnyClass` is enabled, which it isn't (and even then it would fail, since the default instance uses `Generic`, of which there is no instance).

So I think it didn't change the behaviour, however, I've gotten confused about this a few times before, so I might be missing something. This confusion is exactly the reason for using explicit deriving strategies. We should always do it and enforce this (which brings us back to using hlint).

libs/types-common/src/Data/Code.hs
```diff
@@ -67,6 +70,9 @@ instance FromJSON Timeout where
           (pure . Timeout . fromIntegral)
           t
 
+instance Arbitrary Timeout where
+  arbitrary = Timeout . fromIntegral <$> arbitrary @Int
```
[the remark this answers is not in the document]
mheinzel: That was the existing instance. :shrug: 

`NominalDiffTime` could potentially be negative. If we want that a `Timeout` is positive we should use a different type or use smart constructors. Also, the JSON instance parses negative numbers into a `Timeout`, too.

So unless we are sure the value can never be negative, I wouldn't put that invariant into the generator.

libs/types-common/src/Data/Misc.hs
```diff
@@ -165,9 +177,13 @@ instance ToJSON Location where
 
 instance FromJSON Location where
   parseJSON = withObject "Location" $ \o ->
-    location <$> (Latitude <$> o .: "lat")
+    location
+      <$> (Latitude <$> o .: "lat")
       <*> (Longitude <$> o .: "lon")
 
+instance Arbitrary Location where
+  arbitrary = Location <$> arbitrary <*> arbitrary
```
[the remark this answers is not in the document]
mheinzel: It can be, but the generic instance is currently defined in `wire-api`, which is downstream. Didn't seem very high priority to me, but we can definitely move the `GenericUniform` newtype + instance to `types-common` instead.

libs/types-common/src/Data/Misc.hs
mheinzel: Not sure how much better this is, as it still only covers trivial cases. I'd rather make that obvious by using a plain `pure x` instance. A proper instance would be even nicer of course, but I think that would be quite a bit of work.

libs/types-common/src/Data/Misc.hs
mheinzel: I can try removing this instance. We won't get the roundtrip test, but I think in this case it's alright. If I run into trouble, I can write `toJSON = const (toJSON "hunter2")` and the same in the `Arbitrary` instance. What do you think?

libs/wire-api/src/Wire/API/Call/TURN.hs
mheinzel: I'll also send a short announcement that `toEncoding` does not make `toJSON`  unnecessary.

libs/types-common/src/Data/Misc.hs
mheinzel: There is actually a case where we want to send this, when we generate an initial password on the server, see `NewProviderResponse'.

libs/types-common/src/Data/Misc.hs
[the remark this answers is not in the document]
mheinzel: I decided not to move it for now, since the benefit for these one or two instances is very small (and I don't want to touch most of types-common either). Also, I discovered that it would lead to conflict in another branch I have on top of this one.

libs/types-common/src/Data/Misc.hs
fisx: the benefit of having different ones is that it covers tests where we compare two urls that are supposed to be different --

ok, i just finished that thought, and it doesn't make sense anymore.  never mind.  :)

libs/types-common/src/Data/Code.hs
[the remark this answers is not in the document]
fisx: Ok you are right:

```
Prelude Data.Aeson> newtype Y = Y Int deriving newtype (Show, ToJSON)
newtype Y = ...
Prelude Data.Aeson> newtype X = X Int deriving (Show, ToJSON)
newtype X = ...
Prelude Data.Aeson> (X 3, Y 3, encode (X 3, Y 3))
(X 3,3,"[3,3]")  -- (same for ToByteString)
```

I earlier tested the `Show` instance, and there is a difference, but that's why you left this in `deriving stock`.

libs/types-common/src/Data/Misc.hs
fisx: Yes, that's why I added it as a FUTUREWORK.  :)

But now I'm thinking that we're going to use these types also for `servant-client`, so maybe nevermind.

fisx: (I remember having had this discussion in the past, though, so maybe it makes sense to leave a comment that we do want to have ToJSON, and why.)

libs/wire-api/src/Wire/API/Team.hs
```diff
@@ -206,11 +213,19 @@ newTeamJson (NewTeam n i ik _) =
     # "icon_key" .= (fromRange <$> ik)
     # []
 
-deriving instance FromJSON BindingNewTeam
+deriving newtype instance FromJSON BindingNewTeam
+
+-- FUTUREWORK: since new team members do not get serialized, we zero them here.
+-- it may be worth looking into how this can be solved on in the types.
```
fisx, line 219:
```suggestion
-- it may be worth looking into how this can be solved in the types.
```

libs/wire-api/src/Wire/API/User/Auth.hs
```diff
@@ -373,23 +402,16 @@ instance FromJSON RemoveCookies where
 -- | A temporary API access token.
 data AccessToken = AccessToken
   { user :: UserId,
-    access :: LByteString, -- accessTokenValue
-    tokenType :: TokenType, -- accessTokenType
-    expiresIn :: Integer -- accessTokenExpiresIn
+    -- | assumed to be valid UTF-8
```
fisx, line 405:
```suggestion
    -- | TODO: must be valid UTF-8, encode that in the type!
```

libs/wire-api/src/Wire/API/User/Profile.hs
```diff
@@ -66,14 +67,17 @@ import Data.Range
...
 import qualified Data.Text as Text
 import Imports
+import Wire.API.Arbitrary (Arbitrary (arbitrary), GenericUniform (..))
 
 --------------------------------------------------------------------------------
 -- Name
 
 -- | Usually called display name.
 newtype Name = Name
   {fromName :: Text}
```
fisx, line 77:
```suggestion
  {fromName :: Text}  -- TODO: use @Range 1 128 Text@
```

libs/wire-api/test/unit/Test/Wire/API/Roundtrip.hs
```diff
@@ -0,0 +1,320 @@
...
+      testRoundTrip @User.NewUserPublic,
+      testRoundTrip @User.UserIdList,
+      testRoundTrip @User.UserProfile,
+      testRoundTrip @User.User,
+      testRoundTrip @User.SelfProfile,
+      testRoundTrip @User.InvitationCode,
+      testRoundTrip @User.BindingNewTeamUser,
+      -- FUTUREWORK: this should probably be tested individually,
+      -- but NewUserOrigin doesn't have JSON instances, just plain functions.
+      -- testRoundTrip @User.NewUserOrigin,
```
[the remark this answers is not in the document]
mheinzel: > Requires Deferred Type Errors to be enabled for the file it is called in.

:grimacing:

libs/wire-api/test/unit/Test/Wire/API/Roundtrip.hs
```diff
@@ -0,0 +1,320 @@
...
+      testRoundTrip @User.DeleteUser,
+      testRoundTrip @User.VerifyDeleteUser,
+      testRoundTrip @User.DeletionCodeTimeout,
+      testRoundTrip @User.Activation.ActivationKey,
+      -- FUTUREWORK: this should probably be tested individually,
+      -- but ActivationTarget currently doesn't have JSON instances itself.
+      -- testRoundTrip @User.Activation.ActivationTarget,
+      testRoundTrip @User.Activation.ActivationCode,
+      testRoundTrip @User.Activation.Activate,
+      currentlyFailing (testRoundTrip @User.Activation.ActivationResponse), -- because ToJSON doesn't serialize 'sso_id'
```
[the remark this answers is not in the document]
mheinzel: It might make sense to some degree, since SSO users don't need to activate their account (unless I'm mistaken). However, then the types should be adjusted to not allow this. I will leave a comment.

In the meantime, serializing this field (even though it should normally not occur) is probably better than silently dropping it, so I'll do that.

libs/wire-api/src/Wire/API/Event/Conversation.hs
```diff
@@ -461,13 +498,14 @@ instance ToJSON MemberUpdateData where
 
 instance FromJSON MemberUpdateData where
   parseJSON = withObject "member-update event data" $ \m ->
-    MemberUpdateData <$> m .:? "target"
-      -- NOTE: This is really not a maybe and should
+    MemberUpdateData
+      -- NOTE: This (target) is really not a maybe and should
       --       be made compulsory 28 days after the next
       --       release to prod to guaratee that no events
       --       out there do not contain id.
```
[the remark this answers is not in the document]
mheinzel: I think I'll just change the comment to say that it can be removed immediately. I'll also link the ticket: https://github.com/zinfra/backend-issues/issues/1309

libs/wire-api/src/Wire/API/User/Auth.hs
mheinzel: I applied most of these suggestions, but I might change the TODO to a FUTUREWORK.

PR #1076, opened 2020-04-24
Leave domain empty in cookies.

fisx:
This makes UAs default to server host (plus sub-domains).

Fixes https://github.com/zinfra/backend-issues/issues/953

PR #1102, opened 2020-05-18
Leave domain empty in cookies, but account for clients with old cookies.

fisx:
Fixes https://github.com/zinfra/backend-issues/issues/953

#1076 breaks UX: clients attempting to get new tokens from `/access` with cookies they got before this change will receive an error, and will have to login again.

This PR does it right.

When multiple cookies are presented to the backend, _all of them_ will be tested against the database. If any of them is valid, then we pick the "first" one and use that for auth purposes.

### DIFF
--- a/services/brig/brig.integration.yaml
+++ b/services/brig/brig.integration.yaml
@@ -135,7 +135,6 @@ optSettings:
   setNexmo: test/resources/nexmo-credentials.yaml
   # setStomp: test/resources/stomp-credentials.yaml
   setUserMaxConnections: 16
-  setCookieDomain: 127.0.0.1
   setCookieInsecure: true
   setUserCookieRenewAge: 2
   setUserCookieLimit: 5
--- a/services/brig/src/Brig/Options.hs
+++ b/services/brig/src/Brig/Options.hs
@@ -411,9 +411,6 @@ data Settings = Settings
     -- | Max. number of permanent clients per user
     setUserMaxPermClients :: !(Maybe Int),
     -- | The domain to restrict cookies to
-    setCookieDomain :: !Text,
-    -- | Whether to allow plain HTTP transmission
-    --   of cookies (for testing purposes only)
     setCookieInsecure :: !Bool,
     -- | Minimum age of a user cookie before
     --   it is renewed during token refresh
--- a/services/brig/src/Brig/Options.hs
+++ b/services/brig/src/Brig/Options.hs
@@ -410,7 +410,8 @@ data Settings = Settings
     setUserMaxConnections :: !Int64,
     -- | Max. number of permanent clients per user
     setUserMaxPermClients :: !(Maybe Int),
-    -- | The domain to restrict cookies to
+    -- | Whether to allow plain HTTP transmission
+    --   of cookies (for testing purposes only)
     setCookieInsecure :: !Bool,
     -- | Minimum age of a user cookie before
     --   it is renewed during token refresh
--- a/services/brig/src/Brig/Provider/API.hs
+++ b/services/brig/src/Brig/Provider/API.hs
@@ -1045,7 +1045,6 @@ setProviderCookie t r = do
       Cookie.def
         { Cookie.setCookieName = "zprovider",
           Cookie.setCookieValue = toByteString' t,
-          Cookie.setCookieDomain = Nothing,
           Cookie.setCookiePath = Just "/provider",
           Cookie.setCookieExpires = Just (ZAuth.tokenExpiresUTC t),
           Cookie.setCookieSecure = not (setCookieInsecure s),
--- a/services/brig/src/Brig/Provider/API.hs
+++ b/services/brig/src/Brig/Provider/API.hs
@@ -1045,7 +1045,7 @@ setProviderCookie t r = do
       Cookie.def
         { Cookie.setCookieName = "zprovider",
           Cookie.setCookieValue = toByteString' t,
-          Cookie.setCookieDomain = Just $ Text.encodeUtf8 . setCookieDomain $ s,
+          Cookie.setCookieDomain = Nothing,
           Cookie.setCookiePath = Just "/provider",
           Cookie.setCookieExpires = Just (ZAuth.tokenExpiresUTC t),
           Cookie.setCookieSecure = not (setCookieInsecure s),
--- a/services/brig/src/Brig/User/API/Auth.hs
+++ b/services/brig/src/Brig/User/API/Auth.hs
@@ -305,11 +305,17 @@ tokenRequest = opt (userToken ||| legalHoldUserToken) .&. opt (accessToken ||| l
     legalHoldUserToken = cookieErr @ZAuth.LegalHoldUser <$> cookie "zuid"
     accessToken = parse @ZAuth.Access <$> (tokenHeader .|. tokenQuery)
     legalHoldAccessToken = parse @ZAuth.LegalHoldAccess <$> (tokenHeader .|. tokenQuery)
+    --
+    tokenHeader :: r -> Result P.Error ByteString
     tokenHeader = bearer <$> header "authorization"
+    --
+    tokenQuery :: r -> Result P.Error ByteString
     tokenQuery = query "access_token"
+    --
     cookieErr :: ZAuth.UserTokenLike u => Result P.Error (ZAuth.Token u) -> Result P.Error (ZAuth.Token u)
     cookieErr x@Okay {} = x
     cookieErr (Fail x) = Fail (setMessage "Invalid user token" (P.setStatus status403 x))
+    --
     -- Extract the access token from the Authorization header.
     bearer :: Result P.Error ByteString -> Result P.Error ByteString
     bearer (Fail x) = Fail x
@@ -323,6 +329,7 @@ tokenRequest = opt (userToken ||| legalHoldUserToken) .&. opt (accessToken ||| l
                     TypeError
                     (setMessage "Invalid authorization scheme" (err status403))
                 )
+    --
     -- Parse the access token
     parse :: ZAuth.AccessTokenLike a => Result P.Error ByteString -> Result P.Error (ZAuth.Token a)
     parse (Fail x) = Fail x
--- a/services/brig/src/Brig/User/API/Auth.hs
+++ b/services/brig/src/Brig/User/API/Auth.hs
@@ -34,9 +34,9 @@ import qualified Brig.ZAuth as ZAuth
 import qualified Data.ByteString as BS
 import Data.ByteString.Conversion
 import Data.Either.Combinators (leftToMaybe, rightToMaybe)
+import Data.Id
 import Data.List1 (List1)
 import qualified Data.List1 as List1
-import Data.Id
 import Data.Predicate
 import qualified Data.Swagger.Build.Api as Doc
 import qualified Data.ZAuth.Token as ZAuth
@@ -360,8 +360,8 @@ cookies k r =
     [] -> Fail . addLabel "cookie" $ notAvailable k
     cc ->
       case mapMaybe fromByteString cc of
-        []     -> (Fail . addLabel "cookie" . typeError k $ "Failed to get zuid cookies")
-        (x:xs) -> return $ List1.list1 x xs
+        [] -> (Fail . addLabel "cookie" . typeError k $ "Failed to get zuid cookies")
+        (x : xs) -> return $ List1.list1 x xs
 
 notAvailable :: ByteString -> P.Error
 notAvailable k = e400 & setReason NotAvailable . setSource k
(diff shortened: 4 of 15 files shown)


fisx:
I've opened this with the changes from the earlier PR.  Now we just need to figure out how to do it right.

fisx:
@mythsunwind writes:

> The current issues with our QA environments and former-used accounts are happening, because when `/login` is accessed, the browsers do not override the old cookie with `".zinfra.io"` with the new cookie. Instead they then have two cookie. One cookie with a working token and `Domain:"staging-nginz-https.zinfra.io"` and one with an outdated cookie and `Domain:.zinfra.io`. Unfortunately it seems both Firefox and Chrome then send two zuid tokens which results in a 403 on backend.

So is it enough to just allow for the old cookie to be present, but ignore it?

mythsunwind:
> So is it enough to just allow for the old cookie to be present, but ignore it?

Here is an example of the `Cookie` header of a request headers done by Firefox when the staging backend had the issue:
```
zuid=O49aSGfTwRWCjOlIUzJSSI0S2OqQRx0qtj2wjo6YExDrWDKrCN02dd97Pux7PQCmm8bWYoRWTBe_rs1S0mQOBA==.v=1.k=1.d=1593608983.t=u.l=.u=a7adc996-f698-4275-b0e4-1980f7105fba.r=8ac45588; zuid=qqme0VEmQkgR1l6ZIw-o3BO-OBbYsujB5PgTWOXNAFs69ffHrt1sPIiN3ajC1DfBut7l86qi4XHvsgGaP77UDg==.v=1.k=1.d=1594647026.t=u.l=.u=26a193b9-c4de-4671-9140-71470651bf6d.r=1bd7e045
```

I hope this helps you.

fisx:
106816c does this:

```
            test/integration/API/User/Auth.hs:213:
            every login must issue a new cookie
```

I'm probably missing something obvious?

services/brig/test/integration/API/User/Auth.hs
```diff
@@ -183,6 +196,42 @@ testNginzLegalHold b g n = do
...
+-- while: because the domains differ, new ones will not overwrite old ones.  This should be
+-- handled gracefully (ie., one invalid cookie should just be ignored if up to two cookies
+-- with label @"zuid"@ are present).
+testNginzLegacyCookies :: Brig -> Nginz -> Http ()
+testNginzLegacyCookies b n = do
+  u <- randomUser b
+  let Just email = userEmail u
+      dologin :: HasCallStack => Http ResponseLBS
+      dologin = login n (defEmailLogin email) PersistentCookie <!! const 200 === statusCode
+  outdatedCookie <- decodeCookie <$> dologin
```
fisx, line 211: we need to make the outdated cookie look like a legacy cooikie!  ie., different domain.

services/spar/src/Spar/API.hs
```diff
@@ -69,7 +69,6 @@ import Spar.Scim
 import Spar.Scim.Swagger ()
 import Spar.Types
 import qualified URI.ByteString as URI
-import qualified Web.Cookie as Cky
```
fisx, line 72: also test that this behaves as expected!

services/brig/brig.integration.yaml
```diff
@@ -135,7 +135,6 @@ optSettings:
   setNexmo: test/resources/nexmo-credentials.yaml
   # setStomp: test/resources/stomp-credentials.yaml
   setUserMaxConnections: 16
-  setCookieDomain: 127.0.0.1
```
fisx, line 138: move this to integration.yaml

services/brig/test/integration/API/User/Auth.hs
tiago-loureiro: We just need to send 2 cookies

services/brig/brig.integration.yaml
tiago-loureiro: No need.

tiago-loureiro:
This is now ready for review. Could someone please run ormolu against it?

tiago-loureiro:
> This is now ready for review. Could someone please run ormolu against it?

Nevermind, all good

services/brig/src/Brig/User/Auth.hs
```diff
@@ -252,12 +252,34 @@ isPendingActivation ident = case ident of
...
+--   given, we perform the usual checks.
+--   If multiple cookies are given and several are valid, we return the first valid one.
 validateTokens ::
+  ZAuth.TokenPair u a =>
+  [ZAuth.Token u] -> -- FUTUREWORK: This should be a NonEmpty
+  Maybe (ZAuth.Token a) ->
+  ExceptT ZAuth.Failure AppIO (UserId, Cookie (ZAuth.Token u))
+validateTokens [] _ = throwE ZAuth.Invalid
+validateTokens (ut : []) at = validateToken ut at
+validateTokens uts at = do
```
arianvp, line 266: Is this function good enough?

Currently it will fail if _any_ of the cookies is malformed.  But succeed if they are all well-formed but only a few of them is  valid.

Don't we also want to support the usecase of   ` [ malformed, valid, malformed]` too?


Say we make a breaking change in the `zuid` format; then this function would handle that gracefully in the future; which sounds very useful to me.

services/brig/src/Brig/Options.hs
```diff
@@ -411,9 +411,6 @@ data Settings = Settings
     -- | Max. number of permanent clients per user
     setUserMaxPermClients :: !(Maybe Int),
     -- | The domain to restrict cookies to
-    setCookieDomain :: !Text,
-    -- | Whether to allow plain HTTP transmission
-    --   of cookies (for testing purposes only)
```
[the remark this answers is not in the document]
tiago-loureiro: 👍

services/brig/src/Brig/User/Auth.hs
```diff
@@ -252,12 +252,34 @@ isPendingActivation ident = case ident of
...
+  tokens <- forM uts $ \ut -> lift $ runExceptT (validateToken ut at)
```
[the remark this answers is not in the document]
tiago-loureiro: We discussed it f2f; the reason is that we don't want to fail right away so the logic is correct. Will try to tackle the previous point as a cleanup.

services/brig/src/Brig/User/Auth.hs
tiago-loureiro: We discussed it f2f too: I actually said back theb that "if the cookie is not parseable as Zauth, then it's dropped altogether when parsing the header in wai-routing".

So I added a test and... realised I was wrong! I fixed it now with https://github.com/wireapp/wire-server/pull/1102/commits/6841dff9e5615c0816f9681743f4b6c45072197b and realised why during our call I wanted to use `catMaybes` and `map` :P `traverse` fails too early

services/brig/src/Brig/User/Auth.hs
```diff
@@ -252,12 +254,32 @@ isPendingActivation ident = case ident of
...
+  ZAuth.TokenPair u a =>
+  List1 (ZAuth.Token u) ->
+  Maybe (ZAuth.Token a) ->
+  ExceptT ZAuth.Failure AppIO (UserId, Cookie (ZAuth.Token u))
+validateTokens uts at = do
+  tokens <- forM uts $ \ut -> lift $ runExceptT (validateToken ut at)
+  getFirstSuccessOrFirstFail tokens
+  where
+    -- FUTUREWORK: There is surely a better way to do this
+    getFirstSuccessOrFirstFail :: List1 (Either ZAuth.Failure (UserId, Cookie (ZAuth.Token u))) -> ExceptT ZAuth.Failure AppIO (UserId, Cookie (ZAuth.Token u))
```
arianvp, line 271: This function is fully pure now; and can be way more polymorphic.  This also very much signifies the future work.

The callsite above needs to just be replaced with:
```haskell
ExceptT $ getFirstSuccessOrFirstFail tokens
```

```suggestion
    getFirstSuccessOrFirstFail :: List1 (Either ZAuth.Failure (UserId, Cookie (ZAuth.Token u))) -> ExceptT ZAuth.Failure AppIO (UserId, Cookie (ZAuth.Token u))
    getFirstSuccessOrFirstFail :: List1 (Either e a) -> Either e a
```

services/brig/src/Brig/Provider/API.hs
```diff
@@ -1045,7 +1045,7 @@ setProviderCookie t r = do
       Cookie.def
         { Cookie.setCookieName = "zprovider",
           Cookie.setCookieValue = toByteString' t,
-          Cookie.setCookieDomain = Just $ Text.encodeUtf8 . setCookieDomain $ s,
+          Cookie.setCookieDomain = Nothing,
```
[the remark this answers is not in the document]
tiago-loureiro: 👍

services/brig/test/integration/API/User/Auth.hs
```diff
@@ -183,6 +196,33 @@ testNginzLegalHold b g n = do
...
   get (n . path "/notifications" . header "Authorization" ("Bearer " <> (toByteString' t))) !!! const 200 === statusCode
 
+-- | Corner case for 'testNginz': when upgrading a wire backend from the old behavior (setting
+-- cookie domain to eg. @*.wire.com@) to the new behavior (leaving cookie domain empty,
+-- effectively setting it to the backend host), clients may start sending two cookies for a
+-- while: because the domains differ, new ones will not overwrite old ones locally (it was seen
+-- on different browsers) although the cookie does get overriden in the DB). This should be handled
+-- gracefully (ie., one invalid cookie should just be ignored if up to two cookies with label @"zuid"@ are present).
+--
+-- In this test, we actually don't use different domains (testing that correctly is actually pretty
```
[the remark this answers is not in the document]
tiago-loureiro: I explained it better [here](https://github.com/wireapp/wire-server/pull/1102/commits/b05e5366ddb46fb607dfefd900c07ac7564dfa5d)

services/brig/test/integration/API/User/Auth.hs
```diff
@@ -183,6 +196,33 @@ testNginzLegalHold b g n = do
...
+      dologin = login n (defEmailLogin email) PersistentCookie <!! const 200 === statusCode
+  badCookie1 <- (\c -> c {cookie_value = "SKsjKQbiqxuEugGMWVbq02fNEA7QFdNmTiSa1Y0YMgaEP5tWl3nYHWlIrM5F8Tt7Cfn2Of738C7oeiY8xzPHAA==.v=1.k=1.d=1.t=u.l=.u=13da31b4-c6bb-4561-8fed-07e728fa6cc5.r=f844b420"}) . decodeCookie <$> dologin
+  goodCookie <- decodeCookie <$> dologin
+  badCookie2 <- (\c -> c {cookie_value = "SKsjKQbiqxuEugGMWVbq02fNEA7QFdNmTiSa1Y0YMgaEP5tWl3nYHWlIrM5F8Tt7Cfn2Of738C7oeiY8xzPHAA==.v=1.k=1.d=1.t=u.l=.u=13da31b4-c6bb-4561-8fed-07e728fa6cc5.r=f844b420"}) . decodeCookie <$> dologin
+
+  post (n . path "/access" . cookie goodCookie) !!! const 200 === statusCode
+  post (n . path "/access" . cookie badCookie1) !!! const 403 === statusCode
+  post (n . path "/access" . cookie badCookie2) !!! const 403 === statusCode
+  -- Sending both cookies should always work, regardless of the order
+  post (n . path "/access" . cookie badCookie1 . cookie goodCookie . cookie badCookie2) !!! const 200 === statusCode
```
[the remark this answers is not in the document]
tiago-loureiro: https://github.com/wireapp/wire-server/pull/1102/commits/b05e5366ddb46fb607dfefd900c07ac7564dfa5d

services/brig/src/Brig/User/Auth/Cookie.hs
```diff
@@ -235,7 +234,7 @@ setResponseCookie c r = do
       WebCookie.def
         { WebCookie.setCookieName = "zuid",
           WebCookie.setCookieValue = toByteString' (cookieValue c),
-          WebCookie.setCookieDomain = Just $ encodeUtf8 . setCookieDomain $ s,
+          WebCookie.setCookieDomain = Nothing,
```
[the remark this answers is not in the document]
tiago-loureiro: https://github.com/wireapp/wire-server/pull/1102/commits/b05e5366ddb46fb607dfefd900c07ac7564dfa5d

services/brig/src/Brig/User/API/Auth.hs
```diff
@@ -338,3 +345,28 @@ tokenRequest = opt (userToken ||| legalHoldUserToken) .&. opt (accessToken ||| l
 tokenResponse :: ZAuth.UserTokenLike u => Auth.Access u -> AppIO Response
 tokenResponse (Auth.Access t Nothing) = pure $ json t
 tokenResponse (Auth.Access t (Just c)) = Auth.setResponseCookie c (json t)
+
+-- Utilities
```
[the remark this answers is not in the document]
tiago-loureiro: Done.

tiago-loureiro:
This is good to go from my end, all PR feedback was addressed now.

mheinzel (approved):
Just one small formatting nit, not a blocker. Looks good to me! :)

jschaul (approved):
Looks good to me, too.

PR #1105, opened 2020-05-20
Change maxTeamSize to Word32 to allow for larger teams

tiago-loureiro:
Fixes: https://github.com/zinfra/backend-issues/issues/1459


tiago-loureiro:
> LGTM (but why not `Int`?)

I wanted to keep it in sync with the rest of the options; although we're using a mix of `Word` and `Int` which is perhaps not required and requires using `fromIntegral` in many places, perhaps that can be simplified and standardised

PR #1108, opened 2020-05-22
Customer Extensions

fisx:
Fixes https://github.com/zinfra/backend-issues/issues/1399

TODO:

- [x] integration test with more than one entry in the blocked-domain list
- [x] integration test that invitations can still be accepted from blocked domains
- [x] make integration tests pass


fisx:
@tiago-loureiro you said:

> I think u just need to touch `/activate/send`

@julialongtin and I this we should touch `/register` instead:

- UX is bad: user first needs to go read their email, get back with the secret, and *then* get told that they can't create a user here.
- registered, non-activated accounts also have binding ToS, which may cause problems when cleaning them up.
- registered, non-activated accounts need to be cleaned up in the first place.
- blocking `/register` effectively blocks `/activate/send` as well, as the latter requires the former to have recorded a secred in the DB to succeed.

Let me know if we missed something here.

services/brig/test/integration/API/User/Account.hs
```diff
@@ -1011,23 +1011,17 @@ testDomainsBlockedForRegistration :: Opt.Opts -> Brig -> Http ()
...
+  post (brig . path "/activate/send" . contentJson . body (p badEmail1)) !!! do
     const 451 === statusCode
     const (Just "domain-blocked-for-registration") === (^? AesonL.key "label" . AesonL._String) . (responseJsonUnsafe @Value)
-  post (brig . path "/register" . contentJson . body (p badEmail2)) !!! do
+  post (brig . path "/activate/send" . contentJson . body (p badEmail2)) !!! do
     const 451 === statusCode
     const (Just "domain-blocked-for-registration") === (^? AesonL.key "label" . AesonL._String) . (responseJsonUnsafe @Value)
   goodEmail <- randomEmail <&> \e -> e {emailDomain = "good.domain.com"}
-  post (brig . path "/register" . contentJson . body (p goodEmail)) !!! do
+  post (brig . path "/activate/senf" . contentJson . body (p goodEmail)) !!! do
```
tiago-loureiro, line 1021: Typo :P

services/brig/src/Brig/API/Error.hs
```diff
@@ -484,3 +486,13 @@ federationNotImplemented qualified =
     rendered = LT.intercalate ", " . toList . fmap (LT.fromStrict . renderMapping) $ qualified
     renderMapping IdMapping {idMappingLocal, idMappingGlobal} =
       idToText idMappingLocal <> " -> " <> renderQualifiedId idMappingGlobal
+
+-- (the tautological constraint in the type signature is added so that once we remove the
+-- feature, ghc will guide us here.)
+customerExtensionBlockedDomain :: (DomainsBlockedForRegistration ~ DomainsBlockedForRegistration) => Domain -> Wai.Error
```
arianvp, line 492: I miss agda now.

What you want here is a witness that ` (x : Domain) -> (x  <= domainsBlocked)`

But this is a fine workaround :)

fisx: thank you.  :)

PR #1110, opened 2020-05-25
api-client: depend on wire-api only

mheinzel:
Follow up to https://github.com/zinfra/backend-issues/issues/1416

### DIFF
--- a/libs/api-client/src/Network/Wire/Client/API/Push.hs
+++ b/libs/api-client/src/Network/Wire/Client/API/Push.hs
@@ -43,7 +43,7 @@ module Network.Wire.Client.API.Push
 where
 
 import Bilge
-import Brig.Types
+import Brig.Types (Name, User, UserConnection, userEmail)
 import Control.Concurrent (myThreadId)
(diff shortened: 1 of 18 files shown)


libs/api-client/src/Network/Wire/Client/API/Push.hs
```diff
@@ -174,7 +180,8 @@ data Event
   | EMemberLeave (ConvEvent UserIdList)
   | EConnect (ConvEvent Connect)
   | EConvRename (ConvEvent ConversationRename)
-  | EMemberStateUpdate (ConvEvent MemberUpdate)
+  | -- seems like it should be 'Wire.API.Event.Conversation.MemberUpdateData' instead
```
fisx, line 183:
```suggestion
  | -- FUTUREWORK: seems like `MemberUpdate` should be 'Wire.API.Event.Conversation.MemberUpdateData' instead
```

mheinzel:
Think there is only a handful, most of them related to events, which will be cleaned up with https://github.com/zinfra/backend-issues/issues/1417.

PR #1111, opened 2020-05-25
wire-api: extend generic Arbitrary instances with implementation for 'shrink'

mheinzel:
~This builds on top of #1098, so I'll have to rebase once that is merged. The diff should be quite small then.~

The instance for shrink makes it much easier to find the reason for failing test cases, for example the roundtrip tests still marked as failing (see #https://github.com/zinfra/backend-issues/issues/1446). To write the generic instance, we need QuickCheck 2.14.

Another benefit with QuickCheck 2.14 is that many generators became significantly faster.


mheinzel:
I rebased now, please take a look.

PR #1113, opened 2020-05-25
Validate saml emails

fisx:
This is probably good to review commit-by-commit.

Allow for users provisioned by scim and authenticated by saml to have a validated email address in some cases (in the user record in brig).

This will support the use case of provisioning users with scim, but only having a fake idp to authenticate them, and using passwords and email as the main authentication method.  They users being provisioned with an email address will have to (a) respond to the validation email as usual; (b) follow the "forgot password" procedure to get themselves an initial password.

Fixes https://github.com/zinfra/backend-issues/issues/1465

Steps:

- [X] Skeleton of the activation function and the call side.
- [x] Failing integration test.
- [x] Feature flag for teams (no in the config file, just in galley, stern).
- [x] RFC call to the feature flag in `Spar.Intra.Galley`.
- [x] RFC call to brig `PUT /self/email` in `Spar.intra.Brig`.
- [ ] What do we do about expired validation emails?  Change expiration time to 3 weeks globally?  Or just for teams that have this feature turned on?


fisx:
@arianvp what do you think of this?  Please read the description of this PR for context.

We currently store the SAML NameID `Scim.externalId`, whether it is an email or not.  `Scim.User` also contains a field `Scim.emails`.  Let's read that for email(s), and if there are any inside we validate them.  If they are not, no validation takes place.

This way, we don't need a team flag that maintains the old behavior for everybody but the customers who need this flow.

arianvp:
One problem with that. the `SCIM` library currently has no support for `lens`'ing into lists for updates. So fields like `emails` are currently not supported in any shape or form.

`PATCH`'ing and `PUT`'ing lists is rather intricate and is future work that I didn't have time for when I last worked on it.   If you set any of these fields `azure` will send us PATCH requests that we can not parse

arianvp:
This also comes back to the fact that I think our internal scim schema should only reflect and store the things we actually support. See https://github.com/zinfra/backend-issues/issues/1006

services/spar/src/Spar/Intra/Brig.hs
```diff
@@ -169,6 +170,10 @@ createBrigUser suid (Id buid) teamid mbName managedBy = do
       | otherwise ->
         throwSpar . SparBrigError . cs $ "create user failed with status " <> show sCode
 
+updateEmail :: (HasCallStack, MonadSparToBrig m) => UserId -> Email -> m ()
```
arianvp, line 173: We should not introduce any new hacks that call `/self` endpoints with forged authorization headers from other services. Instead add a proper internal endpoint instead. This makes our lives hard when implementing federation; and is also just hacky

services/spar/src/Spar/App.hs
```diff
@@ -208,12 +211,28 @@ autoprovisionSamlUserWithId buid suid mbName managedBy = do
   let teamid = idp ^. idpExtraInfo . wiTeam
   scimtoks <- wrapMonadClient $ Data.getScimTokens teamid
   if null scimtoks
-    then createSamlUserWithId buid suid mbName managedBy
+    then do
+      createSamlUserWithId buid suid mbName managedBy
+      validateEmailIfExists buid suid
```
arianvp, line 216: renamed to `ifNeeded` maybe?

arianvp:
The current approach looks fine to me (with the global setting per team for this) as opposed to using the `emails` field in SCIM

fisx:
quick thought extracted from the above, before I forget:  a number of users won't respond in time before the validation email expires.  how do we resend them?  (related: https://github.com/zinfra/backend-issues/issues/1087)

services/spar/src/Spar/Intra/Brig.hs
fisx: ok, fair point.  i'll add an internal end-point that does exactly the same thing, it's just a few lines of boilerplate.  thanks for reminding me!

services/spar/src/Spar/App.hs
fisx: I'd be ok with `ifAvailable`, or `ifPresent`.  It's not about the need, it's about the presence.

fisx:
> The current approach looks fine to me (with the global setting per team for this) as opposed to using the `emails` field in SCIM

so you actually prefer the current approach?  i thought that using the emails field would have the benefit that we don't need to introduce another team flag that people need to learn how to use.  easier to implement, easier to teach, easier to use.

arianvp:
I don't prefer the current approach; but I don't see us realistically pulling off the other approach in time. SCIM needs love first, and we need to schedule time for it. Multi-valued attributes simply _dont work_ currently so we can not use them

services/spar/src/Spar/Intra/Brig.hs
fisx: there is actually a `put /i/self/email`, which only needs to be augmented with a query flag to allow for sending out emails.

fisx:
this is probably good to review commit-by-commit.

services/galley/schema/src/V42_TeamFeatureValidateSamlEmails.hs
```diff
@@ -0,0 +1,29 @@
...
+  )
+where
+
+import Cassandra.Schema
+import Imports
+import Text.RawString.QQ
+
+migration :: Migration
+migration = Migration 42 "Add feature flag for validation of saml emails" $ do
+  schema' [r| ALTER TABLE team_features ADD validate_saml_emails int; |]
```
fisx, line 29: I just noticed that `validate_saml_emails_status` would be more consistent.  I have no preference, let the reviewer decide.

fisx:
There are two things in this PR that could have gone in separately, but were motivated by the work here: roundtrip-testing for bytestring-conversions in wire-api (e3537dd); and simplification of the database interface for team features (85619c1).

services/spar/test-integration/Util/Scim.hs
```diff
@@ -145,6 +145,21 @@ randomScimUserWithSubjectAndRichInfo richInfo = do
...
+randomScimUserWithEmail :: MonadRandom m => m (Scim.User.User SparTag, Email)
+randomScimUserWithEmail = do
+  suffix <- cs <$> replicateM 7 (getRandomR ('0', '9'))
+  let email = Email ("email" <> suffix) "example.com"
+      externalId = fromEmail email
+  pure
+    ( (Scim.User.empty userSchemas ("scimuser_" <> suffix) (ScimUserExtra (RichInfo mempty mempty)))
+        { Scim.User.displayName = Just ("Scim User #" <> suffix),
+          Scim.User.externalId = Just externalId,
+          Scim.User.emails = mempty,
```
arianvp, line 157: Aren't these fields already set to `mempty`? why have them here? can be removed right?

services/brig/src/Brig/API/Internal.hs
```diff
@@ -288,19 +289,17 @@ deleteUserNoVerify uid = do
...
-data ChangeEmailResponse
-  = ChangeEmailResponseIdempotent
-  | ChangeEmailResponseNeedsActivation
-
-changeSelfEmailNoSend :: UserId -> Email -> Handler ChangeEmailResponse
-changeSelfEmailNoSend u email = do
+changeSelfEmailMaybeSend :: UserId -> Bool -> Email -> Handler ChangeEmailResponse
+changeSelfEmailMaybeSend u True email = do
+  API.changeSelfEmail u email
+changeSelfEmailMaybeSend u False email = do
```
arianvp, line 307: What does the boolean mean here?

services/brig/src/Brig/API/Internal.hs
```diff
@@ -86,8 +86,9 @@ sitemap = do
   post "/i/users" (continue createUserNoVerifyH) $
     accept "application" "json"
       .&. jsonRequest @NewUser
-  put "/i/self/email" (continue changeSelfEmailNoSendH) $
+  put "/i/self/email" (continue changeSelfEmailMaybeSendH) $
```
arianvp, line 94: eventhough we use an `internal` endpoint now; having it use the `zauth` header is still a bit akward to me.. but lets leave this; not for this PR to solve

arianvp:
The tests are failing. Are you sure htis is ready @fisx ?

```
galley-types                     >     permissions for viewing feature flags:                                              FAIL

galley-types                     >       test/unit/Test/Galley/Types.hs:56:

```

services/spar/test-integration/Util/Scim.hs
fisx: 4790d35d4e5ac3bb52b329172a5db4a46e81783

services/brig/src/Brig/API/Internal.hs
fisx: 20d4a3bc784643976fa319b0074b0b7f38256ffb

fisx:
> The tests are failing. Are you sure htis is ready @fisx ?
> 
> ```
> galley-types                     >     permissions for viewing feature flags:                                              FAIL
> 
> galley-types                     >       test/unit/Test/Galley/Types.hs:56:
> ```

ea2ff07643c5ce3352ce25aa7f7c9866a3cacf84

fisx:
I also removed the refactoring that broke a lot of tests, and will add it in another PR.  It has little to do with this PR to begin with.

All tests pass locally now.

PR #1114, opened 2020-05-25
Use wire-api types in public endpoints (brig)

mheinzel:
This guarantees that the public API only depends on types defined in `wire-api` (and currently also `types-common`), so we can't break the public API by changing internal types.

The other services will follow.


services/brig/src/Brig/API/Public.hs
```diff
@@ -938,65 +948,77 @@ getRichInfoH :: UserId ::: UserId ::: JSON -> Handler Response
...
-    sendActivationEmail e u p l (Just (NewTeamCreator (BindingNewTeamUser (Team.BindingNewTeam t) _))) =
-      sendTeamActivationMail e u p l (fromRange $ t ^. Team.newTeamName)
-    sendActivationEmail e u p l _ =
-      sendActivationMail e u p l Nothing
-    sendWelcomeEmail :: Email -> CreateUserTeam -> NewTeamUser -> Maybe Locale -> AppIO ()
+    sendActivationEmail e u p l = \case
+      (Just (Public.NewTeamCreator (Public.BindingNewTeamUser (Public.BindingNewTeam t) _))) ->
+        sendTeamActivationMail e u p l (fromRange $ t ^. Public.newTeamName)
+      _ ->
+        sendActivationMail e u p l Nothing
```
fisx, line 1012: Doesn't this loop?  (If so, the old one does, too.)

Also, I want to have an hlint rule that disallows catch-alls like this.  If there was another `Just _` and then a `Nothing`, it would have been much harder to get confused here.

services/brig/src/Brig/API/Public.hs
```diff
@@ -1158,102 +1178,144 @@ getHandleInfo self handle = do
...
+    ChangeEmailResponseIdempotent -> pure (setStatus status204 empty)
+    ChangeEmailResponseNeedsActivation -> pure (setStatus status202 empty)
 
-createConnectionH :: JSON ::: UserId ::: ConnId ::: JsonRequest ConnectionRequest -> Handler Response
+data ChangeEmailResponse
+  = ChangeEmailResponseIdempotent
+  | ChangeEmailResponseNeedsActivation
+
+changeSelfEmail :: UserId -> Public.Email -> Handler ChangeEmailResponse
+changeSelfEmail u email = do
```
fisx, line 1233: ah, this is the refactoring, not the internal one.  anyway, good change.  i'll rebase onto this.

services/brig/src/Brig/API/Internal.hs
```diff
@@ -290,7 +289,21 @@ deleteUserNoVerify uid = do
   lift $ API.deleteUserNoVerify uid
 
 changeSelfEmailNoSendH :: UserId ::: JsonRequest EmailUpdate -> Handler Response
-changeSelfEmailNoSendH (u ::: req) = changeEmail u req False
+changeSelfEmailNoSendH (u ::: req) = do
```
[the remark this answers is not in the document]
mheinzel: Ah, interesting. Similar ideas. :)

services/brig/src/Brig/API/Public.hs
mheinzel: The naming is very confusing here, so I think you saw two different names as the same? There are three different functions involved:
- sendActivationMail
- sendActivation**E**mail
- send**Team**ActivationMail

So at a first glance it doesn't seem like it's looping, but maybe I can make this less confusing.

mheinzel: I tried improving the pattern matching, but I don't think it really got better:

```diff
@@ -1006,9 +1007,14 @@ createUser (Public.NewUserPublic new) = do
   pure $ CreateUserResponse cok userId (Public.SelfProfile usr)
   where
     sendActivationEmail e u p l = \case
-      (Just (Public.NewTeamCreator (Public.BindingNewTeamUser (Public.BindingNewTeam t) _))) ->
-        sendTeamActivationMail e u p l (fromRange $ t ^. Public.newTeamName)
-      _ ->
+      Just (Public.NewTeamCreator newTeamUser) -> do
+        let teamName = Public._newTeamName . coerce @_ @(Public.NewTeam ()) . Public.bnuTeam $ newTeamUser
+        sendTeamActivationMail e u p l (fromRange teamName)
+      Just _ ->
+        -- not a creater
+        sendActivationMail e u p l Nothing
+      Nothing ->
+        -- not a team user at all
         sendActivationMail e u p l Nothing
     sendWelcomeEmail :: Public.Email -> CreateUserTeam -> Public.NewTeamUser -> Maybe Public.Locale -> AppIO ()
     -- NOTE: Welcome e-mails for the team creator are not dealt by brig anymore
```

So I think I'll just merge this instead of going through another CI cycle.

mheinzel: This is another possibility:

```diff
@@ -1005,10 +1005,12 @@ createUser (Public.NewUserPublic new) = do
     UserAccount _ _ -> lift $ Auth.newCookie @ZAuth.User userId Public.PersistentCookie newUserLabel
   pure $ CreateUserResponse cok userId (Public.SelfProfile usr)
   where
-    sendActivationEmail e u p l = \case
-      (Just (Public.NewTeamCreator (Public.BindingNewTeamUser (Public.BindingNewTeam t) _))) ->
-        sendTeamActivationMail e u p l (fromRange $ t ^. Public.newTeamName)
-      _ ->
+    sendActivationEmail e u p l mTeamUser
+      | Just teamUser <- mTeamUser,
+        Public.NewTeamCreator creator <- teamUser,
+        let Public.BindingNewTeamUser (Public.BindingNewTeam team) _ = creator =
+        sendTeamActivationMail e u p l (fromRange $ team ^. Public.newTeamName)
+      | otherwise =
         sendActivationMail e u p l Nothing
     sendWelcomeEmail :: Public.Email -> CreateUserTeam -> Public.NewTeamUser -> Maybe Public.Locale -> AppIO ()
     -- NOTE: Welcome e-mails for the team creator are not dealt by brig anymore
```

Or I just add a comment like "not only a `NewTeamUser`, but `NewTeamCreator`"?

I could do either of those in #1116, not sure which. Opinions?

PR #1116, opened 2020-05-27
Use wire-api types in public endpoints (galley)

mheinzel:
This guarantees that the public API only depends on types defined in `wire-api` (and currently also `types-common`), so we can't break the public API by changing internal types.


fisx:
@mheinzel do you want to merge this based on having run the tests locally?  it would be nice to have it out of the way..

mheinzel:
Just successfully rebased and ran the integration tests locally. Will merge now!

PR #1120, opened 2020-05-28
Spar: base64-encode error details in browser

fisx:


fisx:
> This looks good to me, can we have a bit more complex examples in the test cases though?

Can you be more specific?  I agree that we should implement the pending tests, but is there anything missing there still?

tiago-loureiro:
> > This looks good to me, can we have a bit more complex examples in the test cases though?
> 
> Can you be more specific? I agree that we should implement the pending tests, but is there anything missing there still?

It was more in the sense of: I cannot see what changed/got fixed looking at the test case, i.e., how does the output look different with your fix.

fisx:
I fixed the error message first.  The test then failed because two of the three substrings that I search for in the error message are now base64-encoded.  Then I added a helper to the text that searches for those substring in the *decoded* base64-text.

So if the base64-encoding of the user-provided content in the error message somehow stops being base64-encoded, the test case cannot succeed any more.

So I think what you ask for is already there, but I'm probably confused.  If you think so, let's get on the phone.  :)

services/spar/test-integration/Test/Spar/APISpec.hs
```diff
@@ -268,12 +269,25 @@ specFinalizeLogin = do
               authnreq
               True
         sparresp <- submitAuthnResponse authnresp
+        let shouldContainInBase64 :: String -> String -> Expectation
+            shouldContainInBase64 hay needle = cs hay'' `shouldContain` needle
```
mheinzel, line 273: This is already merged, but out of curiosity: Why do we search for the needle in the decoded hay, instead of searching for the encoded needle in the normal hay?

This would avoid the whole complication with finding the section between `<pre></pre>`.

fisx: good idea.  just didn't occur to me.

but don't the character bounderies slide over the bits of the base64 encoding?  that would mean that I can't just encode the substring and search it in the encoding, because the bytes in the encoding depend on how many decoded bytes come before that substring.

mheinzel: But wouldn't you get the same issue the other way around?

fisx: I don't understand.  Can you elaborate?

This is the problem: the encoding of `"wef"` depends on what comes before it:

```
$ echo "wef" | base64 
d2VmCg==
$ echo "xwef" | base64 
eHdlZgo=
```

mheinzel: Ah, yeah, it's because we not only encode the input, but `(err, inputs, mcky)`. Cool, makes sense. Thanks!

PR #1115, opened 2020-05-26
Swagger docs.

fisx:
Related: https://github.com/zinfra/backend-issues/issues/1409

PR #1121, opened 2020-05-29
Copy swagger-ui bundle to nginz conf for integration tests

akshaymankar:


fisx:
Huh?  I thought that was already available, see https://github.com/wireapp/wire-server/pull/1115#issuecomment-634497857

akshaymankar:
> Huh? I thought that was already available, see [#1115 (comment)](https://github.com/wireapp/wire-server/pull/1115#issuecomment-634497857)

I saw that and tried it and it was not. The copying of directory was only happening in `deploy/services-demo/demo.sh`. So, those instructions only worked if you had deployed the demo on your machine previously.

PR #1122, opened 2020-05-29
Consolidate team features

akshaymankar:
This is being done as _pre-refactoring_ for https://github.com/zinfra/backend-issues/issues/1408 and another issue which needs to add a feature flag.


libs/brig-types/src/Brig/Types/Team/LegalHold.hs
```diff
@@ -24,10 +24,6 @@ module Brig.Types.Team.LegalHold
     viewLegalHoldService,
     LegalHoldClientRequest (..),
 
-    -- * LegalHoldTeamConfig (re-export)
-    LegalHoldTeamConfig (..),
-    LegalHoldStatus (..),
```
fisx, line 29: note to self: sso equivalent has been removed from galley-types.

libs/galley-types/src/Galley/Types/Teams.hs
```diff
@@ -297,15 +298,13 @@ makeLenses ''FeatureFlags
...
 data HiddenPerm
   = ChangeLegalHoldTeamSettings
-  | ViewLegalHoldTeamSettings
   | ChangeLegalHoldUserSettings
   | ViewLegalHoldUserSettings
-  | ViewSSOTeamSettings -- (change is only allowed via customer support backoffice)
-  | ViewTeamSearchVisibilityAvailable
+  | ViewTeamFeature TeamFeatureName
   | ChangeTeamSearchVisibility
   | ViewTeamSearchVisibility
```
fisx, line 305:
```suggestion
```

services/galley/src/Galley/API/Internal.hs
```diff
@@ -194,7 +193,7 @@ sitemap = do
 
   put "/i/teams/:tid/features/search-visibility" (continue Teams.setTeamSearchVisibilityAvailableInternalH) $
     capture "tid"
-      .&. jsonRequest @TeamSearchVisibilityAvailableView
+      .&. jsonRequest @Public.TeamFeatureStatus
```
fisx, line 196: the 3 put end-points for the 3 features could easily be one.  same for internal and public get.

fisx: (ah, it already happened for the public get.)

services/galley/src/Galley/API/Swagger.hs
```diff
@@ -261,29 +262,17 @@ instance ToSchema ViewLegalHoldServiceInfo where
...
-              "LegalHoldEnabled" -> "enabled"
-          }
-      tweak = fmap $ schema . description ?~ descr
-        where
-          descr =
-            "determines whether admins of a team "
-              <> "are allowed to enable LH for their users."
+      status =
+        mempty
+          & enum_ ?~ [String "enabled", String "disabled"]
```
fisx, line 275: thanks, this is at least as good as the generic-then-mutate instance that was here before, but can you keep the description?

libs/galley-types/src/Galley/Types/Teams.hs
akshaymankar: `ViewTeamSearchVisibility` and `ViewTeamSearchVisibilityAvailable` mean different things for some reason. Only `ViewTeamSearchVisibilityAvailable` corresponds to the flag itself, which is the scope of this PR.

services/galley/src/Galley/API/Swagger.hs
akshaymankar: Do you mean we should combine the descriptions for legalhold, sso and search-visibility and add it here?

libs/galley-types/src/Galley/Types/Teams.hs
```diff
@@ -340,10 +339,10 @@ hiddenPermissionsFromPermissions =
             Set.fromList [ViewSameTeamEmails]
         roleHiddenPerms RoleExternalPartner =
           Set.fromList
-            [ ViewLegalHoldTeamSettings,
+            [ ViewTeamFeature TeamFeatureLegalHold,
+              ViewTeamFeature TeamFeatureSSO,
+              ViewTeamFeature TeamFeatureSearchVisibility,
```
mheinzel, line 344: How likely is it that we will remember updating this list when we add another feature flag? How can we increase it?

libs/galley-types/src/Galley/Types/Teams.hs
mheinzel: I think it's correct that he only removed `ViewTeamSearchVisibilityAvailable`, as this is the team-level feature that can only be enabled via the backoffice (like the other feature flags).
`ViewTeamSearchVisibility` can be enabled by admins.

libs/galley-types/src/Galley/Types/Teams.hs
fisx: I tried this: 76a2ab8

Which had the nice side-effect that I refactored hidden permissions a little: c9f835c

libs/wire-api/src/Wire/API/Team/Feature.hs
```diff
@@ -21,56 +19,62 @@
...
-
-    -- * SSOTeamConfig
-    SSOTeamConfig (..),
-    SSOStatus (..),
-
-    -- * Swagger
-    modelLegalHoldTeamConfig,
-    modelSsoTeamConfig,
+  ( TeamFeatureName (..),
+    typeFeatureName,
```
mheinzel, line 23: I usually kept the Swagger models separate in the export list to avoid the noise. Do you think that makes sense or should we export them next to the types they describe?

fisx: No, your approach is fine, this is an oversight.  I make a separate PR to fix this.

fisx: https://github.com/wireapp/wire-server/pull/1126

PR #1124, opened 2020-06-03
Add documentation for recreating the ES index

akshaymankar:
Fixes https://github.com/zinfra/backend-issues/issues/1286

From discussion with Tiago, we came to a conclusion that it may be easier for some customers to take some downtime and recreate the ES index instead of migrating to a new index while maintaining uptime. This PR adds documentation about how to recreate the index.

### DIFF
--- a/docs/reference/elastic-search.md
+++ b/docs/reference/elastic-search.md
@@ -83,6 +83,9 @@ ES_PORT=<YOUR_PORT> # default is 9200
 ES_SRC_INDEX=<INDEX_NAME_ALREADY_IN_USE>
 ES_DEST_INDEX=<NEW_INDEX_NAME>
 WIRE_VERSION=<VERSION_YOU_ARE_DEPLOYING>
+SHARDS=<NUMBER_OF_SHARDS_FOR_THE_INDEX>
+REPLICAS=<NUMBER_OF_REPLICAS_FOR_THE_INDEX>
+REFRESH_INTERVAL=<REFRESH_INTERVAL_FOR_THE_INDEX>
 ```
 
 1. Create the new index (please fill out values in `<>` as required)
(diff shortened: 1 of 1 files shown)


docs/reference/elastic-search.md
```diff
@@ -106,3 +109,48 @@ WIRE_VERSION=<VERSION_YOU_ARE_DEPLOYING>
 1. Redeploy brig without `elasticsearch.additionalWriteIndex` and with `elasticsearch.index` set to the name of new index
 
 Now you can delete the old index.
+
+## Recreate an index
```
tiago-loureiro, line 113:
```suggestion
## Recreate an index (simpler but requires downtime!)
```

docs/reference/elastic-search.md
```diff
@@ -106,3 +109,48 @@ WIRE_VERSION=<VERSION_YOU_ARE_DEPLOYING>
...
+it is not possible to keep the index running while the changes are applied.
+
+To tackle this, a wire-server operator must either migrate to a new index as
+documented [above](#migrate-to-a-new-index) or allow for some downtime. One
+might want to choose downtime for simplicity. These steps are especially simple
+to do when using [wire-server-deploy](https://github.com/wireapp/wire-server-deploy/).
+
+Here are the steps:
+
+Before starting, please set these environment variables
```
tiago-loureiro, line 126: Can we separate the steps a bit more clearly? As in:

# First step

`1. Delete the old index and the migrations index`

# Second step

## If you use helm charts....

## If you do _not_ use helm charts....

WDYT?

akshaymankar: Hmm, good idea, I wiil try it out to see how it looks.

docs/reference/elastic-search.md
akshaymankar: I added the "simpler" part in documentation for migrating, as this being simpler on top level is a bit confusing.

docs/reference/elastic-search.md
akshaymankar: Looks much better, thanks!

PR #1128, opened 2020-06-03
Count team owners

jschaul:
Some crude metrics about team owners/users. All the logic is in Work.hs, the rest is essentially copied.


fisx:
can this be closed?  (i'm asking because i read "throw-away code" in the comment above.)

PR #1131, opened 2020-06-08
Enable some warnings

mheinzel:
These are [not enable by `-Wall`](https://downloads.haskell.org/ghc/8.6.5/docs/html/users_guide/using-warnings.html#ghc-flag--Wall). More specifically:
- `-Wincomplete-uni-patterns`
- `-Wincomplete-record-updates`
- `-Wpartial-fields`

In tests, I disable the warnings per file, but in production code, I tried resolving the issues instead. We should probably have another look at the specific cases and see if the first fix that came to mind is also the right one.

The only other thing I would consider adding is `-Wredundant-contraints`, but it just avoids noise, doesn't usually prevent bugs, so probably not worth the time to add right now.


libs/brig-types/src/Brig/Types/Intra.hs
```diff
@@ -119,8 +119,9 @@ instance FromJSON UserAccount where
 
 instance ToJSON UserAccount where
   toJSON (UserAccount u s) =
-    let Object o = toJSON u
-     in Object $ M.insert "status" (toJSON s) o
+    case toJSON u of
+      Object o -> Object $ M.insert "status" (toJSON s) o
+      other -> other -- shouldn't happen
```
mheinzel, line 124: probably better to fail here, since this is more likely to get caught in integration tests.

libs/wai-utilities/src/Network/Wai/Utilities/Server.hs
```diff
@@ -202,7 +202,9 @@ errorHandlers =
   [ Handler $ \(x :: Wai.Error) -> pure x,
     Handler $ \(_ :: InvalidRequest) -> pure $ Wai.Error status400 "client-error" "Invalid Request",
     Handler $ \(_ :: TimeoutThread) -> pure $ Wai.Error status408 "client-error" "Request Timeout",
-    Handler $ \(ZlibException (-3)) -> pure $ Wai.Error status400 "client-error" "Invalid request body compression",
+    Handler $ \case
+      ZlibException (-3) -> pure $ Wai.Error status400 "client-error" "Invalid request body compression"
+      ZlibException _ -> pure $ Wai.Error status500 "server-error" "Server Error",
```
mheinzel, line 207: This is pretty much the current behaviour I think, but maybe we want something else?

services/brig/src/Brig/User/Event.hs
```diff
@@ -1,4 +1,6 @@
 {-# LANGUAGE RecordWildCards #-}
+{-# OPTIONS_GHC -Wno-incomplete-record-updates #-}
```
mheinzel, line 2: This is not test code, so maybe we can get rid of this. (we should keep the `partial-fields` one for now instead of changing the `UserEvent` type definition, though)

mheinzel: I got rid of this.

libs/brig-types/src/Brig/Types/Intra.hs
mheinzel: done

libs/brig-types/src/Brig/Types/Intra.hs
```diff
@@ -119,8 +119,11 @@ instance FromJSON UserAccount where
 
 instance ToJSON UserAccount where
   toJSON (UserAccount u s) =
-    let Object o = toJSON u
-     in Object $ M.insert "status" (toJSON s) o
+    case toJSON u of
+      Object o ->
+        Object $ M.insert "status" (toJSON s) o
+      other ->
+        error $ "toJSON UserAccount: not an object: " <> show (encode other)
```
fisx, line 126: Hum, I followed the discussion and thought I agreed, but now I'm actually not sure in which sense this is better, am I missing something?  I suppose it's more explicit when reading the code?  Clearer semantics?  That might be enough to make it a good change.

Does the error message at least still contain the source code location?

mheinzel: In some cases I managed to avoid the incomplete pattern match, in other (like this one), it wasn't as straight-forward. But yeah, as you said, it's at least much easier to see that this is partial.

I'm not sure if the error message contains the source code location, but it has a greppable string.

libs/wai-utilities/src/Network/Wai/Utilities/Server.hs
fisx: Perhaps the idea of the incomplete pattern was that it should *only* throw if the pattern matches, and the handler should not be picked otherwise.  But I think `runHandlers` below does not allow for this behavior.  If the type matches, this is the last handler that's called.

So I'm in favor of leaving this as is.

libs/wire-api/src/Wire/API/User.hs
```diff
@@ -701,11 +701,10 @@ data BindingNewTeamUser = BindingNewTeamUser
...
 instance ToJSON BindingNewTeamUser where
-  toJSON (BindingNewTeamUser t c) =
-    let (Object t') = toJSON t
-     in object $
-          "currency" .= c
-            # HashMap.toList t'
+  toJSON (BindingNewTeamUser (BindingNewTeam t) c) =
+    object $
+      "currency" .= c
+        # newTeamJson t
```
fisx, line 707: huh?  why has the indentation changed here?

services/gundeck/test/unit/ThreadBudget.hs
```diff
@@ -286,9 +287,7 @@ mock (Model Nothing) (Init _) =
   InitResponse <$> genSym
 mock (Model (Just _)) Run {} = pure RunResponse
 mock (Model (Just _)) Wait {} = pure WaitResponse
-mock (Model (Just _)) Measure {} = pure MeasureResponse {..}
-  where
-    rspConcreteRunning = undefined
+mock (Model (Just _)) Measure {} = pure (MeasureResponse undefined)
```
fisx, line 290: ok, this has nothing to do with warnings any more, but it's definitely an improvement.  thanks!  :)

mheinzel: Well, it followed from the warnings. I removed the `rspConcreteRunning` field label, because it's a partial accessor (`-Wpartial-fields` complained). Without it, I was not able to use the `{..}` syntax anymore.

libs/wire-api/src/Wire/API/User.hs
mheinzel: Not just the indentation, but the code changed. :stuck_out_tongue: 

I got rid of the `let`.

PR #1132, opened 2020-06-09
Add team level flag for digital signtaures

akshaymankar:
Fixes https://github.com/zinfra/backend-issues/issues/1408


mheinzel:
Are you also planning to add a few integration tests? I know, the functionality follows the same pattern as for the other feature flags, but a small test could still be useful, especially for the default value.

services/galley/test/integration/API/Teams.hs
```diff
@@ -2077,6 +2083,16 @@ testFeatureFlags = do
     getTeamSearchVisibility tid3 Public.TeamFeatureEnabled
     getTeamSearchVisibilityInternal tid3 Public.TeamFeatureEnabled
 
+  forM_ [Public.TeamFeatureDigitalSignatures, Public.TeamFeatureValidateSAMLEmails] $ \(feature) -> do
```
mheinzel, line 2086: I'm always a bit torn about using this kind of abstraction/parametrization in tests, as it can lead to just re-building the structure of the implementation (sometimes making the tests harder to grasp) and make the error messages less useful (e.g. location).

Not a problem here, though.

akshaymankar: Yeah, unfortunately the line number or the error message won't convey much here apart from one of the feature flag is not behaving as expected.
I was thinking of breaking this whole test into multiple tests and that way it could have conveyed which one and with line number from the expectation failure we would be able to figure out what exactly is broken. But I got lazy, I guess I am not feeling very lazy now, so I will do it in the next PR and while I am at it I will also move these tests to another file, navigating this one was very painful.

fisx: i think it would have helped to break up the one `forM_` loop into two lines with two consecutive calls to the function that's in the `forM_` body.

also, too many parens.

but don't make any more changes now that it's merged, I'm just trying to learn for next time.

PR #1133, opened 2020-06-09
Add federation domain and federator endpoint to service configuration

mheinzel:
Preparation for https://github.com/zinfra/backend-issues/issues/1213


services/brig/src/Brig/API/Util.hs
```diff
@@ -36,13 +37,21 @@ lookupProfilesMaybeFilterSameTeamOnly self us = do
...
 -- FUTUREWORK(federation, #1178): implement function to resolve IDs in batch
 
 -- | this exists as a shim to find and mark places where we need to handle 'OpaqueUserId's.
 resolveOpaqueUserId :: MonadReader Env m => OpaqueUserId -> m (MappedOrLocalId Id.U)
 resolveOpaqueUserId (Id opaque) = do
-  mEnabled <- view (settings . enableFederation)
-  case fromMaybe defEnableFederation mEnabled of
+  isFederationEnabled >>= \case
     False ->
       -- don't check the ID mapping, just assume it's local
```
fisx, line 56: How will this work if we switch an unfederated backend with users to a federated one?  I suppose there cannot be any remote users.

What about switching a federated backend into a non-federated one?  Then we had remote users, but now they are assumed to be local, right?  Should we just clearly state that turning the switch this way is not allowed?  Or is there a reason why it's not a problem?

services/galley/galley.integration.yaml
```diff
@@ -30,6 +34,7 @@ settings:
   concurrentDeletionEvents: 1024
   deleteConvThrottleMillis: 0
   enableIndexedBillingTeamMembers: true
+  # enableFederationWithDomain: wire.com
```
fisx, line 37: so this is redundant in both brig and galley?  *sigh* brig and galley need to be merged.

```suggestion
  # enableFederationWithDomain: wire.com
  # NOTE: enableFederationWithDomain needs to have the same value as the corrresponding setting in brig.
```

(and perhaps a similar comment in brig.)

services/brig/src/Brig/API/Util.hs
mheinzel: Enabling federation is completely fine, turning it off is more complicated. Probably we should make it clear that this is not recommended/supported for now.

At the moment the effect is that the remote users and conversations would just seem to have disappeared. Handling this more elegantly is probably a lot of work for both backend and clients.

services/galley/galley.integration.yaml
mheinzel: added (https://github.com/wireapp/wire-server/pull/1133/commits/6f9d03f052da82305c0e3f06c1ee40ad39c82485)

PR #1134, opened 2020-06-09
Federation: Implement ID mapping (galley)

mheinzel:
Related to https://github.com/zinfra/backend-issues/issues/1178

### DIFF
--- a/services/galley/src/Galley/API/IdMapping.hs
+++ b/services/galley/src/Galley/API/IdMapping.hs
@@ -32,9 +32,9 @@ import Control.Monad.Catch (throwM)
 import qualified Data.ByteString as BS
 import Data.Domain (domainText)
 import qualified Data.Id as Id
-import Data.Id (Id (Id, toUUID), OpaqueConvId, OpaqueUserId)
-import Data.IdMapping (IdMapping (IdMapping), MappedOrLocalId (Local, Mapped))
-import Data.Qualified (Qualified (Qualified, _qDomain, _qLocalPart))
+import Data.Id (Id (Id, toUUID), OpaqueConvId, OpaqueUserId, idToText)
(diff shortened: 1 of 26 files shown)


mheinzel:
I rebased and addressed all the things we talked about. @arianvp do you want to take another look?

I suggest we do a separate PR for the corresponding changes in Brig. The only thing that I might still squeeze in is batch resolving of opaque IDs, but it could also go into a follow-up PR.

mheinzel:
Rebased again, I think it's done from my side now.

services/galley/src/Galley/API/IdMapping.hs
```diff
@@ -0,0 +1,172 @@
...
+
+-- | Used by Brig and Galley to notify each other when they find a new 'IdMapping'.
+postIdMappingH ::
+  JsonRequest PostIdMappingRequest ::: JSON ->
+  Galley Response
+postIdMappingH (req ::: _) =
+  ifFederationIsEnabled $
+    fmap json . postIdMapping =<< fromJsonBody req
+
+-- | Blindly writes to our own the database, unconditionally.
```
julialongtin, line 79: to our own the... ? english error.

services/galley/src/Galley/API/IdMapping.hs
```diff
@@ -0,0 +1,172 @@
...
+-- The mapping is deterministic, so we don't bother reading existing values.
+-- If we overwrite an existing entry, then with the same value as it had before.
```
julialongtin, line 81:
```suggestion
-- If we overwrite an existing entry, then it will be overwritten with the same value as it had before.
```

mheinzel:
Okay, we're having a few issues with CI, but I'm still gonna merge now. The ormolu job didn't get triggered, but it passed on the previous commit and also passes locally.

PR #1140, opened 2020-06-23
Cleanup hspec-wai dependency

fisx:
we could either go with the first commit, or entirely get rid of the for, taking our changes into (a bit arbitrarily) wai-utilities.

third option (not implemented here): set up a tiny package hspec-wai-extra that contains the changes.

oops, and if we go for the second option, we probably need to change some places in which we use the code that has moved.

@arianvp you triggered this, please make a call how to proceed.  my personal favorite is keep fe934bd and revert 26e4f33

### DIFF
--- a/libs/wai-utilities/package.yaml
+++ b/libs/wai-utilities/package.yaml
@@ -36,6 +36,8 @@ dependencies:
 - wai >=3.0
 - wai-predicates >=0.8
 - wai-routing >=0.12
+- wai-extra
+- hspec-wai
 - warp >=3.0
 library:
   source-dirs: src
--- a/libs/wai-utilities/src/Test/Hspec/Wai/Extra.hs
+++ b/libs/wai-utilities/src/Test/Hspec/Wai/Extra.hs
@@ -1,0 +1,49 @@
+-- | <https://github.com/hspec/hspec-wai/pull/49>, <https://github.com/hspec/hspec-wai/pull/41>
+module Test.Hspec.Wai.Extra (
+  bodyContains
+, bodySatisfies
+) where
+
+import           Prelude
+import           Control.Monad
+import           Data.Char
+import qualified Data.Text as TS
+import qualified Data.Text.Encoding as TS
+import           Data.ByteString (ByteString)
+import qualified Data.ByteString as SB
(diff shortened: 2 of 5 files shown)


arianvp:
Heavy preference on  26e4f33
Maintaining a fork for a 3 line change sounds wrong to me and makes things harder to grok for others.

fisx:
```
spar                             > <no location info>: error:
spar                             >     <command line>: can't load .so/.DLL for: /root/.stack/snapshots/x86_64-linux/1de268473d2e94a7073c23577ab35b42b1d1998d267dd6b6a08887a82e40e1b6/8.6.5/lib/x86_64-linux-ghc-8.6.5/libHShscim-0.3.4-8S2wW7i2FGY4aoBFojHElw-ghc8.6.5.so (Error relocating /root/.stack/snapshots/x86_64-linux/1de268473d2e94a7073c23577ab35b42b1d1998d267dd6b6a08887a82e40e1b6/8.6.5/lib/x86_64-linux-ghc-8.6.5/libHShscim-0.3.4-8S2wW7i2FGY4aoBFojHElw-ghc8.6.5.so: hspeczmwaizm0zi9zi2zmKUVzzuzzSDlKr3lG1odv1wrO_TestziHspecziWaiziMatcher_zdwbodySatisfies_info: symbol not found)
```

the only way i could find to fix this is to `rm -rf ~/.stack/`  :(

mheinzel:
> the only way i could find to fix this is to `rm -rf ~/.stack/` :(

Did you also run into it locally? So is it reproducible or some kind of flakiness in Stack?

fisx:
> > the only way i could find to fix this is to `rm -rf ~/.stack/` :(
> 
> Did you also run into it locally? So is it reproducible or some kind of flakiness in Stack?

reproducible.

I've tried `stack clean --full`, `stack exec -- ghc-pkg unregister saml2-web-sso hscim hspec-wai`, and removing some files from deep inside `~/.stack/`, but only `rm -rf ~/.stack` fixed things.

fisx:
I think I'll pull saml2-web-sso into this repo and see if that stops triggering the stack bug.

...  and then I will join @arianvp in lobbying to get rid of stack.

PR #1142, opened 2020-06-23
Fix ormolu script.

fisx:
actually fail if ormolu version check fails.


fisx:
if you approve this, please merge right away.

PR #1143, opened 2020-06-23
Add missing license headers

mheinzel:


mheinzel:
Rebased and pushed again because CI failed (seems like it was just flaky).

PR #1146, opened 2020-06-24
Fix setting team feature status in Stern/backoffice

mheinzel:
See https://github.com/zinfra/backend-issues/issues/1529

### DIFF
--- a/libs/wire-api/src/Wire/API/Team/Feature.hs
+++ b/libs/wire-api/src/Wire/API/Team/Feature.hs
@@ -115,3 +115,15 @@ instance FromJSON TeamFeatureStatusValue where
     "enabled" -> pure TeamFeatureEnabled
     "disabled" -> pure TeamFeatureDisabled
     x -> fail $ "unexpected status type: " <> T.unpack x
+
+instance ToByteString TeamFeatureStatusValue where
+  builder TeamFeatureEnabled = "enabled"
+  builder TeamFeatureDisabled = "disabled"
+
+instance FromByteString TeamFeatureStatusValue where
+  parser = Parser.takeByteString >>= \b ->
+    case T.decodeUtf8' b of
+      Right "enabled" -> pure TeamFeatureEnabled
+      Right "disabled" -> pure TeamFeatureDisabled
+      Right t -> fail $ "Invalid TeamFeatureName: " <> T.unpack t
+      Left e -> fail $ "Invalid TeamFeatureName: " <> show e
--- a/libs/wire-api/src/Wire/API/Team/Feature.hs
+++ b/libs/wire-api/src/Wire/API/Team/Feature.hs
@@ -1,4 +1,5 @@
 {-# LANGUAGE DerivingVia #-}
+{-# LANGUAGE GeneralizedNewtypeDeriving #-}
 {-# LANGUAGE StrictData #-}
 
 -- This file is part of the Wire Server implementation.
@@ -21,11 +22,12 @@
 module Wire.API.Team.Feature
   ( TeamFeatureName (..),
     TeamFeatureStatus (..),
+    TeamFeatureStatusValue (..),
 
     -- * Swagger
-    modelTeamFeatureStatus,
     typeTeamFeatureName,
-    typeTeamFeatureStatus,
+    modelTeamFeatureStatus,
+    typeTeamFeatureStatusValue,
   )
 where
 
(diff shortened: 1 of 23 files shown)


mheinzel:
It worked as expected when I curled it, but I'm still trying to run the backoffice Swagger app.

I'll keep at it, might just fix https://github.com/zinfra/backend-issues/issues/1496 and https://github.com/wireapp/wire-server/tree/develop/tools/stern#how-to-run-stern-together-with-the-rest-of-wire-server as well. :shrug:

mheinzel:
Umhh, so I managed to get this working locally with some trickery, but there is a bit of a usability issue:

Screenshot from 2020-06-24 16-51-53

Normally, we want JSON requests and responses to be objects, but for the backoffice this doesn't really make sense, as it's much more cumbersome now to send requests.

So I guess we should accept a plain JSON string instead (only in Stern)? To do that, we will need to split out a separate type again...

akshaymankar:
Yeah, it will be nicer to accept a plain string in the body, that way we get the dropdown. But this is also fine as few other endpoints in the same swagger-ui require much more complex JSONs and so the users clearly understand this. 
So, what I want to say is that it is upto you if you want to make it nicer or avoid spending time on this.

fisx:
i'm in favor of making this more usable, if it's not more than 3 hours (i don't think it should be.)

fisx:
all we need to do is move the type which was there before from the body to the query, right?

mheinzel:
> all we need to do is move the type which was there before from the body to the query, right?

That's actually good idea. Unfortunately, I already started doing this by splitting the type into one with is just a JSON string and one with the object wrapper around it. Almost done with that...

Might be slightly nicer to have that separation, because then also the types defined for the interface correspond more closely to their JSON structure (which will also be helpful if we want to start automatically deriving JSON stuff). I'll push the update soon,  so you can have a look and shout if you think the query parameter is the better solution.

mheinzel:
@fisx @akshaymankar This should work now.

mheinzel:
> Have you run this locally? If you have, let's document how to do that.

I'll do that as part of https://github.com/wireapp/wire-server/pull/1148.

Where do you think would be the best place to document it? I was first thinking about [the stern README](https://github.com/wireapp/wire-server/tree/develop/tools/stern#how-to-run-stern-together-with-the-rest-of-wire-server), but now I'm not sure anymore.

akshaymankar:
> Where do you think would be the best place to document it? I was first thinking about [the stern README](https://github.com/wireapp/wire-server/tree/develop/tools/stern#how-to-run-stern-together-with-the-rest-of-wire-server), but now I'm not sure anymore.

It would be nice if it started when I run `services/start-services-only.sh` and then we probably don't even need to document it like rest of the services.

mheinzel:
Okay, I'll look into it as part of the other PR. This one should be ready to be merged, though. Do you think the approach with splitting into `TeamFeatureStatus` and `TeamFeatureStatusValue` makes sense?

libs/wire-api/src/Wire/API/Team/Feature.hs
```diff
@@ -69,35 +71,47 @@ instance ToByteString TeamFeatureName where
 typeTeamFeatureName :: Doc.DataType
 typeTeamFeatureName = Doc.string . Doc.enum $ cs . toByteString' <$> [(minBound :: TeamFeatureName) ..]
 
-data TeamFeatureStatus = TeamFeatureEnabled | TeamFeatureDisabled
-  deriving stock (Eq, Show, Generic)
-  deriving (Arbitrary) via (GenericUniform TeamFeatureStatus)
+newtype TeamFeatureStatus = TeamFeatureStatus
+  {teamFeatureStatusValue :: TeamFeatureStatusValue}
```
fisx, line 75: *sigh*

This is rolling back the change we did in tag `image/2.83.22` (wireapp/wire-server#1122).  The reason we changed it was that it made the code simpler all over the backend, as you can tell from all the awkwardness you had to add.  :-P

So, good thing you asked for a second review!

tools/stern/src/Stern/API.hs
```diff
@@ -337,22 +337,21 @@ routes = do
       description "Team ID"
     Doc.parameter Doc.Path "feature" Public.typeTeamFeatureName $
       description "Feature name"
-    Doc.returns Public.typeTeamFeatureStatus
+    Doc.returns (Doc.ref Public.modelTeamFeatureStatus)
```
fisx, line 340: I'd also go for the value here, as you intended, but this way we keep strings consistently out of http bodies.

libs/wire-api/src/Wire/API/Team/Feature.hs
fisx: we discussed this, and came to the conclusion that this code is more beaurocratic, but less surprising.  so we'll keep it.

mheinzel:
I rebase this and tested the backoffice. Ran into an issue with the query param deserialization, which I fixed now.

I will merge once CI passed.

PR #1148, opened 2020-06-25
Run backoffice locally

mheinzel:
See https://github.com/zinfra/backend-issues/issues/1496.

This took me longer than I thought, but it's mostly working now.

- [x] get backoffice endpoints shown in a separate tab
- [x] when sending request from the swagger-ui webapp, they reach stern
- [x] copy changes to the respective `-docker` files where they exist
- [x] document how to run stern locally in the [stern README](https://github.com/wireapp/wire-server/tree/develop/tools/stern#how-to-run-stern-together-with-the-rest-of-wire-server)

Also, there's still a weird bug I haven't figured out yet:

When I load swagger-ui, everythings works. If I reload the page, it still works. When I switch to the `Back Office` tab and expand the endpoints and *then* reload the page, it doesn't work anymore. More specifically, instead of loading `/backoffice/api-docs/`, the JavaScript tries to load  `/backoffice`. After a hard refresh it works again.

The same happens with `/sso` in the `Swagger-2.0` tab, too, so it doesn't seem like a new issue. Might not fix it for now.

### DIFF
--- a/deploy/services-demo/conf/nginz/nginx-docker.conf
+++ b/deploy/services-demo/conf/nginz/nginx-docker.conf
@@ -365,10 +365,10 @@ http {
       proxy_pass http://stern;
     }
 
-    # The trailing slash matters, as it makes sure the `/stern` prefix is removed.
-    location /stern/ {
+    location /stern {
       include common_response_no_zauth.conf;
-      proxy_pass http://stern;
+      # The trailing slash matters, as it makes sure the `/stern` prefix is removed.
+      proxy_pass http://stern/;
     }
 
     #
--- a/deploy/services-demo/conf/nginz/nginx-docker.conf
+++ b/deploy/services-demo/conf/nginz/nginx-docker.conf
@@ -355,10 +355,44 @@ http {
       proxy_pass http://spar;
     }
 
+    # Stern Endpoints
+
+    # We add a `/stern` suffix to the URL to resolve clashes with non-Stern endpoints.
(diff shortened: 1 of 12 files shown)


mheinzel:
As discussed in #1146, I will look into also running stern automatically in `services/start-services-only.sh` (although it's not one of the `services` per se, living in `tools/stern`).

deploy/services-demo/conf/nginz/nginx-docker.conf
```diff
@@ -355,6 +355,22 @@ http {
...
+
+    # We add a `/stern` suffix to the URL to resolve clashes with non-Stern endpoints.
+    rewrite ^/backoffice/api-docs/stern  /stern/api-docs?base_url=http://127.0.0.1:8080/stern/ break;
+
+    location /stern/api-docs {
+      include common_response_no_zauth.conf;
+      proxy_pass http://stern;
+    }
+
+    # The trailing slash matters, as it makes sure the `/stern` prefix is removed.
```
fisx, line 368: you changed this in `deploy/services-demo/conf/nginz/nginx.conf`, are you sure it should stay like this here?

mheinzel:
I now added support for running stern as part of the `services-demo`. In the process I had to resort to a few hacks, mostly to support

1. only optionally running stern (which I think is justifiable since exposing the backoffice API by default is risky in case anyone runs this exposed to the internet)
2. working both with the native and Docker mode of the script

Dropping any of those two would make things easier, so tell me if you would strongly prefer that.

Also, tell me if there is a nicer way of making nginx still start up when the stern Docker container is not running. I didn't find a nicer way, but there might be some tricks I don't know.

PR #1149, opened 2020-06-25
Fix services-demo (Docker)

mheinzel:
The brig and galley containers didn't start due to missing configuration values, which I fixed now. I also aligned some other `-docker` files to the non-Docker version.

### DIFF
--- a/deploy/services-demo/conf/brig.demo-docker.yaml
+++ b/deploy/services-demo/conf/brig.demo-docker.yaml
@@ -84,6 +84,9 @@ zauth:
     sessionTokenTimeout: 604800  # 7 days
     accessTokenTimeout: 900      # 15 minutes
     providerTokenTimeout: 604800 # 7 days
+    legalHoldUserTokenTimeout: 4838400    # 56 days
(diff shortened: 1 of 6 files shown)


fisx:
I haven't checked, but I suppose the two sets of files are different and one can't be just sym-linked to the other?

mheinzel:
Yes, the main difference is that in the service config files, their endpoints are configured differently, e.g.

```
# running services natively, e.g. galley.demo.conf
brig:
  host: brig
  port: 8082

# running services in Docker, e.g. galley.demo-docker.conf
brig:
  host: 127.0.0.1
  port: 8082
```

Similarly for config like `activationUrl` etc.

Not sure if this can be avoided easily (I'd like to avoid extensive templating).

mheinzel:
> Another idea could be that we get rid of docker based demo in favour of K8s/kind based demo env, our helm charts keep getting tested in CI anyways.

Kind would definitely be more heavy-weight to run, but there is still the option to run everything natively, so that might be fine. We don't really have much insight into how much the Docker services-demo is used, do we? Maybe from issues reported...

Anyways, thanks for the review!

PR #1157, opened 2020-07-02
Make cargohold use amazonka instead of aws

mheinzel:
Fixes https://github.com/wireapp/wire-server/issues/876.
Also see https://github.com/zinfra/backend-issues/issues/1481.

This revives https://github.com/wireapp/wire-server/pull/483 with a more recent version of amazonka. Everything seems to work without performance regressions ~except for problems against Minio with the [pre-signed download URLs](https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/private-content-signed-urls.html) we generate ([more docs](https://docs.aws.amazon.com/AmazonS3/latest/API/sigv4-query-string-auth.html)). This still needs to be investigated.~

*edit: I found the issue in amazonka and made a fix*

Before merging:
- [x] make PR against amazonka (https://github.com/brendanhay/amazonka/pull/588)
- [x] properly support `s3DownloadEndpoint`, which is currently ignored
- [x] find out whether we actually want to switch to `http-client-tls` here (and if not, revert that part)


services/cargohold/src/CargoHold/S3.hs
```diff
@@ -706,147 +721,46 @@ getAmzMetaChunkSize :: [(Text, Text)] -> Maybe V3.ChunkSize
...
-x3 = limitRetries 3 <> exponentialBackoff 200000
-
-handlers :: Monad m => [RetryStatus -> Handler m Bool]
-handlers =
-  Retry.httpHandlers
-    ++ [ const $ Handler $ \(S3Error s _ _ _ _ _ _ _ _ _) ->
-           return $ statusIsServerError s
-       ]
+parseMIMEType :: Text -> Maybe MIME.Type
+parseMIMEType = MIME.parseMIMEType
```
tiago-loureiro, line 733: Is it worth having this as it's just an alias?

services/cargohold/src/CargoHold/S3.hs
```diff
@@ -706,147 +721,46 @@ getAmzMetaChunkSize :: [(Text, Text)] -> Maybe V3.ChunkSize
...
 
-parseMIMEType :: ByteString -> Maybe MIME.Type
-parseMIMEType = MIME.parseMIMEType . decodeLatin1
-
-encodeMIMEType :: MIME.Type -> ByteString
-encodeMIMEType = Text.encodeUtf8 . MIME.showType
+encodeMIMEType :: MIME.Type -> Text
+encodeMIMEType = MIME.showType
```
tiago-loureiro, line 736: Same as the above?

services/cargohold/src/CargoHold/AWS.hs
```diff
@@ -0,0 +1,187 @@
...
+  runResourceT . AWST.runAWST e
+    $ AWST.trying AWS._Error
+    $ AWST.send cmd
+
+exec ::
+  (AWSRequest a, AWS.HasEnv r, MonadUnliftIO m, MonadCatch m, MonadThrow m) =>
+  r ->
+  a ->
+  m (Rs a)
+exec e cmd = execCatch e cmd >>= either (throwM . GeneralError) return
```
tiago-loureiro, line 177: These are not used and have functions with the same name on the S3 module; perhaps move those functions here since they are generic and can be used in other modules (although they are not)

services/cargohold/package.yaml
```diff
@@ -30,17 +30,23 @@ dependencies:
...
+  - amazonka-core >=1.3.7
+  - amazonka-s3 >=1.3.7
+  - amazonka-cloudfront >=1.3.7
   - attoparsec >=0.12
   - auto-update >=0.1.4
-  - aws >=0.18
   - byteable >=0.1
   - base16-bytestring >=0.1
+  - blaze-builder >=0.3
+  - binary >=0.8
```
fisx, line 49: Nit-pick: I think the only thing used from `binary` is `toLazyByteString`, which I think is just a re-export from `bytestring`.

services/cargohold/src/Main.hs
```diff
@@ -22,13 +22,11 @@ where
...
 import CargoHold.Run (run)
 import Imports
-import OpenSSL (withOpenSSL)
 import Util.Options
 
 main :: IO ()
-main = withOpenSSL $ do
-  options <- getOptions desc Nothing defaultPath
-  run options
+main = do
```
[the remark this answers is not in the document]
fisx: @franziskuskiefer do you have an opinion on openssl vs. http://hackage.haskell.org/package/tls?

We had this a while back, and after struggling for a while between "In Haskell with a few good habits, many mistakes cannot be expressed in the language" and "openssl has had so many bugs closed, how many more can there be?", we stuck with openssl.

tiago-loureiro:
> But does the V3 in the tests stand for the signature version, and do we need to cover V4 as well? I am missing a lot of context here. :)

V3 is the Cargohold API version, nothing to do with signatures

services/cargohold/src/Main.hs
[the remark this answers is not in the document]
mheinzel: Ah, I undid the changes removing OpenSSL, but must have missed this one. Interestingly, the code using OpenSSL still worked without the initialization here. Good catch!

For this PR I think it makes sense to stick with OpenSSL for consistency with the rest of the codebase, but yeah, we should think about which library we want to use in general.

services/cargohold/src/CargoHold/S3.hs
mheinzel: No, will inline it. Same for `encodeMIMEType`.

services/cargohold/src/Main.hs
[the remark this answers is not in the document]
mheinzel: https://github.com/wireapp/wire-server/pull/1157/commits/10bd3086fa5af2f6488ba91710a760ce8416ff43

services/cargohold/src/CargoHold/S3.hs
```diff
@@ -166,29 +173,29 @@ updateMetadataV3 (s3Key . mkKey -> key) (S3AssetMeta prc tok ct) = do
...
-  let cfg = cfg' {Aws.timeInfo = Aws.ExpiresIn (fromIntegral ttl)}
-  uri <-
-    liftIO $ Aws.awsUri cfg (s3UriOnly e) $
-      getObject b (Text.decodeLatin1 $ toByteString' path)
-  return =<< toUri uri
+  let req = getObject (BucketName b) (ObjectKey . Text.decodeLatin1 $ toByteString' path)
+  signed <-
+    AWS.execute (AWS.useDownloadEndpoint e) $
+      presignURL now (Seconds (fromIntegral ttl)) req
+  return =<< toUri signed
```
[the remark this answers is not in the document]
mheinzel: as part of https://github.com/wireapp/wire-server/pull/1157/commits/877c9a96036ec475d1506ad738621b2ffe417cc6

services/cargohold/src/CargoHold/S3.hs
mheinzel: https://github.com/wireapp/wire-server/pull/1157/commits/877c9a96036ec475d1506ad738621b2ffe417cc6

services/cargohold/src/CargoHold/S3.hs
mheinzel: see above

services/cargohold/src/CargoHold/AWS.hs
mheinzel: It was not quite as easy as copying them, because of the module dependencies (`Cargohold.AWS` doesn't know about `App` and `Env`), but here it is: https://github.com/wireapp/wire-server/pull/1157/commits/959adbcd2455ad895e392d1a656fa10b32dd9336

services/cargohold/package.yaml
mheinzel: Good spot. We also do this in a few other places, so I cleaned them up, too.
https://github.com/wireapp/wire-server/pull/1157/commits/0bfad0234d36b3830174382a945bbc6dfeeea9ef

franziskuskiefer:
> @franziskuskiefer do you have an opinion on openssl vs. http://hackage.haskell.org/package/tls?
> We had this a while back, and after struggling for a while between "In Haskell with a few good habits, many mistakes cannot be expressed in the language" and "openssl has had so many bugs closed, how many more can there be?", we stuck with openssl.

I generally prefer not using openssl for multiple reasons. But that haskell package doesn't appear to be used a lot, which is crucial for a TLS lib (most errors are found while using it).
I'm happy to look into what we could generally use on the BE for TLS instead of openssl. But we should probably handle that separately.

mheinzel:
I addressed all comments and now rebased to resolve conflicts. Looking forward to more feedback!

PR #1161, opened 2020-07-08
Restrict user creation

fisx:
Fixes https://github.com/zinfra/backend-issues/issues/1491

See 209a08f for context.

### DIFF
--- a/docs/reference/user/registration.md
+++ b/docs/reference/user/registration.md
@@ -1,6 +1,6 @@
 # Registration {#RefRegistration}
 
-_Author: Artyom Kazak_
+_Authors: Artyom Kazak, Matthias Fischmann_
 
 ---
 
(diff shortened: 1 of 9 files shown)


fisx:
TODO:

- [ ] add branches to all failing test cases that test for the expected behavior in the `restricted: true` case.
- [ ] add comment to `testRestrictedUserCreation` on how to extend coverage permanently.
- [ ] revert change in `brig.integration.yaml`

fisx:
> double-check that creating guest accounts is still possible.

This refers to ephemeral users invited via a "join this conversation"-link.

This is allowed both with and without restricted user creation enabled: [these two tests](https://github.com/wireapp/wire-server/blob/26f339b10230735dca4fc53f2c35ccfb7bfb18a2/services/brig/test/integration/API/User/Account.hs#L81-L82) have not been changed, and work before and after 2c19f78

fisx:
About [this test](https://github.com/wireapp/wire-server/blob/26f339b10230735dca4fc53f2c35ccfb7bfb18a2/services/brig/test/integration/API/User/Account.hs#L163-L176): doesn't seem to do anything useful.  in particular, it doesn't call `/register` [as promised](https://github.com/wireapp/wire-server/blob/26f339b10230735dca4fc53f2c35ccfb7bfb18a2/services/brig/test/integration/API/User/Account.hs#L79).  Couldn't find anything in the git history either, this function has looked like this since 2017.

Should we just remove it?

fisx:
What is the purpose of [this test](https://github.com/wireapp/wire-server/blob/26f339b10230735dca4fc53f2c35ccfb7bfb18a2/services/brig/test/integration/API/User/Account.hs#L80)?  It just creates a user via the internal end-point and updates it with an email address.  This could use a comment.

fisx:
Am i correct in assuming that *anonymous*, *guest*, and *ephemeral* users are all the same thing?

fisx:
Ready for review!  I have many questions, but none of them are related to this PR.  :)

services/brig/test/integration/API/User/Account.hs
```diff
@@ -1047,6 +1067,30 @@ testDomainsBlockedForRegistration opts brig = withDomainsBlockedForRegistration
   where
     p email = RequestBodyLBS . encode $ SendActivationCode (Left email) Nothing False
 
+-- | FUTUREWORK: This is not very exhaustive.  You can change the flag in
+-- brig.integration.yaml manually and re-run the integration tests for more coverage.  If you
+-- want to have that coverage permanently, grep this module for @setRestrictUserCreation@ and
+-- change all those tests to run with both settings.  Use 'testRestrictedUserCreation' here as
+-- an instructive example.
```
akshaymankar, line 1074: I would like to have these tests be more exhaustive. Are we not doing this now because we don't have time?

services/brig/test/integration/API/User/Account.hs
```diff
@@ -1047,6 +1067,30 @@ testDomainsBlockedForRegistration opts brig = withDomainsBlockedForRegistration
...
+    getActivationCode brig (Left e) >>= \case
+      Nothing -> liftIO $ assertFailure "missing activation key/code"
+      Just (_, c) -> do
+        let Object reg =
+              object
+                [ "name" .= Name "Alice",
+                  "email" .= fromEmail e,
+                  "email_code" .= c
+                ]
+        postUserRegister' reg brig !!! const 403 === statusCode
```
akshaymankar, line 1092: We should also assert on the error label here to ensure any future restrictions don't make this test not test the thing we want to test.

services/brig/test/integration/API/User/Account.hs
```diff
@@ -152,15 +157,19 @@ testCreateUserWithPreverified brig aws = do
...
+          usr <- postUserRegister reg brig
+          let uid = userId usr
+          get (brig . path "/self" . zUser uid) !!! do
+            const 200 === statusCode
+            const (Just e) === (userEmail <=< responseJsonMaybe)
+          liftIO $ Util.assertUserJournalQueue "user activate" aws (userActivateJournaled usr)
   liftIO $ Util.assertEmptyUserJournalQueue aws
 
-testCreateUser :: Brig -> Galley -> Http ()
+testCreateUser :: Brig -> Galley -> Http () -- TODO: this has nothing to do with /register.  what's going on here?
```
akshaymankar, line 172: I think it got refactored into `randomUser` at some point. It may even have been made redundant with all the other tests, maybe we should verify this and delete it. Or, rename the test to ensure it is reflects the assertions which are about the conversations.

docs/reference/user/registration.md
```diff
@@ -1,6 +1,6 @@
 # Registration {#RefRegistration}
 
-_Author: Artyom Kazak_
+_Authors: Artyom Kazak, Matthias Fischmann_
```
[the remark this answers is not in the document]
fisx: This pattern was introduced by @neongreen, and I can't give you a good reason, just wanted to keep it consistent.

services/brig/test/integration/API/User/Account.hs
fisx: I forgot to mention that I ran the test suite with restricted user creation enabled, and changed all the failing tests [like this](https://github.com/wireapp/wire-server/blob/2c19f78df660a4b2bc783dbc2a729e194d441be6/services/brig/test/integration/API/User/Account.hs#L218).  I went through all tests mentioning "/register" in the description, and verified that just the right tests failed without that change.  So I have tested everything exhaustively, kind of, but I didn't persist that test.

The reason is that it would be a lot of code that isn't only now expensive to write, but will impose a recurring cost on everybody who runs or consults or updates our integration tests.  The code that this would test is literally [these two lines](https://github.com/wireapp/wire-server/blob/2c19f78df660a4b2bc783dbc2a729e194d441be6/services/brig/src/Brig/API/User.hs#L326-L327), and we shouldn't be planning to change those without thinking about how to make this more unit-testable.

I think we shouldn't try to do everything that can be done to improve the quality of our code, but we should think about what's most effective.

Is this a good answer?  Are you ok with leaving the comment now?  If you insist that I resolve this FUTUREWORK, every test case I've extended by the one trivial line will become about 1,5-2 times as long and that much harder to understand and change.

services/brig/test/integration/API/User/Account.hs
fisx: 0091efc62fcfe1b45af21941d0946675dc2dc1ab

Thanks for making me do that!  :)  I am emberrassed to admit I was just lazy.

services/brig/test/integration/API/User/Account.hs
fisx: I checked again, and `randomUser` is posting to `/i/user`, not to `/register`.

I don't really want to complicate this PR with a completely unrelated problem, and again I don't think this is a low-hanging fruit for us to attack now.  I just wanted to leave a note for the next person to pass by here that yes, this hasn't made a lot of sense in a while.

fisx:
Thanks for the comments.  I think if we disagree on anything it's priorities and what constitues low-hanging fruits.  I hope you can live with this now?

docs/reference/user/registration.md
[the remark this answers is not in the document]
neongreen: Git blame is easily spoiled by e.g. reformatting.

My intended semantics for `author:` was "this person is a good first choice to bug if you have a question". Yes, the team is small, but — at least for me — having to arbitrarily pick one person is annoying enough that often I don't ask any questions at all.

At least this was the intended usecase. If it doesn't happen in practice, then there are no other reasons for this field, at least from my side.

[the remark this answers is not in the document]
akshaymankar: Thanks for the context!

PR #1162, opened 2020-07-08
Federation: Implement ID mapping (brig)

mheinzel:
See https://github.com/zinfra/backend-issues/issues/1178

Similar to https://github.com/wireapp/wire-server/pull/1134, but for Brig. Unfortunately we still have a lot of different ways of doing things between Galley and Brig, so it needed some modifications. There's also no nice place to abstract this out, so I think some duplication is okay here.

### DIFF
--- a/services/brig/src/Brig/IO/Intra/IdMapping.hs
+++ b/services/brig/src/Brig/IO/Intra/IdMapping.hs
@@ -1,0 +1,46 @@
+-- This file is part of the Wire Server implementation.
+--
+-- Copyright (C) 2020 Wire Swiss GmbH <opensource@wire.com>
+--
+-- This program is free software: you can redistribute it and/or modify it under
+-- the terms of the GNU Affero General Public License as published by the Free
+-- Software Foundation, either version 3 of the License, or (at your option) any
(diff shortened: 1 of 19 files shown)


services/brig/src/Brig/IO/Intra/IdMapping.hs
```diff
@@ -0,0 +1,46 @@
...
+import Imports
+import Network.HTTP.Types (StdMethod (POST), status200, status403)
+import qualified System.Logger.Class as Log
+
+-- | Calls 'Brig.API.IdMapping.postIdMappingH'.
+createIdMappingInGalley :: PostIdMappingRequest -> AppIO ()
+createIdMappingInGalley body = do
+  Log.debug $ remote "galley" . Log.msg (Log.val "Create ID mapping")
+  st <- statusCode <$> galleyRequest POST req
+  when (st /= 403) $ do
```
jschaul, line 37: Shouldn't this condition be reversed? `when (st == 403) $ do` ?

mheinzel:
Rebased and also added three commits with improvements discussed with @jschaul.

PR #1165, opened 2020-07-13
hscim: add license headers

mheinzel:
For consistency, we should add them there, too. The tooling (`make add-license`) already picks it up.


mheinzel:
`unittests` and `ormolu` are not triggered properly again, but this PR doesn't change any code and I ran the checks locally, so I'll just merge now.

PR #1166, opened 2020-07-13
Upgrade stack to 2.3.1

mheinzel:
See https://github.com/zinfra/backend-issues/issues/1466

It seems like this is the only place we need to change for CI. If it turns out that it isn't, I'll fix it and document it for next time. After merging, we need to trigger the `alpine-prebuilder` job.

I also updated the installation instructions for Stack, with two related question I'll add as comments.

### DIFF
--- a/docs/developer/dependencies.md
+++ b/docs/developer/dependencies.md
@@ -63,12 +63,6 @@ Please refer to [Stack's installation instructions](https://docs.haskellstack.or
 
 When you're done, ensure `stack --version` is recent, ideally the same as `STACK_ALPINE_VERSION` in [`build/alpine/Dockerfile.prebuilder`](../../build/alpine/Dockerfile.prebuilder).
 
-You may wish to make executables installed by stack available, by e.g. adding the following to your shell profile:
-
-```bash
-export PATH=~/.local/bin:$PATH
-```
-
 ### Ubuntu / Debian
 _Note_: The packaged versions of `haskell-stack` are too old. It is recommended to follow the generic instructions or to use stack to update stack (`stack upgrade`).
 
--- a/docs/developer/dependencies.md
+++ b/docs/developer/dependencies.md
@@ -17,7 +17,7 @@ sudo dnf install -y pkgconfig haskell-platform libstdc++-devel libstdc++-static 
 
 ### Ubuntu / Debian:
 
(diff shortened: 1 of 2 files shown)


docs/developer/dependencies.md
```diff
@@ -59,16 +59,18 @@ sudo installer -pkg /Library/Developer/CommandLineTools/Packages/macOS_SDK_heade
 
 ## Haskell Stack
 
-When you're done, ensure `stack --version` is >= 1.6.5
+Please refer to [Stack's installation instructions](https://docs.haskellstack.org/en/stable/README/#how-to-install).
+
+When you're done, ensure `stack --version` is recent, ideally the same as `STACK_ALPINE_VERSION` in [`build/alpine/Dockerfile.prebuilder`](../../build/alpine/Dockerfile.prebuilder).
```
mheinzel, line 64: Do we want the Dockerfile to be the main source of truth for the Stack version? We could also add the version we're using here (of course with the cost that we always need to remember to update both places).

docs/developer/dependencies.md
```diff
@@ -59,16 +59,18 @@ sudo installer -pkg /Library/Developer/CommandLineTools/Packages/macOS_SDK_heade
...
 
 You may wish to make executables installed by stack available, by e.g. adding the following to your shell profile:
 
 ```bash
 export PATH=~/.local/bin:$PATH
 ```
```
mheinzel, line 70: Should we also update this?

Our Makefile installs stuff into `dist`, so running `make install` and following this hint will still not make the binaries available.
- Should we change it to `<path_to_project>/dist`?
- Or add a sentence saying that it depends on whether you use the Makefile or plain `stack install`?
- Or leave it out? (`stack install` will by itself already leave a note saying "Warning: Installation path /home/mheinzel/dev/wire-server/dist not found on the PATH environment variable.")

docs/developer/dependencies.md
jschaul: I think it's okay to have the docker image be the "source of truth" so we don't need to change the version in multiple places.

docs/developer/dependencies.md
jschaul: I think we could remove that entire paragraph - I never run `brig` globally. If people use stack elsewhere, or use it to install globally-available binaries, then they can add that to their path, but it isn't needed for wire-server development as far as I can see.

mheinzel: Thanks, I removed it.

PR #1168, opened 2020-07-13
Pending invitations by email lookup

jschaul:
Context: https://github.com/zinfra/backend-issues/issues/1489

### DIFF
--- a/services/brig/brig.cabal
+++ b/services/brig/brig.cabal
@@ -4,7 +4,7 @@ cabal-version: 2.0
 --
 -- see: https://github.com/sol/hpack
 --
--- hash: 55852968c279333fd398a37b642b9e7428944ec63da4b95f2d3e0d02b1d2df28
+-- hash: 569c2e513825afb99305f83cb7146a6dd4c42b06b991923f534ad980a8bc8c2e
 
 name:           brig
 version:        1.35.0
@@ -439,6 +439,7 @@ executable brig-schema
       V58
       V59
       V60_AddFederationIdMapping
+      V61_team_invitation_email
       V9
       Paths_brig
   hs-source-dirs:
--- a/services/brig/src/Brig/Team/API.hs
+++ b/services/brig/src/Brig/Team/API.hs
@@ -39,7 +39,7 @@ import Brig.Team.Util (ensurePermissionToAddUser, ensurePermissions)
 import Brig.Types.Intra (AccountStatus (..))
 import Brig.Types.Team (TeamSize)
 import Brig.Types.Team.Invitation
-import Brig.Types.User (InvitationCode, emailIdentity)
+import Brig.Types.User (Email, InvitationCode, emailIdentity)
 import qualified Brig.User.Search.Index as ESIndex
 import Control.Lens ((^.), view)
 import Data.Aeson hiding (json)
@@ -50,11 +50,11 @@ import Data.Range
 import qualified Data.Swagger.Build.Api as Doc
 import qualified Galley.Types.Teams as Team
 import qualified Galley.Types.Teams.Intra as Team
-import Imports
+import Imports hiding (head)
 import Network.HTTP.Types.Status
 import Network.Wai (Response)
 import Network.Wai.Predicate hiding (and, result, setStatus)
-import Network.Wai.Routing hiding (head)
+import Network.Wai.Routing
 import Network.Wai.Utilities hiding (code, message)
 import Network.Wai.Utilities.Swagger (document)
 import qualified Network.Wai.Utilities.Swagger as Doc
@@ -141,6 +141,18 @@ routesPublic = do
     Doc.returns (Doc.ref Public.modelTeamInvitation)
     Doc.response 200 "Invitation successful." Doc.end
     Doc.errorResponse invalidInvitationCode
+
+  -- FUTUREWORK: Add another endpoint to allow resending of invitation codes
+  head "/teams/invitations/by-email" (continue headInvitationsByEmailH) $
+    accept "application" "json"
+      .&. query "email"
+  document "HEAD" "headInvitationPending" $ do
+    Doc.summary "Check if there is an invitation pending given an email address."
+    Doc.parameter Doc.Query "email" Doc.bytes' $
+      Doc.description "Email address"
+    Doc.response 200 "Pending invitation exists." Doc.end
+    Doc.response 404 "No pending invitations exists." Doc.end
+    Doc.response 409 "Multiple conflicting invitations to different teams exists." Doc.end
 
 routesInternal :: Routes a Handler ()
 routesInternal = do
@@ -274,6 +286,14 @@ getInvitationByCode :: Public.InvitationCode -> Handler Public.Invitation
 getInvitationByCode c = do
   inv <- lift $ DB.lookupInvitationByCode c
   maybe (throwStd invalidInvitationCode) return inv
+
+headInvitationsByEmailH :: JSON ::: Email -> Handler Response
+headInvitationsByEmailH (_ ::: e) = do
+  inv <- lift $ DB.lookupInvitationInfoByEmail e
+  return $ case inv of
+    DB.InvitationByEmail _ -> setStatus status200 empty
+    DB.InvitationByEmailNotFound -> setStatus status404 empty
+    DB.InvitationByEmailMoreThanOne -> setStatus status409 empty
 
 suspendTeamH :: JSON ::: TeamId -> Handler Response
 suspendTeamH (_ ::: tid) = do
--- a/services/brig/src/Brig/Team/DB.hs
+++ b/services/brig/src/Brig/Team/DB.hs
@@ -28,9 +28,12 @@ module Brig.Team.DB
     lookupInvitations,
     lookupInvitationByCode,
     lookupInvitationInfo,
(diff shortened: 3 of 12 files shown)


services/brig/brig.cabal
```diff
@@ -434,6 +434,7 @@ executable brig-schema
       V57
       V58
       V59
+      V60
```
fisx, line 437: akshay started the convention of adding descriptions to the migrations, a la `V60_Invitations_by_Email`.  I like that convention.

services/brig/schema/src/V60.hs
```diff
@@ -0,0 +1,41 @@
...
+
+migration :: Migration
+migration =
+  Migration 60 "Add team_invitation_email table" $
+    schema'
+      [r|
+        CREATE TABLE team_invitation_email
+            ( email text PRIMARY KEY
+            , team set<uuid>
+            , id uuid
```
fisx, line 38:
```suggestion
            , user uuid
```

or `user_id`, or even `uid`, don't care.  just something indicating that yes dear reader, your hunch is correct, this is the user id.

services/brig/schema/src/V60.hs
```diff
@@ -0,0 +1,41 @@
...
+            , codes set<ascii>
+            )
```
fisx, line 40: Move user to the end?  And it would be slightly nicer to know which team id belongs to which invitation code.  So, perhaps:

```
  ( email text
  , code ascii
  , team uuid
  , user uuid
  ) primary key (email, code)
```

services/brig/src/Brig/Team/API.hs
```diff
@@ -142,8 +143,23 @@ routesPublic = do
...
+  document "HEAD" "headInvitationPending" $ do
+    Doc.summary "Check if there is an invitation pending given an email address."
+    Doc.parameter Doc.Query "email" Doc.bytes' $
+      Doc.description "Email address"
+    Doc.response 200 "Pending invitation exists." Doc.end
+    Doc.response 404 "No pending invitation exists." Doc.end
+
 routesInternal :: Routes a Handler ()
 routesInternal = do
+  -- TODO: Just needed for tests? Do we even need this?
```
fisx, line 158: in tests it'd be ok to use the external end-point, no?

services/brig/src/Brig/Team/API.hs
```diff
@@ -275,6 +291,23 @@ getInvitationByCode c = do
   inv <- lift $ DB.lookupInvitationByCode c
   maybe (throwStd invalidInvitationCode) return inv
 
+headInvitationPendingH :: JSON ::: Email -> Handler Response
```
fisx, line 294:
```suggestion
headInvitationsByEmailH :: JSON ::: Email -> Handler Response
```

services/brig/src/Brig/Team/DB.hs
```diff
@@ -174,6 +180,42 @@ lookupInvitationInfo ic@(InvitationCode c)
     cqlInvitationInfo :: PrepQuery R (Identity InvitationCode) (TeamId, InvitationId)
     cqlInvitationInfo = "SELECT team, id FROM team_invitation_info WHERE code = ?"
 
+lookupInvitationByEmail :: (Log.MonadLogger m, MonadClient m) => Email -> m (Maybe Invitation)
+lookupInvitationByEmail e = lookupInvitationEmail e >>= \case
+  Just InvitationInfo {..} -> lookupInvitation iiTeam iiInvId
+  _ -> return Nothing
+
+lookupInvitationEmail :: (Log.MonadLogger m, MonadClient m) => Email -> m (Maybe InvitationInfo)
```
fisx, line 188:
```suggestion
lookupInvitationInfoByEmail :: (Log.MonadLogger m, MonadClient m) => Email -> m (Maybe InvitationInfo)
```

fisx: Also, I'd introduce a custom maybe here with two nothings, and pass that through to the handler where you can map the two different nothings on two different status codes and / or least error labels.

```haskell
data InvitationByEmail =
    InvitationByEmail InvitationInfo
  | InvitationByEmailNotFound
  | InvitationByEmailMoreThanOne
```

services/brig/src/Brig/Team/DB.hs
```diff
@@ -174,6 +180,42 @@ lookupInvitationInfo ic@(InvitationCode c)
...
+      (t : [], code : _code2 : _) ->
+        -- multiple invites, but all from the same team, so can be deemed equivalent (TODO: can they? Different access role?)
+        return $ Just $ InvitationInfo code t invId
+      (_ : _ : _, _ : _ : _) -> do
+        -- edge case: more than one pending invite
+        -- We can treat the "none or multiple pending invites" the same way in the UI:
+        -- "please talk to your Wire administrator to figure out what to do"
+        Log.info $ Log.msg (Log.val "team_invidation_email: multiple pending invites from different team for the same email")
+        return $ Nothing
+      (_, _) -> do
```
fisx, line 212: If you change the schema as I've described, I think you can get rid of this case.

services/brig/schema/src/V60.hs
jschaul: but it's not in this case, it's the invitation id. There's no user created yet at this point. I'll add a comment to the file.

services/brig/brig.cabal
jschaul: :+1:

services/brig/schema/src/V60.hs
fisx: no, this is proof it needs a better name, not a comment!  please call it `invitation`, then.

fisx: you're not calling `teams` `ids` either, right?

jschaul: ok ok! :) I'll rename.

services/brig/schema/src/V60.hs
jschaul: A primary key on two elements is a good idea and simplifies the cases; I went for this now:

```
        CREATE TABLE team_invitation_email
            ( email text
            , team uuid
            , invitation uuid
            , code ascii
            , primary key (email, team)
            )
```

services/brig/src/Brig/Team/DB.hs
jschaul: :+1:

services/brig/src/Brig/Team/DB.hs
jschaul: Indeed.

services/brig/schema/src/V60.hs
fisx: [Nit-pick] that means two invitations from the same team overwrite each other in chronological order.  I think that's good, but should there be a comment about that somewhere?  (I don't think we should leave comments here in the cql migration code, it's hard to find and may be lost during migration squashing.)

services/brig/src/Brig/Team/API.hs
jschaul: yes

services/brig/src/Brig/Team/API.hs
```diff
@@ -142,6 +142,18 @@ routesPublic = do
     Doc.response 200 "Invitation successful." Doc.end
     Doc.errorResponse invalidInvitationCode
 
+  -- FUTUREWORK: Add another endpoint to allow resending of invitation codes
+  head "/teams/invitations/by-email" (continue headInvitationsByEmailH) $
```
jschaul, line 146: Is this a good name for an endpoint? Anyone has suggestions for improvements?

fisx: it's perfect!  :)

services/brig/test/integration/API/Team.hs
```diff
@@ -152,6 +154,40 @@ testInvitationEmail brig = do
...
+  email <- randomEmail
+  (owner, tid) <- createUserWithTeam brig
+  let invite = stdInvitationRequest email (Name "Bob") Nothing Nothing
+  void $ postInvitation brig tid owner invite
+  inv :: Invitation <- responseJsonError =<< postInvitation brig tid owner invite
+  -- expect an invitation to be found querying with email after invite
+  headInvitationByEmail brig email 200
+  void $ registerInvite brig tid inv email
+  -- expect a 404 after invitation has been used.
+  headInvitationByEmail brig email 404
```
fisx, line 184: Is this removed synchronously?  Otherwise you may want to wrap this in a call to [retrying or similar](https://hoogle.zinfra.io/file/root/.stack/snapshots/x86_64-linux/01af7b6eb1c680941d3ef08786685b31d2470c8b36e981c7bdad15d91f453def/8.6.5/doc/retry-0.8.1.0/Control-Retry.html) here.

jschaul: Yes, it's removed synchronously in both the `Brig/Team/API.hs` deleteInvitation(s) endpoints, as well as on user creation after user activation - all synchronously. So no need for retrying wrapping.

jschaul:
This PR now only depends on

* https://github.com/wireapp/wire-server/pull/1162 being merged (due to schema migration conflict), and this being re-based
* another run of `make git-add-cassandra-schema`

jschaul:
CI failure is unrelated to this PR (tracked in  zinfra/backend-issues#1600 ).

PR #1171, opened 2020-07-14
Count suspended team members correctly

akshaymankar:
Fixes https://github.com/zinfra/backend-issues/issues/1523


services/brig/src/Brig/User/Search/Index/Types.hs
```diff
@@ -109,6 +114,7 @@ mkIndexUser u v =
   IndexUser
     { _iuUserId = u,
       _iuVersion = v,
+      _iuAccountStatus = Nothing,
```
mheinzel, line 117: Minor style nit: `_iuAccountStatus` is the last line of the record in all other places.

PR #1174, opened 2020-07-17
Disallow team creation when setRestrictUserCreation is true

tiago-loureiro:
Follow up to https://github.com/wireapp/wire-server/pull/1161 where the team creation case was missed.

Fixes: https://github.com/zinfra/backend-issues/issues/1602

### DIFF
--- a/services/brig/src/Brig/API/User.hs
+++ b/services/brig/src/Brig/API/User.hs
@@ -322,17 +322,17 @@ checkRestrictedUserCreation new = do
   restrictPlease <- lift . asks $ fromMaybe False . setRestrictUserCreation . view settings
   when
     ( restrictPlease
-           && not (isTeamMember nTeam)
-           && not (isEphemeral nExpires)
+        && not (isTeamMember nTeam)
+        && not (isEphemeral nExpires)
     )
     $ throwE UserCreationRestricted
   where
-    isTeamMember (Just (NewTeamMember _))    = True
+    isTeamMember (Just (NewTeamMember _)) = True
     isTeamMember (Just (NewTeamMemberSSO _)) = True
-    isTeamMember _                           = False
-
+    isTeamMember _ = False
     isEphemeral (Just _) = True
-    isEphemeral _        = False
+    isEphemeral _ = False
+
 -------------------------------------------------------------------------------
 -- Update Profile
 
--- a/services/brig/src/Brig/API/User.hs
+++ b/services/brig/src/Brig/API/User.hs
@@ -316,20 +316,22 @@ createUser new@NewUser {..} = do
 -- | docs/reference/user/registration.md {#RefRestrictRegistration}.
 checkRestrictedUserCreation :: NewUser -> ExceptT CreateUserError AppIO ()
 checkRestrictedUserCreation new = do
-  let memail = newUserEmail new
-      mphone = newUserPhone new
-      msso = newUserSSOId new
-      mteam = newUserTeam new
+  let nTeam = newUserTeam new
+      nExpires = newUserExpiresIn new
+
   restrictPlease <- lift . asks $ fromMaybe False . setRestrictUserCreation . view settings
   when
     ( restrictPlease
-        && or [isJust memail, isJust mphone]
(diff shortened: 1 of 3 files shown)


services/brig/src/Brig/API/User.hs
```diff
@@ -324,10 +324,12 @@ checkRestrictedUserCreation new = do
   when
     ( restrictPlease
         && or [isJust memail, isJust mphone]
-        && and [isNothing msso, isNothing mteam]
+        && and [isNothing msso, isNotTeamMember mteam]
```
akshaymankar, line 327: This implies that [the doc](https://github.com/wireapp/wire-server/blob/0e5bf97618cb4331f29c9005002f2d971fc64f4e/docs/reference/user/registration.md#details) added in the previous PR lies about email/phone being present as sufficient indicator of it being either a team creator or personal accounts. Let's also edit that.

services/brig/test/integration/API/User/Account.hs
```diff
@@ -1067,31 +1067,64 @@ testDomainsBlockedForRegistration opts brig = withDomainsBlockedForRegistration
...
+              [ "name" .= Name "Alice",
+                "email" .= fromEmail e,
+                "email_code" .= ("123456" :: Text),
+                "team" .= object ["name" .= ("Alice team" :: Text), "icon" .= ("default" :: Text), "binding" .= True],
+                "password" .= PlainTextPassword "123123123"
+              ]
+    postUserRegister' teamCreator brig !!! do
+        const 403 === statusCode
+        const (Just "user-creation-restricted") === (^? AesonL.key "label" . AesonL._String) . (responseJsonUnsafe @Value)
+
```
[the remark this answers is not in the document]
tiago-loureiro: Yes, good point. I added it with https://github.com/wireapp/wire-server/pull/1174/commits/ff810d1e0bb8f2d4dc7feaee5396d47887e7f8b2

tiago-loureiro:
> Apart from that I have a rough idea about how we can make this easier to comprehend, let's discuss that if we have time:
While I looked into this deeper I encountered the NewUserOrigin object. I think we can simplify this boolean algebra into a case statement which only allows registration if the origin is Just (NewUserOriginTeamUser (NewTeamMember _)) or Nothing. This way our code would say exactly what the requirement is: Allow creation only if it is a team member or an ephemeral account. I could be missing something here, Let me know if it makes sense.

I think this is really quite useful so I went ahead and did it [here](https://github.com/wireapp/wire-server/pull/1174/commits/2f10318e29fcdb4afcbf93e33bde0ec2fa3f5586) - please double check my logic 😅 

If we agree with it, I will update the docs (which are then greatly simplified)

services/brig/src/Brig/API/User.hs
```diff
@@ -316,20 +316,22 @@ createUser new@NewUser {..} = do
...
-      mteam = newUserTeam new
+  let nTeam = newUserTeam new
+      nExpires = newUserExpiresIn new
+
   restrictPlease <- lift . asks $ fromMaybe False . setRestrictUserCreation . view settings
   when
     ( restrictPlease
-        && or [isJust memail, isJust mphone]
-        && and [isNothing msso, isNotTeamMember mteam]
+           && not (or [isTeamMember nTeam, isEphemeral nExpires])
```
akshaymankar, line 325:
```suggestion
           && not (isTeamMember nTeam)
           && not (isEphemeral nExpires)
```

Looks easier to reason about like this to me (because I am very challenged in boolean algebra). Feel free to ignore.

akshaymankar:
> I think this is really quite useful so I went ahead and did it [here](https://github.com/wireapp/wire-server/pull/1174/commits/2f10318e29fcdb4afcbf93e33bde0ec2fa3f5586) - please double check my logic sweat_smile
> 
> If we agree with it, I will update the docs (which are then greatly simplified)

It makes sense to me. I added a minor and  purely aesthetic suggestion, feel free to ignore. I think we can go ahead with updating docs :+1:

tiago-loureiro:
Should be good to go now.

PR #1177, opened 2020-07-22
Discover and expose SFT servers in /calls/config/v2

jschaul:


libs/wire-api/src/Wire/API/Call/TURN.hs
```diff
@@ -93,36 +100,71 @@ import Wire.API.Arbitrary (Arbitrary (arbitrary), GenericUniform (..))
 -- | A configuration object resembling \"RTCConfiguration\"
 --
 -- The \"ttl\" field is a proprietary extension
+-- The \"sft_servers\" field is a proprietary extension
 --
 -- cf. https://developer.mozilla.org/en-US/docs/Web/API/RTCPeerConnection/RTCPeerConnection#RTCConfiguration_dictionary
 data RTCConfiguration = RTCConfiguration
-  { _rtcConfIceServers :: List1 RTCIceServer,
+  { _rtcConfIceServers :: NonEmpty RTCIceServer,
+    _rtcConfSftServers :: Maybe (NonEmpty SFTServer),
```
fisx, line 108: we just talked about how this should never go from `Just` to `Nothing`, but now it occurs to me that: why?  what's the semantics of that case, and what do we report to the client?  could use a type wtih more expressive constructors (as you suggested) plus haddocks on what it means.

services/brig/brig.cabal
```diff
@@ -103,7 +104,7 @@ library
   hs-source-dirs:
       src
   default-extensions: AllowAmbiguousTypes BangPatterns ConstraintKinds DataKinds DefaultSignatures DerivingStrategies DeriveFunctor DeriveGeneric DeriveLift DeriveTraversable EmptyCase FlexibleContexts FlexibleInstances FunctionalDependencies GADTs InstanceSigs KindSignatures LambdaCase MultiParamTypeClasses MultiWayIf NamedFieldPuns NoImplicitPrelude OverloadedStrings PackageImports PatternSynonyms PolyKinds QuasiQuotes RankNTypes ScopedTypeVariables StandaloneDeriving TemplateHaskell TupleSections TypeApplications TypeFamilies TypeFamilyDependencies TypeOperators UndecidableInstances ViewPatterns
-  ghc-options: -O2 -Wall -Wincomplete-uni-patterns -Wincomplete-record-updates -Wpartial-fields -fwarn-tabs -optP-Wno-nonportable-include-path -funbox-strict-fields
+  ghc-options: -O2 -Wall -Wincomplete-uni-patterns -Wincomplete-record-updates -Wpartial-fields -fwarn-tabs -optP-Wno-nonportable-include-path -funbox-strict-fields -fplugin=Polysemy.Plugin
```
fisx, line 107: Should we do this on a per-module basis?  Not sure, perhaps not.  It's more explicit if few modules are affected, and annoying if many are.

services/brig/src/Brig/Run.hs
```diff
@@ -56,9 +60,13 @@ run o = do
     Async.async
       $ AWS.execute (e ^. awsEnv)
       $ AWS.listen throttleMillis q (runAppT e . SesNotification.onEvent)
+  sftDiscovery <- case e ^. sftEnv of
+    Nothing -> Async.async (pure ()) -- TODO: This looks fishy
```
fisx, line 64: it's for backends that don't want to do calling.  maybe we should just handle this differently in the calling config end-points as well?  like, return 404 instead of an empty sft server list?  or should we not allow this case?  force people to run wire with calling always?

fisx: We could make `sftDiscovery` a `Maybe` and `mapM_` into it for cancelling.  That would be slightly nicer; no lingering thread.

libs/dns-srv/src/Wire/Network/DNS/Effect.hs
```diff
@@ -0,0 +1,41 @@
...
+--
+-- This program is distributed in the hope that it will be useful, but WITHOUT
+-- ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+-- FOR A PARTICULAR PURPOSE. See the GNU Affero General Public License for more
+-- details.
+--
+-- You should have received a copy of the GNU Affero General Public License along
+-- with this program. If not, see <https://www.gnu.org/licenses/>.
+
+module Wire.Network.DNS.Effect where
```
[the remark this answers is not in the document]
akshaymankar: I will eliminate duplication as part of this PR.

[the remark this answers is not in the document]
akshaymankar: I think that one is the quickest start you can get.

libs/dns-srv/src/Wire/Network/DNS/Effect.hs
```diff
@@ -0,0 +1,41 @@
...
+runDNSLookupDefault =
+  interpret $ \l -> do
+    rs <- embed $ DNS.makeResolvSeed DNS.defaultResolvConf
+    embed $ DNS.withResolver rs $ \resolver ->
+      case l of
+        LookupSRV domain -> interpretResponse <$> DNS.lookupSRV resolver domain
+
+-- class Monad m => MonadDNSLookup m where
+--   monadLookupSRV :: Domain -> m SrvResponse
+-- I don't know how to make this choice at runtime
```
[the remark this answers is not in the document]
akshaymankar: Yeah, I wrote this comment as a placeholder for the effect system discussion.

libs/wire-api/src/Wire/API/Call/TURN.hs
akshaymankar: The semantics are we don't want to erase the stale data on a failed lookup. This is why I want to create another type which can make these semantics more obvious.

services/brig/brig.cabal
akshaymankar: I think it is fine here, it requires another dependency to be built. If we see polysemy spreading, we could decide later.

fisx:
> a7820b5 

too bad.  :(  can we do this cnoditionally, with a cabal compiler version check or something?

akshaymankar:
> too bad. :( can we do this cnoditionally, with a cabal compiler version check or something?

I think we can have a flag for this, but the thing is completely useless right now. Instead, we should just update our alpine image to have ncurses-dev. I have [this branch](https://github.com/wireapp/wire-server/tree/akshaymankar/alpine-ncurses-dev) that will help us add it back in future, but it requires a lot of compiling and one of us watching it compile. I will get to it after I am done with this PR, if I have enough time before my vacation. Feel free to take it off my hands though :smile:

libs/wire-api/src/Wire/API/Call/Config.hs
```diff
@@ -93,36 +99,71 @@ import Wire.API.Arbitrary (Arbitrary (arbitrary), GenericUniform (..))
...
+      xs -> fail $ "SFTServer can only have exactly one URL, found " <> show (length xs)
+
+sftServer :: HttpsUrl -> SFTServer
+sftServer = SFTServer
+
+modelRtcSftServer :: Doc.Model
+modelRtcSftServer = Doc.defineModel "RTC SFT Server" $ do
+  Doc.description "Inspired by WebRTC 'RTCIceServer' object, contains details of SFT servers"
+  Doc.property "urls" (Doc.array Doc.string') $
+    Doc.description "Array of SFT server addresses of the form 'https://<addr>:<port>'"
```
jschaul, line 166: Can we assume port 443, or will we be explicit and specify 443 in the server URLs?

services/brig/src/Brig/Calling/API.hs
```diff
@@ -103,34 +110,37 @@ getCallsConfig _ _ = do
...
   -- randomize again (as limitedList partially re-orders uris)
   finalUris <- liftIO $ randomize limitedUris
   srvs <- for finalUris $ \uri -> do
     u <- liftIO $ genUsername tTTL prng
-    pure $ Public.rtcIceServer (List1.singleton uri) u (computeCred sha secret u)
-  pure $ Public.rtcConfiguration srvs cTTL
+    pure $ Public.rtcIceServer (uri :| []) u (computeCred sha secret u)
+  sftSrvEntries <- maybe (pure Nothing) (readIORef . sftServers) mSftEnv
+  -- According to RFC2782, the SRV Entries are supposed to be tried in order of
+  -- priority and weight, but the clients try all of them concurrently, so there
```
jschaul, line 128: I thought we had agreed in https://github.com/zinfra/backend-issues/issues/1453#issue-621270722 as well as https://github.com/zinfra/backend-issues/issues/1453#issuecomment-663070245 to randomize the SFT server list? Client behaviour may change, and even if all of them are currently tried concurrently, the behaviour may in the future only try the first two. In addition, there is no real concurrency, and some servers in the list will "win" by returning the request first. Since this is the only place to control the distribution of requests to servers, i.e. "load balancing", I'm still in favour of randomizing the list here.

services/brig/src/Brig/Calling/Internal.hs
```diff
@@ -0,0 +1,41 @@
...
+import qualified Wire.API.Call.Config as Public
+import Wire.Network.DNS.SRV (SrvTarget (..))
+
+sftServerFromSrvTarget :: SrvTarget -> Public.SFTServer
+sftServerFromSrvTarget (SrvTarget host port) =
+  let uriPort = URI.Port (fromIntegral port)
+      uri =
+        [URI.uri|https://|]
+          & URI.authorityL ?~ URI.Authority Nothing (URI.Host (dropTrailingDot host)) (Just uriPort)
+   in either (error "sftServerFromSrvTarget: invalid https URI") Public.sftServer (mkHttpsUrl uri)
```
jschaul, line 35: What happens in the error case here? Does this lead to a 500? I'm not sure there is a test yet covering this case?

services/brig/src/Brig/Options.hs
```diff
@@ -519,6 +524,26 @@ data CustomerExtensions = CustomerExtensions
 newtype DomainsBlockedForRegistration = DomainsBlockedForRegistration [Domain]
   deriving newtype (Show, FromJSON, Generic)
 
+data SFTOptions = SFTOptions
+  { sftBaseDomain :: !DNS.Domain,
+    sftSRVServiceName :: !(Maybe ByteString),
```
jschaul, line 529: Isn't the service name always `_sft` here? Do we need to make this configurable? The defaults/ fallbacks on not specifying would be nice to document, or move closer to the data type, as it's currently hard to see what these values will be when not set.

We have some `defSomething` values at the bottom of this file, can we add the `_sft` and `10 seconds` here as well?

services/brig/src/Brig/Run.hs
```diff
@@ -56,9 +60,13 @@ run o = do
...
+    Just sftEnv' -> Async.async $ Calling.startSFTServiceDiscovery sftEnv'
```
jschaul, line 65: Would it make sense to use a logic like in https://github.com/wireapp/wire-server/blob/develop/services/gundeck/src/Gundeck/ThreadBudget/Internal.hs#L243-L251 to restart that thread if it dies?

services/brig/test/unit/Test/Brig/Calling/Internal.hs
```diff
@@ -0,0 +1,48 @@
...
+import Wire.Network.DNS.SRV (SrvTarget (SrvTarget))
+
+tests :: TestTree
+tests =
+  testGroup "Calling.API" $
+    [ testGroup "sftServerFromSrvTarget" $
+        [ testCase "when srvTarget ends with a dot" $ do
+            let Right expectedServer = sftServer <$> mkHttpsUrl [URI.uri|https://sft1.env.example.com:9364|]
+            assertEqual
+              "the dot should be stripped from sft server"
```
jschaul, line 38: The dot stripping, and converting of srvTarget to HttpsUrl seems more general than only useful in brig calling; i.e. I think we can also make use of that logic for federation.

Do you think this test and the related function could be moved from Brig calling to dns-util ?

services/brig/src/Brig/Options.hs
jschaul: Done in fde677111beaecd2d77b63a86e024cbe62ca79a8

libs/dns-srv/src/Wire/Network/DNS/Effect.hs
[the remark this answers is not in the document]
akshaymankar: This is done now.

services/brig/brig.cabal
akshaymankar: Sadly it had to go away. Resolving this.

services/brig/src/Brig/Calling/Internal.hs
akshaymankar: It would, but I can't think of any way that this would fail. I will think about how this `error` call can be removed.

services/brig/test/unit/Test/Brig/Calling/Internal.hs
akshaymankar: I would say let's move it when we need to, I think the `HttpsUrl` type could use some refactoring to make sure there is at least one way to create it without running into the `Either`.

services/brig/src/Brig/Run.hs
akshaymankar: Looks much better like this, thanks!

libs/wire-api/src/Wire/API/Call/Config.hs
akshaymankar: Stripping the 443 is a bit of work and having it doesn't really cause any issues, so I would say it is fine if we're always explicit about the port.

services/brig/src/Brig/Calling/API.hs
akshaymankar: Implemented in 009961d852839e28bb5b6844a0399775025dbe64

services/brig/src/Brig/Calling/Internal.hs
akshaymankar: Let me know if 7ec72b1b828c2fc4d2b22d24d73acac503a40e49 makes sense.

services/brig/src/Brig/Run.hs
akshaymankar: Yeah, we need to implement something like that for all of the threads that we spawn in this function. I think this PR is running for too long, let's tackle this in https://github.com/zinfra/backend-issues/issues/1647.

libs/dns-srv/src/Wire/Network/DNS/SRV.hs
```diff
@@ -0,0 +1,98 @@
...
+  | SrvAvailable (NonEmpty SrvEntry)
+  | SrvResponseError DNSError
+  deriving (Eq, Show)
+
+interpretResponse :: Either DNSError [(Word16, Word16, Word16, Domain)] -> SrvResponse
+interpretResponse = \case
+  Left err -> SrvResponseError err
+  Right [] -> SrvNotAvailable
+  Right [(_, _, _, ".")] -> SrvNotAvailable -- According to RFC2782
+  Right (r : rs) -> SrvAvailable $ fmap toSrvEntry (r :| rs)
```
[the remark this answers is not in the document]
akshaymankar: This is wrong, we are OK with `rs` being empty, it would signify that there is only one server.

services/brig/test/integration/API/Calling.hs
```diff
@@ -75,6 +86,26 @@ testCallsConfigMultiple b turnUpdater = do
...
+  uid <- userId <$> randomUser b
+  cfg <- getTurnConfigurationV2 uid b
+  liftIO $
+    assertEqual
+      "when SFT discovery is not enabled, sft_servers shouldn't be returned"
+      Nothing
+      (cfg ^. rtcConfSftServers)
+  withSettingsOverrides (opts & Opts.sftL ?~ Opts.SFTOptions "integration-tests.zinfra.io" Nothing (Just 0.001)) $ do
+    cfg1 <- retryWhileN 10 (isNothing . view rtcConfSftServers) (getTurnConfigurationV2 uid b)
+    -- These values are controlled by https://github.com/zinfra/cailleach/tree/77ca2d23cf2959aa183dd945d0a0b13537a8950d/environments/dns-integration-tests
```
fisx, line 100: it's not ideal that we're having another link into a private repo here, but not linking it would be worse IMO, so :+1:

akshaymankar: Yeah, I chose lesser of the two evils. I was thinking of committing that code in this repo, but decided to not invent yet another way to do infra (and make a third evil?).

services/brig/test/integration/API/Calling.hs
```diff
@@ -15,37 +17,46 @@
...
-    testGroup
-      "turn"
-      [ test m "basic /calls/config - 200" $ testCallsConfig b,
-        -- FIXME: requires tests to run on same host as brig
-        test m "multiple servers /calls/config - 200" . withTurnFile turn $ testCallsConfigMultiple b,
-        test m "multiple servers /calls/config/v2 - 200" . withTurnFile turnV2 $ testCallsConfigMultipleV2 b
+import Wire.API.Call.Config
+
+tests :: Manager -> Brig -> Opts.Opts -> FilePath -> FilePath -> IO TestTree
+tests m b opts turn turnV2 = do
```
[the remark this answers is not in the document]
akshaymankar: I think this is because in some integration tests we do some IO things for all the `testGroups` and here we just kept it similar to that. I think it is better have the `tests` functions be an IO everywhere rather than in only some places. But I don't really feel very strongly on this one.

services/brig/src/Brig/PolyLog.hs
```diff
@@ -0,0 +1,19 @@
...
+
+import Imports
+import Polysemy
+import qualified System.Logger as Log
+
+-- | This effect will help us write tests for log messages
+--
+-- FUTUREWORK: Move this to a separate module if it is required
+--
+-- FUTUREWORK: Either write an orphan instance for MonadLogger or provide
```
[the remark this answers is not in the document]
akshaymankar: I think we cannot really get rid of `tinylog`. We can only get rid of all the `MonadLogger` instances. If we make this orphan (or provide similar functions), we can migrate easily.

libs/types-common/src/Data/Misc.hs
```diff
@@ -236,14 +237,17 @@ instance Cql Milliseconds where
...
 
 mkHttpsUrl :: URIRef Absolute -> Either String HttpsUrl
 mkHttpsUrl uri =
   if uri ^. uriSchemeL . schemeBSL == "https"
     then Right $ HttpsUrl uri
     else Left $ "Non-HTTPS URL: " ++ show uri
 
+ensureHttpsUrl :: URIRef Absolute -> HttpsUrl
+ensureHttpsUrl = HttpsUrl . (uriSchemeL . schemeBSL .~ "https")
+
```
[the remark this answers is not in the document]
akshaymankar: `mkHttpsUrl` is already defined above and it returns `Either String HttpsUrl`. `enforce` and `ensure` have same semantic meaning in my head, so I also don't have much strength in my opinion here.

services/brig/src/Brig/PolyLog.hs
```diff
@@ -0,0 +1,19 @@
...
+-- equivalent functions in System.Logger.Class
+data PolyLog m a where
+  PolyLog :: Log.Level -> (Log.Msg -> Log.Msg) -> PolyLog m ()
+
+makeSem 'PolyLog
+
+runPolyLog :: Member (Embed IO) r => Log.Logger -> Sem (PolyLog ': r) a -> Sem r a
```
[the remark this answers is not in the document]
akshaymankar: How do you mean? This function interprets the effect in terms of `IO`, I don't think it should create a logger, since there are a bunch of log settings we might want to control from wherever the app is being started.

services/brig/src/Brig/Options.hs
```diff
@@ -519,6 +524,26 @@ data CustomerExtensions = CustomerExtensions
...
   deriving newtype (Show, FromJSON, Generic)
 
+data SFTOptions = SFTOptions
+  { sftBaseDomain :: !DNS.Domain,
+    sftSRVServiceName :: !(Maybe ByteString), -- defaults to defSftServiceName if unset
+    sftDiscoveryIntervalSeconds :: !(Maybe DiffTime) -- defaults to defSftDiscoveryIntervalSeconds
+  }
+  deriving (Show, Generic)
+
+instance FromJSON SFTOptions where
```
fisx, line 534: (Wild thought: we should have a couple of strings in the test suite and tests that they parse as server config files.)

services/brig/test/unit/Test/Brig/Calling.hs
```diff
@@ -0,0 +1,206 @@
...
+newtype LogRecorder = LogRecorder {recordedLogs :: IORef [(Log.Level, LByteString)]}
+
+newLogRecorder :: IO LogRecorder
+newLogRecorder = LogRecorder <$> newIORef []
+
+recordLogs :: Member (Embed IO) r => LogRecorder -> Sem (PolyLog ': r) a -> Sem r a
+recordLogs LogRecorder {..} = interpret $ \(PolyLog lvl msg) ->
+  modifyIORef' recordedLogs (++ [(lvl, Log.render (Log.renderDefault ", ") msg)])
+
+ignoreLogs :: Sem (PolyLog ': r) a -> Sem r a
```
fisx, line 61: I don't like the name `PolyLog` (it's a nice name, but it doesn't point out what it refers to clearly enough for my taste).  But we can change it later.

I *do* like the fact that we're introducing polysemi!  :)

libs/dns-srv/src/Wire/Network/DNS/SRV.hs
```diff
@@ -0,0 +1,98 @@
...
+-- FUTUREWORK: maybe improve sorting algorithm here? (with respect to performance and code style)
+--
+-- This function orders the SRV result in accordance with RFC
+-- 2782. It sorts the SRV results in order of priority, and then
+-- uses a random process to order the records with the same
+-- priority based on their weight.
+--
+-- Taken from http://hackage.haskell.org/package/pontarius-xmpp (BSD3 licence) and refactored.
+orderTargets :: [SrvEntry] -> IO [SrvEntry]
+orderTargets =
```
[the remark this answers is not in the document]
akshaymankar: We kept the function there as it was written for federation and will be useful anyway. But ended up not using it as it doesn't match up with how clients decide which server to connect to.

libs/wire-api/src/Wire/API/Call/TURN.hs
akshaymankar: There is new type now. Let me know if it makes sense

services/brig/test/unit/Test/Brig/Calling/Internal.hs
akshaymankar: > I think the `HttpsUrl` type could use some refactoring to make sure there is at least one way to create it without running into the `Either`.

I created another smart constructor.

services/brig/src/Brig/Calling.hs
```diff
@@ -0,0 +1,113 @@
...
+--
+-- This program is distributed in the hope that it will be useful, but WITHOUT
+-- ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+-- FOR A PARTICULAR PURPOSE. See the GNU Affero General Public License for more
+-- details.
+--
+-- You should have received a copy of the GNU Affero General Public License along
+-- with this program. If not, see <https://www.gnu.org/licenses/>.
+
+module Brig.Calling where
```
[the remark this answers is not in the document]
akshaymankar: But I want to use the constructors in the tests. I could always translate it to `Maybe` in the tests but that would make them cumbersome. I think it is OK to leave it at this as it clearly states the semantics and doesn't come in the way of testing.

services/brig/test/unit/Test/Brig/Calling/Internal.hs
akshaymankar: > Otherwise, we can also leave a FUTUREWORK here.

I did this.

services/brig/src/Brig/Calling/Internal.hs
jschaul: :+1:

services/brig/src/Brig/Run.hs
jschaul: Sure, good idea.

libs/dns-srv/src/Wire/Network/DNS/SRV.hs
```diff
@@ -0,0 +1,53 @@
...
+--
+-- This program is distributed in the hope that it will be useful, but WITHOUT
+-- ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+-- FOR A PARTICULAR PURPOSE. See the GNU Affero General Public License for more
+-- details.
+--
+-- You should have received a copy of the GNU Affero General Public License along
+-- with this program. If not, see <https://www.gnu.org/licenses/>.
+
+module Wire.Network.DNS.SRV where
```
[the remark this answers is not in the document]
akshaymankar: The RFC defines what SRV records are. The `dns` package interprets an SRV record into a tuple of 4 things, we named them here for it to be easier to comprehend across the code. Not sure if the mention is worth it. We mention the RFC in the ordering function because it might not be as obvious.

services/brig/test/unit/Test/Brig/Calling.hs
```diff
@@ -0,0 +1,206 @@
...
+  let entry1 = SrvEntry 0 0 (SrvTarget "sft4.foo.example.com." 443)
+      entry2 = SrvEntry 0 0 (SrvTarget "sft5.foo.example.com." 443)
+      newEntries = (entry1 :| [entry2])
+
+  fakeDNSEnv <- newFakeDNSEnv (\_ -> SrvAvailable newEntries)
+  discoveryLoop <- Async.async $ runM . ignoreLogs . runFakeDNSLookup fakeDNSEnv $ sftDiscoveryLoop sftEnv
+  void $ retryEvery10MicrosWhileN 2000 (== 0) (length <$> readIORef (fakeLookupCalls fakeDNSEnv))
+  -- We don't want to stop the loop before it has written to the sftServers IORef
+  void $ retryEvery10MicrosWhileN 2000 (== Discovered newEntries) (readIORef (sftServers sftEnv))
+  Async.cancel discoveryLoop
```
fisx, line 157: Make this a helper function?  `runDiscoveryLoopOnce`?  That would (a) make the tests more concise and (b) also explain a lot better what's going on than just the actual code.

fisx:
Hey github, why can i not reply to https://github.com/wireapp/wire-server/pull/1177#discussion_r462909962?

anyway, what i meant was that this should not take the logger as a function argument, but get it out of a reader effect.  we agree that it shouldn't be created on the fly with the wrong parameters.  :)

fisx:
> https://github.com/wireapp/wire-server/pull/1177#discussion_r463007967

Perhaps we should start writing tests in the production code modules, and merely *calling* them from `/package/tests`?  Anyway, other issue, other PR, other year.

services/brig/test/unit/Test/Brig/Calling.hs
```diff
@@ -0,0 +1,206 @@
...
+  testGroup "Calling" $
+    [ testGroup "mkSFTDomain" $
+        [ testCase "when service name is provided" $
+            assertEqual
+              "should use the service name to form domain"
+              "_foo._tcp.example.com."
+              (mkSFTDomain (SFTOptions "example.com" (Just "foo") Nothing)),
+          testCase "when service name is not provided" $
+            assertEqual
+              "should assume service name to be 'sft'"
```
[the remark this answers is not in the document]
fisx: Never mind, i missed that's already there above.

services/brig/test/unit/Test/Brig/Calling.hs
fisx: it was just pointed out to me that this is slightly trickier than i thought: not only env changes between calls, but also conditions of success.  (also "once" is not a good name, since it loops *until* certain conditions are met.)

PR #1178, opened 2020-07-23
Use ++FAILURES as it will not evaluate to 0

akshaymankar:
When an expression evaluates to 0, its exit code is set to 1. Life is always
interesting when you write bash.
More information:
https://unix.stackexchange.com/questions/32250/why-does-a-0-let-a-return-exit-code-1/32251


tiago-loureiro:
> When an expression evaluates to 0, its exit code is set to 1. Life is always
> interesting when you write bash.
> More information:
> https://unix.stackexchange.com/questions/32250/why-does-a-0-let-a-return-exit-code-1/32251

I was wondering why the CI got "stuck" 🤦

PR #1179, opened 2020-07-24
Cleanup (mostly spar)

fisx:
These are a few very low-hanging fruits, please read commit-by-commit when reviewing, it's quite self-contained.

PR #1145, opened 2020-06-24
Upgrade to ormolu 0.1.2.0

mheinzel:
See https://github.com/zinfra/backend-issues/issues/1623

Don't merge yet. We'll wait for a good moment where we don't have many PRs with Haskell changes in progress.
It should not take much time to rebase this PR then (if we just re-create the `run ormolu` commit instead of rebasing it).

Main [Ormolu changes since 0.0.5.0](https://github.com/tweag/ormolu/blob/master/CHANGELOG.md#ormolu-0120) we care about:
  - Blank lines between definitions in let and while bindings are now preserved. [Issue 554](https://github.com/tweag/ormolu/issues/554).
  - Fixed rendering of comments around if expressions. [Issue 458](https://github.com/tweag/ormolu/issues/458).
  - A bunch of idempotence fixes we occasionally ran into.

Changes that cause most diffs:
   - Improved sorting of operators in imports. [Issue 602](https://github.com/tweag/ormolu/issues/602).
   - Added support for formatting linked lists with (:) as line terminator. [Issue 478](https://github.com/tweag/ormolu/issues/478).
   - Renamed the --check-idempotency flag to --check-idempotence. *(this needed fixing in our script as well)*

PR #1180, opened 2020-07-27
Add script to rebase branch onto formatting changes (like Ormolu upgrades)

mheinzel:
Related to https://github.com/zinfra/backend-issues/issues/1623.

Upgrading Ormolu can be painful, because the style will often change in a lot of places (see for example #1145). Any active branches/PRs will then have to rebase onto these changes and resolve the resulting conflicts. It becomes a bit less cumbersome after first squashing all commits of the branch into a single one, but that's often not desirable.

This script will automate the rebasing process, keeping the (linear) history of intact and making the commits appear as if the changes had been applied onto the newly-formatted version all along.

For example, I rebased #1179 onto #1145, resulting in [this branch](https://github.com/wireapp/wire-server/compare/mheinzel/upgrade-ormolu-squashed...fisx/cleanup-rebased-onto-ormolu-0.1.2?expand=1)

You can compare the old and new commits, e.g. [this pre-#1145 change](https://github.com/wireapp/wire-server/commit/0c4d0b898b35c9c103a5d8e36c6924a4271c9d50#diff-9f592dc55a5aee3a8eb533599379916eR411-R419) with [this one post-#1145](https://github.com/wireapp/wire-server/commit/b39cc6fead76dcfc10f58b8c4a33741a35dc7f28#diff-9f592dc55a5aee3a8eb533599379916eR412-R420).

I still want to clean up the script a bit and add instructions and explanations of the idea, but as you see it already works. It was invoked using `./tools/rebase-onto-formatter.sh 178b754c3ba72c5a78745330d53d79b4fc5e34e0 mheinzel/upgrade-ormolu-squashed`, which you can also try on other branches.

### DIFF
--- a/tools/rebase-onto-formatter.sh
+++ b/tools/rebase-onto-formatter.sh
@@ -1,0 +1,44 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+echo "INSTRUCTIONS:"
+echo "1. Make a copy of your branch (or be prepared to salvage it from reflog)."
+echo "2. Find out the what the base commit is (in our case ...)"
+echo "3. Rebase onto the base commit yourself."
+echo "4. Make sure the formatting tool is installed with the correct version and settings (i.e. stack install ormolu)."
+echo "5. Run this script."
+echo ""
+echo "Running the script now. This will take a while."
+[[ "$(read -e -p 'Continue? [y/N]> '; echo $REPLY)" == [Yy]* ]] || exit 0
+
+set -x
+
+# TODO: check existence of tools
+
+# TODO: just show usage/instructions if arguments are not supplied?
+BASE_COMMIT=$1
+TARGET_COMMIT=$2
+FORMATTING_COMMAND='make formatf'
+
+# edit every commit Ci, adding f(Ci) and r(F(Ci))
+git rebase $BASE_COMMIT~1 --exec "$FORMATTING_COMMAND && git commit -am "format" && git revert HEAD --no-edit"
+
+# drop last commit
+git reset HEAD~1 --hard
+
+# now for every Ci, squash with the previous and next commit (i.e. r(f(C(i-1))) and f(Ci))
+# in sequence editor, squash lines 3, 6, 9, ... and fixup lines 4, 7, 10, ...
+# in commit message editor, drop drop first 9 lines
+GIT_SEQUENCE_EDITOR='sed -i -e "3~3s/pick/squash/" -e "4~3s/pick/fixup/"' \
+  GIT_EDITOR='sed -i "1,9d"' \
+  git rebase --interactive $BASE_COMMIT
+
+# rebase onto TARGET_COMMIT.
+# Annoyingly, we still have this first "format" commit that should already be
+# part of the TARGET_COMMIT. So we drop it.
+GIT_SEQUENCE_EDITOR='sed -i "1s/pick/drop/"' \
+  git rebase --interactive $BASE_COMMIT --onto $TARGET_COMMIT
+
+echo "Done."
+echo "Please check that the history looks as it should and all expected commits are there."
--- a/tools/rebase-onto-formatter.sh
+++ b/tools/rebase-onto-formatter.sh
@@ -20,7 +20,7 @@ TODO: explain base commit etc.
 
 INSTRUCTIONS:
 1. Make a copy of your branch (or be prepared to salvage it from reflog).
-2. Find out the what the base commit is.
+2. Find out what the base commit is.
 3. Rebase onto the base commit yourself.
 4. Make sure the formatting tool is installed with the correct version and settings (i.e. stack install ormolu).
 5. Run this script.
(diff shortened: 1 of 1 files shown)


tools/rebase-onto-formatter.sh
```diff
@@ -0,0 +1,44 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+echo "INSTRUCTIONS:"
+echo "1. Make a copy of your branch (or be prepared to salvage it from reflog)."
+echo "2. Find out the what the base commit is (in our case ...)"
```
fisx, line 7:
```suggestion
echo "2. Find out what the base commit is (in our case...); consider using git merge-base for this."
```

(not sure what the `in our case` refers to.)

tools/rebase-onto-formatter.sh
```diff
@@ -0,0 +1,62 @@
...
+
+# edit every commit Ci, adding f(Ci) and r(F(Ci))
+git rebase $BASE_COMMIT~1 --exec "$FORMATTING_COMMAND && git commit -am "format" && git revert HEAD --no-edit"
+
+# drop last commit
+git reset HEAD~1 --hard
+
+# now for every Ci, squash with the previous and next commit (i.e. r(f(C(i-1))) and f(Ci))
+# - in sequence editor, squash lines 3, 6, 9, ... and fixup lines 4, 7, 10, ...
+# - in commit message editor, drop drop first 9 lines (removing the commit message of the revert commit)
```
fisx, line 50:
```suggestion
# - in commit message editor, drop first 9 lines (removing the commit message of the revert commit)
```

tools/rebase-onto-formatter.sh
```diff
@@ -0,0 +1,62 @@
...
+
+Rebase a branch onto changes created by an automated formatter. The script
+will keep the (linear) history of the branch intact and make the commits appear
+as if the changes had been applied onto the newly-formatted version all along.
+
+TODO: explain base commit etc.
+
+INSTRUCTIONS:
+1. Make a copy of your branch (or be prepared to salvage it from reflog).
+2. Find out the what the base commit is.
```
fisx, line 23:
```suggestion
2. Find out what the base commit is.
```

tools/rebase-onto-formatter.sh
```diff
@@ -0,0 +1,62 @@
...
+
+command -v sed  >/dev/null 2>&1 || { echo >&2 "sed is not installed, aborting."; exit 1; }
+
+BASE_COMMIT=$1
+TARGET_COMMIT=$2
+FORMATTING_COMMAND='make formatf'
+USAGE="
+USAGE: $0 BASE_COMMIT TARGET_COMMIT
+    BASE_COMMIT:
+    TARGET_COMMIT:
```
fisx, line 13:
```suggestion
    BASE_COMMIT: if X is the commit that changes code formatting (and ideally nothing else), this should be X~1
    TARGET_COMMIT: X
```

So...  isn't it easier to require the formatting happens in one commit, and then just asking for that commit, which gives us both X and X-1?

mheinzel: Yes, the easiest case would be if there are two commits, the first one changing the version/settings/whatever and the second one purely applying the formatting.
But it doesn't fit with the way we currently merge code (squashing PRs), so this provides a more general solution. You can provide the doing-everything-at-once commit as TARGET_COMMIT and the only-change-the-settings commit can be passed separately as BASE_COMMIT that does not have to be in the history of `develop`. Usually we can just take it from the pre-squashing history of the PR.

It also doesn't make much of a difference for the complexity of the script.

tools/rebase-onto-formatter.sh
```diff
@@ -41,17 +41,56 @@ fi
 
 echo "Running the script now. This might take a while..."
 
-# The idea behind the sscript is ... TODO
+# The general idea is the following:
+#
+# We have a branch consisting of changes C1, C2, ... on top of our BASE_COMMIT 0.
+# Also, from the base commit, a formatting change `f(0)` was made.
```
fisx, line 47:
```suggestion
# Also, from the base commit, a formatting change `f(0)` was made on some base branch (eg., develop, or master).
```

this would have helped me to click slightly faster.

tools/rebase-onto-formatter.sh
```diff
@@ -41,17 +41,56 @@ fi
...
+# ```
+#
+# And we can also get its inverse `r(f(i))` by reverting the commit.
+#
+# So we can obtain `Ci' by composing `r(f(i-1)), `Ci` and `f(i)`.
 
 set -x
 
-# edit every commit Ci, adding f(Ci) and r(F(Ci))
+# edit every commit i, adding f(i) and r(f(i))
```
fisx, line 87: I liked `Ci` better!  :)  it's also more consistent with the diagrams above.

tools/rebase-onto-formatter.sh
mheinzel: thanks.

tools/rebase-onto-formatter.sh
mheinzel: The problem is that talking about a commit conflates two different ideas:
1. the change/diff it introduces to the previous commit
2. the version of the code at that commit (what you get when you check it out)

In the diagram, these correspond to the arrows and objects respectively. I tried to keep them separate by having the arrows `Ci` and objects `i`.

However, I think in this context this makes it unnecessarily complicated. I'm not trying to define this correctly in terms of category theory (which one could do), but documenting a Bash script. :laughing: 
So I simplified the notation a bit (while trying not to say something too incorrect). I hope it makes it more understandable.

mheinzel:
I'll merge it now, we can still tweak it later.

PR #1181, opened 2020-07-28
Allow for scim in teams that have no SAML IdP

fisx:
Fixes https://github.com/zinfra/backend-issues/issues/1365

### DIFF
--- a/libs/wire-api/src/Wire/API/User/Profile.hs
+++ b/libs/wire-api/src/Wire/API/User/Profile.hs
@@ -22,6 +22,7 @@
 module Wire.API.User.Profile
   ( Name (..),
     mkName,
+    defaultName,
     ColourId (..),
     defaultAccentId,
 
@@ -85,6 +86,9 @@ newtype Name = Name
 
 mkName :: Text -> Either String Name
 mkName txt = Name . fromRange <$> checkedEitherMsg @_ @1 @128 "Name" txt
+
+defaultName :: Name
+defaultName = Name "default"
 
 modelUserDisplayName :: Doc.Model
 modelUserDisplayName = Doc.defineModel "UserDisplayName" $ do
--- a/libs/wire-api/src/Wire/API/User/Profile.hs
+++ b/libs/wire-api/src/Wire/API/User/Profile.hs
@@ -59,6 +59,7 @@ import Control.Applicative (optional)
 import Control.Error (hush)
 import Data.Aeson hiding ((<?>))
 import qualified Data.Aeson.Types as Json
+import Data.Attoparsec.ByteString.Char8 (takeByteString)
 import Data.Attoparsec.Text
 import Data.ByteString.Conversion
 import Data.ISO3166_CountryCodes
@@ -285,6 +286,17 @@ instance FromJSON ManagedBy where
     "wire" -> pure ManagedByWire
     "scim" -> pure ManagedByScim
     other -> fail $ "Invalid ManagedBy: " ++ show other
+
+instance ToByteString ManagedBy where
+  builder ManagedByWire = "wire"
+  builder ManagedByScim = "scim"
+
+instance FromByteString ManagedBy where
+  parser =
+    takeByteString >>= \case
+      "wire" -> pure ManagedByWire
+      "scim" -> pure ManagedByScim
+      x -> fail $ "Invalid ManagedBy value: " <> show x
 
 defaultManagedBy :: ManagedBy
 defaultManagedBy = ManagedByWire
--- a/services/spar/src/Spar/Intra/Brig.hs
+++ b/services/spar/src/Spar/Intra/Brig.hs
@@ -220,11 +220,24 @@ getBrigUser buid = do
         if (userDeleted user)
           then Nothing
           else Just user
-    404 -> pure Nothing
+    404 ->
+      error
+        "look at all callers to getBrigUser and make sure they don't forget \
+        \to call getBrigInvitation and construct a scim user from that if \
+        \this fails."
+        >> pure Nothing
     _ -> throwSpar (SparBrigError "Could not retrieve user")
 
-getBrigInvitation :: (HasCallStack, MonadSparToBrig m) => InvitationId -> m (Maybe Inv.Invitation)
-getBrigInvitation = undefined
+getBrigInvitation :: (HasCallStack, MonadSparToBrig m) => TeamId -> InvitationId -> m (Maybe Inv.Invitation)
+getBrigInvitation tid invid = do
+  resp :: ResponseLBS <-
+    call $
+      method GET
+        . paths ["/i/teams/", toByteString' tid, "/invitations/", toByteString' invid]
+  case (statusCode resp, responseJsonMaybe resp) of
+    (200, Just inv) -> pure inv
+    (404, _) -> pure Nothing
+    _ -> rethrow resp
 
 -- | Get a list of users; returns a shorter list if some 'UserId's come up empty (no errors).
 --
--- a/services/spar/src/Spar/Intra/Brig.hs
+++ b/services/spar/src/Spar/Intra/Brig.hs
@@ -176,7 +176,7 @@ createBrigUserInvite ::
   ManagedBy ->
   m UserId
 createBrigUserInvite email teamid uname managedBy = do
-  let invreq = Inv.InvitationRequest Nothing Nothing (Just uname) email Nothing managedBy
+  let invreq = Inv.InvitationRequest Nothing Nothing (Just uname) Nothing email Nothing managedBy
   invresp <- call $ method POST . paths ["/i/teams", toByteString' teamid, "invitations"] . json invreq
   if statusCode invresp `notElem` [200, 201]
     then rethrow invresp
--- a/services/spar/src/Spar/Intra/Brig.hs
+++ b/services/spar/src/Spar/Intra/Brig.hs
@@ -23,7 +23,14 @@ module Spar.Intra.Brig
     fromUserSSOId,
     urefToExternalId,
     mkUserName,
+    UserOrInvitation (JustUser, JustInvitation),
+    userOrInvitationTeam,
+    userOrInvitationId,
+    userOrInvitationHandle,
+    userOrInvitationManagedBy,
+    userOrInvitationScimExternalId,
     getBrigUser,
+    getBrigActualUser,
     getBrigInvitation,
     getBrigUserTeam,
     getBrigUsers,
@@ -36,7 +43,6 @@ module Spar.Intra.Brig
     setBrigUserUserRef,
     setBrigUserRichInfo,
     checkHandleAvailable,
-    bindBrigUser,
     deleteBrigUser,
     createBrigUserSaml,
     createBrigUserInvite,
@@ -83,6 +89,41 @@ import Wire.API.User
 import Wire.API.User.RichInfo as RichInfo
 
 ----------------------------------------------------------------------
+
+data UserOrInvitation
+  = JustUser User
+  | JustInvitation Inv.Invitation
+  deriving (Eq, Show, Generic)
+
+userOrInvitationTeam :: UserOrInvitation -> Maybe TeamId
+userOrInvitationTeam = \case
+  JustUser usr -> userTeam usr
+  JustInvitation inv -> Just $ Inv.inTeam inv
+
+userOrInvitationId :: UserOrInvitation -> UserId
+userOrInvitationId = \case
+  JustUser usr -> userId usr
+  JustInvitation inv -> coerce $ Inv.inInvitation inv
+
+userOrInvitationHandle :: UserOrInvitation -> Maybe Handle
+userOrInvitationHandle = \case
+  JustUser usr -> userHandle usr
+  JustInvitation inv -> Inv.inInviteeHandle inv
+
+userOrInvitationManagedBy :: UserOrInvitation -> ManagedBy
+userOrInvitationManagedBy = \case
+  JustUser usr -> userManagedBy usr
+  JustInvitation inv -> Inv.inManagedBy inv
+
+userOrInvitationScimExternalId :: UserOrInvitation -> Either String Text
+userOrInvitationScimExternalId = \case
+  JustUser usr -> case (userIdentity >=> ssoIdentity $ usr, userEmail usr) of
+    (Just ssoid, _) -> case fromUserSSOId ssoid of
+      Right (SAML.UserRef _ subj) -> maybe (Left "bad uref from brig") Right $ SAML.shortShowNameID subj
+      Left err -> Left err
+    (Nothing, Just email) -> Right $ fromEmail email
+    (Nothing, Nothing) -> Left "brig user without external id"
+  JustInvitation inv -> Right . fromEmail . Inv.inIdentity $ inv
 
 toUserSSOId :: SAML.UserRef -> UserSSOId
 toUserSSOId (SAML.UserRef tenant subject) =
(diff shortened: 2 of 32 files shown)


fisx:
- this is still based on https://github.com/wireapp/wire-server/pull/1179; i will rebase once that is merged.
- i wanted to see how hard this would be, and i think it's not.  but there are a few (hopefully simple) open questions to discuss.

fisx:
I think this may be ready, except for the rebasing to develop (see above), and unless I find a missing test tomorrow.

libs/wire-api/src/Wire/API/User/Profile.hs
```diff
@@ -81,6 +83,15 @@ newtype Name = Name
   deriving newtype (ToJSON, FromByteString, ToByteString)
   deriving (Arbitrary) via (Ranged 1 128 Text)
 
+-- | Truncate input (or make it @" "@ if empty), and parse it into a 'Name'.
```
arianvp, line 86: Why?

services/spar/src/Spar/Intra/Brig.hs
```diff
@@ -96,6 +98,17 @@ toExternalId ssoid = do
...
   pure $ SAML.nameIDToST subj
 
+-- | Take a maybe text, a saml user ref, and construct a 'Name' from it.  If the text isn't
+-- present, use the saml subject (usually an email address).
+mkUserName :: Maybe Text -> SAML.UserRef -> Name
+mkUserName n u = fromJust $ mkUserName' n (Just u)
+
+-- | Partial variant of mkUserName.
+mkUserName' :: Maybe Text -> Maybe SAML.UserRef -> Maybe Name
+mkUserName' (Just n) _ = Just $ mkName n
```
arianvp, line 108: If this is partial anyway; why call the non-partial `mkName` that does weird things with spaces?

libs/wire-api/src/Wire/API/User/Profile.hs
fisx: What about this?:

```suggestion
-- | Truncate input (or make it @" "@ if empty), and parse it into a
-- 'Name'.  This is useful for situations in which you have an
-- arbitrary 'Text' and need a 'Name', but don't care what it is,
-- eg. during creation of users via SCIM.
```

services/spar/src/Spar/Intra/Brig.hs
fisx: Because we sometimes call 'mkName' directly, and then the totalness eliminates one more error case that everybody has to handle.

(You might argue that SCIM admins should indeed get an error if they try to set an empty user name, in which case I would argue that this is a matter of taste, and I disagree.)

services/spar/src/Spar/Scim/User.hs
```diff
@@ -336,9 +318,9 @@ createValidScimUser tokinfo vsu@(ST.ValidScimUser uref handl mbName richInfo act
 
   -- FUTUREWORK(arianvp): these two actions we probably want to make transactional
   lift . wrapMonadClient $ Data.writeScimUserTimes storedUser
-  lift . wrapMonadClient $ Data.insertSAMLUser uref buid
+  (lift . wrapMonadClient . (`Data.insertSAMLUser` buid)) `mapM_` uref
```
arianvp, line 321: Could we use `traverse_` or `for_` instead of `mapM_` ?

services/spar/src/Spar/Scim/User.hs
```diff
@@ -384,21 +357,23 @@ updateValidScimUser tokinfo uid newScimUser = do
...
+      mOldUref <- do
         let extid :: Maybe Text
             extid = Scim.externalId . Scim.value . Scim.thing $ oldScimStoredUser
         idp <- tokenInfoToIdP tokinfo
         mkUserRef idp extid
-      when (olduref /= newuref) $ do
-        lift . wrapMonadClient $ Data.deleteSAMLUser olduref
-        lift . wrapMonadClient $ Data.insertSAMLUser newuref uid
+      when (mOldUref /= mNewUref) $ do
+        forM_ mOldUref $ lift . wrapMonadClient . Data.deleteSAMLUser
```
arianvp, line 367: can we use `for_` instead of `forM_` ?

services/spar/src/Spar/Intra/Brig.hs
arianvp: Silently setting names to ` ` sounds like a bad idea to me.  It's another weird convention we introduce. 

AFAIK the consensus is that the string `"default"` is what's used for 'empty names' across the platform. Perhaps we should use that here too?

services/spar/src/Spar/Scim/User.hs
fisx: why?  but sure:

```suggestion
        for_ mOldUref $ lift . wrapMonadClient . Data.deleteSAMLUser
```

services/spar/src/Spar/Scim/User.hs
fisx:
```suggestion
  for_ uref $ (lift . wrapMonadClient . (`Data.insertSAMLUser` buid))
```

services/spar/src/Spar/Intra/Brig.hs
arianvp: The only place where I see it used is in auto-provisioning saml users.   And there the old behaviuour was to set the users' name to their email address. We now break that behaviour. no? that seems bad.   This means people using SAML auto-provisioning will not see their name in the UI?

fisx: good idea: 7530c9601206c0937ff229583c85cc399cfbe1b3

arianvp: Ah wait I can't read..  the line below sets the name to `uidSubject`

arianvp: i dont see `mkName` called anywhere else in this PR though. So I'm still confused

services/spar/src/Spar/Scim/User.hs
```diff
@@ -298,29 +289,20 @@ createValidScimUser ::
...
-    let inidp = idpConfig ^. SAML.idpMetadata . SAML.edIssuer
-        inuref = uref ^. SAML.uidTenant
-     in assert (inidp == inuref) $ pure ()
+createValidScimUser ScimTokenInfo {stiTeam} vsu@(ST.ValidScimUser uref handl mbName richInfo active) = do
   -- Generate a UserId will be used both for scim user in spar and for brig.
   buid <- Id <$> liftIO UUID.nextRandom
   -- ensure uniqueness constraints of all affected identifiers.
   -- if we crash now, retry POST will just work
-  assertUserRefUnused uref
+  assertUserRefUnused `mapM_` uref
```
arianvp, line 297:
```suggestion
  forM_ uref assertUserRefUnused
```

services/spar/src/Spar/Scim/User.hs
```diff
@@ -336,9 +318,9 @@ createValidScimUser tokinfo vsu@(ST.ValidScimUser uref handl mbName richInfo act
 
   -- FUTUREWORK(arianvp): these two actions we probably want to make transactional
   lift . wrapMonadClient $ Data.writeScimUserTimes storedUser
-  lift . wrapMonadClient $ Data.insertSAMLUser uref buid
+  for_ uref $ (lift . wrapMonadClient . (`Data.insertSAMLUser` buid))
 
-  lift $ validateEmailIfExists buid uref
+  (lift . validateEmailIfExists buid) `mapM_` uref
```
arianvp, line 323:
```suggestion
  for_ uref (lift . validateEmailIfExists buid)
```

services/spar/src/Spar/Scim/User.hs
```diff
@@ -384,21 +357,23 @@ updateValidScimUser tokinfo uid newScimUser = do
...
         let extid :: Maybe Text
             extid = Scim.externalId . Scim.value . Scim.thing $ oldScimStoredUser
         idp <- tokenInfoToIdP tokinfo
         mkUserRef idp extid
-      when (olduref /= newuref) $ do
-        lift . wrapMonadClient $ Data.deleteSAMLUser olduref
-        lift . wrapMonadClient $ Data.insertSAMLUser newuref uid
+      when (mOldUref /= mNewUref) $ do
+        for_ mOldUref $ lift . wrapMonadClient . Data.deleteSAMLUser
+        forM_ mNewUref $ \newUref -> lift . wrapMonadClient $ Data.insertSAMLUser newUref uid
```
arianvp, line 368:
```suggestion
        for_ mNewUref $ \newUref -> lift . wrapMonadClient $ Data.insertSAMLUser newUref uid
```

services/spar/src/Spar/Scim/User.hs
```diff
@@ -384,21 +357,23 @@ updateValidScimUser tokinfo uid newScimUser = do
...
       -- update 'SAML.UserRef' on brig
-      bindok <- lift $ Brig.bindBrigUser uid newuref
-      unless bindok . throwError $
-        Scim.serverError "Failed to update SAML UserRef on brig."
-      -- this can only happen if user is found in spar.scim_user, but missing on brig.
-      -- (internal error?  race condition?)
+      forM_ mNewUref $ \newUref -> do
```
arianvp, line 370:
```suggestion
      for_ mNewUref $ \newUref -> do
```

services/spar/src/Spar/Scim/User.hs
```diff
@@ -359,23 +341,14 @@ updateValidScimUser ::
...
-
   -- TODO: how do we get this safe w.r.t. race conditions / crashes?
 
   -- construct old and new user values with metadata.
   oldScimStoredUser :: Scim.StoredUser ST.SparTag <-
     Scim.getUser tokinfo uid
   oldValidScimUser :: ST.ValidScimUser <-
     validateScimUser tokinfo . Scim.value . Scim.thing $ oldScimStoredUser
-  assertUserRefNotUsedElsewhere (newScimUser ^. ST.vsuUserRef) uid
+  (`assertUserRefNotUsedElsewhere` uid) `mapM_` (newScimUser ^. ST.vsuUserRef)
```
arianvp, line 351:
```suggestion
  (`assertUserRefNotUsedElsewhere` uid) `traverse_` (newScimUser ^. ST.vsuUserRef)
```

services/spar/src/Spar/Intra/Brig.hs
fisx: 8d940cc9caff77c31b52d8bb2f6d6428d90aa383

fisx:
I'm not sure how to create a user that is created with a team invite.  I see two options:

1. fake the invitation procedure without telling the user, and letting the user create a password by "forgot password".
2. initiate the usual invitation procedure.

2 seems nicer and more natural / organic / coherent, but it would mean that if azure pulls the user between the time the user was posted and the time the invitation was accepted, it will see an inactive user, and may wonder why.  may not be a problem, not sure.

@tiago-loureiro @arianvp any opinions?  context: 1b3fc92 and

> [spar] E, request=N/A, Error {code = Status {statusCode = 500, statusMessage = "Internal Server Error"}, label = "server-error", message = "{\"status\":\"500\",\"schemas\":[\"urn:ietf:params:scim:api:messages:2.0:Error\"],\"detail\":\"{\\\"code\\\":400,\\\"message\\\":\\\"Error in $: team_code, team, invitation_code, sso_id are mutually exclusive and sso_id, team_id must be either both present or both absent.\\\",\\\"label\\\":\\\"bad-request\\\"}\"}"}

arianvp:
I assume we have this code in either brig or galley already no?  for creating team members without SSO ?

tiago-loureiro:
> I assume we have this code in either brig or galley already no? for creating team members without SSO ?

I think not because there isn't such a flow; so we protect against that exactly. 

> I'm not sure how to create a user that is created with a team invite. I see two options:

2. seems more elegant to me. Also, once we _do_ set passwords over SCIM (I remember problems being brought up with using it, I think we discussed this elsewhere already) then the problem you mentioned will disappear.

> it will see an inactive user, and may wonder why. may not be a problem, not sure.

Good point. My guess is that it may try to update that attribute/(re)create the user? I'd default to this behaviour (let's sync with product on this decision) and re-visit it if need be.

arianvp:
> seems more elegant to me. Also, once we do set passwords over SCIM (I remember problems being brought up with using it, I think we discussed this elsewhere already) then the problem you mentioned will disappear.

The problem was that we would be storing the password in plain text in the `scim.users` table.

However that problem now goes away.     @fisx do you think it makes sense to support setting password in SCIM in this PR? If this makes your life easier. It also immediately solves a use-case for some customers IIRC

fisx:
I think we need to clarify the requirements.

So what if we fail with a clean error, something like "not supported", and merge this PR and leave the issue open with label "requirements unclear" until we know how we want to create non-SAML users?

tiago-loureiro:
> I think we need to clarify the requirements.
> 
> So what if we fail with a clean error, something like "not supported", and merge this PR and leave the issue open with label "requirements unclear" until we know how we want to create non-SAML users?

So this PR "prepares" the way but does not actually add/change any functionality?

fisx:
> > I think we need to clarify the requirements.
> > So what if we fail with a clean error, something like "not supported", and merge this PR and leave the issue open with label "requirements unclear" until we know how we want to create non-SAML users?
> 
> So this PR "prepares" the way but does not actually add/change any functionality?

Yes.

Not so sure this is a good idea, though.  I just realized this approach will result in untestable production code.

fisx:
summerizing call with @tiago-loureiro - let's take one step back:

> > I assume we have this code in either brig or galley already no? for creating team members without SSO ?
> 
> I think not because there isn't such a flow; so we protect against that exactly.
> 
> > I'm not sure how to create a user that is created with a team invite. I see two options:
> 
>     1. seems more elegant to me. Also, once we _do_ set passwords over SCIM (I remember problems being brought up with using it, I think we discussed this elsewhere already) then the problem you mentioned will disappear.
> 
> 
> > it will see an inactive user, and may wonder why. may not be a problem, not sure.
> 
> Good point. My guess is that it may try to update that attribute/(re)create the user? I'd default to this behaviour (let's sync with product on this decision) and re-visit it if need be.

let's try to do this, ie. treat "create scim user without saml idp" as "create a team invite and send out an email", and try to respond with a scim user that is constructed from the invitation rather than the user.  it'll be inactive, and not contain a whole lot of data, but we may be able to make it consistent.

testing this with azure will be a second step.  but exxon would be fine with this now working, since they already have the work-aorund with a dummy saml idp in place.

fisx:
6 more integration tests failing:

- line 1081 is probably because invitations cannot be found by email search like active users.  that's intended behavior, we just need to figure out what to do with the test.
- line 1081 is *also* about emails not being used as emails.  perhaps we need to set the team flag for email activation?
- i think there are also a few tests that fail because we need to construct scim users with valid emails as external ids in those tests.

services/spar/src/Spar/Intra/Brig.hs
```diff
@@ -84,6 +84,12 @@ import Wire.API.User.RichInfo as RichInfo
 
 ----------------------------------------------------------------------
 
+newtype UserOrInvitation = UserOrInvitation {fromUserOrInvitation :: Either Inv.Invitation User}
```
arianvp, line 87: Why not just use a sumtype?

fisx: good idea, will do.

fisx:
**Q** technical question about invitation lookup.  we currently allow to lookup invitations by phone number or email, which are both encoded into plain texts in the same column, in the hope that the two string sets are disjoint.  i now need to lookup handles, which aren't disjoint from email addresses.  how should i impelment `lookupInvitationByHandle`?

**A** just add a new table `team_invitation_by_handle`.

tiago-loureiro:
> **Q** technical question about invitation lookup. we currently allow to lookup invitations by phone number or email, which are both encoded into plain texts in the same column, in the hope that the two string sets are disjoint. i now need to lookup handles, which aren't disjoint from email addresses. how should i impelment `lookupInvitationByHandle`?
> 
> **A** just add a new table `team_invitation_by_handle`.

They _are_ disjoint from others as well, see the comment at: https://github.com/wireapp/wire-server/blob/develop/libs/types-common/src/Data/Handle.hs#L76-L83

services/spar/src/Spar/Scim/User.hs
```diff
@@ -627,25 +629,63 @@ synthesizeStoredUserU usr = do
...
       richInfo
-      accStatus
+      (Just accStatus)
       createdAt
       lastUpdatedAt
       baseuri
   lift $ writeState tid (BT.userId usr) accessTimes (BT.userManagedBy usr) storedUser
   pure storedUser
 
+synthesizeStoredUserI :: Inv.Invitation -> Scim.ScimHandler Spar (Scim.StoredUser ST.SparTag)
```
arianvp: dont be shy of characters.    `s/I/FromInvitation`

fisx: if you like 1958a8c, this comment doesn't apply any more.  let me know if you don't, then i'll revert it and give this the longer name.

fisx:
this is the wrong approach: invitations are not intended for representing inactive users, and forcing this breaks the cassandra data model (adding lots of race conditions and inefficiencies).  i'm going to try again in (hopefully much smaller) separate PRs, starting with #1195.

PR #1183, opened 2020-07-31
Merge brig:lib and brig-index-lib

akshaymankar:
This unblocks zoogle and avoids other weird issues with stack not handling
internal libraries properly.


services/brig/brig.cabal
```diff
@@ -1,10 +1,10 @@
-cabal-version: 2.0
+cabal-version: 1.12
```
mheinzel, line 1: Why does that happen? Your hpack version seems to be the same, so why does it generate the file for a different Cabal version?

akshaymankar: hpack figures out that without internal libraries, this cabal file will even work with older `cabal-version`, so it updates that.

PR #1184, opened 2020-07-31
Ensure `sft_servers` is never set to `null` in /calls/config

akshaymankar:
Fixes https://github.com/zinfra/backend-issues/issues/1453

### DIFF
--- a/libs/wire-api/package.yaml
+++ b/libs/wire-api/package.yaml
@@ -70,3 +70,4 @@ tests:
     - tasty-expected-failure
     - tasty-hunit
     - tasty-quickcheck
+    - unordered-containers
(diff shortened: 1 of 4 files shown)


fisx:
(did you link the right issue, though?  perhaps, just to attach it to the right place in the scim board?)

akshaymankar:
> (did you link the right issue, though? perhaps, just to attach it to the right place in the scim board?)

SCIM? Do you mean sprint? The issue is still the same, the changes #1177 produce `sft_servers: null` whenever there are none to be shown. We haven't shipped zinfra/backend-issues#1453, so I guess we don't need another issue.

fisx:
> > (did you link the right issue, though? perhaps, just to attach it to the right place in the scim board?)
> 
> SCIM? Do you mean sprint? The issue is still the same, the changes #1177 produce `sft_servers: null` whenever there are none to be shown. We haven't shipped [zinfra/backend-issues#1453](https://github.com/zinfra/backend-issues/issues/1453), so I guess we don't need another issue.

scim is the best from both scrum and sprint.  :)

PR #1189, opened 2020-08-04
Add endpoint to Federator API, cleanup

mheinzel:
part 3 of https://github.com/zinfra/backend-issues/issues/1213#issuecomment-659237907


services/federator/src/Federator/Util.hs
```diff
@@ -21,19 +21,8 @@ module Federator.Util
...
-        if pfx' /= pfx
-          then Nothing
-          else Just sfx
-
--- | This is a partial function; totality of all calls must be verified by roundtrip tests on
--- the aeson instances involved.
-wireJsonOptions :: String -> Options
-wireJsonOptions pfx = defaultOptions {fieldLabelModifier = fromJust . dropPrefix pfx . fmap toLower}
+wireJsonOptions :: Options
+wireJsonOptions = defaultOptions {fieldLabelModifier = camelTo2 '_' . dropWhile (not . Char.isUpper)}
```
jschaul, line 28: This is the same as the one in wire-api-federation. Re-use that definition?

services/federator/src/Federator/API.hs
```diff
@@ -57,13 +76,15 @@ data API route = API
...
 
+-- TODO: rename
 data FUser = FUser
   { _fuGlobalHandle :: !(Qualified Handle),
     _fuFQU :: !(Qualified UserId)
   }
   deriving (Eq, Show, Generic)
 
-deriveJSON (wireJsonOptions "_fu") ''FUser
+-- TODO: use Generics
```
jschaul, line 86: TODO comments that won't be tackled in the PR they are introduced - don't we usually call those FUTUREWORK? Maybe rename?

mheinzel: Yes, you're right about the conventions. I'd still like to fix it before merging this PR.

services/federator/src/Federator/Util.hs
mheinzel: Yeah, I will do that.

services/federator/src/Federator/API.hs
```diff
@@ -45,7 +52,18 @@ data API route = API
         :> "users"
         :> Capture "fqu" (Qualified UserId)
         :> "prekeys"
-        :> Get '[JSON] PrekeyBundle
+        -- FUTUREWORK(federation):
+        -- this should return a version of PrekeyBundle with qualified UserId,
+        -- defined in wire-api-federation
+        :> Get '[JSON] PrekeyBundle,
+    _gapiJoinConversationById ::
```
arianvp, line 59: I'm not sure yet if the Federator needs to be _typed_ . in the sense that it's just a broker that has to decide where to fan out to right?  

The federator does very little at the moment. Do we actually need it?

arianvp: But lets worry about this after we have the entire feature implemented end to end

services/federator/src/Federator/API.hs
mheinzel: All resolved now.

mheinzel:
Integration tests in CI are not starting up, but it's all passing locally. Also, this PR only makes changes to the federator, which is not integration tested at all yet.

Therefore, I'll merge now.

PR #1190, opened 2020-08-04
[experiment] strongly typed events with polysemy

fisx:
I don't expect this PR to go anywhere, just wanted to learn something.  Issue with findings coming up soon!

### DIFF
--- a/services/brig/src/Brig/IO/Intra.hs
+++ b/services/brig/src/Brig/IO/Intra.hs
@@ -1,4 +1,5 @@
 {-# LANGUAGE RecordWildCards #-}
+{-# OPTIONS_GHC -Wno-unused-imports #-}
 
 -- This file is part of the Wire Server implementation.
 --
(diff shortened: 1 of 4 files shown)


services/brig/src/Brig/IO/Intra.hs
```diff
@@ -94,17 +58,44 @@ import Imports
...
+
 -----------------------------------------------------------------------------
 -- Event Handlers
 
-onUserEvent :: UserId -> Maybe ConnId -> UserEvent -> AppIO ()
-onUserEvent orig conn e =
+onUserEvent :: IsUserEvent e => UserId -> Maybe ConnId -> e -> AppIO ()
+onUserEvent orig conn e = do
+  -- I am tempted to try and use 'Chan' to simply drop events, and then let updateSearchIndex,
+  -- dispatchNotifications, and journalEvent all subscribe to duplicates of that 'Chan'.  This
```
fisx, line 93: or maybe http://hackage.haskell.org/package/broadcast-chan

services/brig/src/Brig/IO/Intra.hs
```diff
@@ -154,73 +145,32 @@ onClientEvent orig conn e = do
...
-    Journal.userUpdate orig Nothing (Just loc) Nothing
-  UserIdentityUpdated (UserIdentityUpdatedData _ (Just em) _) ->
-    Journal.userUpdate orig (Just em) Nothing Nothing
-  UserIdentityRemoved (UserIdentityRemovedData _ (Just em) _) ->
-    Journal.userEmailRemove orig em
-  UserDeleted {} ->
-    Journal.userDelete orig
-  _ ->
-    return ()
+-- (this function shows that things get a lot simpler and type info increases, but i think we
```
mheinzel, line 148: > this function shows that things get a lot simpler and type info increases

not sure what exactly you mean. What got simpler and where is the additional type info?

arianvp:
I don't think this gains us a lot. We still have to manually trigger the events in the code. which is ad hoc.

My problem with our events code is not that we dont track it in the type, it's that we manually have to call it all the time; and if you forget you get bugs. Instead event emissions should be a pattern that we abstract.  They should reflect exactly the operations the database is doing. The only thing that is different is that not all state is synced to all users; and that's the only thing we should differentiate on.

arianvp:
Instead of describing in the type what events we can fire by calling callbacks as _input_

functions should return what events they emit as _output_. This seems a way more natural way to express the problem

fisx:
> Instead of describing in the type what events we can fire by calling callbacks as _input_
> 
> functions should return what events they emit as _output_. This seems a way more natural way to express the problem

I don't think we're looking at the same code, because I agree with you, and I think that's what I've outlined in this PR.  :)  I'll try to get on the phone with you about this on Friday.

services/brig/src/Brig/IO/Intra.hs
```diff
@@ -94,17 +58,44 @@ import Imports
 import Network.HTTP.Types.Method
 import Network.HTTP.Types.Status
 import qualified Network.Wai.Utilities.Error as Wai
+import Polysemy
 import System.Logger.Class as Log hiding (name, (.=))
 import Wire.API.Team.Feature (TeamFeatureName (..), TeamFeatureStatus)
 
+----------------------------------------------------------------------
+-- polysemi
```
akshaymankar, line 66:
```suggestion
-- polysemy
```

fisx:
this is not going to get merged, but i've linked it here [here](https://github.com/zinfra/backend-issues/issues/1846).

PR #1195, opened 2020-08-26
spar cleanup (1)

fisx:
Some more or less trivial code simplifications.

This is the first step in taking apart the completely unhinged https://github.com/wireapp/wire-server/pull/1181.

### DIFF
--- a/libs/types-common/src/Data/Range.hs
+++ b/libs/types-common/src/Data/Range.hs
@@ -89,6 +89,10 @@ newtype Range (n :: Nat) (m :: Nat) a = Range
   }
   deriving (Eq, Ord, Show)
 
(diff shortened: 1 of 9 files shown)


fisx:
All tests pass locally.

akshaymankar (approved):
Looks good!
Since this PR does a bunch of different things, can we please not squash it? A rebase merge would be better. If we are doing that, it would be nice if we could write the last two commit messages to better explain the contents.

fisx:
@akshaymankar as discussed, I'll rebase-commit this once, but I'll keep the commit messages.

In general, we agreed to stick to squashing PRs in general, and just make the PRs smaller and more self-contained than this one.

PR #1196, opened 2020-08-27
Refactor Invitations

fisx:
tl;dr:

- rename stuff that doesn't make sense and/or has no effects on behavior.
- move stuff to make upcoming PRs smaller.

slightly more details:

- drop irName, which carried the inviter name, but was never used.
- fix test suite where we used irName wrong, and then ignored it.
- give irEmail, irPhone, inIdentity, inPhone less ambiguous names.
- shuffle record fields.
- swagger.
- move around validateHandle and some other helpers.


fisx:
all integration tests pass locally.

arianvp (approved):

PR #1197, opened 2020-08-27
Fixed logic related to ephemeral users

tiago-loureiro:
Fixes https://github.com/zinfra/backend-issues/issues/1685

Long story short: the logic that checked whether a user was ephemeral was flawed, causing user creation to fail when it should succeed. This PR fixes that.

### DIFF
--- a/docs/reference/user/registration.md
+++ b/docs/reference/user/registration.md
@@ -185,7 +185,8 @@ During registration, we can take advantage of [NewUserOrigin](https://github.com
 So `/register` should only succeed iff at least one of these conditions is true:
 
 ```
-NewUserIsTeamUser OR newUserIsEphemeral
+import Brig.Types.User
+isNewUserTeamMember || isNewUserEphemeral
 ```
 
 The rest of the unauthorized end-points is safe:
--- a/docs/reference/user/registration.md
+++ b/docs/reference/user/registration.md
@@ -180,14 +180,12 @@ We need an option to block 1, 2, 5 on-prem; 3, 4 should remain available (no blo
  * Allow team members to register (via email/phone or SSO)
  * Allow ephemeral users
 
-During registration, we can take advantage of [NewUserOrigin](https://github.com/wireapp/wire-server/blob/a89b9cd818997e7837e5d0938ecfd90cf8dd9e52/libs/wire-api/src/Wire/API/User.hs#L625); we're particularly interested in `NewUserOriginTeamUser` --> only `NewTeamMember` or `NewTeamMemberSSO` should be accepted. In case this is a `Nothing`, we need to check if the user expires, i.e., `newUserExpiresIn` must be a `Just`.
+During registration, we can take advantage of [NewUserOrigin](https://github.com/wireapp/wire-server/blob/a89b9cd818997e7837e5d0938ecfd90cf8dd9e52/libs/wire-api/src/Wire/API/User.hs#L625); we're particularly interested in `NewUserOriginTeamUser` --> only `NewTeamMember` or `NewTeamMemberSSO` should be accepted. In case this is a `Nothing`, we need to check if the user expires, i.e., if the user has no identity (and thus `Ephemeral`).
 
 So `/register` should only succeed iff at least one of these conditions is true:
 
 ```
-newUserTeam == (Just (NewTeamMember _)) OR
-newUserTeam == (Just (NewTeamMemberSSO _)) OR
-newUserExpiresIn == (Just _)
+NewUserIsTeamUser OR newUserIsEphemeral
 ```
 
 The rest of the unauthorized end-points is safe:
--- a/libs/wire-api/src/Wire/API/User.hs
+++ b/libs/wire-api/src/Wire/API/User.hs
@@ -43,6 +43,8 @@ module Wire.API.User
     newUserEmail,
     newUserPhone,
     newUserSSOId,
+    isNewUserEphemeral,
+    isNewUserTeamMember,
 
     -- * NewUserOrigin
     NewUserOrigin (..),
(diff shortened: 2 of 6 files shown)


libs/wire-api/src/Wire/API/User.hs
```diff
@@ -492,6 +494,19 @@ validateNewUserPublic nu
...
+isNewUserEphemeral :: NewUser -> Bool
+isNewUserEphemeral u = case newUserIdentity u of
+  Nothing -> True
+  Just _ -> False
+
+isNewUserTeamMember :: NewUser -> Bool
+isNewUserTeamMember u = case newUserTeam u of
+  Just (NewTeamMember _) -> True
+  Just (NewTeamMemberSSO _) -> True
+  _ -> False
```
tiago-loureiro, line 508: Opinion poll: should we be explicit about all cases to avoid having false assumptions in case we add more constructors to `NewTeamUser`?

fisx: In my head we already have that rule, but I suppose we should write it down somewhere...  :)

docs/reference/user/registration.md
```diff
@@ -180,14 +180,12 @@ We need an option to block 1, 2, 5 on-prem; 3, 4 should remain available (no blo
...
-During registration, we can take advantage of [NewUserOrigin](https://github.com/wireapp/wire-server/blob/a89b9cd818997e7837e5d0938ecfd90cf8dd9e52/libs/wire-api/src/Wire/API/User.hs#L625); we're particularly interested in `NewUserOriginTeamUser` --> only `NewTeamMember` or `NewTeamMemberSSO` should be accepted. In case this is a `Nothing`, we need to check if the user expires, i.e., `newUserExpiresIn` must be a `Just`.
+During registration, we can take advantage of [NewUserOrigin](https://github.com/wireapp/wire-server/blob/a89b9cd818997e7837e5d0938ecfd90cf8dd9e52/libs/wire-api/src/Wire/API/User.hs#L625); we're particularly interested in `NewUserOriginTeamUser` --> only `NewTeamMember` or `NewTeamMemberSSO` should be accepted. In case this is a `Nothing`, we need to check if the user expires, i.e., if the user has no identity (and thus `Ephemeral`).
 
 So `/register` should only succeed iff at least one of these conditions is true:
 
 ```
-newUserTeam == (Just (NewTeamMember _)) OR
-newUserTeam == (Just (NewTeamMemberSSO _)) OR
-newUserExpiresIn == (Just _)
+NewUserIsTeamUser OR newUserIsEphemeral
```
fisx, line 188: What does this mean?  Maybe you mean these?:

```suggestion
import Brig.Types.User
isNewUserTeamMember || isNewUserEphemeral
```

libs/wire-api/src/Wire/API/User.hs
```diff
@@ -492,6 +494,19 @@ validateNewUserPublic nu
   | otherwise =
     Right (NewUserPublic nu)
 
+-- | Any user registering without either an email or a phone is Ephemeral,
+-- i.e. can be deleted after expires_in or sessionTokenTimeout
+isNewUserEphemeral :: NewUser -> Bool
+isNewUserEphemeral u = case newUserIdentity u of
+  Nothing -> True
+  Just _ -> False
```
fisx, line 502:
```suggestion
isNewUserEphemeral = isNothing . newUserIdentity
```

fisx (approved):
Maybe you can replace the catchall pattern as mentioned there, otherwise looks good.

This is a change in requirements, right?  Before, we intended ephemeral users to have a life expectancy, now we don't want that any more?

tiago-loureiro:
> This is a change in requirements, right? Before, we intended ephemeral users to have a life expectancy, now we don't want that any more?

No, no change in requirements; such users (without identity) by default have an expiry time which is our [session timeout](https://github.com/zinfra/hegemony/blob/develop/roles/brig/config/all/brig.yaml.j2#L90) - the `expires_in` can also be set, in addition, with a value [in the range of [1,604800] seconds](https://github.com/wireapp/wire-server/blob/develop/libs/wire-api/src/Wire/API/User.hs#L538)

docs/reference/user/registration.md
tiago-loureiro: > What does this mean? Maybe you mean these?:

yes, thanks for the suggestion :)

fisx:
> > This is a change in requirements, right? Before, we intended ephemeral users to have a life expectancy, now we don't want that any more?
> 
> No, no change in requirements; such users (without identity) by default have an expiry time which is our [session timeout](https://github.com/zinfra/hegemony/blob/develop/roles/brig/config/all/brig.yaml.j2#L90) - the `expires_in` can also be set, in addition, with a value [in the range of [1,604800] seconds](https://github.com/wireapp/wire-server/blob/develop/libs/wire-api/src/Wire/API/User.hs#L538)

hum, is this documented?  if not, do you want to add this comment to the docs file you're already touching here.  it's easy to get confused about this.

tiago-loureiro:
> > > This is a change in requirements, right? Before, we intended ephemeral users to have a life expectancy, now we don't want that any more?
> > 
> > 
> > No, no change in requirements; such users (without identity) by default have an expiry time which is our [session timeout](https://github.com/zinfra/hegemony/blob/develop/roles/brig/config/all/brig.yaml.j2#L90) - the `expires_in` can also be set, in addition, with a value [in the range of [1,604800] seconds](https://github.com/wireapp/wire-server/blob/develop/libs/wire-api/src/Wire/API/User.hs#L538)
> 
> hum, is this documented? if not, do you want to add this comment to the docs file you're already touching here. it's easy to get confused about this.

It's there in the comment of `isNewUserEphemeral` -> do you suggest adding it elsewhere?

fisx:
> > > > This is a change in requirements, right? Before, we intended ephemeral users to have a life expectancy, now we don't want that any more?
> > > 
> > > 
> > > No, no change in requirements; such users (without identity) by default have an expiry time which is our [session timeout](https://github.com/zinfra/hegemony/blob/develop/roles/brig/config/all/brig.yaml.j2#L90) - the `expires_in` can also be set, in addition, with a value [in the range of [1,604800] seconds](https://github.com/wireapp/wire-server/blob/develop/libs/wire-api/src/Wire/API/User.hs#L538)
> > 
> > 
> > hum, is this documented? if not, do you want to add this comment to the docs file you're already touching here. it's easy to get confused about this.
> 
> It's there in the comment of `isNewUserEphemeral` -> do you suggest adding it elsewhere?

ah, no, that's fine.  just need to get used to reading docs.  :)

fisx:
integration and unit tests pass locally.

PR #1200, opened 2020-09-02
Allow SCIM without saml

fisx:
Fixes https://github.com/zinfra/backend-issues/issues/1365, https://github.com/zinfra/backend-issues/issues/1572

### documentation

Before this PR, scim tokens could only be added to teams that already had exactly one SAML IdP.  Now, we also allow SAML-less teams to have SCIM provisioning.  This is an alternative to onboarding via team-settings and produces user accounts that are authenticated with email and password.  (Phone may or may not work, but is not officially supported.)

The way this works is different from team-settings: we don't send invites, but we create active users immediately the moment the SCIM user post is processed.  The new thing is that the created user has neither email nor phone nor a SAML identity, nor a password.

How does this work?

**email:** If no SAML IdP is present, SCIM user posts must contain an externalId that is an email address.  This email address is not added to the newly created user, because it has not been validated.  Instead, the flow for changing an email address is triggered in brig: an email is sent to the address containing a validation key, and once the user completes the flow, brig will add the email address to the user.  We had to add very little code for this in this PR, it's all an old feature.

When SCIM user gets are processed, in order to reconstruct the externalId from the user spar is retrieving from brig, we introduce a new json object for the `sso_id` field that looks like this: `{'scim_external_id': 'me@example.com'}`.

In order to find users that have email addresses pending validation, we introduce a new table in spar's cassandra called `scim_external_ids`, in analogy to `user`.  We have tried to use brig's internal `GET /i/user&email=...`, but that also finds pending email addresses, and there are corner cases when changing email addresses and waiting for the new address to be validated and the old to be removed...  that made this approach seem infeasible.

**password:** once the user has validated their email address, they need to trigger the "forgot password" flow -- also old code.

That's the gist of it!

### DIFF
--- a/docs/reference/spar-braindump.md
+++ b/docs/reference/spar-braindump.md
@@ -21,6 +21,25 @@ documentation answering your questions, look here!
 - if you want to work on our saml/scim implementation and do not have access to [https://github.com/zinfra/backend-issues/issues?q=is%3Aissue+is%3Aopen+label%3Aspar] and [https://github.com/wireapp/design-specs/tree/master/Single%20Sign%20On], please get in touch with us.
 
 
+## design considerations
+
+### SCIM without SAML.
+
+Before https://github.com/wireapp/wire-server/pull/1200, scim tokens could only be added to teams that already had exactly one SAML IdP.  Now, we also allow SAML-less teams to have SCIM provisioning.  This is an alternative to onboarding via team-settings and produces user accounts that are authenticated with email and password.  (Phone may or may not work, but is not officially supported.)
+
+The way this works is different from team-settings: we don't send invites, but we create active users immediately the moment the SCIM user post is processed.  The new thing is that the created user has neither email nor phone nor a SAML identity, nor a password.
+
+How does this work?
+
+**email:** If no SAML IdP is present, SCIM user posts must contain an externalId that is an email address.  This email address is not added to the newly created user, because it has not been validated.  Instead, the flow for changing an email address is triggered in brig: an email is sent to the address containing a validation key, and once the user completes the flow, brig will add the email address to the user.  We had to add very little code for this in this PR, it's all an old feature.
+
+When SCIM user gets are processed, in order to reconstruct the externalId from the user spar is retrieving from brig, we introduce a new json object for the `sso_id` field that looks like this: `{'scim_external_id': 'me@example.com'}`.
+
+In order to find users that have email addresses pending validation, we introduce a new table in spar's cassandra called `scim_external_ids`, in analogy to `user`.  We have tried to use brig's internal `GET /i/user&email=...`, but that also finds pending email addresses, and there are corner cases when changing email addresses and waiting for the new address to be validated and the old to be removed...  that made this approach seem infeasible.
+
+**password:** once the user has validated their email address, they need to trigger the "forgot password" flow -- also old code.
+
+
 ## operations
 
 ### enabling / disabling the sso feature for a team
--- a/services/spar/src/Spar/Intra/Brig.hs
+++ b/services/spar/src/Spar/Intra/Brig.hs
@@ -290,7 +290,11 @@ getBrigUserByEmail email = do
         . path "/i/users"
         . queryItem "email" (toByteString' email)
   case statusCode resp of
-    200 -> listToMaybe <$> parseResponse @[UserAccount] resp
+    200 -> do
+      macc <- listToMaybe <$> parseResponse @[UserAccount] resp
+      case userEmail . accountUser =<< macc of
+        Just email' | email' == email -> pure macc
+        _ -> pure Nothing
     404 -> pure Nothing
     _ -> rethrow resp
 
--- a/services/spar/src/Spar/Intra/Brig.hs
+++ b/services/spar/src/Spar/Intra/Brig.hs
@@ -25,12 +25,13 @@ module Spar.Intra.Brig
     veidFromUserSSOId,
     urefToExternalId,
     urefToEmail,
-    userToScimExternalId,
+    userToExternalId,
     veidFromBrigUser,
     mkUserName,
     renderValidExternalId,
     emailFromSAML,
     emailToSAML,
+    emailToSAMLNameID,
     getBrigUser,
     getBrigUserAccount,
     getBrigUserTeam,
@@ -132,8 +133,8 @@ urefToEmail uref = case uref ^. SAML.uidSubject . SAML.nameID of
   SAML.UNameIDEmail email -> Just $ emailFromSAML email
   _ -> Nothing
 
-userToScimExternalId :: User -> Either String Text
-userToScimExternalId usr = case (userSSOId usr, userEmail usr) of
+userToExternalId :: User -> Either String Text
+userToExternalId usr = case (userSSOId usr, userEmail usr) of
   (Just ssoid, _) -> case fromUserSSOId ssoid of
     Right (SAML.UserRef _ subj) -> maybe (Left "bad uref from brig") Right $ SAML.shortShowNameID subj
     Left err -> Left err
@@ -271,7 +272,7 @@ getBrigUserAccount buid = do
             if userDeleted $ accountUser account
               then Nothing
               else Just account
-        _ -> pure Nothing -- impossible
+        _ -> pure Nothing
     404 -> pure Nothing
     _ -> rethrow resp
 
@@ -410,7 +411,7 @@ getBrigUserRichInfo buid =
           . paths ["/i/users", toByteString' buid, "rich-info"]
     case statusCode resp of
       200 -> parseResponse resp
-      _ -> throwSpar (SparBrigErrorWith (responseStatus resp) "Could not retrieve rich info")
+      _ -> rethrow resp
 
 checkHandleAvailable :: (HasCallStack, MonadSparToBrig m) => Handle -> m Bool
 checkHandleAvailable hnd = do
--- a/services/spar/src/Spar/Intra/Brig.hs
+++ b/services/spar/src/Spar/Intra/Brig.hs
@@ -21,17 +21,23 @@
 module Spar.Intra.Brig
   ( toUserSSOId,
     fromUserSSOId,
-    toExternalId,
+    urefToExternalId,
+    userToScimExternalId,
     mkUserName,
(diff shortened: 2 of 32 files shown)


fisx:
TODO:

- [ ] `/wire-server/libs/wire-api/src/Wire/API/User/Identity.hs:256:9: error: [-Wpartial-fields, -Werror=partial-fields]` (and two more)
- [ ] f88559d
- [ ] any important test cases that I should add?

services/spar/src/Spar/Intra/Brig.hs
```diff
@@ -130,20 +196,20 @@ instance MonadSparToBrig m => MonadSparToBrig (ReaderT r m) where
 createBrigUser ::
   (HasCallStack, MonadSparToBrig m) =>
   -- | SSO identity
-  SAML.UserRef ->
+  ValidExternalId ->
```
fisx, line 199: This is one of the more meaningful changes in the PR.  Above are helpers inspired by this new type, below I mostly re-aligned functions, imrpoved error handling a little, and called better (more internal) end-points.

fisx (commented):
I re-read everything except `Spar.Scim.User` (which is admittedly one of the more important modules).  Some comments below.

fisx:
also, this just failed (non-deterministically?):

```

  test-integration/Test/Spar/Scim/UserSpec.hs:660:10: 
  1) Spar.Scim.User, GET /Users, doesn't list deleted users
       predicate failed on: User {userId = 7b85c17c-30c7-4499-a27e-7920c34b215a, userIdentity = Just (SSOIdentity (UserSSOId "<Issuer xmlns:samlp=\"urn:oasis:names:tc:SAML:2.0:protocol\" xmlns:samla=\"urn:oasis:names:tc:SAML:2.0:assertion\" xmlns:samlm=\"urn:oasis:names:tc:SAML:2.0:metadata\" xmlns:ds=\"http://www.w3.org/2000/09/xmldsig#\" xmlns=\"urn:oasis:names:tc:SAML:2.0:assertion\">https://issuer.net/_23776545-901f-48e8-a044-e2e40d21d77a</Issuer>" "<NameID xmlns:samlp=\"urn:oasis:names:tc:SAML:2.0:protocol\" xmlns:samla=\"urn:oasis:names:tc:SAML:2.0:assertion\" xmlns:samlm=\"urn:oasis:names:tc:SAML:2.0:metadata\" xmlns:ds=\"http://www.w3.org/2000/09/xmldsig#\" Format=\"urn:oasis:names:tc:SAML:1.1:nameid-format:emailAddress\" xmlns=\"urn:oasis:names:tc:SAML:2.0:assertion\">scimuser_extid_3066607@example.com</NameID>") Nothing Nothing), userDisplayName = Name {fromName = "Scim User #3066607"}, userPict = Pict {fromPict = []}, userAssets = [], userAccentId = ColourId {fromColourId = 0}, userDeleted = False, userLocale = en, userService = Nothing, userHandle = Just (Handle {fromHandle = "scimuser_3066607"}), userExpire = Nothing, userTeam = Just b53ce0d8-9032-460c-b6ae-fb15a8a3436e, userManagedBy = ManagedByScim}

  To rerun use: --match "/Spar.Scim.User/GET /Users/doesn't list deleted users/"

Randomized with seed 469431066
```

services/spar/test-integration/Test/Spar/DataSpec.hs
```diff
@@ -274,12 +275,24 @@ testDeleteTeam = it "cleans up all the right tables after deletion" $ do
...
-    let Right uref1 = fromUserSSOId ssoid1
-    mbUser1 <- runSparCass $ Data.getSAMLUser uref1
+    mbUser1 <- case veidFromUserSSOId ssoid1 of
+      Right veid ->
+        runSparCass $
+          runValidExternalId
+            Data.getSAMLUser
+            undefined -- could be @Data.lookupScimExternalId@, but we don't hit that path.
+            veid
+      Left _email -> undefined -- runSparCass . Data.lookupScimExternalId . fromEmail $ _email
```
arianvp, line 285: Why is this case undefined?

services/spar/test-integration/Test/Spar/Scim/UserSpec.hs
```diff
@@ -190,9 +200,75 @@ specCreateUser = describe "POST /Users" $ do
...
+    (_, tid) <- call $ createUserWithTeam (env ^. teBrig) (env ^. teGalley)
+    registerScimToken tid Nothing
+  user <- randomScimUser <&> \u -> u {Scim.User.password = Just "geheim"}
+  createUser_ (Just tok) user (env ^. teSpar) !!! do
+    const 400 === statusCode
+    -- TODO: write a FAQ entry in wire-docs, reference it in the error description.
+    -- TODO: yes, we should just test for error labels consistently, i know...
+    const (Just "Setting user passwords is not supported for security reasons.") =~= responseBody
+
+testCreateUserNoIdP :: TestSpar ()
```
arianvp, line 216: The description of this test says it checks tht the email validation link is sent; but i dont see this being tested here.
How are we making sure this codepath is indeed taken?

services/spar/test-integration/Test/Spar/Scim/UserSpec.hs
```diff
@@ -190,9 +200,75 @@ specCreateUser = describe "POST /Users" $ do
...
+  (_, tid) <- call $ createUserWithTeam (env ^. teBrig) (env ^. teGalley)
+  tok <- registerScimToken tid Nothing
+  scimStoredUser <- createUser tok scimUser
+  let userid = scimUserId scimStoredUser
+      handle = Handle . Scim.User.userName . Scim.value . Scim.thing $ scimStoredUser
+  brigUser <-
+    aFewTimes (runSpar $ Intra.getBrigUser userid) isJust
+      >>= maybe (error "could not find user in brig") pure
+  brigUser `userShouldMatch` WrappedScimStoredUser scimStoredUser
+  liftIO $ userEmail brigUser `shouldBe` Nothing
```
arianvp, line 230: Why is the email `Nothing` ? I would expect it to be the email with which the user is being registered through scim? Wasn't that the point of this PR?

arianvp: Nevermind scrap that; I see if comes later in the test

services/spar/test-integration/Test/Spar/Scim/UserSpec.hs
arianvp: Scrap that ;I can't read :)

arianvp:
It would be nice if we could document the new behaviour in the scim documentation; and also update the PR description to explain what exactly this PR implements. Especially because we changed designs like 3 times.

From what I understand the gist is that we send   email activation emails upon inviting the user through scim; and modified brig to make users not ephemeral.

services/spar/test-integration/Test/Spar/DataSpec.hs
fisx: yes, it never happens.  it's perfectly reasonable to allow this case, and the comment two lines above says how it could be implemented, but i felt that it's better to not write dead code, but rather make it clear where we don't walk.

feel free to turn it into a `fromRight` if you think it's an improvement.

fisx:
> It would be nice if we could document the new behaviour in the scim documentation; and also update the PR description to explain what exactly this PR implements. Especially because we changed designs like 3 times.
> 
> From what I understand the gist is that we send email activation emails upon inviting the user through scim; and modified brig to make users not ephemeral.

Yes.  I added a few more ideas to the description of this PR.  We should probably move them to the docs somewhere, but I don't know where exactly.  Ah, wait, I know!  I'll make another commit.  :)

docs/reference/spar-braindump.md
```diff
@@ -21,6 +21,25 @@ documentation answering your questions, look here!
...
+
+When SCIM user gets are processed, in order to reconstruct the externalId from the user spar is retrieving from brig, we introduce a new json object for the `sso_id` field that looks like this: `{'scim_external_id': 'me@example.com'}`.
+
+In order to find users that have email addresses pending validation, we introduce a new table in spar's cassandra called `scim_external_ids`, in analogy to `user`.  We have tried to use brig's internal `GET /i/user&email=...`, but that also finds pending email addresses, and there are corner cases when changing email addresses and waiting for the new address to be validated and the old to be removed...  that made this approach seem infeasible.
+
+**password:** once the user has validated their email address, they need to trigger the "forgot password" flow -- also old code.
```
arianvp, line 40: This doesn't sound great? Will the reset-password flow be triggered automatically in any way? Without it; the UX of this approach is not good and I foresee a lot of handholding needed

arianvp:
CI is currently failing because this PR uses `Scim.util.deletUser` which has a bug (it expects 200 whilst it should expect 204). Easy fix

docs/reference/spar-braindump.md
```diff
@@ -21,6 +21,25 @@ documentation answering your questions, look here!
...
+
+The way this works is different from team-settings: we don't send invites, but we create active users immediately the moment the SCIM user post is processed.  The new thing is that the created user has neither email nor phone nor a SAML identity, nor a password.
+
+How does this work?
+
+**email:** If no SAML IdP is present, SCIM user posts must contain an externalId that is an email address.  This email address is not added to the newly created user, because it has not been validated.  Instead, the flow for changing an email address is triggered in brig: an email is sent to the address containing a validation key, and once the user completes the flow, brig will add the email address to the user.  We had to add very little code for this in this PR, it's all an old feature.
+
+When SCIM user gets are processed, in order to reconstruct the externalId from the user spar is retrieving from brig, we introduce a new json object for the `sso_id` field that looks like this: `{'scim_external_id': 'me@example.com'}`.
```
tiago-loureiro, line 36:
```suggestion
When SCIM user GETs are processed, in order to reconstruct the externalId from the user spar is retrieving from brig, we introduce a new json object for the `sso_id` field that looks like this: `{'scim_external_id': 'me@example.com'}`.
```

arianvp (approved):
Approved.

No nits on the code;  but still a big question-mark on whether this is how we want to have this feature at all.  I don't think it's very user-friendly and I do not think it's going to save us time in the long run.  However, it will unblock us for now. Which is good. This should be discussed in a retro.

PR #1204, opened 2020-09-14
[skip ci] Upgrade GHC to 8.8.4

akshaymankar:
Fixes https://github.com/zinfra/backend-issues/issues/1622

### DIFF
--- a/stack.yaml
+++ b/stack.yaml
@@ -1,4 +1,4 @@
-resolver: lts-16.7
+resolver: lts-16.14
 
 packages:
 - libs/api-bot
--- a/stack.yaml
+++ b/stack.yaml
@@ -195,8 +195,8 @@ extra-deps:
 
 - git: https://gitlab.com/axeman/redis-io.git
   commit: e113fa4de8fcc3665c3c562a05a07f7a1d63b3cc # ghc-8.8 (Aug 24, 2020)
-- git: https://github.com/zoominsoftware/HaskellNet-SSL
-  commit: 5b4f4ab49a6321b57f6dbeba64a7f4bb8ffa5cbc # allow-newer-network, dpwright/HaskellNet-SSL#24  (Aug 24, 2020)
+- git: https://github.com/dpwright/HaskellNet-SSL
+  commit: ca84ef29a93eaef7673fa58056cdd8dae1568d2d # master (Sep 14, 2020)
 
 # Not in the LTS, not using latest on hackage due to breaking change
 - quickcheck-state-machine-0.6.0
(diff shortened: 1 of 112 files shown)


akshaymankar:
Special pipeline here: https://bridge-ie-concourse.zinfra.io/teams/main/pipelines/wire-server-ghc88.

stack.yaml
```diff
@@ -55,7 +55,7 @@ nix:
 allow-newer: true
 
 extra-deps:
-- swagger2-2.4
+# - swagger2-2.4
```
jschaul, line 58: can this be removed? Or why is it commented?

stack.yaml
```diff
@@ -64,7 +64,7 @@ extra-deps:
   # a version > 1.0.0 of wai-middleware-prometheus is available
   # (required: https://github.com/fimad/prometheus-haskell/pull/45)
 - git: https://github.com/wireapp/saml2-web-sso
-  commit: 687d9ac8ac2994aff8436189c6ecce29faad8500  # master (May 18, 2020)
+  commit: be04baaffb92ff64d8e1d9f400ec45849d754faf  # akshaymankar/ghc-8.8 (Aug 21, 2020)
```
jschaul, line 67: The various forks should have a reference to a stable branch, not a PR branch which will be deleted after merging the PR. Perhaps before merging this PR; merge the upstream ones and use a stable branch commit identifier?

jschaul (approved):
Generally looks good.

Two things:

1. can the dependencies be based on merged commits, rather than hanging branches that may disappear by accident?

> Special pipeline here: https://bridge-ie-concourse.zinfra.io/teams/main/pipelines/wire-server-ghc88.

2. can the ubuntu-based Docker image also be tested with ghc-8.8? The pipeline seems to include an ubuntu image as well; however I cannot find the pipeline code that was used to create this pipeline anywhere. Please link.

stack.yaml
akshaymankar: This one is the only one we control. I created PR here: https://github.com/wireapp/saml2-web-sso/pull/66
No other branches will get deleted on merge, I have set the option to not delete it in Gitlab. For github forks we don't need to worry as github only deletes branches in the same repo. 

I suspect some of the upstream maintainers aren't maintaining the packages very actively, so it might take very long. So, I think it is fine to refer to these packages in their forks.

stack.yaml
akshaymankar: Hmm, I forgot a few comments, I will clean them up, thanks!

Issue #1203, opened 2020-09-09
Cargohold v2.88.0 won't upload objects to s3 because of BadDigest error


obstenu:
**Describe the bug**
After updated to v.2.88.0 Cargohold just won't upload any objects to s3 (I'm using Google Storage, and v.2.87.0 did work for me). After brief research it came out what in v,2.88.0 the following [PR](https://github.com/wireapp/wire-server/pull/1157) has been applied, so it may probably be the reason.

**To Reproduce**
0. Use Google Storage (tbc I haven't tested if it works with AWS S3)
1. Upgrade to v2.88.0
2. See the bug as described above

**Expected behavior**
Everything just works

**Log file cut**
_5-UwRn[REDACTED]VWfX9Sw"),("Content-Type","application/xml; charset=UTF-8"),("Content-Length","262"),("Vary","Origin"),("Date","Wed, 09 Sep 2020 11:41:28 GMT"),("Server","UploadServer")], _serviceCode = ErrorCode "BadDigest", _serviceMessage = Just (ErrorMessage "The MD5 you specified in Content-MD5 or x-goog-hash did not match what we computed."), _serviceRequestId = Nothing}),_

lucendio:
Unfortunately, we do not support the GCP with all it's native resources. After all, Amazonka is *a comprehensive Amazon Web Services SDK*. You may want to replace *Google Storage* with your own deployment of [min.io](https://docs.min.io/) ([Andible role](https://github.com/wireapp/ansible-minio)).

PR #1205, opened 2020-09-15
Cargohold: Log more about AWS errors

mheinzel:
This will help us debug some things we're seeing since #1157, such as https://github.com/zinfra/backend-issues/issues/1665 and https://github.com/wireapp/wire-server/issues/1203.


jschaul (approved):
Looks good.

PR #1206, opened 2020-09-16
limit list of SFT servers to 5 (configurable), independent of the amount of servers …

jschaul:
…in the SRV record

Can be configured with

```yaml
sft:
  sftListLength: 17 # an integer outside the range 1,100 will throw an error at startup time
```

### DIFF
--- a/services/brig/src/Brig/Calling/API.hs
+++ b/services/brig/src/Brig/Calling/API.hs
@@ -1,3 +1,5 @@
+{-# LANGUAGE BlockArguments #-}
+
 -- This file is part of the Wire Server implementation.
 --
 -- Copyright (C) 2020 Wire Swiss GmbH <opensource@wire.com>
@@ -40,7 +42,7 @@ import Data.Text.Strict.Lens
 import Data.Time.Clock.POSIX (getPOSIXTime)
 import Imports hiding (head)
 import Network.Wai (Response)
-import Network.Wai.Predicate hiding ((#), and, result, setStatus)
+import Network.Wai.Predicate hiding (and, result, setStatus, (#))
 import Network.Wai.Routing hiding (toList)
 import Network.Wai.Utilities hiding (code, message)
 import Network.Wai.Utilities.Swagger (document)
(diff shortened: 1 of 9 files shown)


services/brig/src/Brig/Options.hs
```diff
@@ -563,14 +566,17 @@ defSftServiceName = "_sft"
...
     let defaultV = 3600
         bounded = toBoundedInteger n :: Maybe Int64
-     in pure $
-          Timeout $
-            fromIntegral @Int $
-              maybe defaultV fromIntegral bounded
+     in pure
+          $ Timeout
+          $ fromIntegral @Int
+          $ maybe defaultV fromIntegral bounded
```
akshaymankar, line 579: I think you're not using the right version of ormolu.

services/brig/src/Brig/Calling/API.hs
```diff
@@ -124,14 +126,30 @@ newConfig env mSftEnv limit = do
...
+      -- priority and weight, but we internally agreed to randomize the list of
+      -- available servers for poor man's "load balancing" purposes.
+      -- FUTUREWORK: be smarter about list orderding depending on how much capacity SFT servers have.
+      randomizedSftEntries <- liftIO $ mapM randomize sftSrvEntries
+
+      -- Currently (Sept 2020) the client initiating an SFT call will try all
+      -- servers in this list. Limit this list to a smaller subset in case many
+      -- SFT servers are advertised in a given environment.
+      let subsetLength = Calling.sftListLength actualSftEnv
+      return $ subsetSft subsetLength <$> randomizedSftEntries
```
akshaymankar, line 143: Since all of this is _pure-ish_, we can unit test this.

akshaymankar (changes requested):
Looks good, two requests:
1. Please add tests, I think we don't really need to integration test this, but unit testing should be trivial.
1. Please upgrade ormolu (should work with `stack install ormolu` from root of the repo) and reformat.

services/brig/src/Brig/Options.hs
jschaul: Fixed.

services/brig/src/Brig/Calling/API.hs
jschaul: Let's maybe discuss in a call how best to test this without using an integration test; as this happens outside the easily-unit-testable discoveryloop logic.

akshaymankar (approved):

PR #1207, opened 2020-09-16
Manage invited user by SCIM when SSO is enabled

tiago-loureiro:
When a user has been invited to a team with SAML IdP via team settings, and then is searched via SCIM, the user won't be found and bound.  This is most likely because spar doesn't treat the email address that is the `externalId`  as an email, but as a SAML `NameID`.

This PR will fix this.

Fixes https://github.com/zinfra/backend-issues/issues/1572

### DIFF
--- a/services/spar/test-integration/Test/Spar/Scim/UserSpec.hs
+++ b/services/spar/test-integration/Test/Spar/Scim/UserSpec.hs
@@ -1067,7 +1067,7 @@ testUpdateExternalId withidp = do
         storedUser <- createUser tok user
         let userid = scimUserId storedUser
         veid :: ValidExternalId <-
-          either (error . show) pure $ mkUserRef midp (Scim.User.externalId user)
+          either (error . show) pure $ mkValidExternalId midp (Scim.User.externalId user)
         -- Overwrite the user with another randomly-generated user (only controlling externalId)
         user' <- do
           otherEmail <- randomEmail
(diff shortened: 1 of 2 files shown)


services/spar/test-integration/Test/Spar/Scim/UserSpec.hs
```diff
@@ -636,10 +663,10 @@ testFindNonProvisionedUserNoIdP = do
     liftIO $ userManagedBy brigUser `shouldBe` ManagedByWire
     liftIO $ userEmail brigUser `shouldSatisfy` isJust
 
-  byHandle <- listUsers tok (Just (filterBy "userName" (fromHandle handle)))
   byExternalId <- listUsers tok (Just (filterBy "externalId" (fromEmail email)))
+  byHandle <- listUsers tok (Just (filterBy "userName" (fromHandle handle)))
```
fisx, line 670: why?

tiago-loureiro: I just wanted to make sure that the "migration" happened also when looking up by `externalId` and not just by `userName`; when I first looked at the code I was a bit confused and thought: "oh, this works because we migrate the user _first_ by `userName`"

fisx: ah, makes sense.  but in that case, we should just test both independently, no?  i'll see if i can rewrite it some more to make this clearer.

fisx:
I can't reproduce https://github.com/zinfra/backend-issues/issues/1764 on develop, so we should probably look into that before merging this here...

fisx:
I think the errors have been truncated from the CI log.  I'll attach the copy that I got with fly.
[ci-log.txt](https://github.com/wireapp/wire-server/files/5281278/ci-log.txt)

fisx:
integration tests pass locally.

tiago-loureiro:
> integration tests pass locally.

and remotely!

fisx (approved):

PR #1209, opened 2020-09-22
Drop unused types.

fisx:
I'm a bit nervous about this one.

- The types tell me I'm right, and as the haddocks of the `Priority` type state, the evidene is in favor of this. 
- Worst thing that can happen is that we'll want to send low-priority messages in the future for some reason, and won't be able to (first commit), or that we'll want to change the order of preference between different types (second commit); both seem unlikely to me.

But I have little context in this matter.

I'll open it anyway as a PR, since writing it helped me understand the code better, and what belongs to what; perhaps reading it will help, too.  Happy to have it rejected.  :)

### DIFF
--- a/libs/api-bot/src/Network/Wire/Bot/Crypto.hs
+++ b/libs/api-bot/src/Network/Wire/Bot/Crypto.hs
@@ -139,7 +139,6 @@ encryptMessage clt cnv msg =
     <$> encrypt clt cnv msg
     <*> pure False -- Native push?
     <*> pure False -- Transient?
-    <*> pure Nothing -- Priority
     <*> pure Nothing -- Extra data distributed to all recipients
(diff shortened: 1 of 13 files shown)


akshaymankar:
Did you ask the client teams about this?

fisx:
> Did you ask the client teams about this?

I think somebody did at some point, plus [this](https://github.com/wireapp/wire-server/blob/develop/libs/wire-api/src/Wire/API/Message.hs#L125-L135), plus [this](https://github.com/wireapp/wire-server/pull/1209#issue-491113899).

One thing that this may not have caught is protobuf data.  So, good point, thanks!  I'll ask again.

fisx:
https://github.com/zinfra/backend-issues/issues/889#issuecomment-707075287

PR #1210, opened 2020-09-22
cargohold compat target for testing different S3-like buckets

jschaul:
If testing against S3, minio, and others, this allows to run

```
make compat-minio
make compat-s3
...
```

Requires a `compat` folder (currently symlinked to a private repository containing some credentials)

### DIFF
--- a/services/integration.sh
+++ b/services/integration.sh
@@ -62,6 +62,14 @@ if [[ $INTEGRATION_USE_REAL_AWS -eq 1 ]]; then
     [ -z "$AWS_SECRET_ACCESS_KEY" ] && echo "Need to set AWS_SECRET_ACCESS_KEY in your environment" && exit 1;
     "${TOP_LEVEL}"/services/gen-aws-conf.sh
     integration_file_extension='-aws.yaml'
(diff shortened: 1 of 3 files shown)


services/integration.sh
```diff
@@ -88,21 +97,26 @@ function run() {
...
-run brig "" ${green}
-run galley "" ${yellow}
-run gundeck "" ${blue}
-run cannon "" ${orange}
-run cannon "2" ${orange}
-run cargohold "" ${purpleish}
-run spar "" ${orange}
-run federator "" ${blue}
+if [[ $INTEGRATION_CARGOHOLD_ONLY_COMPAT -eq 1 ]]; then
+    source "${CARGOHOLD_COMPAT_CONFIG_FOLDER}/env.sh"
```
mheinzel, line 106: We could also do relevant parts of `check_prerequisites` here, but not a big deal now that this is not run routinely.

mheinzel (commented):
Introduces more conditionals/complexity in `integration.sh`, but that should be worth it. Before merging, let me just make sure it works for me.

mheinzel (approved):
I was able to run it successfully!

PR #1213, opened 2020-09-25
SCIM onboarding flow again

fisx:
Fixes https://github.com/zinfra/backend-issues/issues/1761


libs/wire-api/src/Wire/API/User.hs
```diff
@@ -507,11 +508,29 @@ isNewUserEphemeral u = noId && noScim
...
+    Nothing -> False
+    Just origin ->
+      case origin of
+        (NewUserOriginInvitationCode _) -> False
+        (NewUserOriginTeamUser newTeamUser) ->
+          case newTeamUser of
+            (NewTeamMember _) -> False
+            (NewTeamCreator _) -> False
+            (NewTeamMemberSSO _) -> False
+            (NewTeamMemberScimInvitation _) -> True
```
fisx, line 525:
```suggestion
  case newUserOrigin u of
    Nothing -> False
    Just (NewUserOriginInvitationCode _) -> False
    Just (NewUserOriginTeamUser (NewTeamMember _)) -> False
    Just (NewUserOriginTeamUser (NewTeamCreator _)) -> False
    Just (NewUserOriginTeamUser (NewTeamMemberSSO _)) -> False
    Just (NewUserOriginTeamUser (NewTeamMemberScimInvitation _)) -> True
```

or even:

```suggestion
  case newUserOrigin u of
    Just (NewUserOriginTeamUser (NewTeamMemberScimInvitation _)) -> True
    _ -> False
```

libs/wire-api/src/Wire/API/User.hs
```diff
@@ -688,15 +708,17 @@ parseNewUserOrigin pass uid ssoid o = do
...
+  scimInvTeamid <- o .:? "scim_invitation_team_id"
+  result <- case (invcode, teamcode, team, ssoid, teamid, scimInvTeamid) of
+    (Just a, Nothing, Nothing, Nothing, Nothing, Nothing) -> return . Just . NewUserOriginInvitationCode $ a
+    (Nothing, Just a, Nothing, Nothing, Nothing, Nothing) -> return . Just . NewUserOriginTeamUser $ NewTeamMember a
+    (Nothing, Nothing, Just a, Nothing, Nothing, Nothing) -> return . Just . NewUserOriginTeamUser $ NewTeamCreator a
+    (Nothing, Nothing, Nothing, Just _, Just t, Nothing) -> return . Just . NewUserOriginTeamUser $ NewTeamMemberSSO t
+    (Nothing, Nothing, Nothing, Nothing, Nothing, Just t) -> return . Just . NewUserOriginTeamUser $ NewTeamMemberScimInvitation t
+    (Nothing, Nothing, Nothing, Nothing, Nothing, Nothing) -> return Nothing
+    (_, _, _, Just _, Nothing, Nothing) -> fail "sso_id, team_id must be either both present or both absent."
+    (_, _, _, Nothing, Just _, Nothing) -> fail "sso_id, team_id must be either both present or both absent."
```
fisx, line 720:
```suggestion
    (_, _, _, Just _, Nothing, _) -> fail "sso_id, team_id must be either both present or both absent."
    (_, _, _, Nothing, Just _, _) -> fail "sso_id, team_id must be either both present or both absent."
```

fisx: (amounts to the same thing, but it's more consistent with the other cases.)

services/brig/src/Brig/API/Public.hs
```diff
@@ -1025,6 +1025,8 @@ createUser (Public.NewUserPublic new) = do
         Team.sendMemberWelcomeMail e t n l
       Public.NewTeamMemberSSO _ ->
         Team.sendMemberWelcomeMail e t n l
+      Public.NewTeamMemberScimInvitation _ ->
+        return () -- TODO: send mail here?
```
fisx, line 1029: yes. :)

services/brig/src/Brig/API/User.hs
```diff
@@ -262,14 +257,10 @@ createUser' viaScim new@NewUser {..} = do
...
     handleTeam (Just (NewTeamCreator t)) _ = do
-      -- invariant: viaScim /= CreateUserNoVerifyViaScim
-      -- FUTUREWORK: throw 400 in case of violation
       (Just t,Nothing,) <$> (Just . Id <$> liftIO nextRandom)
     handleTeam (Just (NewTeamMemberSSO tid)) _ = pure (Nothing, Nothing, Just tid)
-    handleTeam Nothing _ = do
-      -- invariant: viaScim /= CreateUserNoVerifyViaScim
-      -- FUTUREWORK: throw 400 in case of violation
-      return (Nothing, Nothing, Nothing)
+    handleTeam (Just (NewTeamMemberScimInvitation tid)) _ = pure (Nothing, Nothing, Just tid)
```
fisx, line 262: not sure about this; note to self: double check!  in what situations is `handleTeam` called?

services/brig/src/Brig/API/Public.hs
smatting: k will fix

services/spar/src/Spar/Intra/Brig.hs
```diff
@@ -244,7 +258,9 @@ updateEmail buid email = do
     _ -> throwError . SAML.CustomServant . waiToServant . responseJsonUnsafe $ resp
 
 getBrigUser :: (HasCallStack, MonadSparToBrig m) => UserId -> m (Maybe User)
-getBrigUser = (accountUser <$$>) . getBrigUserAccount
+getBrigUser =
+  -- TODO: we sometimes need to also retrieve PendingInvitation-statussed users here.
```
fisx, line 262: I'm going to flesh this out a little more now.

services/brig/src/Brig/API/Public.hs
```diff
@@ -1025,6 +1025,8 @@ createUser (Public.NewUserPublic new) = do
         Team.sendMemberWelcomeMail e t n l
       Public.NewTeamMemberSSO _ ->
         Team.sendMemberWelcomeMail e t n l
+      Public.NewTeamMemberScimInvitation _ ->
+        Team.sendMemberWelcomeMail e t n l
```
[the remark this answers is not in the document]
smatting: Hatte hier angenommen, dass welcome-mail eine invitation-mail ist.

services/brig/src/Brig/Data/User.hs
```diff
@@ -128,9 +129,10 @@ newAccount u inv tid = do
     pict = fromMaybe noPict (newUserPict u)
     assets = newUserAssets u
     status =
-      if isNewUserEphemeral u
-        then Ephemeral
-        else Active
+      if
+          | isNewUserEphemeral u -> Ephemeral
+          | isNewUserInvitedViaScim u -> PendingInvitation
```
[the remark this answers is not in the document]
smatting: Is your intention here to remove `isNewUserInvitedViaScim` from `Wire.API.User`? Because it is called from  `validateNewUserPublic` in `Wire.API.User`.

services/brig/src/Brig/API/Public.hs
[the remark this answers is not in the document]
fisx: > Hatte hier angenommen, dass welcome-mail eine invitation-mail ist.

yes, that is correct, but the point is that this handler is public, and takes a `NewUserPublic`, so this case cannot happen.

how should i rephrase the error message to make this easier to understand?

services/brig/src/Brig/Data/User.hs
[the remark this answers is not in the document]
fisx: right.  i'll take it back, then, it would only have made sense if this was the only call to that function.

services/spar/src/Spar/Intra/Brig.hs
```diff
@@ -214,10 +214,24 @@ createBrigUser veid (Id buid) teamid uname managedBy = do
       newUser =
         (emptyNewUser uname)
```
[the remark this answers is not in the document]
smatting: ## `NewUser` value in tradidional invite flow

I'm using the integration test `testInvitationEmailLookupRegister` as "traditional invite flow"

`NewUserPublic` is parsed (in the handler of `/register`) from the json created with `accept`
`services/brig/test/integration/API/Team/Util.hs:295`
```
accept :: Email -> InvitationCode -> RequestBody
accept email code = acceptWithName (Name "Bob") email code

acceptWithName :: Name -> Email -> InvitationCode -> RequestBody
acceptWithName name email code =
  RequestBodyLBS . encode $
    object
      [ "name" .= fromName name,
        "email" .= fromEmail email,
        "password" .= defPassword,
        "team_code" .= code
      ]
```

This results in a `NewUser` value with exactly these fields set (see `instance FromJSON NewUser`) (all other are `Nothing`):
- `newUserIdentity` (from "email"): `Just (Emailidentity email)`
- `newUserOrigin`  (from "team_code", "password"): `Just (NewUserOriginTeamUser (NewTeamMember InvitationCode))`
- `newUserPassword` (from "password"): `Just (PlainTextPassword passwd)`
- `newUserDisplayName` (from "name" field"): `Just (Name name)`

`createUser` (Brig.API.Public) is then called with this `NewUserPublic` value.

## `NewUser` in scim flow
Looking `Intra/Brig.hs` these fields are set during the scim invitation (all other are `Nothing`):

- `newUserIdentity`: `SSOIdentity` or `EmailIdentity` depending on the externalId
- `newUserOrigin`: `NewUserOriginTeamUser (NewTeamMemberScimInvitation (Id teamId))` or `NewUserOriginTeamUser` or `(NewUserOriginTeamUser (NewTeamMemberScimInvitation (Id teamId)))`
- `newUserUUID`: `Just (Id uuid)` depending on the externalId
- `newUserManagedBy`: `Just managedBy`

services/brig/src/Brig/API/Internal.hs
```diff
@@ -302,11 +307,36 @@ createUserNoVerify uData = do
...
+      case (eac, newUserEmail uData) of
+        (Just adata, Just email) -> do
+          void $
+            lift $
+              sendActivationEmail
+                email
+                (newUserDisplayName uData)
+                (activationKey adata, activationCode adata)
+                (newUserLocale uData)
+                Nothing
```
fisx, line 322:
```suggestion
                (Public.newUserTeam uData)
```

services/brig/src/Brig/API/Internal.hs
```diff
@@ -302,11 +307,36 @@ createUserNoVerify uData = do
...
-        code = activationCode adata
-     in API.activate key code (Just uid) !>> actError
+
+  if isNewUserInvitedViaScim uData
+    then do
+      case (eac, newUserEmail uData) of
+        (Just adata, Just email) -> do
+          void $
+            lift $
+              sendActivationEmail
```
fisx, line 317:
```suggestion
        (Just adata, Just email) -> do
          void . lift $ sendActivationEmail
```

not sure if ormolu won't add a newline after `$`, but that's at least one less than here.  :)

fisx:
### Notes from todays call with @smatting, @jschaul.

#### Aren't the columns `activated` and `status` in table `brig.user` redundant?

`status` was introduced for teams, `activated` is older.  `activated` was used to keep users from being created in a flow that is not used by clients any more (See https://github.com/wireapp/wire-server/blob/84ca176fb79e498db50eb9097dcb322ceec91e82/docs/reference/user/registration.md#registration-without-pre-verification-refregistrationnopreverification).

`activated` is also still useful to mark users that have not been fully created, so that in case brig crashes during creation of a user, the user won't be in a broken state.  `activated==false` is morally equivalend to `does not exist yet`; `activated==true` means that the email address stored in the user is validated and the user is ready to go.

#### Will the second part of traditional team invitation flow work out of the box?

Question: From the moment when we have created the user with status `PendingInvitation`, will the rest of the flow work, or do we need to change things?

Short answer: we already have an integration test for this, so we should find out soon enough.  :)

Long answer: the invitation has the same form as in in the team-settings-based flow (except for missing inviter); there is a pending user, but that should be overwritten by `Brig.API.User.createUser`, since the `UserId` is the same as the `InvitationId`, which we get from the `Invitation`.

#### Corner cases

If an invitation expires, the pending user must go away again.  In the naive implementation, she would still be visible for SCIM peers, but she would not be able to login, and re-creating or activating via SCIM wouldn't work either.

So the pending user record must effectively get the same TTL as the invitation.

There are at least two ways to implement this.

1. Give the user record a TTL and and later remove that, when the invitation is accepted.  (We need to find out **and test** if cassandra supports this..)
2. Whenever we find a user with `activated=false, status=PendingInvitation`, check if there is a matching Invitation.  If not, assume its TTL has expired, and remove the user record.  Otherwise, return it.

#### What about re-sending emails?

That's a different issue, we won't address this here: https://github.com/zinfra/backend-issues/issues/1601

#### Tasks

- [ ] make sure that if a user has `status==PendingActivation`, it always has `activated==false`.
- [ ] somehow give users with `status==PendingInvitation` the same TTL as the matching invitations (see above).
- [ ] make sure we test the following
  - create user
  - wait for invitation TTL to expire
  - user doesn't show in scim any more
  - user tries to invite, fails
  - user still doesn't show in scim
  - sending the same `post /scim/Users` request a second time will have the same effect as the first time
- [ ] make sure we test that `get /i/users?ids=...` does not return pending users.

services/spar/test-integration/Test/Spar/Scim/UserSpec.hs
```diff
@@ -660,6 +664,93 @@ testScimCreateVsUserRef = do
...
+      env <- ask
+      let brig = env ^. teBrig
+
+      scimStoredUser <- createUser tok scimUser
+      let handle = Handle . Scim.User.userName . Scim.value . Scim.thing $ scimStoredUser
+
+      inv <- call $ getInvitation brig email
+
+      -- Wait for user and invitation to expire
+      let setTeamInvitationTimeout = 5 -- Must match the value in brig.integration.yaml
```
fisx, line 690:
```suggestion
      let setTeamInvitationTimeout = env ^. teOpts ...
```

ah.  no.  this is annoying.  [this is the value we're looking for](https://github.com/wireapp/wire-server/blob/84ca176fb79e498db50eb9097dcb322ceec91e82/services/brig/brig.integration.yaml#L133), but we can't access that here.

i think we have to change the `TestEnv` type to contain the bring service options next to the spar service options.  it's not a complicated change, just really annoying.

services/spar/test-integration/Test/Spar/Scim/UserSpec.hs
```diff
@@ -660,6 +664,93 @@ testScimCreateVsUserRef = do
...
+      do
+        call $ do
+          void $ do
+            Just inviteeCode <- getInvitationCode brig tid (inInvitation inv)
+            post
+              ( brig . path "/register"
+                  . contentJson
+                  . json (acceptWithName (Name "Bob") email inviteeCode)
+              )
+              <!! const 201 =/= statusCode
```
fisx, line 710:
```suggestion
              <!! const 400 === statusCode
```

...  or something.

weirdly, i just looked at the code, and couldn't see any error being thrown if the user tries to accept an invitation that doesn't exist any more.  i think there may be another bug in the code, where in this situation, a non-team users will be created and returned.

I may be missing something, good that we're writing this test!  :)  so, I say let's check for status code 999 and run the test, an we'll see what status code we actually get, and why.

services/spar/test-integration/Test/Spar/Scim/UserSpec.hs
```diff
@@ -660,6 +664,93 @@ testScimCreateVsUserRef = do
...
+
+      -- User still doesn't show in scim
```
fisx, line 712: I would put this block in a definition under a nice name, and call that definition twice, to reduce code duplication.

It's a question of taste, though.

services/spar/test-integration/Test/Spar/Scim/UserSpec.hs
```diff
@@ -660,6 +664,93 @@ testScimCreateVsUserRef = do
...
+      do
+        listUsers tok (Just (filterBy "userName" $ fromHandle handle)) >>= \users ->
+          liftIO $ users `shouldBe` []
+        listUsers tok (Just (filterBy "externalId" $ fromEmail email)) >>= \users ->
+          liftIO $ users `shouldBe` []
+
+    getInvitation :: BrigReq -> Email -> Http Invitation
```
fisx, line 719: lines 314ff in this module have some of these definitions.  move all of them to `services/spar/test-integration/Util/Invitation.hs` (or some of the existing `Util` modules), make sure it's exported from `Util`, and avoid code duplication.

services/spar/test-integration/Test/Spar/Scim/UserSpec.hs
```diff
@@ -660,6 +664,93 @@ testScimCreateVsUserRef = do
...
+  env <- ask
+
+  email <- randomEmail
+  scimUser <- randomScimUser <&> \u -> u {Scim.User.externalId = Just $ fromEmail email}
+  (_owner, tid) <- call $ createUserWithTeam (env ^. teBrig) (env ^. teGalley)
+  tok <- registerScimToken tid Nothing
+
+  testFlow tid tok scimUser email
+  -- creating the same user again should behave the same
+  testFlow tid tok scimUser email
```
fisx, line 678: this time, though, the invitation should be accepted, and the user should successfully be created, no?

services/spar/test-integration/Test/Spar/Scim/UserSpec.hs
fisx: I'll give this one a shot.

fisx: dcb21fa6602f48e035d6ab4575930fe1874ead8b

services/brig/src/Brig/API/Internal.hs
```diff
@@ -307,35 +307,11 @@ createUserNoVerify uData = do
   let uid = userId usr
   let eac = createdEmailActivation result
   let pac = createdPhoneActivation result
-
-  if isNewUserInvitedViaScim uData
```
smatting: @fisx Removed sending activation email here. Should this be added somewhere else now?

services/brig/src/Brig/API/User.hs
```diff
@@ -180,13 +181,21 @@ createUser new@NewUser {..} = do
   for_ (catMaybes [emKey, phKey]) $ verifyUniquenessAndCheckBlacklist
   -- team user registration
   (newTeam, teamInvitation, tid) <- handleTeam (newUserTeam new) emKey
+
+  -- user was created via scim
+  whenJust newUserUUID $ \uid ->
+    when (isJust teamInvitation) $ do
+      mbAccount <- lift (lookupAccount (Id uid))
+      whenJust mbAccount $ \account ->
+        removeTTL account
```
fisx, line 190:
```suggestion
      whenJust mbAccount removeTTL
```

services/integration.sh
```diff
@@ -163,7 +163,7 @@ while [ "$all_services_are_up" == "" ]; do
 done
 echo "all services are up!"
 
-( ${EXE} "${@:3}" && echo 0 > "${EXIT_STATUS_LOCATION}" && kill_gracefully ) || kill_gracefully &
+( ${EXE} "${@:2}" && echo 0 > "${EXIT_STATUS_LOCATION}" && kill_gracefully ) || kill_gracefully &
```
fisx: now i'm confused.  shouldn't this actually be 4 rather than the 3 i put there?  since we are getting two new arguments sometimes, namely `-b` and `../brig/brig.integration.yaml`?

services/brig/src/Brig/API/User.hs
```diff
@@ -359,7 +359,7 @@ createUserInviteViaScim (NewUserScimInvitation uid tid loc name rawEmail) = (`ca
   account <- lift $ newAccountInviteViaScim uid tid loc name email
   Log.debug $ field "user" (toByteString . userId . accountUser $ account) . field "action" (Log.val "User.createUserInviteViaScim")
   ttl <- round . setTeamInvitationTimeout <$> view settings
-  lift $ Data.insertAccount account Nothing Nothing False (Just ttl)
+  lift $ Data.insertAccount account Nothing Nothing True (Just ttl)
```
fisx, line 362: no.

activated flag has to be false, otherwise we are risking to see it in all kinds of places where we shouldn't.  what does "otherwise on id" mean?

smatting:
@fisx I've figured out the root cause of the `cql-io` error. The first run of the integration creates a user with a TTL which is not 
fully cleaned up after it expires. This leaves a a row `user` in which almost all columns null, breaking the indexing function.
The column which is not null is `handle`. In our flow it is set `Brig.setBrigUserHandle` _after_ the user is created  `Brig.createBrigUser` with a TTL. Apparently you can set a TTL only on non-null values (which makes somewhat sense to me now).

One solution would be to set the TTL on the user at the very end of the flow. However it might be hard to prove that no column will be set from `null` to a non-null value after the flow.

fisx (commented):
I played around with the TTL thing a little more, and I'm beginning to feel that it's better to go for the alternative approach in https://github.com/wireapp/wire-server/pull/1213#issuecomment-709509857, namely to check for invitation on every account with status PendingInvitation produced by a lookup, and if we find such an account, delete it then and there.

fisx:
So, I reverted the two commits we may not want to use (just for the moment, @smatting feel free to fight me on this!), and got back to the `noIdentity` error:

```
       Response {responseStatus = Status {statusCode = 500, statusMessage = "Internal Server Error"}, responseVersion = HTTP/1.1, responseHeaders = [("Transfer-Encoding","chunked"),("Date","Mon, 19 Oct 2020 18:20:56 GMT"),("Server","Warp/3.3.13"),("Content-Type","application/json")], responseBody = Just "{\"code\":500,\"message\":\"{\\\"status\\\":\\\"500\\\",\\\"schemas\\\":[\\\"urn:ietf:params:scim:api:messages:2.0:Error\\\"],\\\"detail\\\":\\\"{\\\\\\\"code\\\\\\\":403,\\\\\\\"message\\\\\\\":\\\\\\\"The user has no verified identity (email or phone number).\\\\\\\",\\\\\\\"label\\\\\\\":\\\\\\\"no-identity\\\\\\\"}\\\"}\",\"label\":\"server-error\"}", responseCookieJar = CJ {expose = []}, responseClose' = ResponseClose}                                          
```

and I found that there are a few more places where this is called, because you can throw error types that *translate* to `noIdentity`, like, here:

```
services/brig/src/Brig/API/Error.hs:70:connError ConnectNoIdentity {} = StdError noIdentity
services/brig/src/Brig/API/Error.hs:126:changePwError ChangePasswordNoIdentity = StdError noIdentity
services/brig/src/Brig/API/Error.hs:130:changeHandleError ChangeHandleNoIdentity = StdError noIdentity
services/brig/src/Brig/API/Error.hs:183:idtError NoIdentity = StdError noIdentity
```

If you grep for `NoIdentity\b`, which is a substring of all the constructors translated from here, [this](https://github.com/wireapp/wire-server/blob/84ca176fb79e498db50eb9097dcb322ceec91e82/services/brig/src/Brig/API/User.hs#L368) is the most promising match.

Which makes sense: [here](https://github.com/wireapp/wire-server/blob/84ca176fb79e498db50eb9097dcb322ceec91e82/services/spar/src/Spar/Scim/User.hs#L314), we create a user with `activated=false`.  And [here](https://github.com/wireapp/wire-server/blob/84ca176fb79e498db50eb9097dcb322ceec91e82/services/spar/src/Spar/Scim/User.hs#L320), we try to change its handle.  But that user won't show in the lookup in the `changeHandle` function (linked above).

Not sure what to do about this.  It seems like if we stick with `activated=false`, there are a few more places where we need to find them where they are currently invisible.  Or we set `activated=true` after all.  Not sure.

services/brig/src/Brig/API/User.hs
```diff
@@ -258,6 +272,12 @@ createUser new@NewUser {..} = do
     handleTeam (Just (NewTeamCreator t)) _ = (Just t,Nothing,) <$> (Just . Id <$> liftIO nextRandom)
     handleTeam (Just (NewTeamMemberSSO tid)) _ = pure (Nothing, Nothing, Just tid)
     handleTeam Nothing _ = return (Nothing, Nothing, Nothing)
+
+    removeTTL account = do
+      let pw = Nothing
+      let activated = False
+      lift $ Data.insertAccount account Nothing pw activated (Just 0)
```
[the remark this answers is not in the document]
smatting: I think I had doubts if `insertAccount` would overwrite everything, but during pairing I think we figured out that it does?

Reading back the entire row and writing it would not solve the problems with nulls anyway:
1. read whole row for which the `handle` is null and write it back with a TTL. Since `handle` is null on the write no TTL is set for this column.
2. set handle to non-null without TTL
3. the row will not expire

services/integration.sh
smatting: I stumbled accross while debugging argument passing to `spar`.

`${@:3}` contains the args starting from the third element, which in our case is (`echo ${EXE} "${@:3}"` )
```
../../dist/spar-integration spar.integration.yaml -i ../integration.yaml -b ../brig/brig.integration.yaml
```
Note how the `-s` parameter is dropped.
This however doesn't result in incorrect arguments to `spar-integration`, because it doesn't feed the arguments verbatim to the argument parser, but splits them (see `services/spar/test-integration/Spec.hs:55`):
```
wireArgs: ["-i","../integration.yaml","-b","../brig/brig.integration.yaml"]
hspecArgs: ["spar.integration.yaml"]
```

With `${@:2}` all arguments are passed to `spar-integration`
```
../../dist/spar-integration -s spar.integration.yaml -i ../integration.yaml -b ../brig/brig.integration.yaml
```
resulting in
```
wireArgs["-s","spar.integration.yaml","-i","../integration.yaml","-b","../brig/brig.integration.yaml"]
hspecArgs[]
```

Both version (`2` and `3`) seem to work (at least for `spar-integration`) .

We should probably revert this commit, because this might break all other integration test.
But I really want to understand what's going on here.

services/brig/src/Brig/API/User.hs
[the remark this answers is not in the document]
smatting: resolved, as we decided not to use TTL anymore to clear pending users

services/brig/src/Brig/API/Internal.hs
```diff
@@ -334,20 +336,39 @@ changeSelfEmailMaybeSend u DoNotSendEmail email = do
...
+      us <- mapM (API.lookupHandle) (fromList hs)
+      byIds (catMaybes us)
   where
-    byIds uids =
-      filter (isJust . userIdentity . accountUser)
-        <$> API.lookupAccounts uids
+    byIds :: [UserId] -> AppIO [UserAccount]
+    byIds uids = catMaybes <$> (API.lookupAccounts uids >>= traverse accountValid)
+
+    accountValid :: UserAccount -> AppIO (Maybe UserAccount)
```
fisx, line 354:
```suggestion
    byIds uids = API.lookupAccounts uids >>= filterM accountValid

    accountValid :: UserAccount -> AppIO Bool
```

services/brig/src/Brig/API/Internal.hs
```diff
@@ -334,20 +336,39 @@ changeSelfEmailMaybeSend u DoNotSendEmail email = do
...
+    accountValid account = case (userIdentity . accountUser $ account) of
+      Nothing -> pure Nothing
+      Just ident
+        | accountStatus account == PendingInvitation ->
+          if includePendingInvitations
+            then case emailIdentity ident of
+              Nothing -> pure (Just account) -- should not happend, scim-invited user always have an email
```
fisx, line 361:
```suggestion
              Nothing -> pure (Just account)
```

This can happen, because this function is called in situations other then scim-invite.

services/brig/src/Brig/API/Internal.hs
```diff
@@ -334,20 +336,39 @@ changeSelfEmailMaybeSend u DoNotSendEmail email = do
...
+      byIds (catMaybes us)
   where
-    byIds uids =
-      filter (isJust . userIdentity . accountUser)
-        <$> API.lookupAccounts uids
+    byIds :: [UserId] -> AppIO [UserAccount]
+    byIds uids = catMaybes <$> (API.lookupAccounts uids >>= traverse accountValid)
+
+    accountValid :: UserAccount -> AppIO (Maybe UserAccount)
+    accountValid account = case (userIdentity . accountUser $ account) of
```
fisx, line 355:
```suggestion
    accountValid account = case userIdentity . accountUser $ account of
```

services/brig/src/Brig/API/Internal.hs
```diff
@@ -334,20 +336,39 @@ changeSelfEmailMaybeSend u DoNotSendEmail email = do
...
+      Nothing -> pure Nothing
+      Just ident
+        | accountStatus account == PendingInvitation ->
```
fisx, line 358: If `if` works just as well, I prefer it over guards, since it is more restrictive and thus more robust.  [feel free to ignore this suggestion.]

services/spar/test-integration/Test/Spar/Scim/UserSpec.hs
smatting: You get 400 with a body of
```
ponse {responseStatus = Status {statusCode = 400, statusMessage = "Bad Request"}, responseVersion = HTTP/1.1, responseHeaders = [("Transfer-Encoding","chunked"),("Date","Tue, 20 Oct 2020 16:24:30 GMT"),("Server","Warp/3.3.13"),("Content-Encoding","gzip"),("Content-Type","application/json")], responseBody = Just "{\"code\":400,\"message\":\"Invalid invitation code.\",\"label\":\"invalid-invitation-code\"}", responseCookieJar = CJ {expose = []}, responseClose' = ResponseClose}
```
The error itself is defined to be 403
```
invalidCode = Wai.Error status403 "invalid-code" "Invalid verification code"
```
but his is as far as I got

services/brig/src/Brig/API/Internal.hs
smatting: thanks :)

services/brig/src/Brig/API/Internal.hs
smatting: This case has additional `accountStatus account == PendingInvitation` condition

services/brig/src/Brig/API/Internal.hs
smatting: not irgnored, makes sense :)

services/spar/test-integration/Test/Spar/Scim/UserSpec.hs
smatting: Ah I think I've understood your verbal test spec wrong here:

> "sending the same post /scim/Users request a second time will have the same effect as the first time"

I'll update the test

services/brig/src/Brig/API/Internal.hs
fisx: oops, you're right.  never mind.

services/brig/src/Brig/Team/API.hs
```diff
@@ -154,6 +156,19 @@ routesPublic = do
...
+    accept "application" "json"
+      .&. query "email"
+
+  -- TODO: check if this correct
+  document "GET" "getInvitationByEmail" $ do
+    Doc.summary "Get a pending invitation for a given an email address."
+    Doc.parameter Doc.Query "email" Doc.bytes' $
+      Doc.description "Email address"
+    Doc.returns (Doc.ref Public.modelTeamInvitation)
+    Doc.response 404 "No pending invitations exists." Doc.end
```
fisx, line 169:
```suggestion
    Doc.response 200 "Pending invitation exists." Doc.end
    Doc.response 404 "No pending invitations exists." Doc.end
```

Looks right to me!  I tried to make a suggestion for the entire block to remove the TODO as well, but github didn't figure out a good diff, so I'm making two separate suggestions.

services/brig/src/Brig/Team/API.hs
```diff
@@ -154,6 +156,19 @@ routesPublic = do
     Doc.response 404 "No pending invitations exists." Doc.end
     Doc.response 409 "Multiple conflicting invitations to different teams exists." Doc.end
 
+  get "/teams/invitations/by-email" (continue getInvitationByEmailH) $
+    accept "application" "json"
+      .&. query "email"
+
+  -- TODO: check if this correct
```
fisx, line 163:
```suggestion
```

services/spar/spar.cabal
```diff
@@ -1,10 +1,10 @@
 cabal-version: 1.12
 
--- This file has been generated from package.yaml by hpack version 0.33.0.
+-- This file has been generated from package.yaml by hpack version 0.34.2.
```
fisx, line 3: This *shouldn't* cause any problems, but if it does, you need to find the right stack version to match the one on our other systems (concourse, mostly, I think?).

But I think it's fine.

services/spar/test-integration/Test/Spar/Scim/UserSpec.hs
fisx: Ok, sorry, you did follow the spec correctly.  But I still like what I *meant* more than what I actually wrote.  :)

services/spar/src/Spar/Scim/User.hs
```diff
@@ -336,12 +360,13 @@ createValidScimUser ScimTokenInfo {stiTeam} vsu@(ST.ValidScimUser veid handl mbN
...
   -- If applicable, trigger email validation procedure on brig.
-  lift $ validateEmailIfExists buid veid
+  lift $ ST.runValidExternalId (validateEmailIfExists buid) (\_ -> pure ()) veid
 
   -- {suspension via scim: if we don't reach the following line, the user will be active.}
   lift $ do
     old <- Brig.getStatus buid
-    let new = ST.scimActiveFlagToAccountStatus old (Just active)
+    let new = ST.scimActiveFlagToAccountStatus old active
+        active = Scim.active . Scim.value . Scim.thing $ storedUser
```
fisx, line 369: explanation for the reviewer: `active` used to come from a function argument, but in the flow implemented here, brig will change that value, and we need to listen to brig, not the scim peer who is creating the user.

services/spar/src/Spar/Scim/User.hs
```diff
@@ -664,9 +684,13 @@ scimFindUserByHandle mIdpConfig stiTeam hndl = do
...
 scimFindUserByEmail :: Maybe IdP -> TeamId -> Text -> MaybeT (Scim.ScimHandler Spar) (Scim.StoredUser ST.SparTag)
 scimFindUserByEmail mIdpConfig stiTeam email = do
+  lift $ lift $ Log.debug (Log.msg $ "scimFindUserByEmail 1: " <> (show (stiTeam, email)))
   veid <- mkValidExternalId mIdpConfig (pure email)
+  lift $ lift $ Log.debug (Log.msg $ "scimFindUserByEmail 2: " <> (show veid))
   uid <- MaybeT . lift $ ST.runValidExternalId withUref withEmailOnly veid
-  brigUser <- MaybeT . lift . Brig.getBrigUserAccount $ uid
+  lift $ lift $ Log.debug (Log.msg $ "scimFindUserByEmail 3: " <> (show uid))
+  brigUser <- MaybeT . lift . Brig.getBrigUserAccount Brig.WithPendingInvitations $ uid
+  lift $ lift $ Log.debug (Log.msg $ "scimFindUserByEmail 4: " <> (show brigUser))
```
fisx, line 693: is this all too much debug information?  (i'm not sure)

fisx:
@jschaul this is kinda ready for review.  we have one failing integration test, which involves ES search, @smatting is investigating.  But that should be a small change.

services/brig/src/Brig/Team/API.hs
```diff
@@ -154,6 +156,18 @@ routesPublic = do
     Doc.response 404 "No pending invitations exists." Doc.end
     Doc.response 409 "Multiple conflicting invitations to different teams exists." Doc.end
 
+  get "/teams/invitations/by-email" (continue getInvitationByEmailH) $
```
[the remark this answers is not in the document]
jschaul: This is a security problem: anyone can register as if they were that user if this endpoint is public, all they need to know is an email address. Please make internal and hide behind `/i/`.

services/brig/src/Brig/User/Auth.hs
```diff
@@ -114,10 +114,11 @@ login (PasswordLogin li pw label) typ = do
   Log.debug $ field "user" (toByteString uid) . field "action" (Log.val "User.login")
   checkRetryLimit uid
   Data.authenticate uid pw `catchE` \case
+    AuthInvalidUser -> loginFailed uid
+    AuthInvalidCredentials -> loginFailed uid
     AuthSuspended -> throwE LoginSuspended
     AuthEphemeral -> throwE LoginEphemeral
-    AuthInvalidCredentials -> loginFailed uid
-    AuthInvalidUser -> loginFailed uid
+    AuthPendingInvitation -> loginFailed uid
```
jschaul, line 121: inconsistent errors: 

line 121: AuthPendingInvitation -> loginFailed uid
line 304: AuthPendingInvitation -> throwE LoginPendingActivation

services/brig/src/Brig/User/Auth.hs
```diff
@@ -296,10 +297,11 @@ ssoLogin (SsoLogin uid label) typ = do
   Data.reauthenticate uid Nothing `catchE` \case
     ReAuthMissingPassword -> pure ()
     ReAuthError e -> case e of
+      AuthInvalidUser -> throwE LoginFailed
       AuthInvalidCredentials -> pure ()
       AuthSuspended -> throwE LoginSuspended
       AuthEphemeral -> throwE LoginEphemeral
-      AuthInvalidUser -> throwE LoginFailed
+      AuthPendingInvitation -> throwE LoginPendingActivation
```
jschaul, line 304: inconsistent errors: 

line 121: AuthPendingInvitation -> loginFailed uid
line 304: AuthPendingInvitation -> throwE LoginPendingActivation

services/spar/Makefile
```diff
@@ -66,18 +66,18 @@ $(DEB_SCHEMA): install
...
+	INTEGRATION_USE_REAL_AWS=1 ../integration.sh $(EXE_IT) -s $(NAME).integration-aws.yaml -i ../integration.yaml -b ../brig/brig.integration.yaml  $(WIRE_INTEGRATION_TEST_OPTIONS)
 
 .PHONY: i-list
 i-list:
-	$(EXE_IT) -s $(NAME).integration.yaml -i ../integration.yaml -l
+	$(EXE_IT) -s $(NAME).integration.yaml -i ../integration.yaml -b ../brig/brig.integration.yaml  -l
 
 i-%:
-	../integration.sh $(EXE_IT) -s $(NAME).integration.yaml -i ../integration.yaml --match="$*" $(WIRE_INTEGRATION_TEST_OPTIONS)
+	../integration.sh $(EXE_IT) -s $(NAME).integration.yaml -i ../integration.yaml -b ../brig/brig.integration.yaml  --match="$*" $(WIRE_INTEGRATION_TEST_OPTIONS)
```
jschaul, line 80: This changeset doesn't look like it belongs into this PR; please extract into separate PR.

services/spar/package.yaml
```diff
@@ -131,6 +131,7 @@ executables:
     dependencies:
       - aeson-qq
       - async
+      - brig
```
jschaul, line 134: hm, this is not nice to depend on another service entirely. Could you extract whichever functions and types you need into `types-common`, or `brig-types`, or `wire-api`, then remove this depdency?

services/spar/spar.cabal
jschaul: Hm, this is not ideal - we can upgrade stack, but then we should upgrade everywhere at the same time, and also update CI (best done in a separate PR). Could you revert your version of stack to  `2.3.1` and re-compile spar / re-generate this file?

services/brig/src/Brig/API/Internal.hs
```diff
@@ -334,20 +337,42 @@ changeSelfEmailMaybeSend u DoNotSendEmail email = do
...
+    Left us -> byIds (fromList us)
+    Right hs -> do
+      us <- mapM (API.lookupHandle) (fromList hs)
+      byIds (catMaybes us)
   where
-    byIds uids =
-      filter (isJust . userIdentity . accountUser)
-        <$> API.lookupAccounts uids
+    byIds :: [UserId] -> AppIO [UserAccount]
+    byIds uids = API.lookupAccounts uids >>= filterM accountValid
```
jschaul, line 354: This still doesn't  work; API.lookupAccounts only returns user accounts with activated=true. You want some accounts that are activated=false, so this logic isn't right, you need to use another (or write a new function), or adjust the existing lookupAccounts function.

jschaul (commented):
I see no brig integration tests. Please add tests for the new endpoints in brig:

* GET /i/users?ids=ABC&includePendingInvitations=true / false
* the new invitation create/get endpoints introduced in brig.

services/brig/src/Brig/Team/API.hs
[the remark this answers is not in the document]
smatting: Done

services/spar/src/Spar/Scim/User.hs
smatting: Yes i think so, I removed it.

services/spar/package.yaml
smatting: I removed the dependency (see comment on Makefile)

services/spar/Makefile
smatting: Removed it. Parsing the brig options also introduced the dependency on brig.

services/spar/spar.cabal
smatting: I reverted to stack 2.3.1 and regenerated it. (Don't know if it's relevant, but I noticed stack provided by the `shell.nix` in the root of `wire-server` is `2.3.3`.)

smatting:
I've removed the `-b brig.integration.yaml` from spar-integration in https://github.com/wireapp/wire-server/pull/1213/commits/a78843dd71307e7bd16117dd3f55c7c2734bbcf0
Since we need the config value of `setTeamInvitationTimeout` from `brig.integration.yaml` in the integration tests in yaml, one option would be to mirror this in the spar config:

`spar.integration.yaml`  @ develop:
```yaml
...
brig:
  host: 127.0.0.1
  port: 8082

galley:
  host: 127.0.0.1
  port: 8085

cassandra:
  endpoint:
    host: 127.0.0.1
    port: 9042
  keyspace: spar_test
...
```

suggested edit `spar.integration.yaml`:
```yaml
...
brig:
  endpoint:
    host: 127.0.0.1
    port: 8082
  setTeamInvitationTimeout: 5 # NOTE: must be the same value as in config for brig

galley:
  host: 127.0.0.1
  port: 8085

cassandra:
  endpoint:
    host: 127.0.0.1
    port: 9042
  keyspace: spar_test
...
```

But I suspect this change (nesting `endpoint`) probably break multiple of processes CI and deployment, so I'm hesitant to do this.
For now the 5 seconds are hard-coded.

services/brig/src/Brig/API/Internal.hs
```diff
@@ -334,20 +337,42 @@ changeSelfEmailMaybeSend u DoNotSendEmail email = do
...
+                Just email -> do
+                  mbInv <- lookupInvitationByEmail email
+                  if isJust mbInv
+                    then pure True
+                    else do
+                      -- user invited via scim should expire together with its invitation
+                      API.deleteUserNoVerify (userId . accountUser $ account)
+                      pure False
+              else pure False
+          else pure True
```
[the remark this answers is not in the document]
smatting: Done

services/brig/src/Brig/API/Internal.hs
fisx: This is not valid any more: we found that it creates bigger problems to have users with `activated==false && status=PendingActivation`, and changed them to `activated==true && status==PendingActivation`.  So this line should do what it's supposed to.

services/brig/src/Brig/User/Auth.hs
fisx: fe81046e5ccd3cb641868b8369b1911f1ef9b98b

services/brig/src/Brig/User/Auth.hs
fisx: fe81046e5ccd3cb641868b8369b1911f1ef9b98b

services/brig/src/Brig/API/Internal.hs
```diff
@@ -334,20 +337,42 @@ changeSelfEmailMaybeSend u DoNotSendEmail email = do
...
+        case (accountStatus account, includePendingInvitations, emailIdentity ident) of
+          (PendingInvitation, False, _) -> pure False
+          (PendingInvitation, True, Just email) -> do
+            hasInvitation <- isJust <$> lookupInvitationByEmail email
+            if hasInvitation
+              then pure True
+              else do
+                -- user invited via scim should expire together with its invitation
+                API.deleteUserNoVerify (userId . accountUser $ account)
+                pure False
```
fisx, line 369:
```suggestion
            unless hasInvitation $ do
                -- user invited via scim should expire together with its invitation
                API.deleteUserNoVerify (userId . accountUser $ account)
            pure hasInvitation
```

arguably a matter of taste...

services/spar/test-integration/Test/Spar/Scim/UserSpec.hs
```diff
@@ -555,6 +616,57 @@ testScimCreateVsUserRef = do
...
+    searchUser :: HasCallStack => Spar.Types.ScimToken -> Scim.User.User tag -> Email -> Bool -> TestSpar ()
+    searchUser tok scimUser email shouldSucceed = do
+      let handle = Handle . Scim.User.userName $ scimUser
+      listUsers tok (Just (filterBy "userName" $ fromHandle handle)) >>= \users ->
+        liftIO $ length users `shouldSatisfy` if shouldSucceed then (> 0) else (== 0)
+      listUsers tok (Just (filterBy "externalId" $ fromEmail email)) >>= \users ->
+        liftIO $ length users `shouldSatisfy` if shouldSucceed then (> 0) else (== 0)
+
+    waitUserExpiration = do
+      let setTeamInvitationTimeout = 5 -- TODO(stefan): clarify if we should add this to spar's config
```
fisx, line 666:
```suggestion
      -- this should be something like @round . Brig.Options.setTeamInvitationTimeout . Brig.Options.optSettings . 
      -- view teBrigOpts $ env@, but if this goes out of sync with the brig config, we will only get benign false 
      -- negatives, and importing brig options into spar integration tests is just too awkward.
      let setTeamInvitationTimeout = 5
```

smatting (dismissed):

fisx:
> I see no brig integration tests. Please add tests for the new endpoints in brig:
> 
>     * GET /i/users?ids=ABC&includePendingInvitations=true / false
> 
>     * the new invitation create/get endpoints introduced in brig.

We're testing all of this, both happy and sad paths.  It's admittedly hard to find, though: 6b60d7f3741495f7bd27a7790d35f12a73553563

fisx:
I think we're still missing some bits & pieces...

The following database functions in `Brig.Data.User` read table `brig.user`, and thus may get confused about the new `PendingInvitation` status:

- [ ] `lookupAccount`
- [ ] `lookupAccounts`
- [ ] `lookupUser`
- [ ] `lookupUsers`
- [ ] `lookupName`
- [ ] `lookupLocale`
- [ ] `lookupPassword`
- [ ] `lookupStatus`
- [ ] `lookupRichInfo`
- [ ] `lookupUserTeam`
- [ ] `lookupUsersTeam`
- [ ] `lookupServiceUsers`
- [ ] `lookupServiceUsersForTeam`

There is also ES index refresh, but we have that covered.  No other mentions of the table, so with this, we should be ok.

- [ ] will the users in status `PendingInvitation`, and the active users created when scim-triggered invitations are accepted, have `ManagedByWire` set?

fisx:
> will the users in status `PendingInvitation`, and the active users created when scim-triggered invitations are accepted, have ~~`ManagedByWire`~~ `ManagedByScim` set?

just fixed that in my last comment.

fisx:
- `lookupPassword` - all calls are safe (only account deletion, password reset, login).
- `lookupStatus` - this *has* to return `PendingInvitation` for spar to work.  all uses are safe.
- `lookupRichInfo` - different table, not touched through `NewUserScimInvitation`
- `lookupUserTeam` - already safe to use for auth (left a comment)
- `lookupUsersTeam` - dead code
- `lookupServiceUsers` - different c*-table
- `lookupServiceUsersForTeam` - different c*-table

smatting:
- When checking for calls to `lookupAccount` I noticed `isPendingInvitation` [Brig/User/Auth.hs:251](https://github.com/wireapp/wire-server/blob/fisx/scim-onboarding-flow-again/services/brig/src/Brig/User/Auth.hs#L251) would probably yield `False`. It's used in `login` via `resolveLoginId` (all from the Auth module). Is this a hypothetical unhappy flow?:
   1. user is created via scim with password
   2. user does NOT validate email
   3. user tries to login with credentials and receives a wrong `LoginError` value`LoginFailed` when instead she should receive `LoginPendingActivation` value
- When checking for calls to `lookupAccount` I noticed `activateKey` [Brig/User/Auth.hs:251](https://github.com/wireapp/wire-server/blob/fisx/scim-onboarding-flow-again/services/brig/src/Brig/Data/Activation.hs#L87) would probably `throwE invalidCode`. It looks like this function is used in `Brig.API.createUser` via `activateWithCurrency` and `activate`, but from the comments it looks like scim-invited users won't trigger the flow

smatting (approved):

fisx:
> ... It's used in `login` via `resolveLoginId`, which will throw an incorrect exception. Is this a hypothetical unhappy flow?

Not entirely hypothetical, so good catch & good fix!

> ... activateWithCurrency and activate, but from the comments it looks like scim-invited users flow won't call these functions

I agree.

fisx:
CI problems are unrelated; integration tests passed locally.

PR #1215, opened 2020-10-06
add libtinfo/ncurses to docs and nix deps

mheinzel:
proto-lens-protoc switched to ghc-source-gen, which requires libtinfo.

Don't know what the equivalent on macOS is and if it's even needed. Didn't find it with some googling.
Also not sure if the Nix dependency is the right one. Please tell me if it's not right.

### DIFF
--- a/docs/developer/dependencies.md
+++ b/docs/developer/dependencies.md
@@ -178,8 +178,3 @@ Using Stack's [Nix integration](https://docs.haskellstack.org/en/stable/nix_inte
 dependencies automatically - including `cryptobox-c`. If new system dependencies are needed, add them to the `stack-deps.nix` file in the project root.
 Just type `$ nix-shell` and you will automatically have `make`, `docker-compose` and `stack` in `PATH`.
 You can then run all the builds, and the native dependencies will be automatically present.
-
-Nix might only ship a more recent minor version of GHC than the current Stackage snapshot specifies. The releases should be totally compatible, but we need to convince stack that this is true. Hence, we can add the following to our global stack config in `~/.stack/config.yaml`
-```bash
-skip-ghc-check: true
-```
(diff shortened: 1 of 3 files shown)


fisx:
@mheinzel writes:

>  Also, we can probably remove
> > We are currently on a snapshot that uses ghc863 but Nix only ships ghc864 as ghc863 is officially deprecated. The releases should be totally compatible, but we need to convince stack that this is true. Hence, we can add the following to our global stack config in ~/.stack/config.yaml

agreed!  do you want to do that here?

fisx (approved):
With or without [this](https://github.com/wireapp/wire-server/pull/1215#issuecomment-704757687), this PR is good to go I think.

I haven't tried the nix change, but then again, I see it working every day on debian + nix-pkg, and I think we also have it working on nixos (or @arianvp would have complained, no?).

arianvp:
The nix bit doesn't work anymore.  Lets see if bumping nixpkgs fixes it....

mheinzel:
> agreed! do you want to do that here?

I didn't remove it, but made it more general since a similar situation could happen again, right?

Does the Nix setup work now?

docs/developer/dependencies.md
```diff
@@ -175,11 +175,11 @@ docker login --username=<MY_DOCKER_USERNAME>
...
-dependencies automatically - including `cryptobox-c`. If new system dependencies are needed, add them to the `shell.nix` file in the project root.
+dependencies automatically - including `cryptobox-c`. If new system dependencies are needed, add them to the `stack-deps.nix` file in the project root.
 Just type `$ nix-shell` and you will automatically have `make`, `docker-compose` and `stack` in `PATH`.
 You can then run all the builds, and the native dependencies will be automatically present.
 
-We are currently on a snapshot that uses `ghc863` but Nix only ships `ghc864` as `ghc863` is officially deprecated. The releases should be totally compatible, but we need to convince stack that this is true. Hence, we can add the following to our global stack config in `~/.stack/config.yaml`
+Nix might only ship a more recent minor version of GHC than the current Stackage snapshot specifies. The releases should be totally compatible, but we need to convince stack that this is true. Hence, we can add the following to our global stack config in `~/.stack/config.yaml`
```
akshaymankar, line 182: I thought we are completely removing this as we don't really need it.

mheinzel: Okay, then I'll remove it completely.

arianvp (approved):

PR #1218, opened 2020-10-07
Migrate rex to this repository

mheinzel:
See https://github.com/zinfra/backend-issues/issues/1264.

I have not tried running it yet, but it compiles with the current snapshot.


tools/rex/Dockerfile
```diff
@@ -0,0 +1,4 @@
+FROM alpine:3.5
```
jschaul, line 1: let's update this

tools/rex/Makefile
```diff
@@ -0,0 +1,43 @@
...
+default: dist
+
+.PHONY: dist
+dist: clean $(ACI)
+
+.PHONY: build
+build: rex
+
+.PHONY: publish
+publish: $(ACI) $(ACI).asc s3-exists/$(ACI) s3-upload/$(ACI)
```
jschaul, line 20: some of those targets don't exist anymore?

tools/rex/rex.cabal
```diff
@@ -0,0 +1,46 @@
+name:                rex
+version:             0.3.0
+synopsis:            Scrape and expose restund metrics for prometheus
+description:         @rex@ = @restund-exporter@ in prometheus jargon
+author:              Kim Altintop
+maintainer:          kim@wire.com
```
jschaul, line 6: this should be changed to Wire Swiss. Also add a license.

tools/rex/.stylish-haskell.yaml
```diff
@@ -0,0 +1,20 @@
...
+      list_padding: 4
+      separate_lists: true
+  - language_pragmas:
+      style: vertical
+      align: true
+      remove_redundant: true
+  - tabs:
+      spaces: 4
+  - trailing_whitespace: {}
+columns: 80
```
akshaymankar, line 20: Let's remove this and ormolu the code.

akshaymankar (changes requested):

tools/rex/.stylish-haskell.yaml
mheinzel: It's ormulized, but yes, I still need to clean up some files.

tools/rex/Dockerfile
mheinzel: done

tools/rex/rex.cabal
mheinzel: done

tools/rex/.stylish-haskell.yaml
mheinzel: removed this and some more stuff.

tools/rex/Makefile
mheinzel: Yeah, and a few more things need fixing.
- there's an include that doesn't work anymore
- we should probably stack install into `dist` as with the other tools

I'm not quite sure how rex is supposed to be packaged/used, so maybe someone else could take care of these outstanding issues with the Makefile?

mheinzel:
I fixed all of the issues except for the Makefile. Once that's done, this PR should be ready.

> 2 out of 3 committers have signed the CLA.

Kim worked for Wire when writing this code, so we can go ahead without the CLA.

jschaul:
I added two commits here to put the Makefile in order, add a README, and also to compile a static version based on the current GHC/alpine docker image.

jschaul (approved):

jschaul:
@akshaymankar any further concerns?

akshaymankar (approved):
Looks good now.

PR #1219, opened 2020-10-08
Fix compilation in fisx's weird dev env.

fisx:
Without this change, I can't compile brig integration tests in my repl any more.  One hypothesis is that this is because I'm loading modules from `/services/brig/src` directly, rather than loading the compiled package, but I haven't confirmed that.

This seems safe to me: it only affects the integration test executable, and that should already have this dependency transitively anyway.

### DIFF
--- a/services/brig/brig.cabal
+++ b/services/brig/brig.cabal
@@ -4,7 +4,7 @@ cabal-version: 1.12
 --
 -- see: https://github.com/sol/hpack
 --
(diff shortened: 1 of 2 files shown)


akshaymankar:
What error do you get while compiling?

fisx:
What the?  good you asked.  I can't reproduce it any more.

Sorry for the noise.

PR #1222, opened 2020-10-08
update minio image

jschaul:
Relates to https://github.com/zinfra/backend-issues/issues/1659#issuecomment-705712161

(I currently can't locally run integration tests; could someone run those locally for cargohold from this branch? `cd services/cargohold && make integration`)


smatting:
I gave this a shot. I started the dockerephemeral docker compose and got this output:
https://gist.github.com/smatting/4e4975dcb98d99acc678e14775a2ac51

fisx:
i have the same issue.  In fact, I think I've had this issue also on develop for a while.  No time yet to look into it.

mheinzel:
Yes, this was because something was borked with amazonka, but it should be fixed on develop for a while already.

mheinzel (commented):
Rebased, this works fine for me now. The upgrade is not directly needed anymore, but it's probably still a good idea to do it.

The main comment I have is that we should point to an image we built ourselves instead, [as that seems to be the preferred way](https://github.com/wireapp/wire-server/blob/develop/docs/reference/make-docker-and-qemu.md).

mdimjasevic:
I am closing this outdated and old PR. If you think it would be worth it, please reopen the PR and bring it up to date.

PR #1226, opened 2020-10-12
Yet more debug logs

fisx:
This is left-over from https://github.com/zinfra/backend-issues/issues/1797.  Could be useful in the future, and should be harmless for the same reasons as in https://github.com/wireapp/wire-server/pull/1220.

Can be read commit-by-commit.

### DIFF
--- a/services/gundeck/gundeck.cabal
+++ b/services/gundeck/gundeck.cabal
@@ -4,7 +4,7 @@ cabal-version: 1.12
 --
 -- see: https://github.com/sol/hpack
 --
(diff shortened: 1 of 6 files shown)


akshaymankar (commented):
Looks good regarding the logs. I am not sure why we needed to add `HasCallStack` to `prepare`.

fisx:
> I am not sure why we needed to add `HasCallStack` to `prepare`.

That was an accident.  Would you prefer to have it removed?  I'll just remove it.

In fact, I think `prepare` doesn't really have `IO`, so I'll remove that, too.

akshaymankar (approved):
:rocket:

PR #1229, opened 2020-10-20
Better brig errors

fisx:
we should be able to distinguish the handful of places where this error is raised [WIP]


akshaymankar:
Can we use a better type than `Int` please? I'd create a special sum type, but even strings are better.

akshaymankar (approved):
Since, this is better than nothing and we don't want to spend time. It is ok to merge.

PR #1234, opened 2020-10-21
cargohold: add compatibility mode for Scality RING S3 implementation

mheinzel:
See https://github.com/zinfra/backend-issues/issues/1659

Not the nicest way to do it and doesn't adhere to the S3 spec, but at least this way it's behind an optional setting.

### DIFF
--- a/services/cargohold/src/CargoHold/Options.hs
+++ b/services/cargohold/src/CargoHold/Options.hs
@@ -21,6 +21,7 @@ module CargoHold.Options where
 
 import CargoHold.CloudFront (Domain (..), KeyPairId (..))
 import Control.Lens hiding (Level)
+import Data.Aeson (FromJSON (..), withText)
 import Data.Aeson.TH
 import Imports
 import System.Logger.Extended (Level, LogFormat)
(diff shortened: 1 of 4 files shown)


jschaul (approved):
I can confirm this works against the different compatibility layers (aws, minio, scality-ring).

services/cargohold/src/CargoHold/Options.hs
```diff
@@ -51,11 +52,24 @@ data AWSOpts = AWSOpts
     _awsS3DownloadEndpoint :: !(Maybe AWSEndpoint),
     -- | S3 bucket name
     _awsS3Bucket :: !Text,
+    -- | Enable this option for compatibility with specific S3 backends.
+    _awsS3Compatibility :: !(Maybe (Last S3Compatibility)),
```
jschaul, line 56: Just wondering: Why wrap this in `Last` ?

jschaul:
(We also need to expose this optional config flag in wire-server-deploy)

services/cargohold/src/CargoHold/Options.hs
mheinzel: Ah, yeah, I saw it used in other options and was wondering about it as well. Forgot to remove it, but it's done now and apparently didn't have any deeper reason.

mheinzel:
> (We also need to expose this optional config flag in wire-server-deploy)

https://github.com/wireapp/wire-server-deploy/pull/364

PR #1238, opened 2020-10-27
Avoid lazily garbage-collected rows in brig.user table.

fisx:
Fixes https://github.com/zinfra/backend-issues/issues/1847

Don't lazily remove `brig.user` records if in status `PendingInvitation` with `Invitation` expired.  Instead, give them TTL values corresponding to the ones for the `Invitations`.

This is slightly complicated by the fact that we're not doing a single, clean insert, but updating the record for handle, and touch a second table for rich info.

### DIFF
--- a/services/brig/src/Brig/API/User.hs
+++ b/services/brig/src/Brig/API/User.hs
@@ -437,6 +437,7 @@ changeHandleWith mbTTL uid mconn hdl = do
     Nothing -> throwE ChangeHandleNoIdentity
     Just u -> claim u
   where
+    claim :: User -> ExceptT ChangeHandleError AppIO ()
     claim u = do
       unless (isJust (userIdentity u)) $
         throwE ChangeHandleNoIdentity
--- a/services/brig/src/Brig/API/User.hs
+++ b/services/brig/src/Brig/API/User.hs
@@ -353,7 +353,9 @@ makeUserPermanent muid teamInvitation = do
   case (muid, teamInvitation) of
     (Just (Id -> uid), Just _) -> do
       mbAccount <- Data.lookupAccount uid
-      (\account -> Data.insertAccount account Nothing Nothing True) `mapM_` mbAccount
+      for_ mbAccount $ \account -> do
+        Data.insertAccount account Nothing Nothing True
+        makeHandlePermanent `mapM_` userHandle (accountUser account)
       mbRichInfo <- Data.lookupRichInfo uid
       Data.updateRichInfo uid `mapM_` mbRichInfo
     _ -> pure ()
--- a/services/brig/src/Brig/API/User.hs
+++ b/services/brig/src/Brig/API/User.hs
@@ -423,10 +423,13 @@ changeManagedBy uid conn (ManagedByUpdate mb) = do
 -- Change Handle
 
 changeHandle :: UserId -> Maybe ConnId -> Handle -> ExceptT ChangeHandleError AppIO ()
-changeHandle = changeHandleWithTTL 0
+changeHandle = changeHandleWith Nothing
 
 changeHandleWithTTL :: Int32 -> UserId -> Maybe ConnId -> Handle -> ExceptT ChangeHandleError AppIO ()
-changeHandleWithTTL ttl uid mconn hdl = do
+changeHandleWithTTL = changeHandleWith . Just
+
+changeHandleWith :: Maybe Int32 -> UserId -> Maybe ConnId -> Handle -> ExceptT ChangeHandleError AppIO ()
+changeHandleWith mbTTL uid mconn hdl = do
   when (isBlacklistedHandle hdl) $
     throwE ChangeHandleInvalid
   usr <- lift $ Data.lookupUser WithPendingInvitations uid
@@ -437,7 +440,7 @@ changeHandleWithTTL ttl uid mconn hdl = do
     claim u = do
       unless (isJust (userIdentity u)) $
         throwE ChangeHandleNoIdentity
-      claimed <- lift $ claimHandleWithTTL ttl (userId u) (userHandle u) hdl
+      claimed <- lift $ claimHandleWithTTL mbTTL (userId u) (userHandle u) hdl
       unless claimed $
         throwE ChangeHandleExists
       lift $ Intra.onUserEvent uid mconn (handleUpdated uid hdl)
--- a/services/brig/src/Brig/API/User.hs
+++ b/services/brig/src/Brig/API/User.hs
@@ -126,7 +126,7 @@ import Control.Lens (view, (^.))
 import Control.Monad.Catch
 import Data.ByteString.Conversion
 import qualified Data.Currency as Currency
-import Data.Handle (Handle)
+import Data.Handle (Handle, fromHandle)
 import Data.Id as Id
 import Data.IdMapping (MappedOrLocalId, partitionMappedOrLocalIds)
 import Data.Json.Util
@@ -362,10 +362,11 @@ makeUserPermanent muid teamInvitation = do
 -- all over the place there, we add a new function that handles just the one new flow where
 -- users are invited to the team via scim.
 createUserInviteViaScim :: UserId -> NewUserScimInvitation -> ExceptT Error.Error AppIO UserAccount
-createUserInviteViaScim uid (NewUserScimInvitation tid loc name handl rawEmail richinfo) = (`catchE` (throwE . Error.newUserError)) $ do
-  email <- either (throwE . InvalidEmail rawEmail) pure (validateEmail rawEmail)
+createUserInviteViaScim uid (NewUserScimInvitation tid loc name rawHandl rawEmail richinfo) = do
+  email <- either (throwE . Error.newUserError . InvalidEmail rawEmail) pure (validateEmail rawEmail)
+  handl <- validateHandle (fromHandle rawHandl)
   let emKey = userEmailKey email
-  verifyUniquenessAndCheckBlacklist emKey
+  verifyUniquenessAndCheckBlacklist emKey `catchE` (throwE . Error.newUserError)
   account <- lift $ Data.newAccountInviteViaScim uid tid loc name email
   Log.debug $ field "user" (toByteString . userId . accountUser $ account) . field "action" (Log.val "User.createUserInviteViaScim")
   let activated =
@@ -375,10 +376,9 @@ createUserInviteViaScim uid (NewUserScimInvitation tid loc name handl rawEmail r
         -- the SCIM user.
         True
   ttl <- round . setTeamInvitationTimeout <$> view settings
-  lift $ do
-    Data.insertAccountWithTTL ttl account Nothing Nothing activated
-    Data.updateHandleWithTTL ttl uid handl
-    Data.updateRichInfoWithTTL ttl uid (RI.unRichInfo richinfo)
+  lift $ Data.insertAccountWithTTL ttl account Nothing Nothing activated
+  changeHandleWithTTL ttl uid Nothing handl `catchE` (throwE . Error.changeHandleError)
+  lift $ Data.updateRichInfoWithTTL ttl uid (RI.unRichInfo richinfo)
   return account
 
 -- | docs/reference/user/registration.md {#RefRestrictRegistration}.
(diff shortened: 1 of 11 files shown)


fisx:
Different error, but still an error:

```
Failures:

  test-integration/Test/Spar/Scim/UserSpec.hs:661:43: 
  1) Spar.Scim.User, POST /Users, team has no SAML IdP, doesn't list users that exceed their invivtation period, and allows recreating them
       uncaught exception: ErrorCall
       Assertions failed:
        1: 201 =/= 500
       
       Response was:
       
       Response {responseStatus = Status {statusCode = 500, statusMessage = "Internal Server Error"}, responseVersion = HTTP/1.1, responseHeaders = [("Transfer-Encoding","chunked"),("Date","Tue, 27 Oct 2020 20:07:50 GMT"),("Server","Warp/3.3.13"),("Content-Type","application/json")], responseBody = Just "{\"code\":500,\"message\":\"{\\\"status\\\":\\\"500\\\",\\\"schemas\\\":[\\\"urn:ietf:params:scim:api:messages:2.0:Error\\\"],\\\"detail\\\":\\\"{\\\\\\\"code\\\\\\\":409,\\\\\\\"message\\\\\\\":\\\\\\\"The given handle is already taken.\\\\\\\",\\\\\\\"label\\\\\\\":\\\\\\\"handle-exists\\\\\\\"}\\\"}\",\"label\":\"server-error\"}", responseCookieJar = CJ {expose = []}, responseClose' = ResponseClose}
         error, called at src/Bilge/Assert.hs:88:5 in bilge-0.22.0-Icg0fwwA3UCTqNbArtIGD:Bilge.Assert
         <!!, called at test-integration/Util/Scim.hs:191:5 in main:Util.Scim
         createUser, called at test-integration/Test/Spar/Scim/UserSpec.hs:661:43 in main:Test.Spar.Scim.UserSpec

  To rerun use: --match "/Spar.Scim.User/POST /Users/team has no SAML IdP/doesn't list users that exceed their invivtation period, and allows recreating them/"

Randomized with seed 138940367

Finished in 7.4317 seconds
1 example, 1 failure
```

fisx:
Made some more progress, but it's still not working.  I left some ideas in the last commit.

fisx:
do we need to make sure the invitation expiry time is 5secs on the ci, too?

services/brig/src/Brig/API/User.hs
```diff
@@ -346,23 +348,37 @@ createUser new@NewUser {..} = do
...
-  account <- lift $ newAccountInviteViaScim uid tid loc name email
+  verifyUniquenessAndCheckBlacklist emKey `catchE` (throwE . Error.newUserError)
+  account <- lift $ Data.newAccountInviteViaScim uid tid loc name email
   Log.debug $ field "user" (toByteString . userId . accountUser $ account) . field "action" (Log.val "User.createUserInviteViaScim")
   let activated =
         -- It would be nice to set this to 'False' to make sure we're not accidentally
         -- treating 'PendingActivation' as 'Active', but then 'Brig.Data.User.toIdentity'
         -- would not produce an identity, and so we won't have the email address to construct
         -- the SCIM user.
         True
```
jschaul, line 377: As we discussed a few weeks ago, setting this to true is problematic for other kinds of queries. Can you not simply write a slightly different accountLookup query to just extract the pendingInvitation users which doesn't filter on activated?

fisx: That would have been an option, but I'm not sure it would result in better code structure.  As it is now, we went through all calls to the existing lookup functions transitively, and found several places where we want to handle pending invitations as well (eg. update, delete).

I'm pretty confident about the current solution.  How badly to you want to fight this battle?  Would it make sense to have a call and discuss it for half an hour or so, looking at the code?

jschaul: There are some things possible now that shouldn't be possible: I can search for, and create a connection request with a person who is "pendingInvitation". But I'm not sure the person joining can then see this connection request after joining. I can also easily create hundreds of accounts that are active without ever validating any emails, so this is prone to spam. There might be some other problematic cases here.

fisx: > I can search for, and create a connection request with a person who is "pendingInvitation".

What's your evidence?  I think we wrote a test case that makes sure this is not the case.

> Note that the current PR attempts to use a delete-account-via-TTL. But an account creates more things than just one row in the users table.

Did you read this PR?  (Not just the first few commits, I mean.  The whole thing.)

jschaul: I admit I didn't look all your code changes; perhaps I don't need to be worried and all the edge cases are covered. Sorry, it's not helpful for me to make comments without having looked in detail. So I'll either stop reviewing this PR, or review it in-depth before adding more comments. With my attention elsewhere at the moment, I think I'll trust you to do the right thing :)

smatting:
Integration tests seem to fail because of indexing.
  
  https://bridge-ie-concourse.zinfra.io/builds/50595#L5f61e587:2788
  
  The test that fails is define here
  `services/brig/test/integration/API/Search.hs:57`
  which has a call to brig's `/i/index/refresh` that fails with 500.

  I think it fails because some other test created a user that is not fully cleaned up.
  The handler of `/i/index/refresh` reads users from the DB.
  When it tries to read the no-fully-cleaned-up user, which has null columns where the data model says it shouldn't have, it crashes.

jschaul:
> Integration tests seem to fail because of indexing.
> 
> https://bridge-ie-concourse.zinfra.io/builds/50595#L5f61e587:2788
> 
> The test that fails is define here
> `services/brig/test/integration/API/Search.hs:57`
> which has a call to brig's `/i/index/refresh` that fails with 500.
> 
> I think it fails because some other test created a user that is not fully cleaned up.
> The handler of `/i/index/refresh` reads users from the DB.
> When it tries to read the no-fully-cleaned-up user, which has null columns where the data model says it shouldn't have, it crashes.

When inserting something with a TTL to cassandra, this TTL will affect both the entire row, as well as all the inserted columns. So if you wish to remove a TTL, all fields need to be set again with a new insert with a TTL of 0. This is potentially error-prone (I suppose as this indexing test shows), as it can regress the next time someone adapt the user table and adds a new field. The TTL also does not allow to remove user information from elasticsearch (not sure if you're writing to elasticsearch upon user creation at the moment? It may happen as part of the insertion of the user to the DB). Personally, I'd prefer an approach of using a separate table to store the users that have not yet accepted their invite, so there is only one table with TTLs and not a mixture of TTLed and non-TTLed users in the main user table, as it can break certain functionality more fundamentally (like not allowing a re-indexing at all). I think it would make the code clearer and the database schema easier to grasp as well (fewer boolean/optional/default values).
Just a suggestion as this PR popped up in my inbox again and it's hard to not read it :)

fisx:
>  Just a suggestion as this PR popped up in my inbox again and it's hard to not read it :)

suggestion appreciated.  :)

Even though it potentially invalidates this PR, and makes it necessary to write a bigger one.  I think we should have involved you in this change a lot earlier on, but perhaps it's still good to have the discussion now.  We did try to make the right call, and we did rule out all the alternatives you suggest for what we thought were good reasons at the time.  I'll schedule a meeting to go over this one more time.

fisx:
Let's start from the beginning.

We need to store some data on users with an invitation, and it needs to expire together with the invitation.

I see a few options:

1. We currently have some users stored with TTL.  We did this here, but the complexity of the changes is not very trust-inspiring; nor is the fact that we got an inconsistent record, even if we can find and fix this particular case where that happens.

2. Add columns to the invitation table.  I tried this before, and the resuling code was even more complex, and I gave up on it.

3. Add a table `brig.users_invitation_pending`, store the extra data there with the same TTL as the invitation, and make sure records are always kept consistent.  I think this has the disadvantages of both 1 and 2.

4. Store the user data of the invitation in `brig.users`, without TTL, and store an extra index that is used by a [cronjob](...) to find and delete those entries in `brig.users` that *should* have a TTL.  With this index, this shouldn't put too much strain on the brig resources, and the operation is idempotent, so it doesn't matter if differnet brigs ([even old versions with weight 0](...)) are fighting for deleting the same records.

I propose to throw away this PR and go for 4.

fisx:
...  and a little bit of pseudo-code:

```
create table if not exists expired_invitations
  ( created_at  timestamp
  , user        uuid
  ) with clustering order by (created_at ASC)
```

Each time an invitation is created by scim, an entry is made in this table.

The batchjob then fetches:

```
select (createdAt, user) from expired_invitations
  where created_at < $(invitation ttl plus 2 days safety margin)
        -- do we need to repeate the order by from the table declaration?
        -- (i think not)
```

And iterates over the returned rows as follows:

```
go [] = pure ()
go ((createdAt, uid) : others) = do
  invExpired <- isNothing <$> lookupInvitation (coerce uid)
  when invExpired $ do
    lookupAccount uid >>= \case
      Just acc | accountState acc == PendingInvitation -> deleteUser uid
      _ -> pure ()
    deleteExpiredInvitation createdAt uid
    go others
```

jschaul:
>     3. Add a table `brig.users_invitation_pending`, store the extra data there with the same TTL as the invitation, and make sure records are always kept consistent.  I think this has the disadvantages of both 1 and 2.

I thought this solution would not be too complex, but perhaps there are more code paths that need to be tied in there and it makes things weird and complex.

> 
>     4. Store the user data of the invitation in `brig.users`, without TTL, and store an extra index that is used by a [cronjob](...) to find and delete those entries in `brig.users` that _should_ have a TTL.  With this index, this shouldn't put too much strain on the brig resources, and the operation is idempotent, so it doesn't matter if differnet brigs ([even old versions with weight 0](...)) are fighting for deleting the same records.

This sounds like it's a fairly easy change haskell code wise, with the advantage of using the exact same user delete code than what is normally used when deleting. So this sounds good. I think the only disadvantage of this approach is that a cron job is needed, and that cron job will need to run A) on the ubuntu-based infrastructure, and we need B) some while true sleep a day; loop over cassandra pod in the kubernetes infrastructure. It may make sense to not have 6 cron jobs deleting the same users when we have 6 brig instances. Perhaps this can be alleviated by using a cron job everywhere that contains a fixed + a random amount of sleeping time to avoid races/duplicate load on cassandra.

fisx:
sorry, my terminology was sloppy, and then i forgot t fill in the links in [my comment above](https://github.com/wireapp/wire-server/pull/1238#issuecomment-730213215).  Fixed now, please click again!

@jschaul also writes elsewhere:

> The only cassandra-specific concern there that I can mention is that deletes are inefficient on cassandra and cause tombstones that make all other queries to the affected table slower (see also the extreme example here: https://www.datastax.com/blog/cassandra-anti-patterns-queues-and-queue-datasets); therefore, it is an anti-pattern to write a lot; then delete or update a lot from the same table. TTLs are 1000x more efficient for that purpose.
> I suspect the amount of data to be inserted for the SCIM pending invites not to be massive; but I would hate for an automated system to have the following behaviour:
> 
> * I, SCIM CI system, make a request to wire.com/scim-endpoint and dump my 5000 employees in there and invite everyone;
> * for 2 weeks this operation remains idempotent (with just some cassandra reads)
> * after 2 weeks the cron job removes 4990 entries since only 10 people accepted that invite as it landed in spam for everyone else
> * I, SCIM CI system, make a new request to wire.com/scim-endpoint and dump my 5000 employees again
> * new invitations are sent again
> * ad infinitum: cassandra insert; cassandra delete to the main users table.
> 
> There is some risk to this, and having a SCIM-pending-users-separate table (option 3 above) will guarantee not to have performance problems for a normal cassandra read on the users table. But I think the risk at the moment is not very high, so I think I find option 4 acceptable at this point also.

fisx:
2 minor suggestions:

```
go [] = pure ()
go ((createdAt, uid) : others) = do
  ~~invExpired <- isNothing <$> lookupInvitation (coerce uid)~~  -- this is already given by the fact that invitations expire and the select filter.
  lookupAccount uid >>= \case
      Just acc | accountState acc == PendingInvitation -> deleteUser uid
      _ -> pure ()
  deleteExpiredInvitation createdAt uid
  go others
```

timestamp should point to the time when it expires, not the time when it's created (thanks @smatting ).

fisx:
replaced by https://github.com/wireapp/wire-server/pull/1264

PR #1241, opened 2020-11-04
Fix content-type headers in saml responses.

fisx:


fisx:
I wrote a `ServerError` body to a file using `Data.ByteString.Lazy.writeFile` and confirmed that it is UTF-8 encoded:

```
$ file /tmp/x
/tmp/x: XML 1.0 document, UTF-8 Unicode text, with very long lines, with no line terminators
```

Interesting: when I insert a `Ümläüts` into a `LByteString` [here](https://github.com/wireapp/wire-server/blob/3af1aa63a9cb4d6594a4e017b0ba38408c0f26d2/services/spar/src/Spar/App.hs#L465), the encoding changes to ISO-8859; if I type the string literal as `Text` and `cs` it, it's back to `UTF-8`.  This may be another reason why we want to close https://github.com/zinfra/backend-issues/issues/1476...

smatting:
From https://hackage.haskell.org/package/bytestring-0.11.0.0/docs/Data-ByteString.html
Beware: fromString truncates multi-byte characters
So you must no have any string literals with non-ascii characters.

fisx:
> Let's use `Text` and be safe.

i don't think that's an option.  it's hard-coded in servant :(

smatting:
Yes, but you can encode only at the last step (see my commit)

fisx:
Sorry, i wasn't reading very carefully.  Got it, thanks for the fix!

I added another nit-pick commit, but I think this should get merged.  @smatting please see if you want to approve, and if so, feel free to merge (squash-rebase, probably) right away.

Spar tests pass locally.

smatting (approved):

PR #1246, opened 2020-11-12
stack yaml lock change - forgotten to commit in previous PR?

jschaul:


smatting:
Yes, my bad... Thanks! How did you notice this?

smatting (approved):

mheinzel:
> Yes, my bad... Thanks! How did you notice this?

Probably because building `wire-server` with Stack updated these hashes, so they showed up in `git status` or `git diff`.

PR #1252, opened 2020-11-18
Make Content-MD5 header optional for asset upload

mheinzel:
See zinfra/backend-issues#1843.

To avoid undetected API breakage as in https://github.com/wireapp/wire-server/pull/1247, I added a testcase that uses a hardcoded multipart request as it is currently sent by clients. The test fails on #1247, but passes now.

Ideally we would not have our own multipart parser in the first place, but the change in `headers` was relatively straight-forward.


akshaymankar (approved):
I want to remove the custom parser we have written, it makes me nervous. But this looks good for now. The new compatibility test is a good idea!

PR #1254, opened 2020-11-23
Add federation aware endpoint for checking user existence

akshaymankar:
Part of [SQCORE-129](https://wearezeta.atlassian.net/browse/SQCORE-129)

Some major changes:
* Use servant for the new endpoints
* Migrate the old route to servant
* Routes in servant will have swagger generated in Swagger 2.0 format. It will be available at `/brig/api-docs`, this requires change to the nginx config.
* The servant routes will be available in a separate tab of swagger-ui. Maybe we should try to merge it with the "Swagger 2.0" tab.


jschaul (approved):
PR looks good to me now.

PR #1256, opened 2020-11-24
Feature flags cleanup

smatting:
This PR fixes some minor issues with feature flag endpoints. See https://wearezeta.atlassian.net/browse/SQSERVICES-80

- Fixes: GET applock on non-existing team yields default config.  Should yield 404
- Fixes: stern: PUT applock on non-existing responds with 500. Should be 404
- Feature: stern api takes all no-config feature flags into one endpoint
- Refactor: Rename `Galley.API.TeamNotifications` to `Galley.API.Teams.Notifications` and introduce `Galley.API.Teams.Features`
- Refactor: DRY up `Galley.API.Teams.Features`
- Refactor: one accessor fn `featureStatus` for both `TeamFeatureStatusNoConfig` and `TeamFeatureStatusWithConfig`
- Feature: make applock defaults configurable on server side.


services/galley/src/Galley/Data/TeamFeatures.hs
```diff
@@ -72,8 +66,7 @@ setFeatureStatusNoConfig ::
   (TeamFeatureStatus a) ->
   m (TeamFeatureStatus a)
 setFeatureStatusNoConfig tid status = do
-  let flag = Public.tfwoStatus status
-  retry x5 $ write (update (Public.knownTeamFeatureName @a)) (params Quorum (flag, tid))
+  retry x5 $ write (update (Public.knownTeamFeatureName @a)) (params Quorum (featureStatus status, tid))
```
fisx, line 69: Can we now get rid of `FeatureHasNoConfig` altogether?  At least in this function and the getter, I think.

fisx: I'm not so convinced this is an improvement.  I thought there was some simplification in there, but in all cases in this commit we just replace a call to a more specialized function by one to a more general function.  And even if we can drop the `HasNoConfig` constraint here, I'm not sure that's a good thing.  We certainly don't need the more general form now.

I'll take another look at this tomorrow and try to remember why I was suggesting this.  Perhaps it was just a mistake?

fisx: > I'm not so convinced this is an improvement. I thought there was some simplification in there, but in all cases in this commit we just replace a call to a more specialized function by one to a more general function. And even if we can drop the `HasNoConfig` constraint here, I'm not sure that's a good thing. We certainly don't need the more general form now.
> 
> I'll take another look at this tomorrow and try to remember why I was suggesting this. Perhaps it was just a mistake?

I checked again, and it looks like there is not much to win here.  I tried to simplify things with `DuplicateRecordFields`, and decided I like the extension, but it doesn't work very well [in our case](https://github.com/wireapp/wire-server/blob/smatting/applock-followup-2/services/galley/src/Galley/Data/TeamFeatures.hs#L69):

```
/home/mf/src/wire/wire-server/services/galley/src/Galley/Data/TeamFeatures.hs:69:78: error:
     Ambiguous occurrence ‘featureStatus’
```

I pushed it anyway and reverted it, for your amusement.

I'll leave it up to you whether to keep 6416c63 or revert it, I have a slight preference for removing it (sorry :)).

fisx (approved):
https://github.com/wireapp/wire-server/pull/1256#discussion_r531445240 and https://github.com/wireapp/wire-server/pull/1256#discussion_r531010069, otherwise :shipit:!

services/galley/src/Galley/API/Teams.hs
```diff
@@ -1037,7 +1037,15 @@ getSearchVisibilityInternalH (tid ::: _) =
 
 getAppLockInternal :: TeamId -> Galley (Public.TeamFeatureStatus 'Public.TeamFeatureAppLock)
 getAppLockInternal tid = do
-  Defaults defaultStatus <- view (options . optSettings . setFeatureFlags . flagAppLockDefaults)
+  mbDefaults <- view (options . optSettings . setFeatureFlags . flagAppLockDefaults)
+  let defaultStatus =
+        maybe
```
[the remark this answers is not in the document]
smatting: Done

services/galley/src/Galley/Data/TeamFeatures.hs
smatting: I like the aesthetics better, but yes, seems superfluous. Reverted

PR #1258, opened 2020-11-25
Galley: Rename endpoint features/ -> features

smatting:


jschaul (approved):

fisx:
Ah, interesting.  If it was an nginz-issue, perhaps the reordering of the end-points wasn't even necessary?  (Either way this way's better.)

smatting:
Yes the reordering was unnecessary, but it looks better in the swagger ui this way.

PR #1261, opened 2020-11-26
Federation config flags: brig & galley

jschaul:
Relates to https://wearezeta.atlassian.net/browse/SQCORE-229

Introduces the following mandatory configuration option in brig and in galley:

```yaml
  #   Federation domain is used to qualify local IDs and handles,
  #   e.g. 0c4d8944-70fa-480e-a8b7-9d929862d18c@wire.com and somehandle@wire.com.
  #   It should also match the SRV DNS records under which other wire-server installations can find this backend:
  #      _wire-server._tcp.<federationDomain>
  #   Once set, DO NOT change it: if you do, existing users may have a broken experience and/or stop working
  #   Remember to keep it the same in Galley.
```

Also removes some IdMapping code (but only partially so as to not make this PR too big) and replaces it with an "always assume idMapping to be local" stub, as we want to remove the idmapping everywhere (but iteratively as we change endpoints)


mheinzel (approved):
Yup, this matches what was discussed. Looks good.

jschaul:
Tests all pass locally.

PR #1264, opened 2020-11-30
Clean up SCIM-invited users with expired invitation

smatting:
Implements the clean up job discussed in https://github.com/wireapp/wire-server/pull/1238

https://wearezeta.atlassian.net/browse/SQSERVICES-26


services/brig/schema/src/V62_users_pending_activation.hs
```diff
@@ -0,0 +1,38 @@
...
+import Text.RawString.QQ
+
+migration :: Migration
+migration =
+  Migration 62 "Add users_pending_activation" $
+    schema'
+      [r|
+        CREATE TABLE users_pending_activation
+        (
+          expires_at_day  int
```
smatting, line 31: This is date of the `expires_at` timestamp/
We use `int` for the encoding instead of `date`, because the cassandra lib doesn't seem to implement `date` literals.

services/brig/src/Brig/Run.hs
```diff
@@ -103,3 +124,51 @@ lookupRequestIdMiddleware :: (RequestId -> Wai.Application) -> Wai.Application
...
+                    if isExpired
+                      then do
+                        isPendingInvitation <- (Just PendingInvitation ==) <$> lookupStatus uid
+                        invExpired <- isNothing <$> Data.lookupInvitation tid (coerce uid)
+                        when (isPendingInvitation && invExpired) $ do
+                          API.deleteUserNoVerify uid
+                        pure (Just uid)
+                      else pure Nothing
+                )
+        unless (null expiredEntries) $
```
smatting, line 154: Hm... this would delete rows for a table inside a pagination of query to that table.
Probably not a good idea... I'll move it outside of the pagination

smatting: Done

fisx:
I restarted the CI: https://bridge-ie-concourse.zinfra.io/teams/main/pipelines/wire-server-pr/jobs/wire-server-integration/builds/1683

services/brig/src/Brig/Run.hs
```diff
@@ -103,3 +124,51 @@ lookupRequestIdMiddleware :: (RequestId -> Wai.Application) -> Wai.Application
...
+
+cleanExpiredPendingInvitations :: AppIO ()
+cleanExpiredPendingInvitations = do
+  safeForever "cleanExpiredPendingInvitations" $ do
+    today <- utctDay <$> (liftIO =<< view currentTime)
+    for_ [0 .. 9] $ \i ->
+      cleanUpDay (addDays (- i) today)
+    let d :: Int = 24 * 60 * 60
+    randomSecs <- liftIO (round <$> randomRIO @Double (0.5 * fromIntegral d, fromIntegral d))
+    threadDelay (randomSecs * 1_000_000)
```
fisx, line 136: This could be a local helper function.

services/brig/src/Brig/Run.hs
```diff
@@ -103,3 +124,51 @@ lookupRequestIdMiddleware :: (RequestId -> Wai.Application) -> Wai.Application
...
+      where
+        go :: [UserId] -> Page PendingActivationExpiration -> AppIO [UserId]
+        go users (Page hasMore result nextPage) = do
+          users' <- (<> users) <$> f result
+          if hasMore
+            then go users' =<< liftClient nextPage
+            else pure users'
+
+    safeForever :: (MonadIO m, MonadLogger m, MonadCatch m) => String -> m () -> m ()
+    safeForever funName action =
```
fisx, line 169: I think this is duplicating existing code.  Can we move the old function somewhere from where we can call it here as well?  Or is it awkward to generalize?  (Not that important.)

services/brig/src/Brig/Run.hs
```diff
@@ -103,3 +124,51 @@ lookupRequestIdMiddleware :: (RequestId -> Wai.Application) -> Wai.Application
...
+                      invExpired <- isNothing <$> Data.lookupInvitation tid (coerce uid)
+                      when (isPendingInvitation && invExpired) $ do
+                        API.deleteUserNoVerify uid
+                      pure (Just uid)
+                    else pure Nothing
+              )
+      unless (null expiredEntries) $
+        removeTrackedExpiration day expiredEntries
+
+    forTrackedExpirations :: Day -> ([PendingActivationExpiration] -> AppIO [UserId]) -> AppIO [UserId]
```
fisx, line 156: It would be good for space usage if you could process one page fully (including deletions) before pulling the next one.  Not sure what cassandra has to say about this (@jschaul, again?  sorry..).  This function could look like this:

```suggestion
    forTrackedExpirations :: Day -> (Page PendingActivationExpiration -> AppIO ()) -> AppIO ()
```

And the conditional call to `removeTrackedExpiration(s)` above could be moved into the callback and run once per page.

If it turns out problematic to paginate through a select query while deleting stuff in the same table, you can just keep pulling the first page, and it'll always be the next one because you just deleted the previous one.  (Not sure that's any better on performance, but note that we're only worried about system resources here, not response time -- it's fine to wait for cassandra to process the deletions if that takes time.)

services/brig/src/Brig/Run.hs
```diff
@@ -103,3 +124,51 @@ lookupRequestIdMiddleware :: (RequestId -> Wai.Application) -> Wai.Application
...
+    let d :: Int = 24 * 60 * 60
+    randomSecs <- liftIO (round <$> randomRIO @Double (0.5 * fromIntegral d, fromIntegral d))
+    threadDelay (randomSecs * 1_000_000)
+  where
+    cleanUpDay :: Day -> AppIO ()
+    cleanUpDay day = do
+      expiredEntries <- forTrackedExpirations day $ \exps ->
+        catMaybes
+          <$> ( for exps $ \(PendingActivationExpiration uid expiresAt tid) -> do
+                  isExpired <- (expiresAt <=) <$> (liftIO =<< view currentTime)
```
fisx, line 143: As I wrote in jira: we can simplify this logic and remove the `expiresAt` field altogether, at the cost of having data dangle for (at most) 24h longer: just process full days, but start with things that expired *some time* yesterday, not today.  I think this will also elminate the need for the `Ord` instance.

services/brig/src/Brig/Run.hs
```diff
@@ -103,3 +124,51 @@ lookupRequestIdMiddleware :: (RequestId -> Wai.Application) -> Wai.Application
...
+                  if isExpired
+                    then do
+                      isPendingInvitation <- (Just PendingInvitation ==) <$> lookupStatus uid
+                      invExpired <- isNothing <$> Data.lookupInvitation tid (coerce uid)
```
fisx, line 147: No need to do this:

```suggestion
```

This is true by construction.

services/brig/src/Brig/Run.hs
```diff
@@ -103,3 +124,51 @@ lookupRequestIdMiddleware :: (RequestId -> Wai.Application) -> Wai.Application
...
+                      when (isPendingInvitation && invExpired) $ do
+                        API.deleteUserNoVerify uid
```
fisx, line 149: Yeah, we should keep an eye on this.

I suggest we:

- add a function `deleteUsersNoVerify` that deletes many users
- we can implement it for now using `mapM_`
- but we should call it only once per page of `PendingActivationExpiration`
- and keep a metric that tells us how often this is called
- ideally, we should also have a metric that tells us how much time & space deletion takes, but that may be too much work for now.

fisx (changes requested):
There are no tests whatsoever.  Not sure how to test this, and where, perhaps we can think about this together.

Sorry, lots of ideas how to change this.  Do you want to go through them together?

services/brig/src/Brig/Run.hs
```diff
@@ -103,3 +124,51 @@ lookupRequestIdMiddleware :: (RequestId -> Wai.Application) -> Wai.Application
 lookupRequestIdMiddleware mkapp req cont = do
   let reqid = maybe def RequestId $ lookupRequestId req
   mkapp reqid req cont
+
+cleanExpiredPendingInvitations :: AppIO ()
+cleanExpiredPendingInvitations = do
+  safeForever "cleanExpiredPendingInvitations" $ do
+    today <- utctDay <$> (liftIO =<< view currentTime)
```
fisx, line 131: Why not `getCurrentTime`?  If this is in the `ReaderT`, does it ever change after boot time?

services/brig/schema/src/V62_users_pending_activation.hs
smatting: I just found out that while [cassandra-util](https://hoogle.zinfra.io/file/wire-server/.stack-work/install/x86_64-linux/dd82fa75c77cff497223fd50d01ad955b5a7ef42b25996a8a0e9a15d7718cbc4/8.8.4/doc/cassandra-util-0.16.5/Cassandra.html#t:Value) doesnt provide a `CqlDate` constructor, [cql](https://hoogle.zinfra.io/file/root/.stack/snapshots/x86_64-linux/de08094f0e7de5d808e5aaad6423b6aea588428a561794bbc07b2b7f1c963979/8.8.4/doc/cql-4.0.2/Database-CQL-Protocol.html#t:Value) does. However it's `CqlDate Int32`, so it's not immediately clear how dates are encoded as integers in cassandra.

libs/types-common/src/Data/Misc.hs
```diff
@@ -322,3 +326,15 @@ instance FromJSON PlainTextPassword where
 instance Arbitrary PlainTextPassword where
   -- TODO: why 6..1024? For tests we might want invalid passwords as well, e.g. 3 chars
   arbitrary = PlainTextPassword . fromRange <$> genRangeText @6 @1024 arbitrary
+
+--------------------------------------------------------------------------------
+-- ModJulianDay
+
+newtype ModJulianDay = ModJulianDay {fromUTCDay :: Day}
```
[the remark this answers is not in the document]
smatting: Mod for "Modified Julian Day". Didn't want to nameclash with [Data.Time.Calendar](https://hoogle.zinfra.io/file/root/.ghcup/ghc/8.8.4/share/doc/ghc-8.8.4/html/libraries/time-1.9.3/Data-Time-Calendar.html#t:Day). Also this newtype was just me sidestepping the issue of writing a general`Cql` instances for `Day`. Now since we know that `CqlDate Int32` exist it might be worth writing these instances.

services/brig/src/Brig/Run.hs
smatting: Ah yes, right. Done

services/brig/src/Brig/Run.hs
smatting: I think `isPendingInvitation <- (Just PendingInvitation ==) <$> lookupStatus uid` can also removed by the same argument

smatting: I removed both conditions

services/brig/src/Brig/Run.hs
smatting: I saw this pattern `liftIO =<< view currentTime` all over brig and assumed it was a convention.
The field is `_currentTime :: IO UTCTime`. I guessed it exists to mock a different time

services/brig/schema/src/V62_users_pending_activation.hs
smatting: .. I did some investigation `2020-12-01` gets encoded to `-2147465051`

services/brig/src/Brig/Run.hs
fisx: You're right, didn't think about that.  Yeay, pairing!  :)

services/brig/src/Brig/Run.hs
fisx: Sorry, you're right.  Didn't think that through.

services/brig/src/Brig/Run.hs
smatting: Checking for invitations with `invExpired <- isNothing <$> Data.lookupInvitation tid (coerce uid)` is required, because I chose not to call `removeTrackedExpiration` when a user accepts their invitation (since it was necessary because of this check).
Even if we decide to call `removeTrackedExpiration` when the user accepts their invitation, we still might want to keep this check as a safety measure.

services/brig/schema/src/V62_users_pending_activation.hs
smatting: Decided not go forward with CqlDate

services/brig/schema/src/Main.hs
```diff
@@ -131,7 +132,8 @@ main = do
       V58.migration,
       V59.migration,
       V60_AddFederationIdMapping.migration,
-      V61_team_invitation_email.migration
+      V61_team_invitation_email.migration,
+      V62_users_pending_activation.migration
```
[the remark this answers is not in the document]
smatting: Done

services/brig/src/Brig/Data/PendingActivation.hs
```diff
@@ -0,0 +1,71 @@
...
+import Cassandra
+import Data.Id (TeamId, UserId)
+import Data.Misc (ModJulianDay (..))
+import Data.Time (UTCTime (utctDay))
+import Data.Time.Calendar (Day)
+import Data.Tuple.Extra (uncurry3)
+import Imports
+
+data PendingActivationExpiration
+  = PendingActivationExpiration
```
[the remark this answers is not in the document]
smatting: Renamed

services/brig/src/Brig/Run.hs
smatting: I've implemented "always pulling the first" page. It will loop until no more entries are returned from the query.

services/brig/src/Brig/Run.hs
smatting: To avoid scheduling to many deletions to fast, I added a throttling delay. With at most 2 items per second (that 170k per day).

services/brig/src/Brig/Run.hs
```diff
@@ -69,10 +87,13 @@ run o = do
       AWS.execute (e ^. awsEnv) $
         AWS.listen throttleMillis q (runAppT e . SesNotification.onEvent)
   sftDiscovery <- forM (e ^. sftEnv) $ Async.async . Calling.startSFTServiceDiscovery (e ^. applog)
+  expiryCleanup <- Async.async (runAppT e cleanExpiredPendingInvitations)
```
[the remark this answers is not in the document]
smatting: Done

services/brig/src/Brig/Run.hs
smatting: Err, checking for missing inivitation really doesn't make any sense, since invitations also get removed when a user accepts them. Checking for `PendingInvitation` however provides an additional guar. Thanks @jschaul

services/brig/test/integration/API/UserPendingActivation.hs
```diff
@@ -0,0 +1,347 @@
...
+  Spar ->
+  -- | Authentication
+  Maybe ScimToken ->
+  -- | User data
+  Scim.User.User SparTag ->
+  -- | Spar endpoint
+  Http ResponseLBS
+createUser_ spar auth user = do
+  -- NB: we don't use 'mkEmailRandomLocalSuffix' here, because emails
+  -- shouldn't be submitted via SCIM anyway.
```
fisx, line 251: But they *have* to be, otherwise where would the invitation emails be sent to?  Can you make this sentence a bit clearer?  Or is this something you just copied?  In that case this may be a message to myself.  :)

services/brig/test/integration/API/UserPendingActivation.hs
```diff
@@ -0,0 +1,347 @@
...
+createScimToken spar' owner = do
+  CreateScimTokenResponse tok _ <-
+    createToken spar' owner $
+      CreateScimToken
+        { createScimTokenDescr = "testCreateToken",
+          createScimTokenPassword = Just defPassword
+        }
+  pure tok
+
+createUser'step :: Spar -> Brig -> ScimToken -> TeamId -> Scim.User.User SparTag -> Email -> HttpT IO (WithMeta (WithId UserId (Scim.User.User SparTag)), Invitation, InvitationCode)
```
fisx, line 99: Why not this?

```suggestion
createUserStep :: Spar -> Brig -> ScimToken -> TeamId -> Scim.User.User SparTag -> Email -> HttpT IO (WithMeta (WithId UserId (Scim.User.User SparTag)), Invitation, InvitationCode)
```

services/brig/test/integration/API/UserPendingActivation.hs
```diff
@@ -0,0 +1,347 @@
...
+createUser'step spar' brig' tok tid scimUser email = do
+  scimStoredUser <- (createUser spar' tok scimUser)
```
fisx, line 101:
```suggestion
  scimStoredUser <- createUser spar' tok scimUser
```

fisx (changes requested):

fisx:
https://github.com/wireapp/wire-server/pull/1264#discussion_r541269069 is the only one I care about, you may ignore the rest if you like.  And feel free to point out how I got this one backwards, too.  :)

fisx:
one more thing: what happens if a user gets deleted by means other than expiring invitation?  it will still have an entry in the new table, right?  will the gc loop choke on that entry?  solution may be to modify the async delete-user routine to also handle the new table.

services/brig/src/Brig/API/User.hs
```diff
@@ -328,6 +332,7 @@ createUser new@NewUser {..} = do
           field "user" (toByteString uid)
             . field "team" (toByteString $ Team.iiTeam ii)
             . msg (val "Accepting invitation")
+        Data.removeTrackedExpiration uid
```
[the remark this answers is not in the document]
smatting: Rename to `usersPendingActivationRemove`

services/brig/src/Brig/API/User.hs
```diff
@@ -363,6 +368,12 @@ createUserInviteViaScim uid (NewUserScimInvitation tid loc name rawEmail) = (`ca
         -- the SCIM user.
         True
   lift $ Data.insertAccount account Nothing Nothing activated
+
+  ttl <- setTeamInvitationTimeout <$> view settings
+  now <- liftIO =<< view currentTime
+  let expiresAt = addUTCTime (realToFrac ttl) $ now
```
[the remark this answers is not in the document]
smatting: Fixed. It's now built-in to hls :) You should see less of this from me from now on

jschaul:
> one more thing: what happens if a user gets deleted by means other than expiring invitation? it will still have an entry in the new table, right? will the gc loop choke on that entry? solution may be to modify the async delete-user routine to also handle the new table.

I think use deletions are idempotent; so you can attempt to delete the same user over and over again with the same result. It might be nice to (if not already done, I didn't look at the code, just replying to the previous comment), upon regular user deletions, as part of the deletion onEvent flow also delete the entry in this new expiry-tracking table (which will be a no-op if no entry under that key exists)

services/brig/src/Brig/Data/UserPendingActivation.hs
```diff
@@ -0,0 +1,66 @@
...
+import Data.Time (UTCTime)
+import Imports
+
+data UserPendingActivation = UserPendingActivation
+  { upaUserId :: !UserId,
+    upaDay :: !UTCTime
+  }
+  deriving stock (Eq, Show, Ord)
+
+-- | Note: Call this function only after an invitation for the user has been created
```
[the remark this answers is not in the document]
smatting: This Note is no longer longer true since we updated the logic (if it ever was true). I'll remove it.

When brig crashes between creating the invitation and making this entry, then the user won't get garbage-collected.

The conditions for the user to get deleted are quite strict now:
- user has entry in table (which is only added after an invitation was created) AND
- after the expiration date the user is in state "PendingActivation"

services/brig/src/Brig/Run.hs
```diff
@@ -69,10 +80,13 @@ run o = do
       AWS.execute (e ^. awsEnv) $
         AWS.listen throttleMillis q (runAppT e . SesNotification.onEvent)
   sftDiscovery <- forM (e ^. sftEnv) $ Async.async . Calling.startSFTServiceDiscovery (e ^. applog)
+  scimInvitationCleanup <- Async.async (runAppT e cleanExpiredPendingInvitations)
```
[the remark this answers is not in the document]
smatting: Renamed consistently

services/brig/test/integration/API/UserPendingActivation.hs
smatting: It's just something I copied. git blame says it's your message :)

smatting:
If the GC loop cannot find a user then `isPendingInvitation` will be `False`:
```haskell
isPendingInvitation <- (Just PendingInvitation ==) <$> API.lookupStatus uid
```
The user then skips deletion and the entry is removed from the expiry table.

Currently there are no calls to remove an entry from the expiry table except for the loop. The GC loop removes entries from the expiry table by itself. I have slight prefrence for not adding it as a hook to the user deletion, because this creates two call sites that remove entries from the table, one of which is not even strictly necessary.

I have another clarification question for @fisx : Is it possible that spar creates a second invitation for a user after the first invitation is expired? If so then I need to implement this: "every user has at most on entry in the expiry table, which will be overwritten with every invitation".

jschaul:
> then I need to implement this: "every user has at most on entry in the expiry table, which will be overwritten with every invitation".

If needed, this implementation can be achieved easily with the following:

```diff
        (
          user            uuid
        , expires_at      timestamp
-        , primary key (user, expires_at)
+        , primary key (user)
        )       
```

I believe at this point you don't need to have multiple entries for the same user, so you don't need a composite primary key.

services/brig/src/Brig/Run.hs
smatting: It's already duplicated 3 times. I created a separate issue for this https://wearezeta.atlassian.net/browse/BE-486

smatting:
> If needed, this implementation can be achieved easily with the following:
> 
> ```diff
>         (
>           user            uuid
>         , expires_at      timestamp
> -        , primary key (user, expires_at)
> +        , primary key (user)
>         )       
> ```
Thanks! Applied.

services/brig/src/Brig/Data/UserPendingActivation.hs
```diff
@@ -38,7 +38,6 @@ data UserPendingActivation = UserPendingActivation
   }
   deriving stock (Eq, Show, Ord)
 
--- | Note: Call this function only after an invitation for the user has been created
```
fisx: 2a64d2825a101c181907417b2df6374035ef7834

services/brig/src/Brig/Data/UserPendingActivation.hs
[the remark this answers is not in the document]
fisx: 2a64d2825a101c181907417b2df6374035ef7834

fisx (approved):
I've committed one more change, pls check whether this makes sense, and if you agree :shipit:

services/brig/src/Brig/Data/UserPendingActivation.hs
smatting: ah, nice. Didn't see that.

PR #1267, opened 2020-12-01
Migrate code for easier ci

jschaul:
* moves some helm charts from wire-server-deploy to wire-server (thereby making it possible to update helm charts and wire-server source code at the same time in the same PR)
* moves integration-test-on-kubernetes helm_vars and utility scripts to this repository as well (in preparation for CI using scripts and helm charts only from this one git repository)
* adjusts a range of bash scripts to work with helm 3 and to simplify certain things.
* easy Makefile target to try it out locally (see its warning though)

First part for https://github.com/zinfra/backend-issues/issues/1860 

This PR is best reviewed commit by commit as the first commit imports charts from wire-server-deploy.

Review appreciated on discussion items (see https://github.com/zinfra/backend-issues/issues/1860#issuecomment-736750989)


lucendio (commented):
I think you raised some great questions in the backend issue. While going through the first step here, I started to wonder, where the wrapper chart `wire-server` should go - the infra or the services repo? And, should the persistence layer charts - even though they are the ephemeral ones and used for integration testing - go into wire-server, too? Even though, they are part of the integration testing, they are also part of *the demo setup*.

arianvp:
The `wire-server` wrapper-chart should definitely live in `wire-server` in my opinion.  It does have some dependencies that don't belong in `wire-server` but we could remove those.

I think the dependencies on `webapp`, `team-settings` `account-pages` and `legalhold` should be put in a `wire-frontend` wrapper chart and stay in `wire-server-deploy` for now.

arianvp:
An idea I had that I want to discuss tomorrow. Just adding it here so I do not forget:

Instead of having `brig`, `galley`, `cannon` etc be top-level charts and then hook them in with `dependencies` of `wire-server` needing this funky recursive `./bin/update.sh` script; we could move all the components charts to `wire-server/charts/{brig,galley,cannon}`

You never install them standalone anyway; and it will make the whole chart updating logic a bit less messy


This is  a documented thing to do:
https://helm.sh/docs/chart_template_guide/subcharts_and_globals/

jschaul:
> An idea I had that I want to discuss tomorrow. Just adding it here so I do not forget:
> 
> Instead of having `brig`, `galley`, `cannon` etc be top-level charts and then hook them in with `dependencies` using `repository: file://../brig` of `wire-server` needing this funky recursive `./bin/update.sh` script; we could move all the components charts to `wire-server/charts/{brig,galley,cannon}`
> 
> You never install them standalone anyway; and it will make the whole chart updating logic a bit less messy
> 
> This is a documented thing to do:
> https://helm.sh/docs/chart_template_guide/subcharts_and_globals/

Yep, that makes sense for wire-server (without the webapp parts)

jschaul:
I will make two separate follow-up PRs.

PR #1270, opened 2020-12-03
Integrate servant-swagger-ui to brig

jschaul:
Relates to https://wearezeta.atlassian.net/browse/SQCORE-232

This could be the first step towards deprecating usage of custom nginx-hosted zwagger-ui

See swagger json under http://localhost:8082/api/swagger.json
and a new and shiny UI under http://localhost:8082/api/swagger-ui

The same routes are also exposed through nginz.

We could keep the existing zwagger-ui as-is, but document new endpoints under a new URL going forward. In the long term, we could remove the old custom nginz-zwagger-ui.

new-ui

### DIFF
--- a/deploy/services-demo/conf/nginz/nginx.conf
+++ b/deploy/services-demo/conf/nginz/nginx.conf
@@ -143,7 +143,12 @@ http {
       proxy_pass http://brig;
     }
 
-    location /brig/api-docs {
+    location /swagger-ui {
+        include common_response_no_zauth.conf;
+        proxy_pass http://brig;
+    }
+
+    location /swagger.json {
         include common_response_no_zauth.conf;
         proxy_pass http://brig;
     }
--- a/deploy/services-demo/conf/nginz/nginx.conf
+++ b/deploy/services-demo/conf/nginz/nginx.conf
@@ -143,12 +143,12 @@ http {
       proxy_pass http://brig;
     }
 
-    location /swagger-ui {
+    location /api/swagger-ui {
         include common_response_no_zauth.conf;
         proxy_pass http://brig;
     }
 
-    location /swagger.json {
+    location /api/swagger.json {
         include common_response_no_zauth.conf;
         proxy_pass http://brig;
     }
--- a/services/brig/brig.cabal
+++ b/services/brig/brig.cabal
@@ -4,7 +4,7 @@ cabal-version: 1.12
 --
 -- see: https://github.com/sol/hpack
 --
--- hash: 9f982304b3e4de02c3a782a8053c4d3d2e0ab19d3e6bd497684f37ceaaf7215a
+-- hash: 58ca66d809d6b5ee7fb4c04547c644b520ca6b60213096123d0a384f72f6e3de
 
 name:           brig
 version:        1.35.0
@@ -195,6 +195,8 @@ library
     , servant
     , servant-server
     , servant-swagger
+    , servant-swagger-ui
+    , servant-swagger-ui-core
     , singletons >=2.0
     , smtp-mail >=0.1
     , sodium-crypto-sign >=0.1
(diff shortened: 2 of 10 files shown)


fisx:
Very happy to see this is happening!  :)

How are you planning to consolidate the different services?  Is brig going to serve the swagger for all services?  Or is there one UI per service?  How are they mapped in nginz?

See also: https://github.com/wireapp/wire-server/pull/757, https://github.com/wireapp/wire-server/pull/861.

services/brig/brig.cabal
```diff
@@ -195,6 +195,8 @@ library
     , servant
     , servant-server
     , servant-swagger
+    , servant-swagger-ui
+    , servant-swagger-ui-core
```
akshaymankar, line 199: Just so we don't forget this adds dependency on the lzma c-library. This will need to be added to the dockerfile(s) and AMIs.

jschaul:
> See also: #757, #861.

Thanks for the conext, maybe some of this will become useful as we port more endpoints.

> How are you planning to consolidate the different services? Is brig going to serve the swagger for all services? Or is there one UI per service? How are they mapped in nginz?

One step at a time. No combined swagger in this PR. But thinking ahead, yes, I was thinking brig could host the swagger jsons of the other services (it seems one could parse into `Swagger` and `<>` them if I understand the library correctly) since brig already has to contact all the other services on e.g. user deletions.

This will only work for swagger 2.0 for the time being (no, I do not plan to support the old swagger on this endpoint, we can keep the old endpoint for the old swagger), so that also means only spar and brig are needed at this point.

services/brig/brig.cabal
jschaul: Done.

jschaul: No, actually, this one: https://github.com/zinfra/hegemony/pull/1781 is also needed.

services/brig/src/Brig/API/Public.hs
```diff
@@ -221,14 +224,24 @@ type OutsideWorldAPI =
     :<|> GetUserUnqualified
     :<|> GetUserQualified
 
-type ServantAPI =
-  "brig" :> "api-docs" :> Get '[Servant.JSON] Swagger
-    :<|> OutsideWorldAPI
+type ServantHandlerAPI = "api" :> SwaggerSchemaUI "swagger-ui" "swagger.json"
+
+type ServantAPI = OutsideWorldAPI
```
akshaymankar, line 229: The names `ServantAPI` and `ServantHandlerAPI` don't communicate much difference to me. Can we rename `ServantHandlerAPI` to something more meaningful, maybe `SwaggerDocsAPI`?

deploy/services-demo/conf/nginz/nginx.conf
```diff
@@ -143,7 +143,12 @@ http {
       proxy_pass http://brig;
     }
 
-    location /brig/api-docs {
+    location /api/swagger-ui {
+        include common_response_no_zauth.conf;
+        proxy_pass http://brig;
+    }
+
+    location /api/swagger.json {
```
akshaymankar, line 151: I think we'll also need to change the HTML file in the old swagger-ui to match this change.

akshaymankar (changes requested):
Overall looks good, the new UI also looks way nicer!
I think this breaks the old swagger-ui, so let's either remove the new endpoint from there or fix it.
Feel free to do the rename refactor I requested in a follow up PR.

services/brig/src/Brig/API/Public.hs
jschaul: good point! Thanks.

deploy/services-demo/conf/nginz/nginx.conf
jschaul: I removed those entries now.

akshaymankar (approved):
:rocket:

PR #1271, opened 2020-12-04
brig: Simplify SRV record fetching

arianvp:
Context:

In https://wearezeta.atlassian.net/browse/SQPIT-63 it was decided to add an option
to brig to use a static URI for sftd as in Kubernetes SFTD will be fronted by a proper load-balancer; which makes the SRV records redundant.

(pull requests: https://github.com/wireapp/wire-server-deploy/pull/382, https://github.com/wireapp/wire-server-deploy/pull/383)

I started working on this change and intended to not touch the SRV code at all and simply add a new field but this was more complicated than I thought given we set the discovered SFT nodes in a separate thread that periodically polls and hooking in a static field sounded complicated. Then I asked why are we having a separate thread polling for results in the first place? It shouldn't matter to just resolve the DNS record on each and every request because caching should make that efficient. Then I realised that we're holding the `dns` library wrong and if we hold it right we indeed get caching for free making all the periodic fetching code unnecessary.

This removed some Polysemy code (sorry) but I think it's less interesting because we can
now assume the implementation (and the tests) of the `dns` library is correct and not have to implement all this logic by ourselves. So in some shape or form eventhough we removed polysemy this code is more pure than it was before. 

The DNS library that we are using already does caching and honors TTLs
published in DNS. The important part to making this work is that all
calls to `resolve` should use the same `Resolver`

In the `dns-util` library that we used before, we would create a new
`Resolver` object for each request, side-stepping the caching
implementation completely.

By reusing the `Resolver` across threads, calls to `resolveSRV` will
only resolve when a cached entry is expired. The default maximum TTL is
300 seconds.

Now there is also no need for running a worker thread in the background
fetching the record periodically.

I have more faith in this library implementing caching correctly than I
have faith in our implementation and it's less code to maintain.

libs/dns-util: remove DNS/Effect.hs

The function is buggy as it allocates a new resolver for every call,
whilst you should reuse the resolver across calls.

Co-Authored-By: Florian Klink <flokli@flokli.de>


arianvp:
This code is under the assumption that the `srv-announcer`  is setting TTLs in the records.

If not then this code is a serious performance regression from what was written before

flokli:
```
srv-announcer --help
[…]
   --srv-record-ttl value       TTL of the RFC2783 SRV Record Set in seconds (default: 60) [$SRV_ANNOUNCER_SRV_RECORD_TTL]
[…]
```

arianvp:
@akshaymankar says:

> I can see a regression though, on error, we decided to serve old entries if we have them, but without caching this outside the dns library, we will not be able to do so.

julialongtin:
I browsed over this.. where does it limit the records the client receives to 5 at random?

services/brig/test/unit/Test/Brig/Calling.hs
```diff
@@ -1,241 +0,0 @@
...
-recordLogs LogRecorder {..} = interpret $ \(PolyLog lvl msg) ->
-  modifyIORef' recordedLogs (++ [(lvl, Log.render (Log.renderDefault ", ") msg)])
-
-ignoreLogs :: Sem (PolyLog ': r) a -> Sem r a
-ignoreLogs = interpret $ \(PolyLog _ _) -> pure ()
-
-{-# ANN tests ("HLint: ignore" :: String) #-}
-tests :: TestTree
-tests =
-  testGroup "Calling" $
```
jschaul, line 70: While it's a good catch to notice that the DNS library can be used differently (and thus simplify some code), it's not a reason to delete most of the SFT tests without replacing them.

We have some behaviour that should have a test in order for it not to disappear in the next refactoring:
- using a maximum of only X SFT servers even if there are more
- randomization per-request (even if the list of SFT in SRV stays the same) for clients
- fail-safe behaviour in times where the SRV lookup fails temporarily (as has happened a few times in the past) to still give out the current list of servers
- no SFT in config when some config variable is not set

Can you please make sure to cover these cases with a test in this PR. Perhaps you need an integration test here now.

services/brig/src/Brig/Calling/API.hs
```diff
@@ -122,15 +134,21 @@ newConfig env mSftEnv limit = do
...
+    (resolver', domain', limit') <- MaybeT (pure sft)
+    -- NOTE: This is safe to do as domainText is ASCII
+    let domain = encodeUtf8 . domainText $ domain'
+    entries <- MaybeT $ do
+      resultsOrError <- liftIO $ Lookup.lookupSRV resolver' domain
+      -- NOTE: we silently discard DNS errors and serve no sft entries.
+      -- TODO(arianvp): Log this at least?
+      pure $ rightToMaybe resultsOrError
+    shuffled <- MaybeT $ do
+      shuffled' <- liftIO $ take limit' <$> shuffleM entries
```
arianvp, line 148: @julialongtin  here

services/brig/src/Brig/Calling/API.hs
```diff
@@ -90,8 +97,13 @@ getCallsConfigV2H (_ ::: uid ::: connid ::: limit) =
 getCallsConfigV2 :: UserId -> ConnId -> Maybe (Range 1 10 Int) -> Handler Public.RTCConfiguration
 getCallsConfigV2 _ _ limit = do
   env <- liftIO =<< readIORef <$> view turnEnvV2
-  sftEnv' <- view sftEnv
-  newConfig env sftEnv' limit
+  resolver' <- view resolver
+  sftDomain <- view (settings . Opt.sftDomain)
+  -- TODO not hardcode this default?
+  sftLimit <- (fromMaybe 5) <$> view (settings . Opt.sftListLength)
```
arianvp, line 103: @julialongtin and here

services/brig/test/unit/Test/Brig/Calling.hs
arianvp: Yep I'll re-add these tests. I agree they're useful to keep.

> fail-safe behaviour in times where the SRV lookup fails temporarily (as has happened a few times in the past) to still give out the current list of servers

@akshaymankar brought this up too.  But I'm not sure breaking cache invalidation in the presence of failures is the best solution here because it's hard to distinguish between true and transient failures. I think the change of behavior is a good one but I'm happy to be convinced otherwise.

 Instead we should tweak [`resolvRetry`](https://hackage.haskell.org/package/dns-4.0.1/docs/Network-DNS-Resolver.html#v:resolvRetry) to better survive transient failures and we should tweak the record's TTL for more  "permanent"  transient failures.  Currently the SRV records have a TTL of 10 seconds. We could increase that to a few minutes instead.  If there is a positive hit the `dns` library will not contact the DNS server again in that TTL time period.

>  (as has happened a few times in the past) 

Do we have these incidents recorded somewhere?  My gut feeling is that tweaking the TTLs is sufficient to avoid these kind of issues  but perhaps I'm wrong. Knowing exactly the problems we encountered would be useful.

arianvp: Is it this test?

https://github.com/wireapp/wire-server/pull/1271/files#diff-100069ac03fa9aae9c8e7e80a089c378c81ace107d2717d60d77c76cb0f04e4bL135-L149

I am now even more confused about the current behavior.  I  would understand that the record is kept if `SrvResponseError` is returned (A transient faiure) 

but on `SrvNotAvailable` I would expect the record to be properly removed. If the DNS server is giving me an authoritive answer that the record is not available we should not ignore that. How do we ever remove SFT servers from the pool otherwise?

I would not consider `SrvNotAvailable` an "unsuccessful" answer. It is a successful answer; just one that tells there are no entries.

arianvp: Ah this only happens in the scenario where there are _no_ records in the response.

https://github.com/wireapp/wire-server/blob/develop/libs/dns-util/src/Wire/Network/DNS/SRV.hs#L90

If there are no records in the response we keep the previously known set.

Why do we do this though? What scenario can be avoided? Why would a DNS server suddenly stop resolving the entries for the `SRV` record?

flokli: Regarding `srv-announcer`'s behaviour: if we were the only record left and to be removed, a `ChangeActionUpsert`, with an empty `ResourceRecordSet` would be sent. I don't know how R53 will behave on that. Assuming it will become an empty set, If there's an empty set in DNS, I'd expect brig to not return any records.

If we were to turn off all of SFT, this should be possible by stopping all announcers.

I don't think "sticking with the old view" when no records are returned is correct behaviour here - if the DNS server authoritatively replies there's no records we should honor that.

On the other side, if we get a SERVFAIL, we can only retain the records for the lifetime of the TTL, and after that, we just don't know. But safeguarding against R53 outages is probably out of scope for this.

In all cases, this is all only state still running brigs know, and won't be retained if they're upscaled. We rely on these internal caches as less as possible.

PR #1272, opened 2020-12-04
brig:  Allow setting a static SFT Server

arianvp:
When SFT is deployed behind a load-balancer we can get away with a
single static entry. We need this for our Kubernetes-based SFT
deployment.

This still needs an integration test

This is needed for https://github.com/wireapp/wire-server-deploy/pull/383 and https://github.com/wireapp/wire-server-deploy/pull/382

If we in the near future move SFT into kubernetes for our own production environment  we could remove the SRV code. Instead I keep it around for now.

This is currently based on https://github.com/wireapp/wire-server/pull/1271  but there is nothing in that PR that is **absolutely necessary** except that (to me) the change was a bit easier to do. 

But "ignore SRV lookups if we have a static URL" could perfectly well be done on develop as well. Lets see how controversial the other PR is; but if it's not and can be quickly merged I prefer to not rebase this :)

So I could rebase this onto develop if need be. It would some adjustments to apply cleanly

Part of  https://wearezeta.atlassian.net/browse/SQPIT-63 and https://wearezeta.atlassian.net/browse/SQPIT-64

Also see the design in https://wearezeta.atlassian.net/wiki/spaces/ENGINEERIN/pages/150930455/SFT+Load+balancing

### DIFF
--- a/services/brig/src/Brig/Calling/API.hs
+++ b/services/brig/src/Brig/Calling/API.hs
@@ -38,6 +38,7 @@ import Data.Id
 import Data.List.NonEmpty (NonEmpty (..))
 import qualified Data.List.NonEmpty as NonEmpty
 import qualified Data.List1 as List1
+import Data.Misc (HttpsUrl)
 import Data.Range
 import qualified Data.Swagger.Build.Api as Doc
 import Data.Text.Ascii (AsciiBase64, encodeBase64)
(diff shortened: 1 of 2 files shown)


jschaul (commented):
This PR is fine. If you want to get it in fast, perhaps rebase it on develop so we can bikeshed longer on that other one :)

Also, a test would be nice...

arianvp:
Superseded by https://github.com/wireapp/wire-server/pull/1277/files which is branched off develop

PR #1274, opened 2020-12-07
IdP permissions

fisx:
Fixes https://github.com/zinfra/backend-issues/issues/931


fisx:
I tested the last 3 commits locally.  the new test case fails as expected, with the fix it passes, and with the adjustment of the other tests, all unit and integration tests pass.

smatting (approved):

PR #1277, opened 2020-12-07
brig: Allow setting a static SFT Server  (rebased  on develop)

arianvp:
When SFT is deployed behind a load-balancer we can get away with a
single static entry. We need this for our Kubernetes-based SFT
deployment.

### DIFF
--- a/services/brig/src/Brig/Options.hs
+++ b/services/brig/src/Brig/Options.hs
@@ -33,6 +33,7 @@ import Data.Aeson.Types (typeMismatch)
 import qualified Data.Char as Char
 import Data.Domain (Domain (..))
 import Data.Id
+import Data.Misc (HttpsUrl)
 import Data.Range
 import Data.Scientific (toBoundedInteger)
 import qualified Data.Text as Text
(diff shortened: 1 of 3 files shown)


services/brig/src/Brig/Options.hs
```diff
@@ -483,7 +484,12 @@ data Settings = Settings
     -- Customer extensions
 
     -- | Customer extensions.  Read 'CustomerExtensions' docs carefully!
-    setCustomerExtensions :: !(Maybe CustomerExtensions)
+    setCustomerExtensions :: !(Maybe CustomerExtensions),
+    -- | When set; instead of using SRV lookups to discover SFTs the calls
+    -- config will always return this entry. This is useful in Kubernetes
+    -- where SFTs are deployed behind a load-balancer.  In the long-run the SRV
+    -- fetching logic can go away completely
+    setSftStaticUrl :: !(Maybe HttpsUrl)
```
arianvp, line 492: From @jschaul  in https://github.com/zinfra/hegemony/pull/1780 

> I hate these config options renaming changes. They make it really hard to move forward: this need to be merged just before the wire-server pr is merged; not before, otherwise SFT breaks on staging; not after, otherwise SFT breaks on staging. Same for production, it makes a tight coupling between this and wire-server.

> What's the rationale to move the config option name around? It's a lot of work for little gain.

The rationale is that `staticUrl` is a static "Options that persist as runtime settings."   . I.e. I can access the config option from the environment; unlike other options.  

This is needed because I need the static URI verbatim  in https://github.com/wireapp/wire-server/pull/1277/files#diff-6acd71212092653fb8863892a5716626958fc2cbd2c73ee7eca7ef80b51ef4e0R95


I thought this was the common convention in Brig. I I could add extra code to `Brig.Run` to special-case the `sft` options  so they are added to the environment manually but that seems of little gain to me if that code is already there for the `Settings` datatype which is there exactly because this pattern happens very often in our codebase.

jschaul (approved):
Looks good as long as CI agrees :)

Thanks for the test!

PR #1278, opened 2020-12-07
add xz / lzma library to docker files

jschaul:
in preparation for https://github.com/wireapp/wire-server/pull/1270

### DIFF
--- a/build/alpine/Dockerfile.deps
+++ b/build/alpine/Dockerfile.deps
@@ -27,4 +27,5 @@ RUN apk add --no-cache \
             llvm-libunwind \
             ca-certificates \
             dumb-init \
+            xz \
             libxml2
--- a/build/alpine/Dockerfile.deps
+++ b/build/alpine/Dockerfile.deps
@@ -27,5 +27,5 @@ RUN apk add --no-cache \
             llvm-libunwind \
             ca-certificates \
             dumb-init \
-            xz \
+            xz-libs \
             libxml2
--- a/build/alpine/Dockerfile.prebuilder
+++ b/build/alpine/Dockerfile.prebuilder
@@ -34,7 +34,7 @@ RUN apk add --no-cache \
         snappy-dev \
         llvm-libunwind-dev \
         bash \
-        xz \
+        xz-dev \
         libxml2-dev \
         git \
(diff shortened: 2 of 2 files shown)


build/alpine/Dockerfile.prebuilder
```diff
@@ -34,7 +34,7 @@ RUN apk add --no-cache \
         snappy-dev \
         llvm-libunwind-dev \
         bash \
-        xz \
+        xz-dev \
```
akshaymankar, line 37: I think we need `xz` for `ghcup` to work.

build/alpine/Dockerfile.deps
```diff
@@ -27,4 +27,5 @@ RUN apk add --no-cache \
             llvm-libunwind \
             ca-certificates \
             dumb-init \
+            xz \
```
akshaymankar, line 30: I looked for liblzma.so.5 and found it in `xz-utils`:
https://pkgs.alpinelinux.org/contents?file=liblzma.so.5&path=&name=&branch=v3.12
TBH, I don't really know the difference between liblzma.so and liblzma.so.5, but my locally compiled binaries depend on `liblzma.so.5`.

akshaymankar (changes requested):

akshaymankar:
I think we don't need this PR because the tests are green for alpine builds for #1270

build/alpine/Dockerfile.prebuilder
jschaul: I thought one included the other, apparently not. Thanks for catching this issue, I now include both.

build/alpine/Dockerfile.deps
jschaul: `xz-libs` is what you meant to write, yes, I added that (though it seems to already get included by something else and is present already in the alpine-deps image).

akshaymankar: > xz-libs is what you meant to write
yes, not enough coffee!

Did you see my other comment? Maybe we don't need this PR, the pipeline is happy.

jschaul: I thought explicit is better than not explicit, but ok :)

PR #1281, opened 2020-12-09
Create endpoint for getting handle info by qualified handle

akshaymankar:
Federation is not implemented yet, so remote handles will always return 404.

### DIFF
--- a/libs/wire-api/src/Wire/API/User.hs
+++ b/libs/wire-api/src/Wire/API/User.hs
@@ -119,7 +119,7 @@ import Data.Misc (PlainTextPassword (..))
 import Data.Proxy (Proxy (..))
 import Data.Qualified
 import Data.Range
-import Data.Swagger (Referenced (Inline), ToSchema (..), genericDeclareNamedSchema, properties, required, schema)
+import Data.Swagger (ToSchema (..), genericDeclareNamedSchema, properties, required, schema)
 import qualified Data.Swagger.Build.Api as Doc
 import Data.Text.Ascii
 import Data.UUID (UUID, nil)
@@ -191,7 +191,7 @@ data UserProfile = UserProfile
 -- TODO: Deal with profileId being backwards compatible
 instance ToSchema UserProfile where
   declareNamedSchema _ = do
-    uuidSchema <- declareNamedSchema (Proxy @UUID)
+    idSchema <- deprecatedUnqualifiedSchemaRef (Proxy @UserId) "qualified_id"
     genericSchema <-
       genericDeclareNamedSchema
         ( swaggerOptions
@@ -209,7 +209,7 @@ instance ToSchema UserProfile where
     pure $
       genericSchema
         & over (schema . required) (List.delete "deleted")
-        & over (schema . properties) (InsOrdMap.insert "id" (Inline (view schema uuidSchema)))
+        & over (schema . properties) (InsOrdMap.insert "id" idSchema)
 
 modelUser :: Doc.Model
 modelUser = Doc.defineModel "User" $ do
--- a/libs/wire-api/src/Wire/API/User.hs
+++ b/libs/wire-api/src/Wire/API/User.hs
@@ -111,6 +111,7 @@ import qualified Data.Currency as Currency
 import Data.Handle (Handle)
 import qualified Data.HashMap.Strict as HashMap
 import qualified Data.HashMap.Strict.InsOrd as InsHashMap
+import qualified Data.HashMap.Strict.InsOrd as InsOrdMap
 import Data.Id
 import Data.Json.Util (UTCTimeMillis, (#))
 import qualified Data.List as List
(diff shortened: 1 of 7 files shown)


libs/wire-api/src/Wire/API/User/Handle.hs
```diff
@@ -42,6 +44,7 @@ import Wire.API.Arbitrary (Arbitrary, GenericUniform (..))
 newtype UserHandleInfo = UserHandleInfo {userHandleId :: UserId}
```
jschaul, line 44: this data type needs to contain a domain as well.

jschaul (changes requested):

libs/wire-api/src/Wire/API/User/Handle.hs
akshaymankar: Yes, I also discovered that we missed it while creating new endpoints to get User, I will fix those too.

akshaymankar:
Bonus change:
* Return qualified id while returning user by ID.

libs/wire-api/src/Wire/API/User.hs
```diff
@@ -111,6 +111,7 @@ import qualified Data.Currency as Currency
 import Data.Handle (Handle)
 import qualified Data.HashMap.Strict as HashMap
 import qualified Data.HashMap.Strict.InsOrd as InsHashMap
+import qualified Data.HashMap.Strict.InsOrd as InsOrdMap
```
jschaul, line 114: I think you only need one of those qualified imports.

libs/wire-api/src/Wire/API/User.hs
```diff
@@ -237,25 +240,26 @@ modelUser = Doc.defineModel "User" $ do
...
-        # "deleted" .= (if profileDeleted u then Just True else Nothing)
-        # "service" .= profileService u
-        # "handle" .= profileHandle u
-        # "locale" .= profileLocale u
-        # "expires_at" .= profileExpire u
-        # "team" .= profileTeam u
-        # "email" .= profileEmail u
-        # []
+    object
+      [ "id" .= _qLocalPart (profileQualifiedId u),
```
jschaul, line 244: FUTUREWORK (maybe in another PR): rename `_qLocalPart` to something nicer, e.g. `localId`

libs/wire-api/src/Wire/API/User/Handle.hs
```diff
@@ -49,14 +53,30 @@ modelUserHandleInfo = Doc.defineModel "UserHandleInfo" $ do
...
+            .~ [ ("user", unqualifiedIdSchema),
+                 ("qualified_user", qualifiedIdSchema)
+               ]
+
 instance ToJSON UserHandleInfo where
   toJSON (UserHandleInfo u) =
     object
-      ["user" .= u]
+      [ "user" .= _qLocalPart u, -- For backwards comptaibility
+        "qualified_user" .= u
```
jschaul, line 74: so here we use qualified_user not qualified_id? Isn't that inconsistent then?

libs/wire-api/src/Wire/API/User/Handle.hs
```diff
@@ -49,14 +53,30 @@ modelUserHandleInfo = Doc.defineModel "UserHandleInfo" $ do
...
+          & properties
+            .~ [ ("user", unqualifiedIdSchema),
+                 ("qualified_user", qualifiedIdSchema)
+               ]
+
 instance ToJSON UserHandleInfo where
   toJSON (UserHandleInfo u) =
     object
-      ["user" .= u]
+      [ "user" .= _qLocalPart u, -- For backwards comptaibility
```
jschaul, line 73:
```suggestion
      [ "user" .= _qLocalPart u, -- For backwards compatibility
```

jschaul (commented):
Minor nitpicks, looks good otherwise. Perhaps we should remain consistent with the naming of the fields though.

libs/wire-api/src/Wire/API/User/Handle.hs
akshaymankar: I am just prepending `qualified_` to whatever we had earlier. Doing this sounds more consistent than changing everything to `qualified_id`. WDYT?

jschaul: hm, I don't know. The old fields (sometimes called 'user' and sometimes called 'id') seemed inconsistent; this could be a time to change that and not keep that inconsistency. But maybe it doesn't matter if clients use the swagger definitions to create their API.

jschaul (approved):

libs/wire-api/src/Wire/API/User.hs
akshaymankar: On second thoughts, this is also used for `Qualified Handle` , so calling it `localId` doesn't make sense. I can't think for a better name than "localPart"

akshaymankar: Overthinking maybe, but `localPart` is also wrong as it could be a remote thing. So, ideally should be called `unqualified` or `unqualifiedPart`.

akshaymankar: Anyway, let's continue this discussion in https://github.com/wireapp/wire-server/pull/1290

mheinzel: I followed the naming of email adresses there: https://en.wikipedia.org/wiki/Email_address#Local-part

PR #1282, opened 2020-12-09
Increase the max allowed search results from 100 to 500.

fisx:
We decided that we'll keep using search in team-settings for now to get some overview of the team (in addition to a csv download button for getting an exhaustive member list).

I'm hoping this is not going to put too much load on ES in prod, but we can keep an eye on it during testing on staging.


fisx:
@akshaymankar you probably have an opinion on this?  is it a bad idea?  a terrible idea?  harmless?  :-)P

akshaymankar:
> @akshaymankar you probably have an opinion on this? is it a bad idea? a terrible idea? harmless? :-)P

Seems harmless, but a lot of harmful things seem like it sometimes. So, I'd say let's test queries with big results in staging before shipping this.

smatting (approved):

PR #1283, opened 2020-12-09
servantify /self endpoint

jschaul:
* move /self endpoint to servant & swagger 2.0
* expose `qualified_id` field for `GET /self`:

```
 "qualified_id": {
    "id": "00000000-0000-0000-0000-000000000000",
    "domain": "string"
  }
```


akshaymankar (approved):
:shark:

PR #1287, opened 2020-12-14
Remove DB tables for ID mapping

mheinzel:
This was unused already since #1261. Now the only thing remaining is to move a few API types from opaque IDs to qualified IDs.


jschaul (approved):
🐸

jschaul (commented):
Ah CI is right: the file V62_ is missing in the commit.

jschaul (approved):
All green now!

PR #1290, opened 2020-12-16
Data.Qualified: Better field names

akshaymankar:
* Typing underscores was getting annoying.
* _qDomain becomes qDomain.
* _qLocalPart becomes qUnqualified. The qualified thing may not be local, so
calling this local seems wrong.
* The `OptionallyQualified` type seems to be unused, I didn't delete it since I thought
  it would be useful when we want to be backwards compatible in request body.


mheinzel:
>     * _qLocalPart becomes qUnqualified. The qualified thing may not be local, so
>       calling this local seems wrong.

I followed the [naming of email adresses there](https://en.wikipedia.org/wiki/Email_address#Local-part), but I also see how it could be mis-interpreted.
Not sure which one is nicer, happy to move forward with `qUnqualified`.

mheinzel (approved):

PR #1291, opened 2020-12-16
Servantify get users by unqualified ids or handles

akshaymankar:


jschaul (commented):
Looks reasonable, though I see 5-6 TODOs still in here, I assume you still want to work on them during this PR?

jschaul:
The wire-api unittests fail with a mismatch between user profile JSON and swagger.

akshaymankar:
> The wire-api unittests fail with a mismatch between user profile JSON and swagger.

Yes, this should've been a draft.

services/brig/src/Brig/API/Public.hs
```diff
@@ -1220,26 +1224,36 @@ getUserDisplayNameH (_ ::: self) = do
...
+    (_, Just handles) ->
+      let normalRangedList = fromCommaSeparatedList $ fromRange handles
+          qualifiedList = (`Qualified` domain) <$> normalRangedList
+          -- Use of unsafeRange here is ok only because we know that 'handles'
+          -- is valid for 'Range 1 4'. However, we must not forget to keep this
+          -- annotation here otherwise a change in 'Public.ListUsersByHandles'
+          -- could cause this code to break.
+          qualifiedRangedList :: Range 1 4 [Qualified Handle] = unsafeRange qualifiedList
+       in listUsersByIdsOrHandles self (Public.ListUsersByHandles qualifiedRangedList)
+    (Nothing, Nothing) -> throwStd $ badRequest "at least one ids or handles must be provided"
```
mheinzel, line 1242: So in case both are provided, we just silently ignore the provided handles? Would be nicer to fail, maybe reducing the boilderplate by using some kind of combinator `Maybe a -> Maybe b -> Handler (Either a b)` or even API combinator.

mheinzel (approved):
It seems a bit weird to convert the unqualified IDs to qualified ones, just to then partition into local (unqualified) and remote ones again, but I think it makes sense in this situation.

There's one point about mutually exclusive parameters that would be good to address soon, but that could also be in a follow-up PR.

jschaul:
Compile issue:

```
[4 of 8] Compiling Federator.Impl
                       
/home/user/Documents/git/wire-server/services/federator/src/Federator/Impl.hs:35:9: error:
    • No instance for (http-api-data-0.4.1.1:Web.Internal.HttpApiData.FromHttpApiData
                         (Data.Qualified.Qualified Data.Id.UserId))
        arising from a use of ‘serve’
    • In the expression: serve api (mock api (Proxy @'[]))
      In an equation for ‘app’:
          app _        
            = serve api (mock api (Proxy @'[]))
            where      
                api = Proxy @(ToServantApi API.Api)
   |                   
35 | app _ = serve api (mock api (Proxy @'[]))
   |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

Also, perhaps it's worthwhile to rebase this PR on current develop.

services/brig/src/Brig/API/Public.hs
akshaymankar: I decided to not do this for backwards compatibility. Currently nothing breaks if a client sends both of these and handles get ignored if uids are present. So, maybe it is ok to live with this in the old endpoint.
That being said I agree that it would be nicer to throw an error and since we're making a new endpoint, I will just make the `parseJSON` for the input fail if both are present.

akshaymankar:
Alright, I think stuff should just work now :crossed_fingers: 
Could one of you please have a look at  b06813b and let me know if the approval still stands?

services/brig/src/Brig/API/Public.hs
mheinzel: Ah, you're right. Didn't know it behaved that way, but agree that it's best to keep it for the deprecated endpoint.

mheinzel (approved):
New commit looks good. It still seems like some `mutuallyExclusive` combinator could make the instances clearer, especially if the pattern shows up in multiple places, but I'm fine with it for now.

libs/wire-api/src/Wire/API/User.hs
```diff
@@ -1025,3 +1030,37 @@ instance ToJSON DeletionCodeTimeout where
...
+
+instance FromJSON ListUsersQuery where
+  parseJSON =
+    withObject "ListUsersQuery" $ \o -> do
+      mUids <- ListUsersByIds <$$> o .: "qualified_ids"
+      mHandles <- ListUsersByHandles <$$> o .: "qualified_handles"
+      case (mUids, mHandles) of
+        (Just uids, Nothing) -> pure uids
+        (Nothing, Just handles) -> pure handles
+        (_, _) -> fail "exactly one of qualifie_ids or qualified_handles must be provided."
```
mheinzel, line 1047:
```suggestion
        (_, _) -> fail "exactly one of qualifief_ids or qualified_handles must be provided."
```

jschaul (commented):
Minor typos for the next PR.

PR #1293, opened 2020-12-16
import all charts from wire-server-deploy/develop as of 2012-12-17

jschaul:
As of commit tree

https://github.com/wireapp/wire-server-deploy/tree/f58b59c745f1f0445077a5e69211048c6a9c8587

This is a single clean import commit. For all of the related script changes, see #1294 

Relates to https://wearezeta.atlassian.net/browse/SQPIT-124

PR #1294, opened 2020-12-16
Migrate code for easier CI

jschaul:
Relates to https://wearezeta.atlassian.net/browse/SQPIT-124

Expects #1293 to be available.

Please review as a whole (not commit-by-commit), I will squash all commits upon merging.

* import and adjust scripts and configuration for integration tests from wire-server-deploy and cailleach.
* Add bash and makefile foo for the two main usecases henceforth supported with this PR.

For instructions, see docs/developer/processes.md and/or the makefile documentation


Makefile
```diff
@@ -198,10 +202,86 @@ db-reset:
...
+	if [ "${DOCKER_TAG}" = "local" ]; then \
+	      echo "Environment variable DOCKER_TAG not set to non-default value"; \
+	    exit 1; \
+	fi
+	make chart-$(*)
+
+.PHONY: chart-%
+chart-%:
+	./hack/bin/copy-charts.sh $(*)
+	./hack/bin/set-wire-server-image-version.sh $(DOCKER_TAG)
```
akshaymankar, line 264: Defaulting to `local` for `DOCKER_TAG` is hazardous due to race conditions among ourselves. Maybe we can use `$USER`?

hack/helm_vars/wire-server/values.yaml
```diff
@@ -0,0 +1,225 @@
...
+    host: cassandra-ephemeral
+    replicaCount: 1
+elasticsearch-index:
+  elasticsearch:
+    host: elasticsearch-ephemeral
+    index: directory_test
+  cassandra:
+    host: cassandra-ephemeral
+
+brig:
```
akshaymankar, line 26: I think we should set `imagePullPolicy` to `Always` for all these images because I imagine we won't be changing the tag very often and kubernetes will refuse to pull the image if it sees it present locally.

akshaymankar (approved):
Few nits, but overall looks good. 
:sheep: it

hack/bin/integration-cleanup.sh
```diff
@@ -0,0 +1,34 @@
...
+    date -j -f "%b %d %H:%M:%S %Y" "$date_str" +"%s"
+  else 
+    date -d "$d" +"%s"
+  fi
+}
+
+date_now=$(date +"%s")
+
+# Since helm only shows information as hard-to-parse text pending https://github.com/kubernetes/helm/pull/2950
+# we are stuck with brittle grep/awk magic.
```
akshaymankar, line 19: This issue is resolved btw, maybe we can change this in future.

hack/bin/integration-cleanup.sh
```diff
@@ -0,0 +1,34 @@
...
+  namespace=$(echo "$release" | awk '{print $2}')
+  date_str=$(echo "$release" | awk '{print $4" "$5" "$6}')
+  date_s=$(date_seconds "$date_str")
+  diff=$(( $date_now - $date_s ))
+  if [ $diff -ge 7200 ]; then
+    echo "test release '$name' older than 2 hours; deleting..."
+    helm delete -n "$namespace" "$name"
+  fi
+
+done
```
akshaymankar, line 33: This will delete developer releases too, if we start using the CI cluster (I don't have enough RAM to run HLS, Firefox and kubernetes :sweat_smile: ). Maybe we can rename the cluster names or make this less aggressive.
Anyway, not a big problem since this will probably only happen time to time.

hack/helm_vars/wire-server/values.yaml
jschaul: These values, at the time of this PR, are meant for CI, and CI will bump versions all the time.

I agree this can become an issue when we want to make use of this for local development (i.e. when we compile docker images locally). That is a future PR. Noted in hack/FUTUREWORK.md

hack/bin/sync.sh
```diff
@@ -0,0 +1,125 @@
...
+
+# Synchronize helm charts in git with the hosted version on S3.
+# The contents of /charts are thus made available under
+# https://s3-eu-west-1.amazonaws.com/public.wire.com/charts
+# To use the charts:
+# helm repo add wire https://s3-eu-west-1.amazonaws.com/public.wire.com/charts
+# helm search wire
+
+# This script uses the helm s3 plugin,
+# for more info see https://github.com/hypnoglow/helm-s3
```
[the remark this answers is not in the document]
jschaul: Good point! Renamed.

hack/bin/integration-cleanup.sh
jschaul: 👍🏿 added to FUTUREWORK.

Makefile
jschaul: Good suggestion, thanks!

hack/bin/integration-cleanup.sh
jschaul: My experience is that there these long-lived developer namespaces then hanging around. This is for ephemeral testing only, not for long-lived things. If it gets cleaned up the next kube-integration takes an extra 3 minutes. I think this may be bearable. But we can increase the kubernetes cluster size so we can host 10-15 ephemeral environments at a time, and then make this less agressibe.

PR #1295, opened 2020-12-18
Public end-point for getting the team size.

fisx:
TODO:

- [ ] add metrics for ES response time, number of calls
- [ ] add load-test *and run it*, or at least do some ad-hoc load-testing
- [ ] if needed, see if the ES query can be tuned
- [ ] if needed, throttle the end-point
- [ ] if that doesn't help enough, keep track of team sizes in cassandra with a counter (incremented and decremented as users are added and removed)

### DIFF
--- a/services/brig/src/Brig/Team/API.hs
+++ b/services/brig/src/Brig/Team/API.hs
@@ -169,7 +169,10 @@ routesPublic = do
       .&. capture "tid"
 
   document "GET" "teamSize" $ do
-    Doc.summary "Returns the number of team members as an integer."
+    Doc.summary
+      "Returns the number of team members as an integer.  \
+      \Can be out of sync by roughly the `refresh_interval` \
+      \of the ES index."
     Doc.returns (Doc.ref Public.modelTeamSize)
     Doc.response 200 "Invitation successful." Doc.end
     Doc.response 403 "No permission (not admin or owner of this team)." Doc.end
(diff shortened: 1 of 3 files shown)


services/brig/src/Brig/Team/API.hs
```diff
@@ -156,6 +157,17 @@ routesPublic = do
     Doc.response 404 "No pending invitations exists." Doc.end
     Doc.response 409 "Multiple conflicting invitations to different teams exists." Doc.end
 
+  get "/teams/:tid/size" (continue teamSizePublicH) $
+    accept "application" "json"
+      .&. header "Z-User"
+      .&. capture "tid"
```
akshaymankar, line 163: Can you please add a note about this endpoint returning potentially stale size as it depends on `refresh_interval` of the index which is in seconds (5 seconds in prod, iirc).

services/brig/src/Brig/Team/API.hs
```diff
@@ -156,6 +157,17 @@ routesPublic = do
...
+
+  document "GET" "teamSize" $ do
+    Doc.summary "Returns the number of team members as an integer."
+    Doc.returns (Doc.ref Public.modelTeamSize)
+    Doc.response 200 "Invitation successful." Doc.end
+    Doc.response 403 "No permission (not admin or owner of this team)." Doc.end
```
akshaymankar, line 169: Can you please add a note about this endpoint returning potentially stale size as it depends on `refresh_interval` of the index which is in seconds (5 seconds in prod, iirc).

fisx:
I would like to move this to staging and take a look at how it performs there.  I'm optimstic, but if I'm wrong we can still turn it off before the next production release.

smatting (approved):
Looks good. From what I understand this is just making an existing endpoint available as a public endpoint.
The ES query also looks trivial: a simple count query with a term query

PR #1296, opened 2020-12-18
SCIM: Allow strings for boolean values

smatting:
Hopefully fixes https://wearezeta.atlassian.net/browse/SQSERVICES-154

### DIFF
--- a/libs/hscim/hscim.cabal
+++ b/libs/hscim/hscim.cabal
@@ -4,7 +4,7 @@ cabal-version: 1.12
 --
 -- see: https://github.com/sol/hpack
 --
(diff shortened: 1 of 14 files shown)


smatting:
There is a test with title  `Post emp1 with string "True"` in the SCIM test suite.
It POSTs a new user with a body
```
{
    "userName": "fobarbaz",
    "active": "True",
...
```
Without the changes from this PR, this test fails with `Error in $.active: expected Bool, but encountered String`.
With the changes the server responds with 201, however the test still fails, because the response contains `{"active": false}`, which is causes the test condition to fail. However you get the same response with a boolean literal for "active" in the request. Is this expected?

fisx (approved):
Looks good after a two small changes.  I'll merge once the tests are green.

fisx:
all tests passed locally.

PR #1303, opened 2020-12-28
Fix bug: handle lost after registration of scim-invited users

smatting:
This PR fixes the following bug:

1. SCIM creates an invitation with a user handle
2. The registers with the invitation. The registration function ignores the `handle` value from the invitation and thereby overwrites it with `null` in the users table. However the user and handle are still contained `user_handle` which leads to the issue described in https://wearezeta.atlassian.net/browse/SQSERVICES-158

This bug can be verified by an additional check in the integeration test for SCIM-invitations.

### DIFF
--- a/services/brig/src/Brig/Data/User.hs
+++ b/services/brig/src/Brig/Data/User.hs
@@ -104,8 +104,8 @@ data ReAuthError
   = ReAuthError !AuthError
   | ReAuthMissingPassword
 
-newAccount :: NewUser -> Maybe InvitationId -> Maybe TeamId -> AppIO (UserAccount, Maybe Password)
-newAccount u inv tid = do
+newAccount :: NewUser -> Maybe InvitationId -> Maybe TeamId -> Maybe Handle -> AppIO (UserAccount, Maybe Password)
+newAccount u inv tid mbHandle = do
   defLoc <- setDefaultLocale <$> view settings
   domain <- viewFederationDomain
   uid <-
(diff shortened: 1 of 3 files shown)


services/spar/test-integration/Test/Spar/Scim/UserSpec.hs
```diff
@@ -283,7 +283,7 @@ testCreateUserNoIdP = do
         >>= maybe (error "could not find user in brig") pure
     liftIO $ accountStatus brigUser `shouldBe` Active
     liftIO $ userManagedBy (accountUser brigUser) `shouldBe` ManagedByScim
-
+    liftIO $ userHandle (accountUser brigUser) `shouldBe` Just handle
```
fisx, line 286: for the record, without the fix this fails as expected:

```
  test-integration/Test/Spar/Scim/UserSpec.hs:286:14: 
  1) Spar.Scim.User, POST /Users, team has no SAML IdP, creates a user with PendingInvitation, and user can follow usual invitation process
       expected: Just (Handle {fromHandle = "scimuser_8444630"})
        but got: Nothing
```

:+1:

services/brig/src/Brig/Data/User.hs
```diff
@@ -104,8 +104,8 @@ data ReAuthError
   = ReAuthError !AuthError
   | ReAuthMissingPassword
```
fisx, line 106:
```suggestion

-- | Preconditions:
--
-- 1. @newUserUUID u == Just inv || isNothing (newUserUUID u)@.
-- 2. If @isJust@, @mbHandle@ must be claimed by user with id @inv@.
--
-- Condition (2.) is essential for maintaining handle uniqueness.  It is guaranteed by the
-- fact that we're setting getting @mbHandle@ from table @"user"@, and when/if it was added
-- there, it was claimed properly.
```

fisx (approved):
Nice!

fisx:
Integration tests passed locally.

PR #1305, opened 2020-12-29
Makefile: allow overriding NAMESPACE for kube-integration target

jschaul:
And minor drive-by fixes to bash/Makefile.


jschaul:
🍹 🎄 🎅🏿

PR #1307, opened 2020-12-29
WIP: Document usage of helm charts

jschaul:
What other things should we document in the README? The local setup is explained under docs/developer/ and inside the Makefile as comments.

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,7 +64,7 @@ It also contains
 - **deploy**: (Work-in-progress) - how to run wire-server in an ephemeral, in-memory demo mode
 - **doc**: Documentation
 - **hack**: scripts and configuration for kuberentes helm chart development/releases mainly used by CI
-- **charts**: kubernetes helm charts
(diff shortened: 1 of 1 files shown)


lucendio:
Would it make sense to put a couple of top-level make targets in the README and/or into the docs?

jschaul:
> Would it make sense to put a couple of top-level make targets in the README and/or into the docs?

There is this page: https://github.com/wireapp/wire-server/blob/develop/docs/developer/processes.md which contains some usage instructions.

Do you have suggestions on renaming or adding some things here, or making this more discoverable?

lucendio:
> Do you have suggestions on renaming or adding some things here, or making this more discoverable?

Maybe before discussing how, I should probably describe the use case, to see whether it's a legitimate one:

As a 3rd party I want to checkout a release and generate the charts to deploy them from filesystem instead of the chart repository. As this 3rd party I would expect to have a `./charts/README.md` with the __Build charts__ section.

Does this seem reasonable? What do you think?

jschaul:
Follow-up work tracked in https://wearezeta.atlassian.net/browse/SER-77

PR #1311, opened 2020-12-30
nginz_disco: docker building consistency

jschaul:
Other docker images get their build context from the top level, nginz_disco is inconsistently expecting a subdirectory (and trips up CI).


jschaul:
🎄 🎆

PR #1313, opened 2021-01-04
Allow SCIM to appropriate existing users. WIP

smatting:


smatting:
PR irrelevant as SCIM implementation is already working correctly

PR #1317, opened 2021-01-06
Add helm chart for the federator

akshaymankar:
https://wearezeta.atlassian.net/browse/SQCORE-364

### DIFF
--- a/charts/federator/templates/service.yaml
+++ b/charts/federator/templates/service.yaml
@@ -1,0 +1,19 @@
+apiVersion: v1
+kind: Service
+metadata:
+  name: federator
+  labels:
+    wireService: federator
+    chart: {{ .Chart.Name }}-{{ .Chart.Version | replace "+" "_" }}
+    release: {{ .Release.Name }}
+    heritage: {{ .Release.Service }}
+spec:
+  type: ClusterIP
+  ports:
+    # TODO: Do we name it something else to get the _correct_ SRV record?
+    - name: http
+      port: {{ .Values.service.externalPort }}
+      targetPort: {{ .Values.service.internalPort }}
+  selector:
+    wireService: federator
+    release: {{ .Release.Name }}
--- a/charts/federator/templates/service.yaml
+++ b/charts/federator/templates/service.yaml
@@ -10,7 +10,6 @@ metadata:
 spec:
   type: ClusterIP
   ports:
-    # TODO: Do we name it something else to get the _correct_ SRV record?
     - name: http
       port: {{ .Values.service.externalPort }}
       targetPort: {{ .Values.service.internalPort }}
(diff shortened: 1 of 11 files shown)


charts/federator/templates/service.yaml
```diff
@@ -0,0 +1,19 @@
...
+  name: federator
+  labels:
+    wireService: federator
+    chart: {{ .Chart.Name }}-{{ .Chart.Version | replace "+" "_" }}
+    release: {{ .Release.Name }}
+    heritage: {{ .Release.Service }}
+spec:
+  type: ClusterIP
+  ports:
+    # TODO: Do we name it something else to get the _correct_ SRV record?
```
jschaul, line 13: I don't quite understand this TODO. Naming this port http seems fine to me. Do we use the port name somewhere else?

charts/federator/values.yaml
```diff
@@ -0,0 +1,24 @@
...
+image:
+  repository: quay.io/wire/federator
+  tag: do-not-use
+
+service:
+  externalPort: 8080
+  internalPort: 8080
+
+resources:
+  # FUTUREWORK: come up with numbers which didn't appear out of thin air
```
jschaul, line 12: I think we can probably leave the resources block entirely empty: `resources: {}`.
But I'm also okay to follow the other charts and to make a change for all charts at once regarding resource usage, instead of only for the federator.

charts/wire-server/values.yaml
```diff
@@ -9,3 +9,4 @@ tags:
   team-settings: false
   account-pages: false
   legalhold: false
+  federator: false
```
jschaul, line 12: Do we want to disable installing this compoent by default, and enable it once the implementation is more "ready"? Or do you want to keep this false also going forward?

jschaul (approved):
Some questions, but in general looks good and fine to merge if CI is green.

charts/wire-server/values.yaml
akshaymankar: Yes, I was thinking we shouldn't enable it by default at least until it is useful. Even when it is useful, maybe we shouldn't deploy it when federation is not enabled.

charts/federator/templates/service.yaml
akshaymankar: When we call this port `http`, kubernetes will create SRV record as `_http._tcp.federator.<namespace>.cluster.local`.
I assume for discovering remote federators, we will rely on an SRV record like `_wire-federator._tcp.<domain>`.
So, if we rename this port from http to `wire-federator` and configure domain as `federator.<namespace>.cluster.local` we will be able to test federation in same kubernetes cluster without having to mess with global DNS records. All of this is obviously not thought through, so I would much rather tackle it later. I will remove this TODO and write this information in some issue in JIRA.

charts/federator/values.yaml
akshaymankar: Yeah, let's do them all together.

PR #1318, opened 2021-01-06
SCIM: Add error logging for CRUD operations with hashed context

smatting:
This PR adds error logging in spar. FUTUREWORK: use structured logs instead of strings.


fisx (approved):
Add a `logScimError` for get and (as you say) a simliar thing for the manual invite via team settings, and we can ship this.

PR #1319, opened 2021-01-06
Use mu-haskell to implement one initial federation request across backends

jschaul:
See https://github.com/wireapp/wire-server/blob/db4c2351476c713f0367cee635faaaea10f9adf5/docs/reference/federation/pull-requests/1319_initial_federation_request_across_backends.md

This PR might be best viewed with a filter only for .hs and .proto files.


fisx (commented):
I'll keep reviewing later, probably tonight.

libs/dns-util/src/Wire/Network/DNS/Helper.hs
```diff
@@ -0,0 +1,29 @@
...
+import Imports
+import Network.DNS
+
+-- | Set up a thread-safe resolver with a global cache. This means that SRV
+-- records will only be re-resolved after their TTLs expire
+mkDnsResolver :: IO Resolver
+mkDnsResolver = do
+  let resolvConf = defaultResolvConf {resolvCache = Just defaultCacheConf}
+  resolvSeed <- makeResolvSeed resolvConf
+  withResolver resolvSeed $ \resolver -> return resolver
```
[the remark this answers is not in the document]
akshaymankar: I didn't check this. But I agree this isn't very futureproof. I will try to make this function go away.

libs/dns-util/src/Wire/Network/DNS/Helper.hs
```diff
@@ -0,0 +1,29 @@
...
+--
+-- You should have received a copy of the GNU Affero General Public License along
+-- with this program. If not, see <https://www.gnu.org/licenses/>.
+
+module Wire.Network.DNS.Helper where
+
+import Imports
+import Network.DNS
+
+-- | Set up a thread-safe resolver with a global cache. This means that SRV
```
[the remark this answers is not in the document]
akshaymankar: Resolved with ccb52b9

libs/dns-util/src/Wire/Network/DNS/Helper.hs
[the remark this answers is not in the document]
akshaymankar: Let me know what you think of this: ccb52b9

libs/wire-api-federation/proto/router.proto
```diff
@@ -0,0 +1,74 @@
...
+
+// the federator will take out LocalCall from the RemoteCall Envelope, and
+// forward the LocalCall object to a federator on another backend (with the wish
+// that this remote federator forwards this to its own local brig)
+message LocalCall {
+  Component component = 1;
+  Method method = 2;
+  bytes path = 3;
+  repeated QueryParam query = 4;
+  bytes body = 5;
```
fisx, line 41: So GRPC is use to encode arbitrary HTTP requests?  Will it be feasible to elegantly encode GRPC requests to brig once that is a thing, rather than rest/json?

I was expecting to see a higher-level protocol encoded here, but I'm beginning to feel this may have been a better idea.  I should probably read the documentation and see if there is a discussion about this giving the reasons for the decision.  Feel free to drop me a link, though.  :)

libs/wire-api-federation/proto/router.proto
```diff
@@ -0,0 +1,74 @@
...
+enum Method {
+  GET = 0;
+  POST = 1;
+  HEAD = 2;
+  PUT = 3;
+  DELETE = 4;
+  TRACE = 5;
+  CONNECT = 6;
+  OPTIONS = 7;
+  PATCH = 8;
```
fisx, line 58: Will we be able to add `CUSTOM` in the future if that should become required?  I suppose we'll always have the protocol versioning problem in that case no matter what we do.  So, can you add version fields to these types so that remote federators can respond with a required version, rather than failing in some unspecified way?  Seems like a cheap change.

libs/wire-api-federation/proto/router.proto
```diff
@@ -0,0 +1,74 @@
...
+}
+
+// The * part of the following network traffic:
+// Brig *-> federator -> federatpr -> brig
```
fisx, line 62:
```suggestion
// brig *-> federator -> federator -> brig
```

fisx, line 62: Also, are you sure you haven't flipped the comments for `RouteToRemote` and `RouteToInternal`?  If no I'm indeed confused by the names.  If yes then I think the names are fine.

fisx (commented):
General question: can we move https://github.com/wearezeta/documentation/blob/master/topics/federation/federation-design.md to wire-server/docs now?  Or to wire-docs?  The documentation repo is deprecated by confluence anyway.

More review chunks coming up!

libs/wire-api-federation/proto/router.proto
akshaymankar: > Will we be able to add CUSTOM in the future if that should become required? 

Yes, we will be able to add `CUSTOM` if we need to. In that case a newer server sending `CUSTOM` requests would just be told that it is talking non-sense by an older server.

>  I suppose we'll always have the protocol versioning problem in that case no matter what we do. So, can you add version fields to these types so that remote federators can respond with a required version, rather than failing in some unspecified way? Seems like a cheap change.

Yes, there is a problem of verifying the version of the protocol, which will have to be figured out later. But the way I see it happening is probably by some protocol negotiation and handling at feature level as to what we can do when talking to an older backend. Similar things might have to happen when talking to a newer backend.

libs/wire-api-federation/proto/router.proto
akshaymankar: > So GRPC is use to encode arbitrary HTTP requests? Will it be feasible to elegantly encode GRPC requests to brig once that is a thing, rather than rest/json?

Yes, it should be fairly simple as in GRPC you just have to say name of the service, name of the procedure and pass the data in. Supporting streaming services might be more interesting, but I can't think of any use case for streaming services.

> I was expecting to see a higher-level protocol encoded here, but I'm beginning to feel this may have been a better idea. I should probably read the documentation and see if there is a discussion about this giving the reasons for the decision. Feel free to drop me a link, though. :)

Initially we were going to encode higher level protocol here, if by that you mean stuff like "getUserByHandle(handle)". But more I thought about it more it didn't make sense to have to repeat all of this everywhere (serving brig, calling from brig, serving from here and calling from here). Plus that would add testing burden, as in for each thing we add we won't be sure it works until it is tested across two backends. This seemed like we were tangling routing concerns (find a remote and talk to it) with application logic concerns (getUserByHandle), which I think was opposite of why we wanted a federator. Trying to separate these things out until writing federated code was as simple as possible got us here (It can still be simpler, but its a start). Unfortunately, we didn't write anything about this, we plan to do this soon. We're hoping that it will go in the `docs/reference` directory.

libs/wire-api-federation/proto/router.proto
akshaymankar: Hehe, we haven't flipped them. These services desperately need better names.

libs/wire-api-federation/proto/router.proto
```diff
@@ -0,0 +1,74 @@
...
+// the envelope thing which is sent from brig to local federator
+message RemoteCall {
+  string domain = 1;
+  LocalCall localCall = 2;
+}
+
+// the federator will take out LocalCall from the RemoteCall Envelope, and
+// forward the LocalCall object to a federator on another backend (with the wish
+// that this remote federator forwards this to its own local brig)
+message LocalCall {
```
jschaul, line 36: My suggestions for renaming LocalCall and RemoteCall:

RemoteCall --> QualifiedRequest
LocalCall --> UnqualifiedRequest

libs/wire-api-federation/proto/router.proto
```diff
@@ -0,0 +1,74 @@
...
+// brig *-> federator -> federator -> brig
+// FUTUREWORK(federation): better naming?
+service RouteToRemote {
+  rpc call (RemoteCall) returns (Response);
+}
+
+// The * part of the following network traffic:
+// Brig -> federator *-> federator -> brig
+// FUTUREWORK(federation): better naming?
+service RouteToInternal {
```
jschaul, line 71: Could we simply call these services InternalService and ExternalService ? Or InternalFederator and ExternalFederator? This would match the port naming; and brig would talk to InternalFederator and give it a QualifiedRequest, and then the federator would send an UnqualifiedRequest to the ExternalFederator service of a federator of another backend.

libs/wire-api-federation/proto/router.proto
akshaymankar: Yeah, sounds great, it will be consistent with rest of the things.

libs/wire-api-federation/proto/router.proto
akshaymankar: I like InternalFederator and ExternalFederator. I think it will still be confusing for somebody new to this because it is never clear if InternalFederator or InternalService means that it is routing **from** internal things or **to** internal things. But it will at least be consistent and there will be fewer confusing words in the domain.

jmatsushita: When this question arises, I usually find that the easiest to understand is to name things from the perspective of the code, or the service running the code.
 - I understand this is an interface, so technically it runs on both ends, but I think the diagram is clear that the Federator is the star of the show. 
 - Also I think it's best to keep with one set of terms either `Local/Remote` or `InternalExternal`, since `Local/Remote` is usually for "not across some network" or "across some network" and for microservices everything is "across some network", I would go for `Internal/External`. 

So something like `FromInternalService`/ `InternalServiceRoute` or `InternalServiceRPC` (my favorite) and `ToExternalFederator` or `RemoteFederatorRoute` or `RemoteFederatorRPC` (my favorite) would be easiest for me to understand.

services/brig/src/Brig/Federation/Client.hs
```diff
@@ -0,0 +1,76 @@
...
+callRemote fedClient call = liftIO $ gRpcCall @'MsgProtoBuf @Proto.RouteToRemote @"RouteToRemote" @"call" fedClient (Proto.validatedRemoteCallToRemoteCall call)
+
+expectOk :: GRpcReply Proto.Response -> Handler Proto.HTTPResponse
+expectOk = \case
+  GRpcTooMuchConcurrency _tmc -> throwStd $ notFound "Too much concurrency"
+  GRpcErrorCode errCode -> throwStd $ notFound $ LT.pack $ "GRPCError: " <> show errCode
+  GRpcErrorString errStr -> throwStd $ notFound $ LT.pack $ "GRPCError: " <> show errStr
+  GRpcClientError clErr -> throwStd $ notFound $ LT.pack $ "GRPC ClietnError: " <> show clErr
+  GRpcOk (Proto.ResponseErr err) -> throwStd $ notFound $ LT.pack $ "Remote component errored: " <> show err
+  GRpcOk (Proto.ResponseHTTPResponse res) -> pure res
```
akshaymankar, line 76: All of this code is only exercised in the test which goes between two Backends. This is not ideal, we should figure out a way to test client side of federated code without needing another backend. We could do this either by mocking the second backend in integration tests or making all of independent of the `Handler` monad and write unit tests.

jschaul: Made this comment into a futurework in https://github.com/wireapp/wire-server/pull/1319/commits/311cc14b7000bba4c28beced6e8868e92f984815

libs/wire-api-federation/proto/router.proto
jschaul: renamed as discussed.

libs/wire-api-federation/proto/router.proto
jschaul: renamed as discussed.

libs/wire-api-federation/proto/router.proto
jschaul: renamed as discussed.

libs/wire-api-federation/proto/router.proto
jschaul: I added some comments to explain in https://github.com/wireapp/wire-server/pull/1319/commits/c2667aef77b3d5ad6868f6a6fd80bbe9f47fa12b - Akshay, feel free to add something there, or to docs/reference/federation.

akshaymankar (approved):
As decided in the call, let's merge this.
@fisx (or others) feel free to continue to review this, let's try to incorporate the changes in further PRs.

PR #1323, opened 2021-01-12
Fix bug: SCIM-Deleting a user retains the externalId

smatting:
This PR fixes this bug:
Deleting a user via SCIM when there is no IdP won't remove the entry from `spar.scim_external_ids`.
This might be a cause for issues such as [SQSERVICES-179](https://wearezeta.atlassian.net/browse/SQSERVICES-179)

The test `testDeletedUsersFreeExternalIdNoIdp` proves the bug and the fix.

The function `userToExternalId` was removed because its name is misleading. The only call site of this function is refactored.

### DIFF
--- a/services/spar/src/Spar/Intra/Brig.hs
+++ b/services/spar/src/Spar/Intra/Brig.hs
@@ -20,11 +20,10 @@
 -- | Client functions for interacting with the Brig API.
 module Spar.Intra.Brig
   ( veidToUserSSOId,
-    veidFromUserSSOId,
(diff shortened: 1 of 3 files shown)


smatting:
All `spar` integration tests pass locally.

smatting:
All spar and brig integration tests pass locally after merge.

smatting:
Failing test is flaky un most likely unrelated https://wearezeta.atlassian.net/browse/SQSERVICES-189

PR #1325, opened 2021-01-13
 sftd: add support for multiple SFT servers

arianvp:
* The ingress assigns an SFT allocation request to a random SFT
* Each sftd pod is made aware of an URL on which it is directly
reachable, and will return the URL in the response to the client.  e.g.
Pod `sftd-0` will be assigned `https://sft.example.com/sfts/sftd-0`
* The client tells this URL to other clients willing to join the call
* Other clients make a request to this URL
* The ingress points requests to `/sfts` to the `join-call` deployment,
which will redirect to the specific pod, such that the client can join
the conference call of the other client

### DIFF
--- a/charts/sftd/README.md
+++ b/charts/sftd/README.md
@@ -1,5 +1,6 @@
 # SFTD Chart
 
+
 ## Deploy
 
 Replace `example.com` with your own domain here.
@@ -37,6 +38,31 @@ server.
 
 You should configure `brig` to hand out the SFT server to clients by setting
 `brig.optSettings.setSftStaticUrl=https://sftd.example.com:443` on the `wire-server` chart
+
+## Parameters
+
+Please see [values.yaml](./values.yaml) for an overview of parameters that can be configured.
+
+
+## Rollout
+
+Kubernetes will shut down pods and start new ones when rolling out a release. Any calls
+that were in progress on said pod will be terminated and will cause the call to drop.
+
+Kubernetes can be configured to wait for a certain amount of seconds before
+stopping the pod. During this timeframe new calls wil not be initiated on the
+pod, but existing calls will also not be disrupted.  If you want to roll out a
+release with minimal impact you can set the
+[`terminationGracePeriodSeconds`](./values.yaml#L18) option to the maximum
+length you want to wait before cutting off calls.
+
+Currently due to the fact we're using a `StatefulSet` to orchestrate update
+rollouts, and `StatefulSet`s will not replace all pods at once but instead
+one-for-one, a rollout of a release will take `oldReplicas * terminationGracePeriodSeconds`
+to complete.
+
+We might switch to using a `Deployment` for `sftd` in the future, to reduce this time to just `terminationGracePeriodSeconds`.
+
 
 
 ## Multiple sftd deployments in a single cluster
--- a/charts/sftd/README.md
+++ b/charts/sftd/README.md
@@ -18,7 +18,7 @@ helm install sftd wire/sftd  \
 Using Cert-manager:
 ```
 helm install sftd wire/sftd \
-  --set host=example.com \
+  --set host=sftd.example.com \
   --set allowOrigin=https://webapp.example.com \
   --set tls.issuerRef.name=letsencrypt-staging
 ```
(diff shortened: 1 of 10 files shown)


charts/sftd/templates/ingress.yaml
```diff
@@ -17,7 +17,11 @@ spec:
     - host: "{{ .Values.host }}"
       http:
         paths:
-          - path: /
+          - path: /sft/
```
arianvp, line 20: This requires change to the brig chart

lucendio (changes requested):
It's a very interesting approach. Amazing work! Some nits and concerns on my end. Not necessarily related to the approach though.
I still have some questions, but that's something for a coffee.

charts/sftd/README.md
```diff
@@ -38,6 +39,65 @@ server.
...
+## Rollout
+
+Kubernetes will shut down pods and start new ones when rolling out a release. Any calls
+that were in progress on said pod will be terminated and will cause the call to drop.
+
+Kubernetes can be configured to wait for a certain amount of seconds before
+stopping the pod. During this timeframe new calls wil not be initiated on the
+pod, but existing calls will also not be disrupted.  If you want to roll out a
+release with minimal impact you can set the
+[`terminationGracePeriodSeconds`](./values.yaml#L18) option to the maximum
```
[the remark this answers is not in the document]
arianvp: I implement a `preStop` already. (You don't see it in the diff as it was already in the code before this PR):

https://github.com/wireapp/wire-server/pull/1325/files#diff-686f0e1005372025c5d72c982348d8798723feb89559827df231e1180f0947b1L99-L105

lucendio (approved):

charts/sftd/README.md
```diff
@@ -1,5 +1,8 @@
 # SFTD Chart
 
+In theory the `sftd` chart can be installed on its own, but it's usually
+installed as part of the `wire-server` umbrella chart.
```
jschaul, line 4: Just a thought: I understand it might be nice to just say "helm install wire-server", though often "more complex" services (stateful ones or those with slower update cycles like SFT due to e.g. not wanting to cut ongoing calls out for 6 hours or so) being bundled into an otherwise fast-to-upgrade chart may slow any wire-server upgrades down going forward. I presume some scripts and/or helmfiles to be in use anyway, as more than a single chart needs to be installed.
Not sure if there are benefits otherwise that I might be missing? Or do I overestimate potential downsides?

(feel free to go ahead with the wrapper chart approach, just wondering about the motivations)

arianvp: I'm really on the fence between the two options.  It was originally the way you just described and I only changed it today... 

Motivation was:

I wanted to do a bit of refactoring in the `wire-server` umbrella chart in a follow-up PR so that there is a `globals.baseDomain` option that all charts can use to remove all the domain duplication in our `values.yml`. As a form of experiment I wanted to make the SFT chart (which needs to know the base domain; and the domain name of the webapp chart) use this method of configuration to set its `host` and `allowOrigin` options.

Also `helm upgrade --wait` only waits for new pods to be spawned; it doesn't wait for old pods to be terminated; so even with a slow update cycle (due to setting `terminationGracePeriodSeconds` to a few hours) upgrades would still be "snappy". 

However now I just realised that's only true for `Deployment`; not for `StatefulSet`; as `StatefulSet` replaces pods one by one (See Future Work section at the bottom of the README).

I'm going to sleep on it and revisit tomorrow. I might end up reverting this again.

arianvp: I was inspired to structure stuff a bit like:

https://gitlab.com/gitlab-org/charts/gitlab
https://docs.gitlab.com/charts/charts/globals.html#configure-host-settings

arianvp: @jschaul I have made it clear in the docs that both ways are possible. `docs.wire.com` currently still instructs to not use the umbrella chart

jschaul: > @jschaul I have made it clear in the docs that both ways are possible. docs.wire.com currently still instructs to not use the umbrella chart

Great! Thanks!

PR #1328, opened 2021-01-14
Add fast-intermediate Dockerfile for faster PR CI

jmatsushita:
Testing make clean fast and new wire-server-pr-fast.yml file for our internal CI.


jmatsushita:
Could you please approve this PR so that the `wire-server-pr-fast` pipeline can run on all PRs and we can see if it's reliable?

hack/FAST
```diff
@@ -0,0 +1 @@
+1
```
akshaymankar, line 1: What does this file do?

jschaul (approved):

hack/FAST
jmatsushita: Just a way to bump the `wire-server-pr-fast` pipeline and trigger it since the `build` folder where the Dockerfile is is ignored.

PR #1329, opened 2021-01-14
Extract and rename PolyLog to a library for reusability

akshaymankar:
polysemy-wire-zoo is inspired by polysemy-zoo. We can use it to keep our pets
which we might need to interface with other libraries.

I named the constructor `Polylog` so `makeSem` will generate a function called `polylog`. Ideally I would have named it `Log` and the function would've been `log`, but that would collide with `Imports`.


akshaymankar:
I extracted this so it could be reused in the Federator

jschaul (approved):

PR #1330, opened 2021-01-14
Accept any query string for externalId

smatting:
This PR will update the behaviour of `GET /scim/v2/Users?filter=externalId eq "<x>"`. Without this PR if `<x>` is not en email the endpoint will return a `400`. With this PR it will return an empty result set.

For context: https://wearezeta.atlassian.net/browse/SQSERVICES-204

### DIFF
--- a/services/spar/src/Spar/Scim/User.hs
+++ b/services/spar/src/Spar/Scim/User.hs
@@ -763,6 +763,12 @@ scimFindUserByHandle mIdpConfig stiTeam hndl = do
 -- successful authentication with their SAML credentials.
 scimFindUserByEmail :: Maybe IdP -> TeamId -> Text -> MaybeT (Scim.ScimHandler Spar) (Scim.StoredUser ST.SparTag)
 scimFindUserByEmail mIdpConfig stiTeam email = do
+  -- Azure has been observed to search for externalIds that are not emails, even if the
+  -- mapping is set up like it should be.  This is a problem: if there is no SAML IdP, 'mkValidExternalId'
+  -- only supports external IDs that are emails.  This is a missing feature / bug in spar tracked in
+  -- https://wearezeta.atlassian.net/browse/SQSERVICES-157; once it is fixed, we should go back to
+  -- throwing errors returned by 'mkValidExternalId' here, but *not* throw an error if the externalId is
+  -- a UUID, or any other text that is valid according to SCIM.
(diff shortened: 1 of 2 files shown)


jschaul (commented):
Any change to the API should also have a test added that makes use of that API functionality to A) document the desired behaviour and B) avoid future regressions.

services/spar/src/Spar/Scim/User.hs
```diff
@@ -762,7 +763,7 @@ scimFindUserByHandle mIdpConfig stiTeam hndl = do
 -- successful authentication with their SAML credentials.
 scimFindUserByEmail :: Maybe IdP -> TeamId -> Text -> MaybeT (Scim.ScimHandler Spar) (Scim.StoredUser ST.SparTag)
 scimFindUserByEmail mIdpConfig stiTeam email = do
-  veid <- mkValidExternalId mIdpConfig (pure email)
+  veid <- MaybeT (either (const Nothing) Just <$> runExceptT (mkValidExternalId mIdpConfig (pure email)))
```
jschaul, line 766: It would also be good to add a comment here (or in the test) why not sending an email is accepted here, linking to the relevant documentation.
Our API should not be microsoft compatible and break with other providers (or have less good guarantees) without that explicitly documented in the code.

jschaul (approved):
(comment would be nice, otherwise looks good!)

services/spar/src/Spar/Scim/User.hs
fisx: I agree, there has to be a comment here.  I would write:

```suggestion
  -- Azure has been observed to search for externalIds that are not emails, even if the
  -- mapping is set up like it should be.  This is a problem: if there is no SAML IdP, 'mkValidExternalId'
  -- only supports external IDs that are emails.  This is a missing feature / bug in spar tracked in
  -- https://wearezeta.atlassian.net/browse/SQSERVICES-157; once it is fixed, we should go back to
  -- throwing errors returned by 'mkValidExternalId' here, but *not* throw an error if the externalId is
  -- a UUID, or any other text that is valid according to SCIM.
  veid <- MaybeT (either (const Nothing) Just <$> runExceptT (mkValidExternalId mIdpConfig (pure email)))
```

fisx (approved):
approved, but please add the something along the lines of the comment i suggested.

PR #1334, opened 2021-01-15
Brig does not always fill user.sso_id correctly.

fisx:
https://wearezeta.atlassian.net/browse/SQSERVICES-207

(I am now running the tests, I *expect* this to fail, but let's see.)


fisx:
yeay :) 

```
Failures:

  test-integration/Test/Spar/Scim/UserSpec.hs:256:14: 
  1) Spar.Scim.User, POST /Users, team has no SAML IdP, creates a user with PendingInvitation, and user can follow usual invitation process
       expected: Just (UserScimExternalId "success+0601014f-6d0d-46fc-82ff-f58cdd1ef2a6@simulator.amazonses.com")
        but got: Nothing
```

so yeah, not necessarily on monday, but that should be fixed, too.  and may require another database migration.

smatting:
I think this fixes the bug. We should however add a data migration that fixes the missing data for exisitng teams.

services/spar/test-integration/Test/Spar/Scim/UserSpec.hs
```diff
@@ -256,7 +256,6 @@ testCreateUserNoIdP = do
     liftIO $ accountStatus brigUserAccount `shouldBe` PendingInvitation
     liftIO $ userEmail brigUser `shouldBe` Just email
     liftIO $ userManagedBy brigUser `shouldBe` ManagedByScim
-    liftIO $ userSSOId brigUser `shouldBe` Just (UserScimExternalId (fromEmail email))
```
fisx: Don't remove, but check that it's `Nothing`.  If that doesn't work either, leave a FUTUREWORK.

services/brig/src/Brig/API/User.hs
```diff
@@ -207,9 +207,16 @@ createUser new = do
             Nothing
           )
         Just existingAccount ->
-          let ident = newIdentity email phone (newUserSSOId new)
-              existingUser = accountUser existingAccount
-           in ( new {newUserManagedBy = Just (userManagedBy existingUser), newUserIdentity = ident},
+          let existingUser = accountUser existingAccount
+              mbSSOid =
+                case (teamInvitation, email, userManagedBy existingUser) of
+                  (Just _, Just em, ManagedByScim) ->
```
fisx, line 213: leave a comment that `isJust teaminvitation` implies that there is no idp?

services/spar/test-integration/Test/Spar/Scim/UserSpec.hs
smatting: Done

services/brig/src/Brig/API/User.hs
smatting: Done

smatting (approved):
LGTM

PR #1339, opened 2021-01-25
Simplify ES upgrade path

smatting:
This PR updates the `elasticsearch-index` chart to simplify the the ES migration path introduced in #1360.

This PR adds an optional config flag `--delete-template TEMPLATE_NAME`  to `brig-index` that deletes an index template `TEMPLATE_NAME` before it creates any index. If `--delete-template` isn't set then the bahaviour of `brig-index` is unchanged.

### DIFF
--- a/docs/reference/elasticsearch-migration-2021-02-15.md
+++ b/docs/reference/elasticsearch-migration-2021-02-15.md
@@ -2,13 +2,28 @@
 
 Release `2021-02-15` of `wire-server` requires creating a new ElasticSearch index for `brig` _before_ deploying the release. Without this new index the user search in TeamSettings will be defunct.
 
-The index that brig is using, is defined at brig's config at `elasticsearch.index`. This config value of the previous deployment is referred to as `<OLD_INDEX>` in the following.
+The index that brig is using, is defined in the `brig` chart's config at `elasticsearch.index`. This config value of the previous deployment is referred to as `<OLD_INDEX>` in the following.
 
 These following instructions describe how to:
 
 1. Create the new index `<NEW_INDEX>` (you can choose any name that is different from `<OLD_INDEX>`)
 2. Populate `<NEW_INDEX>` with brig's data.
 3. Configure brig to use the new index.
+
+Note: The following steps require downtime of brig. If downtime is not acceptable then please skip to section [Create the new index without downtime](#create-the-new-index-without-downtime)
+
+1. Update config for the `elasticsearch-index` chart:
+    - Set `elasticsearch.index` to `<NEW_INDEX>`
+2. Update config for the `brig` chart:
+    - Set `config.elasticsearch.directory` to `<NEW_INDEX>`
+3. Redeploy `wire-server`.
+   The `elasticsearch-index` chart will automatically create and populate `<NEW_INDEX>`
+   before all other services are started. This might take several minutes depending on the
+   number of users.
+   
+You can delete `<OLD_INDEX>` index from ES now.
+
+## Create the new index without downtime
 
 In the following we assume the following env vars are set.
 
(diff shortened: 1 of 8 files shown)


smatting:
Tested the script locally

services/brig/src/Brig/User/Search/Index.hs
```diff
@@ -275,6 +275,12 @@ createIndex' failIfExists settings shardCount = liftIndexIO $ do
     throwM (IndexError "Index already exists.")
   unless ex $ do
     let fullSettings = settings ++ [ES.AnalysisSetting analysisSettings]
+    let templateName = ES.TemplateName "directory"
+    tExists <- ES.templateExists templateName
```
akshaymankar, line 279: It seems dangerous to hardcode the name. Can you check how to the index was added in the first place and add removal code there? Maybe there was some option to name it so we could use that instead.

Code like this should also have a limit as to how long it should be here for. We should add a comment to remind us to delete it after some point (in Jan 2022?).

smatting: I think I found the latest reference to this template here: https://github.com/zinfra/hegemony/pull/1809

fisx:
@akshaymankar would you prefer to log an error in case the template exists, and ask the admin to remove it manually?

i agree that a comment should be added to that line referencing zinfra/hegemony#1809 (or maybe a relevant place in wire-server) and saying it's ok to delete this after jan'22.

smatting:
Closing this, because I've put the instruction to delete into hegemonys CHANGELOG, that should propagate to the release notes.

smatting:
I locally tested that `brig-index create` deletes the template if it exists.

services/brig/src/Brig/User/Search/Index.hs
```diff
@@ -267,14 +269,23 @@ createIndex' ::
...
   m ()
-createIndex' failIfExists settings shardCount = liftIndexIO $ do
+createIndex' failIfExists settings shardCount mbTemplate = liftIndexIO $ do
   idx <- asks idxName
   ex <- ES.indexExists idx
   when (failIfExists && ex) $
     throwM (IndexError "Index already exists.")
   unless ex $ do
     let fullSettings = settings ++ [ES.AnalysisSetting analysisSettings]
+
```
akshaymankar, line 281: Can you please add a comment here explaining why we do this and maybe drop a link to an issue which will delete this code?

services/brig/src/Brig/User/Search/Index.hs
```diff
@@ -257,6 +258,7 @@ createIndex ::
   [ES.UpdatableIndexSetting] ->
   -- | Number of shards
   Int ->
+  Maybe ES.TemplateName ->
```
akshaymankar, line 261: Can you please add a comment explaining that this template name is for a template to be deleted, not to be created as the name of the function `createIndex` might imply something like that.

services/brig/src/Brig/Index/Options.hs
```diff
@@ -95,12 +97,13 @@ makeLenses ''CassandraSettings
 
 makeLenses ''ReindexFromAnotherIndexSettings
 
-mkCreateIndexSettings :: ElasticSettings -> ([ES.UpdatableIndexSetting], Int)
+mkCreateIndexSettings :: ElasticSettings -> ([ES.UpdatableIndexSetting], Int, Maybe ES.TemplateName)
```
akshaymankar, line 100: This tuple was already weird now it has 3 things, I think it is worth creating a type with appropriate names for these things. But I don't feel too strongly of this, hoping that it will be deleted at some point.

akshaymankar (changes requested):
Apart from the the nits in comments. I think we should also make changes to the helm chart in this PR to allow specifying the name of the template. Can you please add that?

services/brig/src/Brig/User/Search/Index.hs
smatting: Done

services/brig/src/Brig/User/Search/Index.hs
smatting: Done

services/brig/src/Brig/Index/Options.hs
smatting: Created a data type for this

smatting:
I added a `elasticsearch.delete_template` template var with default value `directory`.

I've tested that the upgrade path works by updating `elasticsearch-index.elasticsearch.index` and `brig.config.elasticsearch.index` to a new index in `hack/helm_vars/value.yml`. The helm release reached the `deployed` state and I checked the index has being created.

I manually created a index template then repeated the procecudure with a different index name. The logs showed
```
, Delete index template "directory"
I, 200 - {"acknowledged":true}
I, Create index
I, 200 - {"acknowledged":true,"shards_acknowledged":true,"index":"directory_test_new_3"}
I, Put mapping
I, 200 - {"acknowledged":true}
I, Put mapping
I, 200 - {"acknowledged":true}
```
and the helm release reached `deployed` state again.

akshaymankar (approved):
Looks good :shipit:

docs/reference/elasticsearch-migration-2021-02-15.md
```diff
@@ -2,14 +2,29 @@
...
-The index that brig is using, is defined at brig's config at `elasticsearch.index`. This config value of the previous deployment is referred to as `<OLD_INDEX>` in the following.
+The index that brig is using, is defined in the `brig` chart's config at `elasticsearch.index`. This config value of the previous deployment is referred to as `<OLD_INDEX>` in the following.
 
 These following instructions describe how to:
 
 1. Create the new index `<NEW_INDEX>` (you can choose any name that is different from `<OLD_INDEX>`)
 2. Populate `<NEW_INDEX>` with brig's data.
 3. Configure brig to use the new index.
 
+Note: The following steps require downtime of brig. If downtime is not acceptable then please skip to section [Create the new index without downtime](#create-the-new-index-without-downtime)
```
arianvp, line 13: Why would this cause downtime of brig?  I think `elasticsearch-index` gets run and then and only then `brig` will get a rolling release.  (How currently the chart is set up)

smatting: Ah yes, this note is misleading. There is no downtime. If you don't shut down brig before the migration then you risk having missing user updates of brig in ES.

smatting: I think there are 2 ways I could update these instructions:

1) Prepend a step "Shut down brig."
OR
2) Remove the note about downtime and append a step to the instructions
    "Reindex brig's data again by ... This makes sure that the ES index and brig's data are fully in sync"

Which one would you prefer? Or do you see any alternative?

PR #1341, opened 2021-01-26
Federation end2end test scripts and Makefile targets

jschaul:
For easier review, this PR is a small piece from the now-long-running "federation" branch https://github.com/wireapp/wire-server/pull/1319 with the following purpose:

* introduces telepresence as a new development dependency (right now only for running end2end federation tests spanning multiple backends)
* prepares using brig as a place to keep some federation end2end tests spanning more than one backend, by using pattern matching on `brig-federation`: those tests will not run by default locally or on CI, until they are more stable and CI can actually run them.
* Makefile targets and bash scripts to spawn up two backends (one backend will have a namespace-suffix-by-convention)
* Each ephemeral backend will have a unique federationDomain (based on the kubernetes namespace), so that tests spanning multiple backends become possible.

(if you want to try this out for real, the following can be run:

```sh
git fetch
git checkout mu # yeah, not everything is available on this branch yet)
export NAMESPACE=joe-dev # or any other namespace as long as kubernetes still has enough ram/cpu
export DOCKER_TAG=0.0.1-pr.3300
make kube-integration-setup-federation # sets up two backends
make kube-integration-federation # runs one test using telepresence
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -218,6 +218,7 @@ libzauth:
 # Run this again after changes to libraries or dependencies.
 .PHONY: hie.yaml
 hie.yaml:
+	stack build implicit-hie
 	stack exec gen-hie > hie.yaml
 
 #####################################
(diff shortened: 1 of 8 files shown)


akshaymankar (approved):
Looks good, this PR ideally should've just been merged with #1342 some assumptions here are only programmed there.

smatting (approved):

PR #1342, opened 2021-01-26
Brig integration tests

jschaul:
Brig integration/end2end initial test for federation

For easier review, this PR is a small piece from the now-long-running "federation" branch #1319 with the following purpose:

* refactor some brig-integration setup code (remove looking at environment
  variables)
* brig now knows about federator's existence in the config file. This is
  an optional field. (should it be required?)
* introduce a first test that needs two backends. This test is failing
  on this branch (but working to success on #1319 currently), but will
  be disabled-by-default to run both on CI and locally, as long as this PR
  is merged *after* #1341
* prepare some federation chart config. Note there is currently a `federator` and `federatorInternal` config here: the former can be deleted as soon as some PRs involving the actual federator have made their way in, but it's kept in here to keep CI happy for the time being.

### DIFF
--- a/charts/brig/templates/configmap.yaml
+++ b/charts/brig/templates/configmap.yaml
@@ -44,6 +44,15 @@ data:
       host: gundeck
       port: 8080
 
+    # TODO remove this
+    federator:
+      host: federator
+      port: 8080
+
+    federatorInternal:
+      host: federator
+      port: 8080
+
     {{- with .aws }}
     aws:
       prekeyTable: {{ .prekeyTable }}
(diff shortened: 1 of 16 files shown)


akshaymankar (approved):

smatting (approved):

jschaul:
Thanks for the review. Just waiting on CI to make sure it passes before merging.

jschaul:
CI integration test failure (only in spar integration tests) will be fixed with https://github.com/wireapp/wire-server/pull/1343.

PR #1345, opened 2021-01-28
Add endpoint to get client metadata for many users

akshaymankar:
Fixes https://wearezeta.atlassian.net/browse/SQCORE-415

### DIFF
--- a/services/brig/src/Brig/API/Public.hs
+++ b/services/brig/src/Brig/API/Public.hs
@@ -280,7 +280,7 @@ type ListClientsBulk =
   Summary "List all clients for a set of user ids"
     :> ZAuthServant
     :> "users"
-    :> "get-clients-using-post"
+    :> "list-clients"
     :> Servant.ReqBody '[Servant.JSON] (Range 1 MaxUsersForListClientsBulk [Qualified UserId])
     :> Post '[Servant.JSON] (Public.QualifiedUserMap (Set Public.Client))
 
--- a/services/brig/src/Brig/API/Public.hs
+++ b/services/brig/src/Brig/API/Public.hs
@@ -105,6 +105,7 @@ import qualified Wire.API.User.Client.Prekey as Public
 import qualified Wire.API.User.Handle as Public
 import qualified Wire.API.User.Password as Public
 import qualified Wire.API.User.RichInfo as Public
+import qualified Wire.API.UserMap as Public
 
 ---------------------------------------------------------------------------
 -- Sitemap
@@ -273,6 +274,14 @@ type ListUsersByIdsOrHandles =
     :> Servant.ReqBody '[Servant.JSON] Public.ListUsersQuery
     :> Post '[Servant.JSON] [Public.UserProfile]
 
+type ListClientsBulk =
+  Summary "Get clients for a list of users"
+    :> ZAuthServant
+    :> "users"
+    :> "get-clients-using-post"
+    :> Servant.ReqBody '[Servant.JSON] (Range 1 1000 [Qualified UserId])
+    :> Post '[Servant.JSON] (Public.QualifiedUserMap (Set Public.Client))
+
 type OutsideWorldAPI =
   CheckUserExistsUnqualified
     :<|> CheckUserExistsQualified
@@ -283,6 +292,7 @@ type OutsideWorldAPI =
     :<|> GetHandleInfoQualified
     :<|> ListUsersByUnqualifiedIdsOrHandles
     :<|> ListUsersByIdsOrHandles
+    :<|> ListClientsBulk
 
 type SwaggerDocsAPI = "api" :> SwaggerSchemaUI "swagger-ui" "swagger.json"
 
@@ -310,6 +320,7 @@ servantSitemap =
     :<|> getHandleInfoH
     :<|> listUsersByUnqualifiedIdsOrHandles
     :<|> listUsersByIdsOrHandles
+    :<|> listClientsBulk
 
 -- Note [ephemeral user sideeffect]
 -- If the user is ephemeral and expired, it will be removed upon calling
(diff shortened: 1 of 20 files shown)


services/brig/src/Brig/Data/Client.hs
```diff
@@ -202,6 +210,9 @@ updateClientLabelQuery = "UPDATE clients SET label = ? WHERE user = ? AND client
 selectClientIds :: PrepQuery R (Identity UserId) (Identity ClientId)
 selectClientIds = "SELECT client from clients where user = ?"
 
+selectClientsBulk :: PrepQuery R (Identity [UserId]) (UserId, ClientId, ClientType, UTCTimeMillis, Maybe Text, Maybe ClientClass, Maybe CookieLabel, Maybe Latitude, Maybe Longitude, Maybe Text)
+selectClientsBulk = "SELECT user, client, type, tstamp, label, class, cookie, lat, lon, model from clients where user in ?"
```
jschaul, line 214: Using the IN keyword here with 1000 userIds is bad for performance - the targeted cassandra node may need to contact all other nodes in the cluster to return a result.

The userId is the partition key for this table (see https://github.com/wireapp/wire-server/blob/develop/docs/reference/cassandra-schema.cql#L1112-L1124)

For more information see

See https://docs.datastax.com/en/cql-oss/3.3/cql/cql_reference/cqlSelect.html#cqlSelect__selectInNot
and 
https://lostechies.com/ryansvihla/2014/09/22/cassandra-query-patterns-not-using-the-in-query-for-multiple-partitions/

The solution to this performance problem is to mapconcurrently this query. As you probably don't want to spawn 1000 threads, you can use this very handy function from unliftIO called `pooledForConcurrentlyN_` to limit the amount of threads to say 50 or 100.

jschaul (changes requested):
I pushed a commit as this branch wasn't compiling. See inline comments (performance problem and naming) for things that should be addressed before merging.

services/brig/src/Brig/Data/Client.hs
akshaymankar: Aha, thanks! For my education, how did you come up with 50 or 100 threads?

services/brig/src/Brig/API/Public.hs
```diff
@@ -273,6 +274,14 @@ type ListUsersByIdsOrHandles =
     :> Servant.ReqBody '[Servant.JSON] Public.ListUsersQuery
     :> Post '[Servant.JSON] [Public.UserProfile]
 
+type ListClientsBulk =
+  Summary "Get clients for a list of users"
+    :> ZAuthServant
+    :> "users"
+    :> "get-clients-using-post"
```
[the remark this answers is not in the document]
akshaymankar: > Can you confirm this is exposed already at nginz configuration files?

Yeah, the nginz config in  [hegemony](https://github.com/zinfra/hegemony/blob/c3f81436ea72329f52471a4c3bcdc9eb26945466/roles/nginz/vars/main.yml#L54-L55) and [the helm charts](https://github.com/wireapp/wire-server/blob/a7226b610983cc0a7dccfdee59230fca59358b48/charts/nginz/values.yaml#L77-L82) already expose everything under `/users`

> Naming: It sounds like we will have more of these cases, where we do an HTTP POST request but only retrieve data. The other endpoint with a similar semantic was called list-users. I'd like to get a more consistent naming pattern here. How about /users/list-clients ?

I renamed it to `/users/list-clients`. There is one place in galley which looks like `/teams/:tid/get-members-using-post`, but I guess we can let that be the outlier and deprecate and replace it at some point in future.

services/brig/src/Brig/API/Public.hs
```diff
@@ -1082,6 +1093,10 @@ getClientH (zusr ::: clt ::: _) =
     Just c -> json c
     Nothing -> setStatus status404 empty
 
+listClientsBulk :: UserId -> Range 1 1000 [Qualified UserId] -> Handler (Public.QualifiedUserMap (Set Public.Client))
```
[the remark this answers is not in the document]
akshaymankar: I made this 500 and introduced a constant.

akshaymankar:
> there are also some other compilation issues:
> 
> /src/wire-server/libs/api-bot/src/Network/Wire/Bot/Crypto.hs:120:12: error:
> > • Couldn't match type ‘Opaque U’ with ‘U’
> 
> /src/wire-server/services/galley/src/Galley/API/Swagger.hs:328:10: error:
> Duplicate instance declarations:
> instance ToSchema ClientId

I just undid the refactor as it was causing more problems that I thought. Let's tackle it when we mess with those endpoints.

services/brig/src/Brig/Data/Client.hs
akshaymankar: Accouding to [Brig.App](https://github.com/wireapp/wire-server/blob/a7226b610983cc0a7dccfdee59230fca59358b48/services/brig/src/Brig/App.hs#L395-L404), we have
- 4 stripes
- 4 connections per stripe
- 128 streams per connection ([default](https://hackage.haskell.org/package/cql-io-1.1.1/docs/Database-CQL-IO.html#v:defSettings))

So, theoretically, we can have 2048 (4 * 4 * 128) requests going in parallel. If we say each request to get list of clients spawns 50 cassandra queries  in parallel at max, we're way below that limit. 
According to [the datastax java driver](https://docs.datastax.com/en/developer/java-driver/4.10/manual/core/pooling/) getting more streams doesn't improve much on performance as there is only so much a server can serve in parallel. So, this also suggests that our guess of running 50 requests in parallel is fairly conservative. 
As we don't want one type of request congesting the pool, and this request shouldn't get called with many clients very often, I think we should be fine keeping this conservative, if we see this being slow, let's consider bumping the number of requests then. WDYT?

jschaul (approved):

jschaul:
🚢

PR #1349, opened 2021-02-01
[skip ci] faster local docker image building using buildah

jschaul:
buildah-compile.sh compiles wire-server inside an alpine-based container based on quay.io/wire/alpine-builder.
the tool 'buildah' is used to mount some folders in, and to
keep the stack caches of .stack and .stack-work (renamed to avoid conflicts) for the next compilation. 

After compilation, ./buildah-make-images.sh can be used
to bake individual executables into their respective docker images used by kubernetes.

Usage: just run 'make buildah-docker' which will compile and upload
images to quay.io under the docker tag $USER by default.

On my machine, the overhead (from a successful `stack build` of wire-server to a set of ~15 docker images locally) is 1.5 minutes. Pushing to quay.io depends on your internet connection. Pushing to a local kubernetes cluster would be faster.

FUTUREWORK: build only some, not all images. (good luck figuring out the bash syntax to pass space-separated executables as arguments...)
FUTUREWORK: provide buildah using nix
FUTUREWORK: parameterize having or not `--fast` and maybe also use this on CI
FUTUREWORK: remove some docker makefile targets that are somewhat outdated

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,6 +15,7 @@ CHARTS_INTEGRATION    := wire-server databases-ephemeral fake-aws
 # this list could be generated from the folder names under ./charts/ like so:
 # CHARTS_RELEASE := $(shell find charts/ -maxdepth 1 -type d | xargs -n 1 basename | grep -v charts)
 CHARTS_RELEASE        := wire-server databases-ephemeral fake-aws aws-ingress backoffice calling-test demo-smtp elasticsearch-curator elasticsearch-external fluent-bit minio-external cassandra-external nginx-ingress-controller nginx-ingress-services reaper wire-server-metrics sftd
+BUILDAH_PUSH          ?= 1
 
 default: fast
 
@@ -344,5 +345,8 @@ echo-release-charts:
 .PHONY: buildah-docker
 buildah-docker:
 	./hack/bin/buildah-compile.sh
-	# FUTUREWORK: Allow overriding BUILDAH_PUSH
-	BUILDAH_PUSH=1 ./hack/bin/buildah-make-images.sh
+	BUILDAH_PUSH=${BUILDAH_PUSH} ./hack/bin/buildah-make-images.sh
+
+.PHONY: buildah-clean
+buildah-clean:
+	./hack/bin/buildah-clean.sh
--- a/Makefile
+++ b/Makefile
@@ -344,4 +344,5 @@ echo-release-charts:
 .PHONY: buildah-docker
 buildah-docker:
 	./hack/bin/buildah-compile.sh
+	# FUTUREWORK: Allow overriding BUILDAH_PUSH
 	BUILDAH_PUSH=1 ./hack/bin/buildah-make-images.sh
--- a/Makefile
+++ b/Makefile
@@ -340,3 +340,8 @@ upload-charts: charts-release
 .PHONY: echo-release-charts
 echo-release-charts:
 	@echo ${CHARTS_RELEASE}
+
+.PHONY: buildah-docker
+buildah-docker:
+	./hack/bin/buildah-compile.sh
+	BUILDAH_PUSH=1 ./hack/bin/buildah-make-images.sh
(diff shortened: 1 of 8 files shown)


jschaul:
Updated description. Ready for review now.

hack/bin/buildah-make-images.sh
```diff
@@ -0,0 +1,37 @@
+#!/usr/bin/env bash
+
+set -ex
+
+DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+TOP_LEVEL="$(cd "$DIR/../.." && pwd)"
+
+EXECUTABLES="cannon brig cargohold galley gundeck federator brig-index brig-schema galley-schema galley-migrate-data gundeck-schema proxy spar spar-schema"
+CONTAINER_NAME="output"
+DOCKER_TAG=$USER
```
akshaymankar, line 10: I think we should expect this variable to be set already, as have it in the Makefile. Maybe it can look something like:
```suggestion
DOCKER_TAG=${DOCKER_TAG:-$USER}
```

hack/bin/buildah-make-images.sh
```diff
@@ -0,0 +1,37 @@
+#!/usr/bin/env bash
+
+set -ex
+
+DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+TOP_LEVEL="$(cd "$DIR/../.." && pwd)"
+
+EXECUTABLES="cannon brig cargohold galley gundeck federator brig-index brig-schema galley-schema galley-migrate-data gundeck-schema proxy spar spar-schema"
```
akshaymankar, line 8:
```suggestion
# FUTUREWORK: Define this list in the makefile to allow overriding
EXECUTABLES="cannon brig cargohold galley gundeck federator brig-index brig-schema galley-schema galley-migrate-data gundeck-schema proxy spar spar-schema"
```

Makefile
```diff
@@ -340,3 +340,8 @@ upload-charts: charts-release
 .PHONY: echo-release-charts
 echo-release-charts:
 	@echo ${CHARTS_RELEASE}
+
+.PHONY: buildah-docker
+buildah-docker:
+	./hack/bin/buildah-compile.sh
+	BUILDAH_PUSH=1 ./hack/bin/buildah-make-images.sh
```
akshaymankar, line 347:
```suggestion
	# FUTUREWORK: Allow overriding BUILDAH_PUSH
	BUILDAH_PUSH=1 ./hack/bin/buildah-make-images.sh
```

hack/bin/buildah-compile.sh
```diff
@@ -0,0 +1,30 @@
...
+DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+TOP_LEVEL="$(cd "$DIR/../.." && pwd)"
+
+mkdir -p "$TOP_LEVEL"/.stack-root-buildah
+mkdir -p "$TOP_LEVEL"/.stack-work-buildah
+mkdir -p "$TOP_LEVEL"/dist-buildah
+
+CONTAINER_NAME=wire-server-dev
+
+# check for the existence of; or create a working container
```
akshaymankar, line 21:
```suggestion
# check for the existence of; or create a working container
# FUTUREWORK: Allow cleaning up old working container using some env variable or explicit make task like `make buildah-clean`
```

akshaymankar (approved):
Added some FUTUREWORK comments in code. Looks good!

arianvp:
We wanted to move away from `buildkit` to `buildah` in concourse as well.  So I guess this comes in handy. :)

I was under the impression that `buildah` could consume `Containerfile/Dockerfile`s so I'm a bit surprised we are moving away from a `Containerfile` to a shell-script. But I don't have a very strong opinion on the matter.  Is it because `Dockerfile` doesn't support mounting the cache folders?

arianvp (approved):

akshaymankar:
>  Is it because Dockerfile doesn't support mounting the cache folders?

Yep, ability to mount folders into a building container was the main reason to try out buildah.

PR #1352, opened 2021-02-08
buildah: Use correct dist directory while building docker-images [skip-ci]

akshaymankar:


jschaul (approved):
Oops! Good catch, this must have slipped through my refactoring, this was fixed at some point :)

jschaul:
(I think concourse expects `[skip ci]` or `[ci skip]` without a dash)

akshaymankar:
> (I think concourse expects `[skip ci]` or `[ci skip]` without a dash)

I consistently keep falling for this one! 🙈

PR #1353, opened 2021-02-08
buildah: Allow building only a given exec and fix brig templates [skip ci]

akshaymankar:


hack/bin/buildah-make-images.sh
```diff
@@ -6,15 +6,15 @@ DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 TOP_LEVEL="$(cd "$DIR/../.." && pwd)"
 
 # FUTUREWORK: Define this list in the makefile to allow overriding
```
akshaymankar, line 8:
```suggestion
```

jschaul (approved):

PR #1354, opened 2021-02-08
Helper

jschaul:
Can you check if this works? Do you prefer this over symlinks? (feel free to merge if you like it, or ignore otherwise)


akshaymankar (approved):
This makes us unable to use ormolu on this file, but it is small, so why not!

PR #1351, opened 2021-02-04
Galley: Add /teams/:tid/members csv download

smatting:
This PR extends the existing `/teams/:tid/members` so that it if it is requested with `Accept: text/csv` it will respond with a csv file. For details see https://wearezeta.atlassian.net/browse/SQSERVICES-107

PR #1357, opened 2021-02-11
CSV download (again)

fisx:
More left-overs from #1351 #1355:

- [ ] drop check for accept header.
- [ ] auth by cookie, not access token.
- [ ] make sure nginz has the route open and unauthenticated everywhere.
- [ ] add scim externalId or saml nameID.
- [ ] add rich info.

### DIFF
--- a/services/galley/src/Galley/API/Public.hs
+++ b/services/galley/src/Galley/API/Public.hs
@@ -177,7 +177,6 @@ sitemap = do
     -- nginz metrics dashboards easier.
     zauthUserId
       .&. capture "tid"
-      .&. accept "text" "csv"
   document "GET" "getTeamMembersCSV" $ do
     summary "Get all members of the team as a CSV file"
     parameter Path "tid" bytes' $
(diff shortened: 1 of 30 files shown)


services/galley/src/Galley/API/Public.hs
```diff
@@ -177,7 +177,6 @@ sitemap = do
     -- nginz metrics dashboards easier.
     zauthUserId
       .&. capture "tid"
-      .&. accept "text" "csv"
```
fisx, line 180: @Yserz you won't set the accept header in the csv download request, right?  or will you?

Yserz: With an anchor tag I can not modify any headers. The browser would add every mime-type to the header that is acceptable for the browser.

fisx: Ok, let's remove it then, seems safer to me.  Thanks!

smatting:
Closing this because we decided not to do auth by cookie. [context](https://wearezeta.atlassian.net/browse/SQSERVICES-107?focusedCommentId=94058)

PR #1358, opened 2021-02-11
[WIP] grpc python example request

jschaul:
(just a stub to see how working with wire-api-federation would look like when poking it directly from outside in a language other than haskell)

Maybe we could add this as an example somewhere. @akshaymankar


libs/wire-api-federation/client.py
```diff
@@ -0,0 +1,26 @@
...
+channel = grpc.insecure_channel('localhost:8098') # public-facing federator port
+stub = router_pb2_grpc.RouteToInternalStub(channel)
+
+def handle_search(handle):
+    param = QueryParam(key="handle".encode("utf-8"), value=handle.encode("utf-8"))
+    return LocalCall(
+            path="users/by-handle".encode("utf-8"),
+            query=[param],
+            method=GET,
+            component=Brig)
```
jschaul, line 18: figuring out how to write this requires knowing the code in Brig's federation api module. That makes it a bit harder. Perhaps we should consider replacing brig's federation rest API with another protobuf file at some point. Not now, but perhaps after we're done with the user handle lookup story and have a bit more insights on how grpc performs.

jschaul:
Re-opening a fresh PR with these changes rebased on develop soon.

PR #1360, opened 2021-02-15
Release 2021-02-15

smatting:
See [CHANGELOG.md](https://github.com/wireapp/wire-server/blob/release_2021_02_15/CHANGELOG.md)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -87,3 +87,8 @@ hie.yaml
 
 # generated files under .local
 .local
+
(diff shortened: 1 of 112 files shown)


jschaul (changes requested):
* nginz config path for team size and/or csv download might be incorrectly forwarding to the wrong service.
* migration path for ES is not explained in the changelog for kubernetes operators.

jschaul (approved):

PR #1361, opened 2021-02-15
Fix: Missing routes /teams/:tid/size and /teams/:tid/search

smatting:


jschaul (commented):
This change seems correct. What about `path: ~* ^/teams/([^/]*)/search$` - that has been introduced elsewhere but is missing from the helm charts, too.

smatting:
Thank you for checking! I'll make sure I add the [checklist](https://github.com/zinfra/backend-wiki/wiki/Checklists#adding-new-endpoints-in-wire-server) to any future tickets.

I've verified that both routes are correct via `make kube-integration-setup` and telepresence.

akshaymankar (approved):

PR #1367, opened 2021-02-16
Update kubernetes migration path and documentation

smatting:
This PR simplifies the the ES migration path introduced in #1360.
The migration instructions have been updated accordingly.

This PR requires #1339 to merged first.

### DIFF
--- a/charts/elasticsearch-index/templates/create-index.yaml
+++ b/charts/elasticsearch-index/templates/create-index.yaml
@@ -37,6 +37,7 @@ spec:
             - --elasticsearch-shards=5
             - --elasticsearch-replicas=2
             - --elasticsearch-refresh-interval=5
+            - --delete-template directory
       containers:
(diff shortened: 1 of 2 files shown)


charts/elasticsearch-index/templates/create-index.yaml
```diff
@@ -37,6 +37,7 @@ spec:
             - --elasticsearch-shards=5
             - --elasticsearch-replicas=2
             - --elasticsearch-refresh-interval=5
+            - --delete-template directory
```
akshaymankar, line 40: Can you please make this configurable and have it default to `directory`?

akshaymankar (changes requested):
Can you please merge this PR into #1339?
Apart from that I think we should keep the template name configurable, it will allow other template names that could be in the wild to be cleaned up.

smatting:
Closing this because it will be merged into #1339

PR #1366, opened 2021-02-15
Bump ormolu version.

fisx:
No change in formatting of the existing code, but some [nice bug fixes](https://hackage.haskell.org/package/ormolu-0.1.4.1/changelog).  (Ormolu formatting seems to stabilize.  Good.)

PR #1368, opened 2021-02-16
Commit missing stack.yaml.lock

fisx:
forgot to add this in #1366


fisx:
https://bridge-ie-concourse.zinfra.io/teams/main/pipelines/docker-images/jobs/alpine-builder/builds/167 is green, so this should go green, too.

fisx:
This is weird, alpine-builder has been updated a few hours ago, but the error that ormolu needs to be updated is still there.  Do I need to update anything else manually?

akshaymankar (approved):

PR #1372, opened 2021-02-17
Federation: Convert prekeys endpoints

smatting:
https://wearezeta.atlassian.net/browse/SQCORE-466

This PR refactors these 3 brig endpoints and adds a qualified version for each one

- `/users/prekeys`, with qualified `/users/list-prekeys`
- `/users/:uid/prekeys`, with qualified `/users/:domain/:uid/prekeys`
- `/users/:uid/prekeys/:client`, with qualified `/users/:domain/:uid/prekeys/:client`


services/brig/src/Brig/API/Client.hs
```diff
@@ -151,52 +150,46 @@ rmClient u con clt pw =
...
-    throwStd . federationNotImplemented . fmap fst
-  -- FUTUREWORK(federation, #1272): claim keys from other backends, merge maps
-  lift $ UserClientMap . Map.mapKeys makeIdOpaque <$> claimLocalPrekeyBundles localUsers
+claimPrekeyBundle :: Domain -> UserId -> AppIO PrekeyBundle
+claimPrekeyBundle domain uid = do
+  isLocalDomain <- (domain ==) <$> viewFederationDomain
+  if isLocalDomain
+    then claimLocalPrekeyBundle uid
+    else -- FUTUREWORK(federation, #1272): claim keys from other backend
+      error "TODO FUTUREWORK(federation, #1272)"
```
jschaul, line 174: Let's please avoid hard errors like these for things not yet implemented. The previous code already had a `throwStd . federationNotImplemented`, which leads (currently) to a 403. We can discuss if 403 is the best, but that can be done at the definition of federationNotImplemented.

Plain 500s without good error messages are always worrying and hard to debug for both us (we see 500s in the metrics but have no logs here of what's happening) and for client developers.

services/brig/src/Brig/API/Client.hs
```diff
@@ -151,52 +150,46 @@ rmClient u con clt pw =
...
 
-claimLocalPrekeyBundles :: [(UserId, Set ClientId)] -> AppIO (Map UserId (Map ClientId (Maybe Prekey)))
-claimLocalPrekeyBundles = foldMap getChunk . fmap Map.fromList . chunksOf 16
+claimMultiPrekeyBundles :: QualifiedUserClients -> AppIO (QualifiedUserClientMap (Maybe Prekey))
+claimMultiPrekeyBundles quc = do
+  localDomain <- viewFederationDomain
+  res <- forM (Map.toList . qualifiedUserClients $ quc) $ \(domain, userClients) -> do
+    if domain == localDomain
+      then (domain,) <$> getLocal userClients
+      else error "TODO FUTUREWORK(federation, #1272)"
```
jschaul, line 187: same here, please no `error` anywhere in production code.

services/brig/src/Brig/API/Client.hs
```diff
@@ -209,6 +202,13 @@ claimLocalPrekeyBundles = foldMap getChunk . fmap Map.fromList . chunksOf 16
       when (isNothing key) $ noPrekeys u c
       return key
 
+-- FUTUREWORK: delete this in favor of claimMultiPrekeyBundles
```
jschaul, line 205: I don't quite understand at which point this function should be deleted. What needs to change for this to become possible? Can it be done in this PR? If not, what needs to be in place for it to be done?

services/brig/src/Brig/API/Public.hs
```diff
@@ -284,6 +286,63 @@ type ListClientsBulk =
     :> Servant.ReqBody '[Servant.JSON] (Range 1 MaxUsersForListClientsBulk [Qualified UserId])
     :> Post '[Servant.JSON] (Public.QualifiedUserMap (Set Public.Client))
 
+type GetUsersPrekeysClientUnqualified =
+  Summary "Get a prekey for a specific client of a user."
```
jschaul, line 290: Add `(deprecated)` to the summary here.

services/brig/src/Brig/API/Public.hs
```diff
@@ -284,6 +286,63 @@ type ListClientsBulk =
...
+    :> ZAuthServant
+    :> "users"
+    :> Capture "domain" Domain
+    :> CaptureUserId "uid"
+    :> "prekeys"
+    :> CaptureClientId "client"
+    :> Get '[Servant.JSON] Public.ClientPrekey
+
+type GetUsersPrekeyBundleUnqualified =
+  Summary "Get a prekey for each client of a user."
```
jschaul, line 309: Add `(deprecated)` to the summary here.

services/brig/src/Brig/API/Public.hs
```diff
@@ -284,6 +286,63 @@ type ListClientsBulk =
...
+    :> Capture "domain" Domain
+    :> CaptureUserId "uid"
+    :> "prekeys"
+    :> Get '[Servant.JSON] Public.PrekeyBundle
+
+type GetMultiUserPrekeyBundleUnqualified =
+  Summary
+    "Given a map of user IDs to client IDs return a \
+    \prekey for each one. You can't request information for more users than \
+    \maximum conversation size."
```
jschaul, line 329: Add `(deprecated)` to the summary here.

services/brig/src/Brig/API/Public.hs
```diff
@@ -323,6 +388,14 @@ servantSitemap =
...
     :<|> listUsersByIdsOrHandles
     :<|> listClientsBulk
+    :<|> getPrekeyUnqualifiedH
+    :<|> getPrekeyH
+    :<|> getPrekeyBundleUnqualifiedH
+    :<|> getPrekeyBundleH
+    :<|> getMultiUserPrekeyBundleUnqualifiedH
+    :<|> getMultiUserPrekeyBundleH
+
+-- getMultiPrekeyBundlesH
```
jschaul, line 398: Can this line be deleted?

jschaul (commented):
Thanks for adding nginx routes everywhere! Can we move `/list-prekeys` to `/users//list-prekeys` though?

A few nitpicks as inline comments, generally looks good. (I have not looked at the swagger yet though)

charts/nginz/values.yaml
```diff
@@ -255,6 +255,15 @@ nginx_conf:
     - path: /calls
       envs:
       - all
+    - path: ~* ^/users/([^/]+)/([^/]+)/prekeys/([^/]+)$ 
+      envs:
+      - all
+    - path: ~* ^/users/([^/]+)/([^/]+)/prekeys$
+      envs:
+      - all
```
akshaymankar, line 263: This is not required, everything under /users is already being forwarded to brig.

libs/wire-api/src/Wire/API/User/Client/Prekey.hs
```diff
@@ -37,17 +37,30 @@ module Wire.API.User.Client.Prekey
...
 
+instance ToSchema PrekeyId where
+  declareNamedSchema _ = tweak $ declareNamedSchema (Proxy @Int)
+    where
+      tweak = fmap $ schema . description ?~ descr
+        where
+          descr = "in the range [0..65535]."
+
+-- FUTUREWORK: can this be also expressed in swagger, not just in the description?
+
```
akshaymankar, line 63: If you newtype derive ToSchema, it will have the right bounds in the schema by default.

services/brig/src/Brig/API/Client.hs
smatting: This should have been a TODO instead of FUTUREWORK. I removed it. `claimMultiPrekeyBundlesLocal` made code more DRY, so I left it

services/brig/src/Brig/API/Public.hs
```diff
@@ -381,47 +446,6 @@ sitemap o = do
...
-      "Given a map of user IDs to client IDs return a \
-      \prekey for each one. You can't request information for more users than \
-      \maximum conversation size."
-    Doc.notes
-      "Prekeys of all clients of a multiple users. \
-      \The result is a map of maps, i.e. { UserId : { ClientId : Maybe Prekey } }"
-    Doc.body (Doc.ref Public.modelUserClients) $
-      Doc.description "JSON body"
-    Doc.response 200 "Prekey Bundles" Doc.end
-    Doc.errorResponse tooManyClients
```
akshaymankar, line 400: I don't see this in the servant endpoints, am I reviewing this too soon?

smatting: It's refactored as the `GetMultiUserPrekeyBundleUnqualified` endpoint. For the qualified veresion (multiple domains) `GetMultiUserPrekeyBundleQualified` the path i mapped `POST /list-prekeys` which I don't like so much, but it's somewhat consistent with `list-users`.

Not ready yet, I'll re-request a review from you then.

services/brig/src/Brig/API/Public.hs
```diff
@@ -305,6 +307,57 @@ type ListClientsBulk =
...
+    :> "prekeys"
+    :> Get '[Servant.JSON] Public.PrekeyBundle
+
+type GetUsersPrekeyBundleQualified =
+  Summary "Get a prekey for each client of a user."
+    :> "users"
+    :> Capture "domain" Domain
+    :> CaptureUserId "uid"
+    :> "prekeys"
+    :> Get '[Servant.JSON] Public.PrekeyBundle
```
smatting, line 340: The response type doesn't contain the domain, so clients have to stitch requrest+response together, which is okay I think, wdyt? Note that for example /usr/:uid/prekeys also doesn't contain the user in the response.

smatting:
> Thanks for adding nginx routes everywhere! Can we move `/list-prekeys` to `/users/list-prekeys` though?
Done
> A few nitpicks as inline comments, generally looks good. (I have not looked at the swagger yet though)
I added manual examples in swagger

libs/wire-api/src/Wire/API/Message/Proto.hs
```diff
@@ -168,6 +168,7 @@ toOtrRecipients =
   Msg.OtrRecipients . Client.UserClientMap
     . foldl' userEntries mempty
   where
+    -- userEntries :: Map UserId (Map ClientId Text) -> UserEntry -> Map UserId (Map ClientId Text)
```
smatting, line 171: - [ ] remove

services/brig/src/Brig/API/Error.hs
```diff
@@ -524,3 +525,10 @@ federationNotImplemented qualified =
     rendered = LT.intercalate ", " . toList . fmap (LT.fromStrict . renderMapping) $ qualified
     renderMapping IdMapping {_imMappedId, _imQualifiedId} =
       idToText _imMappedId <> " -> " <> renderQualifiedId _imQualifiedId
+
+federationNotImplemented' :: Wai.Error
+federationNotImplemented' =
+  Wai.Error
+    status403
+    "federation-not-implemented"
+    "Federation is not implemented"
```
jschaul, line 534:
```suggestion
    "Federation is not implemented yet for this endpoint"
```

services/brig/src/Brig/API/Public.hs
```diff
@@ -158,6 +158,8 @@ instance
 
 type CaptureUserId name = Capture' '[Description "User Id"] name UserId
 
+type CaptureClientId name = Capture' '[Description "ClientId ID"] name ClientId
```
jschaul, line 161:
```suggestion
type CaptureClientId name = Capture' '[Description "ClientId Id"] name ClientId
```

jschaul (approved):
:shipit: looks good! Swagger also looks good.

libs/wire-api/src/Wire/API/Message/Proto.hs
jschaul: This actually looks like a usefu type declaration to add to the function for improved readability to the next person looking at this.

services/brig/src/Brig/API/Public.hs
smatting: decided not to include it

PR #1373, opened 2021-02-17
Add docs for deriving-swagger2

akshaymankar:


akshaymankar:
Looks like this:

image

libs/deriving-swagger2/src/Deriving/Swagger.hs
```diff
@@ -28,64 +30,128 @@ import GHC.Generics (Generic (Rep))
...
+-- @
+--
+-- This will produce swagger schema with fields @name@, @age@ and @birth_city@.
+--
+-- Similarly constructor tags of sum types can also be modified like this:
+--
+-- @
+-- data AssetSize = AssetComplete | AssetPreview
+--   deriving (Generic)
+--   deriving (ToSchema) via (CustomSwagger '[ConstructorTagModifier (StripPrefix \"Asset\", CamelToSnake)] LoginCodeType)
```
jschaul, line 54:
```suggestion
--   deriving (ToSchema) via (CustomSwagger '[ConstructorTagModifier (StripPrefix \"Asset\", CamelToSnake)] AssetSize)
```

libs/deriving-swagger2/src/Deriving/Swagger.hs
```diff
@@ -28,64 +30,128 @@ import GHC.Generics (Generic (Rep))
 import GHC.TypeLits (ErrorMessage (Text), KnownSymbol, Symbol, TypeError, symbolVal)
 import Imports
 
-newtype CustomSwagger t a = CustomSwagger {unCustomSwagger :: a}
-
-data StripPrefix t
+-- * How to use this library
```
jschaul, line 33: Question about haddock syntax when both `-- |` and `-- *` are in use around the module: The `-- | Type-directed ....` snippet will be shown in hoogle I presume, but will this "how to use this library" also be shown on the https://hoogle.zinfra.io/file/wire-server/.stack-work/install/x86_64-linux/146172d8ac10b98fcd2c377fe880ca327bd038f40123eca7c624740956ba8a30/8.8.4/doc/deriving-swagger2-0.1.0/Deriving-Swagger.html page? Or is it only visible when opening the source code?

libs/deriving-swagger2/src/Deriving/Swagger.hs
```diff
@@ -28,64 +30,128 @@ import GHC.Generics (Generic (Rep))
...
+  ToSchema (CustomSwagger t a)
+  where
+  declareNamedSchema _ = genericDeclareNamedSchema (swaggerOptions @t) (Proxy @a)
 
-data LabelMappings (lmap :: [LabelMapping Symbol Symbol])
+-- ** Specify __what__ to modify
 
-class StringModifier t where
-  getStringModifier :: String -> String
+-- | Modifies field labels only most likely 't' should be anything that
```
jschaul, line 109:
```suggestion
-- | Modifies field labels only. Most likely 't' should be anything that
```

libs/deriving-swagger2/src/Deriving/Swagger.hs
```diff
@@ -28,64 +30,128 @@ import GHC.Generics (Generic (Rep))
...
+-- implments 'StringModifier'
+data FieldLabelModifier t
 
-instance KnownSymbol prefix => StringModifier (StripPrefix prefix) where
-  getStringModifier = fromMaybe <*> stripPrefix (symbolVal (Proxy @prefix))
+-- | Modifies constructor tags only, most likely 't' should be anything that
+-- implments 'StringModifier'.
```
jschaul, line 114:
```suggestion
-- | Modifies constructor tags only. Most likely 't' should be anything that
-- implements 'StringModifier'.
```

jschaul (approved):
Looks good, thanks for documenting this module nicely!

libs/deriving-swagger2/src/Deriving/Swagger.hs
akshaymankar: Using `-- *` creates a section in haddocks, if it is followed by a `-- $<something>` that text appears in the section. All of this is not attached to any particular code in haskell.
`-- |` can only be used before some haskell code and attached the comment to that code. So, the `-- | Type-directed ...` stuff is attached to the module and will be shown at the top. 

The section for "how to use this library" only has `-- $usage`, that comes right after the module description. After this I have created another section for Types, to separate the free form text docs from type specific docs.

The picture in https://github.com/wireapp/wire-server/pull/1373#issuecomment-780470214 shows how it will look in zoogle.

jschaul: Thanks for the explanations!

PR #1374, opened 2021-02-17
Haskell.nix

arianvp:
You can try things out by exploring:

Theres a binary cache at `https://wire-server-test.cachix.org`

It at least has a `ghc-8.8.4` in there. So you do not have to build it fro msource. but i'm pushing more packages as we speak.

Enable it with:
```
$ cachix use wire-server-test
```

Explore what haskell.nix exposes with `nix repl`
```
nix repl ./nix
>  wireHaskellPkgs.<TAB>

> wireContainers.<TAB>
```

Build a container:

```
docker load $(nix-build ./nix -A wireContainers.brig)
```

just build a binary:
```
nix-build ./nix -A wireHaskellPkgs.brig.components.exes
```

Exercise for the reader:

Look at the haskell.nix docs and see if you can make a nix-shell file


akshaymankar:
Why did  you need to delete the dist directories?

arianvp:
`haskell.nix` tripped over them for some reason.  Because it puts build artifacts there inside the nix build (Just like Cabal puts build artifacts there) but the fact that they're not in .gitignore but in the git repo was tripping it up.

Perhaps we should rename `dist` to not clash with the same name that Cabal (is/was) using

arianvp:
Everything is pushed to cache now

arianvp:
Now with a new origin: https://github.com/wireapp/wire-server/pull/1379

PR #1377, opened 2021-02-18
sftd: Make install as part of umbrella-chart opt-in instead of opt-out

arianvp:
This makes adding of `sftd` a non-breaking change


arianvp:
cc @smatting . No need for a changelog entry as it's now a non-breaking change

smatting (approved):

PR #1382, opened 2021-02-23
charts/federator: fix allowList syntax

jschaul:
The current syntax uses 'with' which leads to end results like

    optSettings:
      setFederationStrategy:
        allowedDomains:
          - "[domain1.wire.link domain2.wire.link]"

which is not what we want. Range is the keyword to loop over a list.

Also: document federator allow list configuration

### DIFF
--- a/charts/federator/templates/configmap.yaml
+++ b/charts/federator/templates/configmap.yaml
@@ -42,9 +42,9 @@ data:
         allowAll:
         {{- else if .setFederationStrategy.allowedDomains }}
         allowedDomains:
-          {{- with $domain := .setFederationStrategy.allowedDomains }}
+        {{- range $domain := .setFederationStrategy.allowedDomains }}
           - {{ $domain | quote }}
-          {{- end }}
+        {{- end }}
         {{- else }}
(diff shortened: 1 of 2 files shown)


akshaymankar:
Have you tested this? I noticed that one of `end` statements were misplaced and the `.optSettings` should actually be `.Values.optSettings`. I have fixed it here: https://github.com/wireapp/wire-server/pull/1381/files#diff-f3a66d2bb1386b71d67d96597f4fd8db612cc8e96c8d241bdc413efc3e7db5d1. Do you want to pick that change from there and add it here, in case it is indeed broken?

akshaymankar (approved):

akshaymankar:
> Have you tested this? I noticed that one of `end` statements were misplaced and the `.optSettings` should actually be `.Values.optSettings`. I have fixed it here: https://github.com/wireapp/wire-server/pull/1381/files#diff-f3a66d2bb1386b71d67d96597f4fd8db612cc8e96c8d241bdc413efc3e7db5d1. Do you want to pick that change from there and add it here, in case it is indeed broken?

Nvm, my change is probably wrong.

PR #1359, opened 2021-02-12
Add spar.scim_external table and follow changes from spar.scim_external_ids

smatting:
Note: This PR requires a migration of spar's DB.

https://wearezeta.atlassian.net/browse/SQSERVICES-268

PR #1384, opened 2021-02-24
Add migrate-external-ids tool

smatting:
This pr adds `tools/db/migrate-external-ids`

This tool fills entries in the table `spar.scim_external` (introduced in #1359 and rreleased [2021-02-16](https://github.com/wireapp/wire-server/releases/tag/v2021-02-16)) from table `spar.scim_external_ids` and `brig.user`. It is meant to be used in the migration from `spar.scim_external_ids` to `spar.scim_external`.

### DIFF
--- a/tools/db/migrate-external-ids/src/Work.hs
+++ b/tools/db/migrate-external-ids/src/Work.hs
@@ -95,7 +95,7 @@ runCommand env@Env {..} = do
 
   failCount <- readIORef envFailCount
   when (failCount > 0) $
-    Log.warn envLogger (Log.msg @String (show failCount <> " external ids have *NOT* been migrated.\nAn external id fails to me migrated if the mapped user doesn't exist or doesn't have a team."))
+    Log.warn envLogger (Log.msg @String (show failCount <> " external ids have *NOT* been migrated.\nAn external id fails to be migrated if the mapped user doesn't exist or doesn't have a team."))
 
 type LegacyExternalId = (Text, UserId)
 
--- a/tools/db/migrate-external-ids/src/Work.hs
+++ b/tools/db/migrate-external-ids/src/Work.hs
@@ -1,0 +1,159 @@
+{-# OPTIONS_GHC -Wno-orphans -Wno-unused-imports -Wno-name-shadowing #-}
+
+-- This file is part of the Wire Server implementation.
+--
+-- Copyright (C) 2020 Wire Swiss GmbH <opensource@wire.com>
+--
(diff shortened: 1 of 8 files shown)


smatting:
I tested locally with ~20 external ids created by the spar integration tests.
Tested:
- migration. all external ids are inserted
- dry-run doesn't insert any rows
- brig user with missing team id is handled and reported at the end

tools/db/migrate-external-ids/src/Work.hs
```diff
@@ -0,0 +1,159 @@
...
+    readUserTeam Env {..} uids =
+      runClient envBrig $ do
+        query select (params Quorum (Identity uids))
+      where
+        select :: PrepQuery R (Identity [UserId]) UserTeam
+        select = "SELECT id, team FROM user where id in ?"
+
+isDebug :: Debug -> Bool
+isDebug Debug = True
+isDebug NoDebug = False
```
fisx, line 138: Move to the type definition?  (No strong opinion.)

tools/db/migrate-external-ids/src/Work.hs
```diff
@@ -0,0 +1,159 @@
...
+  runConduit $
+    zipSources
+      (CL.sourceList [(1 :: Int32) ..])
+      (readLegacyExternalIds env)
+      .| CC.mapM (resolveTeam env)
+      .| sinkMain env
+
+  failCount <- readIORef envFailCount
+  when (failCount > 0) $
+    Log.warn envLogger (Log.msg @String (show failCount <> " external ids have *NOT* been migrated.\nAn external id fails to me migrated if the mapped user doesn't exist or doesn't have a team."))
```
fisx, line 98:
```suggestion
    Log.warn envLogger (Log.msg @String (show failCount <> " external ids have *NOT* been migrated.\nAn external id fails to be migrated if the mapped user doesn't exist or doesn't have a team."))
```

fisx: `failures :: [LegacyExternalId]` containing all failed externalIds would be more helpful.  (I'm pretty sure those fit into memory, too.)

On second thought: all the information is contained in stdout/stderr, so we can just capture that.  Furthermore, this script is idempotent, so even if we forget to capture, we can just run it again.

So this looks good to me!

fisx (approved):

fisx:
@smatting good to ship!  feel free to ignore everything i write in the review, it's all nit-picks!

PR #1386, opened 2021-02-25
Forward grpc traffic to federator via ingress (or nginz for local integration tests)

jschaul:
This PR implements:

* integration tests now also deploy an ingress with a self-signed TLS certificate
* changes to allow federation traffic to reach the federator:
   * locally: nginz and its config to forward grpc traffic to federator (generated certificates locally, open an extra port on nginz, and use nginz as the "ingress" for local tests)
   * on real environments: nginx-ingress-services changes to terminate TLS & forward grpc traffic from `federator.<domain>` to federator
   * for integration tests: there is a `federation-test-helper` service which adds an SRV dns record in kubernetes for integration test discovery which redirects traffic to the ingress.
* drive-by fixes to nginz config
* example grpcurl usage bash script under ./hack/federation
* example grpc python client under ./hack/federation


deploy/services-demo/conf/nginz/nginx.conf
```diff
@@ -109,20 +109,28 @@ http {
     listen 8080;
     listen 8081;
 
+    # for nginx-without-tls, we need to use a separate port for http2 traffic,
+    # as nginx cannot handle unencrypted http1 and http2 trafic on the same
+    # port.
+    listen 8090 http2;
```
akshaymankar, line 115: I wonder if we can run brig, galley, etc on http2.

shell.nix
```diff
@@ -63,6 +63,8 @@ in pkgs.mkShell {
     pkgs.gnumake
     pkgs.haskell-language-server
     pkgs.telepresence
+    pkgs.jq
+    pkgs.grpcurl
```
akshaymankar, line 68:
```suggestion
    pkgs.cfssl

```

akshaymankar (approved):
Looks good, ship it if pipeline agrees! :rocket:

PR #1389, opened 2021-03-01
[federation] brig: refactor GET /users/*/clients endpoints

smatting:
This PR refactors 2 endpoints and creates stubs for qualified versions

- `/users/{uid}/clients` with qualified version `/users/{domain}/{uid}/clients`
- `/users/{uid}/clients/{client}` with qualified version `/users/{domain}/{uid}/clients/{client}`


akshaymankar (approved):
I suggested tiny changes. Apart from that it looks good.

PR #1391, opened 2021-03-02
Return PubClient instead of Client from /users/list-clients

akshaymankar:


services/brig/src/Brig/API/Public.hs
```diff
@@ -1117,9 +1118,9 @@ getClientH (zusr ::: clt ::: _) =
     Just c -> json c
     Nothing -> setStatus status404 empty
 
-listClientsBulk :: UserId -> Range 1 MaxUsersForListClientsBulk [Qualified UserId] -> Handler (Public.QualifiedUserMap (Set Public.Client))
-listClientsBulk _zusr limitedUids =
-  API.lookupClientsBulk (fromRange limitedUids) !>> clientError
+listClientsBulk :: UserId -> Range 1 MaxUsersForListClientsBulk [Qualified UserId] -> Handler (Public.QualifiedUserMap (Set Public.PubClient))
+listClientsBulk _zusr limitedUids = do
+  Set.map API.pubClient <$$> API.lookupClientsBulk (fromRange limitedUids) !>> clientError
```
smatting, line 1123: Maybe integrate this mapping into `API.lookupClientsBulk`? It looks like this is the only call site of `API.lookupClientsBulk`.

smatting (approved):

services/brig/src/Brig/API/Public.hs
akshaymankar: Now it makes me wonder why I fetched all that data for upto 500 users from DB in the first place! I will fix this in a follow up PR.

jschaul (approved):

PR #1393, opened 2021-03-02
Allow scim externalId to not be an email in password case

fisx:


services/spar/src/Spar/Intra/Brig.hs
```diff
@@ -78,73 +75,34 @@ import qualified Network.Wai.Utilities.Error as Wai
...
 mkUserName Nothing =
-  runValidExternalId
+  runAuthId
     (\uref -> mkName (SAML.unsafeShowNameID $ uref ^. SAML.uidSubject))
     (\email -> mkName (fromEmail email))
 
-renderValidExternalId :: ValidExternalId -> Maybe Text
-renderValidExternalId = runValidExternalId urefToExternalId (Just . fromEmail)
+renderAuthAsExternalId :: AuthId -> Maybe Text
+renderAuthAsExternalId = \case
```
fisx, line 100:
```suggestion
renderAuthIdAsExternalId = \case
```

services/spar/src/Spar/Scim/User.hs
```diff
@@ -160,6 +161,24 @@ instance Scim.UserDB ST.SparTag Spar where
       )
       $ deleteScimUser tokeninfo uid
 
+authIdFromUser :: TeamId -> User -> Maybe AuthId
+authIdFromUser stiTeam user = do
```
fisx: No, I don't think this function should exist.  But maybe we need to make another change to reading old users back from the database, if a scim-only user is created without `sso_id`?  Let's talk!

services/spar/src/Spar/Scim/User.hs
```diff
@@ -664,16 +725,18 @@ assertHandleNotUsedElsewhere uid hndl = do
...
-synthesizeStoredUser usr veid =
+synthesizeStoredUser :: UserAccount -> Scim.ScimHandler Spar (Scim.StoredUser ST.SparTag)
+synthesizeStoredUser usr =
   logScim
     ( logFunction "Spar.Scim.User.synthesizeStoredUser"
         . logUser (userId . accountUser $ usr)
         . maybe id logHandle (userHandle . accountUser $ usr)
         . maybe id logTeam (userTeam . accountUser $ usr)
-        . maybe id logEmail (veidEmail veid)
+        . maybe id logEmail (authIdEmail =<< userAuthId (accountUser usr))
```
fisx, line 734: This is still the broken authId, right?  Move `authIdFromUser` above this line?

libs/wire-api/src/Wire/API/User/Identity/AuthId.hs
```diff
@@ -114,21 +112,25 @@ data EmailSource
...
 -- (The 'TeamId' could be inferred from the saml issuer, but that would be a database lookup,
 -- which would make the function even less total.)
-urefToExternalId :: TeamId -> SAML.UserRef -> Maybe ExternalId
-urefToExternalId tid uref = ExternalId tid <$> SAML.shortShowNameID (uref ^. SAML.uidSubject)
-
--- | See 'userRefToExternalId'.  'NameID' format is either email (if parseable) or unspecified
--- (if not).
-externalIdToUref :: SAML.Issuer -> ExternalId -> SAML.UserRef
-externalIdToUref = undefined
+authIdExternalId :: TeamId -> AuthId -> Maybe ExternalId
```
smatting, line 127: Unused. Remove
- [ ] done

services/spar/src/Spar/Scim/User.hs
```diff
@@ -664,16 +725,18 @@ assertHandleNotUsedElsewhere uid hndl = do
...
     )
     $ do
+      authId <- maybe (throwError $ Scim.serverError "Server error: user doesn't haven an AuthId") pure (authIdFromUser (accountUser usr))
```
fisx, line 737: Is this a 5xx?  Can it happen if we search for a user by handle, and then find the user has no `AuthId`?  Not sure...

services/spar/src/Spar/Scim/User.hs
```diff
@@ -745,19 +808,34 @@ synthesizeScimUser :: ST.ValidScimUser -> Scim.User ST.SparTag
...
+        { Scim.externalId = Brig.renderAuthAsExternalId $ info ^. ST.vsuAuthId,
           Scim.displayName = Just $ fromName (info ^. ST.vsuName),
-          Scim.active = Just . Scim.ScimBool $ info ^. ST.vsuActive
+          Scim.active = Just . Scim.ScimBool $ info ^. ST.vsuActive,
+          Scim.emails =
+            maybeToList $
+              authIdEmailWithSource (info ^. ST.vsuAuthId) >>= \ews ->
+                case ewsEmailSource ews of
+                  EmailFromExternalIdField -> Nothing
+                  EmailFromEmailField -> emailToScimEmail (ewsEmail ews)
```
fisx, line 818: since we don't have to (and often can't) keep the data we receive unchanged, we could just add the email field here no matter where we have it from.

but this seems like the the choice that may go wrong less often.  so yes, let's keep it this way!

services/spar/src/Spar/Scim/User.hs
```diff
@@ -768,16 +846,29 @@ scimFindUserByHandle mIdpConfig stiTeam hndl = do
...
-  uid <- MaybeT . lift $ ST.runValidExternalId withUref withEmailOnly veid
+
+  let mbUref :: Maybe SAML.UserRef =
+        case sequenceA (createUserRef <$> mIdpConfig <*> pure email) of
+          Left _scimError -> Nothing
+          Right mbUref' -> mbUref'
+
+  let mbEmail :: Maybe Email = parseEmail email
+
+  uid <- case (mbUref, mbEmail) of
```
fisx, line 854:
```suggestion
  uid <- case (mbUref, parseEmail email) of
```

services/spar/src/Spar/Scim/User.hs
smatting: Can't, because the authIdFromUser is wrapped by it. For logging it's fine to get the email from the authId from user I think

services/spar/src/Spar/Scim/User.hs
smatting: In general authIdFromUser can only fail if 1) the user has no identity OR 2) the user has a `PhoneIdentity`

services/spar/src/Spar/Scim/User.hs
```diff
@@ -111,7 +131,8 @@ instance Scim.UserDB ST.SparTag Spar where
             | Scim.isUserSchema schema -> do
               x <- runMaybeT $ case attrName of
                 "username" -> scimFindUserByHandle mIdpConfig stiTeam val
-                "externalid" -> scimFindUserByEmail mIdpConfig stiTeam val
+                "externalid" -> scimFindUserByExternalId mIdpConfig stiTeam val
+                -- TODO: what if we want to search by email, but the email is stored in `emails`, not `externalId`?
```
smatting, line 135: When the externalId is not a valid email `scimFindUserByExternalId` fails to find the user by email. It succeeds however when searched by the external id. Do we need to fix this in this PR?

services/spar/src/Spar/Scim/User.hs
```diff
@@ -260,42 +295,63 @@ validateScimUser' midp richInfoLimit user = do
...
+      throwError err
+
+  pure $ AuthSCIM (ScimDetails (ExternalId tid extid) emailWithSource)
+mkAuth tid (Just idp) (Just extid) _ = do
+  uref <- createUserRef idp extid
   pure $ case parseEmail extid of
-    Just email -> ST.EmailAndUref email uref
-    Nothing -> ST.UrefOnly uref
+    Just email -> AuthBoth tid uref (Just (EmailWithSource email EmailFromExternalIdField))
+    Nothing -> AuthSAML uref -- TODO: also handle external email here!
```
smatting, line 336: What is TODO here?

services/spar/src/Spar/Scim/User.hs
fisx: God no, please don't!  i was merely marking this as TODO to remind me to think about it again later.

services/spar/src/Spar/Scim/User.hs
fisx: 24e092f

fisx: (Not sure if that helped?  Please ask again if not.)

services/spar/src/Spar/Scim/User.hs
fisx: I mean like this: 
e038c92

libs/brig-types/src/Brig/Types/Intra.hs
```diff
@@ -146,28 +147,25 @@ instance ToJSON UserAccount where
 -- NewUserScimInvitation
 
 data NewUserScimInvitation = NewUserScimInvitation
-  { newUserScimInvTeamId :: TeamId,
-    newUserScimInvLocale :: Maybe Locale,
+  { newUserScimDetails :: ScimDetails,
```
fisx, line 150: This won't work: it requires to deploy brig and spar simultaneously, which is impossible.  We need to keep `NewUserScimInvitation` backwards compatible, or introduce a new end-point that has the new behavior, keeping the old end-point alive.

fisx: Would it be an option to break this PR into two?  One for refactoring, maintaining behavior perfectly (or mostly even) towards the outside, and another one that changes behavior?  I think that would be worth a lot, and we already decided we want to keep the two parts in develop at least as commits.

services/spar/src/Spar/App.hs
```diff
@@ -84,12 +85,20 @@ import Spar.Error
...
-import Spar.Scim.Types (ValidExternalId (..), runValidExternalId)
 import Spar.Types
 import qualified System.Logger as Log
 import System.Logger.Class (MonadLogger (log))
 import URI.ByteString as URI
 import Web.Cookie (SetCookie, renderSetCookie)
+import Wire.API.User (ScimDetails (ScimDetails))
+import Wire.API.User.Identity (AuthId (..))
+
+-- replace runAuthId in wire-api with this
```
fisx, line 96:
```suggestion
-- TODO: replace runAuthId in wire-api with this
```

services/spar/src/Spar/Scim/User.hs
```diff
@@ -304,6 +383,21 @@ mkValidExternalId (Just idp) (Just extid) = do
...
+primaryEmail emails = do
+  for (firstScimEmail emails) $ \scimEmail ->
+    maybe
+      (throwError $ Scim.badRequest Scim.InvalidValue (Just "Not a valid email adress (conflict)"))
+      pure
+      (fromEmailAddress . ScimEmail.unEmailAddress . ScimEmail.value $ scimEmail)
+  where
+    firstScimEmail :: [ScimEmail.Email] -> Maybe ScimEmail.Email
+    firstScimEmail [] = Nothing
+    firstScimEmail (e : es) = Just $ maximumOn1 ScimEmail.primary (e :| es)
```
fisx, line 399: This line makes a lot of implicit assumptions and is a bit hard to read.  What about this?

```suggestion
    firstScimEmail :: [ScimEmail.Email] -> Maybe ScimEmail.Email
    firstScimEmail es = listToMaybe $ (filter (unScimBool . ScimEmail.primary) es) <> es
```

(This is why doctest is so awesome.  You can write completely obfuscated one-liners, as long as you can make their behavior intuitively clear with a handful examples.)

fisx: Also, please think of better names for `primaryEmail`, `primaryEmail'`: they are not necessarily primary ("first" was better), and the `'` puns nicely on the `primary`, but otherwise is not very helpful.

What about `firstEmailFromUser`, `firstEmailFromEmailsField`?

services/spar/src/Spar/Scim/User.hs
```diff
@@ -370,25 +462,27 @@ createValidScimUser tokeninfo@ScimTokenInfo {stiTeam} vsu@(ST.ValidScimUser veid
     $ do
       -- ensure uniqueness constraints of all affected identifiers.
       -- {if we crash now, retry POST will just work}
-      assertExternalIdUnused veid
+      assertExternalIdUnused authId
       assertHandleUnused handl
       -- {if we crash now, retry POST will just work, or user gets told the handle
       -- is already in use and stops POSTing}
 
+      liftIO $ putStrLn $ "authId " <> show authId
```
fisx, line 470:
```suggestion
      liftIO $ putStrLn $ "authId " <> show authId  -- TODO: don't merge this!
```

services/spar/src/Spar/Scim/User.hs
```diff
@@ -410,23 +504,25 @@ createValidScimUser tokeninfo@ScimTokenInfo {stiTeam} vsu@(ST.ValidScimUser veid
       -- @AccountStatus == PendingActivation@, which translates to @active == False@.  So we need
       -- to reload the Account from brig.
       storedUser <- do
-        acc <-
-          lift (Brig.getBrigUserAccount Brig.WithPendingInvitations buid)
-            >>= maybe (throwError $ Scim.serverError "Server error: user vanished") pure
-        synthesizeStoredUser acc veid
+        acc <- do
+          mbAccount <- lift (Brig.getBrigUserAccount Brig.WithPendingInvitations buid)
+          liftIO $ putStrLn $ "mbAccount" <> show mbAccount
```
fisx, line 509:
```suggestion
          liftIO $ putStrLn $ "mbAccount" <> show mbAccount  -- TODO: don't merge this
```

services/spar/src/Spar/Scim/User.hs
```diff
@@ -495,20 +591,20 @@ updateValidScimUser tokinfo uid newValidScimUser =
...
   TeamId ->
   UserId ->
-  ST.ValidExternalId ->
-  ST.ValidExternalId ->
+  AuthId ->
+  AuthId ->
   Spar ()
-updateVsuUref team uid old new = do
-  let geturef = ST.runValidExternalId Just (const Nothing)
+updateVsuUref _team uid old new = do
```
fisx, line 597:
```suggestion
updateVsuUref _team uid old new = do  -- TODO: remove `_team` from parameter list, it's unused.
```

libs/wire-api/src/Wire/API/User/Identity/AuthId.hs
fisx: It is used again, but it shouldn't be, I think.

It's turning an `AuthSAML` into an `ExternalId`, which (I think) is only needed for cases where a SAML-only user is getting in contact with SCIM for the first time.  After that, we should always end up in the `AuthBoth` case.  So when connecting a user to SCIM, instead of this function, we should do something that makes it more explicit that we're creating a fresh `ExternalId`.

Also, the `AuthBoth` case is not using the `tid` it already carries,  but one it is being passed as an argument.  This is at best a bad smell.

services/spar/src/Spar/Scim/User.hs
```diff
@@ -741,58 +846,98 @@ synthesizeStoredUser' uid veid dname handle richInfo accStatus createdAt lastUpd
...
 
--- | Construct a 'ValidExternalid'.  If it an 'Email', find the non-SAML SCIM user in spar; if
--- that fails, find the user by email in brig.  If it is a 'UserRef', find the SAML user.
+  -- TODO: remove authIdFromUser but keep same behaviour
+  case authIdFromUser (accountUser brigUser) of
+    Just _authId -> lift $ do
+      synthesizeStoredUser stiTeam brigUser
+    Nothing -> Applicative.empty
+
+-- | Find user by externalId, which is tried as a name id of a SAML User, an externalId spar,
```
fisx, line 887:
```suggestion
-- | Find user by externalId, which is tried as a `SAML.NameID` of a SAML user, an externalId spar,
```

services/spar/src/Spar/Scim/User.hs
```diff
@@ -741,58 +846,98 @@ synthesizeStoredUser' uid veid dname handle richInfo accStatus createdAt lastUpd
...
+-- or an email in brig.
 -- Return the result as a SCIM user.
 --
 -- Note the user won't get an entry in `spar.user`.  That will only happen on their first
 -- successful authentication with their SAML credentials.
-scimFindUserByEmail :: Maybe IdP -> TeamId -> Text -> MaybeT (Scim.ScimHandler Spar) (Scim.StoredUser ST.SparTag)
-scimFindUserByEmail mIdpConfig stiTeam email = do
+scimFindUserByExternalId :: Maybe IdP -> TeamId -> Text -> MaybeT (Scim.ScimHandler Spar) (Scim.StoredUser ST.SparTag)
+scimFindUserByExternalId mIdpConfig stiTeam extId = do
```
fisx, line 894: (This is where I'm giving up for today, can't focus any more.  I have read all of `git diff 0d4e837~..e038c923` except this function and changes under `spar/test-integration`.)

libs/wire-api/src/Wire/API/User/Identity/AuthId.hs
smatting: I inlined the funciton and removed it from wire-api

services/spar/src/Spar/Scim/User.hs
smatting: K, removed the TODO

services/spar/src/Spar/Scim/User.hs
smatting: Thanks, that helped. Interestingly the impelmentation: `AuthBoth` is guaranteed to have a `EmailWithSource` value, even though we desinged the constructor `AuthBoth TeamId SAML.UserRef (Maybe EmailWithSource)`. I think this makes sense, since having an uref and email is exactly the defintion of `AuthBoth`.

services/spar/src/Spar/Scim/User.hs
smatting: Haha, pun not intendend (unfortunately..). I applied your suggestions

libs/brig-types/src/Brig/Types/Intra.hs
smatting: Ah, good catch. I made the NewUserScimInvitation backwards-compatible. With this we can deploy brig first, then spar.

I'm in favor of spearating the PR if not for readability alone. I would however deploy them together.

libs/wire-api/src/Wire/API/User/Activation.hs
```diff
@@ -168,15 +169,11 @@ modelActivationResponse = Doc.defineModel "ActivationResponse" $ do
   Doc.property "first" Doc.bool' $
     Doc.description "Whether this is the first successful activation (i.e. account activation)."
 
--- FUTUREWORK: de-deduplicate work with JSON instance for 'UserIdentity'?
```
fisx, line 171: (This is a good example for something that we should have moved to a separate branch and merged long ago, to make for smaller PRs.)

services/brig/src/Brig/API/Internal.hs
```diff
@@ -474,24 +485,41 @@ addPhonePrefixH (_ ::: req) = do
...
+    Just tid -> updateAuthId uid (fromLegacyAuthId legacyAuthId tid)
+
+deleteSSOIdLegacyH :: UserId ::: JSON -> Handler Response
+deleteSSOIdLegacyH (uid ::: _) = deleteAuthId uid
+
+updateAuthIdH :: UserId ::: JSON ::: JsonRequest AuthId -> Handler Response
+updateAuthIdH (uid ::: _ ::: req) = do
+  updateAuthId uid =<< parseJsonBody req
+
+updateAuthId :: UserId -> AuthId -> Handler Response
```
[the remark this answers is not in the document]
fisx: 76542821bcafe643f3f361c93622d8c320cee076

services/brig/src/Brig/API/User.hs
```diff
@@ -201,7 +202,9 @@ createUser new@NewUser {..} = do
             { newUserManagedBy = case mbExistingAccount of
                 Nothing -> newUserManagedBy
                 Just acc -> Just . userManagedBy . accountUser $ acc,
-              newUserIdentity = ident
+              newUserIdentity = case mbExistingAccount of
```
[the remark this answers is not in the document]
smatting: Yes, if I consult my reflog, the commit that introduced the change had message.  

> Fix:createUser ignores user identity of invitation

I don't remember the details.
I reverted this change (https://github.com/wireapp/wire-server/pull/1393/commits/0e152da188b2ed18306e952538f8b8b59f9fa51f) because all tests still pass without it.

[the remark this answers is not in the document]
fisx: this may be due to my changing the definition of `ident` a few lines above?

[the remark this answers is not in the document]
smatting: No, weirdly enough not. I reverted the `ident` definition and the tests still passed. It must've been some other changes.

services/spar/test-integration/Util/Scim.hs
```diff
@@ -117,26 +119,30 @@ randomScimUserWithSubjectAndRichInfo ::
   m (Scim.User.User SparTag, SAML.UnqualifiedNameID)
 randomScimUserWithSubjectAndRichInfo richInfo = do
```
fisx, line 120: there is a duplicate of this in brig that may also need updating?

also, could you add a FUTUREWORK about using `eml` in both the `0` and the `1` case?

services/spar/test-integration/Util/Scim.hs
```diff
@@ -623,5 +620,21 @@ userShouldMatch u1 u2 = liftIO $ do
...
 -- what we expect a user that comes back from spar to look like in terms of what it looked
 -- like when we sent it there.
-whatSparReturnsFor :: HasCallStack => IdP -> Int -> Scim.User.User SparTag -> Either String (Scim.User.User SparTag)
-whatSparReturnsFor idp richInfoSizeLimit = either (Left . show) (Right . synthesizeScimUser) . validateScimUser' (Just idp) richInfoSizeLimit
+whatSparReturnsFor :: HasCallStack => TeamId -> IdP -> Int -> Scim.User.User SparTag -> Either String (Scim.User.User SparTag)
+whatSparReturnsFor tid idp richInfoSizeLimit =
+  either (Left . show) (Right . synthesizeScimUser)
+    . validateScimUser' tid (Just idp) richInfoSizeLimit
+
+renderAuthIdAsExternalId :: AuthId -> Maybe Text
```
fisx, line 628: move this to wire-api?  (you can refuse, or do it in the next PR.)

libs/wire-api/test/unit/Test/Wire/API/User/Identity/AuthId.hs
```diff
@@ -0,0 +1,194 @@
...
+  describe "AuthId markers" $ do
+    it "validStates, validStatesExpected are in sync" $ do
+      validStates `shouldBe` (fst `Set.map` validStatesExpected)
+    it "all validStateExpected entries have examples" $ do
+      forM_ (Set.toList validStatesExpected) $ \(_, examples) -> do
+        examples `shouldNotBe` mempty
+    prop "arbitrary AuthId values are not missing in validStatesExpected" $ \(authId :: AuthId) -> do
+      allSamples `shouldContain` [Equiv authId]
+
+  describe "what's an email" $ do
```
fisx, line 27:
```suggestion
  xdescribe "what's an email" $ do
```

...  and we can fix this in a self-contained PR.

services/spar/test-integration/Test/Spar/Scim/UserSpec.hs
```diff
@@ -227,12 +232,48 @@ testCreateUserWithPass = do
     -- TODO: yes, we should just test for error labels consistently, i know...
     const (Just "Setting user passwords is not supported for security reasons.") =~= responseBody
 
-testCreateUserNoIdP :: TestSpar ()
-testCreateUserNoIdP = do
+data IdentityTestConfig
+  = EmailExternalId
+  | RandomExternalIdAndEmail
```
fisx, line 237:
```suggestion
  | RandomExternalIdAndEmail  -- the `unspecified` `NameID` case.
```

services/brig/src/Brig/User/Search/Index.hs
```diff
@@ -50,7 +50,7 @@ module Brig.User.Search.Index
 where
 
 import Brig.Data.Instances ()
-import Brig.Index.Types (CreateIndexSettings (..))
+import Brig.Index.Types (CreateIndexSettings (..), UncheckedLegacyAuthId (..))
```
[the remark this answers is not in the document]
smatting: I yhink can't do proper parse error handling, because the parsing happens in a `Cql` instance in DB lookup. This happens in a batch operation, so I'm not sure if it's possible to skip rows here.

services/brig/test/integration/API/Team.hs
```diff
@@ -757,25 +760,25 @@ testConnectionSameTeam brig = do
...
+  authId <- liftIO $ AuthSAML <$> generate arbitrary
   -- creating users requires both sso_id and team_id
-  postUser' True False "dummy" True False (Just ssoid) Nothing brig
-    !!! const 400 === statusCode
-  postUser' True False "dummy" True False Nothing (Just teamid) brig
-    !!! const 400 === statusCode
+  -- postUser' True False "dummy" True False (Just ssoid) Nothing brig
+  --   !!! const 400 === statusCode
+  -- postUser' True False "dummy" True False Nothing (Just teamid) brig
+  --   !!! const 400 === statusCode
```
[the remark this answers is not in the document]
smatting: This should have been deleted,

| Use `postUser' True False\` instead of 'postUser' if you want to send broken bodies to test error

I decided to refactor `postUser'` to send a proper `NewUser` type instead of a `Aeson.Value`, because I think there is not need to test this in a integration test.

This made these tests obsolete. I'll remove them.

services/spar/test-integration/Test/Spar/Scim/UserSpec.hs
```diff
@@ -227,12 +232,48 @@ testCreateUserWithPass = do
...
+  | EmailExternalIdAndEmail
+
+createTestIdentity :: IdentityTestConfig -> TestSpar (Scim.User.User SparTag, Text, Email)
```
fisx, line 240:
```suggestion
-- | FUTUREWORK(fisx): this partially duplicates 'randomScimUser', 'randomScimUserWithSubjectAndRichInfo', consolidate that!
createTestIdentity :: IdentityTestConfig -> TestSpar (Scim.User.User SparTag, Text, Email)
```

services/spar/test-integration/Test/Spar/Scim/UserSpec.hs
```diff
@@ -227,12 +232,48 @@ testCreateUserWithPass = do
     -- TODO: yes, we should just test for error labels consistently, i know...
     const (Just "Setting user passwords is not supported for security reasons.") =~= responseBody
 
-testCreateUserNoIdP :: TestSpar ()
-testCreateUserNoIdP = do
+data IdentityTestConfig
+  = EmailExternalId
+  | RandomExternalIdAndEmail
+  | EmailExternalIdAndEmail
```
fisx, line 238:
```suggestion
  | EmailExternalIdAndEmail -- (two different emails, which is a bit odd, but we want to make sure doing that is no way for a scim peer to break wire.)
```

services/spar/test-integration/Util/Scim.hs
smatting: I thought about this, but since this function is only used in the integration tests I didn't want to elevate its officialness.

services/spar/test-integration/Test/Spar/Scim/UserSpec.hs
```diff
@@ -1081,6 +1109,9 @@ testUpdateRequiresUserId = do
   updateUser_ (Just tok) Nothing user (env ^. teSpar)
     !!! assertTrue_ (inRange (400, 499) . statusCode)
 
+testScimEmailUpdate :: EmailSource -> TestSpar ()
+testScimEmailUpdate _ = pure ()
```
fisx, line 1113:
```suggestion
testScimEmailUpdate _ = liftIO pending
```

(sorry, my bad)

services/spar/test-integration/Test/Spar/Scim/UserSpec.hs
```diff
@@ -1213,27 +1244,27 @@ testUpdateExternalId withidp = do
...
             then do
               (hasChanged, muserid) `shouldBe` (hasChanged, Nothing)
               (hasChanged, muserid') `shouldBe` (hasChanged, Just userid)
             else do
-              (hasChanged, veid') `shouldBe` (hasChanged, veid)
+              (hasChanged, authId') `shouldBe` (hasChanged, authId)
               (hasChanged, muserid') `shouldBe` (hasChanged, Just userid)
 
-      lookupByValidExternalId :: ValidExternalId -> TestSpar (Maybe UserId)
+      lookupByValidExternalId :: AuthId -> TestSpar (Maybe UserId)
```
fisx, line 1262:
```suggestion
      -- FUTUREWORK: rename to lookupByAuthId
      lookupByValidExternalId :: AuthId -> TestSpar (Maybe UserId)
```

services/spar/test-integration/Util/Scim.hs
smatting: Done in ce66

services/spar/test-integration/Test/Spar/Scim/UserSpec.hs
smatting: Removed this test, because it will be implemented as part of https://wearezeta.atlassian.net/browse/SQSERVICES-376

libs/wire-api/test/unit/Test/Wire/API/User/Identity/AuthId.hs
```diff
@@ -24,40 +24,6 @@ specs = do
     prop "arbitrary AuthId values are not missing in validStatesExpected" $ \(authId :: AuthId) -> do
       allSamples `shouldContain` [Equiv authId]
 
-  xdescribe "what's an email" $ do
```
smatting: I removed this only because i'd like to run all other tests on CI first.
- [ ] rever this

services/spar/src/Spar/Scim/User.hs
```diff
@@ -410,23 +495,24 @@ createValidScimUser tokeninfo@ScimTokenInfo {stiTeam} vsu@(ST.ValidScimUser veid
...
+        runAuthId
           (`Data.insertSAMLUser` buid)
-          (\email -> Data.insertScimExternalId stiTeam email buid)
-          veid
+          (\(ScimDetails externalId _ews) -> Data.insertScimExternalId externalId buid)
+          authId
 
       -- If applicable, trigger email validation procedure on brig.
-      lift $ ST.runValidExternalId (validateEmailIfExists buid) (\_ -> pure ()) veid
+      lift $ runAuthId (validateEmailIfExists buid) (\_ -> pure ()) authId
```
fisx, line 515: Shouldn't this be triggered if there is no IdP as well?  (*especially* then?)

services/spar/src/Spar/Scim/User.hs
fisx: Why did you resolve this?  It was still in there, until now: 01f4473ad59b79994a1bfe1d0153b06d13cd21f1

services/spar/src/Spar/Scim/User.hs
```diff
@@ -495,20 +581,20 @@ updateValidScimUser tokinfo@ScimTokenInfo {stiTeam} uid newValidScimUser =
...
   UserId ->
-  ST.ValidExternalId ->
-  ST.ValidExternalId ->
+  AuthId ->
+  AuthId ->
   Spar ()
-updateVsuUref team uid old new = do
-  let geturef = ST.runValidExternalId Just (const Nothing)
+updateVsuUref _team uid old new = do
+  let geturef = runAuthId Just (const Nothing)
```
fisx, line 588: shouldn't this extract an email?

services/spar/src/Spar/Scim/User.hs
```diff
@@ -495,20 +581,20 @@ updateValidScimUser tokinfo@ScimTokenInfo {stiTeam} uid newValidScimUser =
...
   case (geturef old, geturef new) of
     (mo, mn@(Just newuref)) | mo /= mn -> validateEmailIfExists uid newuref
     _ -> pure ()
 
   wrapMonadClient $ do
-    old & ST.runValidExternalId Data.deleteSAMLUser (Data.deleteScimExternalId team)
-    new & ST.runValidExternalId (`Data.insertSAMLUser` uid) (\email -> Data.insertScimExternalId team email uid)
+    old & runAuthId Data.deleteSAMLUser (\(ScimDetails externalId _) -> Data.deleteScimExternalId externalId)
+    new & runAuthId (`Data.insertSAMLUser` uid) (\(ScimDetails externalId _) -> Data.insertScimExternalId externalId uid)
```
fisx, line 595: shouldn't this only happen if the id actually changed?

services/spar/src/Spar/Scim/User.hs
```diff
@@ -664,16 +757,21 @@ assertHandleNotUsedElsewhere uid hndl = do
...
   logScim
     ( logFunction "Spar.Scim.User.synthesizeStoredUser"
         . logUser (userId . accountUser $ usr)
         . maybe id logHandle (userHandle . accountUser $ usr)
         . maybe id logTeam (userTeam . accountUser $ usr)
-        . maybe id logEmail (veidEmail veid)
+        . maybe id logNewAuthId (newUserAuthId (accountUser usr))
     )
     $ do
+      authId <- maybe (throwError $ Scim.serverError "Server error: user doesn't haven an AuthId") pure (newUserAuthId (accountUser usr))
```
fisx, line 770:
```suggestion
      authId <- maybe (throwError $ Scim.serverError "Server error: user doesn't have an AuthId") pure (newUserAuthId (accountUser usr))
```

services/spar/src/Spar/Scim/User.hs
```diff
@@ -745,54 +843,95 @@ synthesizeScimUser :: ST.ValidScimUser -> Scim.User ST.SparTag
...
+  where
+    getOrCreateExternalId :: AuthId -> Maybe Text
+    getOrCreateExternalId = \case
+      AuthSAML uref -> SAML.shortShowNameID (uref ^. SAML.uidSubject)
+      AuthSCIM (ScimDetails (ExternalId _ t) _) -> pure t
+      AuthBoth _ uref _ -> SAML.shortShowNameID (uref ^. SAML.uidSubject)
+
+    emailToScimEmail :: Email -> Maybe ScimEmail.Email
+    emailToScimEmail =
+      fmap scimEmail . toEmailAddress
```
fisx, line 865: I think both the two copies of `unsafeEmailToScimEmail` in the tests and this `emailToScimEmail`, and `toEmailAddress` and `fromEmailAddress` can be made total: the two email types come from different packages (wire-api and email-validate), but they are both validated.

Anyway, that's a FUTUREWORK, and probably related to the question of whether `@handle` is also a valid email address (which it shouldn't be).

services/spar/src/Spar/Scim/User.hs
smatting: I'm sorry, I don't know. I resolve the suggestions after I create a local commit. <aybe I dropped the change during squashing/rebasing. I'll try a new a approach to prevent this from happening: Resolve only after looking at diff of the _pushed _ commits.

services/spar/src/Spar/Scim/User.hs
fisx: 289cfa1

fisx (commented):
Approved!  Let's merge this and never speak of it again!  :-P

(I opened the PR, @smatting could you klick the button for me?)

services/spar/src/Spar/Scim/User.hs
fisx: i'm also fine with not resolving comments, and just merging/rebasing/squashing/whicheverwayresolving the PR eventually.  :)

services/brig/src/Brig/User/Event.hs
```diff
@@ -74,8 +75,8 @@ data UserUpdatedData = UserUpdatedData
     eupHandle :: !(Maybe Handle),
     eupLocale :: !(Maybe Locale),
     eupManagedBy :: !(Maybe ManagedBy),
-    eupSSOId :: !(Maybe UserSSOId),
-    eupSSOIdRemoved :: Bool
+    eupAuthId :: !(Maybe AuthId),
```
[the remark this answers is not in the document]
smatting: Made backwards compatible in `9251`

libs/wire-api/test/unit/Test/Wire/API/User/Identity/AuthId.hs
smatting: Reverted in separate PR #1437

services/spar/src/Spar/Scim/User.hs
smatting: https://wearezeta.atlassian.net/browse/SQSERVICES-379

libs/wire-api/src/Wire/API/User/Identity.hs
```diff
@@ -219,19 +217,13 @@ isValidPhone = either (const False) (const True) . parseOnly e164
...
-{-# DEPRECATED UserSSOId "use AuthId instead: https://wearezeta.atlassian.net/browse/SQSERVICES-264" #-}
-
-data UserSSOId
-  = UserSSOId
-      -- An XML blob pointing to the identity provider that can confirm
-      -- user's identity.
-      Text
-      -- An XML blob specifying the user's ID on the identity provider's side.
-      Text
-  | UserScimExternalId
```
fisx, line 281: https://wearezeta.atlassian.net/browse/SQSERVICES-306

fisx (commented):
Feel free to remove the last commit if you don't want the noise, but I couldn't resist.  :)

Now good to go I think!

smatting (approved):

services/spar/src/Spar/Intra/Brig.hs
```diff
@@ -187,20 +135,20 @@ class (Log.MonadLogger m, MonadError SparError m) => MonadSparToBrig m where
...
+  AuthId ->
   UserId ->
   TeamId ->
   -- | User name
   Name ->
   -- | Who should have control over the user
   ManagedBy ->
   m UserId
-createBrigUserSAML uref (Id buid) teamid uname managedBy = do
+createBrigUserSAML authId (Id buid) teamid uname managedBy = do
```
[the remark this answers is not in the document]
smatting: I refactored this so that all public functions are well-specified https://github.com/wireapp/wire-server/pull/1393/commits/c685a5c823971e3d7b91375997cbd277a1322dca

PR #1397, opened 2021-03-03
Return UserProfile when getting user by qualified handle

akshaymankar:


services/brig/src/Brig/API/Public.hs
```diff
@@ -1444,10 +1444,10 @@ getHandleInfo self handle = do
         Just ownerId -> do
           ownerProfile <- lift $ API.lookupProfile self (Qualified ownerId domain)
           owner <- filterHandleResults self (maybeToList ownerProfile)
-          return $ Public.UserHandleInfo . Public.profileQualifiedId <$> listToMaybe owner
+          return $ listToMaybe owner
     getRemoteHandleInfo = do
-      Log.info $ (Log.msg $ Log.val "getHandleInfo - remote lookup") Log.~~ Log.field "domain" (show (qDomain handle))
-      Federation.getUserHandleInfo handle
+      Log.info $ Log.msg (Log.val "getHandleInfo - remote lookup") Log.~~ Log.field "domain" (show (qDomain handle))
+      undefined $ Federation.getUserHandleInfo handle
```
jschaul, line 1450: I see an undefined in a PR not marked as draft.

jschaul (changes requested):
This is a breaking API change:

UserhandleInfo has `user` whereas userprofile has `id`:

```json
{
   "user" : "ec405119-ab35-4648-92d8-96e9d2747770",
   "qualified_user" : {
      "domain" : "staging.zinfra.io",
      "id" : "ec405119-ab35-4648-92d8-96e9d2747770"
   }
}
```

```json
{
   "handle" : "test",
   "id" : "ec405119-ab35-4648-92d8-96e9d2747770",
    "qualified_id" : {
      "domain" : "staging.zinfra.io",
      "id" : "ec405119-ab35-4648-92d8-96e9d2747770"
   },
   "name" : "test",
   "accent_id" : 0,
   "assets" : [
      {
         "key" : "3-2-774cfe48-517b-4bd1-96af-9f673557eacc",
         "size" : "preview",
         "type" : "image"
      },
      {
         "key" : "3-2-119cbc3f-b389-4194-ab19-eef63afede89",
         "size" : "complete",
         "type" : "image"
      }
   ],
   "picture" : [
      {
         "content_length" : 156995,
         "content_type" : "image/jpg",
         "data" : null,
         "id" : "b37a6b41-4c48-4a93-80dd-e3d14ce75287",
         "info" : {
            "correlation_id" : "ad12d79f-18e5-4f5f-95dd-eed7f372cb22",
            "height" : 280,
            "nonce" : "ad12d79f-18e5-4f5f-95dd-eed7f372cb22",
            "original_height" : 1200,
            "original_width" : 1200,
            "public" : true,
            "tag" : "smallProfile",
            "width" : 280
         }
      },
      {
         "content_length" : 503740,
         "content_type" : "image/jpg",
         "data" : null,
         "id" : "b5c76c68-a212-4e54-94e9-38ac7670dda4",
         "info" : {
            "correlation_id" : "ad12d79f-18e5-4f5f-95dd-eed7f372cb22",
            "height" : 1200,
            "nonce" : "ad12d79f-18e5-4f5f-95dd-eed7f372cb22",
            "original_height" : 1200,
            "original_width" : 1200,
            "public" : true,
            "tag" : "medium",
            "width" : 1200
         }
      }
   ]
}
```

I don't like this much in the current state. I'm happy with a new API endpoint (can this be exposed under another route?), but breaking the API in a subtle way by renaming one of the required and expected fields isn't very nice, even if it will be deprecated and eventually phased out. It's also inconsistent with respect to the old route. Adding both `user` and `id` fields to the json might be one way around this.

services/brig/test/integration/API/Federation.hs
```diff
@@ -41,9 +42,17 @@ testGetUserByHandleSuccess brig = do
   hdl <- randomHandle
   putHandle brig uid hdl
     !!! const 200 === statusCode
-  get (brig . paths ["federation", "users", "by-handle"] . queryItem "handle" (toByteString' hdl)) !!! do
-    const 200 === statusCode
-    const (Just (UserHandleInfo quid)) === (responseJsonMaybe)
+  profile <-
```
jschaul, line 45: It's nice to change the test when changing the API; however, the previous expectation should still hold for backwards compatibility, and parsing the response as UserHandleInfo should still work here. Could you test that the response parses as both the old and the new type if you'd like to change that endpoint?

services/brig/src/Brig/API/Public.hs
akshaymankar: Hmm, some test is missing!

akshaymankar:
> This is a breaking API change:
> 
> UserhandleInfo has `user` whereas userprofile has `id`:

This is a breaking API change only for `/users/handles/<domain>/<handle>` API and not for `/users/handles/<handle>`. 

`/users/handles/<handle>` returns `UserHandleInfo`.
`/users/handles/<domain>/<handle>` shouldn't be used by anyone yet, so it is safe to break.

services/brig/test/integration/API/Federation.hs
akshaymankar: Yes, this is a breaking change to the federation API and I think it is fine to make this in our current state. Is it not?

jschaul: > shouldn't be used by anyone yet, so it is safe to break.

I couldn't find any recent usage of that in the logs, so maybe it's not in use yet indeed.

Still, the userhandle object gives `qualified_user` whereas the userprofile gives `qualified_id`. It's not consistent and it's not clear to me why we can't be consistent in the naming of these things.

services/brig/src/Brig/API/Public.hs
akshaymankar: There is a bug in pipeline! The tests are fine, phew!

services/brig/test/integration/API/User/Handles.hs
```diff
@@ -56,7 +57,9 @@ tests _cl _at conf p b c g =
       test p "handles/race" $ testHandleRace b,
       test p "handles/query" $ testHandleQuery conf b,
       test p "handles/query - team-search-visibility SearchVisibilityStandard" $ testHandleQuerySearchVisibilityStandard conf b,
-      test p "handles/query - team-search-visibility SearchVisibilityNoNameOutsideTeam" $ testHandleQuerySearchVisibilityNoNameOutsideTeam conf b g
+      test p "handles/query - team-search-visibility SearchVisibilityNoNameOutsideTeam" $ testHandleQuerySearchVisibilityNoNameOutsideTeam conf b g,
+      test p "GET /users/handles/<handle>" $ testGetUserByUnqualifiedHandle b,
+      test p "GET /users/handles/<handle>" $ testGetUserByQualifiedHandle b
```
akshaymankar, line 62:
```suggestion
      test p "GET /users/handles/<domain>/<handle>" $ testGetUserByQualifiedHandle b
```

jschaul (approved):
As discussed, since the existing endpoint wasn't in use yet it's okay to make a slight breaking change to it. 

The merge from develop has some issue however: the diff shows the addition of a range of extra unrelated files. Perhaps a clean merge or clean rebase could be done with latest develop?

PR #1406, opened 2021-03-09
stack-deps.nix: Use nixpkgs from niv

akshaymankar:


smatting (approved):
Approved for more reproduceability

PR #1408, opened 2021-03-15
Handle errors which could happen while talking to remote federator

akshaymankar:


services/brig/src/Brig/API/Public.hs
```diff
@@ -186,6 +197,36 @@ instance ToSchema Empty404 where
...
+          & type_ ?~ SwaggerObject
+          & properties
+            .~ InsOrdHashMap.fromList
+              [ ("code", natSchema),
+                ("label", textSchema),
+                ("message", textSchema)
+              ]
+          & required .~ ["code", "label", "message"]
+
+-- declareNamedSchema (Proxy @Text) <&> (schema . description ?~ "user not found")
```
jschaul, line 228: dead code?

services/brig/src/Brig/API/Public.hs
```diff
@@ -382,6 +428,9 @@ type SwaggerDocsAPI = "api" :> SwaggerSchemaUI "swagger-ui" "swagger.json"
 
 type ServantAPI = OutsideWorldAPI
 
+-- type family WithFederationErrors (xs :: [*]) :: [*] where
+--   WithFederationErrors xs = xs :++: [WithStatus 520 ()]
```
jschaul, line 432: dead code?

services/brig/src/Brig/API/Public.hs
```diff
@@ -1417,15 +1466,22 @@ checkHandlesH (_ ::: _ ::: req) = do
...
-getUserByHandleH :: UserId -> Domain -> Handle -> Handler Public.UserProfile
-getUserByHandleH self domain handle =
-  ifNothing (notFound "handle not found")
-    =<< getHandleInfo self (Qualified handle domain)
+  profile <- getUserByHandleH self domain handle
+  case profile of
+    Z (I (WithStatus userProfile)) ->
+      Servant.respond . WithStatus @200 . Public.UserHandleInfo . Public.profileQualifiedId $ userProfile
+    S (Z (I x)) -> Servant.respond x
+    S (S x) -> case x of
```
jschaul, line 1477: This `S (X (I x))` is very confusing syntax. I'd be in favour of either removing this complexity to leak into these method bodies here. Having slightly improved swagger is not worth the tradeoff to have to write this kind of non-intuitive code in my opinion. Also, it's unclear why the profile is converted to a UserHandleInfo here. This is confusing and needs at least some comments. Or instead why not call another Data.lookupfunction that was there before?

jschaul (commented):
Looks good overall, see nitpick inline comments.

Also, I think we should document the list of error codes we came up with and their meanings/usages. How about adding some information to `docs/reference/federation/pull-requests/1408_federation_error_codes.md` ? It might be good to cross-link the PR and the Jira issue, too, in both directions.

libs/wire-api-federation/proto/router.proto
```diff
@@ -20,13 +20,43 @@ message HTTPResponse {
...
+    InvalidCertificate = 4;
+    VersionMismatch = 5;
+    FederationDeniedByRemote = 6;
+    FederationDeniedLocally = 7;
+    RemoteFederatorError = 8;
+    InvalidRequest = 9;
+  }
+
+  ErrorType type = 1;
+  ErrorPayload payload = 1;
```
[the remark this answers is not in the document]
akshaymankar: Yeah, I didn't write any tests for payload so it didn't get caught. I would've hoped this gets caught at compile time.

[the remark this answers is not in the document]
akshaymankar: I am so confused, this is not causing any problems!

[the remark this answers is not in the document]
akshaymankar: Of course it doesn't work with anything apart from a mu-haskell client

services/brig/src/Brig/API/Public.hs
akshaymankar: Oh yeah, I wanted to ask @arianvp and @fisx if they know of a better way to do this.

libs/wire-api-federation/proto/router.proto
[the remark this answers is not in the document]
akshaymankar: Fixed it now.

services/brig/src/Brig/API/Public.hs
akshaymankar: Removed it, let's talk about what it would take for UVerb to be ready.

jschaul (approved):
Looks good if CI agrees! (I have a feeling there may be some unused imports still maybe?)

PR #1414, opened 2021-03-18
Brig: Return correct status phrase and body on error

akshaymankar:
Previously all the servant endpoints were wrongly fabricating the phrase for
status and only putting the message in the body. So, some 404's would return
something like:

```
HTTP/1.1 404 handle-not-found
```

This is allowed in HTTP, but still is a breaking change to our API. So, I think
it should be reverted.

Apart from this we also broke our API by not writing the whole error object in
the body, fortunately nothing broke, but I am pretty sure something would have
broken if we touched the endpoints where clients try to parse the error labels
and messages.

### DIFF
--- a/services/brig/src/Brig/API/Handler.hs
+++ b/services/brig/src/Brig/API/Handler.hs
@@ -43,11 +43,11 @@ import qualified Control.Monad.Catch as Catch
 import Data.Aeson (FromJSON)
 import qualified Data.Aeson as Aeson
 import Data.Default (def)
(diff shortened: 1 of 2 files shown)


jschaul:
Any chance for a test, or a before/after example?

akshaymankar:
> Any chance for a test, or a before/after example?

I added the test, it will fail right now, but should start succeeding after https://github.com/wireapp/wire-server/pull/1408 is merged.

fisx (approved):

PR #1415, opened 2021-03-18
Fix: Debian package for spar-migrate-data

smatting:


services/spar/migrate-data/deb/opt/spar/spar-migrate-data/spar-migrate-data
```diff
@@ -0,0 +1 @@
+../../../../../../../dist/spar-migrate-data
```
akshaymankar, line 1: Best hacks live forever :1st_place_medal: 

Can you please write about this somewhere?

akshaymankar (approved):

PR #1416, opened 2021-03-19
Random prekeys strategy

jmatsushita:
Implement https://wearezeta.atlassian.net/browse/SQPIT-348

TODO:
 - [x] Parse `randomPrekeys: true` into `Maybe ()` instead of `randomPrekeys: []` in `brig.yaml` configuration.
 - [x] Make the DynamoDB endpoint configuration optional
 - [x] Allow both strategies to be tested in the integration tests
   - [x] Local integration tests use the randomPrekey strategy
   - [x] CI integration tests use the DynamoDB lock.
 - [ ] Deploy to staging
   - [ ] Merge to develop
   - [ ] Update brig..yaml in staging
   - [ ] Switch CI integration test to use randomPrekey strategy

### DIFF
--- a/services/brig/brig.integration.yaml
+++ b/services/brig/brig.integration.yaml
@@ -34,9 +34,10 @@ aws:
   # ^ Comment this out if you don't want to journal user events
   prekeyTable: integration-brig-prekeys
   sqsEndpoint: http://localhost:4568 # https://sqs.eu-west-1.amazonaws.com
-  dynamoDBEndpoint: http://localhost:4567 # https://dynamodb.eu-west-1.amazonaws.com
+  # dynamoDBEndpoint: http://localhost:4567 # https://dynamodb.eu-west-1.amazonaws.com
 
-randomPrekeys: []
+# Uncomment to use the randomPrekey allocation strategy instead of dynamoDB
+randomPrekeys: true
 
 # Uncomment this if you want STOMP.
 #
--- a/services/brig/brig.integration.yaml
+++ b/services/brig/brig.integration.yaml
@@ -36,6 +36,8 @@ aws:
   sqsEndpoint: http://localhost:4568 # https://sqs.eu-west-1.amazonaws.com
   dynamoDBEndpoint: http://localhost:4567 # https://dynamodb.eu-west-1.amazonaws.com
 
+randomPrekeys: []
+
 # Uncomment this if you want STOMP.
 #
 # stomp:
--- a/services/brig/src/Brig/App.hs
+++ b/services/brig/src/Brig/App.hs
@@ -173,7 +173,7 @@ data Env = Env
     _digestSHA256 :: Digest,
     _digestMD5 :: Digest,
     _indexEnv :: IndexEnv,
-    _randomPrekeys :: Maybe (),
+    _randomPrekeys :: Bool,
     _randomPrekeyLocalLock :: MVar ()
   }
 
@@ -251,7 +251,7 @@ newEnv o = do
         _digestMD5 = md5,
         _digestSHA256 = sha256,
         _indexEnv = mkIndexEnv o lgr mgr mtr,
-        _randomPrekeys = Just (),
+        _randomPrekeys = fromMaybe False $ Opt.randomPrekeys o,
         _randomPrekeyLocalLock = prekeyLocalLock
       }
   where
--- a/services/brig/src/Brig/App.hs
+++ b/services/brig/src/Brig/App.hs
@@ -55,7 +55,6 @@ module Brig.App
     sftEnv,
     internalEvents,
     emailSender,
-    randomPrekeys,
     randomPrekeyLocalLock,
 
     -- * App Monad
@@ -173,8 +172,7 @@ data Env = Env
     _digestSHA256 :: Digest,
     _digestMD5 :: Digest,
     _indexEnv :: IndexEnv,
-    _randomPrekeys :: Bool,
-    _randomPrekeyLocalLock :: MVar ()
+    _randomPrekeyLocalLock :: Maybe (MVar ())
   }
 
 makeLenses ''Env
@@ -216,7 +214,9 @@ newEnv o = do
     StompQueue q -> pure (StompQueue q)
     SqsQueue q -> SqsQueue <$> AWS.getQueueUrl (aws ^. AWS.amazonkaEnv) q
   mSFTEnv <- mapM Calling.mkSFTEnv $ Opt.sft o
-  prekeyLocalLock <- newMVar ()
+  prekeyLocalLock <- case Opt.randomPrekeys o of
+    Just True -> Just <$> newMVar ()
+    _ -> pure Nothing
   return
     $! Env
       { _cargohold = mkEndpoint $ Opt.cargohold o,
@@ -251,7 +251,6 @@ newEnv o = do
         _digestMD5 = md5,
         _digestSHA256 = sha256,
         _indexEnv = mkIndexEnv o lgr mgr mtr,
-        _randomPrekeys = fromMaybe False $ Opt.randomPrekeys o,
         _randomPrekeyLocalLock = prekeyLocalLock
       }
   where
--- a/services/brig/src/Brig/App.hs
+++ b/services/brig/src/Brig/App.hs
@@ -56,6 +56,7 @@ module Brig.App
     internalEvents,
     emailSender,
     randomPrekeys,
+    randomPrekeyLocalLock,
 
     -- * App Monad
     AppT,
@@ -172,7 +173,8 @@ data Env = Env
     _digestSHA256 :: Digest,
     _digestMD5 :: Digest,
     _indexEnv :: IndexEnv,
-    _randomPrekeys :: Maybe ()
+    _randomPrekeys :: Maybe (),
+    _randomPrekeyLocalLock :: MVar ()
   }
 
 makeLenses ''Env
@@ -214,6 +216,7 @@ newEnv o = do
     StompQueue q -> pure (StompQueue q)
     SqsQueue q -> SqsQueue <$> AWS.getQueueUrl (aws ^. AWS.amazonkaEnv) q
   mSFTEnv <- mapM Calling.mkSFTEnv $ Opt.sft o
+  prekeyLocalLock <- newMVar ()
   return
     $! Env
       { _cargohold = mkEndpoint $ Opt.cargohold o,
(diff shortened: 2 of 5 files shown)


services/brig/src/Brig/Data/Client.hs
```diff
@@ -187,19 +188,44 @@ updatePrekeys u c pks = do
...
+                  . msg (val "last resort prekey used")
+          return $ Just (ClientPrekey c (Prekey i k))
+    removeAndReturnPreKey Nothing = return Nothing
+
+    pickRandomPrekey :: [(PrekeyId, Text)] -> AppIO (Maybe (PrekeyId, Text))
+    pickRandomPrekey [] = return Nothing 
+    -- unless we only have one key left
+    pickRandomPrekey [pk] = return $ Just pk
+    -- pick among list of keys, except lastPrekeyId
+    pickRandomPrekey pks = do
```
franziskuskiefer, line 225: We should not actually pick the last resort prekey unless all other prekeys are used (it gives us less security because it will be re-used). So instead I would suggest to pick the random prekey from all but the last one and only pick the last resort key if there is no key left.

franziskuskiefer (commented):
*Drive by review*

services/brig/src/Brig/Data/Client.hs
jmatsushita: Hey @franziskuskiefer that's exactly what this code does unless I misunderstand 😄 
 - if we have only one key left we return that one, [pattern matching the singleton list on line 223](https://github.com/wireapp/wire-server/pull/1416/files#diff-4902a73060613f1a4272a1912557acc1698ef6bf17271e658e3d6f7f067a65edR223) (it should be the `lastPrekey` but we don't actually check that) and the [`removeAndReturnPreKey`](https://github.com/wireapp/wire-server/pull/1416/files#diff-4902a73060613f1a4272a1912557acc1698ef6bf17271e658e3d6f7f067a65edR208) won't delete it from the database.
 - if we have more than one key left (pattern matching any other list that's neither empty nor singleton):
   - we filter out the `lastPrekeyId` from the list of all prekeys [into a `pks'` list on line 226](https://github.com/wireapp/wire-server/pull/1416/files#diff-4902a73060613f1a4272a1912557acc1698ef6bf17271e658e3d6f7f067a65edR226)
   - and we pick at random in the filtered out list `pks'`.  ([line 227](https://github.com/wireapp/wire-server/pull/1416/files#diff-4902a73060613f1a4272a1912557acc1698ef6bf17271e658e3d6f7f067a65edR227))

So the `lastPrekeyId` never get used until it's the last one remaining.

franziskuskiefer: Ah right. I didn't properly look at line 226, i.e that `pks'` doesn't contain the last resort prekey. Then this is doing the right thing! 🙂

services/brig/brig.integration.yaml
```diff
@@ -34,7 +34,10 @@ aws:
   # ^ Comment this out if you don't want to journal user events
   prekeyTable: integration-brig-prekeys
   sqsEndpoint: http://localhost:4568 # https://sqs.eu-west-1.amazonaws.com
-  dynamoDBEndpoint: http://localhost:4567 # https://dynamodb.eu-west-1.amazonaws.com
+  # dynamoDBEndpoint: http://localhost:4567 # https://dynamodb.eu-west-1.amazonaws.com
```
fisx, line 37: Are you sure you want to keep that line here as a comment if we want people to not do that any more?

services/brig/src/Brig/Options.hs
```diff
@@ -353,6 +353,8 @@ data Opts = Opts
     elasticsearch :: !ElasticSearchOpts,
     -- | AWS settings
     aws :: !AWSOpts,
+    -- | Enable Random Prekey Strategy
+    randomPrekeys :: !(Maybe Bool),
```
fisx, line 357: (I would like this more if it was a `Bool` and would parse an optional json boolean with `fromMaybe False`, but I think your pattern is the one we're using in other places, too.)

services/brig/src/Brig/Data/Client.hs
```diff
@@ -187,10 +189,23 @@ updatePrekeys u c pks = do
...
 
 claimPrekey :: UserId -> ClientId -> AppIO (Maybe ClientPrekey)
-claimPrekey u c = withOptLock u c $ do
-  prekey <- retry x1 $ query1 userPrekeys (params Quorum (u, c))
-  case prekey of
-    Just (i, k) -> do
+claimPrekey u c =
+  ifM
+    (view randomPrekeys)
+    -- Use random prekey selection strategy
```
fisx, line 195: You could make the `then` and `else` blocks functions, give them good names, and avoid having to write comments here.  :)

services/brig/src/Brig/App.hs
```diff
@@ -170,7 +172,9 @@ data Env = Env
     _zauthEnv :: ZAuth.Env,
     _digestSHA256 :: Digest,
     _digestMD5 :: Digest,
-    _indexEnv :: IndexEnv
+    _indexEnv :: IndexEnv,
+    _randomPrekeys :: Bool,
+    _randomPrekeyLocalLock :: MVar ()
```
fisx, line 177: There shouldn't be a local lock for random prekeys if `_randomPrekeys` is `False`.  You could do something like this:

```
    _randomPrekeyLocalLock :: Maybe (MVar ())
```

and then:

```
randomPrekeys :: Getter' Env Bool
randomPrekeys = to $ isJust <$> view randomPrekeyLocalLock
```

In `Brig.Data.Client`, you can then case-switch over the maybe-lock instead of the `ifM`, and you'll have fewer representable bad states.

I think this would be easy, but I'm not sure it's important enough to do it?

fisx (approved):
All good!  I tried really hard to come up with something to complain about, but you can safely ignore all my comments.  :)

services/brig/src/Brig/Data/Client.hs
```diff
@@ -199,7 +214,17 @@ claimPrekey u c = withOptLock u c $ do
...
+    removeAndReturnPreKey Nothing = return Nothing
+
+    pickRandomPrekey :: [(PrekeyId, Text)] -> AppIO (Maybe (PrekeyId, Text))
+    pickRandomPrekey [] = return Nothing
+    -- unless we only have one key left
+    pickRandomPrekey [pk] = return $ Just pk
+    -- pick among list of keys, except lastPrekeyId
+    pickRandomPrekey pks = do
+      let pks' = filter (\k -> fst k /= lastPrekeyId) pks
+      ind <- liftIO $ randomRIO (0, length pks' - 1)
```
fisx, line 226: Ah, one more useless pedanticism: I think cryptonite puts more effort into their PRNGs, but this use is sensitive to attacks against randomness, so it should be fine.

fisx:
...  and maybe have a chat with QA on how they test this?  But now that I think about it, what's the difference in observable behavior for the clients?  Is there any?

services/brig/src/Brig/Data/Client.hs
franziskuskiefer: This is actually a good point. @jmatsushita what PRNG is used here? It's not much of a security issue here that's why I didn't think about it. But with a bad PRNG we of course increase the likelihood of errors seen by clients.

services/brig/brig.integration.yaml
jmatsushita: In my view we should do another pass (particularly with documentation) once we've convinced ourselves that all works well in staging.

services/brig/src/Brig/Options.hs
jmatsushita: Yes I wondered what was the cultural pattern :) In the end having the Env be Bool, but the Opt be Maybe Bool seemed like an ok choice to me.

services/brig/src/Brig/Data/Client.hs
jmatsushita: That's a style question but I did think of that and thought that I like to have the "workhorse" functions not be punted down to a where clause, `withLocalLock` and `withOptLock` are fairly self documenting already, commenting is caring :)

services/brig/src/Brig/App.hs
jmatsushita: It's cleaner I agree. On it.

akshaymankar (commented):
Doesn't have to be in this PR, but can we also stop deploying the fake dynamoDB thing for integration tests and ephemeral environments?

akshaymankar:
> ... and maybe have a chat with QA on how they test this? But now that I think about it, what's the difference in observable behavior for the clients? Is there any?

The change in observable behaviour is that clients will see more key collisions than earlier as we have reduced the guarantees. I am not sure if we ever test the limits of our guarantees.

services/brig/src/Brig/Data/Client.hs
jmatsushita: It uses `randomRIO` from the `random` package https://hackage.haskell.org/package/random-1.2.0/docs/System-Random.html which uses the `splitmix` algorithm https://hackage.haskell.org/package/splitmix which is advertised as safe for randomized data structures (which I think is our case) but not for cryptographic uses. I would think it's a good fit for our use case, but just say the word and I'll change it to use Cryptonite's PRNG.

fisx: I'm wondering now: what's a good definition of a non-cryptographic PRNG?

(No need to change it, I think.)

franziskuskiefer: Non-cryptographic RNGs like this one can for example be predictable. That's fine in a case like this, but fatal when used for any cryptographic operation. Most non-cryptographic RNGs trade this and other security properties for speed.
I don't see the need for a cryptographic one and the one used here appears to be good enough.
One attack that's for example possible with the predictability of this RNG is that an attacker can, after learning the sequence, try to produce a collision for a specific target. But this far from straight forward and would only lead to the victim having to re-establish the attacked session.
So I agree, no need to change it.

franziskuskiefer: Non-cryptographic RNGs like this one can for example be predictable. That's fine in a case like this, but fatal when used for any cryptographic operation. Most non-cryptographic RNGs trade this and other security properties for speed.
I don't see the need for a cryptographic one and the one used here appears to be good enough.
One attack that's for example possible with the predictability of this RNG is that an attacker can, after learning the sequence, try to produce a collision for a specific target. But this far from straight forward and would only lead to the victim having to re-establish the attacked session.
So I agree, no need to change it.

jmatsushita: Excellent, thanks a lot for the review :)

jmatsushita:
Thanks for the suggestion to raise this more clearly with QA. I've created a ticket for the core squad about it, please comment there if you have more thoughts https://wearezeta.atlassian.net/browse/SQCORE-562

PR #1417, opened 2021-03-19
Add spar-migrate-data to cassandara-migrations helm chart

smatting:
This PR adds `spar-migrate-data.yaml` (similar to `galley-migrate-data.yaml` in the same directory) to the `cassandra-migrations` chart.
This should be released after https://github.com/wireapp/wire-server-private/pull/283


jschaul (approved):
Looks good to me. Let's see if CI agrees :)

PR #1424, opened 2021-03-25
Remove unused dependencies

akshaymankar:
I used [prune-juice](https://github.com/dfithian/prune-juice) to discover these.


fisx (approved):
Nice!  Would it make sense / be easy to add a small shell script to `/tools/` and hook it into `make integration`?  Or just the script, if you prefer.  It's just really nice for documentation.

akshaymankar:
> Nice! Would it make sense / be easy to add a small shell script to `/tools/` and hook it into `make integration`? Or just the script, if you prefer. It's just really nice for documentation.

There are a few false postives in the result, so I didn't add a script.

akshaymankar:
I just tried to add it to `stack.yaml`, but it fails to install. It is not even part of nixpkgs, so I couldn't add it to shell.nix either. After installing, it is just the matter of typing `prune-juice`. I think the fact that I wrote it in the jira ticket is enough documentation for us to get to automating this as I don't really want to spend time chasing dependencies right now.

PR #1432, opened 2021-04-06
hack/bin/integration-cleanup.sh: Use jq magic instead of bash

akshaymankar:
For some reason the the bash expression to calculate time fails in CI and helm
now supports outputting json, so I swapped bash magic with jq magic. It more
robust, hopefully it is also more understandable.

### DIFF
--- a/hack/bin/filter-old-releases.jq
+++ b/hack/bin/filter-old-releases.jq
@@ -1,0 +1,9 @@
+.[]
+  | .parsedUpdatedTime = ( .updated
+                         | sub("(?<time>.*)\\.[\\d]+(?<tz>.*)"; "\(.time)\(.tz)")
+                         | strptime("%Y-%m-%d %H:%M:%S %z %Z")
+                         | mktime
+                         )
+  | select ( .parsedTime < (now - 3 * 60 * 60))
+  | [.name, .namespace]
+  | @tsv
--- a/hack/bin/filter-old-releases.jq
+++ b/hack/bin/filter-old-releases.jq
@@ -4,6 +4,6 @@
                          | strptime("%Y-%m-%d %H:%M:%S %z %Z")
                          | mktime
                          )
-  | select ( .parsedTime < (now - 3 * 60 * 60))
+  | select ( .parsedUpdatedTime < (now - 3 * 60 * 60))
   | [.name, .namespace]
   | @tsv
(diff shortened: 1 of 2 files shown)


hack/bin/filter-old-releases.jq
```diff
@@ -0,0 +1,9 @@
+.[]
+  | .parsedUpdatedTime = ( .updated
+                         | sub("(?<time>.*)\\.[\\d]+(?<tz>.*)"; "\(.time)\(.tz)")
+                         | strptime("%Y-%m-%d %H:%M:%S %z %Z")
+                         | mktime
+                         )
+  | select ( .parsedTime < (now - 3 * 60 * 60))
+  | [.name, .namespace]
+  | @tsv
```
lucendio, line 9: question: why tabs?

lucendio (approved):
LGTM

hack/bin/integration-cleanup.sh
```diff
@@ -4,31 +4,15 @@
...
-
-IFS=$'\n'
-for release in $(helm list -A | grep test-); do
-  name=$(echo "$release" | awk '{print $1}')
-  namespace=$(echo "$release" | awk '{print $2}')
-  date_str=$(echo "$release" | awk '{print $4" "$5" "$6}')
-  date_s=$(date_seconds "$date_str")
-  diff=$(( $date_now - $date_s ))
-  if [ $diff -ge 7200 ]; then
-    echo "test release '$name' older than 2 hours; deleting..."
```
fisx, line 29: wanna keep that echo statement in some modified form?

hack/bin/filter-old-releases.jq
akshaymankar: I think jq can either output a json array, a csv or a tsv. Out of these I thought tsv was the most compatible with bash as I can `awk '{print $1}'`. Do you have a better idea?

jschaul (approved):
There were some problems with the script still

- it errors if `releases` is an empty string:

```
./hack/bin/integration-cleanup.sh
test release '' older than 2 hours; deleting...
Error: uninstall: Release name is invalid: 
```

I added a commit which should fix that.

akshaymankar:
The tests are failing because this incorrectly deletes releases deployed less than 2 hours ago, I will look into that.

akshaymankar:
Maybe I was a fool to assume jq is safer than bash :cry: https://github.com/wireapp/wire-server/pull/1432/commits/c225f41a6fe66623a53152a39262648f2efe145a

jschaul:
All green now, it seems!

hack/bin/filter-old-releases.jq
lucendio: I just stumbled upon it and was curious, as `FS` usually defaults to *space*.

PR #1436, opened 2021-04-08
Document how to run multi-backend tests [skip ci]

jschaul:


docs/developer/how-to.md
```diff
@@ -17,3 +17,77 @@ Open your browser at:
...
+latest tag for nginz:
+2.104.11
+```
+
+Let's assume the tags are the same(*) for both, then export an environment variable:
+
+```
+export DOCKER_TAG=2.104.11
+```
+
```
pcapriotti, line 52: Could this be automated?

docs/developer/how-to.md
```diff
@@ -17,3 +17,77 @@ Open your browser at:
...
+Now run the following, which will create two full installations of wire-server on the kubernetes cluster you've configured to connect to, and should take ~10 minutes.
+
+```
+make kube-integration-setup
+```
+
+Next, you can choose to either run all integration tests, which includes running the multi-backend integration tests, and takes another ~10 minutes. Or you can instead choose to *only* run the multi-backend tests.
+
+Run all integration tests (takes ~10 minutes) with: `make kube-integration-tests` - these tests run entirely on kubernetes.
```
pcapriotti, line 61: When I tried this, it worked fine, but I got no output until the very end. Is this expected? If so, maybe add a note to that effect.

pcapriotti (commented):
Documentation looks good, I was able to follow and run the tests.

In the TODO section, I'm not quite sure what kind of workflow that is supposed to help with, so maybe it would be good to add some explanation.

docs/developer/how-to.md
```diff
@@ -17,3 +17,77 @@ Open your browser at:
...
+
+Run only the multi-backend tests:
```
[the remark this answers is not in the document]
jschaul: The tests are indeed run on kubernetes, but in a different way, without using telepresence. I tried to reformulate the text to make that clearer.

docs/developer/how-to.md
jschaul: CI already runs these tests, so there it's already automated. For local purposes, chances are that you don't want to run these tests often, and if you do, you probably want control over which code should be tested. That makes further automating the "test the code on develop from a local computer" probably not very useful.

docs/developer/how-to.md
jschaul: Yes, it's expected, kind of, but a limitation of using the underlying `helm test`. Maybe we want to run integration tests differently at some point, for better control, but that's another topic. I added a comment here.

pcapriotti:
Looks great!

pcapriotti (approved):

PR #1437, opened 2021-04-09
Support chaning email via SCIM

smatting:


mdimjasevic:
@smatting and @fisx , any update on this?

PR #1439, opened 2021-04-12
Update update-helm-charts-s3.sh arguments

smatting:
This PR changes `update-helm-charts-s3.sh `

- new option `-r <wire|wire-develop|wire-custom>` - this was derived from the branch beforehand
- new option `-d CHART_DIR` - this was a positional argument before

Tested: options parsing works.
Not tested: compatibility with uses of the script (CI, where else?)

Make sure to deploy https://github.com/wireapp/wire-server-private/pull/291 after this

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -334,14 +334,14 @@ clean-charts:
 # (assummption: CI sets DOCKER_TAG and HELM_SEMVER)
 .PHONY: upload-chart-%
 upload-chart-%: release-chart-%
-	./hack/bin/upload-helm-charts-s3.sh .local/charts/$(*)
+	./hack/bin/upload-helm-charts-s3.sh -r $(HELM_REPO) -d .local/charts/$(*)
 
 # Usecases for this make target:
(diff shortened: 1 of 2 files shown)


arianvp:
> Not tested: compatibility with uses of the script (CI, where else?)

No-where else. it's only used in our CI pipelines.

arianvp:
This is where the script gets called:

https://github.com/wireapp/wire-server/blob/develop/Makefile#L336

https://github.com/wireapp/wire-server-private/blob/957bbe983fc874728bc678f25cd08f5012121b82/ci/pipelines/tasks/scripts/update-helm-chart-images.sh#L36

I guess we should adjust somehow these scripts?

smatting:
Guard against empty `HELM_REPO` not strictly necessary, because `update-helm-charts-s3.sh`  errors on empty `-r` argument there.

jschaul (approved):

jschaul (commented):
Loving bash yet Stefan? 😆

PR #1441, opened 2021-04-12
Brig: Ensure servant APIs are recorded by the metrics middleware

jschaul:


jschaul:
🥇 Oh wow that treeLookup is more complex than I thought. Change looks good!

PR #1444, opened 2021-04-13
galley: Convert conversation endpoints to servant

smatting:


services/galley/src/Galley/API/Public.hs
```diff
@@ -72,6 +87,65 @@ import qualified Wire.API.Team.SearchVisibility as Public
...
+    Servant.hoistServerWithContext (Proxy @(InternalAuth :> api)) pc nt s
+
+type GetTeamConversationRoles =
+  Summary "Get existing roles available for the given team"
+    :> ZAuthServant
+    :> "teams"
+    :> Capture "tid" TeamId
+    :> "conversations"
+    :> "roles"
+    :> Get '[Servant.JSON] Public.ConversationRolesList
```
akshaymankar, line 129: I think we should use the generics based way of doing these endpoints, similar to what we're doing here: https://github.com/wireapp/wire-server/blob/73b47fce0d53a56b81cf17b7c4e60d0f9f0c80ed/libs/wire-api-federation/src/Wire/API/Federation/API/Brig.hs#L30-L39

This will generate better errors and help us keep all of these `type`s together. I started doing it like this in brig, because I wasn't aware of the other pattern.

smatting: Agree. Should the apis live in a separate package? Currently you'd need on `brig` for the brig api and on `galley` for the api. If we want to use `servant-client` this could create cyclic dependencies.

jschaul: We could put them into galley-types and brig-types, respectively?

akshaymankar: What if the APIs lived in `wire-api`?

pcapriotti:
I've continued work on this PR on top of #1461 in [this branch](https://github.com/wireapp/wire-server/tree/pcapriotti/sqcore-603-plus-typed-schema). Once #1461 is merged, I'll rebase and continue from here.

services/galley/src/Galley/API/Public.hs
```diff
@@ -72,6 +95,239 @@ import qualified Wire.API.Team.SearchVisibility as Public
...
+      routes
+        :- Summary "Get existing roles available for the given conversation"
+        :> ZUser
+        :> "conversations"
+        :> Capture "cnv" ConvId
+        :> Get '[Servant.JSON] Public.ConversationRolesList,
+    getConversationIds ::
+      routes
+        :- Summary "Get all conversation IDs."
+        -- TODO: add note "At most 1000 IDs are returned per request"
```
akshaymankar, line 174: Does this not get added automatically due to the `ToSchema` instance of `Range`?

akshaymankar (approved):
Suggested minor changes. Looks good otherwise. This should've been marked ready for review right?

services/galley/src/Galley/API/Public.hs
pcapriotti: Not yet, but good point, that instance could include the bounds. I'll add a note.

PR #1445, opened 2021-04-14
Use servant-client to make federated calls

akshaymankar:

PR #1446, opened 2021-04-14
More federation-aware API endpoints in brig

pcapriotti:
## Add federation API endpoints:

- [x] users by ids
- [x] prekey
- [x] prekey bundle
- [x] multi-user prekey bundle

## Hook qualified public API to federator

- [x] users by ids
- [x] prekey
- [x] prekey bundle
- [x] multi-user prekey bundle

## Tests

- [x] federation endpoints integration
- [x] multi-backend integration

## TODOs

- [x] ~~make `/list-users` return an object~~ (TBD in another PR)

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -3,4 +3,8 @@
 **/target
 **/*.aci
 **/*.tgz
+.stack-work
+.stack-work-buildah
+.stack-root-buildah
+.local
 services/nginz/src/objs
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,10 +15,6 @@
 -->
 # [2021-03-23]
 
-## Release Notes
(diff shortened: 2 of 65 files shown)


jschaul (approved):
Looks good! See minor inline nitpicks.

I suppose that some of the changes we touched upon in discussions (such as moving federation API in brig to make use of more post requests and extensible types rather than lists) is a refactoring planned for a future PR?

pcapriotti:
> I suppose that some of the changes we touched upon in discussions (such as moving federation API in brig to make use of more post requests and extensible types rather than lists) is a refactoring planned for a future PR?

Yes, I think it's better to do that on top of #1445.

jschaul (approved):

PR #1449, opened 2021-04-15
HLS workaround for optimization flags

smatting:
This PR updates `hie.yaml` Makefile target with @pcapriotti 's  workound for this problem

> I think most of the issues I am having with my lsp setup are caused by the fact that there is a difference in flags between the make build and the one triggered by hie-bios (which is used by the language server).

> I couldn't find a way to pass options to the stack process run by hie-bios, but I've found a workaround: create a separate config file `stack-dev.yaml`, which is a copy of stack.yaml, plus the following:
```
ghc-options:
  "$locals": -O0
```

> then modify `hie.yaml` to look as follows:
```
cradle:
  stack:
    stackYaml: "./stack-dev.yaml"
    components:
      # list of components here
```

> Now hie-bios recompiles with optimisations disabled, and it does seem that its compilation outputs are in sync with whatever make (i.e. `stack build --fast`) produces. This should limit the number of times things get needlessly recompiled.

NOTE: This PR adds the python-based `yq` to the `shell.nix` environment. Can this cause conflicts on CI with our other uses of `yq`?


Makefile
```diff
@@ -218,9 +218,17 @@ libzauth:
 #
 # Run this again after changes to libraries or dependencies.
 .PHONY: hie.yaml
-hie.yaml:
+hie.yaml: stack-dev.yaml
 	stack build implicit-hie
-	stack exec gen-hie > hie.yaml
+	stack exec gen-hie > hie.orig.yaml
+	nix-shell --command 'cat hie.orig.yaml | yq "{cradle: {stack: {stackYaml: \"./stack-dev.yaml\", components: .cradle.stack}}}" > hie.yaml'
```
pcapriotti, line 224: Why not simply
```
stack exec gen-hie | nix-shell --run "yq '...'" > hie.yaml
```
?

pcapriotti (approved):
Looks good!

Makefile
```diff
@@ -218,9 +218,17 @@ libzauth:
...
+
+.PHONY: stack-dev.yaml
+stack-dev.yaml:
+	cp stack.yaml stack-dev.yaml
+	echo -e '\n\nghc-options:\n "$$locals": -O0' >> stack-dev.yaml
```
pcapriotti, line 229: It seems we also want `-Wall` and `-Werror` to replicate `stack build`'s `--pedantic` switch.

pcapriotti:
Unfortunately, this is not quite the whole story. We are also running into [this issue](https://github.com/commercialhaskell/stack/issues/5503) with stack. Someone suggested a workaround there, but it seems a bit painful to set up.

Makefile
smatting: simpler is better, added your suggestion

Makefile
smatting: Added both flags

smatting:
> Unfortunately, this is not quite the whole story. We are also running into [this issue](https://github.com/commercialhaskell/stack/issues/5503) with stack. Someone suggested a workaround there, but it seems a bit painful to set up.

Should be scriptable. We can add it in another PR

PR #1451, opened 2021-04-15
Setup for running tests in kind

akshaymankar:


Makefile
```diff
@@ -350,13 +352,39 @@ echo-release-charts:
...
+.PHONY: kind-delete
+kind-delete:
+	rm -f $(CURDIR)/.local/kind-kubeconfig
+	kind delete cluster --name $(KIND_CLUSTER_NAME)
+
+.PHONY: kind-reset
+kind-reset: kind-delete kind-cluster
+
+.local/kind-kubeconfig:
+	kind get kubeconfig --name $(KIND_CLUSTER_NAME) > .local/kind-kubeconfig
```
jschaul, line 379:
```suggestion
	kind get kubeconfig --name $(KIND_CLUSTER_NAME) > $(CURDIR)/.local/kind-kubeconfig
```

hack/helm_vars/wire-server/values.yaml
```diff
@@ -15,10 +15,12 @@ tags:
   sftd: false
 
 cassandra-migrations:
+  imagePullPolicy: Always
```
jschaul, line 18: Why are these pullPolicy=always suddenly necessary if they were not before? This looks like an unrelated drive-by change.

hack/helm_vars/wire-server/kind-values.yaml
```diff
@@ -0,0 +1,22 @@
+cassandra-migrations:
+  imagePullPolicy: Never
```
jschaul, line 2: what about `spar-migrate-data` and `galley-migrate-data`? Are those missing here for a reason?

jschaul (commented):
Could you update the docs in docs/developer/how-to.md as part of this PR?

Makefile
```diff
@@ -350,13 +352,39 @@ echo-release-charts:
...
-	BUILDAH_PUSH=${BUILDAH_PUSH} EXECUTABLES=$(*) ./hack/bin/buildah-make-images.sh
+	BUILDAH_PUSH=${BUILDAH_PUSH} EXECUTABLES=$(*) KIND_CLUSTER_NAME=${KIND_CLUSTER_NAME} BUILDAH_KIND_LOAD=${BUILDAH_KIND_LOAD} ./hack/bin/buildah-make-images.sh
 
 .PHONY: buildah-clean
 buildah-clean:
 	./hack/bin/buildah-clean.sh
+
+.PHONY: kind-cluster
+kind-cluster:
+	kind create cluster --name $(KIND_CLUSTER_NAME)
```
jschaul, line 368: Would it make sense to use a kind config file like in https://github.com/wireapp/wire-server/pull/1087/files#diff-2b5ffee781d627b4719c83f613ca08512b9cae4919718dc9f535f14ded176b48R1-R11 here?

jschaul: Also, would it make sense to lock a specific version of kubernetes here? With `--image kindest/node:vXYZ`?

Makefile
```diff
@@ -16,6 +16,8 @@ CHARTS_INTEGRATION    := wire-server databases-ephemeral fake-aws nginx-ingress-
 # CHARTS_RELEASE := $(shell find charts/ -maxdepth 1 -type d | xargs -n 1 basename | grep -v charts)
 CHARTS_RELEASE        := wire-server databases-ephemeral fake-aws aws-ingress backoffice calling-test demo-smtp elasticsearch-curator elasticsearch-external fluent-bit minio-external cassandra-external nginx-ingress-controller nginx-ingress-services reaper wire-server-metrics sftd
 BUILDAH_PUSH          ?= 1
```
jschaul, line 18: This PR seems to imply a local workflow, so pushing images to remote seems like not the right default:

```suggestion
BUILDAH_PUSH          ?= 0
```

hack/helm_vars/wire-server/values.yaml
akshaymankar: Yes, it is a drive by change. Earlier, this was not set because it couldn't be set, but for integration tests we should set imagePullPolicy to `Always` for all images that we create during integration test run.

Makefile
akshaymankar: Can we do those things incrementally, as we need them? I would love to keep this as config free as possible.

Makefile
akshaymankar: Hmm, I didn't want to break somebody's workflow. But I guess it is just you and me who use this as of now, so I will break it and also enable BUILDAH_KIND_LOAD.

hack/helm_vars/wire-server/kind-values.yaml
akshaymankar: Just forgot about them, of course things didn't fail because I didn't test with a tag which is not already published :disappointed:

Makefile
akshaymankar: Done.

akshaymankar:
> Could you update the docs in docs/developer/how-to.md as part of this PR?

Done.

Makefile
jschaul: Sure, that works also.

docs/developer/how-to.md
```diff
@@ -107,21 +107,27 @@ Look at a successful job in the `wire-server-pr` pipeline from a job build match
...
+##### (i) Build images
+
+1. Ensure `buildah` is available on your system.
+2. Compile the image using `make buildah-docker`. This will try to upload the
+   images into a `kind` cluster. If you'd prefer uploading images to quay.io,
+   you can run it with `make buildah-docker BUILDAH_PUSH=1 BUILDAH_KIND_LOAD=0`
+
+##### (ii) Run tests in kind
+
+1. Install wire-server using `make kind-integration-setup`.
```
jschaul, line 125:
```suggestion
0. Create a local kind cluster with `make kind-cluster`
1. Install wire-server using `make kind-integration-setup`.
```

jschaul (approved):

PR #1455, opened 2021-04-19
Search: Only perform exact handle match if term is a valid handle

akshaymankar:
This is requried because cassandra doesn't like querying with empty keys.
Ideally we should just not allow making the 'Handle' value if something isn't a
valid handle, a FUTUREWORK has been added for this.


jschaul (approved):

jschaul:
I think this needs an ormolu...

jschaul (approved):

PR #1457, opened 2021-04-19
Add docs/**/wire_scim_token.py

smatting:


fisx:
context: team-settings didn't have a way to create scim tokens until recently, so every team admin who wanted to use scim had to go through a poorly documented, awkward process of either running curl commands themselves or opening a CS ticket and providing admin credentials to their team.

this has been fixed by now, but i would still keep the script for documentation and hacking purposes.

unless @smatting you think it should go away now?

PR #1458, opened 2021-04-19
Ormolu script improvements

pcapriotti:
* Reduce size of ormolu output when running in a terminal

* Make the ormolu script detect Ctrl-C more reliably

### DIFF
--- a/tools/ormolu.sh
+++ b/tools/ormolu.sh
@@ -78,14 +78,20 @@ fi
 
 for hsfile in $(git ls-files | grep '\.hsc\?$'); do
     FAILED=0
-    ormolu --mode $ARG_ORMOLU_MODE --check-idempotence $LANGUAGE_EXTS "$hsfile" || FAILED=1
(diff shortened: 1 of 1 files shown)


jschaul (approved):

tools/ormolu.sh
```diff
@@ -72,14 +72,26 @@ echo "language extensions: $LANGUAGE_EXTS"
 
 FAILURES=0
 
+if [ -t 1 ]; then
```
fisx, line 75: why this trivial conditional?

anyway i like the changes!

pcapriotti: `[ -t 1 ]` returns `0` if the standard output is a `tty`. This check is to prevent the script from printing escape sequences when it runs in CI, or any other environment which is not a shell, basically.

fisx: Ah, my bad: I read `-t` as `isTrue`.  Checking the man page would have helped.  :)

PR #1459, opened 2021-04-20
Add script to colourise test failure output [skip ci]

pcapriotti:
This is useful when equality assertions fail on large objects.


akshaymankar (approved):

jschaul (commented):
How do you best use this script? 👀

pcapriotti:
I just copy the assertion failure output from the integration, run the script, paste in the terminal and hit Ctrl-D. I bit fiddly, but it beats having to find the differences by eye, especially on large objects.

PR #1460, opened 2021-04-23
Fix potential pagination bug

smatting:


fisx (changes requested):
What about [this comment](https://github.com/wireapp/wire-server/blob/2a6fbaa253850d7a79c5d05d8be5f84ee5279dd7/services/galley/src/Galley/Data.hs#L153-L163)?

akshaymankar:
Can a test be added to see if this is indeed a bug or not?

fisx:
> What about [this comment](https://github.com/wireapp/wire-server/blob/2a6fbaa253850d7a79c5d05d8be5f84ee5279dd7/services/galley/src/Galley/Data.hs#L153-L163)?

(it was unrelated.  the `ResultSet` abstraction can be used on any one page, but we still need the `Page` type for moving to the next page.  maybe `ResultSet` was not the best idea ever, it's at the very best confusing.)

fisx:
(it's not a bug, the change has no impact on behavior.  but it's more clear this way.)

fisx (approved):

PR #1461, opened 2021-04-23
RFC: Typed Swagger schemas [skip ci]

pcapriotti:
Defining good `ToSchema` instances is not particularly easy at the moment. One can choose between using the generic-based API in deriving-swagger2, or write instances by hand using the swagger2 Lens interface directly. The former is convenient, but can only be used in limited cases, and it is essentially impossible to attach descriptions to fields. The latter is very tedious and error prone, as there are no type-level guarantees that the schema matches the type in any sense.

So I'm proposing a different API to attach schemas to types. Since approaches based on generics or type classes are very rigid, and make it really hard to specify any metadata which is not already contained in the types (such as documentation), my proposed API is based on value-level combinators, and makes use of the `Applicative` and `Alternative` type classes to reduce the number of combinators needed.

Despite being at the value-level, this is actually *more* type safe than the other approaches. For example, forgetting to specify a field in a record is a type error.

The API is based on the type constructors `Schema` and `NamedSchema`, which are both instances of `Applicative` and `Alternative`. A value of `Schema a` (resp. `NamedSchema a`) represents a Swagger schema (resp. named schema) for the type `a`, as well as all its referenced (named) schemas.

To form a schema for a record type, one can turn an existing schema definition into a field using the `field` combinator, which returns a value of type `Schema a`. Fields can then be combined using the `Applicative` interface of `Schema a`. Similarly, arrays can be specified using the `array` combinator. The `field` and `array` combinators can accept both a `Schema` and a `NamedSchema`.

Even though not strictly necessary, I have included some support for using this API together with type classes. There is a type class `ToTypedSchema` which one can instantiate to define a canonical schema for a type. Accordingly, the `schema` combinator returns the schema canonically associated to a type. There is no need to specify type parameters or pass proxy values, as the type argument of `Schema` makes most types inferrable.

To aid the transition to this API, I've also included a combinator `untypedSchema`, which converts an existing `ToSwagger` instance to a typed schema. This makes it possible to freely schemas defined using the previous approaches with the ones obtained from this API.

I have included in this PR a commit that illustrates the use of the API in practice. This won't be merged with the PR, and is just included here for convenience.

Right now, the type parameter of `Schema` is just a phantom parameter, and the `Applicative` and `Alternative` instances are simply based on the underlying `Monoid` instance of `Swagger.Schema`. However, it should be possible to include some more information in these values so that `ToJSON` and `FromJSON` instances could also be derived automatically.

### DIFF
--- a/libs/wire-api/src/Data/Swagger/Typed.hs
+++ b/libs/wire-api/src/Data/Swagger/Typed.hs
@@ -13,6 +13,8 @@ module Data.Swagger.Typed
     unnamed,
     array,
     enum,
+    handle,
+    coerce,
     schema,
     untypedSchema,
   )
@@ -30,6 +32,41 @@ import Imports
 type Declare = S.Declare (S.Definitions S.Schema)
 
 -- | A typed Swagger schema
+--
+-- /Note/: the 'Applicative' and 'Alternative' instances of 'Schema'
+-- have the same implementation, but different types. For example, the
+-- binary operation of 'Applicative' is:
+-- @
+-- (<*>) :: f (a -> b) -> f a -> f b
+-- @
+-- while that of 'Alternative' is:
+-- @
+-- (<|>) :: f a -> f a -> f a.
+-- @
+-- The simple reason for this distinction is that the types will line
+-- up with the 'Applicative' instance in some cases, and with the
+-- 'Alternative' instance in others, so one needs both.
+--
+-- The deeper reason is that they are conceptually different. A schema like:
+-- @
+-- (,) <$> schemaA <*> schemaB
+-- @
+-- means that you will can have fields from @A@ and @B@ in the
+-- resulting object, while
+-- @
+-- schemaA <|> schemaB
+-- @
+-- is only allowed to have fields from @A@ /or/ from @B@.
+--
+-- Unfortunately, there does not seem to be a way to express the
+-- latter constraint in the resulting schema, and that's why the
+-- implementations are the same. It seems this should be possible in
+-- openapi3, thought.
+--
+-- The difference between 'Applicative' and 'Alternative' also becomes
+-- important when we think about using a typed schema specification to
+-- derive FromJSON instances. Then they would respectively implement
+-- sequencing and alternation of the corresponding parsers.
 newtype Schema (a :: *) = Schema (Declare S.Schema)
   deriving (Functor)
 
@@ -89,9 +126,25 @@ instance HasSchemaRef NamedSchema where
       S.NamedSchema Nothing s ->
         pure (S.Inline s)
 
+-- | Just like 'field', but with a trivial "modifier".
 field' :: HasSchemaRef f => Text -> f a -> Schema a
 field' name = field name id
 
+-- | A schema for an object with a single field.
+--
+-- Multiple fields can be combined into a single object using the
+-- 'Applicative' instance of 'Schema'. Note that 'field' accepts both
+-- a 'NamedSchema' and 'Schema' for the inner field schema.
+--
+-- The second argument is an arbitrary "modifier" function from
+-- 'S.Schema' to itself, to be applied to the field schema. The most
+-- common use for this is to set some values, such as its
+-- description. For example, using lenses:
+-- @
+--     field "name" (description ?~ "The name of this object") schema
+-- @
+-- creates a field called "name" with the given description, with a
+-- type that is instance of 'TypedSchema'.
 field :: HasSchemaRef f => Text -> (S.Schema -> S.Schema) -> f a -> Schema a
 field name f s = Schema $ do
   ref <- schemaRef s
@@ -101,6 +154,10 @@ field name f s = Schema $ do
       & S.type_ ?~ S.SwaggerObject
       & S.properties . at name ?~ ref'
 
+-- | A schema for an array of items.
+--
+-- Just like 'field', it can take both 'NamedSchema' and 'Schema'
+-- definitions.
 array :: HasSchemaRef f => f a -> Schema [a]
 array s = Schema $ do
   ref <- schemaRef s
@@ -109,6 +166,13 @@ array s = Schema $ do
       & S.type_ ?~ S.SwaggerArray
       & S.items ?~ S.SwaggerItemsObject ref
 
+-- | A schema for an enumeration type.
+--
+-- This takes as argument a list of pairs @(value, schema)@ where
+-- @value@ is one of the possible JSON values of the enumeration, and
+-- @schema@ is the corresponding Swagger schema. At the moment, the
+-- schemas are ignored, but they could be used to derive a FromJSON
+-- instance.
 enum :: [(Value, Schema a)] -> Schema a
 enum ss =
   Schema $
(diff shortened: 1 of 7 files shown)


libs/wire-api/src/Data/Swagger/Typed.hs
```diff
@@ -0,0 +1,144 @@
...
+runNamedSchema :: NamedSchema a -> Declare S.NamedSchema
+runNamedSchema (NamedSchema decl descr) = do
+  s <- decl
+  pure $ s & (S.schema . S.description) .~ descr
+
+instance Applicative Schema where
+  pure _ = Schema (pure mempty)
+  Schema a <*> Schema b = Schema ((<>) <$> a <*> b)
+
+instance Alternative Schema where
```
fisx, line 72: So `Applicative` and `Alternative` have the same implementation...  do you introduce both just so you have two operator preferences so you can avoid parenthesis?

libs/wire-api/src/Data/Swagger/Typed.hs
```diff
@@ -0,0 +1,144 @@
...
+untypedSchema :: forall a. S.ToSchema a => NamedSchema a
+untypedSchema = NamedSchema (pure (S.toNamedSchema (Proxy :: Proxy a))) Nothing
+
+class ToTypedSchema a where
+  toTypedSchema :: Proxy a -> NamedSchema a
+
+-- to use with deriving via
+newtype TypedSchema a = TypedSchema a
+
+instance ToTypedSchema a => S.ToSchema (TypedSchema a) where
```
fisx, line 138: Trying to explain this to myself: without `TypedSchema`, this would be an undecidable instance?  Or is there another reason for introducing it?  Can you leave a comment with the motivation?  This may be obvious to normal people, but one line of comment to help out people like me isn't too bad I hope :).

fisx (approved):
Nice!  Are you planning to try and move `Data.Swagger.Typed` upstream into swagger2 (and openapi3, while you're at it)?  (I'm all in favor of keeping it in wire-api for now, but I would like to see a comment there mentioning a PR in the upstream repos, so we remember to remove it when it's time.)

I think you can add descriptions in the ToSchema approach using lenses, and I think I've done it, but I forget the specifics and I'm not sure how elegant or convenient it is.  Your approach certainly makes it easier.

Why don't you want to merge the sample implementations in the same PR?  I think it makes sense to keep the two commits together, the second explains the first very well.

Please get another approval of somebody from somebody more in the loop with federation, but I'm certainly convinced.

pcapriotti:
> Nice! Are you planning to try and move `Data.Swagger.Typed` upstream into swagger2 (and openapi3, while you're at it)? (I'm all in favor of keeping it in wire-api for now, but I would like to see a comment there mentioning a PR in the upstream repos, so we remember to remove it when it's time.)

It probably needs a bit more polishing before it can be upstreamed, so my plan would be to keep it here for the moment, try it out on all of the types that need a Swagger schema, see if it works well, then send a PR upstream.


> I think you can add descriptions in the ToSchema approach using lenses, and I think I've done it, but I forget the specifics and I'm not sure how elegant or convenient it is. Your approach certainly makes it easier.

When using something like `CustomSwagger`, you can add descriptions after the fact by definining an instance manually that calls `defineNamedSchema` for the `CustomSwagger` type, then goes over the fields one by one and adds descriptions using lenses. I've done it myself as well, but

 - the field names you use to attach descriptions are not checked: if you misspell them the descriptions are silently ignored
 - it's not particularly less verbose than just defining the whole schema from scratch.

> Why don't you want to merge the sample implementations in the same PR? I think it makes sense to keep the two commits together, the second explains the first very well.

Ah, sure, I can do that too.

> Please get another approval of somebody from somebody more in the loop with federation, but I'm certainly convinced.

Thanks!

libs/wire-api/src/Data/Swagger/Typed.hs
```diff
@@ -0,0 +1,144 @@
...
+-- | A typed Swagger schema
+newtype Schema (a :: *) = Schema (Declare S.Schema)
+  deriving (Functor)
+
+-- FUTUREWORK: store enough information in 'Schema' to synthesise
+-- ToJSON and FromJSON instances
+
+data NamedSchema a = NamedSchema
+  { _nsSchema :: Declare S.NamedSchema,
+    _nsDescription :: Maybe Text
```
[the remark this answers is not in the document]
pcapriotti: Not sure how, because the `Maybe` is important for the `Applicative` and `Alternative` instances. One could argue that those instances shouldn't be there, but they are convenient in certain cases (e.g. [here](https://github.com/wireapp/wire-server/pull/1461/commits/53a2992e8773451a6940d8808390a454b4202a57#diff-30064c2e38bdfba879d773c11abc7e226ab8ae4d3c7435db650a6a4d95f7fc9dR284)).

Also, if we eventually want this to make it upstream, it should support optional descriptions.

libs/wire-api/src/Data/Swagger/Typed.hs
pcapriotti: The `Applicative` and `Alternative` instances have the same implementation, but different types. For example, the binary operation of `Applicative` is:
```haskell
(<*>) :: f (a -> b) -> f a -> f b
```
while that of `Alternative` is:
```haskell
(<|>) :: f a -> f a -> f a.
```

So the simple reason for the distinction is that the types will line up with the `Applicative` instance in some cases, and with the `Alternative` instance in others, so you need both.

The deeper reason is that they are conceptually different. A schema like:
```haskell
(,) <$> schemaA <*> schemaB
```
means that you will can have fields from `A` and `B` in the resulting object, while
```haskell
schemaA <|> schemaB
```
is only allowed to have fields from `A` *or* from `B`.

Unfortunately, I haven't found a way to actually express the latter constraint in the resulting schema, and that's why the implementations are the same. It seems this should be possible in openapi3, thought.

The difference between `Applicative` and `Alternative` also becomes important when we think about using a typed schema specification to derive `FromJSON` instances. Then they would respectively implement sequencing and alternation of the corresponding parsers.

libs/wire-api/src/Data/Swagger/Typed.hs
pcapriotti: Yes, without the `newtype` it would be an undecidable instance. Besides, an instance like
```haskell
instance ClassA a => ClassB a
```
doesn't really work too well, and can create overlapping instances in certain cases, so it's best avoided.

I guess that's why `deriving via` was introduced. So now you can do:
```haskell
data MyType = ...
  deriving ToSchema via TypedSchema MyType
```
and you get the effect you want with just one short line of boilerplate.

But yes, I will add a comment.

libs/wire-api/src/Data/Swagger/Typed.hs
fisx: Thanks, this all makes perfect sense.  Maybe you can add some of this to the haddocks as well?  (I know documenting haskell code is almost non-idiomatic, but it helped me here, so it may help others.  :))

libs/wire-api/src/Data/Swagger/Typed.hs
```diff
@@ -0,0 +1,144 @@
...
+        S.declare [(n, s)]
+        pure (S.Ref (S.Reference n))
+      S.NamedSchema Nothing s ->
+        pure (S.Inline s)
+
+field' :: HasSchemaRef f => Text -> f a -> Schema a
+field' name = field name id
+
+field :: HasSchemaRef f => Text -> (S.Schema -> S.Schema) -> f a -> Schema a
+field name f s = Schema $ do
```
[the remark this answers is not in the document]
pcapriotti: I added some docs and an example.

libs/wire-api/src/Data/Swagger/Typed.hs
pcapriotti: Sure, I added my explanation from this PR to the haddocks.

akshaymankar:
This looks neat, but I am having trouble with one type of usecase which I suspect is going to be very common in the federation work. Consider this type:

```haskell
data User = User { userQualifiedId :: Qualified Id, userName :: Text }
``` 

Here the `userQualifiedId`  is a replacement of `userId` and for backwards compatibility, we want to have the returned JSON to have both `id` and `qualified_id` fields. If I use the `Applicative` instance for `Schema`, this wouldn't work. Perhaps we can write a helper to attach such fields, or am I holding this completely wrong?

pcapriotti:
> ```haskell
> data User = User { userQualifiedId :: Qualified Id, userName :: Text }
> ```
> 
> Here the `userQualifiedId` is a replacement of `userId` and for backwards compatibility, we want to have the returned JSON to have both `id` and `qualified_id` fields. If I use the `Applicative` instance for `Schema`, this wouldn't work. Perhaps we can write a helper to attach such fields, or am I holding this completely wrong?

Ah, good question. I guess the idea is that the "id" field is just some redundant information, and it wouldn't be enough for parsing? Then, thinking ahead to when these descriptions can be used to parse JSON as well, one can do:

```haskell
instance ToTypedSchema User where
  toTypedSchema _= named "User" $ User
    <$> (  field' "qualified_id" schema
       <|> field' "id" (handle (pure (Left "qualified_id is required")))
        )
    <*> field "username" schema
```

where `handle` is supposed to return a parsing error when it encounters a `Left` value. For now it is the same as `coerce` though (i.e. it just changes the underlying type) because parsing isn't implemented.

pcapriotti:
> ```haskell
>        <|> field' "id" (handle (pure (Left "qualified_id is required")))
> ```

Sorry, this should be:
```haskell
       <|> field' "id" (handle (const (Left "qualified_id is required") <$> schema))
```
so that the "id" field is annotated with the correct type.

pcapriotti:
I have submitted an updated version of this pull request (#1474) so I'm closing this one.

PR #1464, opened 2021-04-26
Add endpoint to list clients which doesn't accept a JSON list in req

akshaymankar:
The new endpoint is at `POST /users/list-clients/v2`, this endpoint consumes a
JSON object with one key `qualified_users` whose value must be a list of
`Qualiified UserId` limited to MaxUsersForListClientsBulk.

### DIFF
--- a/services/brig/src/Brig/API/Public.hs
+++ b/services/brig/src/Brig/API/Public.hs
@@ -84,6 +84,7 @@ import Data.Swagger
     type_,
   )
 import qualified Data.Swagger.Build.Api as Doc
+import qualified Data.Text as T
 import qualified Data.Text as Text
 import qualified Data.Text.Ascii as Ascii
 import Data.Text.Encoding (decodeLatin1)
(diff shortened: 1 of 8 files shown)


pcapriotti (approved):
Looks good.

services/brig/src/Brig/API/Public.hs
```diff
@@ -328,13 +328,22 @@ type GetUserClientQualified =
     :> Get '[Servant.JSON] Public.PubClient
 
 type ListClientsBulk =
-  Summary "List all clients for a set of user ids"
+  Summary "List all clients for a set of user ids (deprecated, use /user/list-clients/v2)"
```
jschaul, line 331:
```suggestion
  Summary "List all clients for a set of user ids (deprecated, use /users/list-clients/v2)"
```

services/brig/src/Brig/API/Public.hs
```diff
@@ -328,13 +328,22 @@ type GetUserClientQualified =
...
     :> Post '[Servant.JSON] (Public.QualifiedUserMap (Set Public.PubClient))
 
+type ListClientsBulkV2 =
+  Summary "List all clients for a set of user ids"
+    :> ZAuthServant
+    :> "users"
+    :> "list-clients"
+    :> "v2"
+    :> Servant.ReqBody '[Servant.JSON] (Public.LimitedQualifiedUserIdList MaxUsersForListClientsBulk)
+    :> Post '[Servant.JSON] (Public.QualifiedUserMap (Set Public.PubClient))
```
jschaul, line 345: While we're at it: is giving back a map a great idea? Wouldn't we have more flexibility for the future if we give back an object containing the map here: `{"qualifiedUserMap" : { "domain" : [ ... ] }}`? 
(feel free to ignore me here and merge the PR as is. I just thought I'd bring that idea up)

jschaul (commented):
Needs an ormolu, and see inline comment/question. Otherwise fine to merge.

services/brig/src/Brig/API/Public.hs
akshaymankar: Hmm, I am going to implement the `Named` thing I did lazily, but in a slightly better way and in wire-api to achieve this.

akshaymankar: Done, LMK what you think.

jschaul (approved):

PR #1465, opened 2021-04-27
Add federator to galley

smatting:
This PR:
- generalizes `executeFederated` and makes it available to the `Galley` monad
- adds `API/Federation/API/Galley.hs` with stubs for api types


services/galley/src/Galley/Federation/Client.hs
```diff
@@ -0,0 +1,53 @@
...
+-- requests, but there are some issues in http2-client which might need some
+-- fixing first. More context here:
+-- https://github.com/lucasdicioccio/http2-client/issues/37
+-- https://github.com/lucasdicioccio/http2-client/issues/49
+mkFederatorClient :: FederationAppIO GrpcClient
+mkFederatorClient = do
+  federatorEndpoint <- view federator !? FederationNotConfigured
+  let cfg = grpcClientConfigSimple (T.unpack (federatorEndpoint ^. epHost)) (fromIntegral (federatorEndpoint ^. epPort)) False
+  createGrpcClient cfg
+    >>= either (throwE . FederationUnavailable . reason) pure
```
pcapriotti, line 46: Would it be possible to move these functions to some common place (maybe wire-api-federation) by making them polymorphic on the monad using mtl constraints?

smatting: Okay, did so in https://github.com/wireapp/wire-server/pull/1465/commits/ef036f1be72b2250fe7251e2b3f5bffd33f39bfd WDYT?

pcapriotti: Nice! Looks good from a quick look. I'll try to give a more in-depth review when this PR is ready.

jschaul (approved):

jschaul:
While this PR adds support for a federator client and creates a federation API stub that's not implemented yet, implementing the list-conversations endpoint stubbed here can be done in a subsequent PR.

PR #1471, opened 2021-04-29
Add tests for the federator client

akshaymankar:


libs/wire-api-federation/test/Test/Wire/API/Federation/ClientSpec.hs
```diff
@@ -0,0 +1,148 @@
...
+instance MonadIO m => MonadState RecievedRequests (MockT m) where
+  get = readIORef =<< ask
+  put x = do
+    ref <- ask
+    writeIORef ref x
+
+runMockT :: IORef RecievedRequests -> MockT m a -> m a
+runMockT ref mock = runReaderT (unMock mock) ref
+
+retryWhileN :: forall a m. (MonadIO m) => Int -> (a -> Bool) -> m a -> m a
```
jschaul, line 136: It sounds like retryWhileN should really move to a lower-level library and not being re-defined again, I think there's already a range of copies in the codebase:

```
ag "retryWhileN ::"
services/gundeck/test/integration/API.hs
936:retryWhileN :: (MonadIO m) => Int -> (a -> Bool) -> m a -> m a

services/brig/test/integration/Util.hs
750:retryWhileN :: (MonadIO m) => Int -> (a -> Bool) -> m a -> m a

services/galley/test/integration/API/Util.hs
1353:retryWhileN :: (MonadIO m) => Int -> (a -> Bool) -> m a -> m a

libs/cassandra-util/src/Cassandra/Schema.hs
251:retryWhileN :: (MonadIO m) => Int -> (a -> Bool) -> m a -> m a
```

libs/wire-api-federation/test/Test/Wire/API/Federation/ClientSpec.hs
```diff
@@ -0,0 +1,148 @@
...
+
+  it "should report federator failures correctly" $ do
+    handle <- generate arbitrary
+
+    (actualResponse, _) <-
+      withMockFederator (error "some IO error!") $
+        Brig.getUserByHandle Brig.clientRoutes handle
+
+    -- This behaviour is actually incorrect, the fix in upstream is not merged
+    -- yet: https://github.com/haskell-grpc-native/http2-grpc-haskell/pull/48
```
jschaul, line 63: Is it possible to depend on the fork with the fix rather than broken upstream?

jschaul (approved):

libs/wire-api-federation/test/Test/Wire/API/Federation/ClientSpec.hs
akshaymankar: Done!

libs/wire-api-federation/test/Test/Wire/API/Federation/ClientSpec.hs
akshaymankar: I couldn't figure out where it should go, then I thought this type of duplication in test code doesn't really hurt us, so I gave up :shrug:

PR #1474, opened 2021-04-29
RFC: Schemas for documented bidirectional JSON encoding

pcapriotti:
This is a new and improved version of the proposal in #1461, adding automatic `ToJSON` and `FromJSON` instance generation for types implementing `ToTypedSchema`. The API itself is pretty similar to the previous proposal, but there are some differences, mostly due to the presence of the serialiser component of the schema.

I've taken the basic idea from [this little-known haskell library](https://github.com/chpatrick/codec), but reworked things quite a bit and added the "documentation" part (i.e. Swagger schema generation).

I've written an [introductory tutorial](https://github.com/wireapp/wire-server/blob/pcapriotti/typed-schema-v2/docs/reference/schemas.md) that shows how to use the library in practice. More details on the inner workings and descriptions of the combinators can be found in the haddock, but that part is not really comprehensive yet.

### TODO

 - [x] Better / more consistent naming (e.g. TypedSchema -> Schema)
 - [x] Move into its own package
 - [x] Implement more examples to make sure the library is powerful enough
 - [x] Rename primed functions
 - [x] Generate JSON files using old instances (moved to #1486)
 - [x] Set up golden tests for new JSON instances (moved to #1486)


libs/wire-api/test/unit/Test/Data/Swagger/Typed.hs
```diff
@@ -0,0 +1,239 @@
...
+import Data.Aeson (FromJSON (..), Result (..), ToJSON (..), Value, fromJSON)
+import Data.Aeson.QQ
+import Data.Swagger.Typed
+import Imports
+import Test.Tasty
+import Test.Tasty.HUnit
+
+tests :: TestTree
+tests =
+  testGroup
```
jschaul, line 16: Is there an easy way to add a test which shows that a type with a field "name" with an instance of `ToTypedSchema` also has a swagger schema that contains "name"?

jschaul (approved):
Looks good to me! Feel free to also merge this PR when you're happy with it and add improvements in a later PR.

libs/wire-api/src/Data/Swagger/Typed.hs
```diff
@@ -0,0 +1,468 @@
+{-# LANGUAGE DeriveFunctor #-}
+{-# LANGUAGE DerivingVia #-}
+{-# LANGUAGE GeneralizedNewtypeDeriving #-}
+{-# LANGUAGE OverloadedLists #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+
+module Data.Swagger.Typed
```
jschaul, line 7: Perhaps you could add a module comment here which links to the tutorial docs?

libs/wire-api/test/unit/Test/Data/Swagger/Typed.hs
```diff
@@ -0,0 +1,239 @@
...
+      testBarAToJSON,
+      testBarAFromJSON,
+      testBarBToJSON,
+      testBarBFromJSON,
+      testAccessToJSON,
+      testAccessFromJSON,
+      testUser1ToJSON,
+      testUser1FromJSON,
+      testUser2ToJSON,
+      testUser2FromJSON
```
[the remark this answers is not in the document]
pcapriotti: The failure messages should be pretty much the same as the ones we were getting before with the hand-rolled instances. I've added a test to check that the error messages mention missing keys.

libs/wire-api/test/unit/Test/Data/Swagger/Typed.hs
```diff
@@ -0,0 +1,239 @@
...
+      (Success exampleUser1)
+      (fromJSON exampleUser1JSON)
+
+testUser2ToJSON :: TestTree
+testUser2ToJSON =
+  testCase "toJSON User" $
+    assertEqual
+      "User2: toJSON should match handwritten JSON"
+      exampleUser2JSON
+      (toJSON exampleUser2)
```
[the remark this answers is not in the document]
pcapriotti: I've changed the missing field tests to work with strings rather than JSON values.

libs/wire-api/src/Data/Swagger/Typed.hs
pcapriotti: Done.

libs/wire-api/test/unit/Test/Data/Swagger/Typed.hs
pcapriotti: Yes, I have added some tests for the swagger schema.

libs/schemas/src/Data/Schema.hs
```diff
@@ -0,0 +1,559 @@
...
+-- 'SchemaP d v w a b'.
+--
+-- The idea of using the profunctor structure of 'SchemaP' is taken
+-- from the [codec](https://github.com/chpatrick/codec) library.
+data SchemaP doc v w a b
+  = SchemaP
+      (SchemaDoc doc a b)
+      (SchemaIn v a b)
+      (SchemaOut w a b)
+  deriving (Functor)
```
akshaymankar, line 185: What does the `P` in `SchemaP` stand for?

libs/schemas/src/Data/Schema.hs
```diff
@@ -0,0 +1,559 @@
...
+
+-- | Like 'object', but apply an arbitrary function to the
+-- documentation of the resulting object.
+object' ::
+  HasObject doc doc' =>
+  Text ->
+  (doc' -> doc') ->
+  ObjectSchema doc a ->
+  ValueSchema doc' a
+object' name modify sch = over doc modify (object name sch)
```
akshaymankar, line 290: This prime makes this function harder to discover. But I can also not come up with nicer name, here are my attempts:
- `ObjectWithModifiedDoc`
- `overDocObject`

Another thing we can do is just expect users to use the lens, or does it make things more complicated?

libs/schemas/src/Data/Schema.hs
```diff
@@ -0,0 +1,559 @@
...
+-- | An optional schema with a specified failure value
+--
+-- This is a more general version of 'opt' that allows a custom
+-- serialisation 'Nothing' value.
+opt' :: w -> SchemaP d v w a b -> SchemaP d v w (Maybe a) (Maybe b)
+opt' w0 sch = SchemaP (SchemaDoc d) (SchemaIn i) (SchemaOut o)
+  where
+    d = schemaDoc sch
+    i = optional . schemaIn sch
+    o = maybe (pure w0) (schemaOut sch)
```
akshaymankar, line 360: Can we please rename this to `optWithDefault`?

libs/schemas/src/Data/Schema.hs
```diff
@@ -0,0 +1,559 @@
...
+text name =
+  named name $
+    mkSchema
+      (pure mempty)
+      (A.withText (T.unpack name) pure)
+      (pure . A.String)
+
+-- | A schema for a textual value with possible failure.
+text' :: Text -> (Text -> Either String a) -> SchemaP NamedSwaggerDoc A.Value A.Value Text a
+text' name parser = text name `withParser` (either fail pure . parser)
```
akshaymankar, line 381: Can we please rename this to `parsedText`?

akshaymankar (approved):
I have suggested couple of renames, feel free to do them in follow-up PRs as this PR blocks some other work.
Apart from that, looks very good. :shipit:

libs/schemas/src/Data/Schema.hs
pcapriotti: Profunctor

libs/schemas/src/Data/Schema.hs
pcapriotti: Using the lens directly is possible, but slightly akwkard, because the super-general lens types make type inference fail if you write it in the obvious way. The reason is that the `doc` lens can change the underlying `doc` type, whereas within `object'` the modifier function is constrained to be an endomorphism `doc' -> doc'`.

Besides renaming, another possible solution would be to make the general `doc` lens internal, rename it, and export a type-restricted `doc` lens. That hopefully should make type inference work in most cases.

If that doesn't work, I'll rename `object'` to one of your suggestions.

pcapriotti: I ended up renaming it to `objectWithDocModifier`. It's useful to have it as a convenience function instead of messing with lenses and having to add a bunch of extra parentheses. Same for `field'`.

akshaymankar: Sounds good. :+1:

pcapriotti:
I have squashed this pull request to a single commit adding the `schemas` library. The previous history is at [this branch](https://github.com/wireapp/wire-server/tree/pcapriotti/typed-schema-v2-port), which I'll soon turn into a new pull request porting some of the existing swagger/json instances to the new library.

akshaymankar:
Hmm, I thought `schemas` will collide with something in hackage and was going to propose a different name, but looks like `schemas` is pretty much the [same thing](https://hackage.haskell.org/package/schemas-0.4.0.2) except it has its own OpenApi2 types :sweat_smile:

jschaul:
CI is green now :)

smatting (approved):

PR #1486, opened 2021-05-04
Golden tests for JSON instances

pcapriotti:
This introduces some [golden tests](https://ro-che.info/articles/2017-12-04-golden-tests) for JSON instances. The way this works is that when the test does not find JSON files to compare against, it fails, but it also generates them. One is supposed to add the files to git, and rerun the tests (which will now mostly likely pass). If in the future the JSON instance changes, the test will hopefully catch it.

The way the tests are generated is a bit hacky and hard to reproduce, but hopefully it shouldn't be done too often. Basically, the wire-server code is patched to add some ad-hoc `Show` instances. Then the `Arbitrary` instances are used to generate random objects, which are then printed and end up in the `Golden.Generated` module as Haskell code. Some of this code is fixed by a sed script after the fact.

So to regenerate `Golden.Generated` one can do the following:

 - apply the patch(es) in `libs/wire-api/test/golden/` with `git am`
 - from libs/wire-api, run the `gentests.sh` script
 - undo the patch (but keep the generated code)

## TODO

- [x] Prune unused imports from generated files
- [x] Group tests by type


akshaymankar (commented):
Looks good, but let's add license headers to all the generated haskell.

libs/wire-api/test/golden/gentests.sh
```diff
@@ -156,6 +156,24 @@ stack build --fast --test --bench --no-run-benchmarks wire-api |
...
+            echo "-- later version."
+            echo "--"
+            echo "-- This program is distributed in the hope that it will be useful, but WITHOUT"
+            echo "-- ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS"
+            echo "-- FOR A PARTICULAR PURPOSE. See the GNU Affero General Public License for more"
+            echo "-- details."
+            echo "--"
+            echo "-- You should have received a copy of the GNU Affero General Public License along"
+            echo "-- with this program. If not, see <https://www.gnu.org/licenses/>."
+            echo ""
```
akshaymankar, line 176: Why not just run `headroom` like ormolu?

akshaymankar (approved):

PR #1487, opened 2021-05-05
Brig: Serve unified swagger of brig and galley

akshaymankar:
Doesn't work yet, because changes are pending in galley.

### DIFF
--- a/services/brig/src/Brig/API/Public.hs
+++ b/services/brig/src/Brig/API/Public.hs
@@ -41,6 +41,7 @@ import qualified Brig.API.Util as API
 import Brig.App
 import qualified Brig.Calling.API as Calling
 import qualified Brig.Data.User as Data
+import qualified Brig.IO.Intra as Intra
 import Brig.Options hiding (internalEvents, sesQueue)
 import qualified Brig.Provider.API as Provider
 import qualified Brig.Team.API as Team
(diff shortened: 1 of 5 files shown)


jschaul:
Idea: Could the same be done for spar, too? Spar already gives back swagger 2.0.

akshaymankar:
> Idea: Could the same be done for spar, too? Spar already gives back swagger 2.0.

Oh yeah, I forgot about spar. I will add it.

akshaymankar:
Hmm, looks like brig never calls spar. This make it more complicated :(

jschaul:
> Hmm, looks like brig never calls spar. This make it more complicated :(

Ah, really?

Hm, but spar calls both brig and galley. So maybe spar could gather the swaggers for the rest. But spar won't call gundeck and cargohold, so we'll anyway have to introduce a new service dependency for that.

The other option could be to write the swagger json(s) to a file when compiling, and move it into a file e.g. brig can read from disk in its container. Then no new http calls need to be made.

akshaymankar:
> The other option could be to write the swagger json(s) to a file when compiling, and move it into a file e.g. brig can read from disk in its container. Then no new http calls need to be made.

If we move the API types to `wire-api`, brig can just have the swagger, we won't have to do any changes to our build system

jschaul:
> > The other option could be to write the swagger json(s) to a file when compiling, and move it into a file e.g. brig can read from disk in its container. Then no new http calls need to be made.
> 
> If we move the API types to `wire-api`, brig can just have the swagger, we won't have to do any changes to our build system

Good point!

akshaymankar:
Superseded by #1508

PR #1488, opened 2021-05-05
wire-api-federation: Mark flaky tests as pending

akshaymankar:


fisx (approved):
Didn't know about fdescribe?

We should have an integration test hook (an hlint rule?) that makes sure we don't have any of those!

akshaymankar:
> Didn't know about fdescribe?
> 
> We should have an integration test hook (an hlint rule?) that makes sure we don't have any of those!

I think we can add an `.hspec` file at the root to make the tests fail if it encounters a focused test.

PR #1489, opened 2021-05-05
Fail on focussed

fisx:
https://github.com/wireapp/wire-server/pull/1488#issuecomment-832654944

### DIFF
--- a/.hspec
+++ b/.hspec
@@ -1,0 +1,2 @@
+--fail-on-focused
+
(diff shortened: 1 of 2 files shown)


fisx:
The directory I've picked is just a shot in the dark.  Needs to be `~` or `.` when the CI pipeline is alive.

fisx:
(Locally I confirmed that stack runs test with package root as wd, so we *could* just add a million hspecs all over the place.)

fisx:
> (Locally I confirmed that stack runs test with package root as wd, so we _could_ just add a million `.hspec` files all over the place.)

fisx:
...  or something like `echo --fail-on-focused >> ~/.hspec` somewhere where the CI pipeline will pick it up?  (where would that be?)

akshaymankar:
I implemeted this in #1491, maybe this can be closed?

fisx:
solved here: https://github.com/wireapp/wire-server/pull/1491

PR #1491, opened 2021-05-05
wire-api-federator: Make client tests reliable

akshaymankar:
Important changes:
* Run the server once for all the tests
* Use Warp.openFreePort to open a free port
* Use Warp.setGracefulCloseTimeout2 to 0 so stopping the server is fast
* Use Warp.setBeforeMainLoop and MVar to ensure tests are only run after the
server has started
* Error if any test is focused


libs/wire-api-federation/test/Test/Wire/API/Federation/ClientSpec.hs
```diff
@@ -47,128 +45,142 @@ import Wire.API.Federation.GRPC.Types (Component (Brig), FederatedRequest (Feder
...
-        Brig.getUserByHandle Brig.clientRoutes handle
-
-    actualResponse `shouldBe` Left (FederationClientRPCError "grpc error: GRPC status indicates failure: status-code=NOT_FOUND, status-message=\"Just testing\"")
+spec = do
+  stateRef <- runIO . newIORef $ MockState [] (error "No mock response provided") (error "server not started") (error "No port selected yet")
+  beforeAll (startMockFederator stateRef) $
+    afterAll_ (stopMockFederator stateRef) $
+      before_ (flushState stateRef) $
+        describe "Federator.Client" $ do
+          -- startMockFederator
```
akshaymankar, line 54:
```suggestion
```

libs/wire-api-federation/.hspec
```diff
@@ -0,0 +1 @@
+--fail-on-focused
```
fisx, line 1: Isn't there a better solution that applies to all packages?  I'm trying in https://github.com/wireapp/wire-server/pull/1489, but no definite idea yet.

anyway it's fine to do it here now, better than nothing.  (unless somebody will expect it to be universally enabled and rely on it.)

libs/wire-api-federation/test/Test/Wire/API/Federation/ClientSpec.hs
```diff
@@ -47,128 +45,141 @@ import Wire.API.Federation.GRPC.Types (Component (Brig), FederatedRequest (Feder
...
-spec = xdescribe "Federator.Client" $ do
-  it "should make correct calls to the federator and parse success response correctly" $ do
-    handle <- generate arbitrary
-    expectedResponse :: Maybe UserProfile <- generate arbitrary
+spec = do
+  stateRef <- runIO . newIORef $ MockState [] (error "No mock response provided") (error "server not started") (error "No port selected yet")
+  beforeAll (startMockFederator stateRef) $
+    afterAll_ (stopMockFederator stateRef) $
+      before_ (flushState stateRef) $
+        describe "Federator.Client" $ do
```
jschaul, line 53: Minor nitpick: Would there be a way to do these nice things in a way that doesn't lead to so much nesting/indentation for all the individual tests?

libs/wire-api-federation/.hspec
akshaymankar: There are four ways:
- Specify it in ~/.hspec: Obviously doesn't work for us
- Specify it in HSPEC_OPTIONS env variable: This variable gets lost when stack runs with nix, but could work for CI
- Specify it in command line args: This doesn't work because not all of our tests are hspec, so passing this in `--test-arguments` for stack would cause other tests to fail.
- Specify it in .hspec at the root of each package: This is what I ended up doing.

libs/wire-api-federation/test/Test/Wire/API/Federation/ClientSpec.hs
akshaymankar: Hmm, maybe using `.` to compose these instead of `$` would make ormolu format them nicely, something like this:

```haskell
  beforeAll (startMockFederator stateRef)
    . afterAll_ (stopMockFederator stateRef)
    . before_ (flushState stateRef)
    $ describe "Federator.Client" $ do
      it "should make correct calls to the federator and parse success response correctly" $ do
```

I don't think there is a way to just have these hooks setup separate from the test blocks (and not wrapping them) like JUnit, RSpec etc.

libs/wire-api-federation/.hspec
akshaymankar: Hmm, I set it in `stack-deps.nix` and it worked! I have now set it in Makefile too and focused a test, if it works, we can use this. Unfortunately, those who don't use nix or make may commit a focused the test, but the CI should catch it.

akshaymankar: [It worked](https://bridge-ie-concourse.zinfra.io/teams/main/pipelines/wire-server-pr/jobs/wire-server-compile-alpine/builds/3058#L606eb6ee:2368:2369)!!

```
#9 268.5 wire-api-federation              > spec: unexpected argument `"--fail-on-focused"' from environment variable HSPEC_OPTIONS
#9 268.5 wire-api-federation              > Test suite spec failed
```

akshaymankar: Hmm, I should read things better!

akshaymankar: The ubuntu job won't fail because it doesn't use the Makefile :(

akshaymankar: [Much better](https://bridge-ie-concourse.zinfra.io/teams/main/pipelines/wire-server-pr/jobs/wire-server-compile-alpine/builds/3061#L606ebea7:2548:2563)

```
#7 294.3 wire-api-federation              > Test.Wire.API.Federation.Client
#7 294.3 wire-api-federation              >   Federator.Client
#7 294.3 wire-api-federation              >     should make correct calls to the federator and parse success response correctly FAILED [1]
#7 294.3 wire-api-federation              > 
#7 294.3 wire-api-federation              > Failures:
#7 294.3 wire-api-federation              > 
#7 294.3 wire-api-federation              >   test/Test/Wire/API/Federation/ClientSpec.hs:54:7: 
#7 294.3 wire-api-federation              >   1) Test.Wire.API.Federation.Client.Federator.Client should make correct calls to the federator and parse success response correctly
#7 294.3 wire-api-federation              >        item is focused; failing due to --fail-on-focused
#7 294.3 wire-api-federation              > 
#7 294.3 wire-api-federation              >   To rerun use: --match "/Test.Wire.API.Federation.Client/Federator.Client/should make correct calls to the federator and parse success response correctly/"
#7 294.3 wire-api-federation              > 
#7 294.3 wire-api-federation              > Randomized with seed 819810423
#7 294.3 wire-api-federation              > 
#7 294.3 wire-api-federation              > Finished in 0.1371 seconds
#7 294.3 wire-api-federation              > 1 example, 1 failure
```

libs/wire-api-federation/test/Test/Wire/API/Federation/ClientSpec.hs
akshaymankar: Done here:  bb2b627

akshaymankar:
@fisx I am going to revert your commit. Let's tackle focussed hspecs in integration tests separately. Can please I get a review of the main work here?

Makefile
```diff
@@ -18,6 +18,7 @@ CHARTS_RELEASE        := wire-server databases-ephemeral fake-aws aws-ingress ba
 BUILDAH_PUSH          ?= 0
 KIND_CLUSTER_NAME     := wire-server
 BUILDAH_KIND_LOAD     ?= 1
+export HSPEC_OPTIONS = --fail-on-focused
```
jschaul, line 21: A comment would be nice, on what this does and how this interacts with unit or integration tests, and why this gets set globally in the top-level makefile but not inside e.g. the services/brig/Makefile. (what are the implications of that for people using one or the other makefile to run tests?)

Also, this focussed override in the makefile and the nix dependencies looks like it could live in a separate PR to the wire-api-federator mock changes.

jschaul (approved):
Mock server startup logic looks good!

Makefile
akshaymankar: > A comment would be nice, on what this does and how this interacts with unit or integration tests, and why this gets set globally in the top-level makefile but not inside e.g. the services/brig/Makefile. (what are the implications of that for people using one or the other makefile to run tests?)

I will add a comment.

> Also, this focussed override in the makefile and the nix dependencies looks like it could live in a separate PR to the wire-api-federator mock changes.

Hmm, I was in the mindset of fixing all the problems addition of this test had created. Creating many small PRs is a little annoying, but I guess that is our process, so in hind sight I agree with you but maybe we should figure out a less annoying process.

PR #1492, opened 2021-05-06
Fix Arbitrary instances and enable corresponding roundtrip tests

pcapriotti:
This changes `Arbitrary` instances for types that had flaky JSON roundtrip tests. Here is the list of changes:
 - `Milliseconds` is a 64 bit integer which is converted to and from `Double` when going through JSON encoding. Its arbitrary instance now limits it to values that JSON can represent exactly (up to 2^53).
 - `Asset`, `ResumableAsset`, `Event` and `ClientMismatch` truncate the times they contain to millisecond precision. The `Arbitrary` instance now reflects that.
 - The "mute status" field of `MemberUpdate` never gets serialised (why?). Therefore, the arbitrary instance always sets it to `Nothing`.
 - `OtherMemberUpdate` cannot be `Nothing`.
 - `TeamUpdateData` cannot have all its fields set to `Nothing`.

### DIFF
--- a/libs/types-common/src/Data/Misc.hs
+++ b/libs/types-common/src/Data/Misc.hs
@@ -76,7 +76,7 @@ import qualified Data.Swagger.Build.Api as Doc
 import qualified Data.Text as Text
 import Data.Text.Encoding (decodeUtf8, encodeUtf8)
 import Imports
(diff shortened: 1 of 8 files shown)


akshaymankar (approved):
Looks good! Can you please follow up with the clients about the mute status? Looks like a bug.

PR #1493, opened 2021-05-06
Federation: Add RPC getConversations

smatting:
This PR:
- adds `getConversations` federation client RPC to galley
- adds a stub `updateConversationMembership` federation client RPC to galley


libs/wire-api-federation/src/Wire/API/Federation/API/Galley.hs
```diff
@@ -19,40 +19,63 @@
...
         :> "conversations"
-        -- usecases:
-        -- - e.g. upon registering a new client to your account, get the list of your conversations
-        :> "list-conversations"
-        :> ReqBody '[JSON] ListConversations
-        :> Post '[JSON] ListConversationsResponse,
-    conversationMemberChange ::
+        :> "get-by-ids"
+        :> ReqBody '[JSON] GetConversationsRequest
+        :> Post '[JSON] [Conversation],
```
jschaul, line 43: Could we wrap the plain list `[Conversation]` here in a json object (e.g. `data SomeResponse = { conversations :: [Conversation] }` for better futureproofing?

libs/wire-api-federation/src/Wire/API/Federation/API/Galley.hs
```diff
@@ -19,40 +19,63 @@
...
+instance ToJSON MembershipUpdateOp
 
-type ListConversations = ()
+instance FromJSON MembershipUpdateOp
 
-type ListConversationsResponse = ()
+data ConversationMemberUpdate = ConversationMemberUpdate
+  { cmuConvId :: Qualified ConvId,
+    cmuUserId :: UserId,
+    cmuOperation :: MembershipUpdateOp
```
jschaul, line 77: We need support for adding multiple users in the same request. Perhaps we could have a list of users to remove and a list of users to add?

libs/wire-api-federation/src/Wire/API/Federation/API/Galley.hs
```diff
@@ -19,40 +19,63 @@
...
-        :> ReqBody '[JSON] ConversationMemberChange
-        :> Post '[JSON] ConversationMemberChangeResponse
+        :> "update-membership"
+        :> ReqBody '[JSON] ConversationMemberUpdate
+        :> Post '[JSON] ()
+  }
+  deriving (Generic)
+
+data GetConversationsRequest = GetConversationsRequest
+  { gcrUserId :: UserId,
```
jschaul, line 57: This needs to be a qualified UserId, as a non-qualified UserId doesn't have a meaning on the other backend.

```suggestion
  { gcrUserId :: Qualified UserId,
```

services/galley/test/integration/API/Federation.hs
```diff
@@ -0,0 +1,75 @@
...
+  carl <- randomUser
+  connectUsers alice (singleton carl)
+  cnv2 <- responseJsonUnsafeWithMsg "conversation" <$> postConv alice [bob, carl] (Just "gossip2") [] Nothing Nothing
+  getConvs alice (Just $ Left [cnvId cnv2]) Nothing !!! do
+    const 200 === statusCode
+    const (Just [cnvId cnv2]) === fmap (map cnvId . convList) . responseJsonUnsafe
+  -- get both
+
+  fedGalleyClient <- view tsFedGalleyClient
+  cs <- Just <$> FedGalley.getConversations fedGalleyClient (GetConversationsRequest alice [cnvId cnv1, cnvId cnv2])
```
jschaul, line 60: If you replace `alice` in this line with another random user not part of that conversation, I suppose this test will fail, right? getConversations should then return a non-200 status code or some other embedded error? Could you add a test for the failure case of the requesting user not being part of the conversation?

I think we need to add some FUTUREWORK here to replace alice with a remote user once we have support for that.

jschaul (approved):
Overall looks good! See nitpicks inline :)

jschaul:
There's an import in `libs/wire-api-federation/test/Test/Wire/API/Federation/APISpec.hs` to a module deleted in this branch tripping up compilation.

libs/wire-api-federation/src/Wire/API/Federation/API/Galley.hs
smatting: Oh, I missed that. I updated the type and created a stub for the non-local case which can't handle atm

services/galley/test/integration/API/Federation.hs
smatting: Good point. I changed the behaviour and added a test: If the user is not part of the conversation it is not included in the result. Rationale: We want to handle this case, because it might result frequent when two backends are not fully synced. This requires lifiting the problem to the value level. Excluding the conversation from the result is also consistent with the case when a conversation id cannot be found.

PR #1496, opened 2021-05-10
Fix: NewTeamMember vs. UserLegalHoldStatus

fisx:
This fixes a bug that has not been in production only because the LH feature has never been officially released, nor turned on anywhere we know of.

If you have turned it on on your own instance, this affects you iff you are worried about custom / rogue clients, ie. software that uses the REST API in ways not anticipated.

If you are still worried: the bug that was fixed here allowed specially crafted clients to set an invalid legalhold state for the attacked user.  This may lead to broken UX, or the appearance that LH has been enabled even though there is no LH device present for that user.


libs/wire-api/src/Wire/API/Team/Member.hs
```diff
@@ -258,11 +259,18 @@ modelNewTeamMember = Doc.defineModel "NewTeamMember" $ do
...
+      mem = Object . HM.fromList . fltr . HM.toList $ o
+      Object o = toJSON . _ntmNewTeamMember $ t
+      fltr = filter ((`elem` ["user", "permissions", "created_by", "created_at"]) . fst)
 
 instance FromJSON NewTeamMember where
-  parseJSON = withObject "add team member" $ \o ->
-    NewTeamMember <$> o .: "member"
+  parseJSON = withObject "add team member" $ \o -> do
+    mem <- o .: "member"
+    if (_legalHoldStatus mem == defUserLegalHoldStatus)
```
jschaul, line 271: I can't find where `defUserLegalHoldStatus` is defined.

In general I wonder: wouldn't it be cleaner to change NewTeamMember to take a different type to `TeamMember` (a new one without legalhold_status), which would then make adding if statements and filtering not necessary. Should be more readable also.
(In case you'd like to stick to filtering and conditionals - could you add some comments explaining why that's done for the the future reader?)

PR #1497, opened 2021-05-11
schema-profunctor: Add ToSchema instance for lists

akshaymankar:


libs/schema-profunctor/src/Data/Schema.hs
```diff
@@ -563,6 +563,8 @@ instance ToSchema String where schema = genericToSchema
 
 instance ToSchema Bool where schema = genericToSchema
 
+instance (S.ToSchema a, A.ToJSON a, A.FromJSON a)=> ToSchema [a] where schema = genericToSchema
```
pcapriotti, line 566: I would prefer this:

```suggestion
instance ToSchema a => ToSchema [a] where schema = array schema
```

akshaymankar: Hmm, is there a difference. This does look better, asking for my education.
I also pushed another instance for `NonEmpty a` I am guessing `array` wouldn't produce a nice ToSchema instance for that, so we should use `genericToSchema` in that case?

libs/schema-profunctor/src/Data/Schema.hs
```diff
@@ -563,6 +564,10 @@ instance ToSchema String where schema = genericToSchema
 
 instance ToSchema Bool where schema = genericToSchema
 
+instance (S.ToSchema a, A.ToJSON a, A.FromJSON a) => ToSchema [a] where schema = genericToSchema
+
+instance (S.ToSchema a, A.ToJSON a, A.FromJSON a) => ToSchema (NonEmpty a) where schema = genericToSchema
```
pcapriotti, line 569: I would also like to define this in terms of `ToSchema a` rather than `S.ToSchema a`. It should be possible to use `array`, and then `withParser` to convert it into a non-empty list.

libs/schema-profunctor/src/Data/Schema.hs
akshaymankar: Hmm, `array schema` doesn't work as it doesn't produce a named schema, so either I need a `TypeRep`, which would be annoying or I need to set it to something like an empty string.

pcapriotti: Ah, you're right, on both counts.

I think it's better not to have this instance. One can just use `array` when you need an array, since lists are indeed most commonly unnamed.

Similarly, I would add a `nonEmptyArray` combinator instead of the `NonEmpty` instance.

libs/schema-profunctor/src/Data/Schema.hs
```diff
@@ -316,6 +318,19 @@ array sch = SchemaP (SchemaDoc s) (SchemaIn r) (SchemaOut w)
...
+  ValueSchema doc (NonEmpty a)
+nonEmptyArray sch = SchemaP (SchemaDoc s) (SchemaIn r) (SchemaOut w)
+  where
+    name = maybe "non-empty array" ("non-empty array of " <>) (getName (schemaDoc sch))
+    r = A.withArray (T.unpack name) $ \arr -> case V.toList arr of
+      [] -> A.parseFail "Unexpected empty array found while parsing a NonEmpty"
+      (x : xs) -> mapM (schemaIn sch) (x :| xs)
+    s = mkArray (schemaDoc sch) & S.minItems ?~ 1
+    w xs = A.Array . V.fromList <$> mapM (schemaOut sch) (NonEmpty.toList xs)
+
```
pcapriotti, line 333: It should be possible to use `array` instead of reimplementing its functionality. Something like:
```haskell
nonEmptyArray sch = NonEmpty.toList .= array sch `withParser` check
  where
    check = maybe (fail "Unexpected empty array found while parsing a NonEmpty") pure
          . NonEmpty.nonEmpty
```

libs/schema-profunctor/src/Data/Schema.hs
```diff
@@ -316,6 +318,19 @@ array sch = SchemaP (SchemaDoc s) (SchemaIn r) (SchemaOut w)
     s = mkArray (schemaDoc sch)
     w x = A.Array . V.fromList <$> mapM (schemaOut sch) x
 
+nonEmptyArray ::
+  forall ndoc doc a.(HasArray ndoc doc, HasName ndoc, S.HasMinItems doc (Maybe Integer)) =>
```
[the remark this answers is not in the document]
akshaymankar: I first added it to `HasArray`, but it didn't look very nice as the lens doesn't depend on the `ndoc` variable and it has to be type applied. I will defined a similar class, that should also work.

libs/schema-profunctor/src/Data/Schema.hs
akshaymankar: I couldn't make the swagger work nicely, thought it was ok to have a small amount of duplication. Maybe we can look at it together after lunch to get it to work?

pcapriotti (approved):
Apart from the possible simplification of the `minItems` lens, looks good to me.

PR #1500, opened 2021-05-11
Federation: Add getUserClients RPC to brig

smatting:
This PR
- adds the `getUserClients` RPC to brig's federation API.
- implements the stubs in 3 public endpoints that retrieve `PubClient` values for remote users


jschaul (commented):
Question: Do you intend to hook this call into the existing client-server API of `/users/{domain}/{uid}/clients` in this or an upcoming PR?

Which information exactly do clients need about other user's clients? The `PubClient` type and the `toPubClient` function are in use for current client-server APIs returning clients. The `UserClients` type seems to so far only be used for internal endpoints and tests (or did I see this wrong? I only did a brief grepping). For message sending and the backend in galley, perhaps we only need the IDs. But other user's clients probably want to also know their device class for displaying in the UI perhaps. Perhaps giving back a map of UserId to PubClient might make sense here?

smatting:
> Question: Do you intend to hook this call into the existing client-server API of `/users/{domain}/{uid}/clients` in this or an upcoming PR?
I hooked it into the the client endpoints. However end2end tests are not included. Should we add them?

> Which information exactly do clients need about other user's clients? ...
I changed the response type to `PubClient`.

jschaul (approved):
Adding one end2end test for this endpoint (`/users/{domain}/{uid}/clients`) might be nice, otherwise this PR looks good I think.

pcapriotti:
Thanks, I was indeed writing an End2End test.

jschaul (approved):

PR #1503, opened 2021-05-12
Client-supported features end-point

fisx:
https://wearezeta.atlassian.net/browse/SQSERVICES-417

changelog update (and possibly a few lines of documentation) coming up.

TODO:

- [x] change `legalhold` keyword to `legalhold-implicit-consent`
- [x] rename `feature` to `capability`


services/brig/src/Brig/API/Client.hs
```diff
@@ -131,6 +137,11 @@ updateClient u c r = do
   unless ok $
     throwE ClientNotFound
   for_ (updateClientLabel r) $ lift . Data.updateClientLabel u c . Just
+  for_ (updateClientSupportedFeatures r) $ \features' -> do
+    features <- fromMaybe [] <$> Data.lookupClientSupportedFeatures u c
+    if Set.fromList features `Set.isSubsetOf` Set.fromList features'
+      then lift . Data.updateClientSupportedFeatures u c . Just $ features'
+      else throwE ClientFeaturesCannotBeRemoved
```
jschaul, line 144: This doesn't make sense. Of course features can be removed, or changed, e.g. support for implicit consent could be removed, and instead support for explicit consent added. Why force clients in 3 years to support half-implemented features from the distant past?

services/brig/src/Brig/API/Public.hs
```diff
@@ -818,6 +818,17 @@ sitemap o = do
     Doc.returns (Doc.ref Public.modelClient)
     Doc.response 200 "Client" Doc.end
 
+  get "/clients/:client/supported-features" (continue getClientSupportedFeaturesH) $
```
jschaul, line 821: Is this endpoint passed through in the [nginz config](https://github.com/zinfra/backend-wiki/wiki/Checklists#adding-new-endpoints-in-wire-server) ?

services/brig/src/Brig/Data/Client.hs
```diff
@@ -233,6 +244,9 @@ insertClient = "INSERT INTO clients (user, client, tstamp, type, label, class, c
 updateClientLabelQuery :: PrepQuery W (Maybe Text, UserId, ClientId) ()
 updateClientLabelQuery = "UPDATE clients SET label = ? WHERE user = ? AND client = ?"
 
+updateClientSupportedFeaturesQuery :: PrepQuery W (Maybe [SupportedClientFeature], UserId, ClientId) ()
```
jschaul, line 247: This should be a cassandra set (`C.Set` from  `ìmport Cassandra as C`) and not a Maybe List. See e.g. https://github.com/wireapp/wire-server/blob/develop/services/galley/src/Galley/Data/Queries.hs#L206-L207

services/brig/test/integration/API/User/Client.hs
```diff
@@ -518,11 +529,38 @@ testUpdateClient opts brig = do
...
+              const (Just "client-feature-cannot-be-removed") === fmap Error.label . responseJsonMaybe
+
+        getClientSupportedFeatures brig uid (clientId c) !!! do
+          const 200 === statusCode
+          const (Just (SupportedClientFeatureList featuresOut)) === responseJsonMaybe
+
+  checkUpdate (Just [ClientSupportsLegalholdImplicitConsent]) True [ClientSupportsLegalholdImplicitConsent]
+  checkUpdate Nothing True [ClientSupportsLegalholdImplicitConsent]
+  checkUpdate (Just []) False [ClientSupportsLegalholdImplicitConsent]
+
```
jschaul, line 563: Can you test that updating the client with only the supportedClientFeature doesn't break the rest (i.e. other parts of the client entry are not erased) ?

docs/reference/cassandra-schema.cql
```diff
@@ -1158,6 +1158,7 @@ CREATE TABLE brig_test.clients (
     lat double,
     lon double,
     model text,
+    supported_features list<int>,
```
jschaul, line 1161: This should be a set, see comment further down.

jschaul (approved):
See inline comments for things to improve; the functionality of storing client input is implemented alright. 
However as I don't really see what the plan is to use this information afterwards, this is only a tentative approval of this particular storing functionality, not of a first part of a general plan. i.e. this code may get reverted entirely in the future if the plan how to make use of these features doesn't materialize.
For upcoming PRs please get reviews on the implementation ideas first before writing a lot of code.

libs/wire-api/src/Wire/API/User/Client.hs
```diff
@@ -555,6 +565,54 @@ instance FromJSON UpdateClient where
       <$> o .:? "prekeys" .!= []
       <*> o .:? "lastkey"
       <*> o .:? "label"
+      <*> o .:? "supported-features"
+
+-- FUTUREWORK: add golden tests for this?
+data SupportedClientFeatureList = SupportedClientFeatureList {fromSupportedClientFeatureList :: [SupportedClientFeature]}
+  deriving stock (Eq, Ord, Show, Generic)
+  deriving (Arbitrary) via (GenericUniform SupportedClientFeatureList)
+  deriving (ToSchema) via (CustomSwagger '[FieldLabelModifier (StripPrefix "fromSupportedClient", CamelToSnake)] SupportedClientFeatureList)
```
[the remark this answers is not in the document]
fisx: proved to be non-trivial, but i'll figure it out here, later: https://github.com/wireapp/wire-server/pull/1512

services/brig/src/Brig/API/Client.hs
fisx: It does make sense if we see this as a data-minimizing substitute for version information.  Versions usually aren't decreasing either, and I think it's reasonable to disallow it.

services/brig/src/Brig/API/Public.hs
fisx: Good point, thanks!  I'll add that.

services/brig/test/integration/API/User/Client.hs
fisx: b9469683055876b3927c255cd1098a06202ebd8b

services/brig/schema/src/V64_ClientFeatures.hs
```diff
@@ -0,0 +1,32 @@
...
+  )
+where
+
+import Cassandra.Schema
+import Imports
+import Text.RawString.QQ
+
+migration :: Migration
+migration = Migration 64 "Add supported client features to clients table" $ do
+  schema' [r| alter columnfamily clients add supported_features list<int>; |]
```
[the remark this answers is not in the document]
fisx: 10f1da8109780c034b41a7345f3fc380b4ffb8b9

services/brig/src/Brig/API/Client.hs
fisx: Another thing that can be said about in favor of disallowing it is that the restriction can be trivially lifted whenever we change our mind.

services/brig/src/Brig/API/Public.hs
fisx: actually, it looks like we're cool:

```yaml
    - path: /clients
      envs:
      - all
```

services/brig/src/Brig/API/Client.hs
jschaul: Okay, that's true, this check can be removed again.

services/brig/src/Brig/API/Public.hs
jschaul: Perfect!

jschaul (approved):

PR #1505, opened 2021-05-13
Make DerivingVia a package default.

fisx:
This has been bugging me for a while.  I'll just add this minimal commit so we won't have to mention `DerivingVia` in the package heades any more in the future, but I'll leave the old ones in there for the sake of laziness.

### DIFF
--- a/libs/brig-types/src/Brig/Types/User/EJPD.hs
+++ b/libs/brig-types/src/Brig/Types/User/EJPD.hs
@@ -1,4 +1,3 @@
-
 -- This file is part of the Wire Server implementation.
 --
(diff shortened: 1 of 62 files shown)


pcapriotti:
Yes, good idea. It would be nice to remove the corresponding pragmas as well. This should do it:
```
fd -ehs --exec sed -i '/LANGUAGE DerivingVia/d'
```

fisx:
fd looks cool, thanks for the pointer!  (package `fd-find` in debian)

(for the record: i deliberately decided against removing it to make this as simple and swift as possible and then move back to other business.  but entering one more command wasn't too bad either.  :))

pcapriotti (approved):
Looks good!

PR #1508, opened 2021-05-14
Unify Swagger 2.0 docs for brig, galley and spar

smatting:
**Note**: this supersedes #1487.

This is a refactoring PR.

- Moves all Servant API types to `wire-api` under `Wire.API.Public`.
- Serves all swagger 2.0 docs from `api/swagger.json` (from brig).
- Removes `sso/api-docs`.
 - `wire-api` has gained many more dependencies, including servant, scim and saml packages.
 - The two basic types modules in `spar` have been moved to `Wire.API.Scim` and `Wire.API.Spar`. Orphan swagger instances are also contained there.
 - The `MakeCustomError` instance for `IdPMetadataInfo` has been orphaned, because it cannot reside in `wire-api`, and it is currently in `Spar.Error`.
 - `spar` endpoints are now using the same servant implementation of `ZAuth` as `brig` and `galley`, ~~so now the `Z-User` header is no longer optional~~.

## TODOs

- [x] Move `GalleyRoutesPublic` and expose swagger (https://wearezeta.atlassian.net/browse/SQSERVICES-371)
- [x] mirror changes to zinfra/zwagger-ui (https://github.com/zinfra/zwagger-ui/pull/4)
- [x] un-orphan swagger instances when possible
- [x] rearrange new modules in `wire-api`


fisx:
Good idea, looking forward to see this land!  I haven't looked at the code yet, but a few remarks on the description:

- orphan swagger instances shouldn't be necessary, just move them to the modules where the types are introduced.
- how do you feel about renaming `Wire.API.Public` to `Wire.API.Routes.Public`?
- the original idea of wire-api is to structure the module hierarchy to accomodate and categorize the data types needed in the, well wire api.  `Wire.API.Scim` and `Wire.API.Spar` don't do that very well.  How about introducing `Wire.API.User.Scim`, `Wire.API.User.SAML`, and moving what doesn't fit into either to `Wire.API.User`, or some new module that makes most sense?  Like, `Wire.API.User.Provisioning`?

pcapriotti:
>     * orphan swagger instances shouldn't be necessary, just move them to the modules where the types are introduced.

Some of them are for types defined in other packages, and they were already orphan before this. I'll try to un-orphan (adopt?) the other ones.

>     * how do you feel about renaming `Wire.API.Public` to `Wire.API.Routes.Public`?

Will do.

>     * the original idea of wire-api is to structure the module hierarchy to accomodate and categorize the data types needed in the, well wire api.  `Wire.API.Scim` and `Wire.API.Spar` don't do that very well.  How about introducing `Wire.API.User.Scim`, `Wire.API.User.SAML`, and moving what doesn't fit into either to `Wire.API.User`, or some new module that makes most sense?  Like, `Wire.API.User.Provisioning`?

Yes, I thought of something like this as well. I'll try to move things around like you suggested.

fisx:
> Some of them are for types defined in other packages, and they were already orphan before this. I'll try to un-orphan (adopt?) the other ones.

One difference is that before, the orphans where in the executable (right?  I mean, in the spar library which is morally an executable), which makes orphanhood less of an issue than in a library.  But even `wire-api` is different from something like `bytestring` or `aeson`, so use your best judgement if something is awkward to un-orphan.

jschaul (approved):

PR #1511, opened 2021-05-17
Improve naming conventions federation RPC calls

pcapriotti:


libs/wire-api-federation/src/Wire/API/Federation/API/Brig.hs
```diff
@@ -58,7 +58,7 @@ data Api routes = Api
         :> "get-by-ids"
         :> ReqBody '[JSON] [UserId]
         :> Post '[JSON] [UserProfile],
-    claimPrekey ::
+    getPrekey ::
```
jschaul, line 61: prekeys are removed from the database as they are given out, so in case that is important, the word claim might be better than get, but that's up for debate.

pcapriotti: Yep, makes sense, I changed them to `claim` everywhere.

smatting (approved):

PR #1512, opened 2021-05-18
Try to use schema-profunctor lib.

fisx:


libs/wire-api/src/Wire/API/User/Client.hs
```diff
@@ -97,6 +97,92 @@ import Wire.API.Arbitrary (Arbitrary (arbitrary), GenericUniform (..), generateE
...
+    ClientSupportsLegalholdImplicitConsent
+  deriving stock (Eq, Ord, Bounded, Enum, Show, Generic)
+  deriving (Arbitrary) via (GenericUniform SupportedClientFeature)
+  deriving (ToJSON, FromJSON, Swagger.ToSchema) via S.Schema SupportedClientFeature
+
+instance S.ToSchema SupportedClientFeature where
+  schema =
+    S.named "SupportedClientFeature" $
+      S.tag _ClientSupportsLegalholdImplicitConsent S.schema
+
```
pcapriotti, line 148: You should use `enum` here. Something like:
```haskell
S.enum @Text "SupportedClientFeature" $
  S.element "legalhold-implicit-consent" ClientSupportsLegalholdImplicitConsent
```

libs/wire-api/src/Wire/API/User/Client.hs
```diff
@@ -97,6 +97,92 @@ import Wire.API.Arbitrary (Arbitrary (arbitrary), GenericUniform (..), generateE
...
+  Doc.description "Hints provided by the client for the backend so it can behavior in a backwards-compatible way."
+  Doc.property "feature_list" (Doc.array typeSupportedClientFeature) $ do
+    Doc.description "Array containing all supported features."
+
+instance ToJSON SupportedClientFeatureList where
+  toJSON (SupportedClientFeatureList l) = object ["feature_list" .= l]
+
+instance FromJSON SupportedClientFeatureList where
+  parseJSON = withObject "SupportedClientFeatureList" $ \obj -> SupportedClientFeatureList <$> obj .: "feature_list"
+
```
pcapriotti, line 185: These can be derived automatically if you write an `S.ToSchema` instance. Something like the following should work (haven't tried compiling it though):
```haskell
S.object "SupportedClientFeatureList" $
  fromSupportedClientFeatureList S..= S.field "feature_list" S.schema
```

pcapriotti (changes requested):
I have left some comments.

If it's not too much of a pain, could you also make the `Schema` import unqualified, and qualify `Aeson`? This is the way I've dealt with imports in other places.

fisx:
thanks, this is helpful!  I tried again, and got stuck a little further down, but i think i'm almost there.  :)

as for the unqualified import of `Data.Schema`, I would like to leave that for a dedicated refactoring PR that transforms the entire module, otherwise it'll get too noisy.

NB: we've tried to establish a policy in the backend chapter that forbids unqualified, implicit imports.  with limited success, admittedly, but there appeared to be a weak consensus when we debated it.

libs/wire-api/src/Wire/API/User/Client.hs
```diff
@@ -97,6 +100,93 @@ import Wire.API.Arbitrary (Arbitrary (arbitrary), GenericUniform (..), generateE
...
+data SupportedClientFeatureList = SupportedClientFeatureList {fromSupportedClientFeatureList :: Set SupportedClientFeature}
+  deriving stock (Eq, Ord, Show, Generic)
+  deriving (Arbitrary) via (GenericUniform SupportedClientFeatureList)
+  deriving (ToJSON, FromJSON, Swagger.ToSchema) via (Schema.Schema SupportedClientFeatureList)
+
+instance Schema.ToSchema SupportedClientFeatureList where
+  schema =
+    Schema.object "SupportedClientFeatureList" $
+      SupportedClientFeatureList
+        <$> (fromSupportedClientFeatureList Schema..= Schema.field "feature_list" Schema.schema)
```
pcapriotti, line 179: The brackets here should not be necessary.

pcapriotti (approved):
Looks good! Agreed about the unqualified import, that can be done separately.

You can also attach descriptions using `objectWithDocModifier "name" (description ?~ "desc")` instead of `object "name"`.

Minor nitpick follows.

fisx:
looks good, but doesn't compile yet!  :)  can you guess the type error from looking at the change?

fisx:
spoiler:

```
/home/mf/src/wire/wire-server/libs/wire-api/src/Wire/API/User/Client.hs:182:12: error:
    • Couldn't match type ‘[SupportedClientFeature]’
                     with ‘Set SupportedClientFeature’
```

pcapriotti:
Ah, sorry, I didn't realise there was a type mismatch there. You can fix it by adding the set to list conversion functions in the right place. Here's an excerpt from the schema for `NewConv`:
```haskell
      <*> (Set.toList . newConvAccess)
        .= ( field "access" (Set.fromList <$> array schema)
               <|> pure mempty
           )
```

PR #1514, opened 2021-05-19
Declare better ToSchema instance for NoContent

akshaymankar:
ToSchema instance for unit gives an empty array, which is not how NoContent
responses are encoded.

### DIFF
--- a/libs/wire-api/src/Wire/API/Routes/Public/Galley.hs
+++ b/libs/wire-api/src/Wire/API/Routes/Public/Galley.hs
@@ -20,6 +20,7 @@
 
 module Wire.API.Routes.Public.Galley where
 
+import Control.Lens ((?~))
 import Data.CommaSeparatedList
 import Data.Id (ConvId, TeamId)
 import Data.Range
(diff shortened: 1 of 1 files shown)


pcapriotti (commented):
This is certainly an improvement, but not the right solution. This instance should really not be needed. I think the way to go is to slightly generalise the `HasSwagger` instance for the recursive case of `UVerb` in servant-swagger to be something like:
```haskell
  ( HasStatus a,
    AllAccept cs,
    SwaggerMethod method,
    HasSwagger (UVerb method cs as),
    HasSwagger (Verb method (StatusOf a) cs a)
```
so that it doesn't require `ToSchema a` at all.

akshaymankar:
Hmm, ok. I will see if I can make a PR with this tomorrow.

akshaymankar:
@pcapriotti This is obsolete now right?

pcapriotti:
> @pcapriotti This is obsolete now right?

If #1657 gets merged, yes.

akshaymankar:
> > @pcapriotti This is obsolete now right?
> 
> If #1657 gets merged, yes.

Cool, I will close this then! Thanks!

PR #1515, opened 2021-05-19
Turn off flaky test.

fisx:
We should really turn off all those flaky tests: if we have an issue that states a test is broken, nothing is gained by keeping it enabled as long as we keep ignoring the failures.

### DIFF
--- a/services/brig/test/integration/API/Search.hs
+++ b/services/brig/test/integration/API/Search.hs
@@ -257,7 +257,7 @@ testReindex brig = do
     mkRegularUser = randomUserWithHandle brig
 
 testOrderName :: TestConstraints m => Brig -> m ()
-testOrderName brig = do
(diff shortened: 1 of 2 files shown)


fisx:
So you're vetoing the PR?  You have a point, I just think it's really painful to have to re-run the CI in what feels like 20-30% of all runs, and it's already slow enough as it is.

But maybe I'm exaggerating.  So if you want to veto this PR, please close it, I won't fight.  :)

akshaymankar:
Honestly, I think taking down safety nets to go fast is not a good idea. We should instead spend some time in making the safety net better. I don't want to veto things without a discussion, so can we please have that?

fisx:
no time for a discussion, sorry, working on improving safety nets in our code as we speak  :)

i agree that it'd be good to unflake tests rather than deactivate them, so let's aim for that.

PR #1517, opened 2021-05-20
Remove OptionallyQualified data type from types-common

mdimjasevic:
This simple PR removes the `Data.Qualified.OptionallyQualified` data type from types-common as it is unused.


jschaul (approved):
Looks good, however the code formatter (ormolu) doesn't agree. You can run `make format` at the top-level directory of wire-server to re-format all haskell files. In general it will be useful to also hook ormolu into your editor config to run automatically on file save.

PR #1518, opened 2021-05-20
RFC: Use sed instead of yq to read yaml files in ormolu script

pcapriotti:
This is probably slightly less robust, but it removes a dev dependency with multiple incompatible incarnations (https://github.com/mikefarah/yq, https://github.com/abesto/yq, https://github.com/kislyuk/yq ...) and different incompatible versions. Also avoids the use of awk in the script.


akshaymankar (approved):
Ship it, if it works :)

fisx:
This used to be perl before it was yq...  :-)  I wonder if this is going to stabilize now?

PR #146, opened 2017-12-11
Let Galley read client list form Brig

elverkilde:
Change the Galley read path for client lookup to go through Brig. The write path is still the same, i.e. Galley still writes its own client mapping to Cassandra. This way we can we test the implications of the higher read latency before making a (potential) full switch.

The current code has two identical definitions for userid to client mappings, one for internal use in Galley and one for serialization. This is changed slightly so that the external representation is wrapped in the internal one (it's not pretty but it works).

PR #1519, opened 2021-05-20
[charts:sftd] Introduce flag to enable TURN discovery

lucendio:
Enabling what [docs will state](https://github.com/wireapp/wire-docs/pull/146/files#diff-3131d91b6a6d39e99b97c9303ac446d36e6d948fd4b2de09f76bf2e919dcbae3R73-R76).

### DIFF
--- a/charts/sftd/values.yaml
+++ b/charts/sftd/values.yaml
@@ -81,3 +81,5 @@ joinCall:
     # Overrides the image tag whose default is the chart appVersion.
     tag: "1.19.5"
 
+# Make the SFT to proxy itself with some TURN server when trying to connect to clients
+turnDiscoveryEnabled: false
--- a/charts/sftd/templates/statefulset.yaml
+++ b/charts/sftd/templates/statefulset.yaml
@@ -83,7 +83,12 @@ spec:
               else
                 ACCESS_ARGS="-A ${EXTERNAL_IP}"
               fi
-              exec sftd  -I "${POD_IP}" -M "${POD_IP}" ${ACCESS_ARGS} -u "https://{{ required "must specify host" .Values.host }}/sfts/${POD_NAME}"
+              exec sftd \
+                      -I "${POD_IP}" \
+                      -M "${POD_IP}" \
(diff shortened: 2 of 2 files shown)


charts/sftd/values.yaml
```diff
@@ -81,3 +81,5 @@ joinCall:
     # Overrides the image tag whose default is the chart appVersion.
     tag: "1.19.5"
 
+# Make the SFT to proxy itself with some TURN server when trying to connect to clients
```
jschaul, line 84: I'm not sure I fully understand everything, so please get confirmation from AVS. But to my understanding, the following might be a more helpful comment:

```suggestion
# When enabled, allow media traffic to flow clients <--> TURN <--> SFT [that allows UDP OR TCP OR TLS between clients and TURN. UDP traffic is still required between TURN and SFT]
# When disabled, only allow media traffic to flow clients <--> SFT [that requires direct UDP packets between clients and SFT]
```

lucendio: Admittedly, the comment is phrased in a weird way. but your suggestion is not true. When setting the flag, it would actually be: clients <--> TURN <--> TURN <--> SFT (with the arrows optionally being TCP-only). See [this PR #146 at wire-docs](https://github.com/wireapp/wire-docs/pull/146/files#diff-3131d91b6a6d39e99b97c9303ac446d36e6d948fd4b2de09f76bf2e919dcbae3R73-R76). I'm still failing to find some better phrasing.

jschaul: Okay sorry, I don't fully understand. Are you able to maybe partially re-use some of the phrasing and replace some of the arrows in question with better arrows?
Or perhaps link to the section in the docs that explain what happen with this is enabled or disabled?

lucendio: I made another small change to the comment in the code, which also refers to a [recent change in the docs](https://github.com/wireapp/wire-docs/pull/146). Would you mind taking another quick look and let me know if there is still sth to be improved.

jschaul (approved):
Looks good!

PR #1520, opened 2021-05-21
Work around a locale issue in Ormolu

arianvp:
Otherwise ormolu doesn't work


mdimjasevic (approved):

mdimjasevic:
@arianvp and @pcapriotti , I just updated the PR to use the `LOCALE_ARCHIVE` variable instead, i.e., this should be a proper workaround for the locale issue.

shell.nix
```diff
@@ -58,6 +58,7 @@ let
   };
 in pkgs.mkShell {
   name = "shell";
+  LOCAL_ARCHIVE=${glibcLocales}/lib/locale/locale-archive; # works around https://github.com/tweag/ormolu/issues/38
```
akshaymankar, line 61: Shouldn't this be `pkgs.glibcLocales`?

mdimjasevic: I am an absolute beginner at Nix. Should I replace "glibcLocales" with "pkgs.glibcLocales"?

pcapriotti (changes requested):
This doesn't look right. `LOCALE_ARCHIVE` should be an environment variable, not an attribute.

arianvp:
@pcapriotti attributes of a derivation are exposed as environment variables.

shell.nix
pcapriotti: You need to wrap it in quotes and qualify `glibcLocales` with `pkgs`.

shell.nix
```diff
@@ -58,6 +58,7 @@ let
   };
 in pkgs.mkShell {
   name = "shell";
+  LOCAL_ARCHIVE="${pkgs.glibcLocales}/lib/locale/locale-archive"; # works around https://github.com/tweag/ormolu/issues/38
```
pcapriotti, line 61:
```suggestion
  LOCALE_ARCHIVE="${pkgs.glibcLocales}/lib/locale/locale-archive"; # works around https://github.com/tweag/ormolu/issues/38
```

pcapriotti (changes requested):
I think it should be `LOCALE_ARCHIVE`, not `LOCAL_ARCHIVE`, right? Have you tried to see if it fixes your issue?

pcapriotti (approved):

PR #1524, opened 2021-05-21
Expose mock federator in wire-api-federation

pcapriotti:
Also use this general mock federator in brig, instead of hand-rolling one.


akshaymankar (approved):
There is a conflict in the cabal file, apart from that this looks good :shark:

jschaul:
CI is green, merging so other PRs can be rebased on develop.

PR #1526, opened 2021-05-24
Block device handshake in case of LH policy conflict

fisx:
https://wearezeta.atlassian.net/browse/SQSERVICES-454

- [x] ask clients: are clients also explicitely requested? on which api?
- [x] !! API Changes. Verify that the endpoints are authenticated (Z-User) in all environments
- [x] Decide whether to implement LH plus federation (includes writing more tests!)

### DIFF
--- a/libs/wire-api-federation/src/Wire/API/Federation/API/Brig.hs
+++ b/libs/wire-api-federation/src/Wire/API/Federation/API/Brig.hs
@@ -30,6 +30,7 @@ import Wire.API.Arbitrary (GenericUniform (..))
 import Wire.API.Federation.Client (FederationClientError, FederatorClient)
 import qualified Wire.API.Federation.GRPC.Types as Proto
 import Wire.API.Message (UserClients)
+import Wire.API.Team.LegalHold (LegalholdProtectee)
 import Wire.API.User (UserProfile)
 import Wire.API.User.Client (UserClientPrekeyMap)
 import Wire.API.User.Client.Prekey (ClientPrekey, PrekeyBundle)
@@ -61,19 +62,19 @@ data Api routes = Api
       routes
         :- "federation"
         :> "claim-prekey"
-        :> ReqBody '[JSON] (UserId, ClientId)
+        :> ReqBody '[JSON] (LegalholdProtectee, UserId, ClientId)
         :> Post '[JSON] (Maybe ClientPrekey),
     claimPrekeyBundle ::
       routes
         :- "federation"
         :> "claim-prekey-bundle"
-        :> ReqBody '[JSON] UserId
+        :> ReqBody '[JSON] (LegalholdProtectee, UserId)
         :> Post '[JSON] PrekeyBundle,
     claimMultiPrekeyBundle ::
       routes
         :- "federation"
         :> "claim-multi-prekey-bundle"
-        :> ReqBody '[JSON] UserClients
+        :> ReqBody '[JSON] (LegalholdProtectee, UserClients)
         :> Post '[JSON] UserClientPrekeyMap,
     searchUsers ::
       routes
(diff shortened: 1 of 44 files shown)


fisx:
>     * [ ]  Update all routing tables with new internal endpoint

what do you mean?  it's an *internal* end-point, nginz isn't supposed to know about it.

libs/wire-api-federation/src/Wire/API/Federation/API/Brig.hs
```diff
@@ -61,19 +62,19 @@ data Api routes = Api
...
         :> "claim-prekey-bundle"
-        :> ReqBody '[JSON] UserId
+        :> ReqBody '[JSON] (LegalholdProtectee, UserId)
         :> Post '[JSON] PrekeyBundle,
     claimMultiPrekeyBundle ::
       routes
         :- "federation"
         :> "claim-multi-prekey-bundle"
-        :> ReqBody '[JSON] UserClients
+        :> ReqBody '[JSON] (LegalholdProtectee, UserClients)
```
fisx, line 77: I don't think this can work.  We need to get `ZUser` from the header, the request body cannot be expected to change, these are external end-points.

services/galley/src/Galley/API/Update.hs
```diff
@@ -980,7 +984,7 @@ handleOtrResponse ::
 handleOtrResponse usr clt rcps membs clts val now go = case checkOtrRecipients usr clt rcps membs clts val now of
   ValidOtrRecipients m r -> go r >> pure (OtrSent m)
   MissingOtrRecipients m -> do
-    guardLegalholdPolicyConflicts usr (missingClients m)
+    guardLegalholdPolicyConflicts (ProtectedUser usr) (missingClients m)
```
fisx, line 987: Are you sure it's always a user here, never a bot?

libs/wire-api-federation/src/Wire/API/Federation/API/Brig.hs
fisx: Ah, get it.  this is between federated instances.  But I still don't know if we can just change it like this.  We shouldn't really touch federation here, it's not supposed to work yet.  I'll try to surgically remove this again.

services/galley/src/Galley/API/Update.hs
fisx: Actually, I think you're right: bots have their own, entirely separate API.  I've always been annoyed by that, but at least this once it works in our favor.  :)

fisx (commented):
Actually, I'm not sure any more about LH+Federation, maybe solving it now is easy enough?  I've started reverting it, let's decide tomorrow.

services/brig/src/Brig/Data/Client.hs
```diff
@@ -200,8 +213,9 @@ updatePrekeys u c pks = do
         Success n -> return (CryptoBox.prekeyId n == keyId (prekeyId a))
         _ -> return False
 
-claimPrekey :: UserId -> ClientId -> AppIO (Maybe ClientPrekey)
-claimPrekey u c =
+claimPrekey :: LegalholdProtectee -> UserId -> ClientId -> AppIO (Maybe ClientPrekey)
+claimPrekey _protectee u c =
+  -- TODO: guard the protectee here!
```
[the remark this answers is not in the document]
smatting: We decided not do it on the database level, but at the callsites of this function

services/galley/src/Galley/Intra/Client.hs
```diff
@@ -14,9 +14,11 @@
 --
 -- You should have received a copy of the GNU Affero General Public License along
 -- with this program. If not, see <https://www.gnu.org/licenses/>.
+{-# OPTIONS_GHC -Wno-deferred-type-errors #-}
```
smatting, line 17: note to self: remove this

smatting (approved):

PR #1527, opened 2021-05-25
Prettier looking golden tests

akshaymankar:
This will enable us to reason about these json and haskell files better.


pcapriotti (commented):
Looks good! 🚀

pcapriotti (approved):

PR #1528, opened 2021-05-25
Add remote users when creating a conversation

mdimjasevic:
This adds remote users when creating a conversation. Several checks are still missing, as noted in a related funuction `addMembers`:

- Remote qualified users must exist before they can be added (a call to the respective backend should be made): Avoid clients making up random Ids, and increase the chances that the updateConversationMembership call suceeds
- A call must be made to the remote backend informing it that this user is now part of that conversation. Use and implement 'updateConversationMemberships'.
-- that call should probably be made *after* inserting the conversation membership happens in this backend.                                                                                                                                                           
-- 'updateConversationMemberships' should send an event to the affected users informing them they have joined a remote conversation.
- Events should support remote / qualified users, too.

Furthermore, the collection of updated tests has to be updated with a non-empty list of remote users.


services/galley/src/Galley/API/Create.hs
```diff
@@ -98,17 +99,21 @@ internalCreateManagedConversation zusr zcon (NewConvManaged body) = do
...
 createRegularGroupConv zusr zcon (NewConvUnmanaged body) = do
   name <- rangeCheckedMaybe (newConvName body)
-  _uids <- checkedConvSize (newConvUsers body) -- currently not needed, as we only consider local IDs
+  localDomain <- viewFederationDomain
   let localUserIds = newConvUsers body
+      qualifiedUserIds = newConvQualifiedUsers body
   ensureConnected zusr localUserIds
-  localCheckedUsers <- checkedConvSize localUserIds
+  let allUsers = map (`Qualified` localDomain) localUserIds <> qualifiedUserIds
+  checkedUsers <- checkedConvSize allUsers
```
jschaul, line 107: The connection check `ensureConnected` is lost here when doing the check only on non-qualified users. That check also needs to happen on all qualified users with the localDomain. (A test to that effect would be great, too). A FUTUREWORK should be added to also do that connection check once we have federated connections.

services/galley/test/integration/API/Util.hs
```diff
@@ -427,6 +427,7 @@ createTeamConvLegacy u tid us name = do
   let convPayload =
         object
           [ "users" .= us,
+            "qualified_users" .= ([] :: [Qualified UserId]),
```
jschaul, line 430: for backwards compatibility, it must be possible to not send qualified_users here. If this line is required for tests to pass, then the fromJSON instance of NewConv is wrong and needs to default to an empty list if not provided.

jschaul (commented):
I see this is still WIP, but I had an initial look and made two inline comments. Generally the approach looks good!

services/galley/src/Galley/API/Create.hs
mdimjasevic: Just to confirm: @jschaul , this is resolved now, right?

mdimjasevic:
@jschaul , I removed the WIP tag from the PR title, but perhaps you have a suggestion what more to add or change.

services/galley/src/Galley/API/Create.hs
jschaul: Yes, that is resolved now (for local users). Remote users and connections to remote users will come later.

services/galley/src/Galley/API/Create.hs
```diff
@@ -118,22 +125,28 @@ createRegularGroupConv zusr zcon (NewConvUnmanaged body) = do
...
-  otherConvMems <-
+  checkedPartitionedUsers' <-
     if cnvManaged tinfo
       then do
         -- ConvMaxSize MUST be < than hardlimit so the conv size check is enough
         maybeAllMembers <- Data.teamMembersForFanout convTeam
         let otherConvMems = filter (/= zusr) $ map (view userId) $ (maybeAllMembers ^. teamMembers)
-        checkedConvSize otherConvMems
+        checkedLocalUsers <- checkedConvSize otherConvMems
+        pure (fmap ([],) checkedLocalUsers)
```
jschaul, line 148: I think I don't understand why an empty list is added here. Maybe add a comment/TODO/futurework?

jschaul (approved):
Looks good when CI agrees.

services/galley/src/Galley/API/Create.hs
mdimjasevic: I left a comment for future work, but I believe @pcapriotti and me got to a conclusion that in this branch there can be no remote users.

PR #1529, opened 2021-05-26
Add remote members to conversations

pcapriotti:
Some preliminary work to make the endpoint for adding remote users to conversations functional. This implements point (1) in the comment of `addMembers`, i.e. it checks if remote users exist before adding them to a conversation. If some remote users do not exist, a new error with status code 400 is returned.

I also added some tests using the mock federator exposed by #1524.

The rest of the functionality required by `addMembers` will be implemented in follow-up PRs.

## TODO

- [x] test deleted user failure scenario
- [x] ~add end2end test~ (Moved to #1538)

### DIFF
--- a/services/galley/test/integration/API.hs
+++ b/services/galley/test/integration/API.hs
@@ -111,8 +111,7 @@ tests s =
           test s "add past members" postMembersOk3,
           test s "fail to add members when not connected" postMembersFail,
           test s "fail to add too many members" postTooManyMembersFail,
-          -- TODO: make it into an end2end test or mock the federator
-          -- test s "add remote members" testAddRemoteMember,
+          test s "add remote members" testAddRemoteMember,
           test s "add remote members on invalid domain" testAddRemoteMemberInvalidDomain,
           test s "remove members" deleteMembersOk,
           test s "fail to remove members from self conv." deleteMembersFailSelf,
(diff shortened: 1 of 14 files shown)


services/galley/test/integration/API.hs
```diff
@@ -888,6 +900,37 @@ testAddRemoteMember = do
...
+  convId <- decodeConvId <$> postConv alice [] (Just "remote gossip") [] Nothing Nothing
+  opts <- view tsGConf
+  g <- view tsGalley
+  liftIO $ do
+    (resp, _) <-
+      withTempMockFederator
+        opts
+        remoteDomain
+        [mkProfile remoteCharlie (Name "charlie")]
+        (postQualifiedMembers' g alice (remoteBob :| []) convId)
```
jschaul, line 920: Do I read this code correctly as "dear Mock Federator, please give back a profile for Charlie when asking for the userId of bob" ? If so, this is an interesting test case that is good to cover but implausible to happen in well-behaved backends. Could you also extend this test to include the (likely more common) case of the domain being correct but that user not existing (should give 404 in the client-server API, not sure what the RPC will give) or that user having been deleted (deleted flag in the profile being set to true)?

jschaul (commented):
Looks good.

I suppose an end2end test would be nice to catch possible misconfigurations here now that real RPCs are made from/to galley.

You can also remove point (1) from that futurework comment now 😸

services/galley/test/integration/API.hs
pcapriotti: I think what I meant to do here is add both Bob and Charlie to the conversation, where the remote only knows about Charlie. But I forgot to include Charlie in the request.

I will add the other tests you suggested, though, they do sound like good scenarios to check.

jschaul (approved):

PR #1532, opened 2021-05-26
Servantify JSON endpoint to send messages

akshaymankar:
This PR contains couple of commits from #1531 which need to be removed when that is merged


fisx:
I'm assuming this will not break the way integration tests call this end-point?

akshaymankar:
I'd be surprised if we have integration tests verifying the time of the mismatch :)

fisx:
> I'd be surprised if we have integration tests verifying the time of the mismatch :)

I'm talking about new tests.  In [a neighbor PR](https://github.com/wireapp/wire-server/pull/1526), we use the old end-point in tests, and hope that the new one and the helper functions in the integration tests will work the same way.  (It would be nice to have integration tests use servant-client, but it would make rebasing or merging very painful in this case.)

pcapriotti:
Nothing should be changed in terms of the API interface, so I don't see why integration tests would be affected.

jschaul (approved):

PR #1533, opened 2021-05-26
Add some missing charts

arianvp:
I noticed these aren't getting bumped and synced anymore

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,7 +14,7 @@ CHARTS_INTEGRATION    := wire-server databases-ephemeral fake-aws nginx-ingress-
 # (e.g. move charts/brig to charts/wire-server/brig)
 # this list could be generated from the folder names under ./charts/ like so:
 # CHARTS_RELEASE := $(shell find charts/ -maxdepth 1 -type d | xargs -n 1 basename | grep -v charts)
-CHARTS_RELEASE        := wire-server redis-ephemeral databases-ephemeral fake-aws fake-aws-s3 fake-aws-sqs aws-ingress backoffice calling-test demo-smtp elasticsearch-curator elasticsearch-external elasticsearch-ephemeral fluent-bit minio-external cassandra-external nginx-ingress-controller nginx-ingress-services reaper wire-server-metrics sftd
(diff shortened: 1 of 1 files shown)


arianvp:
Or are these also "inline local subcharts" so not really needed? (Reading the comment).

Maybe instead I should take a bit of time to refactor the charts a bit...

jschaul (approved):

Makefile
```diff
@@ -14,7 +14,7 @@ CHARTS_INTEGRATION    := wire-server databases-ephemeral fake-aws nginx-ingress-
 # (e.g. move charts/brig to charts/wire-server/brig)
 # this list could be generated from the folder names under ./charts/ like so:
 # CHARTS_RELEASE := $(shell find charts/ -maxdepth 1 -type d | xargs -n 1 basename | grep -v charts)
-CHARTS_RELEASE        := wire-server databases-ephemeral fake-aws aws-ingress backoffice calling-test demo-smtp elasticsearch-curator elasticsearch-external fluent-bit minio-external cassandra-external nginx-ingress-controller nginx-ingress-services reaper wire-server-metrics sftd
+CHARTS_RELEASE        := wire-server redis-ephemeral databases-ephemeral fake-aws fake-aws-s3 fake-aws-sqs aws-ingress backoffice calling-test demo-smtp elasticsearch-curator elasticsearch-external elasticsearch-ephemeral fluent-bit minio-external cassandra-external nginx-ingress-controller nginx-ingress-services reaper wire-server-metrics sftd
```
jschaul, line 17: The `-ephemeral` ones are covered from the databases-ephemeral being their wrapper chart.

I suppose the only new one added here is aws-ingress, which wasn't there before.

arianvp: hmm but the subcharts dont get published "on their own" it seems.  e.g. the latest `redis-ephemeral` chart is `0.130.0` instead of `2.something`.   Whilst the `redis-ephemeral` chart inside `databases-ephemeral` is newer.    IDK if this is a problem; because people should probably always use the wrapper chart; but it looks a bit odd..

I guess most of this gets 'fixed'  as listed in the comment above. Namely not having all the charts top-level anymore; but make clear that a subchart is a subchart by  putting it in the `charts/toplevel-chart/chart/<subchart>` folder..

jschaul: I also don't know if this is a problem either. I'm happy either way, if you'd like to publish some subcharts on their own, that's cool with me. Approving this PR in case you'd like to merge it as-is. Otherwise feel free to make changes.

jschaul (approved):

PR #1534, opened 2021-05-26
Release 2021-05-26

fisx:
See https://github.com/wireapp/wire-server/blob/d65ec3c26/CHANGELOG.md#2021-05-26


jschaul (approved):
Changelog looks good.

PR #1535, opened 2021-05-27
Use Servant.respond instead of explicit SOP

pcapriotti:
This replaces manual lifting to a Union type with a call to `Servant.respond`, helped by some annotations. Thanks @akshaymankar for showing me how to make this work.


akshaymankar (approved):
:tada:

PR #1543, opened 2021-05-27
expose /conversations/{cnv}/members/v2 for federation backends

jschaul:
For (test, non-production) backends explicitly configured to use federation, expose /conversations/{cnv}/members/v2 to clients. Note that the implementation of using this endpoint isn't complete yet, but it can nonetheless be tried out by clients.

This PR adds tests to ensure this endpoint cannot be called with remote users on environments that do not support federation.

### DIFF
--- a/services/galley/test/integration/API.hs
+++ b/services/galley/test/integration/API.hs
@@ -969,15 +969,24 @@ testAddRemoteMemberFederationDisabled = do
       const 400 === statusCode
       const (Just "federation-not-enabled") === fmap label . responseJsonUnsafe
   -- federator endpoint being configured in brig and/or galley, but not being
-  -- available (i.e. no service listing on that IP/port) is the case for
-  -- kubernetes-based deployments that do not explicitly enable the federator
-  -- service.
+  -- available (i.e. no service listing on that IP/port) can happen due to a
+  -- misconfiguration of federator. That should give a 500.
   -- Port 1 should always be wrong hopefully.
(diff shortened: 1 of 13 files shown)


jschaul:
Dear reviewer, if you agree with this PR, please merge it.

pcapriotti (commented):
I don't understand what the point of this change is. Why should we conflate the two errors "federation-not-enabled" and "federation-not-available"? Is there any reason why we can't fix the staging/production environment by either disabling the federator endpoint completely, or by deploying a federator?

services/galley/test/integration/API.hs
```diff
@@ -953,6 +952,42 @@ testAddRemoteMemberInvalidDomain = do
...
+  -- available (i.e. no service listing on that IP/port) can happen due to a
+  -- misconfiguration of federator. That should give a 500.
+  -- Port 1 should always be wrong hopefully.
+  let federatorUnavailable :: Opts = opts & optFederator ?~ Endpoint "127.0.0.1" 1
+  withSettingsOverrides federatorUnavailable $ do
+    postQualifiedMembers' g alice (remoteBob :| []) convId !!! do
+      const 500 === statusCode
+      -- FUTUREWORK: this should be a federation-not-available
+      -- but the error is hidden inside a server-error, confusingly.
+      -- separate task see cryptpad
```
pcapriotti, line 981: Maybe not refer to internal documents in public facing code?

pcapriotti (approved):
Looks good. Minor nitpick follows.

PR #1547, opened 2021-05-28
Qualify users and conversations in Event

pcapriotti:
This makes the conversation ID and originating user ID of `Event` qualified, as well as the new member IDs in the case of `MemberJoin` event.

The change is pretty big, but there is nothing really going on except adjusting all the places where events are used to make everything type-check. Some functions gained a new `Domain` argument which is assumed to be the local domain, in order to be able to qualify IDs.

As more and more things become qualified, the need for these ad-hoc qualifications should hopefully reduce.


services/galley/src/Galley/API/Federation.hs
```diff
@@ -56,12 +55,12 @@ updateConversationMemberships cmu = do
   when (not (null localUsers)) $ do
     Data.addLocalMembersToRemoteConv localUserIds (cmuConvId cmu)
   -- FUTUREWORK: the resulting event should have qualified users and conversations
```
jschaul, line 57: that futurework is done I suppose (feel free to remove this comment only in a follow-up PR though so you don't need to wait for CI again)

jschaul (approved):
It's a big diff to review indeed. I admit I only looked at the Haskell changes and did not review the changes to the 125 json files, but glancing at it it appears backwards compatible. 
Great work!

PR #1548, opened 2021-05-29
Only update ES index at the end of member add loop.

fisx:
I can't think of a reason why an up-to-date index should be required during adding more users?


akshaymankar (commented):
I think this will allow you to create a bigger team than max team size, but maybe no tests depend on that. I see a couple of legalhold tests failed, but only in the second run, due to wrong member count. I don't know if that is related.

akshaymankar:
Hmm, the test failed again, something is definitely fishy.

akshaymankar:
@fisx let's close this?

fisx:
I left a comment to avoid having to have this conversation again in 6 weeks.  @akshaymankar if you agree pls merge right away.  (I don't think this mandates a changelog entry.)

PR #1550, opened 2021-05-30
 Refactoring-friendly FutureWork data type

fisx:
internal change, no need to show in the release notes.  it's pretty small and self-explanatory.

### DIFF
--- a/libs/types-common/src/Data/Misc.hs
+++ b/libs/types-common/src/Data/Misc.hs
@@ -54,6 +54,9 @@ module Data.Misc
 
     -- * Swagger
     modelLocation,
+
+    -- * Typesafe FUTUREWORKS
+    FutureWork (..),
   )
 where
 
@@ -348,3 +351,12 @@ instance FromJSON PlainTextPassword where
 instance Arbitrary PlainTextPassword where
   -- TODO: why 6..1024? For tests we might want invalid passwords as well, e.g. 3 chars
   arbitrary = PlainTextPassword . fromRange <$> genRangeText @6 @1024 arbitrary
+
+-- | Usage:
+-- 1. Use this type in patterns to mark FUTUREWORKS.
+-- 2. Remove the label constructor -> all futureworks become compiler errors
+--
+-- Example:
+-- >>> let (FutureWork @'LegalholdPlusFederationNotImplemented -> _remoteUsers, localUsers)
+-- >>>      = partitionRemoteOrLocalIds domain qualifiedUids
+newtype FutureWork label payload = FutureWork payload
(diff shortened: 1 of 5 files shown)


libs/types-common/src/Data/Misc.hs
```diff
@@ -348,3 +351,12 @@ instance FromJSON PlainTextPassword where
...
   arbitrary = PlainTextPassword . fromRange <$> genRangeText @6 @1024 arbitrary
+
+-- | Usage:
+-- 1. Use this type in patterns to mark FUTUREWORKS.
+-- 2. Remove the label constructor -> all futureworks become compiler errors
+--
+-- Example:
+-- >>> let (FutureWork @'LegalholdPlusFederationNotImplemented -> _remoteUsers, localUsers)
+-- >>>      = partitionRemoteOrLocalIds domain qualifiedUids
+newtype FutureWork label payload = FutureWork payload
```
smatting, line 362: If you plan to use the payload then uwrapping the newtype is a bit noisy
```hs
let FutureWork x' = FutureWork @'FeatNotImplemented x
```
If you also provide an unwrapper `FutureWork -> payload` and combine wrapper and unwrapper for convenience, this degenerates to

```
  let  futurework :: forall a b. b -> b
       futurework = id

  let x = futurework @'FeatNotImplemented (42 :: Int)
```
Having `FutureWork` type however has the advantage that you can pass futurework-values to functions in a typesafe way. So I'm in favor of `FutureWork` now after some thinking :)

smatting (approved):

services/galley/src/Galley/API/Update.hs
```diff
@@ -1000,7 +1001,7 @@ withValidOtrRecipients protectee clt cnv rcps val now go = do
     Data.deleteConversation cnv
     throwM convNotFound
   -- FUTUREWORK(federation): also handle remote members
-  localMembers <- Data.members cnv
+  (FutureWork @'LegalholdPlusFederationNotImplemented -> _remoteMembers, localMembers) <- (undefined,) <$> Data.members cnv
```
akshaymankar, line 1004: Does this have to be `undefined`?

fisx: Do you prefer `()`?  I think I prefer `undefined` because it makes sure this makes it more clear it shouldn't be used anywhere.  Perhaps there is a better idiom yet to be found though; I tried something diferent [here](https://github.com/wireapp/wire-server/blob/a843ef28039ff9cb85d2ef57216076114547f488/services/brig/src/Brig/Provider/API.hs#L865)

PR #1556, opened 2021-05-31
Notify remote backends when users join

pcapriotti:
This makes a backend notify remote backends when users join a conversation.

**Note**: this includes commits from #1570. Rebase them away before merging.

### DIFF
--- a/services/galley/src/Galley/API/Update.hs
+++ b/services/galley/src/Galley/API/Update.hs
@@ -66,9 +66,11 @@ import qualified Brig.Types.User as User
 import Control.Arrow ((&&&))
 import Control.Lens
 import Control.Monad.Catch
+import Control.Monad.Except (runExceptT)
 import Control.Monad.State
 import Data.ByteString.Conversion (toByteString')
 import Data.Code
+import Data.Domain (Domain)
 import Data.Id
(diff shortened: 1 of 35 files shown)


services/galley/src/Galley/API/Update.hs
```diff
@@ -892,16 +892,16 @@ addToConversation (bots, lothers) rothers (usr, usrRole) conn locals remotes c =
...
       qusr = Qualified usr localDomain
   -- FUTUREWORK: parallelise federated requests
   traverse_ (uncurry (updateRemoteConversations now mm qusr qcnv))
     . Map.assocs
     . partitionQualified
-    . map fst
-    $ others
+    . nubOrd
+    . map (unTagged . rmId)
+    $ rmm <> rothers
```
mdimjasevic, line 904: If I am getting this right, this traversing rightfully notifies remote users. Is there an equivalent for local users?

mdimjasevic (approved):

services/galley/src/Galley/API/Update.hs
pcapriotti: This doesn't directly notify users, but it calls the `update-conversation-memberships` federation RPC, which will in turn notify users on its backend.

Notifications for local users are sent in the two statements that follow.

PR #1558, opened 2021-06-01
Parametric mock federator

pcapriotti:
This makes the fake response provided to a mock federator take a request argument. The motivation is that it allows a single mock to be used to provide multiple fake responses depending on the requests received.

For example, when testing the endpoint for adding remote members to conversations, the mock federator will be queried twice: once to fetch remote user profiles, and once to propagate the join event to the remote backend. Therefore, we need to be able to set up different fake responses for the two requests.

See [this commit](https://github.com/wireapp/wire-server/pull/1556/commits/1f69b233508a01378d54bfd694335023ecd1d345) in #1556 for a usage example.

### DIFF
--- a/services/galley/test/integration/API/Util.hs
+++ b/services/galley/test/integration/API/Util.hs
@@ -1666,6 +1666,12 @@ mkProfile quid name =
 
 -- mock federator
 
+-- | Run the given action on a temporary galley instance with access to a mock
+-- federator.
+--
+-- The `resp :: FederatedRequest -> a` argument can be used to provide a fake
+-- federator response (of an arbitrary JSON-serialisable type a) for every
+-- expected request.
 withTempMockFederator ::
   (MonadIO m, ToJSON a) =>
   Opts.Opts ->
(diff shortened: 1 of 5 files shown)


services/galley/test/integration/API/Util.hs
```diff
@@ -1671,19 +1671,19 @@ withTempMockFederator ::
   (MonadIO m, ToJSON a) =>
   Opts.Opts ->
   Domain ->
-  a ->
+  (FederatedRequest -> a) ->
```
jschaul, line 1674: I think I don't quite understand what this does and how the `resp / Request -> a` argument should be used. Could you add some inline comments here to these arguments?

jschaul (commented):
Could you explain what this PR does and how a mock function with a `Request -> a` argument is more useful than one with an `a` argument? It appears whatever the new functionality is, it's not made use of anywhere, so I'm a little bit lost here.

jschaul:
(also, this PR needs a merge/rebase due to conflicts)

pcapriotti:
> Could you explain what this PR does and how a mock function with a `Request -> a` argument is more useful than one with an `a` argument? It appears whatever the new functionality is, it's not made use of anywhere, so I'm a little bit lost here.

Sorry for not writing any PR description. I've added one now, hopefully that clarifies the reason for these changes. There might be some other way to test endpoints involving multiple federated calls, but making the fake response argument depend on the request seems the most straightforward.

jschaul (approved):

PR #1559, opened 2021-06-01
[charts] Update versions of webapp, team-settings, account-pages

lucendio:


jschaul (approved):
We have no tests for this, so I have no idea whether this change may break anything for installations. 
Not sure an approval process for these changes is the most useful though, as there isn't much a reviewer can check.

lucendio:
Agreed. Famous last words: *I created a ticket to track some progress on that end.*

PR #1560, opened 2021-06-01
Add more information to federation errors

pcapriotti:
This adds a `data` field to `Wai.Error`, containing some more specific error data. For now, only federation-specific data is supported. The federation error-conversion functions add domain and path information to the error, which can be used to diagnose which particular federated request failed in case of federation errors.

Most of the diff is about adapting every usage point of `Wai.Error`. Since `Wai.Error` gained a new field, and the constructor was used directly everywhere, I have replaced all uses of the constructor with a call to a smart constructor `mkError` which sets the extra data to `Nothing`.

### DIFF
--- a/libs/wai-utilities/src/Network/Wai/Utilities/Error.hs
+++ b/libs/wai-utilities/src/Network/Wai/Utilities/Error.hs
@@ -18,10 +18,17 @@
 -- You should have received a copy of the GNU Affero General Public License along
 -- with this program. If not, see <https://www.gnu.org/licenses/>.
 
-module Network.Wai.Utilities.Error where
+module Network.Wai.Utilities.Error
+  ( Error (..),
+    mkError,
+    (!>>),
+    byteStringError,
+  )
+where
 
 import Control.Error
 import Data.Aeson hiding (Error)
+import Data.Aeson.Types (Pair)
 import Data.Text.Lazy.Encoding (decodeUtf8)
 import Imports
 import Network.HTTP.Types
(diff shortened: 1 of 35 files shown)


libs/wai-utilities/src/Network/Wai/Utilities/Error.hs
```diff
@@ -18,40 +18,78 @@
...
   }
   deriving (Show, Typeable)
 
+mkError :: Status -> LText -> LText -> Error
+mkError c l m = Error c l m Nothing
+
 instance Exception Error
 
+data ErrorData = FederationErrorData
+  { federrDomain :: !Domain,
```
jschaul, line 52: Is this domain the originating domain or the remote domain? Possibly `FederationErrorData` could be extended in the future to contain both.

libs/wire-api-federation/src/Wire/API/Federation/Client.hs
```diff
@@ -58,26 +59,37 @@ instance KnownComponent 'Proto.Galley where
...
+        domain = targetDomain env
+        mkFailure = FederationClientFailure domain path
+        failure :: MonadError FederationClientFailure n => FederationClientError -> n x
+        failure = throwError . mkFailure
+        rpcFailure = failure . FederationClientRPCError
+        readBody = \case
+          RequestBodyLBS lbs -> pure $ LBS.toStrict lbs
+          RequestBodyBS bs -> pure bs
+          RequestBodySource _ -> failure FederationClientStreamingUnsupported
+    put (Just path)
```
jschaul, line 75: Hm, does this make an actual put request? Shouldn't this only happen a few lines below in `callRemote` ?

jschaul (commented):
Overall looks like a good change, I just have one inline question in Client.hs

libs/wai-utilities/src/Network/Wai/Utilities/Error.hs
pcapriotti: It's the remote domain. I assumed the originating domain would not be needed because it would simply be the one where the instance is running. But yes, it can be added easily.

libs/wire-api-federation/src/Wire/API/Federation/Client.hs
pcapriotti: Good question, I should have explained what this is in a comment. It is not a PUT request, but simply saving the path in the state of the monad stack. The function `put` is [this one](http://hackage.haskell.org/package/mtl-2.2.2/docs/Control-Monad-State-Class.html).

The idea here is that Servant is going to call this function for each request, and call `throwClientError` when some error occurs while translating a response into a Servant response. Since the path we want to add to the error structure is contained in the request, we have no direct way to access it from our implementation of `throwClientError` below.

My solution was to change the monad to keep a `Maybe ByteString` in its state, set it when a request is received, and fetch it when we need to produce an error. As far as I understand, Servant guarantees that `throwClientError` is only called after `runRequestAcceptStatus`, so this should work fine. If for any reason this guarantee does not hold, we get an empty path, but nothing catastrophic.

jschaul: Ah, that makes more sense. Too many functions have the same name and without a qualified import it's not directly obvious. Thanks for the explanation and the comments!

jschaul (approved):

PR #1568, opened 2021-06-03
Introduces the HasGalley class

mdimjasevic:
- It allows to access a Galley instance in a monad implementing the class.


pcapriotti (approved):
Looks good

PR #1569, opened 2021-06-03
Add remote users when creating a conversation

mdimjasevic:
This adds remote users when creating a conversation. Note: several checks are still missing, as noted in a related funuction `addMembers`:

* A call must be made to the remote backend informing it that this user is now part of that conversation. Use 'updateConversationMemberships'.
-- that call should probably be made after inserting the conversation membership happens in this backend.
* Events should support remote / qualified users, too.

This PR supersedes PR #1528.

### DIFF
--- a/libs/wire-api/test/unit/Test/Wire/API/Golden/FromJSON.hs
+++ b/libs/wire-api/test/unit/Test/Wire/API/Golden/FromJSON.hs
@@ -20,6 +20,7 @@ module Test.Wire.API.Golden.FromJSON where
 import Imports
 import Test.Tasty
 import Test.Tasty.HUnit
+import Test.Wire.API.Golden.Generated.NewConvUnmanaged_user
 import Test.Wire.API.Golden.Generated.NewOtrMessage_user
 import Test.Wire.API.Golden.Runner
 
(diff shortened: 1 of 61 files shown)


pcapriotti (commented):
Looks good, but it's missing some "fromJSON" golden tests to check that the schema for `NewConv` is backwards compatible, i.e. it can still parse `NewConv` requests that only have unqualified user ids.

I can add that myself later, if that's ok.

mdimjasevic:
> Looks good, but it's missing some "fromJSON" golden tests to check that the schema for `NewConv` is backwards compatible, i.e. it can still parse `NewConv` requests that only have unqualified user ids.
> 
> I can add that myself later, if that's ok.

Hm, I thought there is a test that does that. Initially I changed that test, but then someone (Joe?) told me it should still parse even when the new field is not given in JSON, so I reverted the change and the test still passed, i.e., it should be backwards compatible.

At any rate, if you think such a test case is missing, feel free to add it.

pcapriotti (approved):

libs/wire-api/test/unit/Test/Wire/API/Golden/FromJSON.hs
```diff
@@ -29,5 +30,8 @@ tests =
     "FromJSON golden tests"
     [ testCase ("NewOtrMessage") $
         testFromJSONObjects
-          [(testObject_NewOtrMessage_user_1, "testObject_NewOtrMessage_user_1.json")]
+          [(testObject_NewOtrMessage_user_1, "testObject_NewOtrMessage_user_1.json")],
+      testCase "NewConv" $
+        testFromJSONObjects
+          [(testObject_NewConvUnmanaged_user_1, "testObject_NewConvUnmanaged_user_1.json")]
```
mdimjasevic, line 36: Thank you for this!

pcapriotti:
:tada:

PR #1570, opened 2021-06-03
federator: rename Brig -> Service and add galley

pcapriotti:
Federator is currently forwarding all its requests to brig, regardless of the value of the `Component` field. This PR adds support for arbitrary internal services, renames the `Brig` effect to `Service` to reflect its generalised function, and adds a configuration option for galley.

### DIFF
--- a/services/federator/src/Federator/Run.hs
+++ b/services/federator/src/Federator/Run.hs
@@ -51,6 +51,7 @@ import qualified System.Logger.Extended as LogExt
 import UnliftIO (bracket)
 import UnliftIO.Async (async, waitAnyCancel)
 import Util.Options
+import Wire.API.Federation.GRPC.Types
 import qualified Wire.Network.DNS.Helper as DNS
 
 ------------------------------------------------------------------------------
@@ -86,8 +87,8 @@ newEnv o _dnsResolver = do
   _applog <- LogExt.mkLogger (Opt.logLevel o) (Opt.logNetStrings o) (Opt.logFormat o)
   let _requestId = def
   let _runSettings = Opt.optSettings o
-  let _brig = mkEndpoint (Opt.brig o)
-  let _brigEndpoint = Opt.brig o
+  let _brig Brig = mkEndpoint (Opt.brig o)
+      _brig Galley = mkEndpoint (Opt.galley o)
   _httpManager <- initHttpManager
   return Env {..}
   where
--- a/services/federator/src/Federator/Run.hs
+++ b/services/federator/src/Federator/Run.hs
@@ -87,12 +87,12 @@ newEnv o _dnsResolver = do
   _applog <- LogExt.mkLogger (Opt.logLevel o) (Opt.logNetStrings o) (Opt.logFormat o)
   let _requestId = def
   let _runSettings = Opt.optSettings o
-  let _brig Brig = mkEndpoint (Opt.brig o)
-      _brig Galley = mkEndpoint (Opt.galley o)
+  let _service Brig = mkEndpoint (Opt.brig o)
+      _service Galley = mkEndpoint (Opt.galley o)
   _httpManager <- initHttpManager
   return Env {..}
   where
-    mkEndpoint service = RPC.host (encodeUtf8 (service ^. epHost)) . RPC.port (service ^. epPort) $ RPC.empty
+    mkEndpoint s = RPC.host (encodeUtf8 (s ^. epHost)) . RPC.port (s ^. epPort) $ RPC.empty
 
 closeEnv :: Env -> IO ()
 closeEnv e = do
--- a/services/federator/src/Federator/Service.hs
+++ b/services/federator/src/Federator/Service.hs
@@ -25,8 +25,9 @@ import Bilge.RPC (rpc')
 import Control.Lens (view)
 import Data.Domain
 import Data.String.Conversions (cs)
+import qualified Data.Text.Lazy as LText
 import Federator.App (Federator, liftAppIOToFederator)
-import Federator.Env (brig)
+import Federator.Env (service)
 import Imports
 import qualified Network.HTTP.Types as HTTP
 import Polysemy
@@ -52,9 +53,9 @@ interpretService ::
   Sem r a
 interpretService = interpret $ \case
   ServiceCall component path body domain -> embed @Federator . liftAppIOToFederator $ do
-    brigReq <- view brig <$> ask
+    serviceReq <- view service <$> ask
     res <-
-      rpc' "brig" (brigReq component) $
+      rpc' (LText.pack (show component)) (serviceReq component) $
         RPC.method HTTP.POST
           . RPC.path path -- FUTUREWORK(federation): Protect against arbitrary paths
           . RPC.body (RPC.RequestBodyBS body)
--- a/services/federator/src/Federator/Service.hs
+++ b/services/federator/src/Federator/Service.hs
@@ -15,10 +15,8 @@
 -- You should have received a copy of the GNU Affero General Public License along
 -- with this program. If not, see <https://www.gnu.org/licenses/>.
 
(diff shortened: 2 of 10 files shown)


jschaul (commented):
Hm, interesting! We didn't catch this before, when we only use mocks... but it would be nice to be able to catch this sort of problem without an end2end test.

As https://github.com/wireapp/wire-server/blob/develop/services/brig/test/integration/Federation/End2end.hs#L236 returns a 200, if galley can't be called yet, I suppose that is because currently only a call to a remote brig is made, and https://github.com/wireapp/wire-server/pull/1556 isn't merged yet? 

(Even though this is only a draft, changes so far look good)

pcapriotti:
> Hm, interesting! We didn't catch this before, when we only use mocks... but it would be nice to be able to catch this sort of problem without an end2end test.

Agreed, but I'm not sure how.

> As https://github.com/wireapp/wire-server/blob/develop/services/brig/test/integration/Federation/End2end.hs#L236 returns a 200, if galley can't be called yet, I suppose that is because currently only a call to a remote brig is made, and #1556 isn't merged yet?

That's right, that test is indeed failing in #1556.

services/federator/src/Federator/Run.hs
```diff
@@ -86,8 +87,8 @@ newEnv o _dnsResolver = do
   _applog <- LogExt.mkLogger (Opt.logLevel o) (Opt.logNetStrings o) (Opt.logFormat o)
   let _requestId = def
   let _runSettings = Opt.optSettings o
-  let _brig = mkEndpoint (Opt.brig o)
-  let _brigEndpoint = Opt.brig o
+  let _brig Brig = mkEndpoint (Opt.brig o)
+      _brig Galley = mkEndpoint (Opt.galley o)
```
mdimjasevic, line 91: How about calling this something other than `_brig` as it is not about Brig only anymore? Perhaps `_componentEndpoint` or something like that?

services/federator/src/Federator/Service.hs
```diff
@@ -32,30 +30,31 @@ import Federator.Env (brig)
...
-  BrigCall path body domain -> embed @Federator . liftAppIOToFederator $ do
+interpretService = interpret $ \case
+  ServiceCall component path body domain -> embed @Federator . liftAppIOToFederator $ do
     brigReq <- view brig <$> ask
     res <-
-      rpc' "brig" brigReq $
+      rpc' "brig" (brigReq component) $
         RPC.method HTTP.POST
           . RPC.path path -- FUTUREWORK(federation): Protect against arbitrary paths
           . RPC.body (RPC.RequestBodyBS body)
```
mdimjasevic, line 60: I do not understand this part, but I see that renaming/generalizing in the PR from a brig to a service did not change this section much. This is not a Brig-specific request anymore, right? So perhaps `serviceReq` instead of `brigReq`? In the same vein, that `"brig"` string seems overly specific; I am not sure what exactly that argument to `rpc'` represents, but my understanding so far is that it is used only in printing/showing an exception, so how about using something like `show component` instead?

services/federator/test/unit/Test/Federator/ExternalServer.hs
```diff
@@ -48,27 +48,27 @@ tests =
...
       embed $ assertEqual "one call to brig should be made" [expectedCall] actualCalls
       embed $ assertEqual "response should be success with correct body" (InwardResponseBody "response body") res
 
 requestBrigFailure :: TestTree
 requestBrigFailure =
   testCase "should translate response from brig to 'InwardResponseBody' when response has status 404" $
-    runM . evalMock @Brig @IO $ do
-      mockBrigCallReturns @IO (\_ _ _ -> pure (HTTP.notFound404, Just "response body"))
+    runM . evalMock @Service @IO $ do
+      mockServiceCallReturns @IO (\_ _ _ _ -> pure (HTTP.notFound404, Just "response body"))
```
mdimjasevic, line 65: Where does this `mockServiceCallReturns` thing come from? I can't find it anywhere in the codebase.

services/federator/test/unit/Test/Federator/ExternalServer.hs
```diff
@@ -48,27 +48,27 @@ tests =
...
       let request = Request Brig "/users" "\"foo\"" exampleDomain
 
-      res <- mock @Brig @IO . noLogs . Polysemy.runReader allowAllSettings $ callLocal request
+      res <- mock @Service @IO . noLogs . Polysemy.runReader allowAllSettings $ callLocal request
 
-      actualCalls <- mockBrigCallCalls @IO
-      let expectedCall = ("/users", "\"foo\"", aValidDomain)
+      actualCalls <- mockServiceCallCalls @IO
```
mdimjasevic, line 70: Can't find `mockServiceCallCalls` either.

services/federator/test/unit/Test/Federator/ExternalServer.hs
```diff
@@ -36,7 +36,7 @@ import Test.Tasty (TestTree, testGroup)
 import Test.Tasty.HUnit (assertEqual, testCase)
 import Wire.API.Federation.GRPC.Types
 
-genMock ''Brig
+genMock ''Service
 
 tests :: TestTree
 tests =
```
mdimjasevic, line 42: Perhaps at some point it would make sense to add tests for Galley as well. If that makes sense, I'm fine with doing that in a separate PR.

mdimjasevic (changes requested):
Overall this looks good. The only thing I am not sure about is the implementation of `interpretService`. @pcapriotti , can you tell if it needs updating?

services/federator/src/Federator/Run.hs
pcapriotti: Good point, I forgot to name the field.

services/federator/test/unit/Test/Federator/ExternalServer.hs
pcapriotti: It's automatically generated by `genMock`.

services/federator/test/unit/Test/Federator/ExternalServer.hs
pcapriotti: I added one test for galley.

pcapriotti:
Thanks for the review. I changed whatever `brig` name was left into `service`. I also updated the label passed to `rpc` with the actual name of the service, and added a test for galley requests.

services/federator/test/unit/Test/Federator/ExternalServer.hs
```diff
@@ -72,6 +73,19 @@ requestBrigFailure =
...
+requestGalleySuccess :: TestTree
+requestGalleySuccess =
+  testCase "should translate response from galley to 'InwardResponseBody' when response has status 200" $
+    runM . evalMock @Service @IO $ do
+      mockServiceCallReturns @IO (\_ _ _ _ -> pure (HTTP.ok200, Just "response body"))
+      let request = Request Galley "/users" "\"foo\"" exampleDomain
+
+      res :: InwardResponse <- mock @Service @IO . noLogs . Polysemy.runReader allowAllSettings $ callLocal request
+      actualCalls <- mockServiceCallCalls @IO
+      let expectedCall = (Galley, "/users", "\"foo\"", aValidDomain)
```
mdimjasevic, line 85: I don't understand this well, but shouldn't this be about a Galley-related endpoint?

mdimjasevic (approved):
Great!

I left a tiny comment in a Galley test you added.

services/federator/test/unit/Test/Federator/ExternalServer.hs
pcapriotti: I guess it should. It's just a mocked request, so it doesn't matter, but I'll turn it into a galley endpoint for clarity.

akshaymankar (approved):

PR #1573, opened 2021-06-07
wire-api: Introduce ErrorDescription

akshaymankar:
This type will have special instances of HasSwagger so we can express the error
responses of a request using UVerb and they will appear correctly in the swagger
documentation.

### DIFF
--- a/libs/wire-api/src/Wire/API/ErrorDescription.hs
+++ b/libs/wire-api/src/Wire/API/ErrorDescription.hs
@@ -1,0 +1,120 @@
+module Wire.API.ErrorDescription where
+
+import Control.Lens (at, over, (.~), (?~))
+import Control.Lens.Combinators (_Just)
+import qualified Data.Aeson as A
+import qualified Data.HashMap.Strict.InsOrd as InsOrdHashMap
+import Data.Schema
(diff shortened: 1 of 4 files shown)


libs/wire-api/src/Wire/API/ErrorDescription.hs
```diff
@@ -0,0 +1,120 @@
...
+
+instance (KnownNat statusCode, KnownSymbol desc) => ToSchema (ErrorDescription statusCode desc) where
+  schema =
+    object "ErrorDescription" $
+      ErrorDescription
+        <$> label .= field "label" schema
+        <*> message .= field "message" schema
+        <* const (natVal (Proxy @statusCode)) .= field "code" schema
+
+-- | This insance works with 'UVerb' only becaue of the following overlapping
```
pcapriotti, line 35: typo: "insance"

pcapriotti (approved):
Nice!

libs/wire-api/src/Wire/API/ErrorDescription.hs
akshaymankar:
```suggestion
-- | This instance works with 'UVerb' only becaue of the following overlapping
```

PR #1575, opened 2021-06-07
minor refactor: runFederated

jschaul:


services/galley/src/Galley/API/Util.hs
```diff
@@ -320,12 +322,22 @@ checkRemoteUsersExist =
...
     throwM unknownRemoteUser
+
+type FederatedGalleyRPC c a = FederatorClient c (ExceptT FederationClientFailure Galley) a
+
+runFederatedGalley :: Domain -> FederatedGalleyRPC 'Galley a -> Galley a
+runFederatedGalley remoteDomain rpc = do
+  runExceptT (executeFederated remoteDomain rpc)
+    >>= either (throwM . federationErrorToWai) pure
+
+runFederatedBrig :: Domain -> FederatedGalleyRPC 'Brig a -> Galley a
```
pcapriotti, line 340: I would maybe consider having a single `runFederated` function here with a `c :: Component` type parameter, and use it as `runFederated @'Brig`. But this is ok too.

pcapriotti (approved):
Looks good. Just one minor comment.

services/galley/src/Galley/API/Util.hs
jschaul: I made a single function, however I can't use it with the `@`syntax. So I left the helper functions runFederatedBrig and runFederatedGalley in there. It seems to be able to resolve the types from the function signature, but if I try to use `runFederated @'Brig` or `runFederated @Brig` it tells me that

```
expected a type, but 'Brig has kind Component
expected a type, but Brig has kind Component
```

I'm a bit fuzzy on these custom `kind` types, especially as they are automagically generated by mu-haskell, there isn't any code definitions I can look at.

Probably there is an obscure language pragma that needs to be enabled to use a particular syntax to call that function. At this point I think it's better to keep regular helper functions around, rather than leave the next person guessing how to actually call runFederated with the Brig component.

pcapriotti: I think it's good already as it is, but for future reference, you can make `c` the first type argument, and then type application would work:
```haskell
runFederated :: forall (c :: Component) a ...
```

jschaul: Aha! That's not obvious, thanks!

PR #1581, opened 2021-06-09
Remove resource requests

pcapriotti:
If I understand correctly, these are only used in tests, so there should be no danger in disabling them. With those requests on, I cannot run two clusters on my machine, so I always have to apply this patch locally, which is inconvenient.


jschaul (approved):
Your assumption is correct, this is only used during tests on CI.

Possibly if too many clusters are running, something will break in a weird way when there isn't enough CPU or memory available, but we can diagnose and possibly set better resource requests following that (the current resource requests are very ad-hoc and not based on any data)

PR #1582, opened 2021-06-09
Import http2 fix

pcapriotti:
This should fix the issue we were experiencing with end2end tests randomly failing on CI. The underlying problem is that the nginx controller proxying the federator uses an HTTP2 feature which was implemented incorrectly in the http2 library, which we indirectly depend on. See https://github.com/kazu-yamamoto/http2/issues/27.

This imports a [fix from our fork](https://github.com/wireapp/http2/tree/header-encoding-0-size-table-backport) into our stackage snapshot.

### DIFF
--- a/stack.yaml
+++ b/stack.yaml
@@ -230,7 +230,8 @@ extra-deps:
     - http2-client-grpc
 
 # Fix in for issue #27: https://github.com/kazu-yamamoto/http2/issues/27
-# Note: the patch is based on version 2.0.6 of http2
+# PR here: https://github.com/kazu-yamamoto/http2/pull/28
+# Note: the commit used here is based on version 2.0.6 of http2
 - git: https://github.com/wireapp/http2 # (2021-06-09) branch: header-encoding-0-size-table-backport
   commit: 7c465be1201e0945b106f7cc6176ac1b1193be13
 
--- a/stack.yaml
+++ b/stack.yaml
@@ -229,6 +229,11 @@ extra-deps:
   subdirs:
     - http2-client-grpc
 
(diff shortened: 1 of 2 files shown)


stack.yaml
```diff
@@ -229,6 +229,11 @@ extra-deps:
   subdirs:
     - http2-client-grpc
 
+# Fix in branch header-encoding-0-size-table-backport for issue #27
+# Note: the patch for #27 is based on version 2.0.6 of http2
```
akshaymankar, line 233: Can you please mention the branch name here?

pcapriotti: You mean instead of the SHA?

akshaymankar: No, you have to write sha in there. But we usually also write the branch name in the comment so if our fork needs any maintenance it is easy to figure out where to do that maintenance.

akshaymankar: For eample: https://github.com/wireapp/wire-server/blob/3ad1409258f04ce4563023924e508ee20d878908/stack.yaml#L92

jschaul (commented):
Could you in the comment add a link to the issue, and could you also maybe make a PR upstream (and reference that one here in the comments, too)?

stack.yaml
pcapriotti: The branch was already in the comment, but I've reformatted to look like your example.

akshaymankar: Ah, silly of me!

pcapriotti:
> Could you in the comment add a link to the issue, and could you also maybe make a PR upstream (and reference that one here in the comments, too)?

Done: https://github.com/kazu-yamamoto/http2/pull/28

jschaul (approved):

PR #1584, opened 2021-06-09
Servantify own client API

pcapriotti:
This just converts a few more endpoints to servant and introduces the corresponding schema instances.

### DIFF
--- a/libs/wire-api/src/Wire/API/Routes/Public/Brig.hs
+++ b/libs/wire-api/src/Wire/API/Routes/Public/Brig.hs
@@ -316,6 +316,13 @@ data Api routes = Api
         :> "clients"
         :> CaptureClientId "client"
         :> UVerb 'GET '[JSON] GetClientResponse,
+    getClientCapabilities ::
+      routes :- Summary "Read back what the client has been posting about itself"
+        :> ZUser
+        :> "clients"
+        :> CaptureClientId "client"
+        :> "capabilities"
+        :> Get '[JSON] ClientCapabilityList,
     searchContacts ::
       routes :- Summary "Search for users"
         :> ZUser
--- a/libs/wire-api/src/Wire/API/Routes/Public/Brig.hs
+++ b/libs/wire-api/src/Wire/API/Routes/Public/Brig.hs
@@ -303,6 +303,11 @@ data Api routes = Api
         :> CaptureClientId "client"
         :> ReqBody '[JSON] RmClient
         :> Delete '[] (EmptyResult 200),
+    listClients ::
+      routes :- Summary "List the registered clients"
+        :> ZUser
+        :> "clients"
+        :> Get '[JSON] [Client],
     searchContacts ::
       routes :- Summary "Search for users"
         :> ZUser
--- a/libs/wire-api/src/Wire/API/Routes/Public/Brig.hs
+++ b/libs/wire-api/src/Wire/API/Routes/Public/Brig.hs
@@ -49,7 +49,9 @@ type CaptureUserId name = Capture' '[Description "User Id"] name UserId
 
 type CaptureClientId name = Capture' '[Description "ClientId"] name ClientId
 
-type ClientResponse = Headers '[Header "Location" ClientId] Client
+type NewClientResponse = Headers '[Header "Location" ClientId] Client
+
+type GetClientResponse = [WithStatus 200 Client, Empty404]
 
 data Api routes = Api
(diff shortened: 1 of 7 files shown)


libs/wire-api/src/Wire/API/Routes/Public/Brig.hs
```diff
@@ -268,6 +274,66 @@ data Api routes = Api
         :> "list-prekeys"
         :> ReqBody '[JSON] QualifiedUserClients
         :> Post '[JSON] QualifiedUserClientPrekeyMap,
+    -- User Client API ----------------------------------------------------
+    -- TODO: another one?
```
akshaymankar, line 278: Forgotten TODO?

libs/wire-api/src/Wire/API/User/Client.hs
```diff
@@ -409,6 +411,20 @@ data Client = Client
...
+      Client
+        <$> clientId .= field "id" schema
+        <*> clientType .= field "id" schema
+        <*> clientTime .= field "id" schema
+        <*> clientClass .= opt (field "id" schema)
+        <*> clientLabel .= opt (field "id" schema)
+        <*> clientCookie .= opt (field "id" schema)
+        <*> clientLocation .= opt (field "id" schema)
+        <*> clientModel .= opt (field "id" schema)
+        <*> clientCapabilities .= (field "id" schema <|> pure mempty)
```
akshaymankar, line 426: Field names look wrong. This also means there is no Swagger test for this.

services/brig/src/Brig/API/Public.hs
```diff
@@ -868,51 +783,36 @@ getMultiUserPrekeyBundleH zusr qualUserClients = do
...
-getClientH (zusr ::: clt ::: _) =
-  getClient zusr clt <&> \case
-    Just c -> json c
-    Nothing -> setStatus status404 empty
+getClient :: UserId -> ClientId -> Handler (Union BrigAPI.GetClientResponse)
+getClient zusr clientId = do
+  mc <- lift $ API.lookupLocalClient zusr clientId
+  case mc of
+    Nothing -> Servant.respond ErrorDescription.clientNotFound
+    Just c -> Servant.respond (WithStatus @200 c)
```
akshaymankar, line 815: Does this mean we used to return empty body earlier and now we return the error description? I hope this doesn't break anything :crossed_fingers:

akshaymankar (changes requested):

libs/wire-api/src/Wire/API/Routes/Public/Brig.hs
pcapriotti: That was already there, so I copy pasted it back. I'm not sure what it means.

libs/wire-api/src/Wire/API/User/Client.hs
pcapriotti: :facepalm: I'll fix it. Thanks for spotting it.

services/brig/src/Brig/API/Public.hs
pcapriotti: Yeah. I'm not sure how else to handle this, since `Servant` really does not allow an empty body unless the list of content types is empty, and we cannot have an empty list of content types, because we return a non-empty body in one of the cases.

akshaymankar: Hmm, I think it is a mistake to set and empty body in this case anyway and the likelihood of this breaking anything is small, so maybe it is ok. Unfortunate that we'll have to wait for a merge to know if this actually breaks anything. Maybe we can ask client devs, if they think this will break stuff?

libs/wire-api/src/Wire/API/Routes/Public/Brig.hs
akshaymankar: Looks like it sneaked in while previous refactoring: https://github.com/wireapp/wire-server/commit/37039b64291ac77be2666f6b74a1bcab39ce33cb#diff-4652702c562703881b30f146d018046eb3d53b00c4c1d94ab753943e9d54bc6aR503

Let's remove it.

libs/wire-api/src/Wire/API/User/Client.hs
pcapriotti: Actually, there are tests, but the old JSON instances were still in place, otherwise they would have indeed caught this. Fixed now.

akshaymankar (approved):

PR #1585, opened 2021-06-09
end2end test getting remote conversation and complete its implementation

jschaul:
* fixes the implementation of getting a conversation by ID: the conversationView and getConversationRPC call was incomplete / wrong.
* also fixes one futurework in one brig integration test

This should have been part of #1566

### DIFF
--- a/services/brig/test/integration/Federation/End2end.hs
+++ b/services/brig/test/integration/Federation/End2end.hs
@@ -20,7 +20,7 @@ module Federation.End2end where
 import API.Search.Util
 import API.User.Util (getUserClientsQualified)
 import Bilge
-import Bilge.Assert ((!!!), (===))
+import Bilge.Assert ((!!!), (<!!), (===))
 import Brig.API.Client (pubClient)
 import qualified Brig.Options as BrigOpts
 import Brig.Types
@@ -30,7 +30,7 @@ import qualified Data.Aeson as Aeson
 import Data.ByteString.Conversion (toByteString')
 import Data.Domain (Domain)
 import Data.Handle
-import Data.Id (ClientId)
+import Data.Id (ClientId, ConvId, UserId)
 import Data.List.NonEmpty (NonEmpty ((:|)))
 import qualified Data.Map as Map
 import Data.Qualified
@@ -59,8 +59,8 @@ import Wire.API.User.Client (QualifiedUserClients (..), mkQualifiedUserClientPre
 -- - Remote discovery succeeds but server doesn't exist
 -- - Remote federator fails to respond in many ways (protocol error, timeout, etc.)
 -- - SRV record has two servers but higher priority one always fails
-spec :: BrigOpts.Opts -> Manager -> Brig -> Galley -> Endpoint -> Brig -> IO TestTree
-spec _brigOpts mg brig galley _federator brigTwo =
+spec :: BrigOpts.Opts -> Manager -> Brig -> Galley -> Endpoint -> Brig -> Galley -> IO TestTree
+spec _brigOpts mg brig galley _federator brigTwo galleyTwo =
   pure $
     testGroup
       "federation-end2end-user"
@@ -71,7 +71,7 @@ spec _brigOpts mg brig galley _federator brigTwo =
         test mg "claim prekey bundle" $ testClaimPrekeyBundleSuccess brig brigTwo,
         test mg "claim multi-prekey bundle" $ testClaimMultiPrekeyBundleSuccess brig brigTwo,
         test mg "list user clients" $ testListUserClients brig brigTwo,
-        test mg "add remote users to local conversation" $ testAddRemoteUsersToLocalConv brig galley brigTwo
+        test mg "add remote users to local conversation" $ testAddRemoteUsersToLocalConv brig galley brigTwo galleyTwo
       ]
 
 -- | Path covered by this test:
@@ -210,8 +210,8 @@ testClaimMultiPrekeyBundleSuccess brig1 brig2 = do
       const 200 === statusCode
       const (Just ucm) === responseJsonMaybe
 
-testAddRemoteUsersToLocalConv :: Brig -> Galley -> Brig -> Http ()
-testAddRemoteUsersToLocalConv brig1 galley1 brig2 = do
+testAddRemoteUsersToLocalConv :: Brig -> Galley -> Brig -> Galley -> Http ()
+testAddRemoteUsersToLocalConv brig1 galley1 brig2 galley2 = do
   alice <- randomUser brig1
   bob <- randomUser brig2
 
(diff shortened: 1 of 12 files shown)


pcapriotti (approved):
Looks good.

services/brig/test/integration/Federation/End2end.hs
```diff
@@ -238,6 +239,29 @@ testAddRemoteUsersToLocalConv brig1 galley1 brig2 = do
...
+      let backend1Domain = qDomain (userQualifiedId alice)
+          -- FUTUREWORK add qualified conversation Id to Conversation data type, then use that from the conversation creation response
+          qualifiedConvId = Qualified convId backend1Domain
+      res' <- getConvQualified galley (userId bob) qualifiedConvId <!! (const 200 === statusCode)
+      let conv' = responseJsonUnsafeWithMsg (backend <> "- get /conversations/domain/cnvId") res'
+          actual' = cmOthers $ cnvMembers conv'
+          expected' = [OtherMember (userQualifiedId bob) Nothing roleNameWireAdmin]
+      liftIO $ actual' @?= expected'
+
+    getConvQualified :: Galley -> UserId -> Qualified ConvId -> Http ResponseLBS
```
mdimjasevic, line 256: This is a generally useful utility function so how about making it a top-level exported function in a `Util` module?

services/brig/test/integration/Federation/End2end.hs
```diff
@@ -238,6 +239,29 @@ testAddRemoteUsersToLocalConv brig1 galley1 brig2 = do
     )
     !!! (const 200 === statusCode)
 
+  -- test GET /conversations/:backend1Domain/:cnv
+  testQualifiedGetConversation galley1 "galley1" alice bob convId
+  testQualifiedGetConversation galley2 "galley2" alice bob convId
```
mdimjasevic, line 244: Shouldn't the second call to `testQualifiedGetConversation` have `alice` and `bob` swapped when compared to the other call to the same function? I would expect so, because I would expect a list of `OtherMember`s to be dependent on the beholder.

services/brig/test/integration/Federation/End2end.hs
```diff
@@ -210,12 +210,12 @@ testClaimMultiPrekeyBundleSuccess brig1 brig2 = do
...
+testAddRemoteUsersToLocalConv :: Brig -> Galley -> Brig -> Galley -> Http ()
+testAddRemoteUsersToLocalConv brig1 galley1 brig2 galley2 = do
   alice <- randomUser brig1
   bob <- randomUser brig2
 
-  let conv = NewConvUnmanaged $ NewConv [] [] (Just "gossip") mempty Nothing Nothing Nothing Nothing roleNameWireAdmin
+  let newConv = NewConvUnmanaged $ NewConv [] [] (Just "gossip") mempty Nothing Nothing Nothing Nothing roleNameWireAdmin
   convId <-
     cnvId . responseJsonUnsafe
       <$> post
```
mdimjasevic, line 221: I was about to propose to use `postConvWithRoleQualified` instead, but then I am not sure how it would fit the `Http` monad and even more importantly, how to make sure `galley1` is used, and not `galley2`. What's perhaps even more important is that this function is located in Galley integration tests.

services/brig/test/integration/Federation/End2end.hs
mdimjasevic: Locally I switched `alice` and `bob` in line 244 and the test still passes, which suggests the test needs improvement.

services/brig/test/integration/Federation/End2end.hs
mdimjasevic: I based my PR on this PR, which is where I already made the change I proposed here. See https://github.com/wireapp/wire-server/pull/1594/files.

jschaul: thanks, I backported your change.

services/brig/test/integration/Federation/End2end.hs
jschaul: yes, good point.

services/brig/test/integration/Federation/End2end.hs
jschaul: At this point we probably want to duplicate some of these helper functions; but I agree it would be nice to consolidate things more. If/when we merge galley+brig+spar, that will become a non-issue.

services/galley/test/unit/Test/Galley/Mapping.hs
```diff
@@ -0,0 +1,209 @@
...
+
+  assertEqual "self:" expectedSelf (cmSelf <$> actual)
+  assertEqual "others:" expectedOthers (cmOthers <$> actual)
+
+runMappingNotAMemberA :: HasCallStack => IO ()
+runMappingNotAMemberA = do
+  let convDomain = Domain "backendA.example.com"
+  let aliceDomain = Domain "backendA.example.com"
+  alice <- flip Qualified aliceDomain <$> randomId
+  requester <- flip Qualified aliceDomain <$> randomId
```
mdimjasevic, line 73: An unstated assumption here is that `alice` and `requester` are two different (random) values. Sure, it is highly unlikely to happen, but there is a possibility of ending up with the same values here.

Given that randomness doesn't buy us anything in this case, and on the other hand can lead to trouble as I just pointed out, I'd simply use fixed, distinct and specific IDs for `alice` and `requester`.

services/galley/test/unit/Test/Galley/Mapping.hs
```diff
@@ -0,0 +1,209 @@
...
+
+mkOtherMember :: Qualified UserId -> OtherMember
+mkOtherMember u = OtherMember u Nothing roleNameWireAdmin
+
+mkRemoteMember :: Qualified UserId -> RemoteMember
+mkRemoteMember u = I.RemoteMember (toRemote u) roleNameWireAdmin
+
+mkInternalConv :: [LocalMember] -> [RemoteMember] -> IO Data.Conversation
+mkInternalConv locals remotes = do
+  creator <- randomId
```
mdimjasevic, line 163: Shouldn't the creator always be part of the conversation? I'd always expect the creator to be on the list of local members (I suppose one cannot create a conversation on a remote backend). Furthermore, given that it is generated at random, we know nothing how it relates to IDs in `locals`, e.g., maybe it is equal to an ID in the list.

Even if the implicit idea behind `Data.Conversation` is that `Data.convCreator` is always a local member (which I'd have to check), I'd turn the creator into an input to `mkInternalConv` to make it more visible at use sites that there is another member of the conversation.

services/galley/test/unit/Test/Galley/Mapping.hs
jschaul: I disagree here, we use UUIDs for a reason - because we believe the chances for collision are so small that we accept that risk. We do this everywhere in production code, so it doesn't make sense to me to be more careful in unit test than in production code.
There is a specific test further down testing the UUID-collision case between alice and bob from two different backends: This is arguably more likely as it can be maliciously crafted this way by a backend that is not well-behaved.

services/galley/test/unit/Test/Galley/Mapping.hs
jschaul: The creator has no role in the unit test I'm try to do here, which is about the conversationView. So any value is fine.

A creator often is part of the localMembers; yes; but in theory a creator can create a conversation, then leave the conversation, leading to a situation where he/she is no longer part of that conversation. I'll add a comment as it may be confusing for the reader.

services/galley/src/Galley/API/Mapping.hs
```diff
@@ -81,6 +93,25 @@ conversationViewMaybe u Data.Conversation {..} = do
...
+          (qUnqualified (Public.omQualifiedId m))
+          Nothing
+          False
+          Nothing
+          Nothing
+          False
+          Nothing
+          False
+          Nothing
+          (Public.omConvRoleName m)
```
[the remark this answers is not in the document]
jschaul: Good suggestion! Will do.

services/galley/src/Galley/API/Mapping.hs
```diff
@@ -81,6 +93,25 @@ conversationViewMaybe u Data.Conversation {..} = do
           Public.omConvRoleName = Internal.rmConvRoleName x
         }
 
+    -- FUTUREWORK(federation): we currently don't store muted, archived etc status for users who are on a different backend than a conversation
+    -- but we should. Once this information is available, the code should be changed to use the stored information, rather than these defaults.
+    incompleteSelfMember :: [Public.OtherMember] -> Maybe Public.Member
```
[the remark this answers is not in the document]
jschaul: Good point. Now that I have some unit tests, refactoring the function has become easier in a way. I made this function simply OtherMember -> Member now.

services/brig/test/integration/Main.hs
```diff
@@ -72,6 +73,7 @@ instance FromJSON BackendConf where
   parseJSON = withObject "BackendConf" $ \o ->
     BackendConf
       <$> o .: "brig"
+      <*> o .: "galley"
       <*> o .: "federatorInternal"
       <*> o .: "federatorExternal"
```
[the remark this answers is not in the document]
jschaul: Good idea, I will do this in a follow-up PR

mdimjasevic (approved):

PR #1588, opened 2021-06-10
Reorganise helper functions for conversation notifications

mdimjasevic:
- Break up addToConversation to its constituents
- Move Galley helper functions to a Util module

### DIFF
--- a/services/galley/src/Galley/API/Update.hs
+++ b/services/galley/src/Galley/API/Update.hs
@@ -62,25 +62,20 @@ where
 
 import Brig.Types.Intra (accountUser)
 import qualified Brig.Types.User as User
-import Control.Arrow ((&&&))
 import Control.Lens
 import Control.Monad.Catch
-import Control.Monad.Except (runExceptT)
 import Control.Monad.State
 import Data.ByteString.Conversion (toByteString')
 import Data.Code
-import Data.Domain (Domain)
 import Data.Id
 import Data.Json.Util (toUTCTimeMillis)
 import Data.LegalHold (UserLegalHoldStatus (UserLegalHoldNoConsent), defUserLegalHoldStatus)
-import Data.List.Extra (nubOrd, nubOrdOn)
 import Data.List1
 import qualified Data.Map.Strict as Map
 import Data.Misc (FutureWork (..))
 import Data.Qualified
 import Data.Range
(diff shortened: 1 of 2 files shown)


services/galley/src/Galley/API/Update.hs
```diff
@@ -885,60 +876,32 @@ rmBot zusr zcon b = do
...
-    . Map.assocs
-    . partitionQualified
-    . nubOrd
-    . map (unTagged . rmId)
-    $ rmm <> rothers
-  let allMembers = nubOrdOn memId (lmm <> lothers)
-  for_ (newPush ListComplete usr (ConvEvent e) (recipient <$> allMembers)) $ \p ->
-    push1 $ p & pushConn ?~ conn
-  void . forkIO $ void $ External.deliver (bots `zip` repeat e)
+  notifyRemotes existingRemotes usr now c lmm rmm
```
pcapriotti, line 902: Note that `notifyRemotes` not only results in remote users getting notifications, but also changes state on the remote end in the form of database updates.

I think the naming of these two functions should reflect the fact that they have sightly different side effects, and their difference is not only in terms of operating locally vs remotely.

services/galley/src/Galley/API/Util.hs
```diff
@@ -342,3 +346,57 @@ runFederated :: forall (c :: Component) a. Domain -> FederatorClient c (ExceptT
...
+          { cmuTime = now,
+            cmuOrigUserId = orig,
+            cmuConvId = cnv,
+            cmuAlreadyPresentUsers = others,
+            cmuUsersAdd = uids,
+            cmuUsersRemove = []
+          }
+  let rpc = FederatedGalley.updateConversationMemberships FederatedGalley.clientRoutes cmu
+  runExceptT (Federation.executeFederated domain rpc)
+    >>= either (throwM . federationErrorToWai) pure
```
pcapriotti, line 402: While you're at it, could you make use of the `runFederated` convenience function defined just above?

pcapriotti (approved):
Looks good. Minor comments follow.

services/galley/src/Galley/API/Util.hs
mdimjasevic: I implemented this.

PR #1589, opened 2021-06-10
Merge Empty200, Empty404 and EmptyResult

pcapriotti:
There is a slight complication here due to the fact that `Empty200` and `Empty404` are used in a `UVerb`, and `UVerb` does not really support the no content type case.

The solution in this patch is to define an overlapping instance for `HasServer UVerb` for an empty list of content types.

Also, the combinators in `ErrorDescription` (taken from `servant-swagger`) have been moved to a new module and reused for the
definition of the `Swagger` instance for a `UVerb` contaning empty results.

### DIFF
--- a/libs/wire-api/package.yaml
+++ b/libs/wire-api/package.yaml
@@ -42,6 +42,7 @@ library:
   - hscim
   - http-api-data
   - http-media
+  - http-types
   - insert-ordered-containers
   - iproute >=1.5
   - iso3166-country-codes >=0.2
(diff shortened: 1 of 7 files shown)


akshaymankar (approved):
I reviewed 57de31f, it looks good.

I think the "Squash and merge" button will do something wrong if you try to do it now, can you please rebase this onto develop and then merge it?

pcapriotti:
> I reviewed [57de31f](https://github.com/wireapp/wire-server/commit/57de31f06d4ce17297742ead628c965a6e819517), it looks good.
> 
> I think the "Squash and merge" button will do something wrong if you try to do it now, can you please rebase this onto develop and then merge it?

Done, I'll wait for CI to go green again and then merge.

PR #1591, opened 2021-06-10
Federation: POST /list-conversations endpoint

jschaul:
This implements a new endpoint `POST /list-conversations` similar to the existing `GET /conversations` but allows specifying parameters, such as a list of remote conversations in its request body. It will also, when called with an empty body, return a user's remote conversations.
This is part of https://wearezeta.atlassian.net/browse/SQCORE-134

## todos:

- [x] implement local conversations behaviour
- [x] same test coverage than GET /conversations
- [x] actually make remote calls
- [x] mockFederator test
- [x] end2end test

## Checklist

 - [x] The description provides context as to why the change should occur and
       what the code contributes to that effect. This could also be a link to a
       JIRA ticket or a Github issue, if there is one.
 - [x] If a component uses a new or changed internal endpoint of another
       component, this is mentioned in the CHANGELOG.md.
 - [x] The CHANGELOG.md file in the *Unreleased* section has been updated to
       explain the change which will be included in the release notes.


libs/wire-api/src/Wire/API/Routes/Public/Galley.hs
```diff
@@ -136,6 +136,13 @@ data Api routes = Api
              "size"
              (Range 1 500 Int32)
         :> Get '[Servant.JSON] (Public.ConversationList Public.Conversation),
+    listConversations ::
+      routes
+        :- Summary "Get all conversations (also returns remote conversations)"
+        :> ZUser
+        :> "list-conversations"
+        :> ReqBody '[Servant.JSON] Public.ListConversations
+        :> Post '[Servant.JSON] (Public.ConversationList Public.Conversation),
```
jschaul, line 145: I'd like to get a review on the endpoint naming here: Is `POST /list-conversations` a good name or would a different route be preferred?

jschaul: Before I make all the changes in nginz configs, perhaps I could get an early feedback on this question @akshaymankar ?

akshaymankar: This is consistent with `list-users`, `list-clients` and `list-prekeys`, so I like it :100:

libs/wire-api/src/Wire/API/Conversation.hs
```diff
@@ -247,6 +249,26 @@ instance FromJSON a => FromJSON (ConversationList a) where
...
+
+instance ToSchema ListConversations where
+  schema =
+    objectWithDocModifier
+      "ListConversations"
+      (description ?~ "A request to list some or all of a user's conversations, including remote ones")
+      $ ListConversations
+        <$> lax (lQualifiedIds .= opt (field "qualified_ids" (nonEmptyArray schema)))
+        <*> lax (lStartId .= opt (field "start_id" schema))
+        <*> lax (lSize .= opt (field "size" schema))
```
pcapriotti, line 270: No need to use `lax` here. `lax` is important when `opt` / `optWithDefault` is used in the inner schema. If you use `opt` on the `field` schema itself, it already accepts a missing field, so `lax` does not do anything.

In fact, after #1621, it is better to just use `optField` here, as in:
```haskell
lQualifiedIds .= optField "qualified_ids" (nonEmptyArray schema)
```
and similarly for the other ones. The difference is that if we get as input a JSON like:
```json
{ "qualified_ids": ["invalid"] }
```
with `opt` we just get `Nothing` for that field, while with `optField` we get an error.

services/galley/src/Galley/API/Query.hs
```diff
@@ -90,19 +93,24 @@ getConversation zusr domain cnv = do
...
-getRemoteConversation zusr (Qualified convId remoteDomain) = do
-  localDomain <- viewFederationDomain
-  let qualifiedZUser = Qualified zusr localDomain
-      req = FederatedGalley.GetConversationsRequest qualifiedZUser [convId]
-      rpc = FederatedGalley.getConversations FederatedGalley.clientRoutes req
-  -- we expect the remote galley to make adequate checks on conversation
-  -- membership and here we just pass through the reponse
-  conversations <- runFederatedGalley remoteDomain rpc
-  case gcresConvs conversations of
+getRemoteConversation zusr qConvId = do
```
pcapriotti, line 96: Maybe use `Remote ConvId` instead of `Qualified UserId`, since we know at this point that `qConvId` is definitely remote.

services/galley/test/integration/API.hs
```diff
@@ -843,6 +846,14 @@ getConvsOk = do
     const 200 === statusCode
     const [toUUID usr] === map (toUUID . cnvId) . decodeConvList
 
+-- same test than getConvsOk, but using the listConversations endpoint
```
pcapriotti, line 849:
```suggestion
-- same test as getConvsOk, but using the listConversations endpoint
```

also below

pcapriotti (approved):
Looks good! Minor comments follow.

EDIT: the end2end test is failing, though. 🤔

libs/wire-api/src/Wire/API/Conversation.hs
jschaul: Thanks! I changed this to use optField. Have you come across places where we don't want to omit the field when generating json though? Otherwise the extra argument to optField (Nothing / Aeson.null in the usages so far) could be simplified to always use Nothing.

services/galley/src/Galley/API/Query.hs
jschaul: 👍🏿

PR #1592, opened 2021-06-10
Add endpoint: Get name, id with for CodeAccess conversations

smatting:
This PR adds the endpoint
`GET /conversations/join?code=xx&value=yy`
It can be used to retrieve the name and id of the conversation that would be joined with `POST /conversation/join` with the same code and value. The endpoint requires the same authorization as the `POST` endpoint.

https://wearezeta.atlassian.net/browse/SQSERVICES-521


libs/wire-api/src/Wire/API/Routes/Public/Galley.hs
```diff
@@ -140,6 +142,16 @@ data Api routes = Api
...
     --             errorResponse Error.notATeamMember
     --             errorResponse (Error.operationDenied Public.CreateConversation)
+    getConversationByReusableCode ::
+      routes
+        :- Summary "Get limited conversation information by key/code pair"
+        :> ZUser
+        :> "conversations"
+        :> "join"
+        :> QueryParam' [Required, Strict] "code" Code.Key
+        :> QueryParam' [Required, Strict] "value" Code.Value
```
Yserz, line 152: Is it possible to change `value` to `key` to keep it in sync with `POST /conversations/join`?

fisx: wait, which one do you mean?

1.
```haskell
        :> QueryParam' [Required, Strict] "code" Code.Key
        :> QueryParam' [Required, Strict] "key" Code.Value
```

2.
```haskell
        :> QueryParam' [Required, Strict] "key" Code.Key
        :> QueryParam' [Required, Strict] "value" Code.Value
```

fisx: I'm wondering if there is a bug here, and it should be:

```haskell
        :> QueryParam' [Required, Strict] "key" Code.Key
        :> QueryParam' [Required, Strict] "code" Code.Value
```

but that's about the haskell types, more of a question to myself.

fisx: I went for the latter: 06900624c4e4aef3faa665774c297c27474859ef

This seems to be consistent with the [existing endpoints](https://github.com/wireapp/wire-server/blob/dde3eb6b550fb29c79757807adedd563fb7d5765/libs/wire-api/src/Wire/API/Conversation/Code.hs#L48-L52).

fisx (approved):

PR #1593, opened 2021-06-10
Add qualified endpoint for sending messages

akshaymankar:
Highlights:
- Add wireapp/generic-message-proto as a submodule
- Expose only Proto endpoint for sending messages to qualified conversations

Todo:
- [x] Add new integration tests for the qualified endpoint
- [x] Implement sending messages to local users, ignore remotes (to be done in a
follow up PR)
- [x] Update changelog


pcapriotti (approved):
Looks good to me.

libs/wire-api/src/Wire/API/Routes/Public/Galley.hs
```diff
@@ -247,13 +254,26 @@ data Api routes = Api
         :> "otr"
         :> "messages"
         :> ReqBody '[Servant.JSON, Proto] Public.NewOtrMessage
+        :> UVerb 'POST '[Servant.JSON] PostOtrResponsesUnqualified,
+    postOtrMessage ::
+      routes
+        :- Summary "Post an encrypted message to a conversation (accepts JSON or Protobuf)"
```
pcapriotti, line 260: The description looks wrong, since this endpoint does not accept JSON.

(observed by @mdimjasevic)

akshaymankar, line 260:
```suggestion
        :- Summary "Post an encrypted message to a conversation (accepts only Protobuf)"
```

PR #1594, opened 2021-06-11
Federation: Notify Remote Users of Being Added to a New Conversation

mdimjasevic:
This PR introduces:

* a gRPC call to a remote Galley to update its own view of a conversation on another backend
* a test that makes sure that the local and remote Galley have the same view of conversation membership

Note 1: This depends on PR #1585.
Note 2: I am not confident in the test's correctness. I based it on @jschaul's similar test.


services/brig/test/integration/Federation/Util.hs
```diff
@@ -108,3 +113,23 @@ assertRight = \case
 
 assertRightT :: (MonadIO m, Show a, HasCallStack) => ExceptT a m b -> m b
 assertRightT = assertRight <=< runExceptT
+
+testQualifiedGetConversation :: Galley -> String -> User -> User -> ConvId -> Http ()
```
pcapriotti, line 117: This utility function is a bit confusing. Could you document what its arguments are? For example, it seems that `alice` is only used for its domain.

services/galley/src/Galley/API/Create.hs
```diff
@@ -327,6 +327,8 @@ notifyCreatedConversation dtime usr conn c = do
   localDomain <- viewFederationDomain
   now <- maybe (liftIO getCurrentTime) pure dtime
   pushSome =<< mapM (toPush localDomain now) (Data.convLocalMembers c)
+  -- Notify remote users of being added to a conversation
+  updateRemoteConversationMemberships [] usr now c [] (Data.convRemoteMembers c)
```
pcapriotti, line 331: This doesn't look right. On the remote end, users will get a `MemberJoin` notification, instead of the expected `ConvCreate`. I think you should either make a new RPC for the new conversation case, or generalise this to be able to send the other kind of event.

pcapriotti (changes requested):
I'm not sure this is behaving correctly in terms of events. See comment below.

libs/wire-api-federation/src/Wire/API/Federation/API/Galley.hs
```diff
@@ -38,7 +41,15 @@ import Wire.API.Federation.Util.Aeson (CustomEncoded (CustomEncoded))
 -- for the current list we need.
 
 data Api routes = Api
-  { getConversations ::
+  { -- | Create a new conversation
+    createConversation ::
+      routes
+        :- "federation"
+        :> Summary "Create a new conversation"
```
pcapriotti, line 48: This description is a bit misleading. This RPC does not really create a conversation, it just registers local users as part of one.

libs/wire-api-federation/src/Wire/API/Federation/API/Galley.hs
```diff
@@ -70,6 +81,30 @@ newtype GetConversationsResponse = GetConversationsResponse
...
+  { -- | The time when the conversation was created
+    ccTime :: UTCTime,
+    -- | The user that created the conversation
+    ccOrigUserId :: Qualified UserId,
+    -- | The qualified conversation ID
+    ccCnvId :: Qualified ConvId,
+    -- | The conversation type
+    ccCnvType :: ConvType,
+    -- | The user that created the conversation
+    ccCnvCreator :: Qualified UserId,
```
pcapriotti, line 95: What's the difference between `ccOrigUserId` and `ccCnvCreator`?

libs/wire-api/src/Wire/API/Conversation.hs
```diff
@@ -99,11 +99,15 @@ import Wire.API.Conversation.Role (RoleName, roleNameWireAdmin)
 data Conversation = Conversation
   { cnvId :: ConvId,
     cnvType :: ConvType,
+    -- FUTUREWORK: Make this a qualified user ID. This will break
+    -- backwards-compatibility with clients.
```
pcapriotti, line 103: Why should this break backwards-compatibility? This is an output-only type, if I'm not mistaken, so we can simply keep producing the unqualified creator id alongside the qualified one for backwards compatibility, like we are already doing for a few other types (e.g. `Event`).

pcapriotti (commented):
This looks ok, but maybe it would have been slightly nicer to add a `Maybe CreateConversation` field to the `updateConversationMemberships` RPC, since they basically do the same thing, and the only difference is the events they send out.

I'm also not sure about the name `CreateConversation`, which sounds like an action, when it should be a noun. Maybe change it to `CreatedConversation`?

By the way, what is wrong with simply reusing the `Conversation` type from `wire-api` instead of introducing a new one?

pcapriotti (approved):

libs/wire-api-federation/src/Wire/API/Federation/API/Galley.hs
mdimjasevic: I updated this. Thanks!

libs/wire-api-federation/src/Wire/API/Federation/API/Galley.hs
mdimjasevic: Per our call, I removed one of them as they're the same.

libs/wire-api/src/Wire/API/Conversation.hs
mdimjasevic: Right, I updated the comment accordingly.

services/galley/src/Galley/API/Util.hs
```diff
@@ -370,7 +374,140 @@ runFederated remoteDomain rpc = do
...
+        Member.Member
+          { memId = qUnqualified . unTagged $ rmId,
+            memService = Nothing,
+            memOtrMuted = False,
+            memOtrMutedStatus = Nothing,
+            memOtrMutedRef = Nothing,
+            memOtrArchived = False,
+            memOtrArchivedRef = Nothing,
+            memHidden = False,
+            memHiddenRef = Nothing,
```
akshaymankar, line 433: This seems wrong, all of this data should be owned by user owning backend and not the conversation owning backend. So, it should never flow between domains.

mdimjasevic: This has been taken care of in PR #1622.

PR #1595, opened 2021-06-11
Handle LH policy conflicts in group conversations

smatting:
This PR ensures that users of these categories

- category 1: users that are under legalhold (are in state
  `UserLegalholdPending` or `UserLegalholdActive`)
- category 2: users that didn't give consent to be in a conversation with user from group 1

cannot be in the same group conversation (1:1 conversations are handled in #1507).

This is enforced by

1) removing users from group conversations and
2) preventing the creation of conversations with users from both categories 1 and 2.

#### 1) Removing users

When a user is transitioning into the `UserLegalholdPending` state all her group
conversations are traversed and users are removed. Who gets removed from the
conversation depends on the admins of the group conversation:

- If any of the admins has given consent then all users (including potential other admins) from category 2 are removed.
- If none of the admins has given consent then the user itself is removed from
  the conversation.

#### 2) Preventing creation  of conversations

This PR adds a guard that prevents conversations with initial participants that
are both from category 1 and 2.

This PR also adds a guard to the invite functionality:

A user from category 2 cannot be invited to a group conversation that contains
participants from category 1.

If a user from category 1 is invited to a group that contains users from
category 2 then the outcome depends:

- If the invitation contains any other users with missing consent then the
  invitation fails. This is the all-or-nothing principle: Either the invitation
  succeeds and all invitees are added to the conversation or it fails and none
  are added.

- If any of the group conversation admins has given consent then all users
(including potential other admins) from category 2 are removed from the group
conversation (Note this is consistent with the rule in 1)) and all invitees are
added to conversation.

- If none of the group conversation admins has given consent then the invitation
  fails. This case is theoretically impossible to occur, because invitations can
  only be created/accepted if the invitee is connected to the admin. The
  business logic ensures however that non-consenting users and legalhold
  activated users cannot be connected.

Internal notes: https://wearezeta.atlassian.net/browse/SQSERVICES-428


jschaul (commented):
Could the documentation of https://github.com/wireapp/wire-server/blob/develop/docs/reference/team/legalhold.md be updated with this PR?
(didn't review this PR yet and won't get to it today)

fisx:
> A user from category 2 cannot be invited to a group conversation
> that contains participants from category 1 and vice versa (the vice versa case
> is impossible to occur with the `whitelist-teams-and-implicit-consent` feature
> flag, because connections are blocked and invitation is not possible without it.
> That's why integration tests for this flag don't list this case).

What if user A, B are connected, A invites B to a group conv, but user C is also in that conv and has a conflict with B?  This does not require a connection between B and C, but still should fail.  Anyway, skipping a test because some non-trivial architecture choice makes it unlikely to fail makes me suspicious...  :)  I'll see if i can add it.

fisx (commented):
Tests look good.  I pushed two more, and changed the shape of `anyLegalholdConsentMissing` a little.  I'll fix all the call sites later, add the two tests, and then continue review.  Let me know if you disagree with anything so far @smatting.

services/galley/src/Galley/API/Query.hs
```diff
@@ -143,6 +149,21 @@ getConversations user mids mstart msize = do
...
+iterateConversations uid pageSize handleConvs = catMaybes <$> go Nothing
+  where
+    go :: Maybe ConvId -> Galley [Maybe a]
+    go mbConv = do
+      convResult <- getConversationsInternal uid Nothing mbConv (Just pageSize)
+      resultHead <- Just <$> handleConvs (convList convResult)
+      resultTail <- case convList convResult of
+        (conv : rest) ->
+          if convHasMore convResult
+            then go (Just (maximum (Data.convId <$> (conv : rest))))
```
fisx, line 162: I'm still not sure I understand this.  Either the convIds come back from cassandra in increasing order, in which case `maximum` should be `last`.  Or they don't, in which case `maximum` won't work, because it may miss arbitrarily many entries.  Please enlighten me!

services/galley/src/Galley/API/LegalHold.hs
```diff
@@ -393,7 +393,11 @@ changeLegalholdStatus tid uid old new = do
     --
     UserLegalHoldDisabled -> case new of
       UserLegalHoldEnabled -> illegal
-      UserLegalHoldPending -> addblocks >> update
+      UserLegalHoldPending -> do
+        update
+        -- outcome of addblocks logic might depend on the update. we'd like to
+        -- fail the transition in case addblocks fails nevertheless
```
fisx, line 399: What if `revert` fails?  Can we construct a hypothetical database state in memory and run `addblocks` on that instead?  Then we could maintain the old order.

fisx (changes requested):
looks good, except:

- [x] the two new tests are still missing
- [ ] i'm a bit oncerned about performance.  i wonder how long it would take to benchmark this a little?
- [ ] joe's request for documentation in https://github.com/wireapp/wire-server/blob/develop/docs/reference/team/legalhold.md
- [ ] the two comments below.

@smatting please look at my commits, and see if you agree.  If you want to add more commits let me know, otherwise I might do that tonight.

services/galley/src/Galley/API/LegalHold.hs
fisx: 596293363c09e2c91c55809fe81086c12c33fccd

services/galley/src/Galley/API/Query.hs
smatting: I agree that `last` would be equivalent. I wasn't sure if cassandra returns them ordered or not, so I didn't want to take any chances.
Why would `maximum` miss anything? Unless I'm mistaking `maximum` always returns the biggest element of a non-empty data structure, no matter if the elements are ordered or not. The default implementation of `Foldable` class is
```haskell
maximum :: forall a . Ord a => t a -> a
maximum = fromMaybe (errorWithoutStackTrace "maximum: empty structure") .
 getMax . foldMap (Max #. (Just :: a -> Maybe a))
```

services/galley/src/Galley/API/LegalHold.hs
smatting: Your solution doesn't rely on exceptions and state :+1:

smatting:
I changed the spec from

`All admins are consenting: all non-consenters get removed from conversation, new user joins (adding new users without consent in the same request fails)`
to
`At least one admin of the group has given consent. If all all users in the invite have given consent then the invite succeeds and all non-consenters from the group get removed`

Rationale for the change `All admins..` -> `At least one admin..`: This makes it consistent with the business logic that governs the removal of users when legalhold is activated for a users. Also the negation of `At least one admin` is `No admin` which was the other test case (`No admins are consenting: all LH..`)

---

I removed the case `No admins are consenting: all LH activated/pending users get removed from conversation (adding new users without consent in the same request succeeds)`, because it cannot occur (see updated PR description).

---

I added a test case
`At least one invited user has activated legalhold. At least one admin of the group has given consent. If any user in the invite has not given consent then the invite fails`. This is consistent with the business logic for _creating_ group conversations with initial participants with conflicting policies.

services/galley/src/Galley/API/Query.hs
fisx: ah, good point.  let's keep it as is then.

fisx (approved):

fisx:
> Could the documentation of https://github.com/wireapp/wire-server/blob/develop/docs/reference/team/legalhold.md be updated with this PR?
> (didn't review this PR yet and won't get to it today)

Documentation plus security review coming up, but not in this PR.

PR #1601, opened 2021-06-16
Add checklist for PRs

akshaymankar:
Hopefully this helps us create a nicer experience for reviewers and people who cut releases.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -8,3 +8,4 @@
        explain the change which will be included in the release notes.
  - [ ] If a component uses a new or changed internal endpoint of another
        component, this is mentioned in the CHANGELOG.md.
+- [ ] If this PR creates a new endpoint, or adds a new configuration flag, the endpoint / config-flag checklist (see Wire-employee only backend [wiki page](https://github.com/zinfra/backend-wiki/wiki/Checklists)) has been followed.
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,10 @@
+## Checklist
+
+ - [ ] Title of this PR explains the impact of the change.
+ - [ ] The description provides context as to why the change should occur and
+       what the code contributes to that effect. This could also be a link to a
+       JIRA ticket or a Github issue, if there is one.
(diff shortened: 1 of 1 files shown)


pcapriotti (approved):
Looks good!

mdimjasevic (approved):

mdimjasevic:
I didn't quite get the part about components and endpoints. I am not sure if that checkbox addresses it, but perhaps it would be worthwhile to have separate changelog considerations for external and internal endpoints.

akshaymankar:
> I didn't quite get the part about components and endpoints. 

So, if one creates a new internal endpoint in galley and consume it in brig. Now, it needs to be ensured that galley gets deployed before brig. This is because we cannot, as of now, guarantee that a new brig will only call new galley. To make sure this information doesn't get lost, we should write it down in the changelog.

> I am not sure if that checkbox addresses it, but perhaps it would be worthwhile to have separate changelog considerations for external and internal endpoints.

Yeah, this is not ensured by the checkbox. IMO changes to internal APIs don't really need to be put in the changelog, as consumers of this changelog don't really care.

.github/pull_request_template.md
```diff
@@ -0,0 +1,10 @@
+## Checklist
+
+ - [ ] Title of this PR explains the impact of the change.
+ - [ ] The description provides context as to why the change should occur and
+       what the code contributes to that effect. This could also be a link to a
+       JIRA ticket or a Github issue, if there is one.
+ - [ ] The CHANGELOG.md file in the *Unreleased* section has been updated to
+       explain the change which will be included in the release notes.
+ - [ ] If a component uses a new or changed internal endpoint of another
+       component, this is mentioned in the CHANGELOG.md.
```
jschaul, line 10:
```suggestion
       component, this is mentioned in the CHANGELOG.md.
- [ ] If this PR creates a new endpoint, or adds a new configuration flag, the endpoint / config-flag checklist (see Wire-employee only backend [wiki page](https://github.com/zinfra/backend-wiki/wiki/Checklists)) has been followed.
```

jschaul (approved):

PR #1603, opened 2021-06-17
Restore old error behaviour for client not found

pcapriotti:
Some clients (webapp at least) are relying on the response body being empty (or at least not a valid JSON error value) when fetching a non-existing client. This restores the previous behaviour by adding a new special constructor for brig handler errors, resulting in an empty response body.


akshaymankar (approved):
Looks good, we tested this on kube-staging by deploying the image built by the PR pipeline.

PR #1615, opened 2021-06-22
Saml IDP: allow to re-use previously used issuer names.

fisx:
**Bug:** If an IdP issuer is changed from A to B and then back to A, the second update fails with the message "A alredy in use".  This is technically correct, since we keep old idp configs around to allow gradual migration of users from old issuer name to new.  

**Fix:** IdPs should be allowed to reuser their *own* issuer names.

**Technical details:** [from memory, I probably should double-check] the idp keeps a list of old idp issuer names around.  If a user tries to auth with a new issuer name, but cannot be found in `spar.user`, we pull the list of old idp issuer names, and do lookups in `spar.user` by those.  If we find a user, it is because that user has not been seen since before one or more idp metadata updates, and the entry in `spar.user` is outdated.  We fix the entry and return the user.

Internal issue: https://wearezeta.atlassian.net/browse/SQSERVICES-537

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -53,6 +53,7 @@
 ## Bug fixes and other updates
 
 * Restore old behaviour for parse errors in request bodies (#1628, #1629).
+* Allow to change IdP Issuer name to previous name (#1615).
 
 # 2021-06-23
 
--- a/services/spar/src/Spar/API.hs
+++ b/services/spar/src/Spar/API.hs
@@ -386,7 +386,7 @@ validateIdPUpdate zusr _idpMetadata _idpId = do
               Nothing -> True
               Just c -> c ^. SAML.idpId == _idpId
         if notInUseByOthers
-          then pure $ (previousIdP ^. SAML.idpExtraInfo) & wiOldIssuers %~ (previousIssuer :)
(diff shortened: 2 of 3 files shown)


fisx:
```
Spar.API
  PUT /identity-providers/:idp
    issuer changed to one that already existed in the same team in the past (but has been updated away)
      changes back to the old one and keeps the new in the `old_issuers` list. FAILED [1]

Failures:

  test-integration/Test/Spar/APISpec.hs:770:20: 
  1) Spar.API, PUT /identity-providers/:idp, issuer changed to one that already existed in the same team in the past (but has been updated away), changes back to the old one and keeps the new in the `old_issuers` list.
       expected: 200
        but got: 400
```

fisx:
spar integration tests pass locally.

akshaymankar:
The new CI pipeline picked this PR up by mistake, I have stopped it, let me know if something is off here.

smatting (approved):

fisx:
> The new CI pipeline picked this PR up by mistake, I have stopped it, let me know if something is off here.

i don't think it was a mistake.  now i don't have a valid ci run on the last commit of this PR, so i'll just poke it again.  please leave it running!  :)

PR #1620, opened 2021-06-23
Message: Disallow unspecified ClientMismatchStrategy

akshaymankar:
## Checklist

### DIFF
--- a/libs/wire-api/src/Wire/API/Message.hs
+++ b/libs/wire-api/src/Wire/API/Message.hs
@@ -359,7 +359,7 @@ data ClientMismatchStrategy
 
 protolensToClientMismatchStrategy :: Maybe Proto.Otr.QualifiedNewOtrMessage'ClientMismatchStrategy -> Either String ClientMismatchStrategy
 protolensToClientMismatchStrategy = \case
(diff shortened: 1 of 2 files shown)


libs/wire-api/src/Wire/API/Message.hs
```diff
@@ -359,7 +359,7 @@ data ClientMismatchStrategy
 
 protolensToClientMismatchStrategy :: Maybe Proto.Otr.QualifiedNewOtrMessage'ClientMismatchStrategy -> Either String ClientMismatchStrategy
 protolensToClientMismatchStrategy = \case
-  Nothing -> Right MismatchReportAll
+  Nothing -> Left "ClientMismatchStrategy not specified!"
```
jschaul, line 362: Is this backwards compatible with old clients who use protobuf on the old endpoint? Or is the unqualified endpoint making use indirectly of this function, too?

jschaul (approved):
Looks good, see comment

libs/wire-api/src/Wire/API/Message.hs
akshaymankar: The unqualified endpoint is not using this at all.

PR #1622, opened 2021-06-23
Federation: Send Minimal Information on Members in a New Conversation gRPC Call

mdimjasevic:
A receiving backend should decide on details of how each of its users has the new conversation configured, and not the conversation-owning backend.

The initial version of the federation RPC for registering a new conversation for each user communicated if they have the conversation archived, muted, or hidden. That way the conversation-owning backend dictated how remote members/users see the conversation. The PR changes that so that each remote user's backend sets initial values for these conversation settings instead.

## Checklist


services/galley/src/Galley/API/Util.hs
```diff
@@ -466,22 +455,29 @@ fromRegisterConversation (Qualified usr localDomain) MkRegisterConversation {..}
...
-        -- make sure not to include 'usr' in the list of others
-        . Map.adjust (filter ((usr /=) . Public.memId)) localDomain
-    memToOther :: Domain -> Member.Member -> OtherMember
-    memToOther d mem =
-      OtherMember
-        { omQualifiedId = Qualified (Member.memId mem) d,
-          omService = Member.memService mem,
-          omConvRoleName = Member.memConvRoleName mem
-        }
+    me :: Set OtherMember -> Galley Public.Member
```
jschaul, line 458:
```suggestion
    -- Currently this function creates a Member with default conversation attributes
    -- FUTUREWORK(federation): retrieve member's conversation attributes (muted, archived, etc) here once supported by the database schema. 
    me :: Set OtherMember -> Galley Public.Member
```

jschaul (approved):

CHANGELOG.md
```diff
@@ -16,9 +16,11 @@
 
 # Unreleased
 
-## Internal changes
+# Internal changes
 
 * schema-profunctor: add `optField` combinator and corresponding documentation (#1621, #1624).
+* [Federation] Let a receiving backend decide membership specifics of its users
+  added to a new conversation via `POST /federation/register-conversation` (#1622)
```
jschaul, line 23:
```suggestion
* [Federation] Let a receiving backend decide conversation attribute specifics of its users
  added to a new conversation via `POST /federation/register-conversation` (#1622)
```

services/galley/src/Galley/API/Util.hs
```diff
@@ -446,7 +435,7 @@ fromRegisterConversation ::
   Qualified UserId ->
   RegisterConversation ->
   Galley Public.Conversation
-fromRegisterConversation (Qualified usr localDomain) MkRegisterConversation {..} = do
+fromRegisterConversation qusr MkRegisterConversation {..} = do
```
pcapriotti, line 438: Couldn't you reuse `conversationView` here?

services/galley/src/Galley/API/Util.hs
```diff
@@ -466,22 +455,31 @@ fromRegisterConversation (Qualified usr localDomain) MkRegisterConversation {..}
...
+                memOtrMuted = False,
+                memOtrMutedStatus = Nothing,
+                memOtrMutedRef = Nothing,
+                memOtrArchived = False,
+                memOtrArchivedRef = Nothing,
+                memHidden = False,
+                memHiddenRef = Nothing,
+                memConvRoleName = omConvRoleName v
+              }
+    others :: Set OtherMember -> [OtherMember]
```
pcapriotti, line 478: Here it would also be nicer to use a higher-level fold (i.e. `foldMap` like above).

pcapriotti (commented):
Looks good, but it would be nice if we could reuse this "conversation view" logic instead of implementing it twice.

services/galley/src/Galley/API/Util.hs
mdimjasevic: It's different types that are involved, though if I think hard enough, maybe I would be able to find a way to unify them.

pcapriotti: Is it possible to use the same technique as in the local implementation? Create *some* `Conversation` object, then pass it to `conversationView` to produce the various views for users.

services/galley/src/Galley/API/Federation.hs
```diff
@@ -61,10 +60,13 @@ registerConversation rc = do
...
-    getLocals localDomain = fromMaybe [] . Map.lookup localDomain . rcMembers $ rc
-    toQualified :: Domain -> Member -> Qualified UserId
-    toQualified domain mem = Qualified (memId mem) domain
+    getLocals :: Domain -> Set OtherMember -> [Qualified UserId]
+    getLocals d = foldl' (addInDomain d) []
+    addInDomain :: Domain -> [Qualified UserId] -> OtherMember -> [Qualified UserId]
+    addInDomain d acc m =
+      if ((== d) . qDomain . omQualifiedId) m
+        then omQualifiedId m : acc
+        else acc
```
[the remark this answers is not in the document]
mdimjasevic: Nice! I'm adding this in an upcoming commit.

services/galley/src/Galley/API/Util.hs
mdimjasevic: Done!

services/galley/src/Galley/API/Util.hs
```diff
@@ -466,22 +455,28 @@ fromRegisterConversation (Qualified usr localDomain) MkRegisterConversation {..}
...
+    others = foldMap (\om -> guard (omQualifiedId om == qusr) $> om)
```
pcapriotti, line 479:
```suggestion
    others = foldMap (\om -> guard (omQualifiedId om /= qusr) $> om)
```

services/galley/src/Galley/API/Util.hs
```diff
@@ -466,22 +455,28 @@ fromRegisterConversation (Qualified usr localDomain) MkRegisterConversation {..}
...
-      OtherMember
-        { omQualifiedId = Qualified (Member.memId mem) d,
-          omService = Member.memService mem,
-          omConvRoleName = Member.memConvRoleName mem
-        }
+    -- Currently this function creates a Member with default conversation attributes
+    -- FUTUREWORK(federation): retrieve member's conversation attributes (muted, archived, etc) here once supported by the database schema.
+    me :: Set OtherMember -> Galley Public.Member
+    me s =
+      case find ((== qusr) . omQualifiedId) . Set.toList $ s of
```
pcapriotti, line 462: This is an O(n) lookup. It's better to do the lookup on the set `s`.

pcapriotti (changes requested):
I've left some comments below.

However, I would suggest rewriting `fromRegisterConversation` so that it is clear that it cannot throw. Maybe we can have:

```haskell
remoteConversationViews :: RegisterConversation -> Galley [(Qualified UserId, Conversation)]
```

so that we can combine looping over `rcMembers` and the corresponding lookup. For example, we can use something like:

```
setHoles :: Ord a => Set a -> [(a, Set a)]
setHoles s = foldMap (\x -> [(x, Set.delete x s)]) s
```

as the basic building block.

Also consider if it needs to live in the `Galley` monad at all.

services/galley/src/Galley/API/Util.hs
pcapriotti: I was wrong: `conversationView` takes a `Data.Conversation`, and it doesn't really make sense to create such a value on the remote backend. Maybe just leave a FUTUREWORK comment to remove this duplicated logic in the future.

services/galley/src/Galley/API/Util.hs
mdimjasevic: True, it is an `O(n)` lookup. Note I need to search via a projection function, and not by equality on `OtherMember`. I couldn't find a function to achieve this goal such that it wouldn't require `O(n)`. If you have an idea how, please let me know!

pcapriotti: Ah, I see. I guess there is no way at this point. But it should be fine if you implement my suggestion above. Then the whole thing is `O(n log n)` instead of `O(n^2)`.

mdimjasevic:
> I've left some comments below.
> 
> However, I would suggest rewriting `fromRegisterConversation` so that it is clear that it cannot throw. 

We have no guarantee that the lookup for `qusr` will succeed, hence the `Galley` monad to allow for throwing. The only alternative I see is to use partial lookup functions.


> ```haskell
> remoteConversationViews :: RegisterConversation -> Galley [(Qualified UserId, Conversation)]
> ```
> 
> so that we can combine looping over `rcMembers` and the corresponding lookup. For example, we can use something like:
> 
> ```
> setHoles :: Ord a => Set a -> [(a, Set a)]
> setHoles s = foldMap (\x -> [(x, Set.delete x s)]) s
> ```
> 
> as the basic building block.

I see. This is a good idea, though that would require changing the type of the `rcMembers` field from `Set OtherMember` to e.g. `[OtherMember]`. I could reorganise code to accommodate for this suggestion.


> Also consider if it needs to live in the `Galley` monad at all.

I mentioned above why I used the `Galley` monad, but if you have a suggestion on how to avoid it altogether, I'll be more than happy to get rid of it.

mdimjasevic:
@pcapriotti , I implemented a more efficient and non-throwing version of `fromRegisterConversation`. Thanks for seeing that better way to achieve the same goal! Please take a look now.

pcapriotti (approved):
Nice!

PR #1626, opened 2021-06-24
Federation: Provide a New Endpoint for Obtaining Classified Domains

mdimjasevic:
The PR implements a new endpoint `GET /teams/:teamId/features/classifiedDomains` that fetches a list of classified domains. There is an internal endpoint equivalent.

## Checklist

- [x] Add a test that checks if a user is authenticated when getting classified domains


docs/reference/config-options.md
```diff
@@ -39,6 +39,7 @@ settings:
     legalhold: disabled-by-default
     teamSearchVisibility: disabled-by-default
     setEmailVisibility: visible_to_self
+    classifiedDomains: disabled-by-default
```
smatting, line 42: `disabled-by-default` is not a valid config value for `classifiedDomains`

hack/helm_vars/wire-server/values.yaml
```diff
@@ -148,6 +148,7 @@ galley:
         sso: disabled-by-default  # this needs to be the default; tests can enable it when needed.
         legalhold: whitelist-teams-and-implicit-consent
         teamSearchVisibility: disabled-by-default
+        classifiedDomains: disabled-by-default
```
smatting, line 151: `disabled-by-default` is not a valid config value for `classifiedDomains`

libs/galley-types/src/Galley/Types/Teams.hs
```diff
@@ -285,9 +289,13 @@ makeLenses ''FeatureFlags
 -- client apps treat permission bit matrices as opaque role identifiers, so if we add new
 -- permission flags, things will break there.
 --
+-- "Hidden" in "HiddenPerm", therefore, refers to a permission hidden from
+-- clients, thereby making it internal to the backend.
+--
 -- The solution: add new permission bits to 'HiddenPerm', 'HiddenPermissions', and make
 -- 'hasPermission', 'mayGrantPermission' polymorphic.  Now you can check both for the hidden
 -- permission bits and the old ones that we share with the client apps.
+-- TODO(md): explain why hidden
```
smatting: leftover TODO

services/galley/src/Galley/API/Public.hs
```diff
@@ -934,5 +935,7 @@ mkFeatureGetAndPutRoute getter setter = do
               description "JSON body"
             response 204 "Team feature status" end
 
-  mkPutRoute True (toByteString' featureName)
-  mkGetRoute False `mapM_` Public.deprecatedFeatureName featureName
+  for_ mbSetter $ \setter -> do
+    mkPutRoute True (toByteString' featureName) setter
+
+-- mkPutRoute False `mapM_` Public.deprecatedFeatureName featureName setter
```
smatting, line 941: I think this should be uncommented

services/galley/test/integration/API/Teams/Feature.hs
```diff
@@ -171,6 +172,34 @@ testSearchVisibility = do
     getTeamSearchVisibility tid3 Public.TeamFeatureEnabled
     getTeamSearchVisibilityInternal tid3 Public.TeamFeatureEnabled
 
+testClassifiedDomains :: TestM ()
```
smatting, line 175: The test depends on the config for galley. So the galley config for both environments should be the same to guarantee that the test succeds on both environments:

- for tests on CI: `hack/helm_vars/wire-server/values.yaml` and `/charts/galley/templates/configmap.yaml` need to be adapted
- for development: `services/galley/galley.integration.yaml`

smatting (changes requested):

docs/reference/config-options.md
mdimjasevic: Right. I see that when other features were documented in this file, that they referred to a specific commit in the `wire-server` repository so that a pinned version can be inspected for how it was implemented. Should I do the same here? If so, I suppose I should leave out changing this documentation file in the PR and use a follow-up PR that can reference a to-be merged commit of this PR.

libs/galley-types/src/Galley/Types/Teams.hs
mdimjasevic: Thanks! I removed it.

services/galley/src/Galley/API/Public.hs
mdimjasevic: You are right! Done.

services/galley/test/integration/API/Teams/Feature.hs
mdimjasevic: Sure, I will align these configuration files.

Should I have this test here that would fetch the classified domains via the endpoint and just assert the values it expects there?

deploy/services-demo/conf/galley.demo.yaml
```diff
@@ -39,6 +39,10 @@ settings:
         config:
           enforceAppLock: false
           inactivityTimeoutSecs: 60
+    classifiedDomains:
+      status: enabled
+      config:
+        domains: ["wire.com"]
```
akshaymankar, line 45:
```suggestion
        domains: ["example.com"]
```

We should try to avoid putting wire.com in the code.

hack/helm_vars/wire-server/values.yaml
```diff
@@ -148,6 +148,9 @@ galley:
         sso: disabled-by-default  # this needs to be the default; tests can enable it when needed.
         legalhold: whitelist-teams-and-implicit-consent
         teamSearchVisibility: disabled-by-default
+        classifiedDomains:
+          status: disabled
+          config:
```
akshaymankar, line 153:
```suggestion
          config: {}
```

Clearer this way.

docs/reference/config-options.md
mdimjasevic: @smatting , just a gentle reminder of this question.

hack/helm_vars/wire-server/values.yaml
mdimjasevic: Done.

deploy/services-demo/conf/galley.demo.yaml
mdimjasevic: Done.

mdimjasevic:
@smatting , should I also add an endpoint in Stern for classified domains, just like the app lock feature has one at https://github.com/wireapp/wire-server/blob/b75778427111ddcad0adb62f7d7bf17b8e37cb9a/tools/stern/src/Stern/API.hs#L352 ?

jschaul (commented):
I ran `helm install --debug --dry-run wire-server .local/charts/wire-server -f hack/helm_vars/wire-server/values.yaml > output.yaml` using a valid kubernetes server config, which results in a final galley config of

```
galley:
  config:
    cassandra:
      host: cassandra-ephemeral
      replicaCount: 1
    enableFederator: true
    journal:
      endpoint: http://fake-aws-sqs:4568
      queue: integration-team-events.fifo
    settings:
      conversationCodeURI: https://kube-staging-nginz-https.zinfra.io/conversation-join/
      enableIndexedBillingTeamMembers: true
      featureFlags:
        classifiedDomains:
          config:
            domains:
            - example.com
          status: enabled
        legalhold: whitelist-teams-and-implicit-consent
        sso: disabled-by-default
        teamSearchVisibility: disabled-by-default
      federationDomain: integration.example.com
```

which appears to be in line with what's in galley.integration.yaml.

Do integration tests pass locally?

charts/galley/values.yaml
```diff
@@ -30,6 +30,9 @@ config:
       sso: disabled-by-default
       legalhold: disabled-by-default
       teamSearchVisibility: disabled-by-default
+      classifiedDomains:
+        status: disabled
+        config: {}
```
jschaul, line 35: This is an invalid configuration:

```
galley: user error (AesonException "Error in $.settings.featureFlags.classifiedDomains.config: key \"domains\" not found")
```

According to the current code, only the following is valid:

```yaml
    classifiedDomains:
      status: disabled
      config:
        domains: []
```

two options: change this default; or make the code lenient in parsing such that if the status is disabled, the config object is ignored if absent.

services/galley/test/integration/API/Teams/Feature.hs
```diff
@@ -42,7 +46,9 @@ tests s =
       test s "LegalHold" testLegalHold,
       test s "SearchVisibility" testSearchVisibility,
       test s "DigitalSignatures" $ testSimpleFlag @'Public.TeamFeatureDigitalSignatures,
-      test s "ValidateSAMLEmails" $ testSimpleFlag @'Public.TeamFeatureValidateSAMLEmails
+      test s "ValidateSAMLEmails" $ testSimpleFlag @'Public.TeamFeatureValidateSAMLEmails,
+      test s "Classified Domains (enabled)" testClassifiedDomainsEnabled,
+      test s "Classified Domains (disabled)" testClassifiedDomainsDisabled
```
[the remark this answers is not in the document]
jschaul: I performed a manual test:

```
curl -s http://localhost:8085/teams/b8509dc2-4b4e-4d2f-afd6-410666bbb464/features -H Z-User:c8b2bb6b-4980-4950-af5d-211d6cb951e9 | json_pp
{
   "appLock" : {
      "config" : {
         "enforceAppLock" : false,
         "inactivityTimeoutSecs" : 60
      },
      "status" : "enabled"
   },
   "classifiedDomains" : {
      "config" : {
         "domains" : [
            "example.com"
         ]
      },
      "status" : "enabled"
   },
   "digitalSignatures" : {
      "status" : "disabled"
   },
   "legalhold" : {
      "status" : "disabled"
   },
   "searchVisibility" : {
      "status" : "disabled"
   },
   "sso" : {
      "status" : "disabled"
   },
   "validateSAMLemails" : {
      "status" : "disabled"
   }
}
```

(and got some user/team ids using `select team, binding, creator from galley_test.team limit 10;` after `make cqlsh` and picked IDs from a binding team)

Still, an integration test here would be nice.

charts/galley/templates/configmap.yaml
```diff
@@ -62,5 +62,6 @@ data:
         sso: {{ .settings.featureFlags.sso }}
         legalhold: {{ .settings.featureFlags.legalhold }}
         teamSearchVisibility: {{ .settings.featureFlags.teamSearchVisibility }}
+        classifiedDomains: {{ .settings.featureFlags.classifiedDomains }}
```
jschaul, line 65: Ah, I found the issue you have. This renders as

```
      featureFlags:
        sso: disabled-by-default
        legalhold: whitelist-teams-and-implicit-consent
        teamSearchVisibility: disabled-by-default
        classifiedDomains: map[config:map[domains:[example.com]] status:enabled]
---
```

so you need to use another helm magic function to get this to render properly. See https://stackoverflow.com/questions/63619667/helm-include-formatted-map-from-values-yml

test it out using:

```diff
--- a/hack/bin/integration-setup.sh
+++ b/hack/bin/integration-setup.sh
@@ -4,22 +4,22 @@ USAGE="Usage: $0"

-charts=( fake-aws databases-ephemeral wire-server nginx-ingress-controller nginx-ingress-services )
+charts=( wire-server )
 
 echo "Installing charts..."
@@ -69,6 +69,8 @@ for chart in "${charts[@]}"; do
         --set brig.config.optSettings.setFederationDomain="$FEDERATION_DOMAIN" \
         --set galley.config.settings.federationDomain="$FEDERATION_DOMAIN" \
         --wait \
+        --debug \
+        --dry-run \
         --timeout "$TIMEOUT"
     set +x
 done

-until resourcesReady; do echo 'waiting for SNS resources'; sleep 1; done
-
-kubectl -n ${NAMESPACE} get pods
 ```
 
 then running `DOCKER_TAG=2.109.27 make kube-integration-setup-sans-federation`

File: charts/galley/values.yaml (hunk 8 above)
Comment (mdimjasevic) [reply charts/galley/values.yaml:35]: Thanks for spotting this! I fixed this in an upcoming commit.

File: services/galley/test/integration/API/Teams/Feature.hs [hunk 11]
```diff
@@ -182,6 +188,66 @@ testSearchVisibility = do
...
+  m ()
+getClassifiedDomainsInternal tid =
+  assertFlagWithConfig @Public.TeamFeatureClassifiedDomainsConfig $
+    Util.getTeamFeatureFlagInternal Public.TeamFeatureClassifiedDomains tid
+
+testClassifiedDomainsEnabled :: TestM ()
+testClassifiedDomainsEnabled = do
+  owner <- Util.randomUser
+  member <- Util.randomUser
+  tid <- Util.createNonBindingTeam "classified domains" owner []
```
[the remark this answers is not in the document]
mdimjasevic: Great, thanks for this! I'm updating this in an upcoming commit.

jschaul (approved):

charts/galley/templates/configmap.yaml
mdimjasevic: This has been taken care of.

mdimjasevic:
@jschaul , I just added an integration test for the all features endpoint.

smatting (approved):
I think this good to merge since the 1 failing test is known to be unrelated.

PR #1630, opened 2021-06-28
charts/nginz: Add restrictive CORS headers to all URLs and explicitly…

arianvp:
Add restrictive CORS headers to all origins and explicitly allowlist allowed origins

There is no reason not to have restrictive CORS on all endpoints; given
our API is currently only consumed by known services. We can loosen this
restriction again in the future (e.g. when we have something like OAuth
integration).

We also move to an explicit allowlist instead of trusting the entirety
of `external_env_domain`. Especially on-prem people might have many more
services running on their base domain and we as Wire have no control
over it. We do not want other services not in our control to be a vector
for e.g. XSS attacks. Hence we require the user of our charts to
explicitly allowlist all the expected origins.  We will update the docs
to guide people into making this update if needed. 

## Checklist

 - [ ] The description provides context as to why the change should occur and what the code contributes to that effect. This could also be a link to a JIRA ticket or a Github issue, if there is one.
 - [ ] The CHANGELOG.md file in the *Unreleased* section has been updated to explain the change which will be included in the release notes.
 - [ ] If a component uses a new or changed internal endpoint of another component, this is mentioned in the CHANGELOG.md.

### DIFF
--- a/changelog.d/0-release-notes/cors
+++ b/changelog.d/0-release-notes/cors
@@ -1,6 +1,5 @@
-
 The `nginz` chart now configures nginx to only allow cross-origin requests from
-an epxlicit allow list of sub-domains. By default these are:
+an explicit allow list of subdomains. By default these are:
 
 ```yaml
 nginz:
@@ -11,5 +10,6 @@ nginz:
     - accounts
 ```
 
-If you changed the names of these services you have to adjust those names in the nginz config as well!
+If you changed the names of these services, you must adjust those names in the
+nginz config as well.
 
--- a/changelog.d/0-release-notes/cors
+++ b/changelog.d/0-release-notes/cors
@@ -1,0 +1,15 @@
+
+The `nginz` chart now configures nginx to only allow cross-origin requests from
+an epxlicit allow list of sub-domains. By default these are:
+
+```yaml
+nginz:
+  nginx_conf:
+    allowlisted_origins:
+    - webapp
+    - teams
+    - accounts
+```
+
+If you changed the names of these services you have to adjust those names in the nginz config as well!
+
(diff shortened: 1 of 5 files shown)


arianvp:
Will have to test this, and write extensive CHANGELOG entry before merging this.

We also have to update the wire installation docs so that new users are sure to not forget this!

franziskuskiefer (approved):
lgtm

arianvp:
Needs rebase

arianvp:
@jmatsushita this is ready for another review

charts/nginz/values.yaml
```diff
@@ -52,6 +52,11 @@ nginx_conf:
   - ~* ^/conversations/([^/]*)/call/state
   - /search/top
   - /search/common
+  # -- The origins from which we allow CORS requests. These are combined with 'external_env_domain' to form a full url
+  allowlisted_origins:
+    - webapp
+    - teams
+    - accounts
```
jschaul, line 59:
```suggestion
    - account
```

changelog.d/0-release-notes/cors
```diff
@@ -0,0 +1,15 @@
+The `nginz` chart now configures nginx to only allow cross-origin requests from
+an explicit allow list of subdomains. By default these are:
+
+```yaml
+nginz:
+  nginx_conf:
+    allowlisted_origins:
+    - webapp
+    - teams
+    - accounts
```
jschaul, line 10:
```suggestion
    - account
```

jschaul (approved):
Minor change: we use and have been using 'account' (without s) everywhere in docs.wire.com and our own environments (except for accounts.wire.com - here we can do an override)

Looks fine, let's merge and deploy this to some of our environments.

Could you, after merging this, also add this to wire-docs somewhere?

charts/nginz/values.yaml
sysvinit: Done

changelog.d/0-release-notes/cors
sysvinit: Done

PR #1633, opened 2021-06-28
[WIP] [RFC] per-request authorization for the example RPC get conversations [skip ci]

jschaul:
Given backend A, B and C with a conversation1@A containing users from A and B. We probably wish to restrict certain actions to be done by malicious backends, such as getting or inserting users into conversation1. If server2server authentication allows us to know "this requests claiming to be from server C is indeed from server C", and our membership check allows us to say "the request contains an originUser Bob@B wanting to add someone to conversation1 which is allowed since Bob belongs to conversation1", we still have a problem, since backend C can create a request with originDomain=C (which is valid) and originUser=Bob@B (which is a lie), and currently the server will accept that request, since nothing connects the domain of the sender backend with the domain of the (supposed) sending user.

Doing a check in each function as done in this first PR would in theory work, but may be brittle as prone to regressions and a missed check for the next endpoint/RPC that gets implemented. 

Can we solve this in a better fashion?

Side question: is there a way to add a header to the servant API, without needing to provide it explicitly when making requests [here](https://github.com/wireapp/wire-server/blob/develop/services/galley/src/Galley/API/Query.hs#L109) since it's already added on the way [here](https://github.com/wireapp/wire-server/blob/develop/services/federator/src/Federator/Service.hs#L64)? 

## Checklist

 - [ ] Title of this PR explains the impact of the change.
 - [ ] The description provides context as to why the change should occur and what the code contributes to that effect. This could also be a link to a JIRA ticket or a Github issue, if there is one.
 - [ ] The CHANGELOG.md file in the *Unreleased* section has been updated to explain the change which will be included in the release notes.
 - [ ] If a component uses a new or changed internal endpoint of another component, this is mentioned in the CHANGELOG.md.

### DIFF
--- a/libs/wire-api-federation/src/Wire/API/Federation/API/Galley.hs
+++ b/libs/wire-api-federation/src/Wire/API/Federation/API/Galley.hs
@@ -32,6 +32,7 @@ import Wire.API.Conversation (Access, AccessRole, ConvType, Conversation, Receip
 import Wire.API.Conversation.Member (OtherMember)
 import Wire.API.Conversation.Role (RoleName)
 import Wire.API.Federation.Client (FederationClientFailure, FederatorClient)
+import Wire.API.Federation.Domain
 import qualified Wire.API.Federation.GRPC.Types as Proto
 import Wire.API.Federation.Util.Aeson (CustomEncoded (CustomEncoded))
 
@@ -50,7 +51,8 @@ data Api routes = Api
         :> Post '[JSON] (),
     getConversations ::
       routes
-        :- "federation"
+        :- DomainHeader
+        :> "federation"
         :> "get-conversations"
         :> ReqBody '[JSON] GetConversationsRequest
         :> Post '[JSON] GetConversationsResponse,
--- a/libs/wire-api-federation/src/Wire/API/Federation/Error.hs
+++ b/libs/wire-api-federation/src/Wire/API/Federation/Error.hs
@@ -130,6 +130,13 @@ federationUnavailable err =
     "federation-not-available"
     ("Local federator not available: " <> LT.fromStrict err)
 
+federationUnauthorized :: Text -> Wai.Error
+federationUnauthorized err =
+  Wai.mkError
+    HTTP.status403
+    "federation-unauthorized"
+    (LT.fromStrict err)
+
 federationRemoteError :: Proto.OutwardError -> Wai.Error
 federationRemoteError err = Wai.mkError status (LT.fromStrict label) (LT.fromStrict msg)
   where
--- a/services/galley/src/Galley/API/Federation.hs
+++ b/services/galley/src/Galley/API/Federation.hs
@@ -16,7 +16,9 @@
 -- with this program. If not, see <https://www.gnu.org/licenses/>.
 module Galley.API.Federation where
 
+import Control.Monad.Catch (throwM)
 import Data.Containers.ListUtils (nubOrd)
+import Data.Domain
 import Data.Qualified (Qualified (..))
 import qualified Galley.API.Mapping as Mapping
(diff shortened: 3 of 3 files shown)


pcapriotti:
Maybe I'm missing something, but isn't it cleaner to just pass a `UserId` instead of a qualified one in `GetConversationsRequest`? This is what I've done in https://github.com/wireapp/wire-server/pull/1596/, where the `Qualified ConvId` is constructed using the `originDomain` header of the RPC.

jschaul:
> Maybe I'm missing something, but isn't it cleaner to just pass a `UserId` instead of a qualified one in `GetConversationsRequest`? This is what I've done in #1596, where the `Qualified ConvId` is constructed using the `originDomain` header of the RPC.

That is certainly a possibility, to avoid passing the domain when making the request, having federator add it, and upon receiving a request, take the domain and add it back into the functions that operator on it.

As you've noticed by now, this pattern leads to the problem that rpcs needs to add a header one more time, as described in your own TODO here: https://github.com/wireapp/wire-server/pull/1638/files#diff-f17e1e3fce300968eff3886ae1411444acf717b7f0eb113d8be97f09e5dbc27bR324
That's why I'm currently not yet convinced of that pattern.

pcapriotti:
> As you've noticed by now, this pattern leads to the problem that rpcs needs to add a header one more time, as described in your own TODO here: https://github.com/wireapp/wire-server/pull/1638/files#diff-f17e1e3fce300968eff3886ae1411444acf717b7f0eb113d8be97f09e5dbc27bR324
> That's why I'm currently not yet convinced of that pattern.

@smatting and I have a fix for that in #1609. Maybe it's worth splitting that into its own PR.

smatting:
> Can we solve this in a better fashion?

I think autorization problems like this one should be solved for every RPC individually and not generally, because I worry that whatever general abstraction/model we need to come up will break (not be flexible enough) with a possible unforseen use case/RPC. Solving authorization on the RPC level is also tried-and-proven strategy for us in all non-federation endpoints. 
Also this keeps the mental model simple: Federator does Authentication (checks that Wire-Origin-Domain is not spoofed), Federation API does Authorization (e.g. can this user add others to this conversation). 

> @smatting and I have a fix for that in #1609: 6e550cc. Maybe it's worth splitting that into its own PR.

@akshaymankar and I reverted that again, because

> A test was failing and that tipped us that maybe not all clients (specially tests) of the Servant API want to use it in a way that they don't specify doamin header. So, we thought it is better to make the default client ask for it explicitly. If we think we want to avoid having to do that we should write come up with some other way. We tried a little with overlapping instances, but it didn't work, so we gave up.  Maybe using AsClient is not the right thing here and we should create something parallel to that to be used in prod code.

I'd favor an approach if every RPC call always has to specify the origin domain if the RPC requires it. There is nothing gained security-wise if it is set implicely by the federator if we assume there are rogue request. Also not every context in which a  federation client is used has an origin domain, for example integration tests. Having the origin domain be explicit also leaves more flexibility if we want to allow multiple domains per host in the future (e.g. domain depending on the team of the user that sends a message).

> This is what I've done in #1596, where the Qualified ConvId is constructed using the originDomain header of the RPC.
+1 This correct-by-construction approach also avoids having to making certain authorization checks altogether. I think we should apply it whenever possible.

akshaymankar:
>  Also not every context in which a federation client is used has an origin domain, for example integration tests. Having the origin domain be explicit also leaves more flexibility if we want to allow multiple domains per host in the future (e.g. domain depending on the team of the user that sends a message).

To add to this, another such context would be the federator. Currently it uses bilge stuff, if we keep this flexibility one day it could also use servant and this way we could even guarantee at type level that federator only calls federation APIs.

jschaul:
I suppose this is obsolete now in the current "valid-by-construction" approach.

PR #1634, opened 2021-06-29
Refactor qualified send message endpoint

pcapriotti:
This is a from-scratch rewrite of the send message endpoint to make it work directly in terms of qualified users and their clients. In the process, we simplified the client verification logic and moved some things away from the `Galley.API.Update` module. Note that this still does not implement any federation logic. This should in theory result in no visible changes for all-local requests.

## Summary of changes

- The qualified endpoint is now implemented directly using new code that does not refer back to the unqualified version.
- The logic for validating and filtering clients now works on "flattened" sets of clients (really, triples of domain, user id and client id), instead of nested maps
- Property tests have been added.
- The integration tests for the qualified endpoint are not checking the reported `redundant` clients, since these are now incorrect because the remote clients are taken into account.
- The `LegalholdProtectee'` has been replaced by a pair of `UserType` (i.e. bot or user) and a `UserId`.
- Legalhold-specific logic has been moved to the `LegalHold` module.
- A `UnionT` monad transformer has been introduced to make it easier to define handlers for `UVerb` endpoints with many branches.

## Checklist


services/galley/test/integration/API.hs
```diff
@@ -244,44 +243,44 @@ postCryptoMessage1 = do
     assertTrue_ (eqMismatch [(eve, Set.singleton ec)] [] [] . responseJsonUnsafe)
   -- Complete
   WS.bracketR2 c bob eve $ \(wsB, wsE) -> do
-    let m2 = [(bob, bc, "ciphertext2"), (eve, ec, "ciphertext2")]
+    let m2 = [(bob, bc, toBase64Text "ciphertext2"), (eve, ec, toBase64Text "ciphertext2")]
```
jschaul, line 246: Is there a backwards-incompatibility regression in the sense that before non-base64 encoded strings were accepted, which is now no longer the case, and only base64encoded test is allowed?
It's not generally a problem to accept only that, but this change/refactor isn't explained in a comment in the code nor PR description so it might be good to add that.

services/galley/test/integration/API.hs
```diff
@@ -498,6 +497,7 @@ postMessageQualifiedLocalOwningBackendMissingClients = do
   (chadOwningDomain, chadClient) <- randomUserWithClientQualified (someLastPrekeys !! 3)
   chadClient2 <- randomClient (qUnqualified chadOwningDomain) (someLastPrekeys !! 2)
   deeId <- randomId
+  -- deeClient <- liftIO $ generate arbitrary
```
jschaul, line 500: remove leftover comment?

services/galley/test/integration/API.hs
```diff
@@ -519,10 +519,18 @@ postMessageQualifiedLocalOwningBackendMissingClients = do
...
+              Map.fromList
+                [ ( owningDomain,
+                    Map.fromList $
+                      [ (bobUnqualified, Set.singleton bobClient),
+                        (chadUnqualified, Set.singleton chadClient2)
+                      ]
+                  )
+                  -- ( remoteDomain,
+                  --   Map.singleton (qUnqualified deeRemote) (Set.singleton deeClient)
+                  -- )
```
jschaul, line 532: work in progress or can this be removed?

services/galley/test/integration/API/Util.hs
```diff
@@ -1673,9 +1673,10 @@ eqMismatchQualified ::
   Maybe Message.MessageSendingStatus ->
   Bool
 eqMismatchQualified _ _ _ Nothing = False
-eqMismatchQualified missing redundant deleted (Just other) = do
+eqMismatchQualified missing _redundant deleted (Just other) = do
   missing == Message.mssMissingClients other
-    && redundant == Message.mssRedundantClients other
+    -- FUTUREWORK: reenable check once remote client discovery is implemented
+    -- && redundant == Message.mssRedundantClients other
```
jschaul, line 1679: I don't understand why this check cannot be done at this stage when the refactor doesn't actually create tests including remote users yet (or at least when *some* tests don't include remote users).

Would it be feasible to have two functions here and only use the commented out on in places where remote users are present?

jschaul (approved):
Generally looks good. Nitpicks and comments inline.

> The integration tests are not checking the reported redundant clients, since these are now incorrect because the remote clients are taken into account.

I think I don't understand this sentence. For integration tests with local users and local clients only, why can the reported redundant clients not be checked anymore? Maybe the sentence could be reworded to state that out of the 17 tests testing message sending, for 3 of them including remote users, that check isn't done. Or the tests could be adapted to continue doing the check where feasible?

services/galley/src/Galley/API/Message.hs
```diff
@@ -0,0 +1,332 @@
...
+  ClientMismatch
+    { cmismatchTime = mssTime status,
+      missingClients = unqualify (mssMissingClients status),
+      redundantClients = unqualify (mssRedundantClients status),
+      deletedClients = unqualify (mssDeletedClients status)
+    }
+  where
+    unqualify = UserClients . Map.findWithDefault mempty localDomain . qualifiedUserClients
+
+unqualifiedResponse :: Domain -> Union PostOtrResponses -> Union PostOtrResponsesUnqualified
```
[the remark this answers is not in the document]
pcapriotti: Unfortunately, `respond` isn't good enough, because it only gives you something of the form `a -> NS as` where `a` is a member of `as`. What we need here is `NS as -> NS bs`, given a way to map the types inside `as` to the types inside `bs`.

After going through the docs for SOP, I've found the `htrans` function, which seems to be general enough for this. It works using an arbitrary type-class which is able to convert `as` into `bs`. For this case, I've created an `Unqualify a b` type class which knows how to convert `a` into `b` given the local domain. Then after giving instances for all the types involved, including a trivial one for `Unqualify a a`, the conversion between the two unions can then be simply implemented in terms of `htrans`.

services/galley/src/Galley/Intra/Push.hs
```diff
@@ -112,7 +116,7 @@ data PushTo user = Push
     _pushNativePriority :: Maybe Gundeck.Priority,
     _pushAsync :: Bool,
     pushOrigin :: Maybe UserId,
-    pushRecipients :: List1 (RecipientBy user),
+    _pushRecipients :: List1 (RecipientBy user),
```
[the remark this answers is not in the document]
pcapriotti: The ones with a leading underscore get corresponding lenses thanks to the call to `makeLenses PushTo` below. We've added `pushRecipients` to the fields that have a lens, because we needed to use it when creating a push to a single client (by default `newPush` creates pushes for all clients).

There should be no problem in adding underscores to all fields (thus getting more consistent naming for fields vs lenses), but I guess it can be done in a separate PR.

services/galley/test/integration/API.hs
pcapriotti: I assumed it was already the case that only base64 encoded text is allowed here. The reason for the present change is just that the type for the message payload was changed from a base64-encoded text to a raw ByteString. Since now we are implementing the old API in terms of the new, we need to convert it back to a ByteString, only to be re-encoded into base64 when creating the events. Therefore, the payload needs to be valid base64 (or least be able to roundtrip back to itself after being decoded as base64) for these tests to work. They used to work before because nothing in the code was actually looking into the message at all, but they were still somewhat incorrect, because the message had to be valid base64 for the events to be parseable by clients.

services/galley/test/integration/API/Util.hs
pcapriotti: It is already pretty much like this. This function is only used within the tests for the new endpoint. Those tests include remote clients, so they all fail if we keep this assertion, because remote clients are (as of this PR) always treated as redundant.

So this PR actually changes the behaviour of the new endpoint, but it is still a refactoring PR as far as requests involving only local clients are concerned.

In any case, I don't think there should any danger in removing this check, because the old endpoint is tested using another utility function for assertions, which is also checking redundants.

pcapriotti:
> I think I don't understand this sentence. For integration tests with local users and local clients only, why can the reported redundant clients not be checked anymore? Maybe the sentence could be reworded to state that out of the 17 tests testing message sending, for 3 of them including remote users, that check isn't done. Or the tests could be adapted to continue doing the check where feasible?

I've updated the PR description to make it explicit that the only integration tests affected are the ones for the new endpoint. Hopefully that makes things clearer.

services/galley/test/integration/API.hs
jschaul: > I assumed it was already the case that only base64 encoded text is allowed here. The reason for the present change is just that the type for the message payload was changed from a base64-encoded text to a raw ByteString. Since now we are implementing the old API in terms of the new, we need to convert it back to a ByteString, only to be re-encoded into base64 when creating the events. Therefore, the payload needs to be valid base64 (or least be able to roundtrip back to itself after being decoded as base64) for these tests to work. They used to work before because nothing in the code was actually looking into the message at all, but they were still somewhat incorrect, because the message had to be valid base64 for the events to be parseable by clients.

Ah, makes sense, thanks for explaining!

services/galley/src/Galley/Intra/Push.hs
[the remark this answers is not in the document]
jschaul: I didn't see the makeLenses call, ok!

jschaul (approved):

PR #1637, opened 2021-06-30
Federation: Types for InwardErrors and federator refactoring

jschaul:
* Introduce a few error types in `InwardResponse`
* Improve readability of ExternalServer in Federator by use of Polysemy.Error also in callLocal
* This also solves an existing TODO whereby federator integration tests didn't work due to InwardResponses, whether an error or an expected return value, were parsed always as InwardResponseBody. This may have been an issue with mu-haskell when parsing (only needed in tests), since the behaviour when using `grpccurl` was correct. This is now sidestepped by using more than a simple string on errors.

This PR is in preparation to sanitize request paths against path traversal attacks (separate PR)

## Checklist


libs/wire-api-federation/src/Wire/API/Federation/Error.hs
```diff
@@ -130,6 +131,19 @@ federationUnavailable err =
     "federation-not-available"
     ("Local federator not available: " <> LT.fromStrict err)
 
+-- FUTUREWORK: Add Domain where the error occured to the WAI error responses here
```
pcapriotti, line 134: If I'm not mistaken, the domain is added by `federationErrorToWai`, so you can just remove this comment (and the one below).

services/federator/test/integration/Test/Federator/InwardSpec.hs
```diff
@@ -73,24 +76,60 @@ spec env =
...
+expectErr :: InwardErrorType -> InwardError -> TestFederator IO ()
+expectErr expectedType err = do
+  case inwardErrorType err of
+    t | t == expectedType -> pure ()
+    _ -> liftIO $ assertFailure $ "expected type '" <> show expectedType <> "' but got " <> show err
+
+asInwardBodyUnsafe :: (HasCallStack, Typeable a, FromJSON a) => GRpcReply InwardResponse -> a
+asInwardBodyUnsafe = either err id . asInwardBodyEither
+  where
+    err parserErr = error . unwords $ ["asInwardBodyUnsafe:"] <> [parserErr]
```
pcapriotti, line 98: Why not assert that the response is a body, instead of using `error`? I found that using `error` in tests can sometimes cause some failures not to be reported, because the test code can run into the error call before it has had a chance to process and print the failures.

pcapriotti (approved):
Looks good! Minor comments below.

libs/wire-api-federation/src/Wire/API/Federation/Error.hs
jschaul: Ah good point! I thought this wasn't done yet.

services/federator/test/integration/Test/Federator/InwardSpec.hs
jschaul: I tried both using `assertFailure` and `error`. The error messages are different, it worked fine locally for me, and I liked the display of using `error` 10% more. We use `error` quite a bit in other integration tests by using `responseJsonUnsafe` and companions.
I switched this PR to make use of assertFailure in https://github.com/wireapp/wire-server/pull/1637/commits/aacb5182eb3cc3cb2fdf304a1ec5ae438d8f6943 (if this is what you had in mind?), but in general I'm curious to see where you've seen this problem that failures are not reported?

services/federator/test/integration/Test/Federator/InwardSpec.hs
```diff
@@ -73,24 +76,70 @@ spec env =
...
+
+asInwardErrorEither :: HasCallStack => GRpcReply InwardResponse -> Either String InwardError
+asInwardErrorEither (GRpcOk (InwardResponseError err)) = Right err
+asInwardErrorEither (GRpcOk (InwardResponseBody bdy)) = Left ("expected InwardError, but got InwardResponseBody: " <> show bdy)
+asInwardErrorEither other = Left $ "GRpc call failed unexpectedly: " <> show other
+
+inwardBrigCall :: (MonadIO m, MonadHttp m, MonadReader TestEnv m, HasCallStack) => ByteString -> LBS.ByteString -> m (GRpcReply InwardResponse)
+inwardBrigCall requestPath payload = do
+  c <- viewFederatorExternalClient
+  let brigCall = Request Brig requestPath (LBS.toStrict payload) "foo.example.com"
```
mdimjasevic, line 131: What is this `"foo.example.com"` argument? Would it make sense to add a note?

mdimjasevic (approved):
This looks reasonable. I can't say I fully understand the changes so it'd be great if you could walk me through the PR.

services/federator/test/integration/Test/Federator/InwardSpec.hs
pcapriotti: I'm not sure where exactly (and it's gone now), but it was indeed a piece of test code using `responseJsonUnsafe`. It looked like:
```haskell
doStuff !!! do
  const 200 === statusCode
  blah === responseJsonUnsafe
```
and I only saw a parse failure error in a situation where the status code was *not* 200. This was confusing, because the assertion that parses the body comes after the one about the status code, but you only get to see the former.

CHANGELOG.md
```diff
@@ -38,6 +38,13 @@
   added to a new conversation via `POST /federation/register-conversation` (#1622).
 * [Federation] Adjust scripts under ./hack/federation to work with recent changes to the federation API (#1632).
 * Refactored Proteus endpoint to work with qualified users (#1634).
+* Refactored Federator InternalServer (#1637)
+
+### Internal Federation API changes
+
+(federation should not be in use anywhere yet, so this should not have any impact)
```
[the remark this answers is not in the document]
jschaul: This comment is meant to apply to the whole section "internal federation API changes". How can I make that clearer?

libs/wire-api-federation/src/Wire/API/Federation/GRPC/Types.hs
```diff
@@ -131,6 +133,30 @@ data ErrorPayload = ErrorPayload
   deriving (Typeable, Show, Eq, Generic, ToSchema Router "ErrorPayload", FromSchema Router "ErrorPayload")
   deriving (Arbitrary) via (GenericUniform ErrorPayload)
 
+type InwardErrorFieldMapping =
+  '[ "inwardErrorType" ':-> "type",
+     "inwardErrorMsg" ':-> "msg"
+   ]
```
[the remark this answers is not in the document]
jschaul: There is a page in the mu-haskell docs, here: https://higherkindness.io/mu-haskell/schema/#mapping-haskell-types that should explain. I'll add a comment here linking to those docs.

services/federator/test/integration/Test/Federator/InwardSpec.hs
jschaul: Ah I see, okay that's a good enough argument to not use error I suppose.

services/federator/test/integration/Test/Federator/InwardSpec.hs
jschaul: It's the originDomain, which is defined as part of Request data type, here: https://github.com/wireapp/wire-server/blob/develop/libs/wire-api-federation/src/Wire/API/Federation/GRPC/Types.hs#L135-L140. Maybe the object can be constructed using its names for clarity, good idea.

libs/wire-api-federation/src/Wire/API/Federation/GRPC/Types.hs
```diff
@@ -78,22 +77,25 @@ instance FromSchema Router "InwardResponse" InwardResponse where
...
      in TRecord (Field (FUnion protoChoice) :* Nil)
 
 instance FromSchema Router "OutwardResponse" OutwardResponse where
   fromSchema (TRecord (Field (FUnion protoChoice) :* Nil)) =
     case protoChoice of
       Z (FSchematic err) -> OutwardResponseError $ fromSchema err
-      S (Z (FPrimitive res)) -> OutwardResponseBody res
-      S (S x) -> case x of
+      S (Z (FSchematic err)) -> OutwardResponseInwardError $ fromSchema err
+      S (S (Z (FPrimitive res))) -> OutwardResponseBody res
```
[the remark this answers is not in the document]
jschaul: There are not yet, but I'm also confused by them, so roundtrip tests would be a good idea for sure. At the moment we only discover some of these things in integration tests (if those tests happen to use all code paths here). I added a FUTUREWORK for that.

PR #1638, opened 2021-07-01
Send messages to remotes

pcapriotti:
This propagates messages to all participants of a conversation, including remote ones. Note that we still cannot handle message sending to conversations in other domains, and attempting to do so still results in a `federation-not-implemented` error.

## Checklist

### DIFF
--- a/libs/wire-api/src/Wire/API/Message.hs
+++ b/libs/wire-api/src/Wire/API/Message.hs
@@ -34,6 +34,9 @@ module Wire.API.Message
     QualifiedNewOtrMessage (..),
     protoToNewOtrMessage,
 
+    -- * Protobuf messages
+    mkQualifiedOtrPayload,
+
     -- * Priority
     Priority (..),
 
@@ -60,11 +63,11 @@ module Wire.API.Message
   )
 where
 
-import Control.Lens (view, (?~))
+import Control.Lens (over, view, (.~), (<>~), (?~))
 import qualified Data.Aeson as A
 import qualified Data.ByteString.Lazy as LBS
 import Data.CommaSeparatedList (CommaSeparatedList (fromCommaSeparatedList))
-import Data.Domain (Domain, mkDomain)
+import Data.Domain (Domain, domainText, mkDomain)
 import Data.Id
 import Data.Json.Util
 import qualified Data.Map.Strict as Map
@@ -77,6 +80,7 @@ import Data.Serialize (runGetLazy)
 import qualified Data.Set as Set
 import qualified Data.Swagger as S
 import qualified Data.Swagger.Build.Api as Doc
+import qualified Data.Text.Read as Reader
 import qualified Data.UUID as UUID
 import Imports
 import qualified Proto.Otr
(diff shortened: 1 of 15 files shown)


libs/wire-api/src/Wire/API/Message.hs
```diff
@@ -206,6 +210,78 @@ protolensToQualifiedNewOtrMessage protoMsg = do
 protolensToClientId :: Proto.Otr.ClientId -> ClientId
 protolensToClientId = newClientId . view Proto.Otr.client
 
+--- functions to generate protobuf messages
+
+mkQualifiedOtrPayload :: ClientId -> [(Qualified UserId, ClientId, ByteString)] -> ByteString -> ClientMismatchStrategy -> Proto.Otr.QualifiedNewOtrMessage
+mkQualifiedOtrPayload sender recipients dat strat =
```
jschaul, line 216: Could these functions here have a unit round trip test? Given a few haskell values, make a proto OTR payload, then decode them again and expect the same?
(I'm fine to add that test in a follow-up PR also; mostly to guard against future regressions)

services/galley/src/Galley/API/Message.hs
```diff
@@ -213,26 +220,110 @@ postQualifiedOtrMessage senderType sender mconn convId msg = runUnionT $ do
...
+  fmap (setOf ((itraversed <. folded) . withIndex . to (\(d, (u, c)) -> (d, u, c)))) . flip Map.traverseWithKey messageMap $ \d msgs ->
+    if d == localDomain
+      then do
+        sendLocalMessages now sender senderClient mconn conv localMemberMap metadata msgs
+        pure mempty
+      else do
+        x <- sendRemoteMessages d now sender senderClient conv metadata msgs
+        case x of
+          Nothing -> pure mempty
+          Just e -> do
```
jschaul, line 263: This code could be refactored a little for improved readability using smaller named helper functions. the lines 253 and 254 look a little bit like they partition the map of domain/user/clients into local and remote, so perhaps a helper function could be created here?

services/galley/src/Galley/API/Message.hs
```diff
@@ -213,26 +220,110 @@ postQualifiedOtrMessage senderType sender mconn convId msg = runUnionT $ do
...
+            FederatedGalley.rmData = Just (toBase64Text (mmData metadata)),
+            FederatedGalley.rmSender = sender,
+            FederatedGalley.rmSenderClient = senderClient,
+            FederatedGalley.rmConversation = conv,
+            FederatedGalley.rmPriority = mmNativePriority metadata,
+            FederatedGalley.rmPush = mmNativePush metadata,
+            FederatedGalley.rmTransient = mmTransient metadata,
+            FederatedGalley.rmRecipients = UserClientMap rcpts
+          }
+  -- TODO: we should not need to pass the local domain to the RPC
```
jschaul, line 324:
```suggestion
  -- FUTUREWORK: we should not need to pass the local domain to the RPC
```

libs/wire-api/src/Wire/API/Message.hs
jschaul: Update: Ah I see these functions were moved from the tests into wire-api. Still, some roundtrip tests might be nice as a FUTUREWORK.

jschaul (approved):
🎆

services/galley/src/Galley/API/Message.hs
pcapriotti: I moved the local/remote distinction logic to a separate function with a guard, and inline the remote error handling in the `sendMessagesRemote` function.

I've also used the opportunity to get rid of the extra flattening/unflattening of `failedToSend`.

Hopefully things are a bit clearer now.

libs/wire-api/src/Wire/API/Message.hs
pcapriotti: Yes, definitely. I added a FUTUREWORK.

services/galley/src/Galley/API/Message.hs
jschaul: Thank you! Looks more readable now.

PR #1639, opened 2021-07-02
Custom Encoded Union

pcapriotti:
This introduces a `CustomEncodedUnion` newtype, to be used mainly with `deriving via` to get JSON instances for `Union as` types. This is useful to expose functions returning unions (such as Servant handlers) through the RPC interface, which cannot handle arbitrary status codes.

The way the instances are generated assumes that the JSON value corresponding to a union is an object with a "status" field containing the status code of a particular case in the union (Servant statically enforces union types to have distinct status codes) and a "value" field containing the JSON serialisation of the corresponding value.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,7 @@
   added to a new conversation via `POST /federation/register-conversation` (#1622).
 * [Federation] Adjust scripts under ./hack/federation to work with recent changes to the federation API (#1632).
 * Refactored Proteus endpoint to work with qualified users (#1634).
+* Added a CustomEncodedUnion newtype to simplify the implementation of RPCs returning unions (#1639).
 
 ## Documentation
 
--- a/libs/wire-api-federation/package.yaml
+++ b/libs/wire-api-federation/package.yaml
@@ -35,6 +35,7 @@ dependencies:
 - servant >=0.16
 - servant-client
 - servant-client-core
+- sop-core
 - template-haskell
 - text >=0.11
 - time >=1.8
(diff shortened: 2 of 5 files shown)


akshaymankar (commented):
@pcapriotti Can you please give more context on this? Are we planning to communicate between backends using json encoded HTTP statuses?

pcapriotti:
> @pcapriotti Can you please give more context on this? Are we planning to communicate between backends using json encoded HTTP statuses?

This is specifically for the `send-message` RPC (#1609). Since that is mostly a proxy call to the message sending endpoint on the other backend, it is useful to reuse its return value, which is a `Union`. However, we cannot simply use `UVerb` in the federator API, because we assume that it always returns 200 on success. With this PR, one can simply use `Post` and set the return type to be a `CustomEncodedUnion`. The response is then always 200, as required, and the actual status is encoded in the JSON body.

I hope this makes sense.

akshaymankar:
> This is specifically for the `send-message` RPC (#1609). Since that is mostly a proxy call to the message sending endpoint on the other backend, it is useful to reuse its return value, which is a `Union`. However, we cannot simply use `UVerb` in the federator API, because we assume that it always returns 200 on success. With this PR, one can simply use `Post` and set the return type to be a `CustomEncodedUnion`. The response is then always 200, as required, and the actual status is encoded in the JSON body.
> 
> I hope this makes sense.

Thanks, this makes sense. However, I remember we decided to not have the federator forward HTTP status from remote was because we wanted to not use HTTP statuses as way of communication. So, doing this would just be getting around our previous decision. I think it is still a good idea to not rely on HTTP statuses, in this case it would be better to just create another sum type which has roundtripping To/FromJSON instances and just write a translator for this new type to our Union in galley.

pcapriotti:
> Thanks, this makes sense. However, I remember we decided to not have the federator forward HTTP status from remote was because we wanted to not use HTTP statuses as way of communication. So, doing this would just be getting around our previous decision. I think it is still a good idea to not rely on HTTP statuses, in this case it would be better to just create another sum type which has roundtripping To/FromJSON instances and just write a translator for this new type to our Union in galley.

But that's pretty much what this PR does. It just uses the status codes as the "tags" to distinguish which case of the sum type we're in. Are you saying that we should be using something other than status codes for tags, but keep the overall approach? Is it worth to try to do this generically for union types like this PR, or should we just make an ad-hoc wrapper for the messaging type?

pcapriotti:
This can probably be closed now.

PR #1640, opened 2021-07-02
Federation: Qualify Conversation IDs

mdimjasevic:
This patch makes a conversation ID qualified. The now makes it impossible to parse from JSON with the old format because a new field `qualified_id` is expected. Conversely, serialising to JSON retains the old `id` field, but in addition also includes `qualified_id`.

* The gist of the PR is in [Conversation.hs](https://github.com/wireapp/wire-server/pull/1640/files#diff-a3da6271c27cfbacc7d2bb7604cbe22f4a8cb336e9bb4a0807b443ac26290e92); the rest of the PR follows mechanically from it.
* Lots of golden tests had to be updated
* Some integration tests had to be updated

## Checklist


pcapriotti (approved):
I haven't looked at all the changes too carefully, but it looks good. One thing to be generally aware of is that this in an incompatible change at the level of the federation API. Not a problem at this stage, I think, but it might be worth pointing this out in the CHANGELOG.

services/galley/test/integration/API.hs
```diff
@@ -1101,8 +1102,11 @@ listConvsPagingOk = do
...
   walk carl [3, 3, 3, 3, 1] -- 11 (group) + 1 (1:1) + 1 (self)
   where
+    walk :: Foldable t => UserId -> t Int -> TestM ()
     walk u = foldM_ (next u 3) Nothing
+    next :: UserId -> Int32 -> Maybe ConvId -> Int -> TestM (Maybe ConvId)
     next u step start n = do
+      -- TODO(md): Check if the FUTUREWORK note is addressed.
       -- FUTUREWORK: support an endpoint to get qualified conversation IDs
       -- (without all the conversation metadata)
       r1 <- getConvIds u (Right <$> start) (Just step) <!! const 200 === statusCode
```
mdimjasevic, line 1112: @jschaul , I am not sure what you had in mind here. Can you please clarify?

jschaul: this futurework comment is unrelated to this PR. It is about creating a new endpoint similar to `GET ​/conversations​/ids Get all conversation IDs.` ([swagger](https://staging-nginz-https.zinfra.io/api/swagger-ui/#/default/get_conversations_ids)).

CHANGELOG.md
```diff
@@ -18,6 +18,8 @@
 
 ## API Changes
 
+* All endpoints that fetch conversation details now also include a new key
+  `qualified_id` for a qualified convevrsation ID (#1640)
```
jschaul, line 22:
```suggestion
  `qualified_id` for a qualified conversation ID (#1640)
```

jschaul (commented):
The golden tests could add this counter-example:

- parsing a json object without a qualified id fails (probably could be added in the https://github.com/wireapp/wire-server/blob/develop/libs/wire-api/test/unit/Test/Wire/API/Golden/FromJSON.hs test)

Is it correct that the conversation data type is always only written, an never parsed by galley/brig/etc ? If it's ever parsed in intra calls, then there might be a deployment order.

Otherwise change looks good.

mdimjasevic:
> The golden tests could add this counter-example:
> 
>     * parsing a json object without a qualified id fails (probably could be added in the https://github.com/wireapp/wire-server/blob/develop/libs/wire-api/test/unit/Test/Wire/API/Golden/FromJSON.hs test)
 
Thank you for the suggestion! I just added this test.

mdimjasevic:
> Is it correct that the conversation data type is always only written, an never parsed by galley/brig/etc ? If it's ever parsed in intra calls, then there might be a deployment order.

@jschaul , I scanned the code base and couldn't find a site where it is parsed from JSON.

jschaul (approved):
Looks good (once CI agrees)!

PR #1642, opened 2021-07-05
Rewrite Team Feature Endpoints to Servant

mdimjasevic:
The PR rewrites team feature endpoints into the Servant web framework.

* Some public PUT endpoints were removed per https://wearezeta.atlassian.net/browse/SQSERVICES-463.
* Tests are added that check if a user is authorized to access GET endpoints of team features.

## Checklist


libs/wire-api/src/Wire/API/Team/Feature.hs
```diff
@@ -197,6 +206,7 @@ type family TeamFeatureStatus (a :: TeamFeatureName) :: * where
 type FeatureHasNoConfig (a :: TeamFeatureName) = (TeamFeatureStatus a ~ TeamFeatureStatusNoConfig) :: Constraint
 
 -- if you add a new constructor here, don't forget to add it to the swagger (1.2) docs in "Wire.API.Swagger"!
+-- TODO: Delete!
```
fisx, line 209: if you want to merge this `TODO` into develop, you should make it a `FUTUREWORK`.  also, could you elaborate a little on what and why?  (`Doc.Model` is swagger 1.2, and the new end-points are documented in 2.0 now, something like that.)

Or can you just remove it in this PR?

(Same questions apply to a few comments above and below.)

libs/wire-api/src/Wire/API/Team/Feature.hs
```diff
@@ -212,60 +222,48 @@ newtype TeamFeatureStatusNoConfig = TeamFeatureStatusNoConfig
...
+  schema =
+    object "TeamFeatureStatusNoConfig" $
+      TeamFeatureStatusNoConfig
+        <$> tfwoStatus .= field "status" schema
 
 ----------------------------------------------------------------------
 -- TeamFeatureStatusWithConfig
 
+-- | FUTUREWORK: Does it ever happen that there is a config for a disabled
+-- feature? If not this type should be made isomorphic to 'Maybe cfg'.
```
fisx, line 243: good catch, but:

```suggestion
-- | The support for disabled features with configs is intentional:
-- for instance, we want to be able to keep the config of a feature
-- that is turned on and off occasionally, and so not force the admin
-- to recreate the config every time it's turned on.
```

services/galley/src/Galley/API/Public.hs
```diff
@@ -92,7 +92,32 @@ servantSitemap =
         GalleyAPI.getTeamConversation = Teams.getTeamConversation,
         GalleyAPI.deleteTeamConversation = Teams.deleteTeamConversation,
         GalleyAPI.postOtrMessageUnqualified = Update.postOtrMessageUnqualified,
-        GalleyAPI.postProteusMessage = Update.postProteusMessage
+        GalleyAPI.postProteusMessage = Update.postProteusMessage,
+        GalleyAPI.teamFeatureStatusSSOGet =
```
fisx, line 96: Why is there no "PUT" counterpart for the sso feature?  The old code looks to me like it has that, but I vaguely remember that we did only allow it in the internal API.  (Same for a few more features below.)

libs/wire-api/src/Wire/API/Team/Feature.hs
```diff
@@ -212,60 +222,48 @@ newtype TeamFeatureStatusNoConfig = TeamFeatureStatusNoConfig
   { tfwoStatus :: TeamFeatureStatusValue
   }
   deriving newtype (Eq, Show, Generic, Typeable, Arbitrary)
+  deriving (ToJSON, FromJSON, S.ToSchema) via (Schema TeamFeatureStatusNoConfig)
 
+-- TODO: Delete
 modelTeamFeatureStatusNoConfig :: Doc.Model
 modelTeamFeatureStatusNoConfig = Doc.defineModel "TeamFeatureStatusNoConfig" $ do
-  Doc.description $ "Configuration for a team feature that has no configuration"
+  Doc.description "Configuration for a team feature that has no configuration"
```
fisx, line 230:
```suggestion
  Doc.description "Team feature that has no configuration beyond the boolean on/off switch."
```

(I know it has been like this, and I may have written this, but I think this is a good time to fix that, too.)

(Come to think of it should this be added to the schema?)

fisx (changes requested):
The new code looks so much better!  But I do have questions (at least one of them may be just because it's late).

libs/wire-api/src/Wire/API/Team/Feature.hs
pcapriotti: I think it should be fine to just delete them in this PR. I don't think they are used anywhere else.

pcapriotti (commented):
Like Matthias noted, it seems the SSO feature needs a PUT. Other than that, looks good to me.

libs/wire-api/src/Wire/API/Team/Feature.hs
akshaymankar: @fisx this can be dealt with by writing a lenient parser which just ignores the config. Does our code ever need to access the config why a feature is disabled?

libs/wire-api/src/Wire/API/Team/Feature.hs
mdimjasevic: I'll remove the comment from here, but I will keep it in that other PR for classified domains and take care of that there (I am not sure at this moment if I can safely delete these).

libs/wire-api/src/Wire/API/Team/Feature.hs
mdimjasevic: @fisx's argument makes sense. @akshaymankar , if you are OK with it, i.e., to keep the configuration in the config file when the feature is disabled, I would accept the code suggestion by Matthias.

services/galley/src/Galley/API/Public.hs
mdimjasevic: One intermediate step in rewriting these endpoints to Servant was to get rid of public endpoints that are not meant for the public, and these were always setters. By inspecting the code, I and whoever paired with me at that point concluded that SSO should not have a public endpoint, and similarly for some other features. Note that they are still available as internal endpoints at `PUT /i/teams/:tid/feature...`.
If this is not a good conclusion, I will be happy to reintroduce removed public endpoints for any/all team features.

libs/wire-api/src/Wire/API/Team/Feature.hs
mdimjasevic: I got rid of these models for Swagger 1.2 and removed TODOs.

libs/wire-api/src/Wire/API/Team/Feature.hs
mdimjasevic: I removed these Swagger 1.2 entries in the module.

services/galley/src/Galley/API/Public.hs
mdimjasevic: @fisx and @pcapriotti , for the context of why certain PUT endpoints were removed, see https://wearezeta.atlassian.net/browse/SQSERVICES-463. I hope that clarifies it!

pcapriotti (approved):

services/galley/src/Galley/API/Public.hs
mdimjasevic: @fisx , I will mark this as resolved per SQSERVICES-463, but if you feel like this is not resolved, please let me know!

fisx: That means https://wearezeta.atlassian.net/browse/SQSERVICES-463 can be closed with this PR, right?

Nice side-effect of the refactoring!  :)

fisx (approved):
> Like Matthias noted, it seems the SSO feature needs a PUT. Other than that, looks good to me.

Actually, [the link in the issue](https://github.com/wireapp/wire-server/blob/develop/libs/galley-types/src/Galley/Types/Teams.hs#L332) Marko quoted above convinced me otherwise.

There are two things I'd like to see, but maybe in a separate PR:

- document why we keep the swagger1.2 around (for stern, correct?  maybe we could even move it to stern, then make that more obvious.)
- as joe noted in https://wearezeta.atlassian.net/browse/SQSERVICES-463, add tests for all end-points.  (maybe we can even do that in a way that throws a compiler warning if we add a new feature config and forget to add a test?)

mdimjasevic:
> There are two things I'd like to see, but maybe in a separate PR:

@fisx , I will do that in a separate PR. Thank you for re-reviewing!

services/galley/src/Galley/API/Public.hs
mdimjasevic: > That means https://wearezeta.atlassian.net/browse/SQSERVICES-463 can be closed with this PR, right?
> 
> Nice side-effect of the refactoring! :)

Yes, that is correct.

fisx:
ah, good catch, thanks.  still :+1:  from me.

PR #1643, opened 2021-07-05
Change conversationCodeURI in galley.yaml.

fisx:
https://wearezeta.atlassian.net/browse/SQSERVICES-535

**Mote context:** (thanks @Yserz, and please correct me if I got anything wrong)

The old url was pointing to the webapp, and that will keep working as well as it works now, namely:

- webapp can create and consume those links;
- mobile clients can create, but not consume them.

In order to allow mobile clients to join conversations with the logged-in user (not a guest, that was harder to implement), the backend should now create new guest-links that point to the account pages under the new path.  For web, this does not change anything, the new link will work as well as the old.  For mobile, this will make things work via the logged-in user.

This PR makes sure that all tests, on-prem installations, ... can make this change.  Until they do, things will work as well as (but not better than) before.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,10 @@
 
 # Unreleased
 
+## Release Notes
+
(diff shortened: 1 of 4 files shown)


arianvp (approved):

arianvp:
Though please also change:

https://github.com/wireapp/wire-server-deploy/tree/develop/values/wire-server
https://github.com/zinfra/cailleach/tree/master/environments

PR #1644, opened 2021-07-06
Golden protobuf tests for QualifiedNewOtrMessage

pcapriotti:
This introduces golden tests for protobuf serialisation/deserialisation and runs them on `QualifiedNewOtrMessage`. It also refactors the protobuf serialisation utilities in `Wire.API.Message` and introduces a `ToProto` typeclass alongside the existing `FromProto`.

There are 20 test objects of type `QualifiedNewOtrMessage`, which have been generated using its `Arbitrary` instance (with some manual cleanup).

## Checklist


libs/wire-api/test/unit/Test/Wire/API/Golden/Runner.hs
```diff
@@ -57,6 +64,44 @@ testObject obj path = do
...
+protoTestObject ::
+  forall m a.
+  (Typeable a, ToProto a, FromProto a, Eq a, Show a, Show m, Eq m, Message m) =>
+  a ->
+  FilePath ->
+  IO Bool
+protoTestObject obj path = do
+  let actual = toProto obj
+  msg <- assertRight (decodeMessage @m (LBS.toStrict actual))
+  let pretty = render (pprintMessage msg)
```
akshaymankar, line 85: I didn't know there was pretty print for protobuf, this is sweet :smile:

libs/wire-api/test/unit/Main.hs
```diff
@@ -14,18 +14,21 @@
 --
 -- You should have received a copy of the GNU Affero General Public License along
 -- with this program. If not, see <https://www.gnu.org/licenses/>.
+{-# OPTIONS_GHC -Wno-unused-imports #-}
```
akshaymankar, line 17: This shouldn't be required, right?

akshaymankar (approved):

libs/wire-api/test/unit/Main.hs
pcapriotti: Ah, thanks, good catch. I just added it while trying some things out.

libs/wire-api/test/unit/Test/Wire/API/Golden/Runner.hs
pcapriotti: Yes, I was really happy to find it. Otherwise, having golden files with binary content would make it quite hard to handle changes in the protocol.

PR #1645, opened 2021-07-06
Document Servant endpoint errors for Galley routes

pcapriotti:
This introduces a `CanThrow` type to be used in Servant APIs, declaring exceptions that can be thrown by the handler. The `CanThrow` type is ignored by Servant, but results in a new response being produced in Swagger.

This PR also arranges for specialised Swagger schemas to be displayed in the response section of endpoints with possible error responses. The specialised schemas have better example and more restrictions on the possible values of the error label. 

## Checklist

### DIFF
--- a/libs/wire-api/src/Wire/API/Routes/Public/Galley.hs
+++ b/libs/wire-api/src/Wire/API/Routes/Public/Galley.hs
@@ -34,7 +34,7 @@ import Servant.Swagger.Internal
 import Servant.Swagger.Internal.Orphans ()
 import qualified Wire.API.Conversation as Public
 import qualified Wire.API.Conversation.Role as Public
-import Wire.API.ErrorDescription (ConversationNotFound, UnknownClient)
+import Wire.API.ErrorDescription
 import qualified Wire.API.Event.Conversation as Public
 import qualified Wire.API.Message as Public
 import Wire.API.Routes.Public (EmptyResult, ZConn, ZUser)
@@ -155,12 +155,12 @@ data Api routes = Api
         :> Post '[Servant.JSON] (Public.ConversationList Public.Conversation),
     -- This endpoint can lead to the following events being sent:
     -- - ConvCreate event to members
-    -- FUTUREWORK: errorResponse Error.notConnected
-    --             errorResponse Error.notATeamMember
+    -- FUTUREWORK: errorResponse Error.notATeamMember
     --             errorResponse (Error.operationDenied Public.CreateConversation)
     getConversationByReusableCode ::
       routes
         :- Summary "Get limited conversation information by key/code pair"
+        :> CanThrow NotConnected
         :> ZUser
         :> "conversations"
         :> "join"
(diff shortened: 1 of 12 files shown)


libs/wire-api/src/Wire/API/Routes/Public/Galley.hs
```diff
@@ -155,22 +155,25 @@ data Api routes = Api
...
         :> "join"
         :> QueryParam' [Required, Strict] "key" Code.Key
         :> QueryParam' [Required, Strict] "code" Code.Value
         :> Get '[Servant.JSON] Public.ConversationCoverView,
-    -- FUTUREWORK: potential errors: codeNotFound, convNotFound, notATeamMember, convAccessDenied
     createGroupConversation ::
       routes
         :- Summary "Create a new conversation"
+        :> CanThrow NotATeamMember
+        :> CanThrow CodeNotFound
```
jschaul, line 174: Hm, isn't the CodeNotFound error thrown on a different endpoint? When requesting or trying to join a conversation which has a conversation code; rather than when creating a group conversation? Same for the ConvAccessDenied. I think the FUTUREWORK comments apply to the handle above the comment, not below the comment.

Maybe you want to swap the error list from createGroupConversation with the one from getConversationByReusableCode.

jschaul (approved):
The swagger looks much better than before. It's generally nice to now have to wield around all the error information in all of the Handler types; it has the downside of potentially being out of sync, but I think having *some* way to document errors in swagger is a big improvement over having no or very little error documentation.

jschaul:
It might be nice to have a `:> CanThrow [Err1, Err2, ...]` syntax at some point in the future, but this is a good incremental step.

jschaul:
labels

Do you think it might be feasible to include the actual more-stable labels (e.g. 'no-conversation') as part of the displayed description, rather than only the message?

libs/wire-api/src/Wire/API/Routes/Public/Galley.hs
pcapriotti: Yes, you're right! I guess the comments went out of whack there after some refactoring. Having the list of exceptions in the type should make this sort of mistake a bit less likely.

pcapriotti:
> It might be nice to have a `:> CanThrow [Err1, Err2, ...]` syntax at some point in the future, but this is a good incremental step.

I considered it, but then decided it wouldn't be worth the extra lines of code. If you have a long list of exceptions, you're going to want to split them in multiple lines anyway, so nothing is gained in terms of vertical space.

> Do you think it might be feasible to include the actual more-stable labels (e.g. 'no-conversation') as part of the displayed description, rather than only the message?

Yeah, that should be possible. Note that the labels are already there as part of the example, as well as the swagger schema (when you click on "Model"), though.

pcapriotti:
How about this?
shot

jschaul:
> How about this?
> shot

Yes, that looks good!

PR #1646, opened 2021-07-06
Federation: sanitize paths to avoid path traversal attacks

jschaul:
At the moment we use the convention that all RPCs between federator -> brig/galley/etc use the "federation" path prefix. This PR adds validation to ensure other paths cannot be called, as that would expose non-federation APIs (including internal, non-authenticated ones) to the world.

See also the discussion in https://wearezeta.atlassian.net/wiki/spaces/CORE/pages/224166764/Limiting+access+to+federation+endpoints

## Checklist

 - [x] rebase to develop after #1637 has landed.

### DIFF
--- a/services/federator/src/Federator/Validation.hs
+++ b/services/federator/src/Federator/Validation.hs
@@ -29,11 +29,10 @@ import Data.Domain (Domain, domainText, mkDomain)
 import Data.String.Conversions (cs)
 import Federator.Options
 import Imports
-import Network.URI
 import Polysemy (Members, Sem)
 import qualified Polysemy.Error as Polysemy
 import qualified Polysemy.Reader as Polysemy
-import qualified URI.ByteString as URIB
+import URI.ByteString
 import Wire.API.Federation.GRPC.Types
 
 -- | Validates an already-parsed domain against the allowList using the federator
@@ -71,16 +70,16 @@ sanitizePath originalPath = do
   when (BS.length originalPath > 200) $
     throwInward IForbiddenEndpoint "path too long"
   -- we parse the path using the URI.ByteString module to make use of its normalization functions
-  uriRef <- case URIB.parseRelativeRef URIB.strictURIParserOptions originalPath of
+  uriRef <- case parseRelativeRef strictURIParserOptions originalPath of
     Left err -> throwInward IForbiddenEndpoint (cs $ show err <> cs originalPath)
     Right ref -> pure ref
 
   -- we don't expect any query parameters or other URL parts other than a plain path
-  when (URIB.queryPairs (URIB.rrQuery uriRef) /= []) $
+  when (queryPairs (rrQuery uriRef) /= []) $
     throwInward IForbiddenEndpoint "query parameters not allowed"
-  when (isJust (URIB.rrFragment uriRef)) $
+  when (isJust (rrFragment uriRef)) $
     throwInward IForbiddenEndpoint "fragments not allowed"
-  when (isJust (URIB.rrAuthority uriRef)) $
+  when (isJust (rrAuthority uriRef)) $
     throwInward IForbiddenEndpoint "authority not allowed"
 
   -- Perform these normalizations:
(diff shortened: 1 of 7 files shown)


services/federator/test/integration/Test/Federator/InwardSpec.hs
```diff
@@ -84,6 +84,24 @@ spec env =
...
+      runTestFederator env $ do
+        let o = object ["name" .= ("fakeNewUser" :: Text)]
+        err <- asInwardErrorUnsafe <$> inwardBrigCall "i/users" (encode o)
+        expectErr IForbiddenEndpoint err
+
+    it "should only accept /federation/ paths (also when there are .. segments)" $
+      runTestFederator env $ do
+        let o = object ["name" .= ("fakeNewUser" :: Text)]
+        err <- asInwardErrorUnsafe <$> inwardBrigCall "federation/../i/users" (encode o)
+        expectErr IForbiddenEndpoint err
```
jschaul, line 103: dear reviewers, I'd love ideas on what other things to test. Would any property tests make sense here?

kkohbrok (commented):
Looks solid already! Another common vector for attacks would be encoding. I assume that `URIB.parseRelativeRef` handles the decoding of the byte string, so we could test the normalization function for its handling of [percent encoding](https://en.wikipedia.org/wiki/Directory_traversal_attack#Percent_encoding_in_URIs). I don't have a lot of experience in webapp security, but it seems to be a common enough attack vector to have some tests for it, i.e. to make sure that the normalizer properly handles percent decoding before sanitization.

services/federator/src/Federator/Validation.hs
```diff
@@ -60,3 +63,30 @@ validateDomain unparsedDomain = do
...
+  -- - eXaMpLe.org -> example.org
+  -- - If the scheme is known and the port is the default (e.g. 80 for http) it is removed.
+  -- - If the path is empty, set it to /
+  -- - Rewrite path from /foo//bar///baz to /foo/bar/baz
+  -- - Sorts parameters by parameter name
+  -- - Remove dot segments as per RFC3986 Section 5.2.4
+  let normalized = URIB.normalizeURIRef' URIB.aggressiveNormalization uriRef
+  -- next, we use the Network.URI library to get path segments after normalization,
+  -- and ensure the first segment in the path is "federation"
+  uri <- case parseURIReference (cs normalized) of
```
pcapriotti, line 85: Here you are converting the parsed URI back to a `String`, then reparsing it. Wouldn't it be better to just take the path (as `ByteString`) and directly check that it begins with (ascii-encoded) "federation/"?

services/federator/test/integration/Test/Federator/InwardSpec.hs
pcapriotti: I think having property tests for `sanitizePath` would be useful, but of course only insofar as the generator produces realistic URL inputs. As for ad-hoc test cases, you could maybe add some with non-ascii character (both percent-encoded and not), query string and/or fragments.

pcapriotti (approved):
Looks good!

services/federator/test/integration/Test/Federator/InwardSpec.hs
```diff
@@ -84,6 +84,24 @@ spec env =
         err <- asInwardErrorUnsafe <$> inwardBrigCall "federation/this-endpoint-does-not-exist" (encode Aeson.emptyObject)
         expectErr IInvalidEndpoint err
 
+    it "should only not accept invalid paths" $
```
mdimjasevic, line 87:
```suggestion
    it "should not accept invalid paths" $
```

services/federator/test/integration/Test/Federator/InwardSpec.hs
mdimjasevic: I agree, some (unit) tests for `sanitizePath` would be good to have, including one that has "federation" as the only path segment.

mdimjasevic (approved):
It looks good. I haven't checked, but I suppose `sanitizePath` is already used for all federator paths?

services/federator/src/Federator/Validation.hs
jschaul: Done.

services/federator/test/integration/Test/Federator/InwardSpec.hs
jschaul: I added a bunch of unit tests, take a look.

pcapriotti: Nice! :+1:

jschaul:
I added a range of tests, and also changed the implementation to guard against more cases. Merging this PR now, but in case someone still has other ideas, feel free to comment.

PR #1657, opened 2021-07-13
Document more error responses in Swagger

pcapriotti:
This is a followup to #1645, adding documentation for both static and dynamic errors thrown from Brig handlers. It uses `MultiVerb`, introduced by #1649, and the `CanThrow` type from #1645. It also removes the "empty error for legacy reason" kludge introduced by #1603, and replaces it with a proper servant error returning a plain text response.

## Checklist

-  [x] If this PR creates a new endpoint, or adds a new configuration flag, the endpoint / config-flag checklist (see Wire-employee only backend [wiki page](https://github.com/zinfra/backend-wiki/wiki/Checklists)) has been followed.


libs/wire-api/src/Wire/API/Routes/Public/Brig.hs
```diff
@@ -45,7 +53,23 @@ import Wire.API.UserMap
 
 type MaxUsersForListClientsBulk = 500
 
-type CheckUserExistsResponse = [EmptyResult 200, EmptyResult 404]
+type UserExistsVerb =
```
jschaul, line 56: There is a `UserExistVerb` in line 65 which seems to be exactly the same type. Remove duplicate?

jschaul (approved):
Looks good!

libs/wire-api/src/Wire/API/Routes/Public/Brig.hs
pcapriotti: Yes, thanks for catching this. It was a mistake in a merge.

PR #1660, opened 2021-07-13
makefile target for helm templating charts

jschaul:
This allows debugging helm templating issues more easily without installing everything for a quicker feedback loop. (e.g. in #1626 this would have saved some debugging time)

## Checklist

   - this doesn't need any mention in the changelog.


mdimjasevic (approved):
I tested this locally and it works.

PR #1665, opened 2021-07-14
download gpg keys for nginz

jschaul:
On fresh development environments, nginx cannot be compiled without the gpg signature verification in place. Before, this had to be done manually; now GPG keys are downloaded once the first time.

Only relevant when compiling nginz locally.

## Checklist

### DIFF
--- a/services/nginz/Makefile
+++ b/services/nginz/Makefile
@@ -114,8 +114,8 @@ zwagger-ui/swagger-ui: $(SWAGGER_BUNDLE)
 $(SWAGGER_BUNDLE):
 	curl -L https://github.com/swagger-api/swagger-ui/archive/v$(SWAGGER_VERSION).tar.gz -o $(SWAGGER_BUNDLE)
 
-src: $(NGINX_BUNDLE)
-	#Find keys on https://nginx.org/en/pgp_keys.html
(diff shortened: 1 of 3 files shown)


mdimjasevic (approved):

services/nginz/Makefile
```diff
@@ -114,8 +114,8 @@ zwagger-ui/swagger-ui: $(SWAGGER_BUNDLE)
 $(SWAGGER_BUNDLE):
 	curl -L https://github.com/swagger-api/swagger-ui/archive/v$(SWAGGER_VERSION).tar.gz -o $(SWAGGER_BUNDLE)
 
-src: $(NGINX_BUNDLE)
-	#Find keys on https://nginx.org/en/pgp_keys.html
+src: $(NGINX_BUNDLE) keys-downloaded
```
arianvp, line 117: Can't we make the rule depend on the GPG keyfile name? Then Make will automatically make sure it's not re-downloaded without us having to touch files manually?

arianvp: something like:

```suggestion
src: $(NGINX_BUNDLE)  $(NGINX_BUNDLE).asc

$(NGINX_BUNDLE).asc:
	./download-gpg-keys.sh
```

jschaul:
obsolete.

PR #1668, opened 2021-07-15
Remove old federation error code

pcapriotti:
Just removing an unused federation error. I haven't added a CHANGELOG entry for this, should I?

## Checklist

 - [ ] Section *Unreleased* of **CHANGELOG.md** contains the following bits of information:
   - [ ] A line with the title and number of the PR in one or more suitable sub-sections.

### DIFF
--- a/services/galley/src/Galley/API/Error.hs
+++ b/services/galley/src/Galley/API/Error.hs
@@ -19,10 +19,7 @@ module Galley.API.Error where
 
 import Control.Monad.Catch (MonadThrow (..))
 import Data.Domain (Domain, domainText)
-import Data.Id (Id)
-import Data.List.NonEmpty (NonEmpty)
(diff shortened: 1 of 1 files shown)


mdimjasevic (approved):
You noticed yesterday this one isn't used, so go ahead with removing it.

mdimjasevic:
> Just removing an unused federation error. I haven't added a CHANGELOG entry for this, should I?

I would be fine without it, but whoever will be cutting a release might be wondering what was this PR about such that it has no entry in the changelog.

akshaymankar:
> You noticed yesterday this one isn't used, so go ahead with removing it.

I think it is fine to not add this. If I was cutting the release the old way (i.e. figuring out changelog from git log), I'd have definitely skipped this one.

PR #1669, opened 2021-07-15
QualifiedCapture type for Servant endpoints

pcapriotti:
This introduces a type `QualifiedCapture`, to be used exactly like `Capture`, but also capturing a `Domain` and providing a `Qualified` value to the handler.

## Checklist


mdimjasevic (approved):
Nice! It makes sense to put together a domain and an ID like this; otherwise they seem unrelated.

jschaul (approved):

PR #1670, opened 2021-07-15
Refactor createUser function

smatting:
This PR is a pure refactoring of `createUser` from `Brig.API.User`. Motivation for this PR was to make function more readable.

## Checklist

### DIFF
--- a/services/brig/src/Brig/API/User.hs
+++ b/services/brig/src/Brig/API/User.hs
@@ -183,7 +183,16 @@ createUser new = do
   (email, phone) <- validateEmailAndPhone new
 
   -- get invitation and existing account
-  (newTeam, teamInvitation, tid) <- handleTeam (newUserTeam new) (userEmailKey <$> email)
+  (newTeam, teamInvitation, tid) <-
+    case (newUserTeam new, userEmailKey <$> email) of
+      (Just (NewTeamMember i), e) ->
+        findTeamInvitation e i
+        >>= return . \case
+          Just (inv, info, tid) -> (Nothing, Just (inv, info), Just tid)
+          Nothing -> (Nothing, Nothing, Nothing)
+      (Just (NewTeamCreator t), _) -> (Just t,Nothing,) <$> (Just . Id <$> liftIO nextRandom)
+      (Just (NewTeamMemberSSO tid), _) -> pure (Nothing, Nothing, Just tid)
+      (Nothing, _) -> return (Nothing, Nothing, Nothing)
   let mbInv = Team.inInvitation . fst <$> teamInvitation
   mbExistingAccount <- lift $ join <$> for mbInv (\(Id uuid) -> Data.lookupAccount (Id uuid))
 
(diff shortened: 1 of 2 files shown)


services/brig/src/Brig/API/User.hs
```diff
@@ -183,7 +183,16 @@ createUser new = do
   (email, phone) <- validateEmailAndPhone new
 
   -- get invitation and existing account
-  (newTeam, teamInvitation, tid) <- handleTeam (newUserTeam new) (userEmailKey <$> email)
+  (newTeam, teamInvitation, tid) <-
+    case (newUserTeam new, userEmailKey <$> email) of
```
fisx: dito: why inline some, and abtract others?

services/brig/src/Brig/API/User.hs
```diff
@@ -185,15 +185,20 @@ createUser new = do
...
-      Just (NewTeamMember i) ->
-        findTeamInvitation (userEmailKey <$> email) i
-          <&> ( \case
-                  Just (inv, info, tid) -> (Nothing, Just (inv, info), Just tid)
-                  Nothing -> (Nothing, Nothing, Nothing)
-              )
-      Just (NewTeamCreator t) -> (Just t,Nothing,) <$> (Just . Id <$> liftIO nextRandom)
-      Just (NewTeamMemberSSO tid) -> pure (Nothing, Nothing, Just tid)
-      Nothing -> return (Nothing, Nothing, Nothing)
+      Just (NewTeamMember i) -> do
```
fisx, line 188: I don't like this change.  I find it more verbose and less readable.

services/brig/src/Brig/API/User.hs
```diff
@@ -239,23 +239,10 @@ createUser new = do
...
-          Log.info $
-            field "user" (toByteString uid)
-              . field "activation.key" (toByteString $ activationKey edata)
-              . msg (val "Created email activation key/code pair")
-          return $ Just edata
-        Just c -> do
-          ak <- liftIO $ Data.mkActivationKey ek
-          void $ activateWithCurrency (ActivateKey ak) c (Just uid) (join (bnuCurrency <$> newTeam)) !>> EmailActivationError
-          return Nothing
+      then handleEmailActivation email uid newTeam
```
fisx: This seems like a behavior-changing change.  I've pushed 87e64c349665176931f9b9e3dc927e390150aab5, let's see what that does.  (This would also mean we have some test coverage issues.  Oh well, another day...)

fisx (approved):
I've pushed a fix that you should doube-check to make sure I'm not seeing things.  Also, a few minor remarks.  In general, I feel this is a clear improvement.

services/brig/src/Brig/API/User.hs
```diff
@@ -292,14 +297,6 @@ createUser new = do
 
       pure (email, phone)
 
-    createTeam :: (UserId -> Bool -> Team.BindingNewTeam -> TeamId -> AppT IO (Maybe CreateUserTeam))
```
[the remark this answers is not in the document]
smatting: To me this function is interesting business logic that I'd rather see in the body. All other functions in the where block are not that interesting business logic-wise and are okay to abstract behind a name.

services/brig/src/Brig/API/User.hs
smatting: To me this function is interesting business logic that I'd rather see in the body. All other functions in the where block are not that interesting business logic-wise and are okay to abstract behind a name.

services/brig/src/Brig/API/User.hs
smatting: It was more for consisentency's sake: In the refactored version every branch ends up in a `pure (x, x, x,)` expression.
Also tuple sections + fmap is harder for me to read than a bind + pure, but I don't feel strongly about.

services/brig/src/Brig/API/User.hs
smatting: No I don't think so. Please see the https://github.com/wireapp/wire-server/pull/1670/commits/cab00b1640faf32b56f1d4cb343e008f78d295cd that introduced the refactoring.

smatting: ... Oh, yes, you were right. I made a mistake in the commit after that: https://github.com/wireapp/wire-server/pull/1670/commits/ed2accb4cdf4f400abd4599160b6beb8c4ff1ed4

services/brig/src/Brig/API/User.hs
smatting: Reverted to previous version in https://github.com/wireapp/wire-server/pull/1670/commits/3a70d46ef17d9bb09d091af17e515950bf1d4e7e

PR #1674, opened 2021-07-19
Document federation errors

pcapriotti:
This adds documentation of federation errors to Swagger. These are not response entries by endpoints, but simply a list of errors (with some brief explanations) divided into categories. They apply more or less to all endpoints that have anything to do with federation.

I also pasted it here so it is possible to see it rendered without building anything: https://gist.github.com/pcapriotti/e50f61c3a8f6abab900e25c9ff3457aa.

## TODO

 - [x] Remove unused errors
 - [x] Move all error label assignments to a single module
 - [x] ~~Rethink some of the error codes~~
        After some discussion, we decided the error codes are mostly fine
 - [x] Add suggested client behaviour for each error

## Checklist

### DIFF
--- a/services/brig/src/Brig/API/Public.hs
+++ b/services/brig/src/Brig/API/Public.hs
@@ -188,11 +188,10 @@ An error in this category indicates an issue with configuration of federation on
 
 Errors in this category are returned in case of communication issues between the local backend and a remote one, or if the remote side encountered an error while processing an RPC.
 
- - **RPC error with code** (status: 533, label: `grpc-error-code`): An RPC to a remote federator resulted in an error. The RPC error code can be found in the error message.
- - **RPC error with string** (status: 533, label: `grpc-error-string`): An RPC to a remote federator resulted in an error. A description of the error can be found in the error message.
- - **RPC error** (status: 500, label: `federation-rpc-error`): There was a non-specified error in the RPC from this backend to another backend. Check the error message for more details.
- - **Connection refused** (status: 521, label: `cannot-connect-to-remote-federator`): The local federator could not connect to a remote one.
  - **Too much concurrency** (status: 533, label: `too-much-concurrency`): Too many concurrent requests on a remote backend.
+ - **gRPC error** (status: 533, label: `grpc-error`): The current federator encountered an error when making an RPC to a remote one. Check the error message for more details.
+ - **Client RPC error** (status: 500, label: `client-rpc-error`): There was a non-specified error when making a request to another backend. Check the error message for more details.
+ - **Connection refused** (status: 521, label: `cannot-connect-to-remote-federator`): The local federator could not connect to a remote one.
  - **Unknown remote error** (status: 500, label: `unknown-federation-error`): An RPC failed but no specific error was returned by the remote side. Check the error message for more details.
 
 ### Backend compatibility errors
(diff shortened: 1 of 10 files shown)


services/brig/src/Brig/API/Public.hs
```diff
@@ -162,6 +162,60 @@ Okta will ask you to provide two URLs when you set it up for talking to wireapp:
...
+### Local federation errors
+
+An error in this category indicates an issue with configuration of federation on the local backend or with the federation-related content of the current client request.
+
+ - **Federation not enabled** (status: 400, label: `federation-not-enabled`): Federation has not been configured for this backend.
+ - **Federation unavailable** (status: 500, label: `federation-not-available`): Federation is configured for this backend, but the local federator cannot be reached.
+ - **Federation not implemented** (status: 403, label: `federation-not-implemented`): Federated behaviour for a certain endpoint is not yet implemented.
+ - **Remote backend not found** (status: 422, label: `srv-record-not-found`): This backend attempted to contact a backend which does not exist or is not properly configured.
+ - **Federation denied locally** (status: 400, label: `federation-not-allowed`): This backend attempted an RPC to a non-whitelisted backend.
+ - **Federator discovery failed** (status: 500, label: `srv-lookup-dns-error`): A DNS occurred during discovery of a remote backend.
```
mdimjasevic, line 184:
```suggestion
 - **Federator discovery failed** (status: 500, label: `srv-lookup-dns-error`): A DNS error occurred during discovery of a remote backend.
```

services/brig/src/Brig/API/Public.hs
```diff
@@ -162,6 +162,60 @@ Okta will ask you to provide two URLs when you set it up for talking to wireapp:
...
+
+### Remote federation errors
+
+Errors in this category are returned in case of communication issues between the local backend and a remote one, or if the remote side encountered an error while processing an RPC.
+
+ - **RPC error with code** (status: 533, label: `grpc-error-code`): An RPC to a remote federator resulted in an error. The RPC error code can be found in the error message.
+ - **RPC error with code** (status: 533, label: `grpc-error-string`): An RPC to a remote federator resulted in an error. A description of the error can be found in the error message.
```
mdimjasevic, line 191:
```suggestion
 - **RPC error with string** (status: 533, label: `grpc-error-string`): An RPC to a remote federator resulted in an error. A description of the error can be found in the error message.
```

mdimjasevic (approved):
A couple of minor suggestions are inlined.

CHANGELOG.md
```diff
@@ -61,6 +61,7 @@ Upgrade nginz (#1658)
 
 ## Federation changes (alpha feature, do not use yet)
 
+* Added documentation of federation errors (#1674).
```
jschaul, line 64: This is a federation-api-backwards-incompatible change (since router.proto changes), which is fine right now, but we'd need to be more careful about this in the future. Could be added to the changelog like here: https://github.com/wireapp/wire-server/blob/develop/CHANGELOG.md#internal-federation-api-changes

jschaul (approved):

CHANGELOG.md
pcapriotti: If we had kept the existing field indices, then I think this sort of change should be backwards compatible. Is this right?

jschaul: Yes, if the meaning and the numbers stay, but only things are added, it's backwards compatible. If indices change or are removed it's not.
So far so good, no need for backwards compatibility concerns just yet.

pcapriotti:
Thanks for approving @jschaul. I marked this PR as a draft, because I wanted to use it to work on reorganising the error codes and clean up error handling. Should I instead merge it as is, and do that in a separate PR?

jschaul:
> Thanks for approving @jschaul. I marked this PR as a draft, because I wanted to use it to work on reorganising the error codes and clean up error handling. Should I instead merge it as is, and do that in a separate PR?

As you wish, I think this is not urgent to get merged, so feel free to keep it around and organize errors a but differently first. Whatever is more convenient.

Another thought crossed my mind: I noticed when doing a manual test that currently the behaviour of the webapp on a 400  (federation not allowed) is, they stop. And on the 525 errors, the webapp retried the request about 10 times. So I wonder if there should be some guidance we can give on these different errors about what e.g. the client could do (retry later, or not), or what the user could do. Some of this discussion could happen with involvement of more parties, but some things could also be added to the description perhaps.

pcapriotti:
> Another thought crossed my mind: I noticed when doing a manual test that currently the behaviour of the webapp on a 400 (federation not allowed) is, they stop. And on the 525 errors, the webapp retried the request about 10 times. So I wonder if there should be some guidance we can give on these different errors about what e.g. the client could do (retry later, or not), or what the user could do. Some of this discussion could happen with involvement of more parties, but some things could also be added to the description perhaps.

Good point, I added it as a TODO item.

PR #1679, opened 2021-07-23
Remove unused HEAD endpoints

pcapriotti:
Servant implements HEAD endpoints automatically in terms of GET endpoints. As it turned out, the explicit implementations we had for `HEAD /users/:uid` (and its qualified counterpart) ended up not being invoked by Servant's routing mechanism.

This simply removes the endpoints, and adds tests showing that the functionality is still present and works as expected.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -54,14 +54,15 @@ Upgrade nginz (#1658)
 
 ## Documentation
 
-* Improved Swagger documentation for endpoints with multiple responses (#1649).
+* Improved Swagger documentation for endpoints with multiple responses (#1649)
 
 ## Internal changes
 
 * Improvements to local integration test setup when using buildah and kind (#1667)
-* The servant-swagger dependency now points to the current upstream master (#1656).
-* Improved error handling middleware (#1671).
+* The servant-swagger dependency now points to the current upstream master (#1656)
+* Improved error handling middleware (#1671)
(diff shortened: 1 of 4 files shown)


jschaul (approved):
Question: I presume removing this endpoint will also remove it from swagger, right? That might not be ideal, as nothing then indicates to client developers that this HEAD endpoint can still be called. Is there a way to sneak this back into swagger?

Thanks for adding tests, strange that those didn't exist.

pcapriotti:
> Question: I presume removing this endpoint will also remove it from swagger, right? That might not be ideal, as nothing then indicates to client developers that this HEAD endpoint can still be called. Is there a way to sneak this back into swagger?

That's right. We could try to somehow generate swagger for this HEAD endpoint, but then why only this one and not all? Would you want to see such an endpoint in swagger for each of our GET endpoints?

jschaul:
> > Question: I presume removing this endpoint will also remove it from swagger, right? That might not be ideal, as nothing then indicates to client developers that this HEAD endpoint can still be called. Is there a way to sneak this back into swagger?
> 
> That's right. We could try to somehow generate swagger for this HEAD endpoint, but then why only this one and not all? Would you want to see such an endpoint in swagger for each of our GET endpoints?

Fine to merge this PR as-is. Generally I think it would introduce noise to document lots of possible HEAD requests that are not used.

PR #1682, opened 2021-07-26
Support client certificates in federator

pcapriotti:
This adds support for client authentication using TLS certificates when making requests to a remote backend.

## Summary of changes

 - Added new configuration options for client certificate and private key
 - Configured demo nginz to require client certificates
 - Read client certificate on federator startup
 - Use client certificate when creating a GRPC client
 - Upgrade haskell packages `tls` and `cryptonite`

## Checklist

 - [x] New configuration options have been documented


charts/federator/templates/configmap.yaml
```diff
@@ -43,7 +43,12 @@ data:
       # Filepath to one or more PEM-encoded server certificates to use as a trust
       # store when making grpc requests to remote backends
       {{- if $.Values.remoteCAContents }}
-      remoteCAStore: "/etc/wire/federator/ca/remote-ca.pem"
+      remoteCAStore: "/etc/wire/federator/conf/remote-ca.pem"
+      {{- end }}
+      useSystemCAStore: false
```
jschaul, line 48:
```suggestion
```
That line exists (in the correct templated way) in line 53. hardcoding this to false doesn't seem like a good idea.

charts/federator/templates/configmap.yaml
```diff
@@ -61,3 +66,11 @@ data:
         {{- end}}
     {{- end }}
     {{- end }}
+
+  # TODO: add validation and fail early during templating: either contents should be provided; or explicitly system trust store enabled
+  {{- if .Values.remoteCAContents }}
+  remote-ca.pem: {{ .Values.remoteCAContents | quote }}
```
jschaul, line 72: I'm not sure this will work: This produces a quoted string containing `\n` characters. Not sure this will lead to actual new lines on disk. So parsing of these certificates might be broken. 

You can test the yaml output using 

```
FEDERATION_DOMAIN_BASE=example.com ./hack/bin/selfsigned-kubernetes.sh
make helm-template-wire-server
```

docs/reference/config-options.md
```diff
@@ -197,14 +197,49 @@ federator:
...
+
+```yaml
+federator:
+  optSettings:
+    federationStrategy:
+      allowedDomains:
+        - server2.example.com
+    useSystemCAStore: false
+    clientCertificate: client.pem
+    clientPrivateKey: client-key.pem
```
jschaul, line 244:
```suggestion
    clientPrivateKey: | 
       -----BEGIN RSA PRIVATE KEY-----
       ...
```
?

services/federator/src/Federator/Options.hs
```diff
@@ -62,9 +62,21 @@ data RunSettings = RunSettings
...
+    clientCertificate :: Maybe FilePath,
+    clientPrivateKey :: Maybe FilePath
   }
-  deriving (Show, Generic)
+  deriving (Eq, Show, Generic)
+
+defRunSettings :: RunSettings
+defRunSettings =
+  RunSettings
+    { federationStrategy = AllowAll,
```
jschaul, line 74: I think the default should be no federation (allow list with empty list) at least at this stage, no?

Tests is another matter, there could be a default test run settings object under ./test/ somewhere.

services/federator/test/unit/Test/Federator/Options.hs
```diff
@@ -63,6 +68,166 @@ parseFederationStrategy =
...
+testSettings :: TestTree
+testSettings =
+  testGroup
+    "settings"
+    [ testCase "parse configuration example (open federation)" $ do
+        assertParsesAs
+          defRunSettings
+          "federationStrategy:\n\
+          \  allowAll:\n\
+          \useSystemCAStore: true",
```
jschaul, line 80: Thanks for the tests! Do you think we could use a nicer multi-line string approach here (with that quasi-quoter) so this becomes easier to read?

jschaul (commented):
I know this is still marked as draft, but I had a look already anyway and made a few suggestions.

charts/federator/templates/configmap.yaml
pcapriotti: The yaml of the expanded configmap contains a quoted string with escaped newlines, as you say, but on the volume using this configmap this translates into a file where those escaped newlines have turned into actual newlines. I've checked it by attaching to a running federator pod.

docs/reference/config-options.md
pcapriotti: No, `clientPrivateKey` should contain the path to the private key.

services/federator/src/Federator/Options.hs
pcapriotti: Yes, this should only be used for tests. I moved it to a test module (8878a177c94020ce1270ddeb4e2995c1bfbdb253).

services/federator/test/unit/Test/Federator/Options.hs
pcapriotti: Yep, I switched those multi-line strings to the interpolate quasiquoter.

charts/nginx-ingress-services/templates/certificate_federator.yaml
```diff
@@ -0,0 +1,30 @@
+{{- if and .Values.tls.enabled .Values.tls.useCertManager .Values.federator.enabled }}
```
akshaymankar, line 1: I think should add a check which disallows disabling TLS when federator is enabled.

charts/nginx-ingress-services/templates/certificate_federator.yaml
```diff
@@ -0,0 +1,30 @@
...
+metadata:
+  name: "federator-{{ include "nginx-ingress-services.zone" . | replace "." "-" }}-csr"
+  namespace: {{ .Release.Namespace }}
+  labels:
+    chart: "{{ .Chart.Name }}-{{ .Chart.Version }}"
+    release: "{{ .Release.Name }}"
+    heritage: "{{ .Release.Service }}"
+spec:
+  issuerRef:
+    name: letsencrypt-staging
```
akshaymankar, line 13:
```suggestion
    name: letsencrypt-http01
```

charts/nginx-ingress-services/templates/staging_issuer.yaml
```diff
@@ -0,0 +1,22 @@
...
+    server: "https://acme-staging-v02.api.letsencrypt.org/directory"
+    email: {{ required "Missing value: certmasterEmail" .Values.certManager.certmasterEmail | quote }}
+    # NOTE: this secret doesnt need to be created, it only gets a name with this
+    privateKeySecretRef:
+      name: letsencrypt-staging-account-key
+    solvers:
+      - http01:
+          ingress:
+            class: nginx
+{{- end -}}
```
akshaymankar, line 22: Why is this needed?

services/federator/src/Federator/Run.hs
```diff
@@ -113,17 +118,48 @@ mkCAStore settings = do
...
+  where
+    noCreds1 :: Either String ()
+    noCreds1
+      | isNothing (clientPrivateKey settings) = pure ()
+      | otherwise = Left "invalid client credentials: no certificate"
+
+    getCreds1 :: Either String FilePath
+    getCreds1 =
+      maybe (Left "invalid client credentials: no private key") pure $
+        clientPrivateKey settings
```
akshaymankar, line 134: Why are these called `noCreds1` and `getCreds1`?

akshaymankar (commented):
Can you please explain whey we have/need let's encrypt staging? Otherwise, the PR looks good.

pcapriotti:
> Can you please explain whey we have/need let's encrypt staging? Otherwise, the PR looks good.

I thought we could merge it with staging first to see whether things have been configured correctly. Or we could deploy it (and cert-manager) on the CI environment and check if things work there. If you think it's ok to use production immediately, I can of course change it.

services/federator/src/Federator/Run.hs
pcapriotti: It's some arbitrary convention of prefixing 1 because they return (or check) credentials without the first field. Somewhat in the same spirit as `foldr1`. In this case it's just the private key, but if you pretend we had more credential fields, then it should make more sense. Hopefully :smile:. If not, I can call them something else.

akshaymankar:
> > Can you please explain whey we have/need let's encrypt staging? Otherwise, the PR looks good.
> 
> I thought we could merge it with staging first to see whether things have been configured correctly. Or we could deploy it (and cert-manager) on the CI environment and check if things work there. If you think it's ok to use production immediately, I can of course change it.

I think it makes sense to try all of this out with staging cert manager first, but to do that we already have `certificate-manager.apiServerURL` config. https://github.com/wireapp/wire-server/blob/31479d9579d4cfc271914574b9ef7d0121417f16/charts/nginx-ingress-services/templates/issuer.yaml#L13

pcapriotti:
> I think it makes sense to try all of this out with staging cert manager first, but to do that we already have `certificate-manager.apiServerURL` config.

But we would also need to set the CA to use, since that's different between staging and production.

akshaymankar:
> > I think it makes sense to try all of this out with staging cert manager first, but to do that we already have `certificate-manager.apiServerURL` config.
> 
> But we would also need to set the CA to use, since that's different between staging and production.

Yes, but we shouldn't need to hardcode the CA into our code. We can make the testing person set that in the `values.yaml` when they are deploying the charts.

charts/nginx-ingress-services/templates/secret_federator.yaml
```diff
@@ -0,0 +1,26 @@
...
+kind: Secret
+metadata:
+  name: federator-certificate-secret
+  labels:
+    chart: "{{ .Chart.Name }}-{{ .Chart.Version }}"
+    release: "{{ .Release.Name }}"
+    heritage: "{{ .Release.Service }}"
+type: kubernetes.io/tls
+data:
+  ca.crt: {{ .Values.secrets.tlsClientCA | b64enc | quote }}
```
akshaymankar, line 12: This should be optional right? Otherwise how can we tell nginx to use system CA?

pcapriotti: I don't know if the client CA can be optional. I didn't find a way to make nginx use the system defaults to verify client certificates. That's why I had Let's Encrypt intermediate CA as the default before.

services/federator/src/Federator/Remote.hs
```diff
@@ -84,11 +91,10 @@ callInward client request =
 -- FUTUREWORK(federation): Consider using HsOpenSSL instead of tls for better
 -- security and to avoid having to depend on cryptonite and override validation
 -- hooks. This might involve forking http2-client: https://github.com/lucasdicioccio/http2-client/issues/76
--- FUTUREWORK(federation): Allow a configurable trust store to be used in TLS certificate validation
 --   See also https://github.com/lucasdicioccio/http2-client/issues/76
```
akshaymankar, line 94:
```suggestion
-- FUTUREWORK(federation): Use openssl
--   See also https://github.com/lucasdicioccio/http2-client/issues/76
```

akshaymankar (commented):
Overall looks good, I am not clear on why the client cert is optional.

charts/federator/values.yaml
```diff
@@ -8,6 +8,11 @@ service:
   internalFederatorPort: 8080
   externalFederatorPort: 8081
 
+tls:
+  # if enabled, federator will get its client certificate and private key from
+  # the secret used by the federator ingress
+  shareFederatorSecret: false
```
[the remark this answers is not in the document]
pcapriotti: c7d717a817fc33016c5ad952a2f79ecc5a89f164

charts/nginx-ingress-services/templates/certificate_federator.yaml
pcapriotti: What's the proper way to do this?

akshaymankar: We can use the `fail` keyword to make a failure in rendering the template. So something like:

```
{{- if and .Values.federator.enabled (not .Values.tls.enabled) }}
{{- fail "When federator is enabled, TLS must also be enabled" }}
{{- end }}
```

akshaymankar (approved):

jschaul:
Post-merge review: Looks good overall. Few minor questions came to mind:

1. What was your usecase for the serve-helm-charts-locally script? Which part of the development flow was this useful for?
2. I see client certificates seem to be optional in the federator configuration, as well as in the config options readme to an operator. Wouldn't the remote backend just reject all requests if there's no valid client certificate passed though? The configuration in the nginx-ingress seems to always turn client certificate validation "on" if I read this correctly (whatever that means; shouldn't this reject the http request if it doesn't contain a client cert?). Perhaps the config-options readme could get an extra line informing a non-backend-engineer server admin on when to provide a client certificate and when this isn't necessary, and what the effects are of not specifying this.

jschaul:
One more thought: the `useSharedFederatorSecret` setting in the federator (or wire-server) helm chart ties the federator/wire-server helm chart to the nginx-ingress-services helm chart. Yet currently those are installed separately. Is there an ordering concern here? Can I install wire-server with this setting to true before nginx-ingress-controller is installed? Or would I need to first install nginx-ingress-services before I can install wire-server helm chart?
In case there is an ugly ordering concern, can we resolve it by including the nginx-ingress-controller chart as part of wire-server's requirements (see https://github.com/wireapp/wire-server/blob/develop/charts/wire-server/requirements.yaml) ?

akshaymankar:
> What was your usecase for the serve-helm-charts-locally script? Which part of the development flow was this useful for?

We wanted to deploy the code in this branch to an environment using `wire-server-deploy`, but we couldn't without either pushing the charts to some repository or serving them locally. The newly created [adhoc-testing environment](https://github.com/zinfra/cailleach/pull/578) can use this by adding `http://localhost:4001` to its repositories list.

> I see client certificates seem to be optional in the federator configuration, as well as in the config options readme to an operator. Wouldn't the remote backend just reject all requests if there's no valid client certificate passed though?

This is a mistake, they were originally optional, but then we realized they shouldn't be. There is no way the federator could work without them. So, the docs in config-options needed update and we forgot.

> The configuration in the nginx-ingress seems to always turn client certificate validation "on" if I read this correctly (whatever that means; shouldn't this reject the http request if it doesn't contain a client cert?).

Yes, setting it to always turn "on" will make any requests without a client cert to be rejected. 

> Perhaps the config-options readme could get an extra line informing a non-backend-engineer server admin on when to provide a client certificate and when this isn't necessary, and what the effects are of not specifying this.

As said earlier, this is a mistake, I will rectify it by marking this as not optional.

> One more thought: the `useSharedFederatorSecret` setting in the federator (or wire-server) helm chart ties the federator/wire-server helm chart to the nginx-ingress-services helm chart. Yet currently those are installed separately. Is there an ordering concern here? Can I install wire-server with this setting to true before nginx-ingress-controller is installed? Or would I need to first install nginx-ingress-services before I can install wire-server helm chart?

If someone turns `useSharedFederatorSecret` to true, it means that secret must exist before wire-server can be deployed. The secret doesn't have to come from the nginx-ingress-services, it can come from anywhere, for instance someone having their own TLS acquiring mechanism. But I guess in most cases, the operators would want to deploy nginx-ingress-services before wire-server.

>  In case there is an ugly ordering concern, can we resolve it by including the nginx-ingress-controller chart as part of wire-server's requirements (see https://github.com/wireapp/wire-server/blob/develop/charts/wire-server/requirements.yaml) ?

This is not good, if we want the ingress-controller business to be optional. This is similar to how deploying wire-server requires cassandra, elasticsearch, etc to be there. But we don't want to mandate that people use our given cassandra and elasticsearch. So, we already have this ordering problem and I guess we solve it by making sure all our docs tell users to install those things before wire-server. Currently, it is not the case with ingress-controller and services, so I will make that change to our docs.

PR #1687, opened 2021-07-28
Server-to-server authentication

pcapriotti:
This is the final bit of logic implementing server-to-server authentication, namely validation of the domain name provided as part of the request against the certificate checked at the point of SSL termination.

## Checklist


services/federator/src/Federator/ExternalServer.hs
```diff
@@ -66,13 +84,27 @@ callLocal = runInwardError . callLocal'
...
+      r
+  ) =>
+  Maybe ByteString ->
+  Request ->
+  Sem r ByteString
+callLocal' mcert req@Request {..} = do
   Log.debug $
     Log.msg ("Inward Request" :: ByteString)
       . Log.field "request" (show req)
+      . maybe id (\cert -> Log.field "certificate" (Text.decodeUtf8 cert)) mcert
```
mdimjasevic, line 105: I forgot, but should this be logging only for now, or should there be some other actions to run? The PR is still in the draft state so perhaps that explains it.

services/federator/src/Federator/ExternalServer.hs
```diff
@@ -85,21 +117,43 @@ callLocal' req@Request {..} = do
...
     code -> do
       let description = "Invalid HTTP status from component: " <> Text.pack (show code) <> " " <> Text.decodeUtf8 (HTTP.statusMessage resStatus) <> " Response body: " <> maybe mempty cs resBody
       throwInward IOther description
 
-routeToInternal :: (Members '[Service, Embed IO, Polysemy.Error ServerError, TinyLog, Polysemy.Reader RunSettings] r) => SingleServerT info Inward (Sem r) _
-routeToInternal = singleService (Mu.method @"call" callLocal)
+routeToInternal ::
+  (Members '[Service, Embed IO, Polysemy.Error ServerError, TinyLog, DiscoverFederator, Polysemy.Reader RunSettings] r) =>
+  Maybe ByteString ->
+  SingleServerT info Inward (Sem r) _
```
mdimjasevic, line 130: At one point you had the return type spelled out, but then put back `_`. Are you planning to put back the type?

services/federator/src/Federator/Validation.hs
```diff
@@ -43,12 +52,62 @@ federateWith targetDomain = do
     AllowAll -> True
     AllowList (AllowedDomains domains) -> targetDomain `elem` domains
 
--- | Validates an unknown domain string against the allowList using the federator startup configuration
-validateDomain :: Members '[Polysemy.Reader RunSettings, Polysemy.Error InwardError] r => Text -> Sem r Domain
-validateDomain unparsedDomain = do
+decodeCertificate ::
+  Members '[Polysemy.Error InwardError] r =>
```
mdimjasevic, line 56: Just nitpicking: given that there is only one effect in the list, you can use `Member` instead of `Members`.

services/federator/test/integration/Test/Federator/InwardSpec.hs
```diff
@@ -109,14 +111,19 @@ inwardBrigCall requestPath payload = do
...
   liftIO $ gRpcCall @'MsgProtoBuf @Inward @"Inward" @"call" c brigCall
 
 viewFederatorExternalClient :: (MonadIO m, MonadHttp m, MonadReader TestEnv m, HasCallStack) => m GrpcClient
 viewFederatorExternalClient = do
   Endpoint fedHost fedPort <- cfgFederatorExternal . view teTstOpts <$> ask
-  client <- createGrpcClient (grpcClientConfigSimple (Text.unpack fedHost) (fromIntegral fedPort) False)
+  exampleCert <- liftIO $ BS.readFile "test/resources/unit/localhost.pem"
+  let cfg =
+        (grpcClientConfigSimple (Text.unpack fedHost) (fromIntegral fedPort) False)
+          { _grpcClientConfigHeaders = [("X-SSL-Certificate", exampleCert)]
```
mdimjasevic, line 124: It crossed my mind, but I am not completely sure about this: should the headers list be expanded here rather than replaced?

services/federator/test/unit/Test/Federator/Validation.hs
```diff
@@ -36,6 +38,11 @@ import Test.Polysemy.Mock (evalMock)
 import Test.Tasty
 import Test.Tasty.HUnit
 import Wire.API.Federation.GRPC.Types
+import Wire.Network.DNS.SRV (SrvTarget (..))
+
+mockDiscoveryTrivial :: Sem (DiscoverFederator ': r) x -> Sem r x
+mockDiscoveryTrivial = Polysemy.interpret $ \(DiscoverFederator dom) ->
+  pure . Right $ SrvTarget (Text.encodeUtf8 (domainText dom)) 443
```
mdimjasevic, line 45: Given that this function is copy-pasted in `ExternalServer.hs`, how about moving it to a `Util` or `Interpret` module?

mdimjasevic (commented):
So far it looks quite good.

services/federator/src/Federator/ExternalServer.hs
pcapriotti: We removed the logging statement now.

services/federator/src/Federator/ExternalServer.hs
pcapriotti: I think `_` is fine, that's what the library docs suggest you do.

services/federator/src/Federator/Validation.hs
pcapriotti: Fixed in acec8903fea4b886e98d415a6418a9e908ef4f55.

services/federator/test/integration/Test/Federator/InwardSpec.hs
pcapriotti: The default is empty, so it makes no difference here.

services/federator/test/unit/Test/Federator/Validation.hs
pcapriotti: I've now imported it from the other module directly.

services/federator/src/Federator/ExternalServer.hs
```diff
@@ -56,8 +62,20 @@ import Wire.API.Federation.GRPC.Types
 --
 -- FUTUREWORK(federation): implement server2server authentication!
 -- (current validation only checks parsing and compares to allowList)
```
akshaymankar, line 64:
```suggestion
```

services/federator/src/Federator/Validation.hs
```diff
@@ -107,3 +162,15 @@ sanitizePath originalPath = do
     throwInward IForbiddenEndpoint ("disallowed path: " <> cs originalPath)
 
   pure normalized
+
+-- Match a hostname against the domain names of a certificate.
```
akshaymankar, line 166:
```suggestion
-- | Match a hostname against the domain names of a certificate.
```

services/federator/test/unit/Test/Federator/Validation.hs
```diff
@@ -77,26 +101,125 @@ validateDomainAllowListFailSemantic =
...
+  testCase "should fail if no client certificate is provided" $
+    runM . evalMock @Remote @IO $ do
+      res :: Either InwardError Domain <-
+        Polysemy.runError
+          . mockDiscoveryTrivial
+          . Polysemy.runReader noClientCertSettings
+          $ validateDomain Nothing "foo.example.com"
+      case res of
+        Left (InwardError IInvalidDomain _) -> pure ()
+        x -> embed $ assertFailure $ "expected IInvalidDomain error, got " <> show x
```
akshaymankar, line 155: Seems weird that everything is `IInvalidDomain`, I think a better name would be `IAuthenticationFailed`, I am going to try and rename this.

services/federator/test/unit/Test/Federator/Validation.hs
```diff
@@ -57,46 +78,135 @@ tests =
...
+          run . Polysemy.runError
+            . mockDiscoveryTrivial
+            . Polysemy.runReader noClientCertSettings
+            $ validateDomain (Just exampleCert) (domainText domain)
+    res @?= Right domain
+
+validateDomainDiscoveryFailed :: TestTree
+validateDomainDiscoveryFailed =
+  testCase "should fail if discovery fails" $ do
+    exampleCert <- BS.readFile "test/resources/unit/localhost.example.com.pem"
```
jschaul, line 189:
```suggestion
    exampleCert <- BS.readFile "test/resources/unit/example.com.pem"
```

services/federator/test/unit/Test/Federator/Validation.hs
```diff
@@ -57,46 +78,135 @@ tests =
...
+  testCase "should fail if the client certificate has a wrong domain" $ do
+    exampleCert <- BS.readFile "test/resources/unit/localhost.example.com.pem"
+    let res :: Either InwardError Domain =
+          run . Polysemy.runError
+            . mockDiscoveryTrivial
+            . Polysemy.runReader noClientCertSettings
+            $ validateDomain (Just exampleCert) "foo.example.com"
+    case res of
+      Left (InwardError IAuthenticationFailed _) -> pure ()
+      x -> assertFailure $ "expected IAuthenticationFailed error, got " <> show x
```
jschaul, line 172: maybe create a little helper function similar to `expectErr` in this file to not duplicate these three lines of case statements for expected errors?

jschaul (approved):
This PR seems to make use of a new coreDNS pod, at least there's new config for that used in Run.hs; that's cool to be able to integration test DNS issues. 

How do these integration tests work on kubernetes though? Also, I see lots of unit tests but no new integration tests. So where is the coreDNS pod actually used? It would be useful to add some comments in the integration tests that make use of DNS which provide some explanation how DNS is used during discovery for A) local integration tests and B) kubernetes integration tests (I assume some implict DNS names are created by some k8s service that match the name in some config file somewhere in the tests?); as issues arising here at some point in the future may not be easy to figure out.

As for documentation: some guidance how to configure federator and nginx-ingress helm charts for operators so that client certificates are made use of (do we allow server2server authentication to actually be turned off? I suppose we should not?) would be nice (in this or a follow-up PR), and some guidance on how to include your friendly backend's CA in your trust chain would be nice. 

As for tests, can we test with a few more scenarios of certificates? E.g. expired, certificate chains of more than one,

services/federator/test/unit/Test/Federator/Validation.hs
akshaymankar: I just replaced this with asserting more precisely on the errors.

akshaymankar:
> This PR seems to make use of a new coreDNS pod, at least there's new config for that used in Run.hs; that's cool to be able to integration test DNS issues.
> 
> How do these integration tests work on kubernetes though? Also, I see lots of unit tests but no new integration tests. So where is the coreDNS pod actually used? It would be useful to add some comments in the integration tests that make use of DNS which provide some explanation how DNS is used during discovery for A) local integration tests and B) kubernetes integration tests (I assume some implict DNS names are created by some k8s service that match the name in some config file somewhere in the tests?); as issues arising here at some point in the future may not be easy to figure out.


With this PR, all the federated requests (even those which bypass nginx, i.e. federator integration tests) need the DNS entry for the origin domain to work. This is not a problem in kubernetes because, as you guessed, there is always an SRV record for the currently deployed federator, so we can just use that to make a request. For local development, however, we don't get this SRV record, so we first considered creating a global record, but decided it was ugly and brittle. So, we chose to just install coredns in our docker-ephmeral setup to facilitate this. The DNS name used by the integration tests is configured in `charts/federator/templates/tests/configmap.yaml` and `services/integration.yaml`.

We didn't need more integration tests because all integration tests started needing to provide this cert and all the failure cases were tested in unit test. But I think it would be nice to add one integration test where federation is denied due to wrong certificate would be good. 


> As for documentation: some guidance how to configure federator and nginx-ingress helm charts for operators so that client certificates are made use of (do we allow server2server authentication to actually be turned off? I suppose we should not?) would be nice (in this or a follow-up PR), and some guidance on how to include your friendly backend's CA in your trust chain would be nice.

Yes, we could certainly do a better job at documenting this. Let's take it up right after we merge this.

> As for tests, can we test with a few more scenarios of certificates? E.g. expired, certificate chains of more than one,

We don't check expiry or even the chains. This is done by nginx. We could add integration tests, but I think it is ok to just trust nginx here.

PR #1688, opened 2021-07-30
Change http response code for `missing-legalhold-consent`.

fisx:
412 is used for "need to add some clients" in message posting; 403 makes client implementations more straight-forward.

It's probably worth mentioning that this only affects users about to get in contact with LH devices.  For those users, we have a new safety check in place that if the user has old clients or is not in a team cleared for LH, she will get a new error response.  403 in that case is better, since it makes it clear to the client that this won't work on retry either.  412 is used for notifying the client that they need to pick up some newly added clients listening in on the conversation, so this will probably break in less nice ways.

tl;dr: only affects users in touch with LH teams that shouldn't, and makes their UX strictly better.  :)

## Checklist

### DIFF
--- a/services/galley/src/Galley/API/Error.hs
+++ b/services/galley/src/Galley/API/Error.hs
@@ -209,6 +209,15 @@ noLegalHoldDeviceAllocated = mkError status404 "legalhold-no-device-allocated" "
 legalHoldCouldNotBlockConnections :: Error
 legalHoldCouldNotBlockConnections = mkError status500 "legalhold-internal" "legal hold service: could not block connections when resolving policy conflicts."
 
+-- | See 'Brig.API.Error.missingLegalholdConsent'.
+--
+-- FUTUREWORK: To avoid this duplication, you could turn this into an `ErrorDescription` and
+-- move it to the `ErrorDescription` module in wire-api, where it can then be used from both
+-- brig and galley. Also, it makes it easier to add it to swagger, if desired, or even later
+-- turn it into a statically checked response with `MultiVerb`.
+--
+-- For examples, see https://github.com/wireapp/wire-server/pull/1657 or
+-- https://github.com/wireapp/wire-server/pull/1657/commits/8e73769d7d4a657fef5fbe9210f885f9ccbcaab6.
 missingLegalholdConsent :: Error
 missingLegalholdConsent = mkError status403 "missing-legalhold-consent" "Failed to connect to a user or to invite a user to a group because somebody is under legalhold and somebody else has not granted consent."
 
--- a/services/galley/src/Galley/API/Error.hs
+++ b/services/galley/src/Galley/API/Error.hs
@@ -210,7 +210,7 @@ legalHoldCouldNotBlockConnections :: Error
 legalHoldCouldNotBlockConnections = mkError status500 "legalhold-internal" "legal hold service: could not block connections when resolving policy conflicts."
 
 missingLegalholdConsent :: Error
-missingLegalholdConsent = mkError status412 "missing-legalhold-consent" "Failed to connect to a user or to invite a user to a group because somebody is under legalhold and somebody else has not granted consent."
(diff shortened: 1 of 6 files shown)


services/galley/src/Galley/API/Error.hs
```diff
@@ -210,7 +210,7 @@ legalHoldCouldNotBlockConnections :: Error
 legalHoldCouldNotBlockConnections = mkError status500 "legalhold-internal" "legal hold service: could not block connections when resolving policy conflicts."
 
 missingLegalholdConsent :: Error
-missingLegalholdConsent = mkError status412 "missing-legalhold-consent" "Failed to connect to a user or to invite a user to a group because somebody is under legalhold and somebody else has not granted consent."
+missingLegalholdConsent = mkError status403 "missing-legalhold-consent" "Failed to connect to a user or to invite a user to a group because somebody is under legalhold and somebody else has not granted consent."
```
pcapriotti, line 213: To avoid this duplication, you could turn this into an `ErrorDescription` and move it to the `ErrorDescription` module in wire-api, where it can then be used from both brig and galley. Also, it makes it easier to add it to swagger, if desired, or even later turn it into a statically checked response with `MultiVerb`.

I've done this for a bunch of errors as part of https://github.com/wireapp/wire-server/pull/1657. This is a typical example: https://github.com/wireapp/wire-server/pull/1657/commits/8e73769d7d4a657fef5fbe9210f885f9ccbcaab6. It can be a bit of work though, depending on how often the error is thrown, so feel free to ignore this.

pcapriotti (approved):
Looks good, and it does make more sense to return 403 instead of 412. Minor suggestion below.

services/galley/src/Galley/API/Error.hs
fisx: thanks, i'll look into it.

fisx: (why is nothing ever easy?  :-))

fisx:
```
            User cannot fetch prekeys of LH users if consent is missing:                                                                      FAIL
              Exception: Assertions failed:
               1: 403 =/= 500
               2: Right (TestErrorLabel {fromTestErrorLabel = "internal-server-error"}) =/= Right (TestErrorLabel {fromTestErrorLabel = "missing-legalhold-consent"})
              
              Response was:
              
              Response {responseStatus = Status {statusCode = 500, statusMessage = "Internal Server Error"}, responseVersion = HTTP/1.1, responseHeaders = [("Transfer-Encoding","chunked"),("Date","Fri, 30 Jul 2021 15:15:03 GMT"),("Server","Warp/3.3.13"),("Content-Encoding","gzip"),("Content-Type","application/json")], responseBody = Just "{\"code\":500,\"message\":\"Internal Server Error\",\"label\":\"internal-server-error\"}", responseCookieJar = CJ {expose = []}, responseClose' = ResponseClose}
                error, called at src/Bilge/Assert.hs:89:5 in bilge-0.22.0-KwPXBVYHFlF3OFbmYHfwv3:Bilge.Assert
                <!!, called at src/Bilge/Assert.hs:107:19 in bilge-0.22.0-KwPXBVYHFlF3OFbmYHfwv3:Bilge.Assert
                !!!, called at test/integration/API/Teams/LegalHold.hs:1176:9 in main:API.Teams.LegalHold
```

seems like i need to do some more integration test fixing...

services/galley/src/Galley/API/Error.hs
fisx: I've solved it like this: 83abb9cae85112193dac8b9e976a7bc8a70612ff

I've made you the commit author @pcapriotti, let me know if that's not ok and I'll change it back.

pcapriotti: Sure, that's fine.

pcapriotti: FYI: #1693 contains the change I was suggesting, since it was useful for another refactoring.

PR #1690, opened 2021-08-02
Release 2021-08-02

fisx:
## Checklist


fisx:
NB: I snuck in ccee508, and linked it in the changelog rather than burying the information only in the latter.

CHANGELOG.md
```diff
@@ -45,6 +92,9 @@ This release requires a manual change in your galley configuration: `settings.co
 
 ## API Changes
 
+* A new team feature for classified domains is available (#1626):
+  - a public endpoint is at `GET /teams/:tid/features/classifiedDomains`
+  - an internal endpoint is at `GET /i/teams/:tid/features/classifiedDomains`
```
akshaymankar, line 97: Why did these things get added to previous release notes? Looks like a mistake.

fisx: e7d682d you're right, thanks!  it was like this on `develop`, but i wasn't careful enough to catch it.

akshaymankar (approved):
:rocket:

CHANGELOG.md
```diff
@@ -34,8 +37,57 @@
...
+
+Upgrade nginz (#1658)
+
+## API Changes
+
+## Features
+
+* A new team feature for classified domains is available (#1626):
+  - a public endpoint is at `GET /teams/:tid/features/classifiedDomains`
+  - an internal endpoint is at `GET /i/teams/:tid/features/classifiedDomains`
```
mdimjasevic, line 57: Should this have been mentioned in the API Changes section too?

fisx: you're right, that would have been better.  in my defense, it's an extension, not a change in existing end-points, and the changelog entry isn't too long to read it end-to-end, and it's already release, so i won't fix it this time.

PR #1693, opened 2021-08-03
Finish integrating `MultiVerb`

pcapriotti:
This PR completely removes `UVerb` from the codebase, as well as the `EmptyResult` kludge that we needed to generate empty responses. Both are now replaced by `MultiVerb`.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,6 +40,7 @@
 ## Internal changes
 
 * Replaced uses of `UVerb` and `EmptyResult` with `MultiVerb` (#1693)
+* Added a mechanism to derive `AsUnion` instances automatically (#1693)
 
 # [2021-08-02]
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,10 +35,11 @@
 
 ## Documentation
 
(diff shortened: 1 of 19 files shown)


mdimjasevic (approved):
I am not familiar with qualified message sending, so probably it is no wonder I don't understand why an endpoint for it can return a client mismatch.
Otherwise, this looks good!

pcapriotti:
> I am not familiar with qualified message sending, so probably it is no wonder I don't understand why an endpoint for it can return a client mismatch.

Because the client fails to include ciphertexts for all possible recipients and their devices, the message is rejected and not sent, so that the client can try again.

pcapriotti:
Good questions, @akshaymankar. I could have used `inject` for the first three constructors, but not for the last two, so what would be the point? In any case, I bit the bullet and implemented [some generic stuff](https://github.com/wireapp/wire-server/pull/1693/commits/0ebf3b6e2b49b680d28de9c78a1a8ce47b69e67e) using Generics.SOP which should let us define `AsUnion` instances automatically in the most common cases.

akshaymankar (approved):
Looks great :100:

PR #1700, opened 2021-08-09
Display helm test logs interactively

pcapriotti:
Use `kubectl logs -f` to display test output while integration tests are running.

## Checklist


akshaymankar (approved):
Looks good, I am paranoid about bash so I would recommend adding `set -euo pipefail` everywhere.

PR #1705, opened 2021-08-12
Minor typo fix, on word "Successfully"

arthurwolf:


fisx (approved):
You can just remove the check list from the description so it doesn't look so unchecked.

PR #1711, opened 2021-08-18
Make mapping between (team) permissions and roles more lenient.

fisx:
shouldn't change behavior for any existing users, but i found a user in our staging env that has an unknown set of permissions, and for that user is should fix https://wearezeta.atlassian.net/browse/SQSERVICES-720.

I've explained things in comments in the code, where we'll be able to find it in the future.

## Checklist

### DIFF
--- a/libs/galley-types/src/Galley/Types/Teams.hs
+++ b/libs/galley-types/src/Galley/Types/Teams.hs
@@ -149,13 +149,14 @@ rolePermissions :: Role -> Permissions
 rolePermissions role = Permissions p p where p = rolePerms role
 
 permissionsRole :: Permissions -> Maybe Role
-permissionsRole (Permissions p p') = if p /= p'
-  then do
-    -- we never did use @p /= p'@ for anything, fingers crossed that it doesn't occur anywhere
-    -- in the wild.  but if it does, this implementation prevents privilege escalation.
-    let p'' = Set.intersection p p'
-     in permissionsRole (Permissions p'' p'')
-  else permsRole p
+permissionsRole (Permissions p p') =
+  if p /= p'
+    then do
+      -- we never did use @p /= p'@ for anything, fingers crossed that it doesn't occur anywhere
+      -- in the wild.  but if it does, this implementation prevents privilege escalation.
+      let p'' = Set.intersection p p'
+       in permissionsRole (Permissions p'' p'')
+    else permsRole p
   where
     permsRole :: Set Perm -> Maybe Role
     permsRole perms =
--- a/libs/galley-types/src/Galley/Types/Teams.hs
+++ b/libs/galley-types/src/Galley/Types/Teams.hs
@@ -161,12 +161,12 @@ permissionsRole (Permissions p p') = if p /= p'
     permsRole perms =
       Maybe.listToMaybe
         [ role
-          | (perms', role) <- reverse . sortOn (length . fst) $ (\r -> (rolePerms r, r)) <$> [minBound ..],
+          | role <- [minBound ..],
             -- if a there is a role that is strictly less permissive than the perms set that
             -- we encounter, we downgrade.  this shouldn't happen in real life, but it has
             -- happened to very old users on a staging environment, where a user (probably)
             -- was create before the current publicly visible permissions had been stabilized.
-            perms' `Set.isSubsetOf` perms
+            rolePerms role `Set.isSubsetOf` perms
         ]
 
 -- | Internal function for 'rolePermissions'.  (It works iff the two sets in 'Permissions' are
--- a/libs/galley-types/src/Galley/Types/Teams.hs
+++ b/libs/galley-types/src/Galley/Types/Teams.hs
@@ -149,13 +149,13 @@ rolePermissions :: Role -> Permissions
 rolePermissions role = Permissions p p where p = rolePerms role
 
 permissionsRole :: Permissions -> Maybe Role
-permissionsRole (Permissions p p')
-  | p /= p' =
+permissionsRole (Permissions p p') = if p /= p'
+  then do
     -- we never did use @p /= p'@ for anything, fingers crossed that it doesn't occur anywhere
     -- in the wild.  but if it does, this implementation prevents privilege escalation.
     let p'' = Set.intersection p p'
      in permissionsRole (Permissions p'' p'')
-permissionsRole (Permissions p _) = permsRole p
+  else permsRole p
   where
     permsRole :: Set Perm -> Maybe Role
     permsRole perms =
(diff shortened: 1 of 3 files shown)


arianvp (changes requested):
Matthias says this doesn't work

fisx:
now it does.  i think.

libs/galley-types/test/unit/Test/Galley/Types.hs
```diff
@@ -54,7 +55,35 @@ tests =
         -- this test, and force future develpers to consider what permissions they want to set.
         assertBool "all covered" (all (roleHasPerm RoleExternalPartner) (ViewTeamFeature <$> [minBound ..])),
       testRoundTrip @FeatureFlags,
-      testRoundTrip @GuardLegalholdPolicyConflicts
+      testRoundTrip @GuardLegalholdPolicyConflicts,
+      testGroup
+        "permissionsRole, rolePermissions"
+        [ testCase "'Role' maps to expected permissions" $ do
+            assertEqual "role type changed" [minBound ..] [RoleOwner, RoleAdmin, RoleMember, RoleExternalPartner]
+            assertEqual "owner" (permissionsRole =<< newPermissions (intToPerms 8191) (intToPerms 8191)) (Just RoleOwner)
```
arianvp, line 63: it might be nicer to use hexadecimal, octal, or binary (does haskell have binary?) notation for these constants. I cant map base-10 to bitmaps in my head

libs/galley-types/src/Galley/Types/Teams.hs
```diff
@@ -149,13 +149,25 @@ rolePermissions :: Role -> Permissions
...
 
 permissionsRole :: Permissions -> Maybe Role
-permissionsRole (Permissions p p') | p /= p' = Nothing
+permissionsRole (Permissions p p')
+  | p /= p' =
+    -- we never did use @p /= p'@ for anything, fingers crossed that it doesn't occur anywhere
+    -- in the wild.  but if it does, this implementation prevents privilege escalation.
+    let p'' = Set.intersection p p'
+     in permissionsRole (Permissions p'' p'')
 permissionsRole (Permissions p _) = permsRole p
```
arianvp, line 158: line 152 and line 158 have overlapping patterns.   Will haskell actually fall back when the guard doesn't match? or would we have to move line 158 into an `| otherwise = ` ?

libs/galley-types/src/Galley/Types/Teams.hs
```diff
@@ -149,13 +149,25 @@ rolePermissions :: Role -> Permissions
...
   where
     permsRole :: Set Perm -> Maybe Role
     permsRole perms =
       Maybe.listToMaybe
-        [role | role <- [minBound ..], rolePerms role == perms]
+        [ role
+          | (perms', role) <- reverse . sortBy (compare `on` (length . fst)) $ (\r -> (rolePerms r, r)) <$> [minBound ..],
```
arianvp, line 164: `sortBy (compare `on` (length . fst))   -> sortOn (length . fst)`

arianvp, line 164: I'm a bit confused why we're sorting on the length of the permission set?  does that really do what we want?  We can have two permission sets that are the same size; but we must consider one permission set strictly less permissive than another (Perhaps based on the ordering of `Perm` ?)


e.g. the `rolePerms` for `RoleAdmin` and `RoleMember` have the same length; they both contain 4 elements; but `RoleMember <= RoleAdmin`

arianvp: Shouldn't we sort by the natural ordering of `Role` ? e.g. just iterate over `[minBound ..]`

So:


```
[ role |  role <- [minBound..],  rolePerms role `Set.isSubsetOf` perms]
```

libs/galley-types/test/unit/Test/Galley/Types.hs
fisx: me neither, but this is just protection against regression, and unrelated to the changes in this PR.

do you think it would be beneficial (*and* worth the time) to have it in hex or binary?  i personally don't think i could extract more useful information from that.

fisx: if you really want this i'm happy to do it in a separate PR that you can review any time in the future, would that be ok?

libs/galley-types/src/Galley/Types/Teams.hs
fisx: good idea, in this particular case it's the same, and the tests will show if it changes: c67cd00da2bd52128c458f88a0fc1a91d5bd0dcf

fisx: 36b591dbd945b06a0cf1fef212d273bc4d835638

libs/galley-types/src/Galley/Types/Teams.hs
fisx: it does fall back, you've hit on something, this here is actually more clear and robust: f12e85c58fdcbc9b00507c148e8e8dffb8a78dd7

libs/galley-types/test/unit/Test/Galley/Types.hs
arianvp: ah these are from the counter example you found in the wild? got it

libs/galley-types/src/Galley/Types/Teams.hs
```diff
@@ -149,13 +149,26 @@ rolePermissions :: Role -> Permissions
...
+      let p'' = Set.intersection p p'
+       in permissionsRole (Permissions p'' p'')
+    else permsRole p
   where
     permsRole :: Set Perm -> Maybe Role
     permsRole perms =
       Maybe.listToMaybe
-        [role | role <- [minBound ..], rolePerms role == perms]
+        [ role
+          | role <- [minBound ..],
```
arianvp, line 165:
```suggestion
          | role <-  reverse [minBound ..],
```

Rereading this this function is now incorrect I think.  It now gives the least permissive role that is a subset of `perms` but we want the most permissive role that is still a subset.   The current implementation maps all users to a `rolePerms minBound` as  `rolePerms minBound` is trivially the subset of almost all permission sets in the wild.

fisx: the tests would have caught it, but you're wrong about `Role`:

```
*Galley.Types.Teams> [minBound..]::[Role]
[RoleOwner,RoleAdmin,RoleMember,RoleExternalPartner]
```

arianvp: Ah so it is ordered exactly opposite of what I expected. All good then :)

arianvp (approved):

PR #1723, opened 2021-08-30
 Avoid remote calls to get conversation when it is not found locally

akshaymankar:
Depends on #1703
Re-work of #1713 

## Checklist

### DIFF
--- a/services/galley/src/Galley/API/Query.hs
+++ b/services/galley/src/Galley/API/Query.hs
@@ -101,15 +101,18 @@ getConversation zusr cnv = do
   localDomain <- viewFederationDomain
   if qDomain cnv == localDomain
     then getUnqualifiedConversation zusr (qUnqualified cnv)
-    else getRemoteConversation zusr (toRemote cnv)
-
-getRemoteConversation :: UserId -> Remote ConvId -> Galley Public.Conversation
-getRemoteConversation zusr remoteConvId = do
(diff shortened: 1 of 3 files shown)


services/galley/src/Galley/API/Query.hs
```diff
@@ -101,15 +101,18 @@ getConversation zusr cnv = do
...
-  conversations <- getRemoteConversations zusr [remoteConvId]
-  case conversations of
-    [] -> throwErrorDescription convNotFound
-    [conv] -> pure conv
-    _convs -> throwM (federationUnexpectedBody "expected one conversation, got multiple")
+    else getRemoteConversation (toRemote cnv)
+  where
+    getRemoteConversation :: Remote ConvId -> Galley Public.Conversation
+    getRemoteConversation remoteConvId = do
+      (foundConvs, _) <- Data.remoteConversationIdOf zusr [remoteConvId]
```
pcapriotti, line 108: There is a compile error here. Should it just be `foundConvs <-`?

pcapriotti (commented):
Looks good, but there is a compile error.

services/galley/src/Galley/API/Query.hs
akshaymankar: Ah yes, I refactored this. Note to self: it is seldom worth it to work on top of unmerged PR.

pcapriotti (approved):
:rocket:

PR #1726, opened 2021-08-30
Servantify POST /connections endpoint; remove deprecated 'message' field

jschaul:
* Servantify `POST /connections`
* Remove the `message` field in connection requests, as it has been long deprecated. Some clients may still send it (as it was a mandatory field), but it was never visible at UI level in the past years, so we can remove it.
* fix an old TODO to make the 256 length check using the Range type, rather than doing it implicitly in the json parser.

The main changeset to types is in  libs/wire-api/src/Wire/API/Connection.hs 

Some golden tests have been removed as they are no longer relevant.

## Checklist


libs/wire-api/src/Wire/API/Connection.hs
```diff
@@ -103,12 +96,23 @@ data UserConnection = UserConnection
...
+
+instance P.ToSchema UserConnection where
+  schema =
+    P.object "UserConnection" $
+      UserConnection
+        <$> ucFrom P..= P.field "from" P.schema
+        <*> ucTo P..= P.field "to" P.schema
+        <*> ucStatus P..= P.field "status" P.schema
+        <*> ucLastUpdate P..= P.field "last_update" P.schema
+        <*> ucConvId P..= P.opt (P.field "conversation" P.schema)
```
pcapriotti, line 113: This is fine, but keep in mind that `P.optField` would behave better in the case where the `conversation` field is present but contains errors.

libs/wire-api/src/Wire/API/ErrorDescription.hs
```diff
@@ -221,6 +221,24 @@ type NotConnected = ErrorDescription 403 "not-connected" "Users are not connecte
...
+
+invalidUser :: InvalidUser
+invalidUser = mkErrorDescription
+
+type NoIdentity = ErrorDescription 403 "no-identity" "The user has no verified identity (email or phone number)."
+
+-- FUTUREWORK: if that additional error number in the message has any importance
+-- (it probably doesn't matter), maybe there's a way to add it to the
+-- description here.
+noIdentity :: Int -> NoIdentity
```
pcapriotti, line 239: If you want to add the error code to the description, you can just use the `ErrorDescription` constructor, which takes a `Text` to use as the description, instead of the static one that is part of the type. You could do something like *(I haven't type-checked this)*:
```haskell
noIdentity :: (NoIdentity ~ ErrorDescription code lbl desc) => Int -> NoIdentity
noIdentity n = ErrorDescription (Text.pack (symbolVal (Proxy @desc)) <> " (code "<> Text.pack (show n) <> ")")
```

libs/wire-api/src/Wire/API/Routes/Public/Util.hs
```diff
@@ -0,0 +1,40 @@
...
+  toUnion (Existed200 x) = Z (I x)
+  toUnion (Created201 x) = S (Z (I x))
+
+  fromUnion (Z (I x)) = Existed200 x
+  fromUnion (S (Z (I x))) = Created201 x
+  fromUnion (S (S x)) = case x of
+
+data ResponseForExistedCreated a
+  = Existed200 !a
+  | Created201 !a
```
mdimjasevic, line 40: I wouldn't put HTTP status codes in the names of these constructors. We've been decoupling HTTP responses from handler responses, and making a connection between the two levels happens in other ways.

services/brig/src/Brig/API/Public.hs
```diff
@@ -1115,13 +1090,12 @@ customerExtensionCheckBlockedDomains email = do
         when (domain `elem` blockedDomains) $
           throwM $ customerExtensionBlockedDomain domain
 
-createConnectionH :: JSON ::: UserId ::: ConnId ::: JsonRequest Public.ConnectionRequest -> Handler Response
-createConnectionH (_ ::: self ::: conn ::: req) = do
-  cr <- parseJsonBody req
+createConnectionH :: UserId -> ConnId -> Public.ConnectionRequest -> Handler (Public.ResponseForExistedCreated Public.UserConnection)
+createConnectionH self conn cr = do
```
mdimjasevic, line 1094: Perhaps this is just nitpicking, but I got the impression that the naming convention where `H` is appended to handler names applies only to old `wai-route` endpoint handlers.

libs/wire-api/src/Wire/API/Routes/Public/Util.hs
pcapriotti, line 40: Please remove the status codes from these names. The point of the last type argument of `MultiVerb` is to decouple HTTP-specific things (like status codes) from the handler.

services/brig/src/Brig/API/Public.hs
```diff
@@ -1115,13 +1090,12 @@ customerExtensionCheckBlockedDomains email = do
...
   rs <- API.createConnection self cr conn !>> connError
   return $ case rs of
-    ConnectionCreated c -> setStatus status201 $ json (c :: Public.UserConnection)
-    ConnectionExists c -> json (c :: Public.UserConnection)
+    ConnectionCreated c -> Public.Created201 c
```
pcapriotti, line 1097: There's no need to have this conversion here. We can just use the `ResponseForExistedCreated` type as the return type of `createConnection`.

services/brig/src/Brig/IO/Intra.hs
```diff
@@ -548,7 +548,7 @@ createConnectConv from to cname mess conn = do
         . zUser from
         . maybe id (header "Z-Connection" . fromConnId) conn
         . contentJson
-        . lbytes (encode $ Connect to (messageText <$> mess) cname Nothing)
+        . lbytes (encode $ Connect to Nothing cname Nothing)
```
pcapriotti, line 551: Can the `message` field be removed from the `Connect` event type? This perhaps belongs in a separate PR, though.

services/galley/schema/src/Main.hs
```diff
@@ -98,6 +98,9 @@ main = do
       V52_FeatureConferenceCalling.migration
       -- When adding migrations here, don't forget to update
       -- 'schemaVersion' in Galley.Data
+      --
+      -- FUTUREWORK: once #1726 has made its way to master/production,
+      -- the 'message' field in connections table can be dropped.
```
pcapriotti, line 103: Probably a stupid question, but why can't it be dropped in this PR?

pcapriotti (changes requested):
This looks good.

I would suggest a minor improvement, possibly in a separate PR: now that you have nicely abstracted out this pattern of "existed200/created201", it would be useful to also abstract out the corresponding `MultiVerb` response list, so that the association between the two responses and the status code is fixed and users cannot get it wrong.

For example, we can add a type to the `Util` module like:
```haskell
type ResponsesForExistedCreated eDesc cDesc a = '[
  Respond 200 eDesc a,
  Respond 201 cDesc a ]
```
and use that as part of the `MultiVerb` type in the routes that use the pattern.

More suggestions below.

services/galley/schema/src/Main.hs
mdimjasevic, line 103: Why does this go as a separate PR, and not part of this PR?

libs/wire-api/src/Wire/API/Routes/Public/Brig.hs
```diff
@@ -309,6 +303,36 @@ data Api routes = Api
...
+    -- - ConvCreate event to self, if creating a connect conversation (via galley)
+    -- - ConvConnect event to self, in some cases (via galley),
+    --   for details see 'Galley.API.Create.createConnectConversation'
+    --
+    createConnection ::
+      routes :- Summary "Create a connection to another user."
+        :> CanThrow MissingLegalholdConsent
+        :> CanThrow InvalidUser
+        :> CanThrow ConnectionLimitReached
+        :> CanThrow NoIdentity
```
mdimjasevic, line 319: If you feel like going an extra mile, it would be great to see these in the return type instead and have tests for them.

mdimjasevic (commented):
I left some comments inlined.

Overall, this looks good, though I wish more things were reflected in types given the move to Servant.

libs/wire-api/src/Wire/API/Connection.hs
jschaul: 👍🏿 good catch, thanks!

libs/wire-api/src/Wire/API/ErrorDescription.hs
```diff
@@ -221,6 +221,24 @@ type NotConnected = ErrorDescription 403 "not-connected" "Users are not connecte
...
+noIdentity _i = mkErrorDescription
```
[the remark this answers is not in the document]
jschaul: I tried that, but failed to get it to type check. 
Apparently it requires a constraint (

```Haskell
forall code lbl desc. (NoIdentity ~ ErrorDescription code lbl desc) =>
```
(thanks @pcapriotti )) for it to compile. I'm not sure why this is, and I have a hard time even googling or hoogling for the `~` operator here. Can anyone point me to documentation on what `~` does and when it's necessary?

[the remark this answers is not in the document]
pcapriotti: It's an equality constraint. Docs should be [here](https://ghc.gitlab.haskell.org/ghc/doc/users_guide/exts/equality_constraints.html?highlight=equality). Basically, when the type checker encounters `a ~ b`, it will try to unify `a` and `b`, and fail if it cannot. As a by-product of this unification, other equality constraints might be generated.

In this case, I suggested to use an equality constraint to get access to the `desc` type variable implicit in the definition of `NoIdentity`. Essentially, this is like pattern matching at the type level. Alternatively, one could define a type family `Description :: * -> Symbol`, instantiate it for `ErrorDescription` and  use that instead, but this seemed quicker.

As for the explicit quantification, I'm also not sure why it's needed. It's most likely just a GHC quirk, don't think too much about it ;)

services/galley/schema/src/Main.hs
jschaul: I wrote some documentation about this, here: https://github.com/wireapp/wire-server/pull/1747/files

libs/wire-api/src/Wire/API/Routes/Public/Brig.hs
jschaul: I'm not a fan. Documenting errors is nice, but making everything complicated by having all possible errors in the type feels very unwieldy.

services/galley/schema/src/Main.hs
jschaul: I wrote some documentation about this, here: https://github.com/wireapp/wire-server/pull/1747/files

pcapriotti: Makes sense, thanks!

services/brig/src/Brig/IO/Intra.hs
jschaul: 👍🏿 I added a futurework for another PR.

services/brig/src/Brig/API/Public.hs
jschaul: Makes sense 👍🏿

libs/wire-api/src/Wire/API/ErrorDescription.hs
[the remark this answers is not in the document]
jschaul: Okay, thanks for the pointer to the docs!

services/brig/src/Brig/API/Public.hs
jschaul: Good point. Fixed.

libs/wire-api/test/unit/Test/Wire/API/Golden/Generated/UserConnection_user.hs
```diff
@@ -61,13 +52,6 @@ testObject_UserConnection_user_3 =
       ucTo = Id (fromJust (UUID.fromString "00000002-0000-0001-0000-000000000001")),
       ucStatus = Pending,
       ucLastUpdate = fromJust (readUTCTimeMillis "1864-05-07T19:25:59.201Z"),
-      ucMessage =
```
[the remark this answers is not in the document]
jschaul: correct 😸

libs/wire-api/src/Wire/API/Connection.hs
```diff
@@ -280,61 +241,37 @@ data ConnectionRequest = ConnectionRequest
...
   deriving (Arbitrary) via (GenericUniform ConnectionUpdate)
+  deriving (FromJSON, ToJSON, S.ToSchema) via (P.Schema ConnectionUpdate)
 
+instance P.ToSchema ConnectionUpdate where
+  schema =
+    P.object "ConnectionUpdate" $
+      ConnectionUpdate
+        <$> cuStatus P..= P.fieldWithDocModifier "status" (P.description ?~ "New relation status") P.schema
+
+-- TODO remove
```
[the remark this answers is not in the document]
jschaul: actually it can't be removed yet as it's still in use; this can only be removed once more endpoints have been servantified (in future PRs)

libs/wire-api/src/Wire/API/Connection.hs
```diff
@@ -103,12 +96,23 @@ data UserConnection = UserConnection
...
+  schema =
+    P.object "UserConnection" $
+      UserConnection
+        <$> ucFrom P..= P.field "from" P.schema
+        <*> ucTo P..= P.field "to" P.schema
+        <*> ucStatus P..= P.field "status" P.schema
+        <*> ucLastUpdate P..= P.field "last_update" P.schema
+        <*> ucConvId P..= P.opt (P.field "conversation" P.schema)
+
+-- TODO remove
```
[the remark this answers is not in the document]
jschaul: actually it can't be removed yet as it's still in use; this can only be removed once more endpoints have been servantified (in future PRs). Removed the comment.

libs/wire-api/src/Wire/API/Connection.hs
```diff
@@ -68,27 +65,23 @@ data UserConnectionList = UserConnectionList
...
+  deriving (FromJSON, ToJSON, S.ToSchema) via (P.Schema UserConnectionList)
 
+instance P.ToSchema UserConnectionList where
+  schema =
+    P.object "UserConnectionList" $
+      UserConnectionList
+        <$> clConnections P..= P.field "connections" (P.array P.schema)
+        <*> clHasMore P..= P.fieldWithDocModifier "has_more" (P.description ?~ "Indicator that the server has more connections than returned.") P.schema
+
+-- TODO remove
```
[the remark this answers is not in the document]
jschaul: actually it can't be removed yet as it's still in use; this can only be removed once more endpoints have been servantified (in future PRs). Removed the comment.

libs/wire-api/src/Wire/API/Connection.hs
```diff
@@ -280,61 +241,37 @@ data ConnectionRequest = ConnectionRequest
   { -- | Connection recipient
     crUser :: UserId,
     -- | Name of the conversation to be created
-    crName :: Text,
-    -- | Initial message
-    crMessage :: Message
+    -- FUTUREWORK investigate: shouldn't this name be optional? Do we use this name actually anywhere?
```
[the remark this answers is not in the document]
jschaul: > I'm just guessing now, but does a connection name ever make it to a one-to-one conversation title?

I'm not sure, I didn't take the time to look more deeply into how this is used and what clients send and receive and what they do with it. That's why there is a FUTUREWORK here. A potential refactor that removes this can be done in a subsequent PR.

libs/wire-api/src/Wire/API/Routes/Public/Util.hs
jschaul: Done.

jschaul: If you insist, okay.

jschaul:
I tried to address all you review comments now.

pcapriotti (approved):
Looks good! :rocket:

PR #1730, opened 2021-08-31
Update client certificates automatically

pcapriotti:
This makes federator update its cached TLS settings when changes to client certificates or the CA store are detected.

Most of the functionality is contained in `Federator.Monitor.Internal`. There is some complicated logic to make sure we get inotify events when the paths containing the credentials change in any possible way, including being modified directly, or overwritten with other files, or their parent directories being replaced with other ones.

## Details

A "monitor" is started when an environment is created. At this point, all the files that need to be watched are collected, and inotify watches are created for each one of them and each of their parent directories up to the filesystem root. Whenever we detect a relevant change, we reload the `TLSSettings` from the filesystem, and update an `IORef`.

For file watches, we listen for the `Close` event, which is sent when a file descriptor in write mode for the corresponding inode is closed. For directory watches, things are more complicated: we listen for `Create` and `MoveIn` events, which happen when a new file appear in the directory, and as soon as it happens (before reloading the settings) we replace the existing watch on the file with a new one (since the inode might have changed).

There is a tiny PR to [hs-certificate](https://github.com/vincenthz/hs-certificate) associated to this PR: https://github.com/vincenthz/hs-certificate/pull/125. This was needed to make sure we are able to catch and log exceptions in inotify handlers. Using `error` is problematic because the exception can be (and indeed is, in my tests) thrown outside of `IO` code we control, and results in the handler thread dying silently. I have pointed `stack.yaml` to the patched version, but it is not essential. Without the patch, we might just lose some logging errors in some cases.

## Testing

In tests, we create a setup using temporary files or directories, then we set up a channel (using `Control.Concurrent.Chan`) to keep track of asynchronous reloads or exceptions thrown within a reload.

## TODO

 - [x] ~~Graceful handling of missing inotify support~~
                 This is probably not needed, since the inotify library cannot even be built for targets that do not support inotify. 
- [x] Test updating a grandparent directory
 - [x] Test modifying a file after a parent directory has been replaced
 - [x] ~~Integration tests~~
                This is not trivial, because it needs an outward service integration test. I'll skip it for now.
 - [x] Unit test directory traversal

## Checklist


services/federator/src/Federator/Monitor/Internal.hs
```diff
@@ -0,0 +1,355 @@
...
+  Sem r a
+logErrors action = Polysemy.catch action $ \err -> do
+  Log.err $
+    Log.msg ("federation setup error while updating certificates" :: Text)
+      . Log.field "error" (showFederationSetupError err)
+  Polysemy.throw err
+
+logAndIgnoreErrors ::
+  Member TinyLog r =>
+  Sem (Polysemy.Error FederationSetupError ': r) () ->
```
mdimjasevic, line 114: Can this be only of the unit return type or can it be any `a` in `Sem (Polysemy.Error FederationSetupError ': r) ()`?

services/federator/src/Federator/Monitor/Internal.hs
```diff
@@ -0,0 +1,355 @@
...
+    embed (readCertificateStore path)
+      >>= maybe (Polysemy.throw (InvalidCAStore path)) pure
+  systemCAStore <-
+    if useSystemCAStore settings
+      then embed getSystemCertificateStore
+      else pure mempty
+  pure (customCAStore <> systemCAStore)
+
+mkCreds ::
+  forall r.
```
mdimjasevic, line 336: Just nit-picking: perhaps this quantification is not used anywhere?

services/federator/test/integration/Test/Federator/Monitor.hs
```diff
@@ -0,0 +1,26 @@
...
+
+module Test.Federator.Monitor (spec) where
+
+import Imports
+import Test.Federator.Util
+import Test.Hspec
+
+spec :: TestEnv -> Spec
+spec _env =
+  describe "Monitor" $ pure ()
```
mdimjasevic, line 26: Is this an unfinished test?

services/federator/src/Federator/Monitor/Internal.hs
pcapriotti: Since errors are ignored, it can only be `()`.

services/federator/src/Federator/Monitor/Internal.hs
pcapriotti: That's right, thanks. Fixed in 0449a7904f1b20726d2cc0bc228c41cfc5d19b78.

services/federator/test/integration/Test/Federator/Monitor.hs
pcapriotti: It was a stub for the integration test I was planning to write. I removed it for now. See 52a357c54377fc039ded6868d4330fce17851e58.

pcapriotti:
With the last few fixes, certificate autoreload works nicely on kubernetes.

stack.yaml
```diff
@@ -243,6 +243,12 @@ extra-deps:
 - git: https://github.com/wireapp/snappy
   commit: b0e5c08af48911caecffa4fa6a3e74872018b258 # master (Sep 03, 2021)
 
+# Error handling fix
+- git: https://github.com/wireapp/hs-certificate
+  commit: e3ea2e1166f0569982a85aad9bc9de8f5b2994c1 # master (Aug 31, 2021)
+  subdirs:
+    - x509-store
```
akshaymankar, line 250: We should upstream this

akshaymankar (changes requested):
I think we shouldn't expose `IORef TLSSettings` to the app.

hack/bin/integration-test-logs.sh
```diff
@@ -1,13 +0,0 @@
...
-if [[ -z "$NAMESPACE" ]]; then
-  echo "NAMESPACE not set"
-  exit 1
-fi
-
-while IFS= read LINE; do
-  if [[ "$LINE" =~ ^Pod\ (.*)\ running$ ]]; then
-    kubectl -n "$NAMESPACE" logs "${BASH_REMATCH[1]}" -f
-  fi
-done
```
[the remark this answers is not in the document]
pcapriotti: We don't need to remove this script. This was just supposed to be a temporary hack. I'll just remove this commit before merging.

hack/bin/integration-test.sh
```diff
@@ -1,11 +1,8 @@
 #!/usr/bin/env bash
-set -euo pipefail
```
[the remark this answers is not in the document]
pcapriotti: See above.

services/federator/src/Federator/Remote.hs
```diff
@@ -115,7 +115,7 @@ mkGrpcClient ::
   Members
     '[ Embed IO,
        Polysemy.Error RemoteError,
-       Polysemy.Reader TLSSettings
+       Polysemy.Reader (IORef TLSSettings)
```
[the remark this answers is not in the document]
pcapriotti: I'm not sure what you have in mind. If `TLSSettings` is just a pure value, how can we "change" it globally from another thread?

stack.yaml
pcapriotti: I sent a PR, I'll include a link in a comment: https://github.com/vincenthz/hs-certificate/pull/125.

hack/bin/integration-test.sh
```diff
@@ -1,11 +1,8 @@
...
 
 NAMESPACE=${NAMESPACE:-test-integration}
 
 echo "Running integration tests on wire-server"
 
-DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 CHART=wire-server
-helm test -n "${NAMESPACE}" "${NAMESPACE}-${CHART}" --timeout 600s |
-  "$DIR/integration-test-logs.sh"
+helm test --logs -n "${NAMESPACE}" "${NAMESPACE}-${CHART}" --timeout 600s
```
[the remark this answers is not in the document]
pcapriotti: Github is diffing from the branching point, not from develop.

services/federator/src/Federator/Remote.hs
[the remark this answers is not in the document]
akshaymankar: Yeah, it can live in the `Env` type as an implementation detail, we don't expose that to the app. The app should always just be able to read it.

hack/bin/integration-test.sh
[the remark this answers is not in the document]
akshaymankar: Ah, makes sense thanks!

services/federator/src/Federator/Remote.hs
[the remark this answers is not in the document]
pcapriotti: Done in 51923d10b86b1702f138de24bd591dd1dc6751c0, but I went with an `Input` effect instead of `Reader`, for two reasons:
 1. It is a more straightforward first order effect, and we don't have to deal with `Local` when interpreting it
 2. There is already a ready-made function to interpret it by running an action every time

akshaymankar (approved):
:rocket:

PR #1746, opened 2021-09-08
Report All Failures for Golden Tests at Once

mdimjasevic:
This PR reports all failures in a golden test group before exiting and reporting the first failure. It turns each test object into a test case, compared to the way it has been, which is presenting a group of test cases as a single test case.

## Checklist


libs/wire-api/test/unit/Test/Wire/API/Golden/Runner.hs
```diff
@@ -34,37 +35,87 @@ import Data.ProtoLens.Message (Message)
...
+  ( Members '[Error (TestOutcomeError a), Embed IO] r,
+    Typeable a,
+    ToJSON a,
+    FromJSON a,
+    Eq a,
+    Show a
+  ) =>
+  a ->
+  FilePath ->
+  Sem r ()
```
akshaymankar, line 101: I think it would be much simpler if this function just returned `IO (Either TestOutcomeError ())` instead of involving polysemy here.

libs/wire-api/test/unit/Test/Wire/API/Golden/Runner.hs
```diff
@@ -34,37 +35,87 @@ import Data.ProtoLens.Message (Message)
...
+        <> show result
+
 testObjects :: forall a. (Typeable a, ToJSON a, FromJSON a, Eq a, Show a) => [(a, FilePath)] -> IO ()
 testObjects objs = do
-  allFilesExist <- and <$> traverse (uncurry testObject) objs
-  assertBool "Some golden JSON files do not exist" allFilesExist
+  allResults <- traverse (run . uncurry testObject) objs
+  let errorResults = foldMap (either pure (const [])) allResults
+      errorMsgs = intercalate "\n" . fmap show $ errorResults
+  assertBool errorMsgs . null $ errorResults
```
akshaymankar, line 83: Instead of traversing and collecting errors, could we make each one these a `testCase`? That way we don't have to invent any of this.

libs/wire-api/test/unit/Test/Wire/API/Golden/Runner.hs
mdimjasevic: That's a perfectly valid choice. In that case we don't have a short-circuiting mechanism of a monad with errors (like we do with `Sem r` that has `Error`s) and instead we have to implement our own nested if-then-else thingy. Should I do that?

akshaymankar: We are pretty much doing the nested conditional logic anyway. But I think it would be more promising if we relied on tasty to do all of this for us by using `testCase` for each test object instead of making all these assertions in one `testCase`.

libs/wire-api/test/unit/Test/Wire/API/Golden/Runner.hs
mdimjasevic: Done!

As a consequence, the output of `stack test wire-api` is much more verbose and (correctly) reports 3549 passing tests (compared to ~570 tests before turning each object into a test case).

libs/wire-api/test/unit/Test/Wire/API/Golden/Runner.hs
mdimjasevic: I just got rid of Polysemy here, and each test object is now a test case on its own.

akshaymankar (approved):

PR #1748, opened 2021-09-09
Fix swagger for `Range` and `Domain`

akshaymankar:
## Checklist

### DIFF
--- a/libs/types-common/src/Data/Range.hs
+++ b/libs/types-common/src/Data/Range.hs
@@ -55,7 +55,7 @@ module Data.Range
 where
 
 import Cassandra (ColumnType, Cql (..), Tagged, retag)
-import Control.Lens ((?~))
+import Control.Lens ((%~), (?~))
 import Data.Aeson (FromJSON (parseJSON), ToJSON (toJSON))
 import Data.Aeson.Types as Aeson (Parser)
 import qualified Data.Attoparsec.ByteString as Atto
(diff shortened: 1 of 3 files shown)


libs/types-common/src/Data/Range.hs
```diff
@@ -121,7 +121,7 @@ instance (Within a n m, FromJSON a) => FromJSON (Range n m a) where
       msg sn sm = fail (errorMsg (fromSing sn) (fromSing sm) "")
 
 rangedSchema ::
-  (Within a n m, Bounds b) =>
+  (Within a n m, Bounds b, HasRangedSchemaDocModifier d b) =>
```
pcapriotti, line 124:
```suggestion
  (Within a n m, HasRangedSchemaDocModifier d b) =>
```

libs/types-common/src/Data/Range.hs
```diff
@@ -131,18 +131,55 @@ rangedSchema sn sm sch = Range <$> untypedRangedSchema (get sn) (get sm) sch
     get = toInteger . fromSing
 
 untypedRangedSchema ::
-  Bounds b =>
+  forall d v w a b.
+  (Bounds b, HasRangedSchemaDocModifier d b) =>
```
pcapriotti, line 135:
```suggestion
  (HasRangedSchemaDocModifier d b) =>
```

pcapriotti (approved):
Looks good.

PR #1749, opened 2021-09-09
New process for CHANGELOG entries

pcapriotti:
This introduces a new process for recording CHANGELOG entries, as previously discussed.

##  tl;dr

Entries have to be written in individual files in `changelog.d`, with omitted PR numbers.

## Details

On every pull request, one is supposed to create a new file in the appropriate subdirectory of `changelog.d`, containing just the text of the corresponding changelog entry. There is no need to explicitly write a PR number, because the `mk-changelog.sh` script will add it automatically at the end. The name of the file does not matter, but please try to make it unique to avoid unnecessary conflicts (e.g. use the branch name).

It is still possible to write the PR number manually if so desired, which is useful in case the entry should refer to multiple PRs. In that case, the script leaves the PR number reference intact, as long as it is at the very end of the entry (no period allowed afterwards!), and in brackets. It is also possible to use the pattern `##` to refer to the current PR number. This will be replaced throughout.

Multiline entries are supported, and should be handled correctly. Again, the PR reference should either be omitted or put at the very end. If multiple entries for a single PR are desired, one should create a different file for each of them.

## Generating a CHANGELOG for a release

Just run the script `changelog.d/mk-changelog.sh` with no arguments. It will print all the entries, nicely formatted, on standard output. The script gets PR numbers from the `git` log. If that fails for any reason (e.g. if an entry was added outside of a PR), make sure that the entry has a manually specified PR number.

There is a branch https://github.com/wireapp/wire-server/tree/akshaymankar/changelog-d-test with some fake merge commits to test that PR number fetching works properly.

## Checklist


changelog.d/mk-changelog.sh
```diff
@@ -0,0 +1,24 @@
...
+
+for d in "$DIR"/*; do
+    if [[ ! -d "$d" ]]; then continue; fi
+
+    echo -n "## "
+    sed '$ a\' "$d/.title"
+    echo ""
+    for f in "$d"/*; do
+        pr=$(getPRNumber $f)
+        sed -r -e '1 { s/^/\* /; }' -e '1 !{ s/^/  /; }' -e "s/##/$pr/g" -e "$ { /^.*\((#.*)\)$/ ! { s/$/ ($pr)/; } }" -e 's/\s+$//' -e '$ a\' "$f"
```
jschaul, line 21: Could you add a comment to describe what this line does? It's a bit daunting to read...

pcapriotti: I added some comments in b97612e7414aa32863289849329795855d18f5a6.

PR #1750, opened 2021-09-09
Increase federator monitor test timeout to 10s

pcapriotti:
Apparently, 1 second is not enough in CI.

## Checklist


mdimjasevic (approved):
I wish we used types for time units to avoid all kinds of time conversion errors.

PR #1751, opened 2021-09-10
Remove otr_muted from Member and related types

pcapriotti:
The `otr_muted` field has been deprecated for some time, and the clients do not need it. Note that the corresponding conversation field in the database is still present, and removing that is left to future work.

## Checklist


services/galley/test/integration/API/Roles.hs
```diff
@@ -182,7 +182,7 @@ wireAdminChecks cid admin otherAdmin mem = do
   let activatedAccess = ConversationAccessUpdate [InviteAccess] NonActivatedAccessRole
   putAccessUpdate admin cid activatedAccess !!! assertActionSucceeded
   -- Update your own member state
-  let memUpdate = memberUpdate {mupOtrMute = Just True}
+  let memUpdate = memberUpdate {mupOtrArchive = Just True}
```
mdimjasevic, line 185: Can you clarify this change? How come you use the archive attribute here instead of the muted status? The same question goes for the change on line 233 below.

mdimjasevic (approved):
This seems like a straightforward and simple erasure change. Just please clarify that change in tests where you tested an archive status instead of muted status.

services/galley/test/integration/API/Roles.hs
pcapriotti: This is just testing that updating works. It doesn't really matter what we update specifically. Since `mupOtrMute` has been removed, I just randomly changed it to `mupOtrArchive`.

PR #1732, opened 2021-09-01
Clean up JSON Golden Tests in wire-api, Part #2

mdimjasevic:
This is the second PR in a series of PRs that clean up automatically generated JSON golden tests in `wire-api`.

## Checklist

 - [x] The **PR Title** explains the impact of the change.
 - [x] The **PR description** provides context as to why the change should occur and what the code contributes to that effect. This could also be a link to a JIRA ticket or a Github issue, if there is one.
 - [x] If end-points have been added or changed: the **endpoint / config-flag checklist** (see Wire-employee only backend [wiki page](https://github.com/zinfra/backend-wiki/wiki/Checklists)) has been followed.
 - [x] If a schema migration has been added, I ran **`make git-add-cassandra-schema`** to update the cassandra schema documentation.
 - [x] Section *Unreleased* of **CHANGELOG-draft.md** contains the following bits of information:
   - [x] A line with the title and number of the PR in one or more suitable sub-sections.
   - [x] If /a: measures to be taken by instance operators.
   - [x] If /a: list of cassandra migrations.
   - [x] If public end-points have been changed or added: does nginz need upgrade?
   - [x] If internal end-points have been added or changed: which services have to be deployed in a specific order?

PR #1762, opened 2021-09-13
Clean up JSON Golden Tests, Part #5

mdimjasevic:
This patch further cleans up JSON golden tests. This is continuation of PRs #1729, #1732, #1733 and #1756.

## Checklist


jschaul (approved):
Nitpick: For the title, using the `#` symbol for parts causes github to link to PRs with that number in the changelog later, which is confusing. Perhaps we can omit this symbol from the title and changelog entry?

PR #1763, opened 2021-09-13
#1755 follow-up

fisx:
Stuff that should have been part of #1755, but wasn't due to time pressure.  See CHANGELOG and individual commits for details.  It probably makes sense to read this changelog.d first, then commit-by-commit.

## Checklist


services/spar/src/Spar/API.hs
```diff
@@ -375,7 +375,7 @@ validateNewIdP apiversion _idpMetadata teamId mReplaces = withDebugLog "validate
     Data.GetIdPFound idp' {- same team -} -> handleIdPClash (Right idp')
     Data.GetIdPNotFound -> pure ()
     res@(Data.GetIdPDanglingId _) -> throwSpar . SparIdPNotFound . ("validateNewIdP: " <>) . cs . show $ res -- database inconsistency
-    res@(Data.GetIdPNonUnique _) -> throwSpar . SparIdPNotFound . ("validateNewIdP: " <>) . cs . show $ res -- impossible
+    Data.GetIdPNonUnique ids' {- same team didn't yield anything, but there are at least two other teams with this issuer already -} -> handleIdPClash (Left ids')
```
pcapriotti, line 378: It seems unnecessary to distinguish between the "non unique" case and the "wrong team", since "non unique" implies "wrong team", and they can still be distinguished by the size of the list.

services/spar/src/Spar/App.hs
```diff
@@ -199,19 +200,39 @@ insertUser uref uid = wrapMonadClient $ Data.insertSAMLUser uref uid
...
+          | isNothing (userTeam user) -> pure GetUserNoTeam
+          | mbteam /= userTeam user -> pure GetUserWrongTeam
+          | otherwise -> pure $ GetUserFound user
+
+data GetUserResult usr
+  = GetUserFound usr
+  | GetUserNotFound
+  | GetUserNoTeam
+  | GetUserWrongTeam
+  deriving (Eq, Show)
```
pcapriotti, line 225: I think you can just `deriving Functor`.

pcapriotti (approved):
Discussed in a call.

services/spar/src/Spar/App.hs
fisx: Thanks, will fix in an upcoming PR.  (Also for `GetIdPResult` in `Spar.Data`.)

services/spar/src/Spar/API.hs
fisx: Good idea, but separate PR.  (And I think this may become more obvious of a change if I make the CRUD API for IdPs in `Spar.Data` less abstract, and expose functions that do precisely what's needed on the call side, and nothing else.)

PR #1771, opened 2021-09-15
Servantify the DELETE /self endpoint

mdimjasevic:
This patch rewrites the `DELETE /self` endpoint to Servant.

## Checklist


libs/wire-api/src/Wire/API/Routes/Public/Brig.hs
```diff
@@ -16,16 +15,19 @@
 --
 -- You should have received a copy of the GNU Affero General Public License along
 -- with this program. If not, see <https://www.gnu.org/licenses/>.
+{-# OPTIONS_GHC -fno-warn-orphans #-}
```
mdimjasevic, line 18: I don't like this so if you have any suggestions where to put that instance for `AsUnion DeleteSelfResponses (Maybe Timeout)`, I am all ears.

libs/types-common/src/Data/Code.hs
```diff
@@ -95,20 +95,27 @@ instance ToHttpApiData Value where
...
 -- | A 'Timeout' is rendered as an integer representing the number of seconds remaining.
 instance ToByteString Timeout where
   builder (Timeout t) = builder (round t :: Int32)
 
 -- | A 'Timeout' is rendered in JSON as an integer representing the
 -- number of seconds remaining.
-instance ToJSON Timeout where
-  toJSON (Timeout t) = toJSON (round t :: Int32)
+instance A.ToJSON Timeout where
+  toJSON (Timeout t) = A.toJSON (round t :: Int32)
```
pcapriotti, line 113: Remove this instance and derive it `via Schema Timeout`.

libs/types-common/src/Data/Code.hs
```diff
@@ -95,20 +95,27 @@ instance ToHttpApiData Value where
...
 
 -- | A 'Timeout' is parsed from JSON as an integer representing the
 -- number of seconds remaining.
-instance FromJSON Timeout where
-  parseJSON = withScientific "Timeout" $ \n ->
+instance A.FromJSON Timeout where
+  parseJSON = A.withScientific "Timeout" $ \n ->
```
pcapriotti, line 118: Same as above.

services/brig/src/Brig/API/Error.hs
```diff
@@ -37,16 +37,16 @@ import Network.HTTP.Types.Header
 import Network.HTTP.Types.Status
 import qualified Network.Wai.Utilities.Error as Wai
 import Servant.API.Status
-import Wire.API.ErrorDescription
+import qualified Wire.API.ErrorDescription as ErrDesc
```
pcapriotti, line 40: Why?

pcapriotti (approved):
Looks good! Minor comments follow.

services/brig/src/Brig/API/Error.hs
mdimjasevic: I introduced a few `ErrorDescription`s with the same names as their counterparts in this module. Should I handle this in some other way?

libs/types-common/src/Data/Code.hs
mdimjasevic: Done in 5413dd1.

libs/types-common/src/Data/Code.hs
mdimjasevic: Done in 5413dd1.

PR #1772, opened 2021-09-15
Servantify Galley's DELETE /i/user endpoint

mdimjasevic:
This Servantifies the `DELETE /i/user` endpoint in Galley.

## Checklist


libs/wire-api/src/Wire/API/Routes/Public.hs
```diff
@@ -61,6 +61,8 @@ type ZConn = ZAuthServant 'ZAuthConn InternalAuthDefOpts
 
 type ZOptUser = ZAuthServant 'ZAuthUser '[Servant.Strict]
 
+type ZOptConn = ZAuthServant 'ZAuthConn '[Servant.Optional, Servant.Strict]
```
pcapriotti, line 64: Could we either add an explicit `Optional` to the type above as well, for consistency? Or alternatively remove it from this one.

pcapriotti (approved):
Looks nice and straightforward. Minor nitpick below.

libs/wire-api/src/Wire/API/Routes/Public.hs
mdimjasevic: I prefer it explicit; done in dc6ea077565ab523dad5d9713bfb5d9ef163734d.

PR #1775, opened 2021-09-16
Add polysemy to spar; promote the SAMLUser CRUD interface to an effect

isovector:
This PR changes the core `Spar`  monad from 

```haskell
newtype Spar a = Spar {fromSpar :: ReaderT Env (ExceptT SparError IO) a}
```

to

```haskell
newtype Spar r a = Spar {fromSpar :: Member (Final IO) r => ReaderT Env (ExceptT SparError (Sem r)) a}
```

--- in essence, replacing the hard-coded `IO` with polysemy. The usage of `Sem` bakes in an assumption that `IO` is available inside of the monad, so that all client code doesn't need to be retrofitted with a `Member (Final IO) r` constraint. However, because of the new `r` (effect row) type parameter, client code can *add* effects to `r`.

As mid-term goals, we should get rid of the `ReaderT Env` and `ExceptT SparError` wrappers inside of `Spar`, and replace them with finer-grained effects. As a longer term goal, we should get rid of the baked-in `Final IO` effect --- most of the benefits of polysemy come from *not* having access to IO.

For all intents and purposes, this change should be a complete no-op.

In addition to doing all of the plumbing to enable polysemy, this PR also promotes the `SAMLUser` CRUD interface into an effect (see `Spar.Sem.SAMLUser`). It comes with a single interpreter right now (`Spar.Sem.SAMLUser.Cassandra`), which runs the exact same code path. However, this allows us to add additional interpreters: in the short term, one which doesn't talk to Cassandra, but keeps users directly in memory.


isovector:
@fisx please review

services/spar/src/Spar/App.hs
```diff
@@ -1,5 +1,7 @@
 {-# LANGUAGE GeneralizedNewtypeDeriving #-}
 {-# LANGUAGE RecordWildCards #-}
+{-# OPTIONS_GHC -fno-warn-orphans #-}
```
isovector, line 3: Needed for the MonadThrow/Catch instances for Sem. Do you have a better place to put orphans?

services/spar/src/Spar/App.hs
```diff
@@ -94,8 +101,30 @@ import Wire.API.User.IdentityProvider
 import Wire.API.User.Saml
 import Wire.API.User.Scim (ValidExternalId (..))
 
-newtype Spar a = Spar {fromSpar :: ReaderT Env (ExceptT SparError IO) a}
-  deriving (Functor, Applicative, Monad, MonadIO, MonadReader Env, MonadError SparError)
+newtype Spar r a = Spar {fromSpar :: Member (Final IO) r => ReaderT Env (ExceptT SparError (Sem r)) a}
+  deriving (Functor)
```
isovector, line 105: The `Member (Final IO) r` dict inside of `Spar r a` prevents us from newtype-deriving these instances, unfortunately.

services/spar/src/Spar/App.hs
```diff
@@ -134,46 +163,59 @@ toLevel = \case
...
-wrapMonadClientWithEnv :: forall a. ReaderT Data.Env (ExceptT TTLError Cas.Client) a -> Spar a
+wrapMonadClientWithEnv :: forall r a. ReaderT Data.Env (ExceptT TTLError Cas.Client) a -> Spar r a
 wrapMonadClientWithEnv action = do
   denv <- Data.mkEnv <$> (sparCtxOpts <$> ask) <*> (fromTime <$> getNow)
   either (throwSpar . SparCassandraTTLError) pure =<< wrapMonadClient (runExceptT $ action `runReaderT` denv)
 
+instance Member (Final IO) r => Catch.MonadThrow (Sem r) where
+  throwM = embedFinal . Catch.throwM @IO
+
+instance Member (Final IO) r => Catch.MonadCatch (Sem r) where
```
isovector, line 198: These orphan instances are only used for the call to `Catch.catch` on line 212. Is it worth just inlining them there rather than defining the orphans?

services/spar/src/Spar/Scim.hs
```diff
@@ -115,32 +116,33 @@ apiScim =
...
-              { schemas = [Scim.Schema.Error20],
-                status = Scim.Status $ errHTTPCode err,
-                scimType = Nothing,
-                detail = Just . cs $ errBody err
-              }
-        Right (Right x) -> do
-          -- No exceptions! Good.
-          pure $ Right x
+    wrapScimErrors :: Spar r a -> Spar r a
+    wrapScimErrors act = Spar $
```
isovector, line 120: This change looks bigger than it is --- it's mostly whitespace. It's a result of removing the `_Spar` iso, which is too impredicative to define with the packed `Member (Final IO)` dict.

services/spar/src/Spar/Sem/SAMLUser.hs
```diff
@@ -0,0 +1,17 @@
...
+
+data SAMLUser m a where
+  Insert :: SAML.UserRef -> UserId -> SAMLUser m ()
+  Get :: SAML.UserRef -> SAMLUser m (Maybe UserId)
+  GetAnyByIssuer :: SAML.Issuer -> SAMLUser m (Maybe UserId)
+  GetSomeByIssuer :: SAML.Issuer -> SAMLUser m [(SAML.UserRef, UserId)]
+  DeleteByIssuer :: SAML.Issuer -> SAMLUser m ()
+  Delete :: UserId -> SAML.UserRef -> SAMLUser m ()
+
+-- TODO(sandy): Inline this definition --- no TH
```
isovector, line 16: Happy to do this now, but otherwise I'll wait and do all of the to-be-defined effects' TH at the same time.

fisx: Let's make progress with polysemization first, and tweak TH away later iff we have time?

services/spar/src/Spar/App.hs
fisx: I suppose that will chip away some of the work of making this an effect as well?  In that case: yes, pls inline; otherwise let's wait for this to be refactored away.

services/spar/src/Spar/App.hs
fisx: Not really, but they are a temporary thing, right?

fisx (changes requested):
Looks great!  A couple minor requests:

- Could you branch off `develop` in the future, and resolve the conflicts for this PR now before I go through the changes one more time?
- spar.cabal has been regenerated and needs to be committed.

isovector:
Rebased off of `develop`, sorry about the mix-up there. I'll work on regenerating the cabal files. PTAL

PR #1781, opened 2021-09-17
 Add polysemy to spar; promote the SAMLUser CRUD interface to an effect

fisx:
Cloned from #1775 (forks are not picked up by our CI); see description there.

## Checklist

### DIFF
--- a/services/spar/src/Spar/App.hs
+++ b/services/spar/src/Spar/App.hs
@@ -1,5 +1,7 @@
 {-# LANGUAGE GeneralizedNewtypeDeriving #-}
 {-# LANGUAGE RecordWildCards #-}
+{-# OPTIONS_GHC -fno-warn-orphans #-}
+{-# OPTIONS_GHC -fplugin=Polysemy.Plugin #-}
 
 -- This file is part of the Wire Server implementation.
 --
@@ -56,6 +58,8 @@ import qualified Data.UUID.V4 as UUID
 import Imports hiding (log)
 import qualified Network.HTTP.Types.Status as Http
 import qualified Network.Wai.Utilities.Error as Wai
+import Polysemy
+import Polysemy.Final
 import SAML2.Util (renderURI)
 import SAML2.WebSSO
   ( Assertion (..),
@@ -80,11 +84,14 @@ import qualified SAML2.WebSSO as SAML
 import qualified SAML2.WebSSO.Types.Email as SAMLEmail
 import Servant
 import qualified Servant.Multipart as Multipart
-import qualified Spar.Data as Data
+import qualified Spar.Data as Data hiding (deleteSAMLUser, deleteSAMLUsersByIssuer, getSAMLAnyUserByIssuer, getSAMLSomeUsersByIssuer, getSAMLUser, insertSAMLUser)
 import Spar.Error
 import qualified Spar.Intra.Brig as Intra
 import qualified Spar.Intra.Galley as Intra
 import Spar.Orphans ()
+import Spar.Sem.SAMLUser (SAMLUser)
+import qualified Spar.Sem.SAMLUser as SAMLUser
+import Spar.Sem.SAMLUser.Cassandra (interpretClientToIO, samlUserToCassandra)
 import qualified System.Logger as Log
 import System.Logger.Class (MonadLogger (log))
 import URI.ByteString as URI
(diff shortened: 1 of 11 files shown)


services/spar/src/Spar/App.hs
```diff
@@ -166,22 +195,35 @@ instance SPStoreIdP SparError Spar where
...
 
+instance Member (Final IO) r => Catch.MonadThrow (Sem r) where
+  throwM = embedFinal . Catch.throwM @IO
+
+instance Member (Final IO) r => Catch.MonadCatch (Sem r) where
+  catch m handler = withStrategicToFinal @IO $ do
+    m' <- runS m
+    st <- getInitialStateS
+    handler' <- bindS handler
+    pure $ m' `Catch.catch` \e -> handler' $ e <$ st
```
mdimjasevic, line 211: I'm not familiar with strategies, so I don't get the combination of `runS`, `getInitialStateS` and `bindS`. @fisx , do you understand this part?

mdimjasevic (approved):
Overall, this looks great!

services/spar/src/Spar/App.hs
fisx: I just found this: https://reasonablypolymorphic.com/blog/tactics/

I don't think it's very important, though: this is something very unlikely to get any bugs past the type system *and* the tests, and the instance will go away as we make more progress with refactoring.  Then again I suspect understanding tactical and strategy may help us understand later versions of this as well.  :)

services/spar/src/Spar/App.hs
```diff
@@ -1,5 +1,7 @@
 {-# LANGUAGE GeneralizedNewtypeDeriving #-}
 {-# LANGUAGE RecordWildCards #-}
+{-# OPTIONS_GHC -fno-warn-orphans #-}
+{-# OPTIONS_GHC -fplugin=Polysemy.Plugin #-}
```
fisx, line 4: We probably want to add this to the cabal file for the entire project?

services/spar/src/Spar/App.hs
```diff
@@ -95,8 +102,30 @@ import Wire.API.User.IdentityProvider
 import Wire.API.User.Saml
 import Wire.API.User.Scim (ValidExternalId (..))
 
-newtype Spar a = Spar {fromSpar :: ReaderT Env (ExceptT SparError IO) a}
-  deriving (Functor, Applicative, Monad, MonadIO, MonadReader Env, MonadError SparError)
+newtype Spar r a = Spar {fromSpar :: Member (Final IO) r => ReaderT Env (ExceptT SparError (Sem r)) a}
+  deriving (Functor)
+
+liftSem :: (Member (Final IO) r => Sem r a) -> Spar r a
+liftSem r = Spar $ lift $ lift r
```
fisx, line 109:
```suggestion
liftSem sem = Spar $ lift $ lift sem
```

or, even more wire style:

```suggestion
liftSem = Spar . lift . lift
```

(but it's not important)

services/spar/src/Spar/App.hs
```diff
@@ -166,22 +195,35 @@ instance SPStoreIdP SparError Spar where
...
   Spar $ do
     ctx <- asks sparCtxCas
-    runClient ctx action
+    lift (lift $ embedFinal @IO $ runClient ctx action)
       `Catch.catch` (throwSpar . SparCassandraError . cs . show @SomeException)
 
-insertUser :: SAML.UserRef -> UserId -> Spar ()
-insertUser uref uid = wrapMonadClient $ Data.insertSAMLUser uref uid
+wrapMonadClientSem :: Sem r a -> Spar r a
+wrapMonadClientSem action = liftSem $ action
```
fisx, line 223: Why the alias, and not just use `liftSem`?

services/spar/src/Spar/App.hs
isovector: These are low-level poslysemy constructs. Inside `withStrategicToFinal`, `runS` spiritually turns `Sem r a` into `IO a`, by running all *future* interpreters on it. Bind is the same, but for kleisli arrows.

The `getInitialStateS` lets us look at all of the accumulated functorial state up to this point in the program, and manipulate it as we please. For example, if there is a `State s` effect still to be interpreteted, we need to keep track of what state it has!

But @fisx is correct in that we'll very clearly see problems in CI if they are here.

isovector (approved):

services/spar/src/Spar/Sem/SAMLUser/Cassandra.hs
```diff
@@ -0,0 +1,28 @@
...
+    embed . \case
+      Insert ur uid -> Data.insertSAMLUser ur uid
+      Get ur -> Data.getSAMLUser ur
+      GetAnyByIssuer is -> Data.getSAMLAnyUserByIssuer is
+      GetSomeByIssuer is -> Data.getSAMLSomeUsersByIssuer is
+      DeleteByIssuer is -> Data.deleteSAMLUsersByIssuer is
+      Delete uid ur -> Data.deleteSAMLUser uid ur
+
+interpretClientToIO :: Member (Final IO) r => ClientState -> Sem (Embed Client ': r) a -> Sem r a
+interpretClientToIO ctx = interpret $ \case
```
fisx, line 27: Naming: should both be have `interpret` in their name?  Or neither?  Or should we just export the composite?

fisx (commented):
In summary I'm also happy with this (and thanks @mdimjasevic for taking a look!).

fisx:
I'll merge when CI is done.  (It's obvious the failure was not related to this PR, and integration tests passed locally on my machine, but more green's always nice.)

fisx:
be1ec48  associativity is hard, it seems...

services/spar/src/Spar/App.hs
isovector: Nice catch --- we unintentionally changed the logic here! This should still be catching the exceptions like `wrapMonadClient` does.

PR #1786, opened 2021-09-17
Make sure to actually wrap the action in 'wrapMonadClientSem'

isovector:
Looking through #1781, I noticed we unintentionally changed the logic with respect to how exceptions are handled. We forgot to do any wrapping in `wrapMonadClientSem`!

## Checklist


fisx (approved):
I think this is just translating one internal server error into another one, but the idea was to keep the behavior completely indistinguishable, so, good catch!

fisx (approved):

isovector:
@fisx good to merge, but apparently I don't have the right bits to do it myself.

fisx:
all tests passed locally.

jschaul:
@isovector @fisx please stop merging PRs until CI is green; #1781 introduced a new system-level dependency which resulted in breaking all the CI tests for everyone when it was merged:

```
docker run -it --entrypoint sh quay.io/wire/spar-schema:latest
 # ldd /usr/bin/spar-schema
	/lib/ld-musl-x86_64.so.1 (0x79110c8fc000)
Error loading shared library libncursesw.so.6: No such file or directory (needed by /usr/bin/spar-schema)
	libxml2.so.2 => /usr/lib/libxml2.so.2 (0x79110c7d3000)
	libstdc++.so.6 => /usr/lib/libstdc++.so.6 (0x79110c63a000)
	libz.so.1 => /lib/libz.so.1 (0x79110c620000)
	libssl.so.1.1 => /lib/libssl.so.1.1 (0x79110c59f000)
	libcrypto.so.1.1 => /lib/libcrypto.so.1.1 (0x79110c320000)
	libgmp.so.10 => /usr/lib/libgmp.so.10 (0x79110c2b9000)
	libc.musl-x86_64.so.1 => /lib/ld-musl-x86_64.so.1 (0x79110c8fc000)
	libgcc_s.so.1 => /usr/lib/libgcc_s.so.1 (0x79110c2a5000)
	liblzma.so.5 => /usr/lib/liblzma.so.5 (0x79110c28
```

> all tests passed locally.

That's not entirely correct, ./deploy/dockerephemeral/run.sh fails with the same issue.

pcapriotti:
> That's not entirely correct, ./deploy/dockerephemeral/run.sh fails with the same issue.

To be fair, if you have the library installed locally and you were already running the ephemeral services in the background, you wouldn't notice the issue.

But I agree, we shouldn't merge when CI is red.

fisx:
Sorry, will look into it now.

I my defense (@isovector had nothing to do with this): I saw and error that I didn't understand, but that did look like something related to concourse or docker trouble, which *has* happened in the past.  When I did a rerun and tried to look at the logs again, concourse wouldn't show them (page wouldn't load).  That's when I decided to take the risk.

PR #1790, opened 2021-09-20
[Federation] When Deleting a User, Remove Them From Remote Conversations

mdimjasevic:
This makes sure to remove a user from their remote conversation when the user's account is being deleted.

## Checklist

### DIFF
--- a/services/galley/src/Galley/API/Internal.hs
+++ b/services/galley/src/Galley/API/Internal.hs
@@ -29,9 +29,11 @@ import qualified Cassandra as Cql
 import Control.Exception.Safe (catchAny)
 import Control.Lens hiding ((.=))
 import Control.Monad.Catch (MonadCatch)
+import Data.Data (Proxy (Proxy))
+import Data.Domain (Domain)
 import Data.Id as Id
 import Data.List1 (maybeList1)
-import Data.Qualified (Qualified (Qualified))
+import Data.Qualified (Local, Qualified (..), Remote, partitionRemoteOrLocalIds', toLocal)
 import Data.Range
 import Data.String.Conversions (cs)
 import GHC.TypeLits (AppendSymbol)
@@ -72,6 +74,7 @@ import Servant.API.Generic
 import Servant.Server
 import Servant.Server.Generic (genericServerT)
 import System.Logger.Class hiding (Path, name)
+import Wire.API.Conversation (ConvIdsPage (..), GetPaginatedConversationIds (..))
 import Wire.API.ErrorDescription (MissingLegalholdConsent)
 import Wire.API.Routes.MultiVerb (MultiVerb, RespondEmpty)
 import Wire.API.Routes.Public (ZOptConn, ZUser)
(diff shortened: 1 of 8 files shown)


services/galley/src/Galley/API/Internal.hs
```diff
@@ -438,29 +440,49 @@ sitemap = do
...
-      cc <- Data.conversations (Cql.result ids)
+      cc <- Data.localConversations ids
       pp <- for cc $ \c -> case Data.convType c of
         SelfConv -> return Nothing
         One2OneConv -> Data.removeMember user (Data.convId c) >> return Nothing
         ConnectConv -> Data.removeMember user (Data.convId c) >> return Nothing
         RegularConv
+          -- TODO(md): see if this membership checking makes any sense here.
+          -- Perhaps we can simply call 'Update.removeMember' and let it handle
+          -- everything.
```
pcapriotti: Why would that be better? `removeMember` would query the whole conversation all over again, whereas here we already have a conversation value.

services/galley/src/Galley/API/Internal.hs
```diff
@@ -438,29 +440,50 @@ sitemap = do
 
 rmUser :: UserId -> Maybe ConnId -> Galley ()
 rmUser user conn = do
-  let n = unsafeRange 100 :: Range 1 100 Int32
-  tids <- Data.teamIdsForPagination user Nothing (rcast n)
+  let n = toRange (Proxy @100) :: Range 1 100 Int32
+      nRange1000 = rcast n :: Range 1 1000 Int32
+  tids <- Data.teamIdsForPagination user Nothing n
```
pcapriotti, line 445: Don't you need to loop through the pages here too?

pcapriotti (approved):
Looks good. I left a question, because I'm not sure about how the part about removing the user from teams works. Feel free to merge if there is no problem there.

services/galley/src/Galley/API/Internal.hs
mdimjasevic: To me this looks like duplicating the logic from `removeMember`, and it feels like checks are performed only partially. But I could be wrong.

services/galley/src/Galley/API/Internal.hs
mdimjasevic: I left the team deletion code the same (modulo replacing `rcast n` with `n`, where `n` is slightly differently defined, but should have the same value.

Perhaps it would be worth paging through teams as well. My hunch is that whoever wrote this code, assumed a user wouldn't be in more than 100 teams, hence there shouldn't be the next page to process.

services/galley/src/Galley/API/Internal.hs
mdimjasevic: I'll drop this to-do item and leave the code as is; this is slightly out of scope of this PR.

services/galley/src/Galley/API/Internal.hs
```diff
@@ -438,29 +441,53 @@ sitemap = do
...
+      leaveRemoteConversations lusr remoteConvs
+      when (mtpHasMore page) $ do
+        nextState <- case (mtpsState . mtpPagingState) page of
+          Nothing -> pure Nothing
+          Just bs -> case parseConversationPagingState bs of
+            Left _ -> throwErrorDescriptionType @BadPagingState
+            Right p -> pure . Just $ p
+        let usr = lUnqualified lusr
+            nextQuery = GetPaginatedConversationIds nextState range
+        newCids <- Query.conversationIdsPageFrom usr nextQuery
```
mdimjasevic, line 467: @pcapriotti or @jschaul , can you take a look at this? I changed this after getting a PR approval, because the underlying paging things changed on `develop` in the meantime. I am not sure if I'm doing the paging state threading properly, let alone if there's a nicer way to do it.

PR #1787, opened 2021-09-17
Polysemy: Extract IdP effect from Spar

isovector:
Part of the effort to polysemize `spar`.

This PR makes a new `IdP` effect, corresponding to more direct-Cassandra-access functions in `Spar.Data`. In addition, it pulls some functions out of `Spar.Data` which are only combinators on top of the effects, moving them to `Spar.App`.

To be merged after #1786 lands (and then rebased onto `develop`).

## Checklist

 - [x] The **PR Title** explains the impact of the change.
 - [x] The **PR description** provides context as to why the change should occur and what the code contributes to that effect. This could also be a link to a JIRA ticket or a Github issue, if there is one.
 - [x] **changelog.d** contains the following bits of information:
   - [x] A file with the changelog entry in one or more suitable sub-sections. The sub-sections are marked by directories inside `changelog.d`.

PR #1793, opened 2021-09-20
Pure interpreter for IdP effect

fisx:
New polysemy interpreter for the effect introduced in #1787.

I suppose this would be even more useful if we had a `quickcheck-state-machine`-based "proof" that the two interpreters are equivalent, but that can be done in a separate PR.

## Checklist

### DIFF
--- a/services/spar/src/Spar/Sem/IdP/Mem.hs
+++ b/services/spar/src/Spar/Sem/IdP/Mem.hs
@@ -1,0 +1,124 @@
+{-# OPTIONS_GHC -fplugin=Polysemy.Plugin #-}
+
+module Spar.Sem.IdP.Mem (idPToMem) where
+
+import Control.Exception (assert)
+import Control.Lens ((%~), (.~), (^.), _1, _2)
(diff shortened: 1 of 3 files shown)


services/spar/src/Spar/Sem/IdP/Mem.hs
```diff
@@ -0,0 +1,124 @@
...
+  forall r a.
+  Sem (Eff.IdP ': r) a ->
+  Sem r a
+idPToMem = evState . evEff
+  where
+    evState :: Sem (State IS : r) a -> Sem r a
+    evState = evalState mempty
+
+    evEff :: Sem (Eff.IdP ': r) a -> Sem (State IS ': r) a
+    evEff = reinterpret @_ @(State IS) $ \case
```
isovector, line 32: Type application isn't necessary with the signature.

```suggestion
    evEff = reinterpret $ \case
```

isovector (approved):
The polysemy looks good. I'll take your word for it that the logic is right!

PR #1794, opened 2021-09-21
Remove explicit instantiations of ErrorDescription values

pcapriotti:
So far, every `ErrorDescription` type has been accompanied by a corresponding value created with `mkErrorDescription`. This PR removes all those definitions, and adds some convenience functions (to be used with type application) that make it easier to refer to the error without having an explicit typed instantiation available.

## Checklist

### DIFF
--- a/services/brig/src/Brig/API/Error.hs
+++ b/services/brig/src/Brig/API/Error.hs
@@ -107,7 +107,7 @@ instance ToJSON Error where
 
 connError :: ConnectionError -> Error
 connError TooManyConnections {} = StdError (errorDescriptionTypeToWai @ConnectionLimitReached)
-connError InvalidTransition {} = StdError (errorDescriptionToWai invalidTransition)
+connError InvalidTransition {} = StdError (errorDescriptionTypeToWai @InvalidTransition)
 connError NotConnected {} = StdError (errorDescriptionTypeToWai @NotConnected)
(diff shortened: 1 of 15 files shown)


services/brig/src/Brig/API/Error.hs
```diff
@@ -37,21 +37,31 @@ import Network.HTTP.Types.Header
...
 import Wire.API.Federation.Error
 
 errorDescriptionToWai ::
   forall (code :: Nat) (lbl :: Symbol) (desc :: Symbol).
   (KnownStatus code, KnownSymbol lbl) =>
-  ErrDesc.ErrorDescription code lbl desc ->
+  ErrorDescription code lbl desc ->
   Wai.Error
-errorDescriptionToWai (ErrDesc.ErrorDescription msg) =
+errorDescriptionToWai (ErrorDescription msg) =
```
fisx, line 49: should this be hidden from export?  we don't want to use it any more outside of this module, right?

fisx (approved):
nice!

services/brig/src/Brig/API/Error.hs
pcapriotti: Some errors are parametrised, and in that case we need to use this function to convert them to `Wai.Error`, because `mkErrorDescription` is not sufficient.

PR #1776, opened 2021-09-16
Work around imprecisions in ES result ordering

pcapriotti:
This sorts the result list after the fact by putting exact handle and name searches at the beginning, so we don't have to rely on ES being precise with the ordering of exact matches.

Hopefully this fixes the flakyness of the `order-name` integration test, and makes #1774 unnecessary.

## Checklist

 - [x] The **PR Title** explains the impact of the change.
 - [x] The **PR description** provides context as to why the change should occur and what the code contributes to that effect. This could also be a link to a JIRA ticket or a Github issue, if there is one.
 - [x] **changelog.d** contains the following bits of information:
   - [x] A file with the changelog entry in one or more suitable sub-sections. The sub-sections are marked by directories inside `changelog.d`.

PR #1774, opened 2021-09-15
Add some retry delay logic to Search integration tests to hopefully deflake them

jschaul:
Another flaky instance of this test was seen on the CI run from #1770, quoted [here](https://github.com/wireapp/wire-server/pull/1770#issuecomment-920321326)

Perhaps using retry over a maximum of 4 seconds alleviates this problem somewhat. We care about eventual consistency I suppose, not immediate consistency.

## Checklist

 - [x] The **PR Title** explains the impact of the change.
 - [x] The **PR description** provides context as to why the change should occur and what the code contributes to that effect. This could also be a link to a JIRA ticket or a Github issue, if there is one.
 - [x] **changelog.d** contains the following bits of information:
   - [x] A file with the changelog entry in one or more suitable sub-sections. The sub-sections are marked by directories inside `changelog.d`

PR #1798, opened 2021-09-21
Remove flaky search order test

pcapriotti:
This test is flaky for reasons that are probably out of our control. Since what is testing is not a particularly essential piece of functionality, I think we are better off removing it altogether, at least until a good solution is found.

See also https://github.com/wireapp/wire-server/pull/1774 and https://github.com/wireapp/wire-server/pull/1776.

## Checklist

### DIFF
--- a/changelog.d/5-internal/remove-flaky-test
+++ b/changelog.d/5-internal/remove-flaky-test
@@ -1,0 +1,1 @@
+Remove one flaky integration test about ordering of search results
--- a/services/brig/test/integration/API/Search.hs
+++ b/services/brig/test/integration/API/Search.hs
@@ -72,7 +72,6 @@ tests opts mgr galley brig = do
         test mgr "reindex" $ testReindex brig,
         testWithBothIndices opts mgr "no match" $ testSearchNoMatch brig,
         testWithBothIndices opts mgr "no extra results" $ testSearchNoExtraResults brig,
-        testWithBothIndices opts mgr "order-name (prefix match)" $ testOrderName brig,
         testWithBothIndices opts mgr "order-handle (prefix match)" $ testOrderHandle brig,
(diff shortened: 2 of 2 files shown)


jschaul (approved):
This is fine for the time being.

One more thought: Maybe ES is to blame; but maybe also we're dealing with a race condition here related to cassandra. When creating a user then assigning them a handle in a second step, there are three inserts made to cassandra: one to create a not-activated user, one to activate them, one to set the handle. All write to the same table as upserts. The handle request will trigger a `onUserUpdate` / `updateSearchIndex`. There, the `writetime` in `lookupForIndex` of the cassandra fields are used to determine whether or not anything in the index of ES should change or not.
I wouldn't exclude the possibility of these writetimes being too close to each other for some updates to not make it correctly into ES in some cases. But maybe this is an ES-only problem also and has nothing to do with the cassandra writetimes.
Anyhow; can be debugged at a later point in time, and this is good to merge.

pcapriotti:
Thanks @jschaul. I'll merge, then.

akshaymankar:
>  I wouldn't exclude the possibility of these writetimes being too close to each other for some updates to not make it correctly into ES in some cases. But maybe this is an ES-only problem also and has nothing to do with the cassandra writetimes.

In my previous attempts, I never tried this. @jschaul Can you please create a ticket for this, so we don't forget?

jschaul:
> > I wouldn't exclude the possibility of these writetimes being too close to each other for some updates to not make it correctly into ES in some cases. But maybe this is an ES-only problem also and has nothing to do with the cassandra writetimes.
> 
> In my previous attempts, I never tried this. @jschaul Can you please create a ticket for this, so we don't forget?

Done, here: https://wearezeta.atlassian.net/browse/BE-523

PR #1802, opened 2021-09-22
Make sure that unnamed schemas are referenced

pcapriotti:
This changes the Swagger implementation of `unnamed` in schema-profunctor to drop the name, but still declare the corresponding named schema. Now even unnamed schemas that are not actually referenced anywhere will appear in the list of schemas in Swagger.

Also changed the name of the schema for `SimpleMember` to make it reflect the Haskell type name.

shot

Thanks @atomrc for reporting this issue.

## Checklist

### DIFF
--- a/changelog.d/4-docs/reference-schemas
+++ b/changelog.d/4-docs/reference-schemas
@@ -1,0 +1,1 @@
+All named Swagger schemas are now displayed in the Swagger UI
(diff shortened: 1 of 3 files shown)


jschaul (approved):
Looks good. I didn't try this out, but does this change still pass the online swagger validation tool when unreferenced things are present?

pcapriotti:
It does not, but nor did before. I'll fix it in another PR, if that's ok. We should also think about incorporating this check in the tests, because it's hard to remember to do it, and many possible things affect Swagger.

jschaul:
> It does not, but nor did before. I'll fix it in another PR, if that's ok. We should also think about incorporating this check in the tests, because it's hard to remember to do it, and many possible things affect Swagger.

Sure, that's fine. Yes about adding tests. Can we download a validation library and make it part of the tooling? (I wouldn't want to rely on the uptime of online testing websites in integration tests).

PR #1803, opened 2021-09-22
Split `MultiTablePaging` module and add haddocks

pcapriotti:
Splitting the state encoding functions to a separate module makes it impossible to use them by mistake with the wrong bytestring (like the Cassandra paging token).

## Checklist

### DIFF
--- a/libs/wire-api/src/Wire/API/Routes/MultiTablePaging.hs
+++ b/libs/wire-api/src/Wire/API/Routes/MultiTablePaging.hs
@@ -39,7 +39,7 @@ import Wire.API.Routes.MultiTablePaging.State
 --
 --  * @name@   Name of the resources being paginated through
 --  * @tables@ A (usually finite) type that represent the table currently being
---             used (must be an instance of 'PagingTable'
+--             used (must be an instance of 'PagingTable')
 --  * @max@    Maximum page size
 --  * @def@    Default page size
 --
@@ -141,7 +141,7 @@ instance
         <*> mtpPagingState .= field "paging_state" schema
 
 -- | A type to be used as the @tables@ argument of 'GetMultiTablePageRequest'
--- when the resources being paginate through are split into local and remote.
+-- when the resources being paginated through are split into local and remote.
 data LocalOrRemoteTable
   = PagingLocals
   | PagingRemotes
--- a/libs/wire-api/src/Wire/API/Routes/MultiTablePaging.hs
+++ b/libs/wire-api/src/Wire/API/Routes/MultiTablePaging.hs
@@ -1,10 +1,30 @@
-module Wire.API.Routes.MultiTablePaging where
+-- This file is part of the Wire Server implementation.
+--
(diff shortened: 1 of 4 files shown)


libs/wire-api/src/Wire/API/Routes/MultiTablePaging.hs
```diff
@@ -13,7 +33,17 @@ import qualified Data.Swagger as S
...
 import GHC.TypeLits
 import Imports
-
+import Wire.API.Routes.MultiTablePaging.State
+
+-- | A request for a page of results from the database. Type arguments:
+--
+--  * @name@   Name of the resources being paginated through
+--  * @tables@ A (usually finite) type that represent the table currently being
+--             used (must be an instance of 'PagingTable'
```
mdimjasevic, line 42:
```suggestion
--             used (must be an instance of 'PagingTable')
```

libs/wire-api/src/Wire/API/Routes/MultiTablePaging/State.hs
```diff
@@ -0,0 +1,73 @@
...
+encodePagingState (MultiTablePagingState table state) =
+  let encodedTable = encodePagingTable table
+      encodedState = fromMaybe "" state
+   in BS.cons encodedTable encodedState
+
+parseConversationPagingState :: PagingTable tables => ByteString -> Either String (MultiTablePagingState name tables)
+parseConversationPagingState = AB.parseOnly conversationPagingStateParser
+
+conversationPagingStateParser :: PagingTable tables => AB.Parser (MultiTablePagingState name tables)
+conversationPagingStateParser = do
```
mdimjasevic, line 55: I know these were just moved to a separate module, but is there anything specific about conversations so that these two functions should have the word 'conversation' in their name? To me they look more general.

mdimjasevic (approved):

libs/wire-api/src/Wire/API/Routes/MultiTablePaging.hs
```diff
@@ -139,6 +140,8 @@ instance
         <*> mtpHasMore .= field "has_more" schema
         <*> mtpPagingState .= field "paging_state" schema
 
+-- | A type to be used as the @tables@ argument of 'GetMultiTablePageRequest'
+-- when the resources being paginate through are split into local and remote.
```
jschaul, line 144:
```suggestion
-- when the resources being paginated through are split into local and remote.
```

jschaul (approved):

libs/wire-api/src/Wire/API/Routes/MultiTablePaging/State.hs
pcapriotti: You're right, this is probably a leftover name from before it was abstracted.

pcapriotti: Done in 6155f0f.

PR #1805, opened 2021-09-23
Use helmfile's parallelism to speed up integration test setup time

jschaul:
## Checklist


hack/bin/integration-setup-federation.sh
```diff
@@ -1,19 +1,56 @@
...
+# Sadly, even with helmfile, we still need to use use this recursive update
+# script beforehand on all relevant charts to download the nested dependencies
+# (e.g. cassandra from underneath databases-ephemeral)
+echo "updating recursive dependencies ..."
+charts=(fake-aws databases-ephemeral wire-server nginx-ingress-controller nginx-ingress-services)
+for chart in "${charts[@]}"; do
+    "$DIR/update.sh" "$CHARTS_DIR/$chart"
+done
 
 set -e
```
akshaymankar, line 24: This should be on top, also while we're at it, we should make it `set -euo pipefail`

akshaymankar (approved):

hack/bin/integration-setup-federation.sh
akshaymankar: `man bash` says `-o errexit` is same as `-e`. I am curious why it is frowned upon.

PR #1792, opened 2021-09-20
Polysemy: Separate more Cassandra effects from Spar

isovector:
This PR rips out a bunch more effects from Spar for polysemization and eventual unit-testing support: 

* `ScimExternalIdStore`
* `ScimUserTimesStore`
* `ScimTokenStore`
* `DefaultSsoCode`

Also, it renames `SAMLUser` to `SAMLUserStore`.

To be rebased onto `develop` after #1787 lands.

## Checklist

 - [x] The **PR Title** explains the impact of the change.
 - [x] The **PR description** provides context as to why the change should occur and what the code contributes to that effect. This could also be a link to a JIRA ticket or a Github issue, if there is one.
 - [x] **changelog.d** contains the following bits of information:
   - [x] A file with the changelog entry in one or more suitable sub-sections. The sub-sections are marked by directories inside `changelog.d`.

PR #1806, opened 2021-09-23
Spar Polysemy: Final Cassandra effects

isovector:
This PR migrates the rest of Spar off depending directly on Casasndra.

To be rebased onto `develop` when #1792 lands.

## Checklist


services/spar/src/Spar/Sem/AssIDStore/Cassandra.hs
```diff
@@ -0,0 +1,29 @@
...
+import Wire.API.User.Saml (Opts, TTLError)
+
+assIDStoreToCassandra ::
+  forall m r a.
+  (MonadClient m, Members '[Embed m, Error TTLError, Embed IO, Reader Opts] r) =>
+  Sem (AssIDStore ': r) a ->
+  Sem r a
+assIDStoreToCassandra =
+  interpret $ \case
+    Store itla t -> do
```
fisx, line 22: No need to fix this now, but in general this is about the scope size where I start to not like one-letter identifiers.

services/spar/test-integration/Test/Spar/DataSpec.hs
```diff
@@ -221,37 +225,39 @@ spec = do
           idp2' <- runSpar $ liftSem (IdPEffect.getConfig (idp1 ^. idpId))
           liftIO $ idp2' `shouldBe` Nothing
 
+-- TODO(sandy): This function should be more polymorphic over it's polysemy
+-- constraints than using 'RealInterpretation' in full anger.
```
fisx, line 229: not sure that would be an improvement?  it's a local function, and the type is obviously flexible enough for it's use cases.

fisx (approved):
Two small things:

- Could you run `make add-license` (I keep forgetting this myself)
- It would be nice to have explicit export lists in `Spar.Sem.*`.  (Optional; we don't do this consisently, but I think at some point we agreed that we'd like to have that.  If you don't do it I might make  PR to resolve `Spar.Data` into `Spar.Sem.*`, and do it there.  I'm hoping to find some dead code that way.  :)

fisx:
Random question: when I'm blocked in review, I usually base branches on top of each other, not all of them on `develop`.  That way rebasing is usually trivial.  Any reason why you don't do that?

fisx:
> Could you run `make add-license` (I keep forgetting this myself)

I just realized that we're not the only people not doing this.  :')  Never mind, I'll do it in a separate PR.

isovector:
> Random question: when I'm blocked in review, I usually base branches on top of each other, not all of them on `develop`. That way rebasing is usually trivial. Any reason why you don't do that?

This branch _is_ based on #1792 :)

services/spar/test-integration/Test/Spar/DataSpec.hs
isovector: The point being that it doesn't need all of `RealInterpreter`, it really just needs a way to run its constraints in `IO`. Making this more polymorphic would mean we could use the same function to test pure interpreters too.

isovector:
> * It would be nice to have explicit export lists in `Spar.Sem.*`

Mind if we do this in a separate PR?

PR #1810, opened 2021-09-24
Spar Polysemy: Separate out Brig and Galley effects

isovector:
This PR pulls apart the `Spar.Intra.(Brig|Galley)` modules into polysemy effects, as part of my ongoing work to excise all `IO` from Spar. 

For the most part, this is an unnotable change, except for the addition of the `Spar.Intra.BrigApp` module (name welcome to bike-shedding.) I left the original talk-to-Brig code in `Spar.Intra.Brig`, and pulled out any combinators/application code into `Spar.Intra.BrigApp` in order to avoid a dependency cycle.

Interpreting the Brig and Galley effects is rather involved, due to all of this `MonadSparTo*` business in their implementations. I implemented a new monad that has all of the necessary instances, and then run that thing to `IO` and embed it. It's not beautiful, but it works, and we can spend some time in the future getting rid of `MonadSparTo*`, which will significantly simplify things.

## Checklist


fisx:
you forgot some pieces of the code, eg.:

```
/src/wire-server/services/spar/test/Test/Spar/Intra/BrigSpec.hs:57:7: error:
Variable not in scope:
veidToUserSSOId :: ValidExternalId -> UserSSOId
```

isovector:
Sorry about that. Should be fixed now.

services/spar/src/Spar/Sem/GalleyAccess/Http.hs
```diff
@@ -0,0 +1,68 @@
...
+
+instance MonadSparToGalley RunHttp where
+  call modreq = do
+    req <- asks rheRequest
+    httpLbs req modreq
+
+instance MonadSparToBrig RunHttp where
+  call modreq = do
+    req <- asks rheRequest
+    httpLbs req modreq
```
fisx, line 53: This is a bit weird: the two instances only do what they should because they are carrying the correct resp. hostnames in the `RunHttpEnv` i the reader.  Nothing guarantees that's the case.  Right?  Time to kill these classes, I think...

(I give you the original code doesn't look much better.  Polysemy's an improvement!  :))

fisx (approved):
I tried to move all ocde from Spar.Intra.Brig into Spar.Sem.BrigAccess.Http, which *almost* worked fairly well, but I ran into a wall with cyclical deps between galley and brig.

Then I tried to eliminate the `MonadSparTo*` type classes, and got very lost and decided this should all be done in separate PRs.  :)

So, this PR's progress.  All good!

services/spar/src/Spar/Intra/Brig.hs
```diff
@@ -94,59 +71,6 @@ veidToUserSSOId = runValidExternalId urefToUserSSOId (UserScimExternalId . fromE
 urefToUserSSOId :: SAML.UserRef -> UserSSOId
 urefToUserSSOId (SAML.UserRef t s) = UserSSOId (cs $ SAML.encodeElem t) (cs $ SAML.encodeElem s)
```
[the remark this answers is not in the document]
isovector: Oops, nice catch!

services/spar/src/Spar/Sem/GalleyAccess/Http.hs
isovector: Yeah, things are currently a bit janky here. I'm trying to change the implementation code as little as possible in order to get the polysemization done ASAP. Getting everything under the polysemy umbrella is the next thing on the critical path here; once it's out of the way, we can do cleanup and testing in parallel.

PR #1815, opened 2021-09-27
Spar Polysemy: Random effect

isovector:
This PR minimizes the `MonadIO` footprint still in Spar, by creating an effect that can generate random things.

## Checklist

 - [x] The **PR description** provides context as to why the change should occur and what the code contributes to that effect. 

### DIFF
--- a/services/spar/src/Spar/Sem/Random.hs
+++ b/services/spar/src/Spar/Sem/Random.hs
@@ -1,9 +1,9 @@
 module Spar.Sem.Random where
 
-import Imports
-import Polysemy
 import Data.Id (ScimTokenId)
 import Data.UUID (UUID)
+import Imports
+import Polysemy
 
 data Random m a where
   Bytes :: Int -> Random m ByteString
@@ -11,4 +11,3 @@ data Random m a where
   ScimTokenId :: Random m ScimTokenId
 
 makeSem ''Random
-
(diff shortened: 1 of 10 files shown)


services/spar/src/Spar/Sem/Random.hs
```diff
@@ -0,0 +1,14 @@
+module Spar.Sem.Random where
```
fisx, line 1: What about this one?  https://hackage.haskell.org/package/polysemy-zoo-0.7.0.1/docs/Polysemy-Random.html

I think this is fine for now, but I don't know if I like the abstraction.  I would have made a more datatype-oblivious effect and then written application logic that uses that, and implements the functions below.

The way it looks now, I have to pull in the effect for generating scim tokens when i only want bytes for something completely unrelated.

I suppose you did this because it's easy to do and revisit and understand later?  I'm happy with keeping it as is, just checking :)

fisx (approved):

services/spar/src/Spar/Sem/Random.hs
isovector: In general, I think application-level effects should be as specific as possible. There's not much value in writing a really abstract effect, just to immediately specialize it. I'd be sympathetic to the argument that maybe we should expose these as `Input UUID`, or something, but I'm not entirely sure what it would buy us.

PR #1820, opened 2021-09-29
Fix check on conversation size

pcapriotti:
Before, we were checking whether the number of people added to a conversation was strictly less than the limit - 1, which was causing conversations with maximum size to be rejected.

This was introduced by the refactoring in #1801.

## Checklist


mdimjasevic (approved):

mdimjasevic:
The test looks good.

PR #1826, opened 2021-09-30
[Federation] Include Remote Connections in Listing All Connections

mdimjasevic:
The PR extends the `POST /list-connections` endpoint to also list remote connections. As with conversations, first it paginates through local connections and only then through remote connections.

The PR depends on PR #1819.

## Checklist


services/brig/src/Brig/API/Public.hs
```diff
@@ -1134,11 +1141,23 @@ listConnections uid req = do
...
           mtpPagingState = Public.ConnectionPagingState table (LBS.toStrict . C.unPagingState <$> pwsState)
         }
+
     mkState :: ByteString -> C.PagingState
     mkState = C.PagingState . LBS.fromStrict
 
-    convertedState :: Maybe C.PagingState
-    convertedState = fmap mkState . Public.mtpsState =<< Public.gmtprState req
+    localsAndRemotes :: Local UserId -> Maybe C.PagingState -> Range 1 500 Int32 -> Handler Public.ConnectionsPage
+    localsAndRemotes self pagingState size = do
```
pcapriotti, line 1149: Is it worth trying to abstract out this logic, since it is the same for connections and conversations? In any case, let's do it in a separate PR.

services/brig/src/Brig/API/Public.hs
```diff
@@ -1134,11 +1141,23 @@ listConnections uid req = do
...
+      localPage <- pageToConnectionsPage Public.PagingLocals <$> Data.lookupLocalConnectionsPage self pagingState (rcast size)
+      let remainingSize = fromRange size - fromIntegral (length (Public.mtpResults localPage))
+      if Public.mtpHasMore localPage || remainingSize <= 0
+        then pure localPage {Public.mtpHasMore = True} -- We haven't check the remotes yet, so has_more must always be True here.
```
pcapriotti, line 1153: Typo: check → checked. Also in the original code this was pasted from.

pcapriotti (approved):
Looks good. Minor comments follow.

services/brig/src/Brig/API/Public.hs
mdimjasevic: Thanks! Fixed in d2d0507.

PR #1827, opened 2021-09-30
Spar Polysemy: SAML2 effect

isovector:
This PR adds some new effects, completely isolating us from the `saml2-web-sso` interface:

- `SAML2`
- `Now`
- `SparRoute`

It also does some refactoring around the canonical interpretation of `Spar`, moving it into its own module --- necessary to break some cyclic dependencies, and a nice touch all around.

As of this change, all the necessary parts of `saml2-web-sso` are packaged as an effect, meaning we can aggressively remove instances on `Spar`. But these instances aren't going far, they're now part of the implementation inside of `Spar.Sem.SAML2.SAML2WebSso` --- at least until we polysemize `saml2-web-sso` also.

## Checklist


services/spar/src/Spar/Sem/SAML2/SAML2WebSso.hs
```diff
@@ -0,0 +1,136 @@
...
+       Final IO
+     ]
+    r =>
+  Inspector f ->
+  Sem (SAML2 : r) (f b) ->
+  Blah r b
+inspectOrBomb ins get_a = do
+  fa <- Blah $ saml2ToSaml2WebSso get_a
+  maybe
+    (error "saml2ToSaml2WebSso called with an uninspectable weaving functor")
```
isovector, line 134: Scary! This case is impossible under normal circumstances, but can be triggered if any of the monadic arguments to the `SAML2` actions fail via `Polysemy.Error.throw`. A better solution here would be to instead throw an error of our own, but I wasn't sure which constructor to use --- due to a technical limitation in Polysemy, we can't rethrow (but the original error will arrive upstream anyway!)

What do you think is the right play here, @fisx?

isovector: It's worth noting that this is entirely an artifact of the typeclass approach to `saml2-web-sso`, and will go away when we polysemize that library.

services/spar/src/Spar/API.hs
```diff
@@ -115,9 +121,15 @@ api ::
...
        SAMLUserStore,
        Random,
+       Error SparError,
+       SAML2,
+       Now,
+       SparRoute,
        Logger String,
        Logger (Msg -> Msg),
-       Error SparError
+       Error SparError,
```
isovector, line 130: duplicate

services/spar/src/Spar/Sem/SparRoute.hs
```diff
@@ -0,0 +1,13 @@
+module Spar.Sem.SparRoute where
+
+import Data.Id (TeamId)
+import Imports
+import Polysemy
+import qualified SAML2.WebSSO.Types as SAML
+import qualified URI.ByteString as URI
+
+data SparRoute m a where
```
isovector, line 9: note to self: fisx had a better name for this

fisx: I forgot, but today I shall call this `SamlProtcolSettings`.

services/spar/src/Spar/Sem/SAML2/SAML2WebSso.hs
fisx: how many places are there with monadic arguments?  can we refactor those away somehow?  or would that be harder than polysemizing saml2-web-sso?

services/spar/src/Spar/Scim/User.hs
```diff
@@ -609,13 +614,13 @@ toScimStoredUser' createdAt lastChangedAt baseuri uid usr =
...
 
 updScimStoredUser ::
-  forall m.
-  (SAML.HasNow m) =>
+  forall r.
+  Member Now r =>
   Scim.User ST.SparTag ->
   Scim.StoredUser ST.SparTag ->
-  m (Scim.StoredUser ST.SparTag)
+  Spar r (Scim.StoredUser ST.SparTag)
```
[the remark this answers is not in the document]
isovector: Yeah, it is. I'm working mostly mindlessly on these, and plan to just replace `Spar` with `Sem` in a PR or two.

services/spar/src/Spar/Sem/SAML2/SAML2WebSso.hs
isovector: The issue are these two effects:

```haskell
data SAML2 m a where
  AuthReq ::
    NominalDiffTime ->
    m Issuer ->
    IdPId ->
    SAML2 m (FormRedirect AuthnRequest)
  AuthResp ::
    Maybe TeamId ->
    m Issuer ->
    m URI ->
    (AuthnResponse -> AccessVerdict -> m resp) ->
    AuthnResponseBody ->
    SAML2 m resp
```

their `m Blah` arguments, in particular.

The tactics effect in Polysemy lets us lift functions `a -> m b` into functions `f a -> m (f b)`, where `f` holds all of the accumulated state from other effects. In a world where `saml2-web-sso` were more flexible with the types of `authreq` and `authresp`, we could just push that `f` all the way through.

The behavior here is not surprising though. Imagine in everyday MTL land, where we pass `throwError blah` as the argument to `authreq`. That too would fail. The types in tactics just make it very clear (and, simultaneously, very obscure) to see where we have these sorts of data dependencies.

IMO, the solution here is to make this a `throw` instead of `error`. But I don't expect to see it, since `authreq` and `authresp` are only ever called called with `SparRoute` actions. 

Life (and this interpreter) would be much easier if these were just pure values instead of monadic ones, which we could do without polysemizing `saml2-web-sso` --- though maybe some callers depend on this being monadic?

services/spar/src/Spar/App.hs
```diff
@@ -421,69 +379,6 @@ bindUser buid userref = do
...
-     Logger String,
-     Logger (TinyLog.Msg -> TinyLog.Msg),
-     Input Opts,
-     Input TinyLog.Logger,
-     Random,
-     Embed IO,
-     Final IO
-   ]
-
-instance r ~ RealInterpretation => SPHandler SparError (Spar r) where
```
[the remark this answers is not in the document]
isovector: As best I can tell, it's only used to give the natural transformation, which we do ourselves anyway.

services/spar/src/Spar/App.hs
```diff
@@ -160,9 +137,6 @@ instance MonadError SparError (Spar r) where
 instance MonadIO (Spar r) where
   liftIO m = Spar $ lift $ embedFinal m
 
-instance Members '[Input Opts, Logger String] r => HasLogger (Spar r) where
```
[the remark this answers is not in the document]
isovector: We're getting tantalizingly close!

services/spar/src/Spar/Error.hs
```diff
@@ -100,6 +100,7 @@ data SparCustomError
   | SparIdPIssuerInUse
   | SparProvisioningMoreThanOneIdP LT
   | SparProvisioningTokenLimitReached
+  | SparInternalError LT
```
fisx, line 103: We have a catch-all that makes this a little redundant.
- https://github.com/wireapp/wire-server/blob/48f7747c84662302676b5362804c33e04e7e03c3/services/spar/src/Spar/Run.hs#L118
- https://github.com/wireapp/wire-server/blob/48f7747c84662302676b5362804c33e04e7e03c3/libs/wai-utilities/src/Network/Wai/Utilities/Server.hs#L182-L198

I think it makes sense to catch it here, but we should add context:

```suggestion
  | -- | FUTUREWORK(fisx): this could also be caught by 'catchErrors' (see "Spar.Run"), maybe we want to remove this case again?  not sure.
    SparInternalError LT
```

services/spar/src/Spar/Sem/SAML2/Library.hs
```diff
@@ -0,0 +1,137 @@
...
+       Final IO
+     ]
+    r =>
+  Inspector f ->
+  Sem (SAML2 : r) (f b) ->
+  SPImpl r b
+inspectOrBomb ins get_a = do
+  fa <- SPImpl $ saml2ToSaml2WebSso get_a
+  maybe
+    (SPImpl . throw @SparError $ SAML.CustomError $ SparInternalError "saml2ToSaml2WebSso called with an uninspectable weaving functor")
```
fisx, line 135: This can only possibly happen in the verdict handler, and I think it's fine.  Everything thrown via the error effects (mtl or polysemy) is caught before this, and everything thrown elsewhere is a legitimate internal error.

services/spar/src/Spar/Sem/SAML2/Library.hs
```diff
@@ -0,0 +1,137 @@
...
+      x <- raise $ unSPImpl $ SAML.authreq @_ @SparError n (inspectOrBomb ins get_a) i
+      s <- getInitialStateT
+      pure $ x <$ s
+    AuthResp mitlt ma mb mc ab -> do
+      get_a <- runT ma
+      get_b <- runT mb
+      get_c <- bindT $ uncurry mc
+      ins <- getInspectorT
+      s <- getInitialStateT
+
```
fisx, line 104:
```suggestion
```

services/spar/src/Spar/Sem/SAML2/Library.hs
```diff
@@ -0,0 +1,137 @@
...
+  storeIdPConfig = SPImpl . App.runSparInSem . App.storeIdPConfig
+  getIdPConfig = SPImpl . App.runSparInSem . App.getIdPConfig
+  getIdPConfigByIssuerOptionalSPId a = SPImpl . App.runSparInSem . App.getIdPConfigByIssuerOptionalSPId a
+
+instance Member (Error SparError) r => MonadError SparError (SPImpl r) where
+  throwError = SPImpl . throw
+  catchError m handler = SPImpl $ catch (unSPImpl m) $ unSPImpl . handler
+
+saml2ToSaml2WebSso ::
```
fisx, line 75: I'm going to trust you on this, but:

```suggestion
-- | To learn more about polysemy tactics, read this:
--   * https://reasonablypolymorphic.com/blog/freer-higher-order-effects/
--   * https://reasonablypolymorphic.com/blog/tactics/
saml2ToSaml2WebSso ::
```

fisx (approved):
I have some minor suggestions, you decide whether they make sense, and whether you want to act on them here or in a different PR.

Otherwise all good!

services/spar/src/Spar/API.hs
```diff
@@ -78,10 +79,15 @@ import qualified Spar.Sem.GalleyAccess as GalleyAccess
 import qualified Spar.Sem.IdP as IdPEffect
 import Spar.Sem.Logger (Logger)
 import qualified Spar.Sem.Logger as Logger
+import Spar.Sem.Now (Now)
 import Spar.Sem.Random (Random)
 import qualified Spar.Sem.Random as Random
+import Spar.Sem.SAML2 (SAML2)
+import qualified Spar.Sem.SAML2 as SAML2
```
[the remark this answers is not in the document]
isovector: Sounds good to me. OK to tackle it after we're done the polysemization though? Then I don't have to worry about conflicts :)

services/spar/src/Spar/Error.hs
isovector: Sorry, I don't understand this comment. I added this constructor because I need to throw _something_ from the `SAML2` interpreter in case the inspector doesn't work, and none of the existing error messages seemed like a good fit.

services/spar/src/Spar/Sem/SAML2/Library.hs
isovector: It's nightmarish code, sorry. Polysemy v2 promises to clean this up, and in my offtime I'm going to port this function to see how improved things are.

Happy to write some comments about WTF is going on here.

services/spar/src/Spar/API.hs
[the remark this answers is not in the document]
fisx: yes, absolutely.  should have labelled this "future work / note to self" more clearly.

services/spar/src/Spar/Error.hs
fisx: My point is that if we don't throw anything here, `catchErrors` will handle it in `Spar.Run`.  Do you disagree with that?  (Also I would assume that this is also what would happen to those errors before this change, no?)

isovector: Added a slightly changed comment

services/spar/src/Spar/Sem/SAML2/Library.hs
fisx: great idea.  if you're doing this right, perhaps you can point other people to this when they're asking about tactics?  :)

PR #1833, opened 2021-10-02
Spar Polysemy: Fully polysemize Spar

isovector:
This PR replaces `Spar`:

```haskell
newtype Spar r a = Spar { Member (Final IO) r => ExceptT (Sem r) a }
```

and instead uses `Sem` directly. In doing so, it means `Spar` is now completely polysemized :tada: :tada: 

Sorry for the size here, I got excited about being as close to the finish line as I was. That being said, the commits are well-organized, so they're a reasonable way to approach this monstrosity.

Most of this PR is boring; replacing `throwSpar` with `throwSparSem` and adding `Member (Error SparError)` constraints. Also, replacing `Spar` with `Sem`. But there are a few changes of interest:

1. Removing `wrapMonadClientSem`, which previously was used to guard all of the `Cassandra` effects. This function has been removed (it required `MonadIO`), but instead, it is now used to wrap each DB action in `interpretClientToIO`. 
2. `Spar.Scim.wrapScimErrors` really depended on having an `ExceptT` and `MonadIO` around. Instead, we now do some direct manipulation of the polysemy stack, to locally introduce an `Error SomeException` effect, and immediately interpret it. This lets us catch Haskell exceptions directly in polysemy, though it's not clear to me that these exceptions can ever happen here (see point 1). While this works, it does float a `Final IO` effect up for a few layers, but no application code has real access to it.

## Checklist

### DIFF
--- a/services/spar/src/Spar/Scim/User.hs
+++ b/services/spar/src/Spar/Scim/User.hs
@@ -447,9 +447,8 @@ createValidScimUser tokeninfo@ScimTokenInfo {stiTeam} vsu@(ST.ValidScimUser veid
           -- up.  We should consider making setUserHandle part of createUser and
           -- making it transactional.  If the user redoes the POST A new standalone
           -- user will be created.}
-          do
-            BrigAccess.setHandle buid handl
-            BrigAccess.setRichInfo buid richInfo
+          BrigAccess.setHandle buid handl
+          BrigAccess.setRichInfo buid richInfo
           pure buid
 
       -- {If we crash now,  a POST retry will fail with 409 user already exists.
@@ -576,9 +575,8 @@ updateVsuUref team uid old new = do
     (mo, mn@(Just newuref)) | mo /= mn -> validateEmailIfExists uid newuref
     _ -> pure ()
 
-  do
-    old & ST.runValidExternalId (SAMLUserStore.delete uid) (ScimExternalIdStore.delete team)
-    new & ST.runValidExternalId (`SAMLUserStore.insert` uid) (\email -> ScimExternalIdStore.insert team email uid)
+  old & ST.runValidExternalId (SAMLUserStore.delete uid) (ScimExternalIdStore.delete team)
+  new & ST.runValidExternalId (`SAMLUserStore.insert` uid) (\email -> ScimExternalIdStore.insert team email uid)
 
   BrigAccess.setVeid uid new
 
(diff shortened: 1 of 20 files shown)


services/spar/src/Spar/Scim/User.hs
```diff
@@ -431,23 +431,23 @@ createValidScimUser tokeninfo@ScimTokenInfo {stiTeam} vsu@(ST.ValidScimUser veid
...
-          liftSem $ Logger.debug ("createValidScimUser: brig says " <> show buid)
+          Logger.debug ("createValidScimUser: brig says " <> show buid)
 
           -- {If we crash now, we have an active user that cannot login. And can not
           -- be bound this will be a zombie user that needs to be manually cleaned
           -- up.  We should consider making setUserHandle part of createUser and
           -- making it transactional.  If the user redoes the POST A new standalone
           -- user will be created.}
-          liftSem $ do
+          do
```
isovector, line 450: Unnecessary block

services/spar/src/Spar/Scim/User.hs
```diff
@@ -570,18 +569,18 @@ updateVsuUref ::
...
-  Spar r ()
+  Sem r ()
 updateVsuUref team uid old new = do
   let geturef = ST.runValidExternalId Just (const Nothing)
   case (geturef old, geturef new) of
     (mo, mn@(Just newuref)) | mo /= mn -> validateEmailIfExists uid newuref
     _ -> pure ()
 
-  wrapMonadClientSem $ do
+  do
```
isovector, line 579: Unnecessary block

services/spar/src/Spar/Scim/User.hs
```diff
@@ -447,9 +447,8 @@ createValidScimUser tokeninfo@ScimTokenInfo {stiTeam} vsu@(ST.ValidScimUser veid
           -- up.  We should consider making setUserHandle part of createUser and
           -- making it transactional.  If the user redoes the POST A new standalone
           -- user will be created.}
-          do
-            BrigAccess.setHandle buid handl
-            BrigAccess.setRichInfo buid richInfo
+          BrigAccess.setHandle buid handl
+          BrigAccess.setRichInfo buid richInfo
```
fisx, line 451: I disagree with this commit: the `do`s are here intentionally to make it clear the contained lines form a self-contained thing.  But I think I'm a minority with this, so fine, keep this it.

fisx (approved):
No more IO ever again!  :-)

services/spar/src/Spar/Scim/User.hs
isovector: I'm pretty sure I introduced the `do` a few commits back to avoid two calls to `wrapMonadClientSem`.

fisx: Oops, sorry, reflex.  I've been trying to defend this pattern before.  Never mind, my bad.

PR #1834, opened 2021-10-04
Typo

fisx:
Make online help in stern/backoffice more helpful and less wrong.  I don't want to clutter the release notes with this, should I?


pcapriotti (approved):
Looks good. No need to add a CHANGELOG entry.

PR #1837, opened 2021-10-05
Prevent legalhold and federation to be enabled at the same time

smatting:
## Checklist

This PR enforces the decision to not support Legalhold and Federation at the same time.
Galley won't start if both features are configured to be enabled.

### DIFF
--- a/services/galley/src/Galley/App.hs
+++ b/services/galley/src/Galley/App.hs
@@ -82,6 +82,7 @@ import Galley.API.Error
 import qualified Galley.Aws as Aws
 import Galley.Options
 import qualified Galley.Queue as Q
+import Galley.Types.Teams (flagLegalHold)
 import qualified Galley.Types.Teams as Teams
 import Imports
 import Network.HTTP.Client (responseTimeoutMicro)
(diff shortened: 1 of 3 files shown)


services/galley/src/Galley/App.hs
```diff
@@ -182,6 +184,20 @@ validateOptions l o = do
...
     error "setMaxTeamSize cannot be < setTruncationLimit"
 
+  when
+    ( isJust (o ^. optFederator)
+        && settings ^. setFeatureFlags . flagLegalHold /= Teams.FeatureLegalHoldDisabledPermanently
+    )
+    $ do
+      let message =
+            ( "Legalhold cannot be activated together with Federation. \
+              \To resolve this problem either set the feature \"legalhold\" to \"disabled-by-default\" \
```
akshaymankar, line 194:
```suggestion
              \To resolve this problem either set the feature \"legalhold\" to \"disabled-permanently\" \
```

pcapriotti (approved):
Looks good.

hack/helm_vars/wire-server/values.yaml
```diff
@@ -146,7 +146,7 @@ galley:
       federationDomain: integration.example.com
       featureFlags:
         sso: disabled-by-default  # this needs to be the default; tests can enable it when needed.
-        legalhold: whitelist-teams-and-implicit-consent
```
smatting, line 149: We need to check if this makes it impossible to test the LH features on CI

PR #1845, opened 2021-10-08
Avoid linear lookup when checking new members

pcapriotti:
Just a small refactoring to avoid quadratic runtime building a user team membership map when adding new members to a conversation.

No CHANGELOG entry.

## Checklist


mdimjasevic (approved):
Oh, I forgot about this one! It looks to me like it'll need conflict resolving due to the Polysemy effect changes to Galley.

PR #1852, opened 2021-10-11
Upgrade to GHC 8.10 and use Ubuntu 20.04 as base image for haskell docker images

fisx:
As the only bindists for GHC 8.10.7 on Alpine Linux are for integer-simple and migrating to integer-simple is some work, this PR swaps out the Alpine docker images with Ubuntu 20.04 based docker images. As Ubuntu is a much more widely used distribution among haskellers, we hope for better support from community for this.

## Checklist

 - [x] The **PR Title** explains the impact of the change.
 - [x] The **PR description** provides context as to why the change should occur and what the code contributes to that effect. This could also be a link to a JIRA ticket or a Github issue, if there is one.
 - [x] **changelog.d** contains the following bits of information:
   - [x] A file with the changelog entry in one or more suitable sub-sections. The sub-sections are marked by directories inside `changelog.d`.

PR #1853, opened 2021-10-12
Use cabal to build wire-server (opt-in)

smatting:
This PR should enable us to try using cabal instead of stack locally.

This PR:
- adds nix files that provide the `stack2cabal` tool
- instructions how to convert the stack project to a cabal project
- Adds a `cabal.project` and `cabal.freeze` file generated by stack2cabal
- Adds a flag `WIRE_BUILD_WITH_CABAL` that is, When it is set `cabal` will be used for `make fast` and `make install` from the root Makefile.
- Adds two make targets that use `cabal` to build:
```
   make c package=brig # to build and install all of brig's executables
   make c package=brig test=1 # also run unit tests
   make ci package=brig pattern="delete" # build and run brig's integration tests
```
- Deletes `package.yaml` of `types-common-journal` and `wire-message-proto-lens`, because the cabal files generated with `hpack` don't work with cabal.

For details see `tools/convert-to-cabal/README.md` included in this PR.

Building with cabal has been tested only locally. For building on CI/withou nix follow-up PRs are needed.

## Checklist


fisx:
this will affect https://github.com/wireapp/wire-server/pull/1852, but i think we can work on it in parallel.  getting all the dependencies in shape for ghc10 and all the new packages will still be required, right?

smatting:
> this will affect #1852, but i think we can work on it in parallel. getting all the dependencies in shape for ghc10 and all the new packages will still be required, right?

Yes, I suggest we finish #1852 first and then rerun stack2cabal (included in this PR) again.

hack/bin/install-cabal-all.sh
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env bash
+# copies all executables from ./dist-newstyle to ./dist
+set -euo pipefail
+
+DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+TOP_LEVEL="$(cd "$DIR/../.." && pwd)"
+DIST="$TOP_LEVEL/dist"
+
+find_exes() {
+    find ./dist-newstyle/build -executable -type f \( -path "*\/x\/*" -or -path '*\/b\/*' -or -path '*\/t\/*' \)
```
akshaymankar, line 10: Instead of this, we should use [`cabal-plan list-bins`](https://github.com/haskell-hvr/cabal-plan). (if we cannot use `cabal install --install-dir=<DIR>`

cabal.project.freeze
```diff
@@ -0,0 +1,2528 @@
+constraints: any.AC-Angle ==1.0,
```
mdimjasevic, line 1: I am not familiar with the purpose of this file, but it looks like version pinning of all direct and indirect dependencies. How do we bump these versions? Is there a tool that does it?

cabal-deps.nix
```diff
@@ -0,0 +1,27 @@
+{ pkgs ? import ./nix }:
+
+with pkgs;
+
+mkShell {
```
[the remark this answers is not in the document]
smatting: Done

hack/bin/install-cabal-all.sh
```diff
@@ -0,0 +1,15 @@
...
+}
+
+for exe in $(find_exes); do
+    cp -v "$exe" "$DIST"
+done
```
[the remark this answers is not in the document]
smatting: It works but it is slow.

hack/bin/install-cabal-all.sh
smatting: Done

cabal.project.freeze
smatting: The freeze file has been created by the `stack2cabal` tool. We don't know yet how easy it is to upgrade without stackage yet. This PR doesn't commit it us to using cabal however.

Makefile
```diff
@@ -39,12 +39,34 @@ init:
 # Build all Haskell services and executables, run unit tests
 .PHONY: install
 install: init
+ifeq ($(WIRE_BUILD_WITH_CABAL), 1)
+	cabal build all
+	cabal test all
+	cabal install --install-method=copy --installdir=dist all
```
akshaymankar, line 45: `cabal install` builds things even after `cabal build` has built it. I am going to see if I can do something to influence that, otherwise we can maybe just use `cabal-plan list-bins` to copy all the binaries to dist.

tools/convert-to-cabal/README.md
```diff
@@ -0,0 +1,85 @@
...
+package foo
+  ghc-options: -O0
+```
+
+- With new cabal build there doesn't seem to be any way of running tests as part of a build. You have to run the tests manually.
+      https://github.com/haskell/cabal/issues/7267
+
+- Nix integration (`nix: True` in `~/.cabal/config`) is not supported in new-build.
+  https://github.com/haskell/cabal/issues/4646
+  That's why you have to enter the environment defined by `cabal-deps.nix` manually (or via direnv) to use cabal.
```
akshaymankar, line 72:
```suggestion
  That's why you have to enter the environment defined by `direnv.nix` manually (or via direnv) to use cabal.
```

Makefile
```diff
@@ -39,12 +39,34 @@ init:
...
 # Build all Haskell services and executables with -O0, run unit tests
 .PHONY: fast
 fast: init
+ifeq ($(WIRE_BUILD_WITH_CABAL), 1)
+	make install
+else
 	stack install --pedantic --test --bench --no-run-benchmarks --local-bin-path=dist --fast $(WIRE_STACK_OPTIONS)
+endif
+
+.PHONY: c
```
akshaymankar, line 59: Does c stand for cabal?

Makefile
akshaymankar: Resolved by using `cabal-plan` and a couple of tiny shell scripts.

Makefile
akshaymankar: I assumed this and wrote a comment.

akshaymankar (approved):

PR #1863, opened 2021-10-18
polysemy-check laws for IdP effect

isovector:
This PR enables `polysemy-check`, and adds a test of some laws for the `IdP` effect.

## Checklist


fisx:
I have no idea what I was thinking when I wrote `storeConfig`, but I think the filter should just go away:

```haskell
storeConfig :: IP.IdP -> TypedState -> TypedState
storeConfig iw = M.insert (iw ^. SAML.idpId) iw
``` 

At least that's what the cassandra interpreter is doing, right?  If you try that and it passes the tests, I'd say that's a good change.  :)

isovector:
> About `*ReplacedBy`: I suppose the way to handle these situations would be to add a few unit tests, until we feel that the behavior is reasonably exhaustively described? I can do that in a separate PR.

I don't even know what a unit test here would look like. I think we should add a `getReplacedBy` action :)

fisx (approved):
LGTM!  I have some ideas how to make tests more exhaustive, but that doesn't have to happen in this PR.  (Merge conflict is only in the cabal file, regenerating fixes that.)

services/spar/test/Test/Spar/Sem/IdPSpec.hs
```diff
@@ -0,0 +1,158 @@
...
+module Test.Spar.Sem.IdPSpec where
+
+import Arbitrary ()
+import Control.Lens
+import Imports
+import Polysemy
+import Polysemy.Check
+import qualified SAML2.WebSSO as SAML
+import SAML2.WebSSO.Types
+import qualified Spar.Sem.IdP as E
```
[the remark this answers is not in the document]
isovector: Effect, I guess? `IdP` is a bad name for an effect because it conflicts with the `IdP` type. Worth changing in a followup PR.

fisx:
Hum: https://github.com/wireapp/wire-server/pull/1895

```
Test.Spar.Sem.IdP
  Config Actions
    storeConfig/getConfig FAILED [1]
    storeConfig/getIdByIssuerWithoutTeam FAILED [2]
  Raw Metadata Actions
    storeRawMetadata/storeRawMetadata FAILED [3]
    storeRawMetadata/getRawMetadata FAILED [4]
    storeRawMetadata/deleteRawMetadata FAILED [5]
    deleteRawMetadata/getRawMetadata FAILED [6]
```

isovector:
CI was green as of 725b142

services/spar/test/Test/Spar/Sem/IdPSpec.hs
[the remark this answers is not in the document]
fisx: I think in the begging we tried the pattern `<datatype>Eff`, so `IdPEff`.  But yes, different PR.

fisx:
Hum, I haven't fully understood polysemy-check yet, but is it possible that there is a conceptual flaw?  It looks like arbitrary is run twice for the two traces that are then compared, and the failure is because arbitrary is, well, not deterministic.  That would explain that all tests are consistently failing.

fisx:
> CI was green as of [725b142](https://github.com/wireapp/wire-server/commit/725b142d54be29da40bb1948947e1ddd67b2878d)

That sounds like I'm doing something wrong, though.  But what?

fisx:
Can you merge my mini-PR and see if the CI picks it up again?

fisx:
> CI was green as of [725b142](https://github.com/wireapp/wire-server/commit/725b142d54be29da40bb1948947e1ddd67b2878d)

Ah, but if you click on the checkbox, you'll see that it didn't run any tests.

Can't you just `stack test --fast spar` locally?

isovector:
@fisx PTAL

services/spar/test/Test/Spar/Sem/IdPSpec.hs
```diff
@@ -0,0 +1,158 @@
+{-# OPTIONS_GHC -Wno-orphans #-}
+{-# OPTIONS_GHC -fplugin=Polysemy.Plugin #-}
+
+module Test.Spar.Sem.IdPSpec where
```
fisx, line 4: In order to run these on the cassandra interpreter, they would have to be defined under `.../src`, not `.../test`.  I propose `/services/spar/src/Spar/Sem/IdP/Spec.hs`.  It's not an interpreter, but I think ti's fine to put all interpreters and the properties that they share all in one directory.

fisx (approved):
What does "PTAL" mean?

services/spar/test/Test/Spar/Sem/IdPSpec.hs
isovector: PTAL: please take a look :)

isovector: I'll pull them out in a subsequent PR --- though we need the orphan instances from `Arbitrary ()`. OK to pull those too?

PR #1876, opened 2021-10-21
Replace shell.nix with a direnv + nixpkgs.buildEnv based setup

akshaymankar:
This ensures we have similar setup to wire-server-deploy and we do not pollute the environment with all the variables nix-shell exports.

## Checklist


smatting:
@akshaymankar could you maybe add your emacs config to e.g. `docs/develop/` on how to use the hls? For neovim maybe @jschaul ?

PR #1884, opened 2021-10-26
Cassandra: use LOCAL_QUORUM

jschaul:
As per https://wearezeta.atlassian.net/browse/BM-22 using LOCAL_QUORUM should:

* not change anything for a single-datacentre cassandra setup (which is our case atm)
* allow to make a datacentre migration that can be rolled back. 

Note: only merge this PR after the next release has been cut.

## Checklist


jschaul:
Rebased; the 'cql' and 'cql-io' modules are now only imported in one place: cassandra-util. And in that module, we only export LocalQuorum, so trying to write Quorum will lead to a compiler error everywhere.

akshaymankar (approved):
:diya_lamp:

PR #1886, opened 2021-10-26
Allow filtering cassandra nodes by datacentre

jschaul:
Context: https://wearezeta.atlassian.net/browse/BM-14 and related PR https://github.com/zinfra/cailleach/pull/703

- cassandra-util: add a dcAwarePolicy for talking to cassandra
- Spar/brig/galley/gundeck: allow filtering nodes by datacentre

Example: galley/gundeck don't configure this option, spar configures a valid datacentre, brig configures an invalid datacentre:

```
[galley] I, Using the following cassandra load balancing options ('Policy'):, filter_datacentre=Nothing
[gundeck] I, Using the following cassandra load balancing options ('Policy'):, filter_datacentre=Nothing
[spar] I, Using the following cassandra load balancing options:, filter_datacentre=Just "datacenter1"

[brig] I, Using the following cassandra load balancing options ('Policy'):, filter_datacentre=Just "datacenter2"
[brig] I, logger=cassandra.brig, Known hosts: []
[brig] I, logger=cassandra.brig, New control connection: datacenter1:rack1:127.0.0.1:9042#<socket: 11>
[brig] E, logger=cassandra.brig, No known hosts.
[brig] I, logger=cassandra.brig, Renewing control connection with initial contacts ...

(...)
brig: Failed to read schema version from meta table. Error was: cql-io: no host available
[brig] E, logger=cassandra.brig, No known hosts.
[brig] I, logger=cassandra.brig, Renewing control connection with initial contacts ...
```

So the behaviour is:

- if not configured, fallback to random policy (i.e. no change from now)
- if configured but wrongly, the service will fail to start up
- if configured correctly, only nodes in the datacentre that is configured will be used as coordinator nodes, so another datacentre going down will have no effect.

## Checklist


libs/types-common/src/Util/Options.hs
```diff
@@ -156,27 +162,6 @@ parseConfigPath defaultPath desc = do
 parseAWSEndpoint :: ReadM AWSEndpoint
 parseAWSEndpoint = readerAsk >>= maybe (error "Could not parse AWS endpoint") return . fromByteString . fromString
 
-cassandraParser :: Parser CassandraOpts
```
jschaul, line 159: This was dead code not used anywhere.

akshaymankar (approved):

PR #1891, opened 2021-10-28
Federation: Notify remote participants when a user leaves a conversation because they were deleted

smatting:
https://wearezeta.atlassian.net/browse/FS-41

## Checklist


akshaymankar (approved):
:100:

PR #1894, opened 2021-10-29
Add -Werror to package-defaults.yaml

smatting:
This PR makes all local packages build with `-Werror` by default. The motivation for this PR is the principle that building with `stack` or `cabal` without any special flags should build in production configuration.

Another motivation is that currently we invoke `stack` with `--pedantic` in all Makefiles to achieve the same thing. Cabal doesn't have a flag like this, which requires using `--ghc-options`, which has its quirks, e.g. specifying `--ghc-options` twice doesn't work as expected.

## Checklist

### DIFF
--- a/changelog.d/5-internal/werror
+++ b/changelog.d/5-internal/werror
@@ -1,0 +1,1 @@
+Add -Werror to package-defaults.yaml
--- a/libs/api-bot/api-bot.cabal
+++ b/libs/api-bot/api-bot.cabal
@@ -4,7 +4,7 @@ cabal-version: 1.12
 --
 -- see: https://github.com/sol/hpack
 --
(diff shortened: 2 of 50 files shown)


fisx (approved):

akshaymankar (commented):
This will make it very hard for someone to use wire packages as libraries with newer versions of GHC, as things are likely to cause warnings in future. Isn't there something we could do with cabal.project?

akshaymankar:
> This will make it very hard for someone to use wire packages as libraries with newer versions of GHC, as things are likely to cause warnings in future. Isn't there something we could do with cabal.project?

That said, only I try to fiddle with this kind of stuff when I am bored, so feel free to ignore me :)

smatting:
@akshaymankar Yes, specifying
```
package wire-api
   ghc-options: -Werror
```
in `cabal.project` works as well. I'll add it to #1853 instead.

PR #1897, opened 2021-11-01
Update webapp version in Helm chart [skip ci]

zebot:
Image tag: `2021-11-01-production.0-v0.28.29-0-d919633`
Release: [`2021-11-01-production.0`](https://github.com/wireapp/wire-webapp/releases/tag/2021-11-01-production.0)


jschaul:
@rohan-wire @atomrc @tmspzz This is the automatic pull request that normally would have caused no extra work for on-premise releases. For future releases, it would be good to have this automatic PR from the web side a few days before a backend release is triggered (or needs to be triggered). It depends essentially on the web version being marked as 'production' and being deployed on our cloud.

As this was 'late', manual work had to happen resulting in https://github.com/wireapp/wire-server/pull/1892

jschaul (approved):

PR #1900, opened 2021-11-01
Update documentation of dependency setup

battermann:
## Checklist

### DIFF
--- a/docs/developer/dependencies.md
+++ b/docs/developer/dependencies.md
@@ -63,7 +63,13 @@ sudo installer -pkg /Library/Developer/CommandLineTools/Packages/macOS_SDK_heade
 
 Please refer to [Stack's installation instructions](https://docs.haskellstack.org/en/stable/README/#how-to-install).
 
-When you're done, ensure `stack --version` is recent, ideally the same as `STACK_ALPINE_VERSION` in [`build/alpine/Dockerfile.prebuilder`](../../build/alpine/Dockerfile.prebuilder).
+When you're done, ensure `stack --version` is the same as `STACK_ALPINE_VERSION` in [`build/alpine/Dockerfile.prebuilder`](../../build/alpine/Dockerfile.prebuilder).
+
+If you have to, you can downgrade stack with this command:
+
+```bash
+stack upgrade --binary-version <version>
+```
 
 ### Ubuntu / Debian
 _Note_: The packaged versions of `haskell-stack` are too old. It is recommended to follow the generic instructions or to use stack to update stack (`stack upgrade`).
--- a/docs/developer/dependencies.md
+++ b/docs/developer/dependencies.md
@@ -11,6 +11,30 @@ In addition to the information below, you can also consult the Dockerfiles for A
 
 ## General package dependencies (needed to compile Haskell services)
 
+### Nix + Direnv
+
+Using Stack's [Nix integration](https://docs.haskellstack.org/en/stable/nix_integration/), Stack will take care of installing any system
+dependencies automatically - including `cryptobox-c`. If new system dependencies are needed, add them to the `stack-deps.nix` file in the project root.
+
+If you have `direnv` and `nix`, you will automatically have `make`, `docker-compose` and `stack` in `PATH` once you `cd` into the project root and `direnv allow`.
+You can then run all the builds, and the native dependencies will be automatically present.
+
+1. Install [Nix](https://nixos.org/download.html)
+   * MacOS users with a recent Mac might need to follow [these
+   instructions](https://nixos.org/nix/manual/#sect-macos-installation)
+   * Debian users can use their distro's `nix` package, and should remember
+   
+   to add their user to the `nix-users` group in /etc/group, and re-start
+   their login session.
+2. Install [Direnv](https://direnv.net/).
+   * On debian, you can install the `direnv` package. On MacOS use `brew install direnv`.
+   * On NixOS with home-manager, you can set `programs.direnv.enable = true;`.
+   * Make sure direnv is hooked into your shell via it's appripriate `rc` file.
+     Add `eval "$(direnv hook bash|zsh|fish)"` to your ~/.(bash|zsh|fish)rc .
+   * When successfully installed and hooked, direnv should ask you to `direnv allow`
+     the current `.envrc` when you cd to this repository.
+     See the [Installation documentation](https://direnv.net/docs/installation.html) for further details.
+
 ### Fedora:
 
 ```bash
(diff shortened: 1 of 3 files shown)


flokli:
Can we drop the Fedora/Ubuntu instructions? Realistically everyone is using the Nix/direnv approach, and the other stuff isn't guaranteed to work (or bitrot over time).

sysvinit:
The current stack configuration doesn't enable the Nix integration for me automatically on Debian (though it does for @flokli on NixOS for some reason, and I haven't looked into why). Per [the stack docs](https://docs.haskellstack.org/en/stable/nix_integration/#additions-to-your-stackyaml), we should add `nix.enable: true` to `stack.yaml` if we want to enable the Nix integration unconditionally; without this I have to pass the `--nix` flag to stack to enable the Nix integration.

The direnv configuration only pulls in stack itself, and non-Haskell build dependencies like GMP and cryptobox are only pulled in through stack's Nix integration, which meant that the build failed for me due to missing dependencies without locally amending the stack config to turn that on.

I agree about the platform-specific dependency setup docs being at risk of bitrotting if nobody's regularly testing them, getting rid of them is probably not a bad idea. I'm not sure whether we want to manage non-Haskell dependencies through Nix+direnv or stack+Nix, but I think that's a separate discussion.

jschaul:
> Can we drop the Fedora/Ubuntu instructions? Realistically everyone is using the Nix/direnv approach, and the other stuff isn't guaranteed to work (or bitrot over time).

Hm, I think a potentially partially-out-of-date list of packages to install for some operating systems is still useful. We don't add new system-level dependencies to our services often (rather like once per year or so). At the moment the nix setup really only installs cryptobox, all the other compile-time and run-time dependencies are mentioned nowhere else and in this readme. 

Taking for instance brig, it has these runtime dependencies:

```
❯ ldd dist/brig
	linux-vdso.so.1 (0x00007ffcff386000)
	libm.so.6 => /lib64/libm.so.6 (0x0000700a7718f000)
	libicuuc.so.67 => /lib64/libicuuc.so.67 (0x0000700a76fa4000)
	libicui18n.so.67 => /lib64/libicui18n.so.67 (0x0000700a76c9b000)
	libicudata.so.67 => /lib64/libicudata.so.67 (0x0000700a75182000)
	libsodium.so.23 => /lib64/libsodium.so.23 (0x0000700a75129000)
	libcryptobox.so => /home/user/.wire-dev/lib/libcryptobox.so (0x0000700a74e90000)
	liblzma.so.5 => /lib64/liblzma.so.5 (0x0000700a74e62000)
	libxml2.so.2 => /lib64/libxml2.so.2 (0x0000700a74cd7000)
	libstdc++.so.6 => /lib64/libstdc++.so.6 (0x0000700a74aef000)
	libz.so.1 => /lib64/libz.so.1 (0x0000700a74ad5000)
	librt.so.1 => /lib64/librt.so.1 (0x0000700a74aca000)
	libutil.so.1 => /lib64/libutil.so.1 (0x0000700a74ac5000)
	libdl.so.2 => /lib64/libdl.so.2 (0x0000700a74abc000)
	libpthread.so.0 => /lib64/libpthread.so.0 (0x0000700a74a9a000)
	libssl.so.1.1 => /lib64/libssl.so.1.1 (0x0000700a749fd000)
	libcrypto.so.1.1 => /lib64/libcrypto.so.1.1 (0x0000700a74710000)
	libgmp.so.10 => /lib64/libgmp.so.10 (0x0000700a7466b000)
	libc.so.6 => /lib64/libc.so.6 (0x0000700a744a0000)
	libgcc_s.so.1 => /lib64/libgcc_s.so.1 (0x0000700a74483000)
	/lib64/ld-linux-x86-64.so.2 (0x0000700a772ef000)
```

The compile-time dependencies for alpine are in https://github.com/wireapp/wire-server/blob/develop/build/alpine/Dockerfile.prebuilder and the run-time dependencies are in https://github.com/wireapp/wire-server/blob/develop/build/alpine/Dockerfile.deps - but that only covers alpine. Any new person starting to develop on wire-server will likely not run alpine, the nix dependencies don't cover even half of what's needed; so what's the harm of keeping a list of packages around for a few more OSes? If we're afraid this can bitrot, we could add them to a script file and add a docker image that installs those then tries to run `make` - and add the whole thing to CI to run once per week or so.

Am I the only one who thinks this is useful? Alternatively the nix setup could be changed in such a way to actually cover all dependencies; and that same setup is used for the docker images we build and ship in the end. But that's not there yet, so I'd wait before we remove some docs that are helpful to some new people starting to work on this git repository.

docs/developer/dependencies.md
```diff
@@ -11,6 +11,30 @@ In addition to the information below, you can also consult the Dockerfiles for A
 
 ## General package dependencies (needed to compile Haskell services)
 
+### Nix + Direnv
+
+Using Stack's [Nix integration](https://docs.haskellstack.org/en/stable/nix_integration/), Stack will take care of installing any system
```
jschaul, line 16: Maybe a note should be added that this needs to be enabled manually by passing the `--nix` flag to stack or set a config in stack.yaml locally. Maybe someone for whom the nix+stack setup works out of the box could add a sentence or two here.

docs/developer/dependencies.md
```diff
@@ -11,6 +11,30 @@ In addition to the information below, you can also consult the Dockerfiles for A
 
 ## General package dependencies (needed to compile Haskell services)
 
+### Nix + Direnv
```
jschaul, line 14:
```suggestion
*Note: all the below sections for getting compile-time dependencies necessary to compile all of wire-server may potentially go out of date; if you spot a mistake please open an issue or PR*

### Nix + Direnv
```

jschaul (approved):
I think overall this changeset is an improvement over the status quo, so feel free to merge this (even if it can be improved further)

tools/nginz_disco/README.md
```diff
@@ -2,5 +2,5 @@
 
 Due to nginx not supporting DNS names for its list of upstream servers (unless you pay extra), the nginz-disco container is a simple bash script to do DNS lookups and write the resulting IPs to a file. Nginz reloads on changes to this file. 
 
-This is useful as a sidecar container to nginz in kubernetes. See also [wire-server-deploy/nginz](https://github.com/wireapp/wire-server-deploy/charts/nginz/)
+This is useful as a sidecar container to nginz in kubernetes. See also [wire-server-deploy/nginz](https://github.com/wireapp/wire-server-deploy/charts/nginz/) <!-- todo: this link is broken >
```
jschaul, line 5: this chart has been merged into wire-server, so the reference should be `./charts/nginz/`

PR #1901, opened 2021-11-01
Fix: push events when AppLock or SelfDeletingMessages config change.

fisx:
- Emit event when SelfDeletingMessages config changes (oversight in https://github.com/wireapp/wire-server/pull/1857)
- Same for AppLock config (oversight back then)
- Remove unused event (classified domains are not changed via rest api, so event cannot occur)

## Checklist

### DIFF
--- a/changelog.d/3-bug-fixes/pr-1901
+++ b/changelog.d/3-bug-fixes/pr-1901
@@ -1,0 +1,1 @@
+Push events when AppLock or SelfDeletingMessages config change.
--- a/services/galley/src/Galley/API/Teams/Features.hs
+++ b/services/galley/src/Galley/API/Teams/Features.hs
@@ -79,7 +79,13 @@ import Servant.API ((:<|>) ((:<|>)))
 import qualified Servant.Client as Client
 import qualified System.Logger.Class as Log
 import Util.Options (Endpoint, epHost, epPort)
-import Wire.API.Event.FeatureConfig (EventData (EdFeatureWithoutConfigChanged))
+import Wire.API.Event.FeatureConfig
+  ( EventData
+      ( EdFeatureApplockChanged,
+        EdFeatureSelfDeletingMessagesChanged,
+        EdFeatureWithoutConfigChanged
+      ),
+  )
 import qualified Wire.API.Event.FeatureConfig as Event
 import qualified Wire.API.Routes.Internal.Brig as IAPI
 import Wire.API.Team.Feature (AllFeatureConfigs (..), FeatureHasNoConfig, KnownTeamFeatureName, TeamFeatureName)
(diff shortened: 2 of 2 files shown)


libs/wire-api/src/Wire/API/Event/FeatureConfig.hs
```diff
@@ -72,7 +71,6 @@ taggedEventDataSchema =
       TeamFeatureDigitalSignatures -> tag _EdFeatureWithoutConfigChanged (unnamed schema)
       TeamFeatureAppLock -> tag _EdFeatureApplockChanged (unnamed schema)
       TeamFeatureFileSharing -> tag _EdFeatureWithoutConfigChanged (unnamed schema)
-      TeamFeatureClassifiedDomains -> tag _EdFeatureClassifiedDomainsChanged (unnamed schema)
       TeamFeatureConferenceCalling -> tag _EdFeatureWithoutConfigChanged (unnamed schema)
       TeamFeatureSelfDeletingMessages -> tag _EdFeatureSelfDeletingMessagesChanged (unnamed schema)
```
battermann, line 75: My linter says that the pattern match is not exhaustive and `TeamFeatureClassifiedDomains` is not matched. Not sure if this is an issue or not?

battermann (approved):
LGTM AFAICT
There is a non-exhaustive pattern match which is not really part of this PR, but anyway I think it is good to keep an eye on this.

libs/wire-api/src/Wire/API/Event/FeatureConfig.hs
battermann: This actually causes the build to fail.

libs/wire-api/src/Wire/API/Event/FeatureConfig.hs
```diff
@@ -52,7 +52,6 @@ instance ToSchema EventType where
 data EventData
   = EdFeatureWithoutConfigChanged TeamFeatureStatusNoConfig
   | EdFeatureApplockChanged (TeamFeatureStatusWithConfig TeamFeatureAppLockConfig)
```
battermann, line 54: Removing `TeamFeatureClassifiedDomainsConfig` causes build to fail due to unused imports.

battermann (changes requested):
Looks like there are two issues that cause the build to fail, see comments.

fisx:
Right.  These events aren't emitted anywhere, but removing them from the code is trickier than I thought.  So let's stick with defining events for all `TeamFeatureName`s exhaustively for now, even for the ones that aren't emitted.

battermann (approved):
LGTM

battermann (approved):

PR #1908, opened 2021-11-04
Provide ormolu via direnv, run `make formatc` in github actions

flokli:
This will provide ormolu via direnv, and do some CI steps in GH Actions:

 - ensuring the direnv environment (still) builds
 - uploading it to the cachix cache
 - run `make formatc` and provide feedback if code is not formatted.

## Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,27 @@
+on:
+ pull_request:
+ push:
+   branches: [master]
+
+jobs:
+  build-dev-env:
+   name: Build dev env
+   strategy:
+     matrix:
+       os:
+         - ubuntu-latest
+         # This is too expensive
+         # - macos-latest
+   runs-on: ${{ matrix.os }}
+   steps:
+     - uses: actions/checkout@v2
+       with:
+         submodules: true
+     - uses: cachix/install-nix-action@v14.1
+     - uses: cachix/cachix-action@v10
+       with:
+         name: wire-server
+         signingKey: '${{ secrets.CACHIX_SIGNING_KEY }}'
+         authToken: '${{ secrets.CACHIX_AUTH_TOKEN }}'
+     - name: Build the wire-server-direnv
+       run: nix-build --no-out-link direnv.nix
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -0,0 +1,31 @@
...
+       with:
+         name: wire-server
+         signingKey: '${{ secrets.CACHIX_SIGNING_KEY }}'
+         authToken: '${{ secrets.CACHIX_AUTH_TOKEN }}'
+     - name: Build the wire-server-direnv
+       run: nix-build --no-out-link direnv.nix
+     - name: Install the wire-server-direnv
+       run: nix-env -f direnv.nix -i
+     - name: Ensure everything is formatted
+       run: make formatc
```
jschaul, line 31: We already have a concourse check for the formatc thing; so these last two lines are not really necessary I'd say; but sure we can have the check run twice in case one CI is down.

jschaul (approved):
Could you add a changelog line under changelog.d/5-internal/ ?

.github/workflows/ci.yml
flokli: Concourse CI Logs aren't visible externally, and in the past, they were red for quite some time without getting fixed. Having CI definitons in the same repo, and public readable logs allows everyone to send a PR to fix it.

It'd be pretty simple to require that gh actions pipeline to be green before merging. Adding a separate concourse pipeline only for that purpose would be much more involved.

flokli:
I added the changelog entry. PTAL.

.github/workflows/ci.yml
jschaul: The ormolu check from concourse is behaving just fine and hasn't had any problems I'm aware of. As for the "visible logs for externals", there is a believe an ongoing project for a concourse-on-k8s-for-CI from the PIT team that would solve that part, but not just for ormolu, but also for all the other checks that currently run on CI.

I already approved the PR so I'm not against adding this check a second time, it's just providing little extra value at this point IMO.

PR #1912, opened 2021-11-08
Do not fail user deletion when a remote notification fails

smatting:
## Checklist

https://wearezeta.atlassian.net/browse/FS-124

### DIFF
--- a/services/brig/src/Brig/IO/Intra.hs
+++ b/services/brig/src/Brig/IO/Intra.hs
@@ -93,6 +93,7 @@ import Data.Coerce (coerce)
 import qualified Data.Conduit.List as C
 import qualified Data.Currency as Currency
 import Data.Domain
+import Data.Either.Combinators (whenLeft)
 import qualified Data.HashMap.Strict as M
 import Data.Id
 import Data.Json.Util (UTCTimeMillis, (#))
@@ -275,10 +276,8 @@ notifyUserDeletionRemotes deleted = do
         Just rangedUids -> do
           luidDeleted <- qualifyLocal deleted
           eitherFErr <- runExceptT (notifyUserDeleted luidDeleted (qualifyAs uids rangedUids))
-          case eitherFErr of
-            Left fErr -> do
-              logFederationError (tDomain uids) fErr
-            Right () -> pure ()
+          whenLeft eitherFErr $
+            logFederationError (tDomain uids)
 
     logFederationError :: Domain -> FederationError -> AppT IO ()
     logFederationError domain fErr =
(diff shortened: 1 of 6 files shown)


pcapriotti (approved):
Looks good to me.

services/brig/src/Brig/IO/Intra.hs
```diff
@@ -278,9 +278,6 @@ notifyUserDeletionRemotes deleted = do
           case eitherFErr of
             Left fErr -> do
               logFederationError (tDomain uids) fErr
-              -- FUTUTREWORK: Do something better here?
-              -- FUTUREWORK: Write test that this happens
-              throwM $ federationErrorToWai fErr
             Right () -> pure ()
```
mdimjasevic, line 281: A pattern like this is encapsulated by `whenLeft :: Applicative m => Either a b -> (a -> m ()) -> m ()`.

services/brig/test/integration/API/User/Account.hs
```diff
@@ -1257,6 +1258,56 @@ testDeleteWithRemotes opts brig = do
...
+          then F.OutwardResponseError (F.OutwardError F.ConnectionRefused "mocked connection problem with b domain")
+          else OutwardResponseBody (cs $ Aeson.encode EmptyResponse)
+
+  let galleyHandler :: ReceivedRequest -> MockT IO Wai.Response
+      galleyHandler (ReceivedRequest requestMethod requestPath _requestBody) =
+        case (requestMethod, requestPath) of
+          (_methodDelete, ["i", "user"]) -> do
+            let response = Wai.responseLBS Http.status200 [(Http.hContentType, "application/json")] (cs $ Aeson.encode EmptyResponse)
+            pure response
+          _ -> error "not mocked"
```
mdimjasevic, line 1288: There's no need to pattern match on `requestMethod` as it is vacuous, right?

services/galley/test/integration/API.hs
```diff
@@ -3239,23 +3248,34 @@ removeUser = do
...
+          | otherwise = error "unmocked domain"
+
     (_, fedRequests) <-
-      withTempServantMockFederator (const emptyFederatedBrig) galleyApi localDomain $
+      withTempMockFederator' handler $
         deleteUser alexDel' !!! const 200 === statusCode
 
     liftIO $ do
-      assertEqual ("expect exactly 5 federated requests in : " <> show fedRequests) 5 (length fedRequests)
+      assertEqual ("expect exactly 5 federated requests in : " <> show fedRequests) 7 (length fedRequests)
```
mdimjasevic, line 3278: Should the quoted string also have 7 in it instead of 5?

mdimjasevic (approved):
This looks good. A few smaller comments are left inlined.

services/brig/src/Brig/IO/Intra.hs
smatting: Ah nice, done.

services/brig/test/integration/API/User/Account.hs
smatting: Yes, you're right. I changed this to a proper pattern match in https://github.com/wireapp/wire-server/pull/1912/commits/d92e7fa976992f97f17733d0147c2fecd0d7e88b

PR #1914, opened 2021-11-08
Ingress tls1.3 [WIP]

jschaul:
https://wearezeta.atlassian.net/browse/FS-33
version support: https://github.com/kubernetes/ingress-nginx#support-versions-table

This PR, once working (still WIP)

* allows TLS 1.3 connections (while maintaining TLS 1.2 support)
* upgrades the nginx-ingress chart to a recent version that supports TLS 1.3. This, as a result, requires kubernetes >= 1.19 (added a changelog entry for that)
* Any overrides will need to be adjusted (added a changelog entry for that)

Current issues:

* `Error: Service "test-bw4377dj3tiz-ic-2-ingress-nginx-controller" is invalid: spec.ports[0].nodePort: Invalid value: 31772: provided port is already allocated`: seems like we can't do an in-place upgrade with helm on the existing controller; instead a migration needs to be devised & documented.

TODO:

* [ ] test on different k8s versions
* [ ] maybe also update api version of ingress definitions to match k8s 1.19+

## Checklist

### DIFF
--- a/charts/nginx-ingress-controller/values.yaml
+++ b/charts/nginx-ingress-controller/values.yaml
@@ -1,5 +1,6 @@
 # Default values for nginx-ingress-controller
 
+# TODO: this should probably be ingress-nginx now
 nginx-ingress:
   controller:
     config:
--- a/charts/nginx-ingress-controller/values.yaml
+++ b/charts/nginx-ingress-controller/values.yaml
@@ -8,7 +8,9 @@ nginx-ingress:
       # As a Wire employee, for Wire-internal discussions and context see
       # * https://wearezeta.atlassian.net/browse/FS-33
       # * https://wearezeta.atlassian.net/browse/FS-444
+      # TODO add TLS 1.3 ciphers to the list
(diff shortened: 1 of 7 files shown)


charts/nginx-ingress-controller/values.yaml
```diff
@@ -5,6 +5,7 @@ nginx-ingress:
     config:
       # NOTE: These are some sane defaults, you may want to overrride them on your own installation
       ssl-ciphers: "ECDHE-ECDSA-AES256-GCM-SHA384:ECDHE-RSA-AES256-GCM-SHA384:ECDHE-ECDSA-CHACHA20-POLY1305:ECDHE-RSA-CHACHA20-POLY1305:ECDHE-ECDSA-AES128-GCM-SHA256:ECDHE-RSA-AES128-GCM-SHA256"
```
comawill, line 7: TLS1.3 ciphers are currently missing here

jschaul:
Superseded by https://github.com/wireapp/wire-server/pull/3140

PR #1916, opened 2021-11-09
Lock/Unlock Payment Status for Self Deleting Messages

battermann:
## Checklist


services/galley/src/Galley/API/Internal.hs
```diff
@@ -244,6 +247,16 @@ type IFeatureStatusPut featureName =
...
     :> Put '[Servant.JSON] (Public.TeamFeatureStatus featureName)
 
+type IFeatureStatusUnlockPaymentPut featureName =
+  Summary (AppendSymbol "Unlock payment for " (Public.KnownTeamFeatureNameSymbol featureName))
+    :> "i"
+    :> "teams"
+    :> Capture "tid" TeamId
+    :> "features"
+    :> Public.KnownTeamFeatureNameSymbol featureName
+    :> "unlock"
```
fisx, line 257: note to self: we agreed this should be a type variable `lockOrUnlock` to avoid having to duplicate this type alias.

services/galley/src/Galley/API/Internal.hs
```diff
@@ -318,6 +332,14 @@ iPutTeamFeature ::
   Galley r (Public.TeamFeatureStatus a)
 iPutTeamFeature setter = Features.setFeatureStatus @a setter DontDoAuth
 
+iUnlockPayment ::
+  forall a r.
+  (Public.KnownTeamFeatureName a) =>
+  (TeamId -> Galley r Public.PaymentStatus) ->
```
fisx, line 338: we probably don't need this.  it's only there in `getSelfDeletingMessagesInternal` etc. because of the different configuration types for different features.  payment status has no heterogeneous config attribute.

services/galley/src/Galley/API/Internal.hs
```diff
@@ -318,6 +332,14 @@ iPutTeamFeature ::
   Galley r (Public.TeamFeatureStatus a)
 iPutTeamFeature setter = Features.setFeatureStatus @a setter DontDoAuth
 
+iUnlockPayment ::
```
fisx, line 335: not sure at this point, but just to not forget this: we may want to make this `iLockOrUnlock`, and pass a `Public.PaymentStatus` as an extra argument.  then we don't need to write `iLockPayment`.  Same probably goes for other places further down the call graph like the function to write into cassandra etc.

services/galley/src/Galley/API/Teams/Features.hs
```diff
@@ -496,6 +497,11 @@ getSelfDeletingMessagesInternal = \case
       TeamFeatures.getSelfDeletingMessagesStatus tid
         <&> maybe Public.defaultSelfDeletingMessagesStatus id
 
+setSelfDeletingMessagesPaymentUnlockInternal ::
```
fisx, line 500: we may want to keep this definition for symmetry with the rest of the code, but call it from `iUnlockPayment` directly instead of passing it into `iUnlockPayment` as an argument.  (not sure about this thought either.)

fisx (commented):
looks good so far!

libs/wire-api/src/Wire/API/Team/Feature.hs
```diff
@@ -451,6 +456,82 @@ defaultSelfDeletingMessagesStatus =
...
+instance ToSchema PaymentStatus where
+  schema =
+    object "PaymentStatus" $
+      PaymentStatus
+        <$> paymentStatus .= field "paymentStatus" schema
+
+modelPaymentStatus :: Doc.Model
+modelPaymentStatus =
+  Doc.defineModel "PaymentStatus" $ do
+    Doc.property "paymentStatus" typePaymentStatusValue $ Doc.description "payment status: locked | unlocked"
```
fisx, line 492:
```suggestion
    Doc.property "paymentStatus" typePaymentStatusValue
```

the description doesn't add any information to `typePaymentStatusValue`, no?

libs/wire-api/src/Wire/API/Team/Feature.hs
```diff
@@ -451,6 +456,82 @@ defaultSelfDeletingMessagesStatus =
     TeamFeatureEnabled
     (TeamFeatureSelfDeletingMessagesConfig 0)
 
+----------------------------------------------------------------------
+-- PaymentStatus
+
+data PaymentStatusUpdate = Lock | Unlock
```
fisx, line 462: Why don't you use `PaymentStatus` for this?  Then you could also use `FromByteString` to implement the `FromHttpApiData` instance.

libs/wire-api/test/unit/Test/Wire/API/Roundtrip/Aeson.hs
```diff
@@ -206,6 +206,7 @@ tests =
       testRoundTrip @(Team.Feature.TeamFeatureStatus 'Team.Feature.TeamFeatureConferenceCalling),
       testRoundTrip @(Team.Feature.TeamFeatureStatus 'Team.Feature.TeamFeatureSelfDeletingMessages),
       testRoundTrip @Team.Feature.TeamFeatureStatusValue,
+      testRoundTrip @Team.Feature.PaymentStatus,
```
fisx, line 209: I'm assuming you deliberately skipped `PaymentStatusValue` since it's implicit?

services/galley/schema/src/V55_TeamFeatureSelfDeletingMessagesPaymentStatus.hs
```diff
@@ -0,0 +1,34 @@
...
+  )
+where
+
+import Cassandra.Schema
+import Imports
+import Text.RawString.QQ
+
+migration :: Migration
+migration = Migration 55 "Add payment status config for self-deleting messages" $ do
+  -- 0: Payment Status locked, 1: Payment Status unlocked
```
fisx, line 29: This is clear from the Cql instance, and a comment here is at risk of getting outdated.  Just remove?  Or link the haskell type or its Cql instance?

services/galley/schema/src/Main.hs
```diff
@@ -99,7 +100,8 @@ main = do
       V51_FeatureFileSharing.migration,
       V52_FeatureConferenceCalling.migration,
       V53_AddRemoteConvStatus.migration,
-      V54_TeamFeatureSelfDeletingMessages.migration
+      V54_TeamFeatureSelfDeletingMessages.migration,
+      V55_TeamFeatureSelfDeletingMessagesPaymentStatus.migration
       -- When adding migrations here, don't forget to update
       -- 'schemaVersion' in Galley.Data
```
fisx, line 106: You forgot to update `schemaVersion` as explained here.

services/galley/src/Galley/API/Internal.hs
```diff
@@ -245,6 +248,16 @@ type IFeatureStatusPut featureName =
     :> ReqBody '[Servant.JSON] (Public.TeamFeatureStatus featureName)
     :> Put '[Servant.JSON] (Public.TeamFeatureStatus featureName)
 
+type IFeatureStatusPaymentStatusPut featureName =
+  Summary (AppendSymbol "Unlock payment for " (Public.KnownTeamFeatureNameSymbol featureName))
```
fisx, line 252:
```suggestion
  Summary (AppendSymbol "(Un-)lock payment for " (Public.KnownTeamFeatureNameSymbol featureName))
```

services/galley/src/Galley/Cassandra/TeamFeatures.hs
```diff
@@ -140,13 +140,27 @@ setSelfDeletingMessagesStatus tid status = do
           <> ", self_deleting_messages_ttl) "
           <> "values (?, ?, ?)"
 
+setSelfDeletingMessagesPaymentStatus ::
```
fisx, line 143:
```suggestion
-- FUTUREWORK: this function can be made polymorphic in the `TeamFeatureName` in the spirit
-- of `setFeatureStatusNoConfig` (defined futher up in this module).
setSelfDeletingMessagesPaymentStatus ::
```

It's ok not to do that now, but when we get to making the next feature (un-)lockable we don't want to duplicate this function, but re-use it.

Come to think of it: we should make this polymorphic now, then the implementation of the handler function that's calling this will also become easier: `setPaymentStatus` won't need the second argument, but can always call this polymorphic function without worrying about the concrete `TeamFeatureName`.

services/galley/src/Galley/API/Teams/Features.hs
```diff
@@ -550,6 +572,18 @@ getSelfDeletingMessagesInternal = \case
...
+setSelfDeletingMessagesPaymentStatusInternal ::
+  Member TeamFeatureStore r =>
+  TeamId ->
+  Public.PaymentStatusUpdate ->
+  Galley r Public.PaymentStatus
+setSelfDeletingMessagesPaymentStatusInternal tid paymentStatusUpdate =
+  liftSem $ TeamFeatures.setFeatureSelfDeletingMessagesPaymentStatus tid (Public.PaymentStatus paymentStatus)
+  where
+    paymentStatus = case paymentStatusUpdate of
+      Public.Lock -> Public.PaymentLocked
```
fisx, line 584: Yes, as mentioned earlier, if we have just `PaymentStatus`, and not `PaymentStatusUpdate` on top of it, things will get easier all over the place...

fisx (changes requested):
Looks good, but of course I have comments.  :)

libs/wire-api/test/unit/Test/Wire/API/Roundtrip/Aeson.hs
battermann: Ye, but I will add it, too. Doesn't hurt.

libs/wire-api/src/Wire/API/Team/Feature.hs
battermann: That is true. I'll leave it empty then.

services/galley/src/Galley/API/Teams/Features.hs
battermann: Fixed it

services/galley/src/Galley/API/Internal.hs
battermann: Fixed

services/galley/schema/src/Main.hs
battermann: Fixed

services/galley/schema/src/V55_TeamFeatureSelfDeletingMessagesPaymentStatus.hs
battermann: Done

libs/wire-api/src/Wire/API/Team/Feature.hs
battermann: Done

services/galley/src/Galley/API/Teams/Features.hs
battermann: removed

services/galley/src/Galley/Cassandra/TeamFeatures.hs
battermann: done

libs/wire-api/src/Wire/API/Team/Feature.hs
```diff
@@ -535,6 +536,12 @@ instance Cass.Cql PaymentStatusValue where
   toCql PaymentLocked = Cass.CqlInt 0
   toCql PaymentUnlocked = Cass.CqlInt 1
 
+getPaymentStatus ::
+  forall (a :: TeamFeatureName).
+  TeamFeatureStatus a ->
+  Maybe PaymentStatusValue
+getPaymentStatus _ = error "todo: implement"
```
fisx, line 543: Why not use `tfwcPaymentStatus` instead of `getPaymentStatus`?

services/galley/src/Galley/API/Error.hs
```diff
@@ -457,6 +459,9 @@ noLegalHoldDeviceAllocated = mkError status404 "legalhold-no-device-allocated" "
 legalHoldCouldNotBlockConnections :: Error
 legalHoldCouldNotBlockConnections = mkError status500 "legalhold-internal" "legal hold service: could not block connections when resolving policy conflicts."
 
+setTeamFeatureConfigPaymentStatusLocked :: Error
+setTeamFeatureConfigPaymentStatusLocked = mkError status409 "payment-status-locked" "feature config cannot be updated when the payment status is locked"
```
fisx, line 463: https://wearezeta.atlassian.net/browse/SQSERVICES-960?focusedCommentId=97383

services/galley/src/Galley/API/Teams/Features.hs
```diff
@@ -124,23 +124,29 @@ setFeatureStatus ::
...
       r
   ) =>
+  (GetFeatureInternalParam -> Galley r (Public.TeamFeatureStatus a)) ->
   (TeamId -> Public.TeamFeatureStatus a -> Galley r (Public.TeamFeatureStatus a)) ->
   DoAuth ->
   TeamId ->
   Public.TeamFeatureStatus a ->
   Galley r (Public.TeamFeatureStatus a)
-setFeatureStatus setter doauth tid status = do
+setFeatureStatus getter setter doauth tid status = do
```
fisx, line 138: I don't think we have to (or want to) change the signature of `iPutTeamFeature`.  Instead of `getter`, I would write a `get`-variant of `Galley.Cassandra.TeamFeatures.setPaymentStatus` and call that directly.  All you need for that is `a` from the type signature here.

This would make a lot of noise in this change evaporate.

services/galley/src/Galley/API/Error.hs
battermann: https://wearezeta.atlassian.net/browse/SQSERVICES-960?focusedCommentId=97409

services/galley/src/Galley/API/Teams/Features.hs
battermann: I wasn't sure, but yes this is an option, and I'll go for that then. Thanks!

services/galley/src/Galley/API/Teams/Features.hs
```diff
@@ -529,7 +558,7 @@ getClassifiedDomainsInternal _mbtid = do
   let config = globalConfig
   pure $ case Public.tfwcStatus config of
     Public.TeamFeatureDisabled ->
-      Public.TeamFeatureStatusWithConfig Public.TeamFeatureDisabled (Public.TeamFeatureClassifiedDomainsConfig [])
+      Public.TeamFeatureStatusWithConfig Public.TeamFeatureDisabled (Public.TeamFeatureClassifiedDomainsConfig []) (Just Public.PaymentLocked)
```
fisx, line 561:
```suggestion
    defaultClassifiedDomains
```

services/galley/src/Galley/Cassandra/TeamFeatures.hs
```diff
@@ -73,13 +73,13 @@ getApplockFeatureStatus tid = do
   let q = query1 select (params LocalQuorum (Identity tid))
   mTuple <- retry x1 q
   pure $
-    mTuple >>= \(mbStatusValue, mbEnforce, mbTimeout) ->
-      TeamFeatureStatusWithConfig <$> mbStatusValue <*> (TeamFeatureAppLockConfig <$> mbEnforce <*> mbTimeout)
+    mTuple >>= \(mbStatusValue, mbEnforce, mbTimeout, mbPaymentStatusValue) ->
+      TeamFeatureStatusWithConfig <$> mbStatusValue <*> (TeamFeatureAppLockConfig <$> mbEnforce <*> mbTimeout) <*> Just mbPaymentStatusValue
```
fisx, line 77: Does this do the right thing for `null` values?  I don't think it does, since the `<*>` catches the `Just` you've added, and not the one from within the `mbPaymentStatusValue`.

services/galley/src/Galley/Cassandra/TeamFeatures.hs
```diff
@@ -140,13 +140,61 @@ setSelfDeletingMessagesStatus tid status = do
...
+  ( MonadClient m,
+    HasPaymentStatusCol a
+  ) =>
+  Proxy a ->
+  TeamId ->
+  PaymentStatus ->
+  m PaymentStatus
+setPaymentStatus _ tid (PaymentStatus paymentStatus) =
+  case paymentStatusCol @a of
+    Nothing -> pure $ PaymentStatus PaymentUnlocked -- todo: what should we do here? probably return an error, UpdatePaymentStatusNotAllowed or sth. similar
```
fisx, line 154:
```suggestion
    Nothing -> pure $ PaymentStatus PaymentUnlocked -- TODO(battermann): what should we do here? probably return an error, UpdatePaymentStatusNotAllowed or sth. similar
```

And yes, we should throw an error here.

We could avoid this runtime error by splitting up `HasPaymentStatusCol` into one for read access and one for write access.  The one for read access returns a `Maybe String`, and the one for write access is just instantiated for those features that actually do write their payment status to cassandra.  Then the type checker make sure we can't write this line in any reasonable way.

services/galley/src/Galley/Cassandra/TeamFeatures.hs
```diff
@@ -140,13 +140,61 @@ setSelfDeletingMessagesStatus tid status = do
...
+    Just col -> do
+      retry x5 $ write insert (params LocalQuorum (tid, paymentStatus))
+      pure (PaymentStatus paymentStatus)
+      where
+        insert :: PrepQuery W (TeamId, PaymentStatusValue) ()
+        insert =
+          fromString $
+            "insert into team_features (team_id, "
+              <> col
+              <> ") values (?, ?)"
```
fisx, line 164:
```suggestion
        insert = fromString $ "insert into team_features (team_id, " <> col <> ") values (?, ?)"
```

services/galley/src/Galley/Cassandra/TeamFeatures.hs
```diff
@@ -140,13 +140,61 @@ setSelfDeletingMessagesStatus tid status = do
...
+  forall (a :: TeamFeatureName) m.
+  ( MonadClient m,
+    HasPaymentStatusCol a
+  ) =>
+  Proxy a ->
+  TeamId ->
+  m (Maybe PaymentStatus)
+getPaymentStatus _ tid =
+  case paymentStatusCol @a of
+    Nothing -> pure Nothing
```
fisx, line 176: I would have expected `unlocked` here instead of nothing, but you can also reasonably decide to handle this further up the call tree.

fisx (changes requested):
tests missing, and some more changes, but you're getting there!  :)

services/galley/src/Galley/Cassandra/TeamFeatures.hs
battermann: Yes, it was my intention to push this to the caller.

services/galley/test/integration/API/Teams/Feature.hs
```diff
@@ -446,19 +449,21 @@ testAllFeatures = do
...
             .= Public.TeamFeatureStatusNoConfig confCalling,
           toS TeamFeatureSelfDeletingMessages
-            .= ( Public.TeamFeatureStatusWithConfig @Public.TeamFeatureSelfDeletingMessagesConfig
-                   TeamFeatureEnabled
-                   (Public.TeamFeatureSelfDeletingMessagesConfig 0)
-               )
+            .= Public.TeamFeatureStatusWithConfig @Public.TeamFeatureSelfDeletingMessagesConfig
+              TeamFeatureEnabled
+              (Public.TeamFeatureSelfDeletingMessagesConfig 0)
+              Nothing
```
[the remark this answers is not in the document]
battermann: This resolved itself after introducing the feature status with and without payment status.

services/galley/src/Galley/Cassandra/TeamFeatures.hs
battermann: This is also resolved by the with/without payment status change.

fisx:
obsoleted by https://github.com/wireapp/wire-server/pull/1937

PR #1919, opened 2021-11-09
Report QualifiedCapture correctly in prometheus metrics

akshaymankar:
For prometheus metrics middleware to be able to replace
/users/4f7cbd8c-5fe3-4818-94c0-8ae68460ba13 with /users/:uid, it needs to know
the paths in servant that exist. This is generated statically using the class
`RoutesToPaths`. This class had an overlappable instance for everything, this
caused to not notice when we created the QualifiedCapture type. In order to
ensure that we instantiate this class correctly, this commit removes this
catch-all instance and instantiate the class for every type that needs it
explicitly.

https://wearezeta.atlassian.net/browse/FS-224

## Checklist

### DIFF
--- a/changelog.d/3-bug-fixes/metrics-handlers
+++ b/changelog.d/3-bug-fixes/metrics-handlers
@@ -1,0 +1,1 @@
+Ensure that all endpoints have a correct handler in prometheus metrics
--- a/libs/extended/extended.cabal
+++ b/libs/extended/extended.cabal
@@ -4,7 +4,7 @@ cabal-version: 1.12
 --
 -- see: https://github.com/sol/hpack
 --
--- hash: 18004f2559de4d4ac804d0a36a2d2780d0bbc10d79ebf78f337c9e6ba7a4ff3f
+-- hash: 65015665656bc1ae721971ef3e88ed707aa7a2be02ba04cf4aab39ac6188714a
 
 name:           extended
 version:        0.1.0
@@ -42,6 +42,7 @@ library
     , extra
     , http-types
     , imports
+    , metrics-wai
     , optparse-applicative
     , servant
     , servant-server
--- a/libs/extended/package.yaml
+++ b/libs/extended/package.yaml
@@ -27,6 +27,7 @@ dependencies:
 # for servant's 'ReqBodyCustomError' type defined here.
 - errors
 - http-types
+- metrics-wai
 - servant
 - servant-server
 - servant-swagger
--- a/libs/extended/src/Servant/API/Extended.hs
+++ b/libs/extended/src/Servant/API/Extended.hs
@@ -23,6 +23,7 @@ module Servant.API.Extended where
 
 import qualified Data.ByteString.Lazy as BL
 import Data.EitherR (fmapL)
+import Data.Metrics.Servant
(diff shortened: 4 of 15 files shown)


pcapriotti:
Personally, I'd prefer to keep the overlappable instances around, and maybe have some more tests that check that the captures are collected correctly. Also, I think it's better to test against the actual galley/brig API, instead of an artificial one. The way it currently is will fail if we say accidentally delete the instance for `QualifiedCapture` (unlikely), but not if we introduce some other capturing combinator without the corresponding instance (much more likely).

smatting (approved):

akshaymankar:
> Personally, I'd prefer to keep the overlappable instances around, and maybe have some more tests that check that the captures are collected correctly. Also, I think it's better to test against the actual galley/brig API, instead of an artificial one. The way it currently is will fail if we say accidentally delete the instance for `QualifiedCapture` (unlikely), but not if we introduce some other capturing combinator without the corresponding instance (much more likely).

I removed the overlappable instances exactly because I want compiler to tell us about the missing instance when we introduce a new capturing combinator. Even if we delete the instance for `QualifiedCapture`, the compilation would fail (not just of the test, but at the point where we try to use the middleware).
I wanted to originally test against the brig/galley API, but I couldn't figure out any way to do that without creating another golden test and just ask people to be careful when it breaks. That said, please let me know if you have any ideas.

pcapriotti:
> I removed the overlappable instances exactly because I want compiler to tell us about the missing instance when we introduce a new capturing combinator. Even if we delete the instance for `QualifiedCapture`, the compilation would fail (not just of the test, but at the point where we try to use the middleware). 

I understand that, but my point was that perhaps we can achieve the same effect with a test, instead of having all these trivial instances around, which need to be extended every time we add a new combinator, even if it has no connection to metrics.

> I wanted to originally test against the brig/galley API, but I couldn't figure out any way to do that without creating another golden test and just ask people to be careful when it breaks. That said, please let me know if you have any ideas.

Why? Couldn't we call `getRoutes` on say the brig api, and assert that the result contains a few of the things that we expect?

In any case, this is a minor point. If you disagree that having all these trivial instances is annoying, feel free to merge anyway.

akshaymankar:
> Why? Couldn't we call getRoutes on say the brig api, and assert that the result contains a few of the things that we expect?

How would this ensure that a new capture type has a correct instance? I guess it would tell us if we replace our current type, but if we use it in a new endpoint, we will still not know.

>  If you disagree that having all these trivial instances is annoying

I don't disagree, the trivial instances are annoying. But I don't know how to not write a good enough test which ensures that all the endpoints are correctly reported in metrics.

pcapriotti:
> How would this ensure that a new capture type has a correct instance? I guess it would tell us if we replace our current type, but if we use it in a new endpoint, we will still not know.

Such a test might have caught the problem solved by this PR though, because we created a new combinator and used in existing routes. But you're right, it won't catch cases where we create a new combinator and only use it in new routes.

> I don't disagree, the trivial instances are annoying. But I don't know how to not write a good enough test which ensures that all the endpoints are correctly reported in metrics.

Feel free to merge, it's really not a big deal.

PR #1923, opened 2021-11-10
Define and expose Federation API version

smatting:
https://wearezeta.atlassian.net/browse/FS-127

Perhaps a set of known compatible versions should be exposed too.

## Checklist

### DIFF
--- a/libs/wire-api/src/Wire/API/Routes/Public/Federation.hs
+++ b/libs/wire-api/src/Wire/API/Routes/Public/Federation.hs
@@ -1,0 +1,64 @@
+-- This file is part of the Wire Server implementation.
+--
+-- Copyright (C) 2021 Wire Swiss GmbH <opensource@wire.com>
+--
+-- This program is free software: you can redistribute it and/or modify it under
+-- the terms of the GNU Affero General Public License as published by the Free
+-- Software Foundation, either version 3 of the License, or (at your option) any
+-- later version.
+--
+-- This program is distributed in the hope that it will be useful, but WITHOUT
+-- ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+-- FOR A PARTICULAR PURPOSE. See the GNU Affero General Public License for more
+-- details.
+--
+-- You should have received a copy of the GNU Affero General Public License along
+-- with this program. If not, see <https://www.gnu.org/licenses/>.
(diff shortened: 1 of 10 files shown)


libs/wire-api/src/Wire/API/Routes/Public/Federation.hs
```diff
@@ -0,0 +1,64 @@
...
+import Servant.API.Generic
+import Servant.Swagger (toSwagger)
+
+-- | Version of Wire's Federation API
+--
+-- Two releases of Wire are fully compatible in all federated features if they
+-- share the same federation version.
+--
+-- FUTUREWORK: Define a compatability matrix when version 2 is released.
+type FederationVersion = 1
```
jschaul, line 36: From FS-127: "Automation can read the federation number from the source code." - is this a good place to keep this version, or should it be in a separate file maybe?

mdimjasevic: I'm open to suggestions. Would it be a separate Haskell file or some other file type? Perhaps it would be misleading to have it in a configuration file as it would suggest it is up to the operator to tweak it, while in reality it's a version constant.

@pcapriotti , any comments?

rohan-wire: From our discussions, I thought that we would provide a range or list of supported API versions. A simplest implementation of this would be to advertise a minimum implemented API version number and a maximum implemented API version number. Both values would be 1 with this initial release.

pcapriotti: I think it's fine to put it in a haskell module like this, as long as it's easy to grep, just so that it's easy to access it without running anything, in case that's needed at some point.

mdimjasevic: > From our discussions, I thought that we would provide a range or list of supported API versions. A simplest implementation of this would be to advertise a minimum implemented API version number and a maximum implemented API version number.

Wouldn't this leave room for misalignment in version compatibility? If we let a backend claim arbitrary lower and upper bounds in this endpoint, then that does not have to be in line with our view of version compatibility based on our test suite.

smatting:
This has been superseeded by other attempts

PR #1928, opened 2021-11-15
Handle Galley errors in federation RPCs

pcapriotti:
Errors when leaving a conversation are now correctly handled instead of resulting in a generic federation error.

Tracked by https://wearezeta.atlassian.net/browse/FS-40.

## Checklist


services/galley/test/integration/API.hs
```diff
@@ -183,6 +183,8 @@ tests s =
           test s "delete conversations/:domain/:cnv/members/:domain/:usr - local conv with locals and remote, delete local" deleteLocalMemberConvLocalQualifiedOk,
           test s "delete conversations/:domain/:cnv/members/:domain/:usr - local conv with locals and remote, delete remote" deleteRemoteMemberConvLocalQualifiedOk,
           test s "delete conversations/:domain/:cnv/members/:domain/:usr - remote conv, leave conv" leaveRemoteConvQualifiedOk,
+          test s "delete conversations/:domain/:cnv/members/:domain/:usr - remote conv, leave conv, non-existent" leaveNonExistentRemoteConv,
+          test s "delete conversations/:domain/:cnv/members/:domain/:usr - remote conv, leave conv, non-existent" leaveRemoteConvDenied,
```
mdimjasevic, line 187: Shouldn't the last part of the test description be about being denied instead of non-existent?

mdimjasevic (approved):

services/galley/test/integration/API.hs
pcapriotti: Fixed, thanks!

PR #1930, opened 2021-11-15
Replace gRPC with HTTP2 in federator

smatting:
This PR changes the federation API to be based on HTTP2 instead of gRPC.

The dependency on gRPC has been completely removed from the code base.  Federator now works by accepting and forwarding HTTP2 requests directly.  Failures are propagated using normal HTTP error codes, and using Wai.Error values as responses.

Federator accepts the following routes:

 * federation client / outward service:
      `/rpc/:domain/:component/:rpc`

 * inward ingress:
      `/federation/:component/:rpc`

The origin domain is passed as a header directly in the form expected by the federation API in brig and galley (i.e. a `Wire-Origin-Domain` header).  Server-to-server authentication is not affected, and the client certificate is received by the inward service as a header, exacly as it was before.

The federation API is still following the same conventions as before, with the exception that now we require that every endpoint consists of a single path segment. It is still accessible at the `/federation` path prefix, for both galley and brig, although the Servant API types do not contain this prefix explicitly anymore.

Federator client has been completely rewritten, and now implements HTTP2 client functionality on top of the http2 library. The low-level http2 client functionality that establishes a connection supports both plaintext and TLS, but not connection reuse. It is used both by federator client and the outward service of federator itself.

The origin domain header type in Servant is now a special combinator that works just like a header on the server side, but is removed from the resulting client type. Federator client knows to add the origin domain header on every request, whether or not the origin domain is present in the API type.

Validation of federated requests has been simplified. In particular, path sanitisation has been removed, since the RPC path is now a normal HTTP path, which is assumed to have already been validated and sanitised by the HTTP server (and probably all the intermediate proxies and load balancers).

Tracked by https://wearezeta.atlassian.net/browse/FS-262.

## Checklist


services/federator/test/unit/Test/Federator/Validation.hs
```diff
@@ -76,224 +76,167 @@ tests =
...
-        ]
-  expectNormalized <$> paths
-  where
-    expectNormalized :: (ByteString, ByteString) -> TestTree
-    expectNormalized (input, output) = do
-      testCase ("Should allow " <> cs input <> " and normalize to " <> cs output) $ do
-        runSanitize input @?= Right output
-
-validatePathForbidden :: [TestTree]
-validatePathForbidden = do
```
jschaul, line 250: Could you leave these tests (or tests similar in spirit) in? Assuming some other component already validated a path so that disallowed paths cannot be called is not a good assumption I'd say. What prevents me from calling federator with /federation/brig/<double percent encoded form of /i/delete-user> where maybe one set of encoding is removed from e.g. the ingress, but another encoding makes its way through?

Maybe there is no way this can happen, but then these tests should do no harm.

The alternative would be to validate incoming paths to a known list of path segments for that component, e.g. extract all the /federation/some-path segments out from the servant API and ensure incoming requests are an exact match of one of those.

At the moment, the code blindly trusts this text component after /federation/ to be valid, as long as it's not empty. That leaves a risk that someone can craft some weird path allowing an external to use an otherwise-unreachable path brig or galley exposes. 

Am I holding things wrong?

pcapriotti: I think it's highly unlikely that there could be any attack like this, now that the RPC is just a path segment of the initial request, but we did leave most of those tests around to be extra safe. They are in `testInvalidPaths` in the `ExternalService` tests.

Here's what we removed (roughly):

 - Fragments being disallowed: this is slightly annoying because the Wai stuff does not handle fragments appropriately (it just treats `#` as a normal character). Anyway, they are innocuous AFAICT.
 - Length limits This will happen at some earlier point anyway.
 - ASCII only. There is no harm in allowing non-ASCII characters. They are roundtripped as UTF-8, and they will just result in a 404 downstream.

Path traversals via percent encoding are still tested, but the corresponding paths are allowed instead of being rejected. They get percent encoded again before being forwarded, so they are safe.

I'm actually wondering if there is actually any logic anywhere that interprets `..` as parent directory, enabling traversals. I don't think Wai or Servant do this. Were any path traversal attacks ever possible at any point?

smatting: I added a validation step that enforces characters on rpc name in https://github.com/wireapp/wire-server/pull/1930/commits/356c8480b14819977a56a7d921dbcec9e85007fe

pcapriotti (approved):

PR #1940, opened 2021-11-23
Polysemy Spar: Laws for DefaultSsoCode, Now, IdP and ScimExternalIdStore

isovector:
This PR adds polysemy-check laws for `Now`, `DefaultSsoCode`, `IdP` and `ScimExternalIdStore`. `Now` is just a check that the function is monotonic, and the other three are the usual lens laws (+ delete). 

The remaining "store" effects in `spar` have weird TTL semantics that I haven't yet worked out what the laws should be. I'll follow up with those when I figure the semantics out.

## Checklist


fisx:
> The remaining "store" effects in `spar` have weird TTL semantics that I haven't yet worked out what the laws should be. I'll follow up with those when I figure the semantics out.

you should interview your client on those.  :)  you can talk to either me or @smatting to learn more.

isovector:
CI seems to be genuinely broken. I guess I'll brave setting up the integration tests locally tomorrow :scream:

smatting:
@isovector FYI, the README.md has a section on how to run them locally.

isovector:
Turns out CI was flakey. Everything is fixed now. @pcapriotti, @fisx , PTAL.

services/spar/test/Test/Spar/Sem/NowSpec.hs
```diff
@@ -0,0 +1,68 @@
...
+  PropConstraints r f =>
+  (forall a. Sem r a -> IO (f a)) ->
+  Property
+prop_nowNow =
+  -- NOTE: This @Input ()@ effect is a workaround to an oversight in
+  -- @polysemy-check@. 'prepropLaw' wants to synthesize some actions to run
+  -- before and after its generators, and check their results for equality. We
+  -- can't use 'Now' as this effect, because 'E.get' won't return equivalent
+  -- results! And we can't keep it empty, because that triggers a crash in
+  -- @polysemy-check@. Thus @Input ()@, which isn't beautiful, but works fine.
```
fisx, line 63: if it's an oversight, is there a ticket in polysemy-check for allowing empty effect lists?  do you have an idea how to fix it?

services/spar/test/Test/Spar/Sem/ScimExternalIdStoreSpec.hs
```diff
@@ -0,0 +1,165 @@
...
+import Test.QuickCheck
+
+deriveGenericK ''E.ScimExternalIdStore
+
+propsForInterpreter ::
+  PropConstraints r f =>
+  String ->
+  (forall a. Sem r a -> IO (f a)) ->
+  Spec
+propsForInterpreter interpreter lower = do
```
fisx, line 24: why is it called "lower"?

services/spar/test/Test/Spar/Sem/ScimExternalIdStoreSpec.hs
```diff
@@ -0,0 +1,165 @@
...
+prop_lookupInsert =
+  prepropLaw @'[E.ScimExternalIdStore] $ do
+    tid <- arbitrary
+    email <- arbitrary
+    pure
+      ( do
+          E.lookup tid email >>= maybe (pure ()) (E.insert tid email),
+        do
+          pure ()
+      )
```
fisx, line 80: The value being equal is tautological, so you're after the fact here that nothing is written to the state if lookup fails?  What's the difference between your test and this one?

```suggestion
      ( E.lookup tid email,
        pure Nothing
      )
```

This question applies to (some of) the other modules as well.

services/spar/test/Test/Spar/Sem/IdPSpec.hs
```diff
@@ -9,64 +10,237 @@ import Imports
...
+  (forall a. Sem r a -> IO (f a)) ->
+  Property
+prop_storeStoreInterleave =
+  prepropLaw @'[E.IdP] $ do
+    s <- arbitrary
+    s' <- arbitrary
+    !_ <-
+      if s ^. SAML.idpId == s' ^. SAML.idpId
+        then discard
+        else pure ()
```
fisx, line 87:
```suggestion
      when (s ^. SAML.idpId == s' ^. SAML.idpId) discard
```

services/spar/test/Test/Spar/Sem/IdPSpec.hs
```diff
@@ -9,64 +10,237 @@ import Imports
...
+prop_deleteGet =
+  prepropLaw @'[E.IdP] $ do
+    s <- arbitrary
+    pure
+      ( do
+          E.deleteConfig s
+          E.getConfig $ s ^. SAML.idpId,
+        do
+          E.deleteConfig s
+          pure Nothing
```
fisx, line 127: I think this test may be pointless if you haven't been careful about the Arbitrary instance.  Does it return the same UUID twice in a few calls with non-negligible probability?

services/spar/test/Test/Spar/Sem/IdPSpec.hs
```diff
@@ -9,64 +10,237 @@ import Imports
...
+  (forall a. Sem r a -> IO (f a)) ->
+  Property
+prop_storeGetByIssuer =
+  prepropLaw @'[E.IdP] $
+    do
+      s <- arbitrary
+      pure
+        ( do
+            E.storeConfig s
+            E.getIdByIssuerWithoutTeam $ s ^. idpMetadata . edIssuer,
```
fisx, line 156: What about "with team"?  (Extract the team to pass to the getter from s.  If there are two idps with the same issuer and different team, it returns the right one.)

fisx: If you don't want to implement it, please leave this as `FUTUREWORK`.

services/spar/test/Test/Spar/Sem/IdPSpec.hs
```diff
@@ -9,64 +10,237 @@ import Imports
...
+    do
+      replaced <- arbitrary
+      replacing <- arbitrary
+      pure
+        ( do
+            E.setReplacedBy replaced replacing
+            E.clearReplacedBy replaced,
+          do
+            E.clearReplacedBy replaced
+        )
```
fisx, line 177: One more test: pull the IdPs after the fact vs. construct the changes from the initially generated values.

services/spar/test/Test/Spar/Sem/IdPSpec.hs
```diff
@@ -9,64 +10,237 @@ import Imports
...
+prop_setGet =
+  prepropLaw @'[E.IdP] $
+    do
+      replaced_id <- arbitrary
+      let replaced = E.Replaced replaced_id
+      replacing_id <- arbitrary
+      let replacing = E.Replacing replacing_id
+      pure
+        ( do
+            E.setReplacedBy replaced replacing
```
fisx, line 241: see above: store IdPs first before you manipulate them like this.

services/spar/test/Test/Spar/Sem/IdPSpec.hs
```diff
@@ -9,64 +10,237 @@ import Imports
...
+            getReplacedBy replaced_id,
+          do
+            E.setReplacedBy replaced replacing
+            (Just replacing_id <$) <$> E.getConfig replaced_id
```
fisx, line 245: ...  and I'm not sure this is making too much sense.  What are you trying to test here?

services/spar/test/Test/Spar/Sem/IdPSpec.hs
```diff
@@ -9,64 +10,237 @@ import Imports
...
+  (forall a. Sem r a -> IO (f a)) ->
+  Property
+prop_setClear =
+  prepropLaw @'[E.IdP] $
+    do
+      replaced <- arbitrary
+      replacing <- arbitrary
+      pure
+        ( do
+            E.setReplacedBy replaced replacing
```
fisx, line 173: Store IdPs first before you manipulate them like this.

services/spar/test/Test/Spar/Sem/IdPSpec.hs
```diff
@@ -9,64 +10,237 @@ import Imports
...
+  PropConstraints r f =>
+  (forall a. Sem r a -> IO (f a)) ->
+  Property
+prop_getGet =
+  prepropLaw @'[E.IdP] $
+    do
+      idpid <- arbitrary
+      pure
+        ( do
+            liftA2 (,) (E.getConfig idpid) (E.getConfig idpid),
```
fisx, line 189: Are you sure this is ever going to pull anything?  I think I've asked this question once above, so I'll stop now.  :)

fisx (changes requested):
Would be nice to run this against the cassandra implementation as well.  I would like to have the specs in `services/src/Spar/Sem/*/Spec.hs`, and have a very small integration test suite that just sticks it all together.  But this can be done in the a different PR.

But there are some tests in here that I don't understand.  Maybe it's time to pair again?

services/spar/test/Test/Spar/Sem/ScimExternalIdStoreSpec.hs
isovector: because it lowers the effect into IO and `run` was already taken :)

services/spar/test/Test/Spar/Sem/ScimExternalIdStoreSpec.hs
isovector: This test is that storing what you got is identity. It's the equivalent of this lens law `set l (view l s) s  ≡ s`. In essence it's saying there is no "junk" inside the `insert` method --- eg it's not allowed to change any other keys' values.

isovector: It's not tautological --- we're also checking equality of the _underlying stores_ after the fact.

services/spar/test/Test/Spar/Sem/IdPSpec.hs
isovector: This shows that `getReplacedBy` actually gives you back what you replaced it by, right?

services/spar/test/Test/Spar/Sem/IdPSpec.hs
isovector: You're right --- it almost certainly doesn't. What would you say are the odds of the generators respecting the size parameter?

services/spar/test/Test/Spar/Sem/IdPSpec.hs
isovector: Maybe we can pair on this one. The semantics weren't immediately clear to me.

services/spar/test/Test/Spar/Sem/IdPSpec.hs
isovector: I think that might be thinking too "domain"-ly. Really we are trying to show the equivalence of these two programs. You're right that this is unlikely to hit a `Just` case in the given implementation, but it really is the property we'd like to test. I think the solution is not to muck with this, but instead to scale the generators running the properties to ensure we get collisions. What do you think?

services/spar/test/Test/Spar/Sem/ScimExternalIdStoreSpec.hs
fisx: ah, that makes sense.  :)  *much* better than `run'`!

services/spar/test/Test/Spar/Sem/ScimExternalIdStoreSpec.hs
fisx: Ok, I did figure out your second point myself, just needed help with the first one.  Ok, now I want those all to stay, cool!

services/spar/test/Test/Spar/Sem/IdPSpec.hs
fisx: I have no idea.  :)  experiment?  I'm not sure that's the right approach, though.  We want the keys to come from a small set so that collisions are common, but not anything else.  If we didn't use quickcheck but [registry](https://hackage.haskell.org/package/registry), we could "easily" replace the underlying UUID and Issuer generators, but with quickcheck I'm not sure what to do.

(Note to self: the Issuer generator may have not enough entropy, instead of too much.  With some of the generators I've been quite lazy.)

services/spar/test/Test/Spar/Sem/IdPSpec.hs
fisx: ...  or fix the generators.  I think that's what we should do instead.

services/spar/test/Test/Spar/Sem/IdPSpec.hs
fisx: Yes, but why the `getConfig`?  Why just `pure $ Just replacing_id`?

services/spar/test/Test/Spar/Sem/IdPSpec.hs
fisx: see above.

services/spar/test/Test/Spar/Sem/IdPSpec.hs
isovector: Ah, yeah. Because this is a `Maybe (Maybe IdPId)` where the outer maybe depends on whether the thing exists in the first place.

isovector:
@fisx good catch --- coverage checking pointed out that we were hitting lots of pathological cases. I think I've the worrisome ones.

isovector:
@fisx I'd like to get this merged.

services/spar/test/Test/Spar/Sem/IdPSpec.hs
```diff
@@ -95,13 +95,10 @@ prop_storeStoreInterleave =
     s <- arbitrary
     s' <- arbitrary
     !_ <-
-      if s ^. SAML.idpId == s' ^. SAML.idpId
-        then discard
-        else pure ()
+      when (s ^. SAML.idpId == s' ^. SAML.idpId) discard
```
fisx, line 100: didn't we say this has negligible probability?

fisx (approved):

changelog.d/5-internal/in-mem-specs
```diff
@@ -0,0 +1 @@
+Added laws for DefaultSsoCode, Now, IdP and ScimExternalIdStore
```
fisx, line 1: this may be a bit confusing for the casual reader of the release notes.  i'll fix it on develop after merging.

PR #1941, opened 2021-11-24
Fix docs.

fisx:
The diff says it all!

## Checklist

### DIFF
--- a/docs/reference/provisioning/scim-via-curl.md
+++ b/docs/reference/provisioning/scim-via-curl.md
@@ -160,7 +160,7 @@ curl -X GET \
 ### get all users
 
 There is a way to do this in the SCIM protocol, but it's not
-implemented in wire for performance reasons.  If you need a complete
+implemented in wire for performance reasons. If you need a complete
 list of your team members, try the CSV download button in the team
 management app.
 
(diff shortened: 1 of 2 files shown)


docs/reference/provisioning/scim-via-curl.md
```diff
@@ -159,12 +159,10 @@ curl -X GET \
...
 ### get all users
 
-```bash
-curl -X GET \
-  --header "Authorization: Bearer $SCIM_TOKEN" \
-  --header 'Content-Type: application/json;charset=utf-8' \
-  $WIRE_BACKEND/scim/v2/Users/
-```
+There is a way to do this in the SCIM protocol, but it's not
+implemented in wire for performance reasons.  If you need a complete
```
battermann, line 163:
```suggestion
implemented in wire for performance reasons. If you need a complete
```

battermann (changes requested):

docs/reference/provisioning/scim-via-curl.md
fisx: Ok, I'm old, fine.  :)  https://www.thesaurus.com/e/writing/is-there-1-space-or-2-after-a-period/

battermann (approved):

PR #1943, opened 2021-11-24
 Activate/Deactivate Conference Calling Team Feature for Back Office [skip ci]

fisx:
## Checklist

### DIFF
--- a/libs/galley-types/src/Galley/Types/Teams.hs
+++ b/libs/galley-types/src/Galley/Types/Teams.hs
@@ -215,7 +215,7 @@ data FeatureFlags = FeatureFlags
     _flagAppLockDefaults :: !(Defaults (TeamFeatureStatus 'WithoutLockStatus 'TeamFeatureAppLock)),
     _flagClassifiedDomains :: !(TeamFeatureStatus 'WithoutLockStatus 'TeamFeatureClassifiedDomains),
     _flagFileSharing :: !(Defaults (TeamFeatureStatus 'WithoutLockStatus 'TeamFeatureFileSharing)),
-    _flagConferenceCalling :: !(Defaults (TeamFeatureStatus 'WithoutLockStatus 'TeamFeatureConferenceCalling)),
+    _flagConferenceCalling :: !(Defaults (TeamFeatureStatus 'WithLockStatus 'TeamFeatureConferenceCalling)),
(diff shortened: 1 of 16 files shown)


libs/galley-types/src/Galley/Types/Teams.hs
```diff
@@ -261,7 +261,7 @@ instance FromJSON FeatureFlags where
       <*> (fromMaybe (Defaults defaultAppLockStatus) <$> (obj .:? "appLock"))
       <*> (fromMaybe defaultClassifiedDomains <$> (obj .:? "classifiedDomains"))
       <*> (fromMaybe (Defaults (TeamFeatureStatusNoConfig TeamFeatureEnabled)) <$> (obj .:? "fileSharing"))
-      <*> (fromMaybe (Defaults (TeamFeatureStatusNoConfig TeamFeatureEnabled)) <$> (obj .:? "conferenceCalling"))
+      <*> (fromMaybe (Defaults defaultConferenceCalling) <$> (obj .:? "conferenceCalling"))
```
fisx, line 264: I think this feature needs a default config that depends on the lock status.  It is always either locked/disabled or unlocked/enabled.  No further changes are possible.  This can be accomplished with something like:

```haskell
class HasDefaultConfig (a :: TeamFeatureName) where
  defaultConfig :: LockStatusValue -> a
```

It can then be either constant and ignore the argument (like in self deleting messages), or consider it (like here).

services/galley/src/Galley/API/Teams/Features.hs
```diff
@@ -336,7 +336,7 @@ getSSOStatusInternal ::
 getSSOStatusInternal =
   either
     (const $ Public.TeamFeatureStatusNoConfig <$> getDef)
-    (getFeatureStatusNoConfig @'Public.WithoutLockStatus @'Public.TeamFeatureSSO getDef)
+    (getFeatureStatusNoConfig @'Public.TeamFeatureSSO getDef)
```
fisx, line 339: I don't understand this change.  The internal end-point has to return the lock status, the external shouldn't.  Both handlers call this function, so this function has to take an `IncludeLockStatus` as argument and then pass it along to `getFeatureStatusNoConfig`.

It's analogous to the with-config case.

battermann:
@fisx I would say that this PR can be closed. Once we pick up the work on the [ticket](https://wearezeta.atlassian.net/browse/SQSERVICES-729) again, we should rewrite it because I think parts of this has already implemented been implemented, and it should be faster than resolving all the conflicts.

PR #1944, opened 2021-11-25
Use minio helm chart from charts.min.io instead of helm.min.io

akshaymankar:
The repository at helm.min.io is down. But the one at charts.min.io is up. Unfortunately, the one at charts.min.io provides a slightly different helm chart. So, some options need to be adjusted accordingly.

## Checklist


akshaymankar:
The new helm chart doesn't retain the users on restart, we have to also create the user in the reaper.

smatting (approved):
I looked over https://github.com/minio/minio/blob/master/helm/minio/values.yaml to see what we might be missing.

PR #1948, opened 2021-11-29
SQSERVICES 1099 Public API end-point for re-sending email validation

battermann:
## Checklist

### DIFF
--- a/changelog.d/1-api-changes/brig-team-owner-can-set-user-email
+++ b/changelog.d/1-api-changes/brig-team-owner-can-set-user-email
@@ -1,1 +1,1 @@
-A new endpoint is added to Brig (`put /users/:uid/email`) that allows a team owner to change/set a user email. It can also be used to (re-) send the activation email.
+A new endpoint is added to Brig (`put /users/:uid/email`) that allows a team owner to initiate changing/setting a user email by (re-)sending an activation email.
--- a/changelog.d/1-api-changes/brig-team-owner-can-set-user-email
+++ b/changelog.d/1-api-changes/brig-team-owner-can-set-user-email
@@ -1,0 +1,1 @@
+A new endpoint is added to Brig ('put /users/:uid/email') that allows a team owner to change/set a user email. It can also be used to (re-) send the activation email.
--- a/changelog.d/1-api-changes/brig-team-owner-can-set-user-email
+++ b/changelog.d/1-api-changes/brig-team-owner-can-set-user-email
@@ -1,1 +1,1 @@
-A new endpoint is added to Brig ('put /users/:uid/email') that allows a team owner to change/set a user email. It can also be used to (re-) send the activation email.
+A new endpoint is added to Brig (`put /users/:uid/email`) that allows a team owner to change/set a user email. It can also be used to (re-) send the activation email.
--- a/libs/wire-api/src/Wire/API/Routes/Public/Brig.hs
+++ b/libs/wire-api/src/Wire/API/Routes/Public/Brig.hs
@@ -137,7 +137,7 @@ data Api routes = Api
         :> MultiVerb 'DELETE '[JSON] DeleteSelfResponses (Maybe Timeout),
     updateUserEmailValidation ::
       routes
-        :- Summary "Resend email validation email."
+        :- Summary "Resend email address validation email."
         :> Description "If the user has a pending email validation, the validation email will be resent."
         :> ZUser
         :> "users"
--- a/libs/wire-api/src/Wire/API/Routes/Public/Brig.hs
+++ b/libs/wire-api/src/Wire/API/Routes/Public/Brig.hs
@@ -135,7 +135,7 @@ data Api routes = Api
         :> "self"
         :> ReqBody '[JSON] DeleteUser
         :> MultiVerb 'DELETE '[JSON] DeleteSelfResponses (Maybe Timeout),
-    updateUserEmailValidation ::
+    updateUserEmail ::
       routes
         :- Summary "Resend email address validation email."
         :> Description "If the user has a pending email validation, the validation email will be resent."
@@ -143,6 +143,7 @@ data Api routes = Api
         :> "users"
         :> CaptureUserId "uid"
         :> "email"
+        :> ReqBody '[JSON] EmailUpdate
         :> Put '[JSON] (),
     getHandleInfoUnqualified ::
       routes
--- a/libs/wire-api/src/Wire/API/Routes/Public/Brig.hs
+++ b/libs/wire-api/src/Wire/API/Routes/Public/Brig.hs
@@ -135,6 +135,15 @@ data Api routes = Api
         :> "self"
         :> ReqBody '[JSON] DeleteUser
         :> MultiVerb 'DELETE '[JSON] DeleteSelfResponses (Maybe Timeout),
+    updateUserEmailValidation ::
+      routes
+        :- Summary "Resend email validation email."
+        :> Description "If the user has a pending email validation, the validation email will be resent."
+        :> ZUser
+        :> "users"
+        :> CaptureUserId "uid"
+        :> "email"
+        :> Put '[JSON] (),
     getHandleInfoUnqualified ::
       routes
         :- Summary "(deprecated, use /search/contacts) Get information on a user handle"
--- a/services/brig/src/Brig/API/Public.hs
+++ b/services/brig/src/Brig/API/Public.hs
@@ -1199,7 +1199,8 @@ verifyDeleteUserH (r ::: _) = do
 updateUserEmailValidation :: UserId -> UserId -> Handler ()
 updateUserEmailValidation zuserId emailOwnerId = do
   maybeZuserTeamId <- lift $ Data.lookupUserTeam zuserId
-  checkPerm maybeZuserTeamId
+  whenM (isNothing <$> runMaybeT (checkPerm maybeZuserTeamId)) $
+    throwStd insufficientTeamPermissions
   maybeEmailOwnerTeamId <- lift $ Data.lookupUserTeam emailOwnerId
   checkSameTeam maybeZuserTeamId maybeEmailOwnerTeamId
   maybeEmail <- lift $ Data.lookupUserUnverifiedEmail emailOwnerId
@@ -1212,18 +1213,13 @@ updateUserEmailValidation zuserId emailOwnerId = do
       when (Just zuserTeamId /= maybeEmailOwnerTeamId) $ throwStd insufficientTeamPermissions
     checkSameTeam Nothing _ = throwStd insufficientTeamPermissions
 
-    checkPerm :: Maybe TeamId -> Handler ()
-    checkPerm =
-      \case
-        Just teamId -> do
-          maybeTeamMember <- lift $ Intra.getTeamMember zuserId teamId
-          case maybeTeamMember of
-            Just teamMember ->
-              if hasPermission teamMember ChangeTeamMemberProfiles
-                then pure ()
-                else throwStd insufficientTeamPermissions
-            Nothing -> throwStd insufficientTeamPermissions
-        Nothing -> throwStd insufficientTeamPermissions
+    checkPerm :: Maybe TeamId -> MaybeT Handler ()
+    checkPerm maybeTeamId = do
+      teamId <- hoistMaybe maybeTeamId
+      teamMember <- MaybeT $ lift $ Intra.getTeamMember zuserId teamId
+      if teamMember `hasPermission` ChangeTeamMemberProfiles
+        then pure ()
+        else throwStd insufficientTeamPermissions
 
 -- activation
 
--- a/services/brig/src/Brig/API/Public.hs
+++ b/services/brig/src/Brig/API/Public.hs
@@ -1199,7 +1199,7 @@ verifyDeleteUserH (r ::: _) = do
 updateUserEmail :: UserId -> UserId -> Public.EmailUpdate -> Handler ()
 updateUserEmail zuserId emailOwnerId (Public.EmailUpdate email) = do
   maybeZuserTeamId <- lift $ Data.lookupUserTeam zuserId
-  whenM (not . fromMaybe False <$> runMaybeT (checkPerm maybeZuserTeamId)) $ throwStd insufficientTeamPermissions
+  whenM (not <$> assertHasPerm maybeZuserTeamId) $ throwStd insufficientTeamPermissions
   maybeEmailOwnerTeamId <- lift $ Data.lookupUserTeam emailOwnerId
   checkSameTeam maybeZuserTeamId maybeEmailOwnerTeamId
   void $ API.changeSelfEmail emailOwnerId email API.AllowSCIMUpdates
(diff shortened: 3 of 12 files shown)


fisx (changes requested):
Good progress!  :+1:

libs/wire-api/src/Wire/API/Routes/Public/Brig.hs
```diff
@@ -135,6 +135,15 @@ data Api routes = Api
         :> "self"
         :> ReqBody '[JSON] DeleteUser
         :> MultiVerb 'DELETE '[JSON] DeleteSelfResponses (Maybe Timeout),
+    updateUserEmailValidation ::
+      routes
+        :- Summary "Resend email validation email."
+        :> Description "If the user has a pending email validation, the validation email will be resent."
+        :> ZUser
+        :> "users"
+        :> CaptureUserId "uid"
```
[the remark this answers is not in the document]
pcapriotti: > In my limited understanding of the subject, this will force you to handle the qualifying domain in which the user lives.  In this end-point, we can throw an error "only supported for local domain" here.

That's correct, and of course it also makes the API require a domain. In galley there is an `ensureLocal` helper which throws `federationNotImplemented` if the resulting qualified value is not local, and otherwise converts it into a `Local x` value (i.e. a qualified value with a guarantee that its domain is local).

fisx:
(ci failure seems unrelated)

libs/wire-api/src/Wire/API/Routes/Public/Brig.hs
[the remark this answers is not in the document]
fisx: Misunderstanding: we don't want to implement the federated version later, we want to make sure that this can only be called from local.  Does that mean keep the code as it is?

(Your plan would be better though if we plan to have federated teams, where some members are on other instances than others.  But I don't think that's a plan, right?)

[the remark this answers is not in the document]
pcapriotti: Ah, sorry, I haven't actually looked at what this endpoint is about. :) If it doesn't make sense for remote values, then I guess you shouldn't use the qualified version. You're right, federated teams are not a thing for the time being.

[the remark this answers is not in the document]
battermann: @fisx please resolve this conversation if I can leave it like this.

services/brig/src/Brig/API/Public.hs
```diff
@@ -1193,6 +1197,24 @@ verifyDeleteUserH (r ::: _) = do
...
+    -- TODO(leif): check if we should use AllowSCIMUpdates or ForbidSCIMUpdates
+    Just email -> void $ API.changeSelfEmail emailOwnerId email API.AllowSCIMUpdates
+    Nothing -> throwStd $ notFound "pending validation email of email owner not found"
+  where
+    checkPermissions zuser emailOwner =
+      void $ case (Public.userTeam zuser, Public.userTeam emailOwner) of
+        (Just zuserTeamId, Just emailOwnerTeamId) -> do
+          when (zuserTeamId /= emailOwnerTeamId) $ throwStd insufficientTeamPermissions
+          ensureHiddenPermissions zuserId zuserTeamId [Team.ChangeTeamMemberProfiles]
+        (_, _) -> throwStd insufficientTeamPermissions
```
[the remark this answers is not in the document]
battermann: IMO returning insufficient team permissions error does not reveal anything about the existence of the user, does it?

battermann:
TODO: Integration tests.

services/brig/src/Brig/API/Public.hs
```diff
@@ -1193,6 +1196,31 @@ verifyDeleteUserH (r ::: _) = do
   API.verifyDeleteUser body !>> deleteUserError
   return (setStatus status200 empty)
 
+updateUserEmailValidation :: UserId -> UserId -> Handler ()
+updateUserEmailValidation zuserId emailOwnerId = do
+  maybeZuserTeamId <- lift $ Data.lookupUserTeam zuserId
```
battermann, line 1201: Maybe it is even cleaner, if we put the whole function into `MaybeT`?

services/brig/src/Brig/API/Public.hs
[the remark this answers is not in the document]
fisx: Yes, but that wasn't my point.

There are two changes, one is cosmetic (mbEmailOwnerTeamId instead of emailOwnerTeamId), the other one important: `_` should contain an error that results in a response indistinguishable from success.

The `ensureHiddenPermissions` is fine.

services/brig/src/Brig/API/Public.hs
fisx: let's pair today!

services/brig/src/Brig/Data/User.hs
```diff
@@ -404,9 +406,18 @@ lookupRichInfoMultiUsers users = do
...
 lookupUserTeam u =
-  join . fmap runIdentity
+  (runIdentity =<<)
     <$> retry x1 (query1 teamSelect (params LocalQuorum (Identity u)))
 
+lookupUserUnverifiedEmail :: MonadClient m => UserId -> m (Maybe Email)
+lookupUserUnverifiedEmail userId =
+  (runIdentity =<<) <$> retry x1 (query1 cql (params LocalQuorum (Identity $ toUUID userId)))
+  where
+    cql :: PrepQuery R (Identity UUID) (Identity (Maybe Email))
```
fisx, line 416: I think you could still use `UserId` here, but it's not important, UUID is also fine.

services/brig/src/Brig/Data/User.hs
```diff
@@ -404,9 +406,18 @@ lookupRichInfoMultiUsers users = do
...
+    cql =
+      "SELECT email \
+      \FROM vcodes WHERE account = ?"
```
fisx, line 419: ah.  problem.  we can't use `account` in the where clause because it's not a key.  two solutions:

- add another database table
- make clients send the email after all, but look it up in vcode and assert that it matches on the right userid.

i strongly prefer the second.  easier on server load, easier to implement, more consistent with the other validation end-points no real downsides.

sorry, should have seen that earlier.  :(

services/brig/test/integration/API/User/Account.hs
```diff
@@ -141,9 +141,59 @@ tests _ at opts p b c ch g aws =
...
         "temporary customer extensions"
         [ test' aws p "domains blocked for registration" $ testDomainsBlockedForRegistration opts b
+        ],
+      testGroup
+        "re-send activation email"
+        [ test' aws p "put /users/:uid/email" $ testReSendActivationEmail b
         ]
     ]
 
+testReSendActivationEmail :: Brig -> Http ()
```
fisx, line 151: It's not really *re*send any more, though, right?  That's just the use case we're doing this for now, but "send" and "re-send" are indistinguishable.  Do you think we should rename it?  Is there "re-" anywhere in the production code?  (If not I don't care all that much.)

services/brig/test/integration/API/User/Account.hs
```diff
@@ -141,9 +141,59 @@ tests _ at opts p b c ch g aws =
...
+testReSendActivationEmail brig = do
+  (_, teamOwner, emailOwner : otherTeamMember : _) <- createPopulatedBindingTeamWithNamesAndHandles brig 2
```
fisx, line 153: People have told me I'm a java developer at heart for coming up with this name.  :)

("Unrelated" question: do you enjoy non-actionable comments in reviews, or should I avoid them?)

services/brig/test/integration/API/User/Account.hs
```diff
@@ -141,9 +141,59 @@ tests _ at opts p b c ch g aws =
...
+  checkLetActivationExpire newEmail
+  checkActivationCode newEmail False
+  checkSetUserEmail teamOwner emailOwner newEmail 200
+  checkActivationCode newEmail False
+  checkUnauthorizedRequests emailOwner otherTeamMember teamOwnerDifferentTeam newEmail
+  checkActivationCode newEmail False
+  where
+    checkLetActivationExpire :: Email -> Http ()
+    checkLetActivationExpire email = do
+      -- assumption: email activation timeout is set to 5 sec, so we wait until it expires
```
fisx, line 177:
```suggestion
      -- assumption: `optSettings.setActivationTimeout = 5` in `brig.yaml`
```

services/brig/test/integration/API/User/Account.hs
```diff
@@ -141,9 +141,59 @@ tests _ at opts p b c ch g aws =
...
+      threadDelay (5100 * 1000)
+      checkActivationCode email False
+
+    checkActivationCode :: Email -> Bool -> Http ()
+    checkActivationCode email shouldExist = do
+      maybeActivationCode <- Util.getActivationCode brig (Left email)
+      if shouldExist
+        then void $ lift $ assertBool "activation code should exists" (isJust maybeActivationCode)
+        else void $ lift $ assertBool "activation code should not exists" (isNothing maybeActivationCode)
```
fisx, line 186:
```suggestion
      void $ lift $ if shouldExist
        then assertBool "activation code should exists" (isJust maybeActivationCode)
        else assertBool "activation code should not exists" (isNothing maybeActivationCode)
```

services/brig/test/integration/API/User/Account.hs
```diff
@@ -141,9 +141,59 @@ tests _ at opts p b c ch g aws =
...
+  checkLetActivationExpire newEmail
+  checkSetUserEmail teamOwner emailOwner newEmail 200
+  checkActivationCode newEmail True
+  checkLetActivationExpire newEmail
+  checkUnauthorizedRequests emailOwner otherTeamMember teamOwnerDifferentTeam newEmail
+  checkActivationCode newEmail False
+  checkSetUserEmail teamOwner emailOwner newEmail 200
+  checkActivationCode newEmail True
+  activateEmail brig newEmail
+  -- apparently activating an email does not invalidate the activation code
```
fisx, line 167: ah, interesting.  i haven't checked the code, but this makes sense as deletion in cassandra is expensive, the vcode entry will evaporate once it expires, and there is no harm in re-activating an already activated email address (idempotency).

services/brig/test/integration/API/User/Account.hs
```diff
@@ -141,9 +141,59 @@ tests _ at opts p b c ch g aws =
...
+  checkLetActivationExpire newEmail
+  checkActivationCode newEmail False
```
fisx, line 169:
```suggestion
  checkActivationCode newEmail True
```

services/brig/test/integration/API/User/Account.hs
fisx: Can you rewrite this test so that "letexpire" is only called once?  It's probably move expensive than the test put together by about an order of magnitude.

fisx (commented):
All but 3 of these comments are just because it's late and I'm chatty, hope that's not too annoying.  I think you're close!  Let's see if concourse agrees.

fisx:
Ah, and can you try and complete the checkbox list?  In the next pairing session we can go through the things you'd prefer to do together before anything else.

services/brig/src/Brig/Data/User.hs
```diff
@@ -471,7 +471,7 @@ lookupFeatureConferenceCalling uid = do
   pure $ ApiFt.TeamFeatureStatusNoConfig <$> mStatusValue
   where
     select :: PrepQuery R (Identity UserId) (Identity (Maybe ApiFt.TeamFeatureStatusValue))
-    select = fromString $ "select feature_conference_calling from user where id = ?"
+    select = fromString "select feature_conference_calling from user where id = ?"
```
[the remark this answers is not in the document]
battermann: I know, but this makes the barrier for improving the code on the fly much higher.

changelog.d/1-api-changes/brig-team-owner-can-set-user-email
```diff
@@ -0,0 +1 @@
+A new endpoint is added to Brig (`put /users/:uid/email`) that allows a team owner to change/set a user email. It can also be used to (re-) send the activation email.
```
fisx, line 1:
```suggestion
A new endpoint is added to Brig (`put /users/:uid/email`) that allows a team owner to initiate changing/setting a user email by (re-)sending an activation email.
```

services/brig/test/integration/API/User/Account.hs
```diff
@@ -141,9 +141,57 @@ tests _ at opts p b c ch g aws =
...
+  checkSetUserEmail teamOwner emailOwner newEmail 200
+  checkActivationCode newEmail True
+  checkUnauthorizedRequests emailOwner otherTeamMember teamOwnerDifferentTeam newEmail
+  activateEmail brig newEmail
+  -- apparently activating the email does not invalidate the activation code
+  -- ideally we would let the activation code expire again (which is too expensive time-wise)
+  -- and then assert that if the email is verified there is no new activation code created
+  -- when the set email function is called again (idempotency)
```
fisx, line 167:
```suggestion
  -- when the set email function is called again
```

It's not idempotency if the behavior changes depending on the state of the backend, right?  (One time it does send an email, the next time it doesn't, because the first one has been answered and the user activated?)

services/brig/test/integration/API/User/Account.hs
```diff
@@ -141,9 +141,57 @@ tests _ at opts p b c ch g aws =
...
+  checkSetUserEmail teamOwner emailOwner newEmail 200
+  checkUnauthorizedRequests emailOwner otherTeamMember teamOwnerDifferentTeam newEmail
```
fisx, line 169: Can we remove this line?  ff not, can you add a comment helping me to find out what this is testing?

services/brig/test/integration/API/User/Account.hs
```diff
@@ -141,9 +141,57 @@ tests _ at opts p b c ch g aws =
...
+  checkActivationCode newEmail False
+  checkSetUserEmail teamOwner emailOwner newEmail 200
+  checkActivationCode newEmail True
+  checkUnauthorizedRequests emailOwner otherTeamMember teamOwnerDifferentTeam newEmail
+  activateEmail brig newEmail
+  -- apparently activating the email does not invalidate the activation code
+  -- ideally we would let the activation code expire again (which is too expensive time-wise)
+  -- and then assert that if the email is verified there is no new activation code created
+  -- when the set email function is called again (idempotency)
+  checkSetUserEmail teamOwner emailOwner newEmail 200
```
fisx, line 168: sorry, changed my mind again, i think you were right before :(

```suggestion
  checkLetActivationExpire newEmail
  checkSetUserEmail teamOwner emailOwner newEmail 200
```

if you've had enough just ignore me, but without this line you might as well remove line 163ff altogether, right?

fisx (approved):
LGTM!  I have two more change requests, but neither is essential, so I've (pre-)approved this PR and leave you with deciding if you think you should follow up on them.

services/brig/test/integration/API/User/Account.hs
battermann: I am absolutely ok with removing this. But I think it is idempotent because it doesn't change the state anymore once the email was verified by the email owner, regardless of how many times this is triggered again.

services/brig/test/integration/API/User/Account.hs
battermann: Maybe these are not the most useful tests, but they don't hurt.

services/brig/test/integration/API/User/Account.hs
battermann: I feel indifferent about both approaches. I would suggest to keep it like it is and extend the comment, if that is ok with you?

services/brig/test/integration/API/User/Account.hs
battermann: As discussed in another conversation, I added the expiration of the token again, so this is obsolete, I think.

fisx (approved):

fisx:
```
#9 1351.2 Copied executables to /wire-server/dist:
#9 1351.2 - api-loadtest
#9 1351.2 - api-smoketest
[...]
#9 1351.2 - stern
#9 1351.2 - zauth
#9 1351.2
#9 1351.2 Warning: Installation path /wire-server/dist not found on the PATH environment variable.
#9 DONE 1360.5s
[...]
#13 exporting to oci image format
#13 sha256:69a2560eef4d3ece902c3b5149d142e9bd132f25db0bc9e35b94201534c415d2
#13 exporting layers
#13 exporting layers 92.7s done
#13 exporting manifest sha256:a1159bd4b920fd682b9ccef8f31c9c69b7632a8d2d0f18dec909c0b736e5adff 0.0s done
#13 exporting config sha256:6440c8a581fb380640fd6ad9d9c4cc058f4b20d7a92c4dd56afad7f41897b1fa 0.0s done
#13 ERROR: no active session for 7hrhca6ne62i9uyj4ksy826la: context deadline exceeded
------
 > exporting to oci image format:
------
error: failed to solve: rpc error: code = DeadlineExceeded desc = no active session for 7hrhca6ne62i9uyj4ksy826la: context deadline exceeded
FATA[1653] failed to build: build: exit status 1        
FATA[1653] failed to run task: exit status 1            
```

not sure where this rpc error is coming from, but probably not from a compilation error?

PR #1951, opened 2021-11-30
Add new LogFormat: StructuredJSON

akshaymankar:
This format is intended as an imporvement over the JSON format. It tries to
improve in these ways:

- Explicitly encode log level as a key in the main JSON object logged

- Encode field values as key-value pairs in the main JSON object

Encoding logs like this makes it easy for external tools like fluent-bit and
elasticsearch to understand the log level and the fields and index them as such
without requiring bonanza to be present while processing logs.

https://wearezeta.atlassian.net/browse/FS-62

## Checklist

### DIFF
--- a/libs/extended/src/System/Logger/Extended.hs
+++ b/libs/extended/src/System/Logger/Extended.hs
@@ -33,10 +33,11 @@ where
 
 import Cassandra (MonadClient)
 import Control.Monad.Catch
-import Data.Aeson
+import Data.Aeson as Aeson
 import Data.Aeson.Encoding (list, pair, text)
 import qualified Data.ByteString.Lazy.Builder as B
 import qualified Data.ByteString.Lazy.Char8 as L
+import qualified Data.Map.Lazy as Map
 import Data.String.Conversions (cs)
 import GHC.Generics
 import Imports
@@ -50,7 +51,7 @@ instance FromJSON LC.Level
 instance ToJSON LC.Level
 
 -- | The log formats supported
-data LogFormat = JSON | Plain | Netstring
+data LogFormat = JSON | Plain | Netstring | StructuredJSON
   deriving stock (Eq, Show, Generic)
   deriving anyclass (ToJSON, FromJSON)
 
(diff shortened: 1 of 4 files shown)


jschaul (approved):
This just adds the new log format, right? But doesn't turn it on anywhere yet - how/where is it turned on to log that way?

The first part, the log level - could that not also be a separate structural element rather than the first item in the list of msgs? Or do you configure fluent-bit to parse the log level as first element?

akshaymankar:
> This just adds the new log format, right? But doesn't turn it on anywhere yet - how/where is it turned on to log that way?

Yes, it only adds a new log format. I will turn it on in our federation environments as a test. If all looks fine after a month or so, I will propose that we make it the default.

> The first part, the log level - could that not also be a separate structural element rather than the first item in the list of msgs? Or do you configure fluent-bit to parse the log level as first element?

The way tinylog works, it sends all the logs as `[Element]` to the renderer, this does contain the log level encoded as 1 character string (`I`, `D`, etc.). But it also sends the log level separately. Instead of relying on this contract of parsing the first 1 character string, I decided  to just ignore it and put the log level separately sent by tiny log into the final JSON as `level`. I will comment inline to show where. This ensures that no configuration on fleunt-bit is required.

libs/extended/src/System/Logger/Extended.hs
```diff
@@ -77,6 +78,43 @@ collect = foldr go (Element' mempty [])
...
+--
+-- >>> logElems = [Bytes "I", Bytes "The message", Field "field1" "val1", Field "field2" "val2", Field "field1" "val1.1"]
+-- >>> B.toLazyByteString $ structuredJSONRenderer "," iso8601UTC Info logElems
+-- "{\"msgs\":[\"I\",\"The message\"],\"field1\":[\"val1\",\"val1.1\"],\"field2\":\"val2\",\"level\":\"Info\"}"
+structuredJSONRenderer :: Renderer
+structuredJSONRenderer _sep _dateFmt lvl logElems =
+  let structuredJSON = toStructuredJSONOutput logElems
+   in fromEncoding . toEncoding $
+        object
+          ( [ "level" Aeson..= lvl,
```
akshaymankar, line 94: @jschaul This line adds the level to the final JSON.

jschaul: Cool, thanks!

akshaymankar:
Turns out [this doc](https://gitlab.com/twittner/tinylog/-/blob/7e980a4fc2c3cf419e3361e4fd8c3123202f0310/src/System/Logger/Settings.hs#L185-186) is a lie. The actual `LogLevel` passed here is just the configured setting, so all logs are currently being marked whatever log level is configured for the service :facepalm: 
The only way to get the log level of the event is to parse it from the list of messages. I will create a follow up PR.

akshaymankar:
I also created an issue upstream: https://gitlab.com/twittner/tinylog/-/issues/5

PR #1952, opened 2021-11-30
Upgrade kibana and fluent-bit helm charts

akshaymankar:
This change removes dependency on the deprecated https://charts.helm.sh repository for kibana and fluent-bit. As a result we get v6.8.18 of kibana. Kibana 6.8.x requires elasticsearch 6.8.x, so this change also upgrades the default version of elasticsearch to 6.8.18 when installing with the elasticsearch-ephemeral chart.  

https://wearezeta.atlassian.net/browse/FS-62

## Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,13 +8,13 @@ DOCKER_TAG            ?= $(USER)
 # default helm chart version must be 0.0.42 for local development (because 42 is the answer to the universe and everything)
 HELM_SEMVER           ?= 0.0.42
 # The list of helm charts needed for integration tests on kubernetes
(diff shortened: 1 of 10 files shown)


jschaul (approved):
Looking at the code, the changeset looks good, (though I didn't try it out yet).

akshaymankar:
> 

I tried it in adhoc-testing and kind, worked ok. Only manual thing I needed to do was to select which index-pattern is default. Kibana API is a bit clunky, so I decided to not research this too much. Kibana almost forces you to select one on first use anyway, so I thought it is ok.

PR #1937, opened 2021-11-23
(Un-)lock Status for Self-Deleting Messages

battermann:
## Checklist

 - [x] The **PR Title** explains the impact of the change.
 - [x] The **PR description** provides context as to why the change should occur and what the code contributes to that effect. This could also be a link to a JIRA ticket or a Github issue, if there is one.
 - [x] If HTTP endpoint paths have been added or renamed, the **endpoint / config-flag checklist** (see Wire-employee only backend [wiki page](https://github.com/zinfra/backend-wiki/wiki/Checklists)) has been followed.
 - [x] If a cassandra schema migration has been added, I ran **`make git-add-cassandra-schema`** to update the cassandra schema documentation.
 - [x] **changelog.d** contains the following bits of information ([details](https://github.com/wireapp/wire-server/blob/develop/docs/developer/changelog.md)):
   - [x] A file with the changelog entry in one or more suitable sub-sections. The sub-sections are marked by directories inside `changelog.d`.
   - [x] If new config options introduced: added usage description under docs/reference/config-options.md
   - [x] If new config options introduced: recommended measures to be taken by on-premise instance operators.
   - [x] If a cassandra schema migration is backwards incompatible (see also [these docs](https://github.com/wireapp/wire-server/blob/develop/docs/developer/cassandra-interaction.md#cassandra-schema-migrations)), measures to be taken by on-premise instance operators are explained.
   - [x] If a data migration (not schema migration) introduced: measures to be taken by on-premise instance operators.
   - [x] If public end-points have been changed or added: does nginz need un upgrade?
   - [x] If internal end-points have been added or changed: which services have to be deployed in a specific order?

PR #1963, opened 2021-12-03
(Un-)lock Status for Self-Deleting Messages (#1937 re-run)

fisx:
https://wearezeta.atlassian.net/browse/SQSERVICES-1079

## Checklist


charts/nginz/values.yaml
```diff
@@ -363,6 +363,11 @@ nginx_conf:
       - staging
       disable_zauth: true
       basic_auth: true
+    - path: ~* /i/teams/([^/]*)/features/([^/]*)/(un)?locked$
+      envs:
+      - staging
+      disable_zauth: true
+      basic_auth: true
```
fisx, line 370:
```suggestion
```

The `basic_auth` hack only works on staging (hegemony).  It's not used in the k8s environment.  (There was no way for you to know this, but it may actually be pretty bad to add this here, since the `basic_auth` will (probably?) be ignored, and that means that the end-point will be publicly accessible.)

fisx (commented):
One last change, then this is good to go!  You need to formally approve, but you can merge right after.

battermann (approved):

PR #1966, opened 2021-12-06
Streaming federator

pcapriotti:
This PR is about allowing federator and federator client to stream responses to their clients. Instead of using explicit CPS everywhere, this is achieved by replacing `IO` with `Codensity IO` in the interpreters for the `Service` and `Remote` effects. The streaming response type is `StreamingResponse` from `Servant`, even when using the low-level HTTP2 client API (`withHTTP2Request`).

The underlying `SourceT` type is slightly more complicated than necessary for our purposes (e.g. we do not need its `Error` and `Skip` constructors), but it is convenient not to have to define our own streaming abstraction.

Tracked by https://wearezeta.atlassian.net/browse/FS-322.

## Checklist


services/federator/src/Federator/Service.hs
```diff
@@ -38,9 +38,11 @@ import Wire.API.Federation.Domain (originDomainHeaderName)
 newtype ServiceError = ServiceErrorInvalidStatus HTTP.Status
   deriving (Eq, Show)
 
-data Service m a where
+type ServiceLBS = Service (Maybe LByteString)
+
+data Service body m a where
   -- | Returns status and body, 'HTTP.Response' is not nice to work with in tests
-  ServiceCall :: Component -> ByteString -> LByteString -> Domain -> Service m (HTTP.Status, Maybe LByteString)
+  ServiceCall :: Component -> ByteString -> LByteString -> Domain -> Service body m (HTTP.Status, body)
```
mdimjasevic: Why is the additional body argument needed for the `Service` type constructor? In the definition of `ServiceCall`, couldn't you have said it is of type `Service m (HTTP.Status, a)`?

services/federator/test/unit/Test/Federator/ExternalServer.hs
```diff
@@ -14,7 +14,7 @@
 --
 -- You should have received a copy of the GNU Affero General Public License along
 -- with this program. If not, see <https://www.gnu.org/licenses/>.
-{-# OPTIONS_GHC -Wno-orphans #-}
+{-# OPTIONS_GHC -Wno-orphans -fprint-potential-instances #-}
```
mdimjasevic, line 17: Is this a leftover?

services/federator/src/Federator/Service.hs
pcapriotti: Because otherwise any interpreter should be able to return an `a` as a body for all possible types `a`, which clearly won't work. The type parameter here is not that important, since the effect is currently only used with `a = StreamingResponse`.

services/federator/test/unit/Test/Federator/Client.hs
```diff
@@ -181,6 +214,41 @@ testResponseHeaders = do
...
+      responseStatusCode resp @?= HTTP.status200
+      lookup "X-Foo" (toList (responseHeaders resp)) @?= Just "bar"
+
+testStreaming :: IO ()
+testStreaming = withInfiniteMockServer $ \port -> do
+  let req = HTTP2.requestBuilder HTTP.methodPost "test" [] mempty
+  withHTTP2Request Nothing req "127.0.0.1" port $ \resp -> do
+    let expected = mconcat (replicate 512 "Hello\n")
+    actual <- takeSourceT (fromIntegral (LBS.length expected)) (responseBody resp)
+    putStrLn "go!"
```
mdimjasevic, line 227: I suppose this print statement is a leftover from debugging.

services/federator/test/unit/Test/Federator/ExternalServer.hs
```diff
@@ -72,59 +71,79 @@ exampleRequest certFile path = do
...
+data Call = Call
+  { cComponent :: Component,
+    cPath :: ByteString,
+    cBody :: LByteString,
+    cDomain :: Domain
+  }
+  deriving (Eq, Show)
+
+mockService ::
+  Members [Output Call, Embed IO] r =>
```
mdimjasevic, line 83: How come this works without using a value-level list? I was expecting a type-level list here.

mdimjasevic (approved):

services/federator/test/unit/Test/Federator/Client.hs
pcapriotti: Yes, thanks for catching this.

services/federator/test/unit/Test/Federator/ExternalServer.hs
pcapriotti: It's not a value level list. The promotion tick is optional when the type is unambiguous. For the case of lists, this means that `[A, B]` is interpreted as something of kind `[*]` (i.e. the same as `'[A, B]`, aka `A ': B ': '[]`), but `[A]` is interpreted as something of type `*` (i.e. the type of lists of `A`).

We can turn on `-Wunticked-promoted-constructors` to turn this into an error, if desired. Should we?

PR #1973, opened 2021-12-07
Add federation component for Cargohold

pcapriotti:
This PR simply adds Cargohold as a new federated component in Federator, and a federated API stub in Cargohold itself.

**Note**: this contains the first few commits from #1966.

## Checklist

 - [x] The **PR Title** explains the impact of the change.
 - [x] The **PR description** provides context as to why the change should occur and what the code contributes to that effect. This could also be a link to a JIRA ticket or a Github issue, if there is one.
 - [x] If HTTP endpoint paths have been added or renamed, the **endpoint / config-flag checklist** (see Wire-employee only backend [wiki page](https://github.com/zinfra/backend-wiki/wiki/Checklists)) has been followed.
 - [x] **changelog.d** contains the following bits of information ([details](https://github.com/wireapp/wire-server/blob/develop/docs/developer/changelog.md)):
   - [x] A file with the changelog entry in one or more suitable sub-sections. The sub-sections are marked by directories inside `changelog.d`.

PR #1964, opened 2021-12-03
SQSERVICES-547 add feature config for guest links

battermann:
Implementation for https://wearezeta.atlassian.net/browse/SQSERVICES-1045 and https://wearezeta.atlassian.net/browse/SQSERVICES-1046 for [Epic](https://wearezeta.atlassian.net/browse/SQSERVICES-547)

## Checklist

 - [x] The **PR Title** explains the impact of the change.
 - [x] The **PR description** provides context as to why the change should occur and what the code contributes to that effect. This could also be a link to a JIRA ticket or a Github issue, if there is one.
 - [x] If HTTP endpoint paths have been added or renamed, the **endpoint / config-flag checklist** (see Wire-employee only backend [wiki page](https://github.com/zinfra/backend-wiki/wiki/Checklists)) has been followed.
 - [x] If a cassandra schema migration has been added, I ran **`make git-add-cassandra-schema`** to update the cassandra schema documentation.
 - [x] **changelog.d** contains the following bits of information ([details](https://github.com/wireapp/wire-server/blob/develop/docs/developer/changelog.md)):
   - [x] A file with the changelog entry in one or more suitable sub-sections. The sub-sections are marked by directories inside `changelog.d`.
   - [x] If new config options introduced: added usage description under docs/reference/config-options.md
   - [x] If new config options introduced: recommended measures to be taken by on-premise instance operators.
   - [x] If a cassandra schema migration is backwards incompatible (see also [these docs](https://github.com/wireapp/wire-server/blob/develop/docs/developer/cassandra-interaction.md#cassandra-schema-migrations)), measures to be taken by on-premise instance operators are explained.
   - [x] If a data migration (not schema migration) introduced: measures to be taken by on-premise instance operators.
   - [x] If public end-points have been changed or added: does nginz need un upgrade?
   - [x] If internal end-points have been added or changed: which services have to be deployed in a specific order?

PR #1975, opened 2021-12-08
Fix brig doc module generation

pcapriotti:
Follow-up to #1956, fixing issues reported by @fisx:

 - remove obsolete data-files stanza from brig's cabal file;
 - generate doc modules on repl invocation, as well as on build.

No CHANGELOG entry.

## Checklist

### DIFF
--- a/services/brig/brig.cabal
+++ b/services/brig/brig.cabal
@@ -1,13 +1,13 @@
-cabal-version: 1.24
+cabal-version: 2.0
 
 -- This file has been generated from package.yaml by hpack version 0.33.0.
 --
 -- see: https://github.com/sol/hpack
 --
--- hash: c37f84d005641650921717e1f340e18d82edd30045196f89649cefd53e549dfb
+-- hash: 513c0f5104342fb14b0246f7c44733a84ec36fae97633fc55cb209e0e0bcd087
 
 name:           brig
-version:        1.35.0
+version:        2.0
 synopsis:       User Service
 category:       Network
 author:         Wire Swiss GmbH
@@ -16,6 +16,8 @@ copyright:      (c) 2017 Wire Swiss GmbH
 license:        AGPL-3
 license-file:   LICENSE
 build-type:     Custom
+extra-source-files:
+    docs/swagger.md
 
 custom-setup
   setup-depends:
@@ -119,6 +121,8 @@ library
       Main
   other-modules:
       Paths_brig
+      Brig.Docs.Swagger
+  autogen-modules:
       Brig.Docs.Swagger
   hs-source-dirs:
       src
--- a/changelog.d/4-docs/update-fed-error-docs
+++ b/changelog.d/4-docs/update-fed-error-docs
@@ -1,1 +1,1 @@
-Update federation error documentation after changes to the federation API (#1956, #1978)
+Update federation error documentation after changes to the federation API (#1956, #1975, #1978)
--- a/services/brig/Setup.hs
+++ b/services/brig/Setup.hs
@@ -22,6 +22,7 @@ import Data.Maybe
 import Distribution.Simple
 import Distribution.Simple.BuildPaths
 import Distribution.Simple.LocalBuildInfo
+import Distribution.Types.PackageDescription
 import System.Directory
 import System.FilePath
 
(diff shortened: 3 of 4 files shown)


services/brig/brig.cabal
```diff
@@ -16,8 +16,6 @@ copyright:      (c) 2017 Wire Swiss GmbH
 license:        AGPL-3
 license-file:   LICENSE
 build-type:     Custom
-data-files:
-    docs/swagger.md
```
fisx, line 20: don't files that are required for compilation also count as data files?

fisx, line 20: and why did you modify the cabal file instead of package.yaml?

fisx:
> No CHANGELOG entry.

You could manually add the two PRs to the one changelog entry you probably (hopefully?) made about the change you're fixing here?

fisx (approved):
I still get the same error, but I'm not sure how to reproduce it without a 200 year old emacs haskell-mode and a self-written ghci wrapper script.  `stack repl` works fine.

But...  how?!  And why did having the data file in the cabal file break things?

fisx:
concourse failed for the same reason https://github.com/wireapp/wire-server/pull/1964 failed earlier today:

```
  test/integration/Test/Federator/InwardSpec.hs:77:9: 
  1) Federator.API.Inward should be able to call cargohold
       uncaught exception: ErrorCall
       Assertions failed:
        1: 403 =/= 500
       
       Response was:
       
       Response {responseStatus = Status {statusCode = 500, statusMessage = "Internal Server Error"}, responseVersion = HTTP/1.1, responseHeaders = [("Transfer-Encoding","chunked"),("Date","Wed, 08 Dec 2021 13:56:38 GMT"),("Server","Warp/3.3.17"),("Content-Type","application/json")], responseBody = Just "{\"code\":500,\"message\":\"Federation is not yet implemented for this endpoint\",\"label\":\"federation-not-implemented\"}", responseCookieJar = CJ {expose = []}, responseClose' = ResponseClose}
         error, called at src/Bilge/Assert.hs:89:5 in bilge-0.22.0-FhngfA5b3MW5Tg8nd5Ja7l:Bilge.Assert
         <!!, called at src/Bilge/Assert.hs:107:19 in bilge-0.22.0-FhngfA5b3MW5Tg8nd5Ja7l:Bilge.Assert
         !!!, called at test/integration/Test/Federator/InwardSpec.hs:77:9 in main:Test.Federator.InwardSpec
```

is it possible that #1956 or #1973 broke develop?

services/brig/brig.cabal
pcapriotti: This only ended up by mistake in the cabal file, not in the package.yaml file.

pcapriotti: I don't think so.

pcapriotti:
> I still get the same error, but I'm not sure how to reproduce it without a 200 year old emacs haskell-mode and a self-written ghci wrapper script. `stack repl` works fine.

If you can share a script that breaks, I'm happy to look into it. But both stack repl and cabal repl work for me.

> But... how?! And why did having the data file in the cabal file break things?

No, that was unrelated, just some harmless leftover. The fix is to generate the doc files on the repl hook, as well as on the build hook.

services/brig/brig.cabal
fisx: but they should make it into the tar ball created by `cabal sdist`, no?  and that's only happening if they are listed here.

```
$ cabal sdist
remote: Enumerating objects: 60, done.
remote: Counting objects: 100% (60/60), done.
remote: Compressing objects: 100% (25/25), done.
remote: Total 60 (delta 40), reused 55 (delta 35), pack-reused 0
Unpacking objects: 100% (60/60), 14.09 KiB | 721.00 KiB/s, done.
From https://github.com/wireapp/http2
 * [new branch]      aeson-2             -> origin/aeson-2
 * [new branch]      client-status-fixme -> origin/client-status-fixme
 * [new branch]      preface-race        -> origin/preface-race
 * [new branch]      smatting/debugging  -> origin/smatting/debugging
Previous HEAD position was 7c465be Better support for setting header table size to 0
HEAD is now at 1ee1ce4 Make sure connection preface is always sent first
cabal: Error: Could not find module: Brig.Docs.Swagger with any suffix:
["gc","chs","hsc","x","y","ly","cpphs","hs","lhs","hsig","lhsig"]. If the
module is autogenerated it should be added to 'autogen-modules'.
```

ah, the last line is interesting!

fisx: (yes, we don't do sdists, but we might as well pretend! :))

fisx (approved):

PR #1976, opened 2021-12-08
SQSERVICES-547 Enforce disabling guest links when conversation is joined

battermann:
https://wearezeta.atlassian.net/browse/SQSERVICES-1057

This PR handles only `conversations/join`

## Checklist

 - [x] The **PR Title** explains the impact of the change.
 - [x] The **PR description** provides context as to why the change should occur and what the code contributes to that effect. This could also be a link to a JIRA ticket or a Github issue, if there is one.
 - ~~If HTTP endpoint paths have been added or renamed, the **endpoint / config-flag checklist** (see Wire-employee only backend [wiki page](https://github.com/zinfra/backend-wiki/wiki/Checklists)) has been followed.~~
 - ~~If a cassandra schema migration has been added, I ran **`make git-add-cassandra-schema`** to update the cassandra schema documentation.~~
 - [x] **changelog.d** contains the following bits of information ([details](https://github.com/wireapp/wire-server/blob/develop/docs/developer/changelog.md)):
   - [x] A file with the changelog entry in one or more suitable sub-sections. The sub-sections are marked by directories inside `changelog.d`.
   - ~~If new config options introduced: added usage description under docs/reference/config-options.md~~
   - ~~If new config options introduced: recommended measures to be taken by on-premise instance operators.~~
   - ~~If a cassandra schema migration is backwards incompatible (see also [these docs](https://github.com/wireapp/wire-server/blob/develop/docs/developer/cassandra-interaction.md#cassandra-schema-migrations)), measures to be taken by on-premise instance operators are explained.~~
   - ~~If a data migration (not schema migration) introduced: measures to be taken by on-premise instance operators.~~
   - ~~If public end-points have been changed or added: does nginz need un upgrade?~~
   - ~~If internal end-points have been added or changed: which services have to be deployed in a specific order?~~

PR #1980, opened 2021-12-09
SQSERVICES-547 post/get conversation/:cnv/code returns error if feature disabled

battermann:
https://wearezeta.atlassian.net/browse/SQSERVICES-1057

## Checklist

 - ~~If HTTP endpoint paths have been added or renamed, the **endpoint / config-flag checklist** (see Wire-employee only backend [wiki page](https://github.com/zinfra/backend-wiki/wiki/Checklists)) has been followed.~~
 - ~~If a cassandra schema migration has been added, I ran **`make git-add-cassandra-schema`** to update the cassandra schema documentation.~~
   - ~~If new config options introduced: added usage description under docs/reference/config-options.md~~
   - ~~If new config options introduced: recommended measures to be taken by on-premise instance operators.~~
   - ~~If a cassandra schema migration is backwards incompatible (see also [these docs](https://github.com/wireapp/wire-server/blob/develop/docs/developer/cassandra-interaction.md#cassandra-schema-migrations)), measures to be taken by on-premise instance operators are explained.~~
   - ~~If a data migration (not schema migration) introduced: measures to be taken by on-premise instance operators.~~
   - ~~If public end-points have been changed or added: does nginz need un upgrade?~~
   - ~~If internal end-points have been added or changed: which services have to be deployed in a specific order?~~

### DIFF
--- a/changelog.d/2-features/pr-1980
+++ b/changelog.d/2-features/pr-1980
@@ -1,1 +1,1 @@
-If guest links team feature is disabled `post /conversations/:cnv/code` and `get /conversations/:cnv/code` will return an error.
+Revoke guest links if feature is disabled. If the guest links team feature is disabled `get /conversations/join`, `post /conversations/:cnv/code`, and `get /conversations/:cnv/code` will return an error.
--- a/changelog.d/2-features/pr-1980
+++ b/changelog.d/2-features/pr-1980
@@ -1,0 +1,1 @@
+If guest links team feature is disabled `post /conversations/:cnv/code` and `get /conversations/:cnv/code` will return an error.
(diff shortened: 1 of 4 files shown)


fisx (changes requested):
Both this and #1976 look great, but please add integration tests.

- https://github.com/wireapp/wire-server/blob/402bdf7a0fc5fc2cc0435f63b9913bad04082af5/services/galley/test/integration/API.hs#L216 - add code (i think?)
- not so sure where `get` is called?
- I also just realized that if a conversation is created, there is the option of adding a guest link right away.  That should also trigger the same error (can be done here or in a separate PR; I'll add it to the ticket.)

fisx (approved):
LGTM!  (I probably would have tried to eliminate code duplication in the two tests, but this also works.)

PR #1869, opened 2021-10-19
add first mapping

julialongtin:
https://wearezeta.atlassian.net/browse/SQPIT-737

## Checklist

 - [x] The **PR Title** explains the impact of the change.
 - [x] The **PR description** provides context as to why the change should occur and what the code contributes to that effect. This could also be a link to a JIRA ticket or a Github issue, if there is one.
 - [x] **changelog.d** contains the following bits of information:
   - [x] A file with the changelog entry in one or more suitable sub-sections. The sub-sections are marked by directories inside `changelog.d`.

PR #1985, opened 2021-12-13
SQSERVICES-1126 Missing Clients (BSI Docs)

battermann:
This covers list item 2 from [this ticket](https://wearezeta.atlassian.net/browse/SQSERVICES-1126)

## Checklist


battermann:
I tried to make the test as self-explanatory as possible, so that there is no need to write extensive comments.

battermann:
In the ticket it says that the user should get a notification if sending of the message failed. Is there a way to test this, too. Is it necessary at all?

services/galley/test/integration/API.hs
```diff
@@ -516,6 +517,29 @@ postCryptoMessage4 = do
   postProtoOtrMessage alice (ClientId "172618352518396") conv m
     !!! const 403 === statusCode
 
+-- @SF.Separation @TSFI.RESTfulAPI @S2
+postMessageMissingClients :: TestM ()
+postMessageMissingClients = do
```
fisx, line 522:
```suggestion
postMessageRejectIfMissingClients :: TestM ()
postMessageRejectIfMissingClients = do
```

Purpose, Procedure and Expected Result.

Now the name arguably covers Purpose and Expected Result.  I agree that Procedure should be appearent from the code, let's see if we're right.

services/galley/test/integration/API.hs
```diff
@@ -220,7 +220,8 @@ tests s =
           test s "cannot join private conversation" postJoinConvFail,
           test s "remove user with only local convs" removeUserNoFederation,
           test s "remove user with local and remote convs" removeUser,
-          test s "iUpsertOne2OneConversation" testAllOne2OneConversationRequests
+          test s "iUpsertOne2OneConversation" testAllOne2OneConversationRequests,
+          test s "missing clients" postMessageMissingClients
```
fisx, line 224:
```suggestion
          test s "missing clients" postMessageRejectIfMissingClients
```

fisx (approved):

fisx:
@mythsunwind FYI.  (I'll also make sure the tickets will have all PRs linked by the time we consider ourselves done, then you can take a last look and second-guess me. :))

fisx:
I merged #1989 and hereby re-approve.  I think just tried locally, and I think this can be merged independently of #1869 in any order.

PR #1986, opened 2021-12-13
Split cannon benchmarks and tests

pcapriotti:
Cannon benchmarks were running as part of the test suite. This creates a separate component for cannon benchmarks, so that they are not run alonside the other unit tests.

## Checklist


services/cannon/bench/Main.hs
```diff
@@ -0,0 +1,31 @@
...
+
+import Bench
+import qualified Cannon.API
+import Data.Metrics.Test (pathsConsistencyCheck)
+import Data.Metrics.WaiRoute (treeToPaths)
+import Imports
+import Network.Wai.Utilities.Server (compile)
+import qualified Test.Cannon.Dict as D
+import Test.Tasty
+import Test.Tasty.HUnit
```
supersven, line 28: Do you need all these imports? The only imported symbol I see is `benchmark`... :thinking:

supersven: I think `import Bench` should be sufficient.

supersven (changes requested):

services/cannon/bench/Main.hs
pcapriotti: Yes, fixed in aed403d69ca497f01d590ac4efdb036cf0a418d3.

supersven:
Looks good to me. :+1: 

Let's wait for the CI results. When they're positive, I'll approve.

supersven (approved):

PR #1991, opened 2021-12-14
Improve Test Documentation for BSI (SQSERVICES-1127)

mdimjasevic:
This PR addresses the Haskell part of https://wearezeta.atlassian.net/browse/SQSERVICES-1127.

## Checklist

### DIFF
--- a/changelog.d/4-docs/sqservices-1127
+++ b/changelog.d/4-docs/sqservices-1127
@@ -1,0 +1,1 @@
+Add the description to several test cases
--- a/services/brig/test/integration/API/User/Auth.hs
+++ b/services/brig/test/integration/API/User/Auth.hs
@@ -400,7 +400,15 @@ testThrottleLogins conf b = do
   login b (defEmailLogin e) SessionCookie !!! const 200 === statusCode
 
 -- The testLimitRetries test conforms to the following testing standards:
--- @SF.Provisioning @TSFI.RESTfulAPI @S2
+-- @SF.Channel @TSFI.RESTfulAPI @S2
+--
+-- The following test tests the login retries. It checks that a user can make
+-- only a prespecified number of attempts to log in with an invalid password,
+-- after which the user is unable to try again for a configured amount of time.
+-- After the configured amount of time has passed, the test asserts the user can
+-- successfully log in again. Furthermore, the test asserts that another
+-- unrelated user can successfully log-in in parallel to the failed attempts of
+-- the aforementioned user.
 testLimitRetries :: HasCallStack => Opts.Opts -> Brig -> Http ()
 testLimitRetries conf brig = do
   let Just opts = Opts.setLimitFailedLogins . Opts.optSettings $ conf
@@ -910,6 +918,12 @@ testRemoveCookiesByLabelAndId b = do
 
 -- The testTooManyCookies test conforms to the following testing standards:
 -- @SF.Provisioning @TSFI.RESTfulAPI @S2
+--
+-- The test asserts that there is an upper limit for the number of user cookies
+-- per cookie type. It does that by concurrently attempting to create more
+-- persistent and session cookies than the configured maximum.
+-- Creation of new cookies beyond the limit causes deletion of the
+-- oldest cookies.
 testTooManyCookies :: Opts.Opts -> Brig -> Http ()
 testTooManyCookies config b = do
   u <- randomUser b
--- a/services/brig/test/integration/API/User/Client.hs
+++ b/services/brig/test/integration/API/User/Client.hs
@@ -89,6 +89,7 @@ tests _cl _at opts p b c g =
       test p "delete /clients/:client - 200 (pwd)" $ testRemoveClient True b c,
       test p "delete /clients/:client - 200 (no pwd)" $ testRemoveClient False b c,
       test p "delete /clients/:client - 400 (short pwd)" $ testRemoveClientShortPwd b,
+      test p "delete /clients/:client - 403 (incorrect pwd)" $ testRemoveClientIncorrectPwd b,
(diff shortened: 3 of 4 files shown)


changelog.d/4-docs/sqservices-1127
```diff
@@ -0,0 +1 @@
+Improve documentation for several test cases per SQSERVICES-1127
```
fisx, line 1: The changelog's target audience is the public, so I would like to avoid internal jira issues.

services/brig/test/integration/API/User/Auth.hs
```diff
@@ -910,6 +918,12 @@ testRemoveCookiesByLabelAndId b = do
 
 -- The testTooManyCookies test conforms to the following testing standards:
 -- @SF.Provisioning @TSFI.RESTfulAPI @S2
+--
+-- The test asserts that there is an upper limit for the number of user cookies
+-- per cookie type. It does that by concurrently attempting to create more than
+-- configured both the persistent and session cookie types. As expected, no more
```
fisx, line 924: grammar

changelog.d/4-docs/sqservices-1127
mdimjasevic: A fair point. I updated the change log so it does not point to the ticket.

services/brig/test/integration/API/User/Auth.hs
mdimjasevic: I hope it reads better now.

services/brig/test/integration/API/User/Client.hs
```diff
@@ -670,6 +683,10 @@ testMissingClient brig = do
 -- The testAddMultipleTemporary test conforms to the following testing standards:
 -- @SF.Provisioning @TSFI.RESTfulAPI @S2
 -- Legacy (galley)
+--
+-- The test validates that creating multiple temporary clients succeeds. It does
+-- so by invoking both a public and an internal endpoints. The expected outcome
+-- is that each call to each of the endpoints returns only one client.
```
fisx, line 689:
```suggestion
-- Add temporary client, check that all services (both galley and
-- brig) have registered it.  Add second temporary client, check
-- again.  (NB: temp clients replace each other, there can always be
-- at most one per account.)
```

services/brig/test/integration/API/User/Client.hs
```diff
@@ -481,6 +490,10 @@ testRemoveClient hasPwd brig cannon = do
 
 -- The testRemoveClientShortPwd test conforms to the following testing standards:
 -- @SF.Provisioning @TSFI.RESTfulAPI @S2
+--
+-- The test checks if a client can be deleted by providing a too short password.
+-- This is done by using a single-character password, whereas the minimum is 6
+-- characters. The client deletion attempt fails as expected.
```
fisx, line 496: your description is correct, but what a weird test to have!  i suggest we add a test with a syntactically valid but wrong password, and make sure that password verification also takes place?  (not strictly in the scope of this PR, but it's easy to add now, it makes sense, and it will potentially save us the trouble of doing another review round later.)

services/brig/test/integration/API/User/Client.hs
```diff
@@ -440,6 +445,10 @@ testTooManyClients opts brig = do
 
 -- The testRemoveClient test conforms to the following testing standards:
 -- @SF.Provisioning @TSFI.RESTfulAPI @S2
+--
+-- This test validates creating and deleting a client. A client is created and
+-- consequently deleted. It cannot be deleted twice. Prekeys and cookies are not
+-- there anymore once the client is deleted.
```
fisx, line 451:
```suggestion
-- This test validates creating and deleting a client. A client is created and
-- consequently deleted. Deleting a second time yields response 404 not found.
-- Prekeys and cookies are not there anymore once the client is deleted.
```

services/brig/test/integration/API/User/Client.hs
```diff
@@ -418,6 +418,11 @@ testMultiUserGetPrekeysQualified brig opts = do
 
 -- The testTooManyClients test conforms to the following testing standards:
 -- @SF.Provisioning @TSFI.RESTfulAPI @S2
+--
+-- The test validates the upper bound on the number of permanent clients per
+-- user. It does so by trying to create one permanent client more than allowed.
+-- The expected outcome is that all the clients up to the limit are successfully
+-- created, but the one over the limit is not.
```
fisx, line 425:
```suggestion
-- created, but the one over the limit is not (error `404 too-many-clients`).
```

services/brig/test/integration/API/User/Auth.hs
```diff
@@ -910,6 +918,12 @@ testRemoveCookiesByLabelAndId b = do
 
 -- The testTooManyCookies test conforms to the following testing standards:
 -- @SF.Provisioning @TSFI.RESTfulAPI @S2
+--
+-- The test asserts that there is an upper limit for the number of user cookies
+-- per cookie type. It does that by concurrently attempting to create more
+-- persistent and session cookies than the configured maximum. As expected, no
+-- more than the configured number of cookies per cookie type is allowed to be
+-- created.
```
fisx, line 926:
```suggestion
-- persistent and session cookies than the configured maximum.
-- Creation of new cookies beyond the limit causes deletion of the
-- oldest cookies.
```

fisx (changes requested):

services/brig/test/integration/API/User/Client.hs
mdimjasevic: This sounds like a good idea! I'll add the test you proposed as part of this PR.

fisx (approved):

PR #2001, opened 2021-12-16
Add end-tags for test tag parser (BSI audit).

fisx:
https://wearezeta.atlassian.net/browse/SQSERVICES-1128

## Checklist

### DIFF
--- a/services/brig/test/integration/API/User/Client.hs
+++ b/services/brig/test/integration/API/User/Client.hs
@@ -444,6 +444,8 @@ testTooManyClients opts brig = do
       const (Just "too-many-clients") === fmap Error.label . responseJsonMaybe
       const (Just "application/json;charset=utf-8") === getHeader "Content-Type"
 
+-- @END
+
 -- The testRemoveClient test conforms to the following testing standards:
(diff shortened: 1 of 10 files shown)


mdimjasevic (approved):

services/brig/test/integration/API/User/Client.hs
```diff
@@ -489,6 +489,8 @@ testRemoveClient hasPwd brig cannon = do
           newClientCookie = Just defCookieLabel
```
mythsunwind, line 489: I cannot mark the correct line in this file but there is a `@END` tag missing around line 446

mythsunwind: @fisx I see also that in this file there are spaces missing between `@SF.Channel` and `@TSFI.RESTfulAPI` in the lines before:
* "rejects saml responses with invalid issuer entity id"
* "rejects saml responses signed with the wrong private key"
* "rejects saml responses to requests not in cassandra:spar.authreq"
* "rejects saml responses already seen (and recorded in cassandra:spar.authresp)"

Additionally please change `@SF.CHANNEL` into `@SF.Channel`

mythsunwind (approved):

PR #2002, opened 2021-12-17
Federated Cargohold API

pcapriotti:
This PR makes the asset API federation-aware. This is accomplished by creating qualified endpoints for downloading and deleting assets, as well as returning the local domain as part of the response object on upload.

Federated behaviour is not yet implemented, but the new endpoints can already be used just like the old ones for local assets (i.e. when the domain is the local one).

Tracked by https://wearezeta.atlassian.net/browse/FS-306

## Checklist


services/cargohold/test/integration/API/V4.hs
```diff
@@ -0,0 +1,282 @@
...
+  let sets = def : map (\r -> def & setAssetRetention ?~ r) rets
+  mapM_ simpleRoundtrip sets
+  where
+    simpleRoundtrip sets = do
+      uid <- liftIO $ Id <$> nextRandom
+      uid2 <- liftIO $ Id <$> nextRandom
+      -- Initial upload
+      let bdy = (applicationText, "Hello World")
+      r1 <-
+        uploadSimple (c . path "/assets/v3") uid sets bdy
```
mdimjasevic: is v3 intended here, or perhaps v4?

services/cargohold/test/integration/API/V4.hs
```diff
@@ -0,0 +1,282 @@
...
+
+testSimpleTokens :: TestSignature ()
+testSimpleTokens c = do
+  uid <- liftIO $ Id <$> nextRandom
+  uid2 <- liftIO $ Id <$> nextRandom
+  -- Initial upload
+  let sets = defAssetSettings & set setAssetRetention (Just AssetVolatile)
+  let bdy = (applicationText, "Hello World")
+  r1 <-
+    uploadSimple (c . path "/assets/v3") uid sets bdy
```
mdimjasevic: ditto

services/cargohold/test/integration/API/V4.hs
```diff
@@ -0,0 +1,282 @@
...
+  let key = view assetKey ast
+  let Just tok = view assetToken ast
+  -- No access without token from other user (opaque 404)
+  get (c . path loc . zUser uid2 . noRedirect)
+    !!! const 404 === statusCode
+  -- No access with wrong token (opaque 404)
+  get (c . path loc . zUser uid2 . header "Asset-Token" "acb123" . noRedirect)
+    !!! const 404 === statusCode
+  -- Token renewal fails if not done by owner
+  post (c . paths ["assets", "v3", toByteString' (qUnqualified key), "token"] . zUser uid2) !!! do
```
mdimjasevic: Isn't there a utility function for making this request? And should it be v3 or v4?

services/cargohold/test/integration/API/V4.hs
```diff
@@ -0,0 +1,282 @@
...
+    const 403 === statusCode
+    const (Just "unauthorised") === fmap label . responseJsonMaybe
+  -- Token renewal succeeds if done by owner
+  r2 <-
+    post (c . paths ["assets", "v3", toByteString' (qUnqualified key), "token"] . zUser uid)
```
mdimjasevic: ditto

services/cargohold/test/integration/API/V4.hs
```diff
@@ -0,0 +1,282 @@
...
+    assertEqual "user mismatch" uid (decodeHeader "x-amz-meta-user" r4)
+    assertEqual "data mismatch" (Just "Hello World") (responseBody r4)
+  -- Verify access without token if the request comes from the creator.
+  get (c . path loc . zUser uid . noRedirect)
+    !!! const 302 === statusCode
+  -- Verify access with new token from a different user.
+  get (c . path loc . header "Asset-Token" (toByteString' tok') . zUser uid2 . noRedirect)
+    !!! const 302 === statusCode
+  -- Delete Token fails if not done by owner
+  delete (c . paths ["assets", "v3", toByteString' (qUnqualified key), "token"] . zUser uid2) !!! do
```
mdimjasevic: v3 or v4?

services/cargohold/test/integration/API/V4.hs
```diff
@@ -0,0 +1,282 @@
...
+    const 403 === statusCode
+    const (Just "unauthorised") === fmap label . responseJsonMaybe
+  -- Delete Token succeeds by owner
+  delete (c . paths ["assets", "v3", toByteString' (qUnqualified key), "token"] . zUser uid) !!! do
```
mdimjasevic: ditto

services/cargohold/test/integration/API/V4.hs
```diff
@@ -0,0 +1,282 @@
...
+      . lbytes bs
+
+deleteAsset :: CargoHold -> UserId -> Qualified AssetKey -> Http (Response (Maybe Lazy.ByteString))
+deleteAsset c u k = delete $ c . zUser u . paths ["assets", "v3", toByteString' (qUnqualified k)]
+
+-- Utilities ------------------------------------------------------------------
+
+decodeHeader :: FromByteString a => HeaderName -> Response b -> a
+decodeHeader h =
+  fromMaybe (error $ "decodeHeader: missing or invalid header: " ++ show h)
```
mdimjasevic: I'd prefer the return type to be `Maybe a` or similar rather than invoking `error`, but perhaps this ain't that bad given it's test code.

mdimjasevic (commented):
I'll continue later with reviewing. Some smaller comments are inlined.

mdimjasevic (approved):

libs/wire-api/src/Wire/API/Asset.hs
```diff
@@ -18,21 +20,345 @@
...
+import Data.ByteString.Conversion
+import qualified Data.ByteString.Lazy as LBS
+import Data.Id
+import Data.Json.Util (UTCTimeMillis (fromUTCTimeMillis), toUTCTimeMillis)
+import Data.Proxy
+import Data.Qualified
 import Data.SOP
+import Data.Schema
+import qualified Data.Swagger as S
 import qualified Data.Swagger as Swagger
```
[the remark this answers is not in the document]
supersven: This is now done.

services/cargohold/test/integration/API/V4.hs
supersven: The uploading of assets is still `v3` for users.

services/cargohold/test/integration/API/V4.hs
supersven: That's indeed a good question, because both - `v3` and `v4` - are available. :thinking:

I'll check that nothing is untested in that case.

services/cargohold/test/integration/API/V4.hs
supersven: The idea is that if you're looking up a header in a test, you'll likely either want to assert it's there or use it as some kind of test setup and cannot reasonably resume without it.

While looking at it again, I realized that it might make sense to add a `HasCallStack` constraint, to get more meaningful backtraces in case of error.

@mdimjasevic , do you think we should change the name; communicating that this might fail? E.g. to `decodeHeaderOrFail`?

mdimjasevic: @supersven , I agree with adding `HasCallStack`. Yes, the renaming suggestion also sounds good to me.

services/cargohold/test/integration/API/V4.hs
supersven: It's a bit unfortunate, but this utility function will be introduced and used with the next PR: https://github.com/wireapp/wire-server/pull/2004/files#diff-e9ecc1bf7fe0d5557306c1ab67374434916fe268af016613313a503dc19f2cd5R146

@mdimjasevic , would it be okay to keep this for now in this PR? As the proposed change is already available in the upcoming / next PR...

services/cargohold/test/integration/API/V4.hs
supersven: Deleting an asset is still `v3`.

services/cargohold/test/integration/API/V4.hs
mdimjasevic: Sure, introducing it in the next PR sounds good!

PR #2005, opened 2021-12-18
Move Spar Polysemy Specs into spar:lib

isovector:
This PR moves the spar `polysemy-check` specs into `spar:lib`, so that we can reuse them between the mock and real interpreters.

## Checklist


services/spar/test/Test/Spar/Sem/IdPRawMetadataStoreSpec.hs
```diff
@@ -1,136 +1,20 @@
...
-                pure Nothing,
-              lawPrelude =
-                [ E.store idpid t
-                ],
-              lawPostlude = [] @(Sem _ ())
-            }
-    )
 
 testInterpreter :: Sem '[E.IdPRawMetadataStore] a -> IO (RawState, a)
 testInterpreter = pure . run . idpRawMetadataStoreToMem
```
fisx, line 15: we're going to need this in the production code as well in order to compare it to the production interpreter in `test-integration`, but it's easy enough to move it then.

services/spar/test/Test/Spar/Sem/ScimExternalIdStoreSpec.hs
```diff
@@ -1,192 +1,16 @@
...
-    prop "delete/lookup" $ prop_deleteLookup (Just $ show . (() <$) . extract) lower
-    prop "delete/insert" $ prop_deleteInsert Nothing lower
-    prop "lookup/insert" $ prop_lookupInsert Nothing lower
-    prop "insert/delete" $ prop_insertDelete Nothing lower
-    prop "insert/lookup" $ prop_insertLookup (Just $ show . (() <$) . extract) lower
-    prop "insert/insert" $ prop_insertInsert (Just $ show . (() <$) . extract) lower
 
 spec :: Spec
 spec = modifyMaxSuccess (const 1000) $ do
   propsForInterpreter "scimExternalIdStoreToMem" snd $ pure . run . scimExternalIdStoreToMem
```
fisx, line 15: same here.

but i realize now that i haven't though this through.  i know we have something close to a proof that effects don't interfere with each other, but i guess i always thought that we have two interpreters that handle all the effects under `Spar.Sem`: one for production code and one for the unit tests; and that we're running all the tests, including these here, the unit tests covering the application logic, and the intergration tests checking for equivalence between the two interpreters.

does this idea even make sense?  is it reasonable?  would it be more expensive than hand-picking interpreters for just the one effect you're dealing with in every particular module?

fisx (approved):

services/spar/test/Test/Spar/Sem/ScimExternalIdStoreSpec.hs
isovector: I'm not sure I understand the question. My plan is to still test the entire application via test interpreters, but that only finds bugs exercised by the existing test suite. These property tests should give us a good idea if there are bugs hiding in the existing effects themselves.

PR #2009, opened 2021-12-21
WIP: Servantify remaining Brig endpoints

isovector:
## Checklist


libs/wire-api/src/Wire/API/User/RichInfo.hs
```diff
@@ -244,16 +248,17 @@ instance Monoid RichInfoAssocList where
...
-  toJSON (RichInfoAssocList l) =
-    object
-      [ "fields" .= l,
-        "version" .= (0 :: Int)
-      ]
+instance S.ToSchema RichInfoAssocList where
+  schema =
+    S.object "RichInfoAssocList" $
+      fmap snd $ (,)
+          -- TODO(sandy): check the version is equal to 0
```
isovector, line 255: @pcapriotti The `FromJSON` instance immediately after this checks to ensure the version is 0, failing otherwise. Is this a check I can replicate in schema-profunctor?

Furthermore, this `version` doesn't actually end up in the datatype. Is there a better way than this ugly `fmap snd $ (,) ...` thing I'm doing here?

pcapriotti: Yes to both.

To add a check, you can use `withParser`, with a function that performs the check and fails with `fail` in the `Aeson.Parser` monad. `withParser` allows you to change the type, so that answers your second question as well.

So for example you can have something like:
```haskell
withParser ((,) <$> field1 <*> field2) $ \(v, xs) -> checkVersion v $> xs
```

Also, there's no need to use `dimap` here. `const 0 .= schema` should work fine (the annotation can probably be removed once you add `withParser` in).

libs/wire-api/src/Wire/API/User/RichInfo.hs
```diff
@@ -290,30 +295,24 @@ modelRichField = Doc.defineModel "RichField" $ do
...
 
+instance S.ToSchema RichField where
+  -- NB: "name" would be a better name for 'richFieldType', but "type" is used because we
+  -- also have "type" in SCIM; and the reason we use "type" for SCIM is that @{"type": ...,
+  -- "value": ...}@ is how all other SCIM payloads are formatted, so it's quite possible
+  -- that some provisioning agent would support "type" but not "name".
+  schema =
+    S.object "RichField" $
+      RichField
+        <$> richFieldType S..= S.field "type" (S.dimap CI.original CI.mk S.schema)
```
pcapriotti, line 313: Same here about `dimap`: `CI.original .= CI.mk <$> schema` is how I normally write these things.

libs/wire-api/src/Wire/API/Routes/Public/Brig.hs
```diff
@@ -73,6 +77,27 @@ type DeleteSelfResponses =
...
+
+-- TODO(sandy): Not sure about this instance -- does false correspond to the
+-- password being set?
+instance AsUnion CheckPasswordExistsResposes Bool where
+  toUnion False = Z (I ())
+  toUnion True = S (Z (I ()))
+  fromUnion (Z (I x)) = False
+  fromUnion (S (Z (I x))) = True
+  fromUnion (S (S ns)) = case ns of
+
```
pcapriotti, line 100: There should be no need for this. If you follow the convention is that "error" or "negative" responses go first, then there is already an instance of `AsUnion` for this case, and you should be able to use
```haskell
'[ Respond 404 ...,
  Respond 200 ...
```
as the list of responses.

libs/wire-api/src/Wire/API/User.hs
```diff
@@ -872,35 +866,29 @@ modelChangePassword = Doc.defineModel "ChangePassword" $ do
...
-instance FromJSON PasswordChange where
-  parseJSON = A.withObject "PasswordChange" $ \o ->
-    PasswordChange
-      <$> o A..:? "old_password"
-      <*> o A..: "new_password"
+instance ToSchema PasswordChange where
+  schema =
+    object "PasswordChange" $
+      PasswordChange
+        <$> cpOldPassword .= maybe_ (optField "old_password" schema)
```
pcapriotti, line 873: This should be
```haskell
cpOldPassword .= optField "old_password" (maybeWithDefault A.Null schema)
```

The difference is that now the field will be output even when its value is `Nothing`, and it will be given a value of `null`, whereas before it was omitted altogether. It probably makes no difference in practice, but to avoid any issues we should try to preserve the previous behaviour exactly if possible.

libs/wire-api/src/Wire/API/User.hs
```diff
@@ -872,35 +866,29 @@ modelChangePassword = Doc.defineModel "ChangePassword" $ do
...
 
-instance ToJSON LocaleUpdate where
-  toJSON l = A.object ["locale" A..= luLocale l]
-
-instance FromJSON LocaleUpdate where
-  parseJSON = A.withObject "locale-update" $ \o ->
-    LocaleUpdate <$> o A..: "locale"
+instance ToSchema LocaleUpdate where
+  schema =
+    object "locale-update" $
```
pcapriotti, line 889: Better to use camel case object names, for consistency in the swagger documentation.

libs/wire-api/src/Wire/API/User.hs
```diff
@@ -929,37 +917,36 @@ instance FromJSON EmailUpdate where
...
 
-instance ToJSON PhoneUpdate where
-  toJSON p = A.object ["phone" A..= puPhone p]
-
-instance FromJSON PhoneUpdate where
-  parseJSON = A.withObject "phone-update" $ \o ->
-    PhoneUpdate <$> o A..: "phone"
+instance ToSchema PhoneUpdate where
+  schema =
+    object "phone-update" $
```
pcapriotti, line 930: Please use a camel case object name.

libs/wire-api/src/Wire/API/User.hs
```diff
@@ -929,37 +917,36 @@ instance FromJSON EmailUpdate where
...
+        <$> puPhone .= field "phone" schema
 
 newtype HandleUpdate = HandleUpdate {huHandle :: Text}
   deriving stock (Eq, Show, Generic)
   deriving newtype (Arbitrary)
+  deriving (ToJSON, FromJSON, S.ToSchema) via (Schema HandleUpdate)
+
+instance ToSchema HandleUpdate where
+  schema =
+    object "handle-update" $
```
pcapriotti, line 941: Again, camel case object name.

libs/wire-api/src/Wire/API/User/Handle.hs
```diff
@@ -84,6 +86,13 @@ modelCheckHandles = Doc.defineModel "CheckHandles" $ do
     Doc.description "Desired number of free handles to return (1 - 10). Default 1."
     Doc.optional
 
+instance ToSchema CheckHandles where
+  schema =
+    object "CheckHandles" $
+      CheckHandles
+        <$> (checkHandlesList .= field "handles" (fromRange .= rangedSchema sing sing (array schema)))
+        <*> (checkHandlesNum .= field "return" schema)
```
pcapriotti, line 94: Extra brackets here.

libs/wire-api/src/Wire/API/User/Identity.hs
```diff
@@ -242,13 +242,11 @@ validateEmail =
...
+  deriving (ToJSON, FromJSON, S.ToSchema) via Schema Phone
 
-instance FromJSON Phone where
-  parseJSON (A.String s) = case parsePhone s of
-    Just p -> return p
-    Nothing -> fail "Invalid phone number. Expected E.164 format."
-  parseJSON _ = mempty
+instance ToSchema Phone where
+  -- TODO(sandy): Validate the phone number
+  schema = dimap fromPhone Phone schema
```
pcapriotti, line 249: There's a convenience function `parsedText` for these sorts of schemas, which are very common.

pcapriotti (commented):
Looking great so far. I've left a few comments with some minor suggestions.

One other small (but possibly annoying) thing: could you please import `Data.Schema` unqualified, and qualify Swagger and Aeson stuff instead? There is often very little need to use those two modules directly, and there are unfortunately a lot of conflicting bindings. Mostly you can get away with:

```
import Data.Aeson (FromJSON (..), ToJSON(..))
import Data.Schema
import qualified Data.Swagger as S
```
and sometimes you need to `import qualified Data.Aeson.Types as A`.

libs/wire-api/src/Wire/API/Routes/Public/Brig.hs
pcapriotti: Sorry, I meant `RespondEmpty` instead of `Respond`.

PR #2011, opened 2021-12-23
Fix doubtless HLint issues in Federator

supersven:
## Checklist

### DIFF
--- a/services/federator/src/Federator/ExternalServer.hs
+++ b/services/federator/src/Federator/ExternalServer.hs
@@ -107,14 +107,14 @@ parseRequestData req = do
   when (Wai.requestMethod req /= HTTP.methodPost) $
     throw InvalidRoute
   -- No query parameters are allowed
-  when (not . BS.null . Wai.rawQueryString $ req) $
+  unless (BS.null . Wai.rawQueryString $ req) $
     throw InvalidRoute
   -- check that the path has the expected form
   (componentSeg, rpcPath) <- case Wai.pathInfo req of
     ["federation", comp, rpc] -> pure (comp, rpc)
     _ -> throw InvalidRoute
 
-  when (not (Text.all isAllowedRPCChar rpcPath)) $
+  unless (Text.all isAllowedRPCChar rpcPath) $
     throw InvalidRoute
 
   when (Text.null rpcPath) $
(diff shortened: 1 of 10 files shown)


services/federator/src/Federator/ExternalServer.hs
```diff
@@ -107,14 +107,14 @@ parseRequestData req = do
...
-  when (not . BS.null . Wai.rawQueryString $ req) $
+  unless (BS.null . Wai.rawQueryString $ req) $
     throw InvalidRoute
   -- check that the path has the expected form
   (componentSeg, rpcPath) <- case Wai.pathInfo req of
     ["federation", comp, rpc] -> pure (comp, rpc)
     _ -> throw InvalidRoute
 
-  when (not (Text.all isAllowedRPCChar rpcPath)) $
+  unless (Text.all isAllowedRPCChar rpcPath) $
```
fisx, line 117: I like this change, but I doubt that it's non-controversial.  :)  Let's see if anybody complains.

services/federator/test/unit/Test/Federator/Client.hs
```diff
@@ -14,6 +14,7 @@
 --
 -- You should have received a copy of the GNU Affero General Public License along
 -- with this program. If not, see <https://www.gnu.org/licenses/>.
+{-# LANGUAGE LambdaCase #-}
```
fisx, line 17: this is in `/package-defaults.yaml` and should need to be added in the module.  i don't mind the noise (in this case), just saying.

fisx (approved):

fisx:
(please add a changelog line.)

services/federator/src/Federator/ExternalServer.hs
supersven: They seem to be exactly the opposite of each other:

```haskell
when p s = if p then s else pure ()

unless p s =  if p then pure () else s
```

So, I'm wondering how this could be controversial. :thinking:

services/federator/test/unit/Test/Federator/Client.hs
supersven: Good spot :+1: 

This is now fixed.

services/federator/src/Federator/ExternalServer.hs
fisx: it's clearly still correct, just a question of style: is code easier to read if with fewer keywords, or with less complexity in boolean expressions?  i'm also leaning towards `unless`, and i think we're way too deep into a bike shedding thing again here, so just ignore me.  if somebody feels strongly, they can make a PR with an hlint rule that rolls this back, *then* we can bike shed.  :)

fisx:
good to merge!

PR #2019, opened 2022-01-03
Sftd helm chart: provide a list of all sftd urls over http(s) using sftd_disco sidecar process.

jschaul:
Related to https://wearezeta.atlassian.net/browse/FS-266.

Implements querying the list of all SFT servers from the sftd's sidecar nginx pod. This is a different take on the above issue and if this list provided by sft servers is made use of; then that makes some of the work introduced in #2012 #2014 #2015 obsolete (sorry 😟 ). 

From brig, using the url configured in `setSftStaticUrl`, and calling `/sft_servers_all.json` on that URL, this PR provides a list of all SFT servers available. The list is computed on the SFTD chart itself (using a sftd_disco sidecar container). This has the advantage of allowing brig to be hosted on a separate kubernetes cluster to sftd, and moves the business logic of knowing sftd servers to sftd itself.

From a brig pod on a test cluster:

```
 # cat /etc/wire/brig/conf/brig.yaml | grep sft
  setSftStaticUrl: https://sftd.a.adhoc-testing.wire.link:443
/ # curl -ks https://sftd.a.adhoc-testing.wire.link:443/sft_servers_all.json
{
  "sft_servers_all": [
    "https://sftd.a.adhoc-testing.wire.link/sfts/wire-server-a-sftd-0"
  ]
}
```

## Checklist

 - [x] The **PR Title** explains the impact of the change.
 - [x] The **PR description** provides context as to why the change should occur and what the code contributes to that effect. This could also be a link to a JIRA ticket or a Github issue, if there is one.
 - [x] **changelog.d** contains the following bits of information ([details](https://github.com/wireapp/wire-server/blob/develop/docs/developer/changelog.md)):
   - [x] A file with the changelog entry in one or more suitable sub-sections. The sub-sections are marked by directories inside `changelog.d`.

PR #2020, opened 2022-01-03
brig->sft lookup: sft service internally talks http not https

jschaul:
(note: this code is based on some deployment assumptions that don't hold and should be refactored again soon; see #2019 for upcoming work)

Fixup for the changes introduced in #2015 #2012 #2014


jschaul:
This doesn't solve the problem, so no need to merge. There are no logs when calling /calls/config so those could be added to gain better visibility here.

mdimjasevic (commented):
How about I do away with this IP-based approach and do as you suggested (branching on whether a load balancer is configured)?

mdimjasevic (approved):

PR #2022, opened 2022-01-04
Remove uses of servant-generic in public Galley API

pcapriotti:
The generic mechanism in Servant for defining API types is cumbersome and slow to compile. This commit replaces the API record for the public Galley API with a tree explicitly built using the `:<|>` operator.

To make up for the lack of record labels, it also introduces a `Named` combinator for servant API types, which can be used to tag an endpoint with a name (usually a symbol). The name must match the one with which the handler is wrapped.

Compile times are indeed better with this change, but unfortunately not by much.

## Checklist

### DIFF
--- a/libs/wire-api/src/Wire/API/Routes/Named.hs
+++ b/libs/wire-api/src/Wire/API/Routes/Named.hs
@@ -1,0 +1,40 @@
+-- This file is part of the Wire Server implementation.
+--
+-- Copyright (C) 2022 Wire Swiss GmbH <opensource@wire.com>
+--
+-- This program is free software: you can redistribute it and/or modify it under
+-- the terms of the GNU Affero General Public License as published by the Free
+-- Software Foundation, either version 3 of the License, or (at your option) any
+-- later version.
+--
+-- This program is distributed in the hope that it will be useful, but WITHOUT
+-- ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+-- FOR A PARTICULAR PURPOSE. See the GNU Affero General Public License for more
+-- details.
+--
+-- You should have received a copy of the GNU Affero General Public License along
+-- with this program. If not, see <https://www.gnu.org/licenses/>.
+
+module Wire.API.Routes.Named where
+
+import Data.Metrics.Servant
+import Data.Proxy
+import Imports
+import Servant.Server
+import Servant.Swagger
+
+newtype Named named x = Named {unnamed :: x}
+  deriving (Functor)
+
+instance HasSwagger api => HasSwagger (Named name api) where
+  toSwagger _ = toSwagger (Proxy @api)
+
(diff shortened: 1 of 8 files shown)


libs/wire-api/src/Wire/API/Routes/Public/Galley.hs
```diff
@@ -83,501 +83,577 @@ type RemoveFromConversationVerb =
...
     (Maybe Event)
 
-data Api routes = Api
-  { -- Conversations
+type ServantAPI =
+  ConversationAPI
+    :<|> TeamConversationAPI
+    :<|> MessagingAPI
+    :<|> TeamAPI
+    :<|> FeatureAPI
```
fisx, line 91: I like the more tree-shaped structure!  Much nicer than the comments over sections of a huge record.

libs/wire-api/src/Wire/API/Routes/Named.hs
```diff
@@ -0,0 +1,40 @@
...
+
+module Wire.API.Routes.Named where
+
+import Data.Metrics.Servant
+import Data.Proxy
+import Imports
+import Servant.Server
+import Servant.Swagger
+
+newtype Named named x = Named {unnamed :: x}
```
fisx, line 26: wasn't the named record fields thing coincidental, and the true purpose of using records was to have better error messages?

what do the error messages look like with named routes?

fisx (approved):
No objections from me, just a few questions.  Let's see what other people think!   Maybe wait for another approval or two?

libs/wire-api/src/Wire/API/Routes/Named.hs
pcapriotti: The `Named` stuff compensates for the fact that handlers have to be matched up with their types in order. If you swap two similar unnamed handlers by mistake, you won't get an error if their types are the same. By using `Named` types and handlers, we get a name mismatch error (which actually looks reasonably nice). I don't mind getting rid of the `Named` trick altogether, if people think it's not necessary.

Errors messages are in general horrible. GHC unhelpfully spews out the whole API type multiple times in the output, forcing you to use some kind of pager to be able to see the top of the error message. It wasn't much better before, but I have to admit this change did not improve things in this regard.

For example, if you forget a handler, GHC points you to the the last one, instead of the spot where the handler is missing. If you misplace a handler, it just says that some types don't match, with no indication to where the problem is.

There are probably ways to improve this, but I'm in general a bit sceptical about meddling with error messages, as it can result in very confusing messages sometimes (e.g. in servant's Union stuff).

fisx: ok, thanks for the background.  now i think i prefer the 'Named' thing, but only slightly.

no further questions!  once we've had a few more eyes i'm happy with these changes.

PR #2023, opened 2022-01-04
Return specific error when a domain is blocked

battermann:
https://wearezeta.atlassian.net/browse/SQSERVICES-996

## Checklist

### DIFF
--- a/changelog.d/5-internal/pr-2023
+++ b/changelog.d/5-internal/pr-2023
@@ -1,0 +1,1 @@
+When sending an activation code, the blocked domains are checked before the whitelist. This only affects staging because there is no whitelist configuration in prod.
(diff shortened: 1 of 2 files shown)


changelog.d/5-internal/pr-2023
```diff
@@ -0,0 +1 @@
+When sending an activation code, the blocked domains are checked before the whitelist. This only affects staging because there is no whitelist configuration in prod.
```
fisx, line 1:
```suggestion
When sending an activation code, the blocked domains are checked before the whitelist. This only affects the wire SaaS staging environment (there is no whitelist configuration in prod, and blocked domains are not applicable to on-prem installations).
```

fisx (approved):
please update the changelog to cover on-prem users, otherwise good to go!

(did hlint ask you to change the anon lambda?)

battermann:
> (did hlint ask you to change the anon lambda?)

No, I find it less noisy to look at and therefore easier to read.

PR #2026, opened 2022-01-05
[FS-266] Fix SFT URL fecthing based on plain HTTP via IPv4

mdimjasevic:
This is part of https://wearezeta.atlassian.net/browse/FS-266:

- It allows for using a plain HTTP connection given an IPv4 address of an SFT server,
- It adds a proper integration test for `sft_servers_all` as the previously extended test silently ignored a code path.

## Checklist


jschaul (approved):
The existing code introduced in #2012  #2014 #2015 seems to be broken, so this PR may improve on the status quo.

PR #2029, opened 2022-01-07
Servantify Cannon's internal API

supersven:
Servantify the internal API of Cannon and use `Named` for the public API, too.

It would be good if the reviewer would especially look closely at the streaming parts, because I'm not 100% confident there.

## Checklist


changelog.d/5-internal/servantify-cannon-public-api
```diff
@@ -0,0 +1 @@
+Migrate the public API of Cannon to Servant.
```
supersven, line 1: Remove this entry as the public endpoint's code is already merged.

services/cannon/src/Cannon/API/Internal.hs
```diff
@@ -16,71 +18,107 @@
...
+           :<|> Named
+                  "get-preferences"
+                  ( "presences"
+                      :> Capture "uid" UserId
+                      :> Capture "conn" ConnId
+                      :> MultiVerb
+                           'HEAD
+                           '[PlainText]
+                           '[ RespondEmpty 404 "Not found: Presence not registered.",
+                              Respond 200 "Presence checked successfully." String
```
supersven, line 96: Use `RespondEmpty` here. (HTTP head returns no body by convention.)

services/cannon/src/Cannon/API/Internal.hs
```diff
@@ -16,71 +18,106 @@
...
+                  "head-status"
+                  ( "status"
+                      :> NoContentWithStatusCodeVerb 'HEAD 200
+                  )
+           :<|> Named
+                  "push-notification"
+                  ( "push"
+                      :> Capture "user" UserId
+                      :> Capture "conn" ConnId
+                      :> StreamBody NoFraming OctetStream (SourceIO ByteString)
```
pcapriotti, line 71: Did you mean to use `PushNotificationStream` here instead of `SourceIO ByteString`? Otherwise, what is the point of that `newtype`?

services/cannon/src/Cannon/WS.hs
```diff
@@ -234,6 +237,23 @@ sendMsgIO :: L.ByteString -> Websocket -> IO ()
 sendMsgIO m c = do
   recoverAll retry3x $ const $ sendBinaryData (connection c) m
 
+sendMsgConduit :: Key -> Websocket -> ConduitT ByteString Void (ResourceT WS) ()
```
pcapriotti, line 240: Looks like this replaces `sendMsg`, so that function can be deleted.

services/cannon/src/Cannon/WS.hs
```diff
@@ -234,6 +237,23 @@ sendMsgIO :: L.ByteString -> Websocket -> IO ()
...
+      lift $ traceLog m'
+      liftIO $ sendToWebsocket m'
+    Nothing -> pure ()
+  where
+    traceLog :: ByteString -> (ResourceT WS) ()
+    traceLog m = lift $ trace $ client kb . msg (val "sendMsg: \"" +++ Data.ByteString.take 128 m +++ val "...\"")
+
+    kb = key2bytes k
+
+    sendToWebsocket :: ByteString -> IO ()
```
pcapriotti, line 254: You could make `sendMsgIO` polymorphic in the input type (i.e. take an `a` constrained by `WebSocketsData a`), and use that instead of redefining it here.

pcapriotti (approved):
Nice job!

If I understand correctly, the `push` endpoint was not streaming before, but now it is, which seems to be a nice improvement. But I guess `bulkpush` is still not streaming.

I haven't checked the test coverage. Hopefully it is good enough to catch potential mistakes in the refactored handlers.

Anyway, looks good generally. Some minor comments follow.

libs/wire-api/src/Wire/API/Routes/NoContentWithStatusCodeVerb.hs
```diff
@@ -0,0 +1,27 @@
+module Wire.API.Routes.NoContentWithStatusCodeVerb where
```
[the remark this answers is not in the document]
supersven: Seems to be a good step as it reduces the code we've to maintain. :+1:

services/cannon/src/Cannon/API/Internal.hs
```diff
@@ -16,71 +18,106 @@
...
+    :> ( Named
+           "get-status"
+           ( "status"
+               :> NoContentWithStatusCodeVerb 'GET 200
+           )
+           :<|> Named
+                  "head-status"
+                  ( "status"
+                      :> NoContentWithStatusCodeVerb 'HEAD 200
+                  )
```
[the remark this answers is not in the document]
supersven: I've removed the endpoint and - to double check - made a `curl --head`. Works. :+1:

services/cannon/src/Cannon/WS.hs
supersven: Indeed, `sendMsg` is a left-over.

services/cannon/src/Cannon/WS.hs
supersven: :+1:

pcapriotti (approved):
Looks great!

PR #2030, opened 2022-01-07
[FS-266] Replace IPv4-HTTP-only Approach to SFT Server Lookup with /sft_servers_all.json

mdimjasevic:
This is part of https://wearezeta.atlassian.net/browse/FS-266. It is a clean-up that replaces IPV4 HTTP-only approach with SFT discovery based on either the SFT static URL (with the `GET /sft_servers_all.json` endpoint) or the full list of SFT servers from SRV records.

This cleans up various aspects of #2012 , #2014 and #2015 . Furthermore, it directly builds on top of #2019 .

**Note**: I tested this PR via https://github.com/zinfra/cailleach/pull/787 on Anta and it works.

## Checklist


pcapriotti (approved):
LGTM

PR #2033, opened 2022-01-11
Use named routes in Federation API

pcapriotti:
This refactors the federation API routing tables to use the `Named` convention introduced in #2022. It also implements a mechanism to construct Servant clients for specific endpoints by name.

This is preliminary refactoring work for the implementation of [API versioning](https://github.com/wireapp/wire-server/blob/develop/docs/developer/api-versioning.md).

Tracked by https://wearezeta.atlassian.net/browse/FS-127.

## Checklist

### DIFF
--- a/changelog.d/6-federation/named-fed-api-routes
+++ b/changelog.d/6-federation/named-fed-api-routes
@@ -1,0 +1,1 @@
+Use `Named` routes for the federation API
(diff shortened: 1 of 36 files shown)


mdimjasevic (approved):
This looks great! I like how you simplified the federation API and how you used type families to check if an endpoint given by a name exist in an API.

PR #2034, opened 2022-01-11
Add restund chart as dependency of wire-server

smatting:
This is a follow up to https://github.com/wireapp/wire-server/pull/2003.

## Checklist


smatting:
Closing this, because we don't want to encourage installing restund on the same cluster as wire-server

PR #2035, opened 2022-01-11
SQSERVICES-1083 Separate Access Control For Guests And Services

battermann:
https://wearezeta.atlassian.net/browse/SQSERVICES-1083

## Checklist


libs/wire-api/src/Wire/API/Conversation.hs
```diff
@@ -436,9 +446,47 @@ data AccessRole
...
+toAccessRole accessRoles
+  | Set.null accessRoles = NonActivatedAccessRole
+  | Set.fromList [TeamMemberAccessRole] == accessRoles = TeamAccessRole
+  | Set.fromList [GuestAccessRole] == accessRoles = NonActivatedAccessRole
+  | Set.fromList [ServiceAccessRole] == accessRoles = NonActivatedAccessRole
+  | Set.fromList [TeamMemberAccessRole, GuestAccessRole] == accessRoles = NonActivatedAccessRole
+  | Set.fromList [TeamMemberAccessRole, ServiceAccessRole] == accessRoles = NonActivatedAccessRole
+  | Set.fromList [GuestAccessRole, ServiceAccessRole] == accessRoles = NonActivatedAccessRole
+  | Set.fromList [TeamMemberAccessRole, GuestAccessRole, ServiceAccessRole] == accessRoles = NonActivatedAccessRole
+  | otherwise = ActivatedAccessRole
```
fisx, line 475: It would be really bad if this function couldn't be total, but I think it can be.  (Doesn't the tech spec say how it can?)

if you use a `ViewPattern`, the runtime doesn't have to convert the set as many times as there are cases here, and you can use `case` with its completeness warnings, and expand the `otherwise` case:

```suggestion
toAccessRole :: Set AccessRoleV2 -> AccessRole
toAccessRole accessRoles = case Set.fromList accessRoles of
  [] -> NonActivatedAccessRole
  [TeamMemberAccessRole] -> TeamAccessRole
  [...] -> ...
```

libs/wire-api/test/golden/Test/Wire/API/Golden/Generated/Event_user.hs
```diff
@@ -54,9 +55,10 @@ testObject_Event_user_2 =
     (Qualified (Id (fromJust (UUID.fromString "00006a88-0000-2acb-0000-6aa0000061b2"))) (Domain "faraway.example.com"))
     (read "1864-06-05 23:01:18.769 UTC")
     ( EdConvAccessUpdate
-        ( ConversationAccessData
-            { cupAccess = [InviteAccess, LinkAccess, PrivateAccess, InviteAccess, InviteAccess],
-              cupAccessRole = ActivatedAccessRole
+        ( ConversationAccessDataResponse
+            { cadrAccess = [InviteAccess, LinkAccess, PrivateAccess, InviteAccess, InviteAccess],
+              cadrAccessRole = ActivatedAccessRole,
+              cadrAccessRolesV2 = Set.fromList [TeamMemberAccessRole, GuestAccessRole]
```
fisx, line 61: No.  You should keep `ConversationAccessData`, drop `cupAccessRole`, add `AccessRolesV2`, and leave the handling of the old field to the parsers.

services/galley/src/Galley/API/Action.hs
```diff
@@ -359,23 +359,24 @@ instance IsConversationAction ConversationAccessData where
...
+    -- -- Team conversations incur another round of checks
+    -- case convTeam conv of
+    --   Just _ -> do
+    --     -- Access mode change might result in members being removed from the
+    --     -- conversation, so the user must have the necessary permission flag
+    --     ensureActionAllowed RemoveConversationMember self
+    --   Nothing ->
+    --     when (cupAccessRole target == TeamAccessRole) $
+    --       throw InvalidTargetAccess
+    error "todo(leif):implement"
```
fisx, line 379: If at all possible could you try to change the indentation less?  Makes it much easier to read.

services/galley/src/Galley/Data/Conversation/Types.hs
```diff
@@ -37,6 +37,7 @@ data Conversation = Conversation
     convName :: Maybe Text,
     convAccess :: [Access],
     convAccessRole :: AccessRole,
+    convAccessRoles :: Set AccessRoleV2,
```
fisx, line 40: Same here: `convAccessRole` shouldn't be part of the type, the resp. database column should be translated away in the `Cql` instance.

services/galley/src/Galley/Data/Conversation.hs
```diff
@@ -82,20 +84,22 @@ convAccessData :: Conversation -> ConversationAccessData
...
-maybeRole :: ConvType -> Maybe AccessRole -> AccessRole
-maybeRole SelfConv _ = privateRole
-maybeRole ConnectConv _ = privateRole
-maybeRole One2OneConv _ = privateRole
+maybeRole :: ConvType -> Maybe (Set AccessRoleV2) -> Set AccessRoleV2
+maybeRole SelfConv _ = privateAccessRole
+maybeRole ConnectConv _ = privateAccessRole
+maybeRole One2OneConv _ = privateAccessRole
 maybeRole RegularConv Nothing = defRole
 maybeRole RegularConv (Just r) = r
```
fisx, line 97: - remove `defRole`
- rename `maybeRole` to `defRole` and move it to `libs/wire-api/src/Wire/API/Conversation.hs`
- then you don't need to export `{private,team,activated,nonActivated}AccessRole` from there.
- you will need to call the new `defRole` in a few places where you now use `{private,team,activated,nonActivated}AccessRole`, but I think that's ok?

to be more precise:

```suggestion
maybeRole :: ConvType -> Maybe (Set AccessRoleV2) -> Set AccessRoleV2
maybeRole = fromMaybe . defRole

defRole :: ConvType -> Set AccessRoleV2
defRole SelfConv = privateAccessRole
defRole ConnectConv = privateAccessRole
defRole One2OneConv = privateAccessRole
defRole RegularConv = activatedAccessRole
```

services/galley/src/Galley/Data/Conversation.hs
```diff
@@ -82,20 +84,22 @@ convAccessData :: Conversation -> ConversationAccessData
...
-privateRole :: AccessRole
-privateRole = PrivateAccessRole
-
 defRegularConvAccess :: [Access]
 defRegularConvAccess = [InviteAccess]
+
+parseAccessRoles :: Maybe FromAccessRoleLegacy -> Maybe (Set AccessRoleV2) -> Maybe (Set AccessRoleV2)
+parseAccessRoles mbLegacy mbV2 = case mbV2 of
+  Just _ -> mbV2
+  Nothing -> farlAccessRoles <$> mbLegacy
```
fisx, line 105:
```suggestion
parseAccessRoles mbLegacy mbV2 = mbV2 <|> farlAccessRoles <$> mbLegacy
```

libs/wire-api/test/unit/Test/Wire/API/Conversation.hs
```diff
@@ -0,0 +1,50 @@
...
+import Test.Tasty
+import Test.Tasty.QuickCheck (testProperty, (===))
+import Wire.API.Conversation
+
+tests :: TestTree
+tests =
+  testGroup
+    "Conversation"
+    [ fromLegacyToV2ToLegacy,
+      fromV2ToLegacyToV2
```
fisx, line 34:
```suggestion
    [ acessRoleFromLegacyToV2ToLegacy,
      acessRoleFromV2ToLegacyToV2
```

libs/wire-api/src/Wire/API/Conversation.hs
```diff
@@ -421,25 +430,82 @@ typeAccess = Doc.string . Doc.enum $ cs . A.encode <$> [(minBound :: Access) ..]
...
     PrivateAccessRole
   | -- | Team-only conversation
     TeamAccessRole
   | -- | Conversation for users who have activated
-    --   email or phone
+    --   email or phone or SSO
     ActivatedAccessRole
   | -- | No checks
     NonActivatedAccessRole
   deriving stock (Eq, Ord, Show, Generic)
```
fisx, line 444:
```suggestion
  deriving stock (Eq, Ord, Bounded, Enum, Show, Generic)
```

(needed elsewhere in the changes I suggested.)

libs/wire-api/test/unit/Test/Wire/API/Conversation.hs
fisx: (Also, it's probably much faster if we don't run QC here but enumerate all test cases exhaustively.  But I think we should probably resist, and do that more exhaustively once we don't have anything better to do.)

libs/wire-api/test/unit/Test/Wire/API/Conversation.hs
```diff
@@ -0,0 +1,50 @@
...
+      accessRoleLegacy === (toAccessRoleLegacy . fromAccessRoleLegacy) accessRoleLegacy
+
+fromV2ToLegacyToV2 :: TestTree
+fromV2ToLegacyToV2 =
+  testProperty "Access role conversion from v2 to legacy to v2 - original should be a subset of roundtrip converted" p
+  where
+    p originalV2 =
+      let convertedToLegacy = toAccessRoleLegacy originalV2
+          convertedBackToV2 = fromAccessRoleLegacy convertedToLegacy
+       in originalV2 `Set.isSubsetOf` convertedBackToV2
```
fisx, line 50: With something like `[minBound..] \\ [convertedToLegacy..]` you can construct all legacy values for which `originalV2` should be a *superset*.  Then we can make sure that it's actually in between the two legacy values where we expect it to be, instead of just testing one bound.

services/brig/src/Brig/Provider/API.hs
```diff
@@ -816,7 +816,7 @@ addBot zuid zcon cid add = do
     throwStd tooManyMembers
   -- For team conversations: bots are not allowed in managed and in
   -- team-only conversations
-  when (cnvAccessRole cnv == TeamAccessRole) $
+  unless (Set.member ServiceAccessRole (cnvAccessRoles cnv)) $
```
fisx, line 819: This is the reason for the failing tests in brig.

services/galley/src/Galley/API/Action.hs
```diff
@@ -36,6 +36,8 @@ module Galley.API.Action
   )
 where
 
+-- import qualified Brig.Types.User as User
+
```
fisx, line 40:
```suggestion
```

I tend to:
1. Commit only chunks I've looked at again individually (there are lots of good tools for that besides emacs's magit)
2. Remove imports I don't need rather than commenting them out, and if I change my mind get them back by undoing the change in git.

git is nice!  :)

services/galley/src/Galley/API/Action.hs
```diff
@@ -362,20 +364,17 @@ instance IsConversationAction ConversationAccessData where
...
     -- Team conversations incur another round of checks
     case convTeam conv of
       Just _ -> do
         -- Access mode change might result in members being removed from the
         -- conversation, so the user must have the necessary permission flag
         ensureActionAllowed RemoveConversationMember self
       Nothing ->
-        when (cupAccessRole target == TeamAccessRole) $
-          throw InvalidTargetAccess
+        when (Set.fromList [TeamMemberAccessRole] == cupAccessRole target) $ throw InvalidTargetAccess
```
fisx, line 376:
```suggestion
        when (TeamMemberAccessRole `Set.member` cupAccessRole target) $ throw InvalidTargetAccess
```

services/galley/src/Galley/API/Create.hs
```diff
@@ -647,8 +647,8 @@ toUUIDs a b = do
   b' <- U.fromUUID (toUUID b) & note InvalidUUID4
   return (a', b')
 
-accessRole :: NewConv -> AccessRole
-accessRole b = fromMaybe Data.defRole (newConvAccessRole b)
+accessRoles :: NewConv -> Set AccessRoleV2
+accessRoles b = fromMaybe Data.defRole (newConvAccessRoles b)
```
fisx, line 651: If you follow [these suggestions](https://github.com/wireapp/wire-server/pull/2035#discussion_r786602136), you may want to expose `defRole` after all for this.  My mistake, sorry.

services/galley/src/Galley/Data/Conversation/Types.hs
```diff
@@ -51,7 +51,7 @@ data NewConversation = NewConversation
   { ncType :: ConvType,
     ncCreator :: UserId,
     ncAccess :: [Access],
-    ncAccessRole :: AccessRole,
+    ncAccessRole :: Set AccessRoleV2,
```
fisx, line 54:
```suggestion
    ncAccessRoles :: Set AccessRoleV2,
```

fisx: Ah, maybe you didn't add the `s` because it makes the JSON schema harder?  In that case I don't feel so strongly about this.  [Also see here](https://github.com/wireapp/wire-server/pull/2035#discussion_r786634545)

fisx (commented):
...

libs/wire-api/src/Wire/API/Conversation.hs
```diff
@@ -146,18 +171,17 @@ conversationMetadataObjectSchema =
         (description ?~ "The creator's user ID")
         schema
     <*> cnvmAccess .= field "access" (array schema)
-    <*> cnvmAccessRole .= field "access_role" schema
+    <*> cnvmAccessRoles .= accessRolesSchema
```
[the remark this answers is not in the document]
battermann: This is defined in the `accessRolesSchema`. When we deserialize, at least one of them has to be set, and when we serialize, we return both.

libs/wire-api/src/Wire/API/Conversation.hs
```diff
@@ -421,25 +430,82 @@ typeAccess = Doc.string . Doc.enum $ cs . A.encode <$> [(minBound :: Access) ..]
...
+-- todo(leif): add docs
+instance ToSchema AccessRoleV2 where
+  schema =
+    (S.schema . description ?~ "Which users/services can join conversations") $
+      enum @Text "AccessRoleV2" $
+        mconcat
+          [ element "team_member" TeamMemberAccessRole,
+            element "non_team_member" NonTeamMemberAccessRole,
+            element "guest" GuestAccessRole,
+            element "service" ServiceAccessRole
```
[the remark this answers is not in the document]
battermann: Will do that.

libs/wire-api/src/Wire/API/Conversation.hs
```diff
@@ -421,25 +430,82 @@ typeAccess = Doc.string . Doc.enum $ cs . A.encode <$> [(minBound :: Access) ..]
...
+    accessRolesLegacyAsc =
+      [ (PrivateAccessRole, privateAccessRole),
+        (TeamAccessRole, teamAccessRole),
+        (ActivatedAccessRole, activatedAccessRole),
+        (NonActivatedAccessRole, nonActivatedAccessRole)
+      ]
+
+-- | Wrapper around `Set AccessRoleV2` for Cassandra Cql instance
+-- that converts legacy access role into access role V2
+newtype FromAccessRoleLegacy = FromAccessRoleLegacy {farlAccessRoles :: Set.Set AccessRoleV2}
```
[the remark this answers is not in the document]
battermann: :+1:

libs/wire-api/src/Wire/API/Conversation.hs
```diff
@@ -421,25 +430,82 @@ typeAccess = Doc.string . Doc.enum $ cs . A.encode <$> [(minBound :: Access) ..]
...
+  maybe NonActivatedAccessRole fst $ find (allMember accessRoles . snd) accessRolesLegacyAsc
+  where
+    allMember :: Ord a => Set a -> Set a -> Bool
+    allMember lhs rhs = all (`Set.member` rhs) lhs
+    accessRolesLegacyAsc =
+      [ (PrivateAccessRole, privateAccessRole),
+        (TeamAccessRole, teamAccessRole),
+        (ActivatedAccessRole, activatedAccessRole),
+        (NonActivatedAccessRole, nonActivatedAccessRole)
+      ]
```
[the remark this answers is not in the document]
battermann: nice

libs/wire-api/test/unit/Test/Wire/API/Conversation.hs
battermann: I like the renaming.
Regarding QC, all wire-api unit tests together take 1.5s to run. I don't see a problem.
Checking properties has the benefit that if we add additional access roles later, they are automatically checked.
We can't ensure exhaustiveness on sets anyway, AFAIK.

services/brig/src/Brig/Provider/API.hs
battermann: Yes, this is fixed now after adding `ServiceAccessRole` to the mapping of `ActivedAccessRole`.

services/galley/src/Galley/API/Action.hs
battermann: I think the original logic is correct, I added a comment.

services/galley/test/integration/API.hs
```diff
@@ -1210,7 +1210,7 @@ testGetCodeRejectedIfGuestLinksDisabled = do
   galley <- view tsGalley
   (owner, teamId, []) <- Util.createBindingTeamWithNMembers 0
   let createConvWithGuestLink = do
-        convId <- decodeConvId <$> postTeamConv teamId owner [] (Just "testConversation") [CodeAccess] (Just ActivatedAccessRole) Nothing
+        convId <- decodeConvId <$> postTeamConv teamId owner [] (Just "testConversation") [CodeAccess] (Just (Set.fromList [TeamMemberAccessRole, NonTeamMemberAccessRole])) Nothing
```
fisx, line 1213: this looks like the module where you can test the new cases (in particular services-not-guests, and guests-not-services).  maybe this can be even randomized somehow?

services/galley/src/Galley/API/Action.hs
```diff
@@ -401,38 +401,50 @@ instance IsConversationAction ConversationAccessData where
...
+          then pure bm
+          else pure $ bm {bmBots = mempty}
+
+      maybeRemoveGuests :: Member BrigAccess r => BotsAndMembers -> Sem r BotsAndMembers
+      maybeRemoveGuests bm =
+        if Set.member GuestAccessRole (cupAccessRoles action)
+          then pure bm
+          else do
+            activated <- map User.userId <$> E.lookupActivatedUsers (toList (bmLocals bm))
+            -- FUTUREWORK: should we also remove non-activated remote users?
```
fisx, line 426: Federation people: can users joining via a guest link without account be non-local?

services/galley/src/Galley/API/Util.hs
```diff
@@ -74,19 +74,18 @@ type JSON = Media "application" "json"
...
   [(UserId, Maybe TeamMember)] ->
   Sem r ()
-ensureAccessRole role users = case role of
-  PrivateAccessRole -> throw ConvAccessDenied
-  TeamAccessRole ->
-    when (any (isNothing . snd) users) $
-      throwED @NotATeamMember
-  ActivatedAccessRole -> do
+ensureAccessRole roles users = do
+  unless (Set.member TeamMemberAccessRole roles) $ throw ConvAccessDenied
```
fisx, line 81:
```suggestion
  when (Set.null roles) $ throw ConvAccessDenied
```

services/galley/src/Galley/API/Util.hs
```diff
@@ -74,19 +74,18 @@ type JSON = Media "application" "json"
...
+  unless (Set.member NonTeamMemberAccessRole roles) $
+    when (any (isNothing . snd) users) $ throwED @NotATeamMember
+  unless (Set.member GuestAccessRole roles) $ do
     activated <- lookupActivatedUsers $ map fst users
-    when (length activated /= length users) $
-      throw ConvAccessDenied
-  NonActivatedAccessRole -> return ()
+    when (length activated /= length users) $ throw ConvAccessDenied
+  -- todo(leif): check if we need to check users for bots/services
+  unless (Set.member ServiceAccessRole roles) $ return ()
```
fisx, line 88: yes, you've collapsed both filters into the first.  try this:

```haskell
  unless (Set.member GuestAccessRole roles && Set.member ServiceAccessRole roles) $ do
    activated <- lookupActivatedUsers $ map fst users
    when (length activated /= length users) $ do
      let guestOrBotIds = map fst users \\ (userId <$> activated)
      guestOrBot <- lookupUsers guestOrBotIds
      let (guests, bots) = partition (isJust . userService) guestOrBot
      unless (null bots || Set.member ServiceAccessRole roles) $ throw ConvAccessDenied
      unless (null guests || Set.member GuestAccessRole roles) $ throw ConvAccessDenied
```

fisx: ah, you can't run `lookupUsers` from galley.  so, more pseudo-code:

```
   get "/i/users" (continue listActivatedAccountsH) $
     accept "application" "json"
       .&. (param "ids" ||| param "handles")
+      .&. def True (flag "active_only")
       .&. def False (query "includePendingInvitations")
```

libs/wire-api/src/Wire/API/Conversation.hs
```diff
@@ -500,7 +500,15 @@ instance ToSchema AccessRoleV2 where
...
-        \This replaces the deprecated field `access_role`."
+        \This replaces the deprecated field `access_role`\
+        \and allows for a more fine grained configuration of access roles\
+        \in particular a separation of guest and services access.\
+        \Maps to `access_role` as follows:\
+        \`[]` => `private` - nobody can be invited to this conversation (e.g. it's a 1:1 conversation)\
+        \`[team_member]` => `team` - team-only conversation\
+        \`[team_member, non_team_member, service]` => `activated` - conversation for users who have activated email, phone or SSO and services\
+        \`[team_member, non_team_member, service, guest]` => `non_activated` - all allowed, no checks.\
+        \All other configurations of `access_role_v2` are mapped to the smallest superset containing all given access roles."
```
fisx, line 511: Would it make more sense to document the mapping in the deprecated type, then they will be deleted together?

changelog.d/0-release-notes/pr-2035
```diff
@@ -0,0 +1 @@
+Backend now separates conversation access control for guests and services. The old access roles are still supported but it is encouraged to upgrade clients since mapping between the old access roles and the new access roles are is not isomorphic. For more details refer to the API changes changelog, the Swagger docs, or the technical specification.
```
fisx, line 1: If we already know which versions of clients support this, we should add those here, but I suspect we're not certain yet.  Bring up in the next standup?

fisx (approved):
#2057 should (does?) provide some tests that are essential for #2045, but you may have tested this manually?  otherwise LGTM!

PR #2037, opened 2022-01-11
Fix Remote Asset Downloading via Better Resource Handling

pcapriotti:
Client teams reported that via PR #2004 merged, they cannot download remote assets of size 20 KB and larger. This fixes the issue by addressing incorrect resource handling. The limitation is that if an error happens before streaming starts, it will result in an empty response. @pcapriotti is working on addressing that case in a follow-up PR.

## Checklist


mdimjasevic (approved):
I'll have to do some reading on this resource handling to better understand this, but this looks good.

PR #2041, opened 2022-01-13
 Remove sft_servers_all in non-federated setting

smatting:
## Checklist

https://wearezeta.atlassian.net/browse/FS-338


jschaul (commented):
Not sure this is a good idea as it stands, see comments on the jira issue.

smatting:
decided not to do this

PR #2044, opened 2022-01-14
cabal-run-integration.sh - remove Makefile indirection

smatting:
This PR improves `cabal-run-integration.sh`:

1. calls `integration.sh` directly instead of calling the Makefile of the service to-be-tested. This makes passing test pattern arguments such as ``-p '$2 == "provider" && $3 == "account"'`` feasable which was almost impossible with gnumake's escaping rules.
2. Allows passing any argument to the integration test binary, not just patterns

## Checklist


pcapriotti (approved):
Nice!

PR #2045, opened 2022-01-14
[FS-333] Tag Test Cases for M2

mdimjasevic:
This implements https://wearezeta.atlassian.net/browse/FS-333. 

Still to do:

- [x] Tag a test case for getting a conversation

## Checklist

### DIFF
--- a/services/federator/test/unit/Test/Federator/Remote.hs
+++ b/services/federator/test/unit/Test/Federator/Remote.hs
@@ -120,6 +120,11 @@ testValidatesCertificateSuccess =
             Right _ -> assertFailure "Congratulations, you fixed a known issue!"
     ]
 
+-- @SF.Federation @TSFI.RESTfulAPI @S2
+--
+-- This is a group of test cases where refusing to connect with the server is
+-- checked. The second test case refuses to connect with a server when it has no
+-- X509v3 Extended Key Usage extension in the certificate.
 testValidatesCertificateWrongHostname :: TestTree
 testValidatesCertificateWrongHostname =
   testGroup
@@ -132,10 +137,6 @@ testValidatesCertificateWrongHostname =
             Left (RemoteError _ (FederatorClientTLSException _)) -> pure ()
             Left x -> assertFailure $ "Expected TLS failure, got: " <> show x
             Right _ -> assertFailure "Expected connection with the server to fail",
-      -- @SF.Federation @TSFI.RESTfulAPI @S2
-      --
-      -- Refuse to connect with a server when it has no X509v3 Extended Key
-      -- Usage extension in the certificate.
       testCase "when the server's certificate does not have the server key usage flag" $
         withMockServer certWithoutServerKeyUsage $ \port -> do
           tlsSettings <- mkTLSSettingsOrThrow settings
@@ -145,6 +146,8 @@ testValidatesCertificateWrongHostname =
             Left x -> assertFailure $ "Expected TLS failure, got: " <> show x
             Right _ -> assertFailure "Expected connection with the server to fail"
     ]
+
+-- @END
 
 testConnectionError :: TestTree
 testConnectionError = testCase "connection failures are reported correctly" $ do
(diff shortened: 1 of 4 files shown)


services/federator/test/unit/Test/Federator/Remote.hs
```diff
@@ -132,6 +132,10 @@ testValidatesCertificateWrongHostname =
             Left (RemoteError _ (FederatorClientTLSException _)) -> pure ()
             Left x -> assertFailure $ "Expected TLS failure, got: " <> show x
             Right _ -> assertFailure "Expected connection with the server to fail",
+      -- @SF.Federation @TSFI.RESTfulAPI @S2
+      --
+      -- Refuse to connect with a server when it has no X509v3 Extended Key
+      -- Usage extension in the certificate.
```
mythsunwind, line 138: The END tag is missing here

mythsunwind (changes requested):
One END tag is missing. Otherwise great work nicely documented with your comments!

services/federator/test/unit/Test/Federator/Remote.hs
mdimjasevic: In that earlier ticket and PR, I did not have an END tag for a nested test case like this. If I try to add it here, because of automatic formatting we have in the `wire-server` codebase it gets weird indentation and it does not match indenting of that introductory @-tag: instead of 6 spaces, it gets 12 spaces. There is no way around it with this automatic formatting in place as far as I know.

Therefore, can you make it work without the @END tag? If not, I'll see what I can do with this automatic formatting.

mythsunwind: Unfortunately the parser does not work without the END tag. What would happen if you add the comment in front of the line `testValidatesCertificateWrongHostname`? Would it work to then with the automatic formatting?

mdimjasevic: > Unfortunately the parser does not work without the END tag. What would happen if you add the comment in front of the line `testValidatesCertificateWrongHostname`? Would it work to then with the automatic formatting?

Sure, I've placed the tags around the top-level code. Let me know if this cuts it.

mythsunwind: Works like a charm! Thanks @mdimjasevic

mythsunwind (approved):

pcapriotti (changes requested):

services/federator/test/unit/Test/Federator/Remote.hs
```diff
@@ -120,6 +120,11 @@ testValidatesCertificateSuccess =
             Right _ -> assertFailure "Congratulations, you fixed a known issue!"
     ]
 
+-- @SF.Federation @TSFI.RESTfulAPI @S2
+--
+-- This is a group of test cases where refusing to connect with the server is
+-- checked. The second test case refuses to connect with a server when it has no
+-- X509v3 Extended Key Usage extension in the certificate.
```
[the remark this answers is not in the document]
mdimjasevic: I'm confused. Are we talking about this test:

```haskell
      testCase "when the server's certificate does not have the server key usage flag" $
```

This test case uses a certificate that does not have "TLS Web Server Authentication" as one of the extensions. Or is my added comment in lines 126-127 wrong?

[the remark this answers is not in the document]
pcapriotti: You wrote it "refuses to connect [...] when it has no X509V3 Extended Key Usage extension". This is not correct as stated. I think you are confusing what *extensions* a certificate has with what *purposes* the extended key usage extension lists (see https://datatracker.ietf.org/doc/html/rfc2459#section-4.2.1.13 for more details).

Maybe write something like "[...] when it the extended key usage extension, if present, does not list `ServerAuth` (or whatever its official name is) among the purposes".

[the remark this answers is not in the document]
mdimjasevic: Yes, you're right, I was confusing the presence of an extension with the purposes the extension lists! I'll fix that now.

pcapriotti (approved):

PR #2061, opened 2022-01-20
Fix HLint issues in Gundeck

supersven:
Fixing these issues has two benefits:
- The code is (hopefully) easier to read.
- HLint issues that might indicate errors are easier to spot. (Less "noise" in HLint's output.)

## Checklist


services/gundeck/test/integration/API.hs
```diff
@@ -71,54 +71,54 @@ tests s =
...
-        [ test s "Register a user" $ addUser,
-          test s "Delete a user" $ removeUser,
-          test s "Replace presence" $ replacePresence,
-          test s "Remove stale presence" $ removeStalePresence,
-          test s "Single user push" $ singleUserPush,
+        [ test s "Register a user" addUser,
+          test s "Delete a user" removeUser,
+          test s "Replace presence" replacePresence,
+          test s "Remove stale presence" removeStalePresence,
+          test s "Single user push" singleUserPush,
```
fisx, line 78: I don't like this change, because I'm obsessed with symmetry, but I like HLint too much to object.  :)

services/gundeck/test/unit/Push.hs
```diff
@@ -34,7 +33,8 @@ import Test.Tasty.QuickCheck
 
 tests :: TestTree
 tests =
-  testGroup "bulkpush" $
+  testGroup
+    "bulkpush"
```
fisx, line 37: Same here.  (The use of `$` here is to keep the first two arguments in the same line.)

fisx (approved):
HLInt is nice!

fisx:
@supersven do you want to merge this?

PR #2065, opened 2022-01-25
Fix (some) HLint issues in wire-api

supersven:
Fixing these issues has two benefits:

- The code is (hopefully) easier to read.
- HLint issues that might indicate errors are easier to spot. (Less "noise" in HLint's output.)

## Checklist

### DIFF
--- a/changelog.d/5-internal/fix-hlint-issues-in-wire-api
+++ b/changelog.d/5-internal/fix-hlint-issues-in-wire-api
@@ -1,0 +1,1 @@
+Fix some HLint issues in libs/wire-api.
--- a/libs/wire-api/src/Wire/API/Arbitrary.hs
+++ b/libs/wire-api/src/Wire/API/Arbitrary.hs
@@ -1,6 +1,5 @@
-{-# LANGUAGE GeneralizedNewtypeDeriving #-}
+{-# LANGUAGE DisambiguateRecordFields #-}
 {-# LANGUAGE KindSignatures #-}
-{-# LANGUAGE RecordWildCards #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# OPTIONS_GHC -fno-warn-orphans #-}
 
--- a/libs/wire-api/src/Wire/API/Conversation/Member.hs
+++ b/libs/wire-api/src/Wire/API/Conversation/Member.hs
@@ -252,13 +252,12 @@ instance Arbitrary MemberUpdate where
 
 validateMemberUpdate :: MemberUpdate -> Either String MemberUpdate
 validateMemberUpdate u =
-  if ( isJust (mupOtrMuteStatus u)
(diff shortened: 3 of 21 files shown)


libs/wire-api/src/Wire/API/Arbitrary.hs
```diff
@@ -1,6 +1,5 @@
-{-# LANGUAGE GeneralizedNewtypeDeriving #-}
+{-# LANGUAGE DisambiguateRecordFields #-}
```
fisx, line 1: oops?  why did you turn this on if it obviously compiled before?

fisx: ah, probably some overlap with RecordWildCards?

libs/wire-api/src/Wire/API/Conversation.hs
```diff
@@ -153,11 +153,10 @@ conversationMetadataObjectSchema =
...
     <*> cnvmMessageTimer
-      .= ( optFieldWithDocModifier
-             "message_timer"
-             (description ?~ "Per-conversation message timer (can be null)")
-             (maybeWithDefault A.Null schema)
-         )
+      .= optFieldWithDocModifier
+        "message_timer"
+        (description ?~ "Per-conversation message timer (can be null)")
+        (maybeWithDefault A.Null schema)
```
fisx, line 159: @battermann and this is why you should not mix hlint changes and feature work in the same PR!  :)  (see #2035)

fisx (approved):

changelog.d/5-internal/fix-hlint-issues-in-wire-api
```diff
@@ -0,0 +1 @@
+Fix HLint issues in libs/wire-api. There are a lot left, but these ones should be safe to not change the application's behavior.
```
pcapriotti, line 1: I would just write "Fix some HLint issues in wire-api". Your phrasing does not seem appropriate for the release notes, but feel free to use that text as commit message for the merge.

pcapriotti (approved):
Looks good.

libs/wire-api/src/Wire/API/Arbitrary.hs
supersven: Indeed, `DisambiguateRecordFields` is implied by `RecordWildCards`, but `DisambiguateRecordFields` is sufficient here.

Probably, this doesn't buy us much, but reducing HLint warnings helps to make the more important ones easier to spot...

changelog.d/5-internal/fix-hlint-issues-in-wire-api
supersven: Yes, that sounds reasonable. :+1:

supersven: (Fixed according to the comment)

PR #2076, opened 2022-01-27
SQSERVICES 1174 enforce access roles when adding a participant or generating a guest link

battermann:
https://wearezeta.atlassian.net/browse/SQSERVICES-1174

## Checklist

### DIFF
--- a/changelog.d/0-release-notes/pr-2076
+++ b/changelog.d/0-release-notes/pr-2076
@@ -1,0 +1,1 @@
+Enforce conversation access roles more tightly on the backend: if a guests are not allowed, block guest link creation (new behavior) as well as ephemeral users joining (old behavior).
--- a/changelog.d/0-release-notes/pr-2076
+++ b/changelog.d/0-release-notes/pr-2076
@@ -1,1 +1,1 @@
-Enforce conversation access roles more tightly on the backend: if a guests are not allowed, block guest link creation (new behavior) as well as ephemeral users joining (old behavior).
+Enforce conversation access roles more tightly on the backend: if a guests or non-team-members are not allowed, block guest link creation (new behavior) as well as ephemeral users joining (old behavior).
--- a/changelog.d/2-features/pr-2076
+++ b/changelog.d/2-features/pr-2076
@@ -1,1 +1,1 @@
-A guest link cannot be generated if the conversation access role doesn’t allow guests.
+Enforce conv access roles more tightly on the backend: if a guests are not allowed, block guest link creation (new behavior) as well as ephemeral users joining (old behavior).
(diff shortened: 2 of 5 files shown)


changelog.d/2-features/pr-2076
```diff
@@ -0,0 +1 @@
+A guest link cannot be generated if the conversation access role doesn’t allow guests.
```
fisx, line 1:
```suggestion
Enforce conv access roles more tightly on the backend: if a guests are not allowed, block guest link creation (new behavior) as well as ephemeral users joining (old behavior).
```

I would also put this either in bug fixes (it's not really a feature) or in release notes (it's not really a bug, just something to be aware of in general, just in case).  What do you think?

services/galley/src/Galley/API/Update.hs
```diff
@@ -586,6 +587,9 @@ addCode lusr zcon lcnv = do
     createCode :: Code -> Sem r ConversationCode
     createCode code = do
       mkConversationCode (codeKey code) (codeValue code) <$> E.getConversationCodeURI
+    ensureGuestsAllowed :: Data.Conversation -> Sem r ()
+    ensureGuestsAllowed conv =
+      unless (GuestAccessRole `Set.member` convAccessRoles conv) $ throw ConvAccessDenied
```
fisx, line 592: Or `NonTeamMemberAccessRole`.  Either is enough to allow creating codes, since either group may want one.

Given this change, please go through all additions of the guest access role again and see if it's necessary.

fisx (dismissed):

services/galley/src/Galley/API/Update.hs
fisx: never mind, i was still confused.  let's keep it like it is!

fisx: Ok, I remember now: the logical thing to do would be to require "either NonTeamMember or Guests" in order to allow this.  This would also be consistent with clients who consistently only ever set both or none, but it would allow future clients to change this into something more flexible, ie., two buttons for "allow ephemeral", "allow non-team".  Then they could keep using the existing backend and we wouldn't have to touch this again.

So:

```suggestion
      unless (GuestAccessRole `Set.member` convAccessRoles conv || NonTeamMemberAccessRole `Set.member` convAccessRoles conv) $ throw ConvAccessDenied
```

changelog.d/0-release-notes/pr-2076
```diff
@@ -0,0 +1 @@
+Enforce conversation access roles more tightly on the backend: if a guests are not allowed, block guest link creation (new behavior) as well as ephemeral users joining (old behavior).
```
fisx, line 1: :+1:

pcapriotti (approved):
Looks good to me.

PR #2077, opened 2022-01-27
Fix bug: Calls config crashes when SFTs are unreachable

smatting:
Followup to https://github.com/wireapp/wire-server/pull/2030

## Checklist

 - [x] The **PR Title** explains the impact of the change.
 - [x] The **PR description** provides context as to why the change should occur and what the code contributes to that effect. This could also be a link to a JIRA ticket or a Github issue, if there is one.
 - [x] **changelog.d** contains the following bits of information ([details](https://github.com/wireapp/wire-server/blob/develop/docs/developer/changelog.md)):
   - [x] A file with the changelog entry in one or more suitable sub-sections. The sub-sections are marked by directories inside

PR #2062, opened 2022-01-23
Add licenses to new Spar work

isovector:
This PR calls `make add-license` for all of my work on Spar polysemization.

## Checklist

 - [x] The **PR Title** explains the impact of the change.
 - [x] The **PR description** provides context as to why the change should occur and what the code contributes to that effect. This could also be a link to a JIRA ticket or a Github issue, if there is one.
 - [x] **changelog.d** contains the following bits of information ([details](https://github.com/wireapp/wire-server/blob/develop/docs/developer/changelog.md)):
   - [x] A file with the changelog entry in one or more suitable sub-sections. The sub-sections are marked by directories inside `changelog.d`.

PR #2080, opened 2022-01-27
Release 2022-01-27

fisx:
## Release notes

* The `nginz` chart now configures nginx to only allow cross-origin requests from an explicit allow list of subdomains. By default these are:

  ```yaml
  nginz:
    nginx_conf:
      allowlisted_origins:
      - webapp
      - teams
      - account
  ```

  If you changed the names of these services, you must adjust those names in the nginz config as well. (#1630)

* Backend now separates conversation access control for guests and services. The old access roles are still supported but it is encouraged to upgrade clients since mapping between the old access roles and the new access roles is not isomorphic. For more details refer to the API changes below or the Swagger docs.
  Old clients are fully supported; if new clients and old clients are mixed, to old clients, either guests of services may appear to be enable if they are not, which may lead to error messages (confusing but harmless). (#2035)

## API changes

* Endpoints that recently have accepted `access_role` in their payload will now accept `access_role_v2` as well which will take precedence over `access_role`. See Swagger docs for how values are mapped. Endpoints that recently have returned `access_role` in their payload will now additionally return the `access_role_v2` field. (#2035)

## Features

* Conversation access roles now distinguish between guests and services. (#2035)

* There is now an explicit CORS allow list for *all* endpoints. In previous releases, all subdomains were accepted, however they must now be listed explicitly. This is a **breaking change**, as now only known Javascript applications may access the backend. (#1630)
* Prevent 500s when SFTs are not reachable from Backend (#2077)

## Internal changes

* Bump hsaml2 package version (#2075)
* Separate Spar.Data module into smaller Cassandra interpreters (#2064)
* Fix some HLint issues in libs/wire-api. (#2065)
* Fix broken build process of package "old-time" for some environments (#2056)
* Refresh license headers (#2062)
* Rename Spar.Sem.ScimTokenStore.GetByTeam to LookupByTeam (#2068)
* (Try syntax change in config file that breaks nginz (#2073, reverted in a4a6193f9494))

## Federation changes

* Tag several federation tests cases for the M2 release (#2045)


julialongtin (approved):

charts/nginz/templates/conf/_nginx.conf.tpl
```diff
@@ -125,7 +125,9 @@ http {
 
   map $http_origin $cors_header {
       default "";
-      "~^https://([^/]+\.)?{{ .Values.nginx_conf.external_env_domain | replace "." "\\." }}(:[0-9]{2,5})?$" "$http_origin";
+    {{ range $origin := .Values.nginx_conf.allowlisted_origins }}
+      "https://{{ $origin }}.{{ .Values.nginx_conf.external_env_domain}} "$http_origin";
```
julialongtin, line 129:
```suggestion
      "https://{{ $origin }}.{{ .Values.nginx_conf.external_env_domain}}" "$http_origin";
```

fisx: 116988c  (i also kept the `$` from #2073)

smatting: Sorry for that :/

PR #2081, opened 2022-01-28
Servantify cannon internal api

supersven:
Second try (first one washttps://github.com/wireapp/wire-server/pull/2029, reverted by https://github.com/wireapp/wire-server/pull/2071).

The issue with the first PR was, that only the first chunk of the notification message stream was sent. This is now solved by consuming the whole stream with `sinkLazy` first and then sending it via `sendBinaryData`.

I've added an integration test for large notifications (resulting in internally chunked streams) and manually tested the endpoint by sending and receiving big notifications to and from (respectively) Cannon. E.g. Goethe's Faust was completely sent and received (https://www.gutenberg.org/files/63203/63203-0.txt). 

## Checklist

 - [X] **changelog.d** contains the following bits of information ([details](https://github.com/wireapp/wire-server/blob/develop/docs/developer/changelog.md)):


pcapriotti (approved):
Looks good.

fisx:
@supersven @pcapriotti for next time: i would prefer to see work on the internal end-points only once we've converted the complete public API.  people are waiting for the swagger docs of the latter, the former is just for us.  (ok, "just".  of course it's nice to have a service completely covered.)

PR #2084, opened 2022-01-28
SQSERVICES-1057 Ensure guest links enabled when join by code

battermann:


mdimjasevic:
@battermann , based on the PR's title, I'd say it implements SQSERVICES-1057. The PR's title and the ticket don't seem to line up. Would you mind clarifying, i.e., what is this PR about?

battermann:
> @battermann , based on the PR's title, I'd say it implements SQSERVICES-1057. The PR's title and the ticket don't seem to line up. Would you mind clarifying, i.e., what is this PR about?

Sorry, Matthias who usually reviews has all the context. It is related to SQSERVICES-1057 as it fixes a bug (there is no extra bug ticket). The requrement is that if the guest link team feature is disabled, it should not be possible to join a conversation by code (which was generated while the feature was previously enabled). This PR provides the implementation plus test.

mdimjasevic (dismissed):
I think there is a subtlety that makes this wrong. See the inlined comment.

Any way, it looks like we're missing a test case that would demonstrate why the change is wrong as it is now: if user A invites user B to join a conversation when the guest link feature is disable for the conversation (i.e., other means than by code), user B should be able to join. Can you please check if we have such a test case and add it if it is not there?

mdimjasevic:
@battermann , I am not sure if you've seen my comment https://github.com/wireapp/wire-server/pull/2084#discussion_r798443783. That is the main reason I asked for a change in the PR. Do you think the concern I raised there is valid?

battermann:
> @battermann , I am not sure if you've seen my comment [#2084 (comment)](https://github.com/wireapp/wire-server/pull/2084#discussion_r798443783). That is the main reason I asked for a change in the PR. Do you think the concern I raised there is valid?

@mdimjasevic Yes, your concern is valid. The affected implementation covers two ways of joining a conversation, by code and by id. We do not want to allow someone joining by code if the feature is disabled. I am currently evaluating if joining by id should also be forbidden if the feature is disabled. If that is the case, there should not be any concerns regarding this change. Does this answer your question?

mdimjasevic:
> > @battermann , I am not sure if you've seen my comment [#2084 (comment)](https://github.com/wireapp/wire-server/pull/2084#discussion_r798443783). That is the main reason I asked for a change in the PR. Do you think the concern I raised there is valid?
> 
> @mdimjasevic Yes, your concern is valid. The affected implementation covers two ways of joining a conversation, by code and by id. We do not want to allow someone joining by code if the feature is disabled. I am currently evaluating if joining by id should also be forbidden if the feature is disabled. If that is the case, there should not be any concerns regarding this change. Does this answer your question?

All I'm saying is that, just like in the first review, the check should be in the join-by-code-specific function, and not in a general one that covers all the ways to join a conversation (see the `Wire.API.Conversation.Access` type).

battermann:
I think there are two things to figure out:

1. Do we want to enforce the guest links team feature when the endpoint `POST /conversation/:convId/join` is called as well?
2. Do we need to enforce the feature when `POST /conversations/code-check` (https://github.com/wireapp/wire-server/blob/develop/services/galley/src/Galley/API/Public/Servant.hs#L54) is called?

Regarding 1. according to @johnxnguyen this endpoint is not used in the iOS client and according to Timothy it is alos not used anywhere in the web client. It is possible that this endpoint is not used at all and can be deprecated (that would be beyond the scope of this feature/PR). The question remains if we should extend the check to this endpoint or not? I would say, yes.

Regarding 2. I think that in case guest links are disabled for the team/server, we should have the `code-check` endpoint to return an error (probably 404 CodeNotFound which is the same error that is returned if conversation code access was revoked).

mdimjasevic:
> Regarding 1. according to @johnxnguyen this endpoint is not used in the iOS client and according to Timothy it is alos not used anywhere in the web client. It is possible that this endpoint is not used at all and can be deprecated (that would be beyond the scope of this feature/PR). The question remains if we should extend the check to this endpoint or not? I would say, yes, and that is how it is currently implemented.

I don't see how the `/conversations/:convId/join` endpoint is related to guest links. It could be they are related, but I don't see how.

> Regarding 2. I think that in case guest links are disabled for the team/server, we should have the `code-check` endpoint to return an error (probably 404 CodeNotFound which is the same error that is returned if conversation code access was revoked).

That's a good point. There is a function `Galley.API.Util.verifyReusableCode` that actually checks if a code is valid. It's called from `Galley.API.Query.getConversationByReusableCode`. In the later function you already have a check if the guest links feature is enabled so I am not sure what else is there to add. If the feature is not enabled, I would throw `CodeNotFound` too, i.e., I wouldn't reveal to the API user that the feature is disabled. Currently, you have that `GuestLinksDisabled :: ConversationError` error that you've introduced, but I'm inclined to suggest to get rid of that data constructor and make `Galley.API.Query.ensureGuestLinksDisabled` have `Error CodeError` as a constraint instead of `Error ConversationError` and throw `CodeNotFound` if the feature is disabled.

battermann:
> That's a good point. There is a function Galley.API.Util.verifyReusableCode that actually checks if a code is valid. It's called from Galley.API.Query.getConversationByReusableCode. In the later function you already have a check if the guest links feature is enabled so I am not sure what else is there to add. If the feature is not enabled, I would throw CodeNotFound too, i.e., I wouldn't reveal to the API user that the feature is disabled. Currently, you have that GuestLinksDisabled :: ConversationError error that you've introduced, but I'm inclined to suggest to get rid of that data constructor and make Galley.API.Query.ensureGuestLinksDisabled have Error CodeError as a constraint instead of Error ConversationError and throw CodeNotFound if the feature is disabled.

I completely agree. To respond with `CodeNotFound` was my initial idea, but it was decided to respond with 409 (https://wearezeta.atlassian.net/wiki/spaces/SER/pages/537756256/Tech+spec+Guest-link+feature+config#Failure-response). It would make things easier if we would return 404 because it would not involve any client changes, and it would not leak any  information to the API use as you mentioned.

services/galley/src/Galley/API/Query.hs
```diff
@@ -541,22 +544,37 @@ getConversationByReusableCode lusr key value = do
...
-ensureGuestLinksEnabled ::
+ensureGuestLinksEnabledWithError ::
   forall r.
   ( Member (Error ConversationError) r,
+    Member (Error CodeError) r,
     Member TeamFeatureStore r,
     Member (Input Opts) r
   ) =>
-  Data.Conversation ->
+  Either ConversationError CodeError ->
```
fisx, line 554:
```suggestion
  Either ConversationError CodeError {- FUTUREWORK: "a `Member` [ConversationError, CodeError] => e"? -} ->
```

this may work or not, and i don't think we should spend the time to sort this out now, but if(f) you make any more commits you might as well leave this comment?

services/galley/test/integration/API.hs
```diff
@@ -1250,32 +1251,55 @@ testJoinTeamConvGuestLinksDisabled = do
...
-  -- works by default
-  getJoinCodeConv userNotInTeam (conversationKey cCode) (conversationCode cCode) !!! do
+  let checkFeatureStatus fstatus =
+        Util.getTeamFeatureFlagWithGalley Public.TeamFeatureGuestLinks galley owner teamId !!! do
+          const 200 === statusCode
+          const (Right (Public.TeamFeatureStatusNoConfigAndLockStatus fstatus Public.Unlocked)) === responseJsonEither
+
+  -- guest can join if guest link feature is enabled
+  checkFeatureStatus Public.TeamFeatureEnabled
+  getJoinCodeConv eve (conversationKey cCode) (conversationCode cCode) !!! do
```
fisx, line 1267: also try out bob here?

fisx (approved):

libs/wire-api/src/Wire/API/Routes/Public/Galley.hs
```diff
@@ -333,8 +337,12 @@ type ConversationAPI =
            )
     :<|> Named
            "code-check"
-           ( Summary "Check validity of a conversation code"
+           ( Summary
+               "Check validity of a conversation code.\
+               \If the guest links team feature is disabled, this will fail with 404 CodeNotFound.\
+               \Note that this is currently inconsistent (for backwards compatibility reasons) with `POST /conversations/join` which responds with 409 GuestLinksDisabled if guest links are disabled."
                :> CanThrow CodeNotFound
+               :> CanThrow ConvNotFound
```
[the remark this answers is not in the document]
fisx: @battermann as discussed off-line: this is thrown because the handler now does a conversation lookup where it didn't have to before, so this is all in order.  (which might mean that clients will have to make a trivial adjustment to their error handler.)

fisx (approved):
good to go!  (all changes suggested in earlier review are optional)

fisx:
@mdimjasevic I think your concerns have been handled?  (tl;dr: the other end-points affected by the changes here are dead code and will be removed soon: https://wearezeta.atlassian.net/browse/SQSERVICES-1290

PR #2059, opened 2022-01-20
Sqservices 1168 add lockstatus to feature configs

battermann:
https://wearezeta.atlassian.net/browse/SQSERVICES-1168

## Checklist

 - [x] The **PR Title** explains the impact of the change.
 - [x] The **PR description** provides context as to why the change should occur and what the code contributes to that effect. This could also be a link to a JIRA ticket or a Github issue, if there is one.
 - [x] If HTTP endpoint paths have been added or renamed, the **endpoint / config-flag checklist** (see Wire-employee only backend [wiki page](https://github.com/zinfra/backend-wiki/wiki/Checklists)) has been followed.
 - [x] If a cassandra schema migration has been added, I ran **`make git-add-cassandra-schema`** to update the cassandra schema documentation.
 - [x] **changelog.d** contains the following bits of information ([details](https://github.com/wireapp/wire-server/blob/develop/docs/developer/changelog.md)):
   - [x] A file with the changelog entry in one or more suitable sub-sections. The sub-sections are marked by directories inside `changelog.d`.
   - [x] If new config options introduced: added usage description under docs/reference/config-options.md
   - [x] If new config options introduced: recommended measures to be taken by on-premise instance operators.

PR #2086, opened 2022-01-28
Cleanup internal endpoints

pcapriotti:
This moves internal endpoints to the `Internal` module hierarchy in `wire-api`, so that they are not included in the generated Swagger, and removes usages of servant-generic from Galley's internal API.

## Checklist

### DIFF
--- a/libs/wire-api/src/Wire/API/Routes/Public/Galley.hs
+++ b/libs/wire-api/src/Wire/API/Routes/Public/Galley.hs
@@ -48,8 +48,6 @@ import Wire.API.Team.Conversation
 import Wire.API.Team.Feature
 import Wire.API.Team.Permission (Perm (..))
 
--- import Wire.API.Team.Permission (Perm (..))
-
 instance AsHeaders '[ConvId] Conversation Conversation where
   toHeaders c = (I (qUnqualified (cnvQualifiedId c)) :* Nil, c)
   fromHeaders = snd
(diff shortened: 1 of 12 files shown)


libs/wire-api/src/Wire/API/Routes/Public/Galley.hs
```diff
@@ -731,59 +729,64 @@ type FeatureAPI =
     :<|> FeatureConfigGet 'WithLockStatus 'TeamFeatureSelfDeletingMessages
     :<|> FeatureConfigGet 'WithLockStatus 'TeamFeatureGuestLinks
 
-type FeatureStatusGet featureName =
+type FeatureStatusGet f =
```
fisx, line 732: please, no one-letter names :(

fisx (commented):
I suspect this will conflict with https://github.com/wireapp/wire-server/pull/2059.  Not sure what to do about it, though.

libs/wire-api/src/Wire/API/Routes/Public/Galley.hs
pcapriotti: Why are you picking on this particular one? The codebase is full of one-letter names. Even for this very purpose (see e.g. `Galley.API.Teams.Features`). I'm actually not sure what the problem is with one-letter names when they stand for very abstract things in a very limited scope (e.g. a single expression, like in this example). Also, the name that was there before isn't even correctly describing what the variable stands for (it's not a "feature name").

mdimjasevic (approved):
Overall this looks good.

I'm just curious: what is the motivation for removing internal endpoints from Swagger? Should we maybe put it into a different root/folder, but still have Swagger for internal endpoints?

services/galley/src/Galley/API/Internal.hs
```diff
@@ -101,247 +97,124 @@ import Wire.API.Federation.API.Galley
...
-  { iStatusGet ::
-      routes
-        :- "i"
-        :> "status"
-        :> Get '[Servant.JSON] NoContent,
-    iStatusHead ::
-      routes
-        :- "i"
-        :> "status"
-        :> Verb 'HEAD 200 '[Servant.JSON] NoContent,
```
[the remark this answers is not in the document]
pcapriotti: `HEAD` endpoints are unnecessary. Servant implements `HEAD` automatically for each `GET` endpoint.

pcapriotti:
> Overall this looks good.
> 
> I'm just curious: what is the motivation for removing internal endpoints from Swagger? Should we maybe put it into a different root/folder, but still have Swagger for internal endpoints?

Because the current swagger is supposed to present the API as visible from the outside (e.g. the `Z-User` header is not documented directly). Also, internal endpoints can overlap, and are only meaningful for a given service. For example, it would be nice to see a bunch of `/i/status` endpoints with no indication of which service they belong to.

PR #2090, opened 2022-01-31
Stay consistent with #2079

jschaul:
Follow-up to #2079. Just for cleanup, old and new parameter works as #2079 has a fallback to 'replicaCount', but going forward replicationFactor is the better naming.

### DIFF
--- a/hack/helm_vars/wire-server/values.yaml.gotmpl
+++ b/hack/helm_vars/wire-server/values.yaml.gotmpl
@@ -18,7 +18,7 @@ cassandra-migrations:
   imagePullPolicy: {{ .Values.imagePullPolicy }}
   cassandra:
     host: cassandra-ephemeral
-    replicaCount: 1
+    replicationFactor: 1
 elasticsearch-index:
   imagePullPolicy: {{ .Values.imagePullPolicy }}
   elasticsearch:


hack/helm_vars/wire-server/values.yaml.gotmpl
```diff
@@ -18,7 +18,7 @@ cassandra-migrations:
   imagePullPolicy: {{ .Values.imagePullPolicy }}
   cassandra:
     host: cassandra-ephemeral
-    replicaCount: 1
+    replicationFactor: 1
```
fisx, line 21: aren't there any call sides that need to be changed in sync with this?

jschaul: The change in the previous PR is backwards compatible, so nothing *has* to be changed, but it's nice if it's changed so eventually the old stuff can be phased out.

There is https://github.com/wireapp/wire-server-deploy/pull/541

There are also references in cailleach, but those require a new version deployed to the respective environment first. I'll open separate PRs there.

akshaymankar (approved):

hack/helm_vars/wire-server/values.yaml.gotmpl
jschaul: Created https://wearezeta.atlassian.net/browse/SQPIT-884 to track other deployments. Good concern!

PR #2091, opened 2022-02-01
Brig: Servantify self API

akshaymankar:
https://wearezeta.atlassian.net/browse/SQCORE-1167

## Checklist


akshaymankar:
The `GET /self/name` endpoint was found to be unused. I will delete it in a separate PR.

libs/wire-api/src/Wire/API/User.hs
```diff
@@ -845,22 +850,33 @@ modelUserUpdate = Doc.defineModel "UserUpdate" $ do
...
+   ]
+
+instance AsUnion PutSelfResponses (Maybe UpdateProfileError) where
+  toUnion (Just ProfileNotFound) = Z (I mkErrorDescription)
+  toUnion (Just DisplayNameManagedByScim) = S (Z (I mkErrorDescription))
+  toUnion Nothing = S (S (Z (I ())))
+  fromUnion (Z (I _)) = Just ProfileNotFound
+  fromUnion (S (Z (I _))) = Just DisplayNameManagedByScim
+  fromUnion (S (S (Z (I _)))) = Nothing
+  fromUnion (S (S (S x))) = case x of
```
pcapriotti: This should be probably easy to automate using `GenericAsUnion`.

You could also consider creating a type like `UpdateProfileResult` with 3 constructors instead, and then you could directly use `deriving via` to get the `AsUnion` instance.

pcapriotti (approved):
Looks good!

As commented below, I think we could get rid of the boilerplate (which is quite a lot!) by changing the result types to include all responses (including the successful ones) and deriving `AsUnion` via `GenericAsUnion`.

If you prefer to keep the `Maybe ErrorType` pattern, you can also use `GenericAsUnion`, by deriving an `AsUnion` instance for the error type, and then having a couple lines of boilerplate to combine it into an instance for `Maybe` (using `eitherToUnion` and `eitherFromUnion`).

This is all a bit unfortunate, but managing error types is always a pain. Once we switch to using polysemy in brig, we might be able to improve the situation.

libs/wire-api/src/Wire/API/User.hs
akshaymankar: Thanks for the tip. I created `maybeToUnion` and `maybeFromUnion` (this one uses eitherFromUnion) to define these instances. Let me know if they look ok.

pcapriotti (approved):
Looks good. It's a bit unfortunate that GHC doesn't let you define an instance for a type family application, even when it fully reduces. I'm not sure why this restriction exists.

akshaymankar:
> Looks good. It's a bit unfortunate that GHC doesn't let you define an instance for a type family application, even when it fully reduces. I'm not sure why this restriction exists.

I found this when I was confused about the error: https://gitlab.haskell.org/ghc/ghc/-/issues/3485

pcapriotti:
> I found this when I was confused about the error: https://gitlab.haskell.org/ghc/ghc/-/issues/3485

In that example the type family is stuck, and I can understand that something like that would not be allowed. However, if the type family application computes to something that is accepted as as the type of an instance, why not accept the application itself?

PR #2099, opened 2022-02-03
SQSERVICES 1239 remove unused locale setting in brig's values.yaml

battermann:
https://wearezeta.atlassian.net/browse/SQSERVICES-1239

## Checklist


pcapriotti (approved):
I don't know the context for this, but I'm assuming it's just an option that is not used anymore?

battermann:
> I don't know the context for this, but I'm assuming it's just an option that is not used anymore?

Kind of, the option has a hard coded default. And since overwriting by helm template is not supported, it can better be removed to avoid confusion.

PR #2104, opened 2022-02-07
Make RespondEmpty responses show up without a schema in swagger

akshaymankar:
They appeared as empty arrays, which is the schema for the unit type.

Before this fix:
image

After this fix:
image

## Checklist

### DIFF
--- a/libs/wire-api/src/Wire/API/Routes/MultiVerb.hs
+++ b/libs/wire-api/src/Wire/API/Routes/MultiVerb.hs
@@ -196,7 +196,7 @@ simpleResponseSwagger = do
       & S.schema ?~ ref
 
 instance
-  (KnownStatus s, KnownSymbol desc, S.ToSchema a) =>
+  (KnownSymbol desc, S.ToSchema a) =>
   IsSwaggerResponse (Respond s desc a)
   where
   responseSwagger = simpleResponseSwagger @a @desc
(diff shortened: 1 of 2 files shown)


libs/wire-api/src/Wire/API/Routes/MultiVerb.hs
```diff
@@ -244,11 +244,21 @@ instance KnownStatus s => IsResponse cs (RespondAs '() s desc ()) where
...
 
 instance
-  (KnownStatus s, KnownSymbol desc, S.ToSchema a) =>
+  (KnownSymbol desc, S.ToSchema a) =>
   IsSwaggerResponse (RespondAs ct s desc a)
   where
   responseSwagger = simpleResponseSwagger @a @desc
 
+instance
+  {-# OVERLAPPING #-}
```
pcapriotti, line 253: If we specify the kind of `ct` in  the other instance (with `ct :: *`), then the overlapping pragma should is not necessary.

libs/wire-api/src/Wire/API/Routes/MultiVerb.hs
```diff
@@ -244,11 +244,21 @@ instance KnownStatus s => IsResponse cs (RespondAs '() s desc ()) where
...
+  (KnownSymbol desc) =>
+  IsSwaggerResponse (RespondAs '() s desc ())
```
pcapriotti, line 255:
```suggestion
  IsSwaggerResponse (RespondEmpty s desc)
```

You can use the type synonym here.

pcapriotti (approved):
Looks good! Minor comments below.

PR #2116, opened 2022-02-09
Minimal client API versioning

pcapriotti:
This PR implements the bare minimum interface for [client API version negotiation](https://wearezeta.atlassian.net/wiki/spaces/ENGINEERIN/pages/555090587/Use+case+deciding+which+backend+API+version+to+use).

The changes in the nginz chart are based on https://github.com/zinfra/cailleach/pull/879 and result in the following [diff](https://gist.github.com/pcapriotti/1e0680ad99f5c128234ba0b876176542).

Tracked by https://wearezeta.atlassian.net/browse/FS-440.

## Checklist

### DIFF
--- a/libs/wire-api/src/Wire/API/VersionInfo.hs
+++ b/libs/wire-api/src/Wire/API/VersionInfo.hs
@@ -1,0 +1,27 @@
+-- This file is part of the Wire Server implementation.
+--
+-- Copyright (C) 2022 Wire Swiss GmbH <opensource@wire.com>
+--
+-- This program is free software: you can redistribute it and/or modify it under
+-- the terms of the GNU Affero General Public License as published by the Free
+-- Software Foundation, either version 3 of the License, or (at your option) any
+-- later version.
+--
+-- This program is distributed in the hope that it will be useful, but WITHOUT
+-- ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+-- FOR A PARTICULAR PURPOSE. See the GNU Affero General Public License for more
+-- details.
+--
+-- You should have received a copy of the GNU Affero General Public License along
+-- with this program. If not, see <https://www.gnu.org/licenses/>.
+
+module Wire.API.VersionInfo
+  ( vinfoSchema,
+  )
+where
+
+import Data.Schema
+import Imports
+
+vinfoSchema :: ValueSchema NamedSwaggerDoc v -> ValueSchema NamedSwaggerDoc [v]
+vinfoSchema sch = object "VersionInfo" $ field "supported" (array sch)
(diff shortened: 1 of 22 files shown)


fisx (commented):
it just occurred to me that the nginz changes are a lot or work but pretty straight-forward.  just add a `(/v(\d+))?` in front of every path and make them all regexps.  but everybody except me probably already thought of that.  :)  https://xkcd.com/564/

pcapriotti:
> You should also have a flag in the server config(s) that keeps supported versions from being advertised, so we can allow on-prem backends to upgrade without forcing them to make their network infrastructure federation-ready:

I don't follow. Version 1 can be supported without having to enable federation or make any changes to the infrastructure. This change has not much to do with federation, really.

mdimjasevic (approved):

services/brig/src/Brig/API/Public.hs
```diff
@@ -108,6 +108,7 @@ import qualified Wire.API.Routes.Public.Galley as GalleyAPI
 import qualified Wire.API.Routes.Public.LegalHold as LegalHoldAPI
 import qualified Wire.API.Routes.Public.Spar as SparAPI
 import qualified Wire.API.Routes.Public.Util as Public
+import Wire.API.Routes.Version
```
fisx, line 111: qualify this, and call `versionSwagger` `swaggerDoc`?

libs/wire-api/src/Wire/API/VersionInfo.hs
```diff
@@ -0,0 +1,27 @@
...
+
+module Wire.API.VersionInfo
+  ( vinfoSchema,
+  )
+where
+
+import Data.Schema
+import Imports
+
+vinfoSchema :: ValueSchema NamedSwaggerDoc v -> ValueSchema NamedSwaggerDoc [v]
```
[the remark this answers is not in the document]
pcapriotti: `VersionInfo` is more general, and can be used for other API version lists, whereas `Wire.API.Routes.Version` is specific for the client API.

services/brig/src/Brig/API/Public.hs
pcapriotti: I don't like to create name clashes on purpose and resolve them with qualified imports. It makes everything more awkward and I don't see the benefit.

libs/wire-api/src/Wire/API/VersionInfo.hs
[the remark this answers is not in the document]
fisx: What about `Version`, then?  Will there be a type `Version` for the federation server-to-server API as well?

services/brig/src/Brig/API/Public.hs
fisx: I won't insist, but I think we agreed at some point (we should document these things) that imports should either be explicit or qualified, but not neither.  Doesn't have to happen now, since we're not consistent, but can you get on board with the general idea?

pcapriotti: Well, I disagree with the general idea, but if it's something everyone else agrees on, I'll comply.

libs/wire-api/src/Wire/API/VersionInfo.hs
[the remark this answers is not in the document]
pcapriotti: Yes, that was the idea, as well as for other APIs that need to be versioned separately.

PR #2119, opened 2022-02-10
Build ubuntu20 docker images with cabal instead of stack

smatting:
This PR changes the way we build Docker images: `stack` is replaced by `cabal` to build all binaries.

This affects
- Docker images for PR pipelines (`Dockerfile.fast-intermediate`)
- Docker images for on-prem and other k8s-based deployments (`Dockerfile.intermediate`)

## Checklist

### DIFF
--- a/build/ubuntu/Dockerfile.fast-intermediate
+++ b/build/ubuntu/Dockerfile.fast-intermediate
@@ -1,5 +1,5 @@
 ARG builder=quay.io/wire/ubuntu20-builder
-ARG deps=quay.io/wire/ubuntu-deps
+ARG deps=quay.io/wire/ubuntu20-deps:develop
 
 #--- Builder stage ---
 FROM ${builder} as builder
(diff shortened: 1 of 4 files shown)


smatting:
[Build](https://bridge-ie-concourse.zinfra.io/teams/main/pipelines/wire-server-pr/jobs/wire-server-compile-ubuntu20/builds/301) with cabal succeeded. Build time (20m) is on-par with build times with stack.

build/ubuntu/Dockerfile.fast-intermediate
```diff
@@ -1,5 +1,5 @@
 ARG builder=quay.io/wire/ubuntu20-builder
-ARG deps=quay.io/wire/ubuntu-deps
+ARG deps=quay.io/wire/ubuntu20-deps:develop
```
smatting, line 2: @akshaymankar is this correct?
In contrast `Dockerfile.intermediate` has `deps=quay.io/wire/ubuntu20-deps`

akshaymankar: I think both of them should have the `develop` tag, IIRC the `latest` tag is not published.

akshaymankar (approved):
Are you also planning on making cabal the default build tool when using `make`?

build/ubuntu/Dockerfile.fast-intermediate
smatting: Ok, changed tag to `develop`

smatting:
> Are you also planning on making cabal the default build tool when using `make`?

Yes, I would also like to get rid of most Makefiles, if possible. I think this can be explored in the next step where I'm planning to change pipelines for the debian packages

PR #2121, opened 2022-02-10
Brig: Servantify `POST /register` and `POST /i/users` endpoint

akshaymankar:
https://wearezeta.atlassian.net/browse/SQCORE-1168

Noteworthy things:

- I servantified `POST /i/users` because it had pretty much the same API and was using the same types.
- While registering a user into a team, if the team has legal hold, the user can only be added if the fanoutLimit has not been reached. This check happens in galley, which generates a `Wai.Error`, brig just throws this error and so isn't aware of internals. This kind of error cannot be transformed into and `ErrorDescription`. So, for now (until galley's internal API is servantiified), brig throws this error using `throwM`, this error gets caught and handled by the `Handler` monad when it executes requests. This is of course not ideal. But hopefully we get to servantify Galley's internal endpoints and we can get rid of this hack.
- The parser for `NewUser` is fairly complex some fields depend on other fields. To help with this, I created `NewUserRaw` which just parses individual fields. The schema for `NewUser` type uses `withParser` to do the validation and make composite fields.
-  Remove ToJSON/FromJSON for UserIdentity, use Schema for User. The ToJSON for UserIdentity encoded nulls, but every other object which used it didn't encode nulls. So, it was better to remove it.

## Checklist


services/brig/src/Brig/API/Handler.hs
```diff
@@ -140,10 +152,16 @@ parseJsonBody req = parseBody req !>> StdError . badRequest
 
 -- | If a whitelist is configured, consult it, otherwise a no-op. {#RefActivationWhitelist}
 checkWhitelist :: Either Email Phone -> Handler ()
-checkWhitelist key = do
+checkWhitelist = checkWhitelistWithError (StdError whitelistError)
+
+checkWhitelistWithError :: (Monad m, MonadReader Env m, MonadIO m, Catch.MonadMask m, MonadHttp m, MonadError e m) => e -> Either Email Phone -> m ()
+checkWhitelistWithError e key = do
+  ok <- isWhiteListed key
+  unless ok (throwError e)
```
fisx, line 160:
```suggestion
  unlessM (isWhiteListed key) (throwError e)
```

https://hoogle.zinfra.io/file/wire-server/.stack-work/install/x86_64-linux-tinfo6/4477baab320e495f60baecbf311b8b6399a53d09f9ec1ecba83646a2c65c551e/8.10.7/doc/imports-0.1.0/Imports.html#v:unlessM

fisx (commented):
Looks good, but I haven't gone enough into the details to confidently approve yet.  Have you taken a look at the swagger output?

libs/wire-api/src/Wire/API/User.hs
```diff
@@ -802,6 +835,24 @@ data NewTeamUser
   deriving stock (Eq, Show, Generic)
   deriving (Arbitrary) via (GenericUniform NewTeamUser)
 
+newTeamUserCode :: NewTeamUser -> Maybe InvitationCode
```
pcapriotti, line 838: These can be autogenerated by `makePrisms`. For example, instead of `newTeamUserCode` you can use `preview _NewTeamMember`.

libs/wire-api/src/Wire/API/User/Activation.hs
```diff
@@ -181,7 +184,7 @@ instance ToJSON ActivationResponse where
 instance FromJSON ActivationResponse where
```
pcapriotti, line 184: Why doesn't this have a `ToSchema` instance?

libs/wire-api/src/Wire/API/Util/Aeson.hs
```diff
@@ -43,3 +45,7 @@ instance (Generic a, GToJSON Zero (Rep a)) => ToJSON (CustomEncoded a) where
 
 instance (Generic a, GFromJSON Zero (Rep a)) => FromJSON (CustomEncoded a) where
   parseJSON = fmap CustomEncoded . genericParseJSON @a customEncodingOptions
+
+eitherToParser :: Either String a -> Parser a
+eitherToParser (Left e) = fail e
+eitherToParser (Right a) = pure a
```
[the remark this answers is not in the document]
pcapriotti: I agree, we can just use `either fail pure` directly. Short and clear.

pcapriotti (approved):
Looks good.

One thing I'm not sure of is the huge list of error responses for the register endpoint. Even with polysemy, that corresponds to a huge list of individual error effects, which is a pain to manage. It's also not so great documentation, IMHO, since it shows way too much detail.

I think it would be nicer to have groupings of errors, like "registration error", "authentication error" etc... Then we could document each group separately (this could even be somewhat automated). We have done this sort of thing with federation errors, where it was pretty obvious that showing 20 possible error responses for basically each endpoint would have been insane.

fisx:
> I think it would be nicer to have groupings of errors, like "registration error", "authentication error" etc... Then we could document each group separately (this could even be somewhat automated).

How do you guarantee that all errors in a group have the same code and label?  Because that is the information we want to have for swagger code generation, no?

fisx:
> where it was pretty obvious that showing 20 possible error responses for basically each endpoint would have been insane.

Not obvious to me.  Is it a run-time or compile-time performance issue?  Or are you just unhappy with the length of the type signatures in characters?

PR #2123, opened 2022-02-10
Extend test suite slightly

fisx:
Trying to reproduce https://wearezeta.atlassian.net/browse/SQSERVICES-1279, but failing: everything seems to be in order.  I would like to add these tests anyway.
 
## Checklist

### DIFF
--- a/changelog.d/5-internal/i-tests
+++ b/changelog.d/5-internal/i-tests
@@ -1,0 +1,1 @@
+Add integration tests for scim/saml user creation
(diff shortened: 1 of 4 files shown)


smatting (approved):

smatting (changes requested):
That test seems to be failing on CI
```
  3799         user is created once, then deleted in team settings, then can login again.
   3800 1:D,6:logger,1:=,14:cassandra.spar,74:Connection established: datacenter1:rack1:10.233.113.168:9042#<soc>
   3801           responds with 'allowed' FAILED [2]
   ```

smatting (approved):

PR #2125, opened 2022-02-10
Refactor Proteus Conversation Creation Handlers and Drop Managed Conversations

mdimjasevic:
This PR does a couple of refactorings:

- It unifies endpoint handlers for Proteus conversation creation for a regular group conversation and for a team conversation into a single function,
- Managed conversations are dropped altogether as they are not used at all. As a consequence, an internal endpoint for creating a managed conversation no longer exists.

*Open question*: can we drop the *managed* column from the `team_conv` table?

## Checklist


jschaul (commented):
> Open question: can we drop the managed column from the team_conv table?

TL;DR: Yes, but not in this PR, but in a separate PR only opened once this PR 2125 has made its way to master and deployed to production (and possibly deployed to customers)

Context: Doing so (i.e. introducing a schema migration that drops this column inside this PR) would be a backwards incompatible change, which would lead to the problems outlined in https://github.com/wireapp/wire-server/blob/develop/docs/developer/cassandra-interaction.md#backwards-incompatible-schema-changes

What this PR *should* be doing which as far as I can see it does not yet do, is stop writing and reading the `managed` attribute from cassandra. i.e., change the cql statements [here](https://github.com/wireapp/wire-server/blob/develop/services/galley/src/Galley/Cassandra/Queries.hs#L54-L61) and [here](https://github.com/wireapp/wire-server/blob/develop/services/galley/src/Galley/Cassandra/Queries.hs#L141-L142) to stop taking the `managed` attribute. Doing this will allow a future PR (in a few weeks or month's time) to introduce an `ALTER TABLE team_conv DROP managed`. You can leave a FUTUREWORK somewhere linking back to this PR and/or comment and/or the cassandra doc.

If you're not sure if clients might parse the `managed` attribute at the API level, you can leave it in at the API level (always False).

Thanks for this! +166 -609 is great! 🚀

pcapriotti (approved):

services/galley/src/Galley/API/Update.hs
```diff
@@ -1465,8 +1465,7 @@ addBot lusr zcon b = do
     teamConvChecks :: ConvId -> TeamId -> Sem r ()
     teamConvChecks cid tid = do
       tcv <- E.getTeamConversation tid cid
-      when (maybe True (view managedConversation) tcv) $
-        throw NoAddToManaged
+      when (isNothing tcv) $ throw NoAddToManaged
```
mdimjasevic, line 1468: @pcapriotti , does this check make sense at all? We're throwing out managed conversations with this PR, yet here we say that a user or bot cannot be added to a managed conversation directly. Any objection to gettting rid of this check as well?

services/galley/src/Galley/Cassandra/Queries.hs
```diff
@@ -138,8 +138,8 @@ selectUserTeamsFrom = "select team from user_team where user = ? and team > ? or
 insertTeam :: PrepQuery W (TeamId, UserId, Text, Text, Maybe Text, TeamStatus, TeamBinding) ()
 insertTeam = "insert into team (team, creator, name, icon, icon_key, deleted, status, binding) values (?, ?, ?, ?, ?, false, ?, ?)"
 
-insertTeamConv :: PrepQuery W (TeamId, ConvId, Bool) ()
-insertTeamConv = "insert into team_conv (team, conv, managed) values (?, ?, ?)"
+insertTeamConv :: PrepQuery W (TeamId, ConvId) ()
+insertTeamConv = "insert into team_conv (team, conv) values (?, ?)"
```
mdimjasevic, line 142: @pcapriotti , this looks suspicious to me: can we even insert without giving a value for each column? I guess a `false` should be inserted where the `managed` input placeholder used to be.

pcapriotti: We want to remove all uses of the `managed` column, so we can delete the column later.

mdimjasevic: Akshay explained this in a chat (this is an OK Cassandra insertion statement).

services/galley/src/Galley/API/Update.hs
pcapriotti: Yes, this throw is unreachable, since at this point we already know that it is a team conversation (look at the call site).
Please remove this whole function and the `NoAddToManaged` constructor.

PR #2126, opened 2022-02-14
Cleanup after review

supersven:
According to the boy scout rule, we cleaned up some tests a bit (to increase their maintainability, readability, ...).

## Checklist

### DIFF
--- a/services/galley/test/integration/API/Teams/LegalHold.hs
+++ b/services/galley/test/integration/API/Teams/LegalHold.hs
@@ -713,6 +713,9 @@ testInWhitelist = do
           UserLegalHoldPending
           userStatus
     do
+      -- owner cannot approve legalhold device
+      withLHWhitelist tid (approveLegalHoldDevice' g (Just defPassword) owner member tid) !!! testResponse 403 (Just "access-denied")
+    do
       -- approve works
       withLHWhitelist tid (approveLegalHoldDevice' g (Just defPassword) member member tid) !!! testResponse 200 Nothing
       UserLegalHoldStatusResponse userStatus lastPrekey' clientId' <- withLHWhitelist tid (getUserStatusTyped' g member tid)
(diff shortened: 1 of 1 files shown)


fisx (approved):
Thanks!  (Unrelated to the merge-ability of this PR: did you solve the mystery of the LH device that could still be reached after its destruction?)

supersven:
Hey @fisx ,

> Thanks! (Unrelated to the merge-ability of this PR: did you solve the mystery of the LH device that could still be reached after its destruction?)

I've solved the mystery:
- The `ClientId` of the legalhold device is just noise in this call. (Spoiler first :wink:)
- It can be replaced by anything, e.g. `ClientId "42"`.
- What happens is that `Galley.API.Message.postQualifiedOtrMessage` pulls a fresh set of clientIds for all members (identified by their `UserId`) from the database.
- Then some logic constructs an empty list of messages to be sent and loops over it for sending. Because executing nothing cannot fail, the result is a success response.
- However, the `QualifiedNewOtrMessage` contains `qualifiedNewOtrClientMismatchStrategy = MismatchReportAll`, so I'm wondering if this isn't a bug... :thinking:

supersven:
For reference:
- The strange call is made here: https://github.com/wireapp/wire-server/blob/b4b8bc519b01b22af58d3301f28b639239c3f407/services/galley/test/integration/API/Teams/LegalHold.hs#L931
- Lookup of clients (depending on conversation members) happens here: https://github.com/wireapp/wire-server/blob/0a233d4dd93421ad0b866254351f6d7a15aa2530/services/galley/src/Galley/API/Message.hs#L406

fisx:
> * However, the `QualifiedNewOtrMessage` contains `qualifiedNewOtrClientMismatchStrategy = MismatchReportAll`, so I'm wondering if this isn't a bug... thinking

my money would be on "yes, but in the test suite, not in the prod code".  but it's probably worth being certain?

supersven:
@fisx , you can keep you money :wink: 

As shown in the adjusted test (https://github.com/wireapp/wire-server/pull/2131/files#diff-c7405593bda0c55eab527c9e369a5bd3d6f09874b5e213652c6a77d0d9742262R939), the device is reported.

Sorry, I was a bit unsure was "reporting" means in this case. It showed that it means: *Respond with success, but add the superfluous clientId to the body.* IMHO that should be fine. Out main concern - that messages could be delivered to deleted legalhold devices - proved wrong.

PR #2127, opened 2022-02-14
Drop the `managed` Column from `team_conv` Table in Galley

mdimjasevic:
(Do not merge before Sept 2022 and/or before most deployments are on chart version >= 4.1.0. In case this is merged and deployed before 4.1.0 was deployed first, a transient period of the time it takes to roll out a deployment (in the order of minutes usually) of 'HTTP status 500 internal server errors' on client API calls to conversations can happen.)

This PR is a follow-up to PR #2125 that was merged on Feb 14, 2022.

The PR also removes the `managed` key from JSON instances in client API version 2. This makes it a client **API breaking change** so versioning is used; version 1 is kept intact.

## Checklist

 - [x] Rebase on top of the latest `develop`. Make sure to update `schemaVersion` in Galley and rename `services/galley/schema/src/V60_DropManagedConversations.hs` accordingly.


pcapriotti:
I'm not sure how the version bounds meant for servant routes ended up in API types, but I would recommend not doing that. As discussed, I don't think we need separate types for older versions, because this feature never actually worked and was never used. We only need to support the interface in V1, not the functionality. So I would only have the types for v2, and just make the JSON instances backward compatible. No messing around with versions should be necessary.

mdimjasevic:
> I'm not sure how the version bounds meant for servant routes ended up in API types, but I would recommend not doing that. As discussed, I don't think we need separate types for older versions, because this feature never actually worked and was never used. We only need to support the interface in V1, not the functionality. So I would only have the types for v2, and just make the JSON instances backward compatible. No messing around with versions should be necessary.

I've reverted the changes to the `ConvTeamInfo` and `NewConv` types to incorporate your feedback: they are never used in output/responses, so it doesn't hurt to have `managed` in their `ToJSON` instances.

However, `TeamConversation` and `TeamConversationList` do end up in output/endpoint responses, so for both of them I've introduced a version boundary type index so we can provide different `ToJSON` instances depending on the version used.

Can you take a look again?

libs/wire-api/src/Wire/API/Team/Conversation.hs
```diff
@@ -36,76 +36,143 @@ module Wire.API.Team.Conversation
...
+import Wire.API.Routes.Version
 import Wire.Arbitrary (Arbitrary, GenericUniform (..))
 
 --------------------------------------------------------------------------------
 -- TeamConversation
 
-newtype TeamConversation = TeamConversation
+-- FUTUREWORK: The version tag type argument can be dropped as soon as versions
+-- below V2 are not supported any more.
+newtype TeamConversation (v :: *) = TeamConversation
```
mdimjasevic, line 54:
```suggestion
-- Derivation of various JSON instances is verbose with this version tag type variable, but this can be simplified once the old version can be dropped.
newtype TeamConversation (v :: *) = TeamConversation
```

services/brig/src/Brig/IO/Intra.hs
```diff
@@ -953,8 +954,15 @@ getTeamConv usr tid cnv = do
     200 -> Just <$> decodeBody "galley" rs
     _ -> pure Nothing
   where
+    v2 = toByteString' (toLower <$> show V2)
```
mdimjasevic, line 957:
```suggestion
    -- Ensure we get a type error, and not a runtime error once we drop V2.
    v2 = toByteString' (toLower <$> show V2)
```

pcapriotti:
> However, `TeamConversation` and `TeamConversationList` do end up in output/endpoint responses, so for both of them I've introduced a version boundary type index so we can provide different `ToJSON` instances depending on the version used.

I still don't think this is necessary. Just output a `managed` field with a value of `true` regardless of the version, and make it deprecated. Ideally, hide it from swagger if that's not too hard to achieve, so we can drop it when V1 becomes unsupported. Also, combinators like `From V2` are not intended to be used as type arguments for data types like that. If you ever actually need a type that depends on a version, just use the version itself or simply make a different type with a V2 suffix or similar.

pcapriotti (approved):
Looks good to me.

libs/wire-api/src/Wire/API/Conversation.hs
```diff
@@ -746,7 +746,7 @@ modelTeamInfo = Doc.defineModel "TeamInfo" $ do
   Doc.property "teamid" Doc.bytes' $
     Doc.description "Team ID"
   Doc.property "managed" Doc.bool' $
-    Doc.description "Is this a managed team conversation?"
+    Doc.description "(Not parsed any more) Is this a managed team conversation?"
```
elland, line 749: I thought we were changing this description as well?

libs/wire-api/src/Wire/API/Team/Conversation.hs
```diff
@@ -53,17 +53,25 @@ newtype TeamConversation = TeamConversation
...
+  schema =
+    objectWithDocModifier
+      "TeamConversation"
+      (description ?~ "Team conversation data")
+      $ TeamConversation
+        <$> _conversationId .= field "conversation" schema
+        <* const ()
+          .= fieldWithDocModifier
+            "managed"
+            ( description ?~ "This field MUST not be used by clients. "
```
elland, line 68: Capitalise NOT as well? :)

libs/wire-api/src/Wire/API/Conversation.hs
mdimjasevic: > I thought we were changing this description as well?

Sure, let me update this one as well. I'm not sure the model is used anywhere, but I'll update it anyway.

libs/wire-api/src/Wire/API/Team/Conversation.hs
mdimjasevic: I'll do that.

PR #2129, opened 2022-02-14
Make helm-template script use helmfile

pcapriotti:
Adapt the `helm-template.sh` script to use `helmfile` instead of `helm` directly.  After #1805, the values files don't exist anymore, so this PR uses `helmfile` for rendering all helm charts using the actual templated values file.

## Checklist

### DIFF
--- a/hack/bin/helm-template.sh
+++ b/hack/bin/helm-template.sh
@@ -13,9 +13,11 @@ set -e
 DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 TOP_LEVEL="$DIR/../.."
 CHARTS_DIR="${TOP_LEVEL}/.local/charts"
-: "${FEDERATION_DOMAIN:=example.com}"
+: "${FEDERATION_DOMAIN_BASE:=example.com}"
+: "${FEDERATION_DOMAIN:=namespace1.example.com}"
 : "${NAMESPACE:=namespace1}"
 
+export FEDERATION_DOMAIN_BASE
 export FEDERATION_DOMAIN
 export NAMESPACE
 
--- a/hack/bin/helm-template.sh
+++ b/hack/bin/helm-template.sh
@@ -4,25 +4,20 @@
 # hack/helm_vars as overrrides, if available.  This allows debugging helm
 # templating issues without actually installing anything, and without needing
 # access to a kubernetes cluster
-USAGE="Usage: $0"
 
 set -e
-
-chart=${1:?$USAGE}
 
 DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 TOP_LEVEL="$DIR/../.."
 CHARTS_DIR="${TOP_LEVEL}/.local/charts"
+: "${FEDERATION_DOMAIN:=example.com}"
+: "${NAMESPACE:=namespace1}"
 
-valuesfile="${DIR}/../helm_vars/${chart}/values.yaml"
-certificatesfile="${DIR}/../helm_vars/${chart}/certificates.yaml"
-declare -a options=()
-if [ -f "$valuesfile" ]; then
-    options+=(-f "$valuesfile")
-fi
-if [ -f "$certificatesfile" ]; then
-    options+=(-f "$certificatesfile")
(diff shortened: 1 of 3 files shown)


hack/bin/helm-template.sh
```diff
@@ -4,25 +4,20 @@
...
 
 set -e
 
-chart=${1:?$USAGE}
-
 DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 TOP_LEVEL="$DIR/../.."
 CHARTS_DIR="${TOP_LEVEL}/.local/charts"
+: "${FEDERATION_DOMAIN:=example.com}"
+: "${NAMESPACE:=namespace1}"
```
jschaul, line 14: I'm not familiar with this syntax. Is that the same as

```
NAMESPACE=${NAMESPACE:-"namespace1"}
```
? i.e. set-to-default-if-not-exists? Or this hardcodes/overrides this variable to always be that?

hack/bin/helm-template.sh
```diff
@@ -4,25 +4,20 @@
 # hack/helm_vars as overrrides, if available.  This allows debugging helm
 # templating issues without actually installing anything, and without needing
 # access to a kubernetes cluster
-USAGE="Usage: $0"
 
 set -e
 
-chart=${1:?$USAGE}
```
jschaul, line 11: what happened to this variable?

jschaul (changes requested):
Doesn't work.

```
make helm-template-wire-server
```

```
(... snip for brevity...)
/home/user/Documents/git/wire-server/hack/bin/selfsigned-kubernetes.sh: line 29: FEDERATION_DOMAIN_BASE: you must provide a FEDERATION_DOMAIN_BASE env variable
make: *** [Makefile:527: helm-template-wire-server] Error 1
```

hack/bin/helm-template.sh
pcapriotti: Yes, it's the same as that. `${x:=foo}` sets `$x` to `foo` if not set, and returns the value of `$x`. The colon is just an empty statement. It's a way to evaluate an expression (usually with side effects), without executing any command.

hack/bin/helm-template.sh
pcapriotti: I don't know how to render a single chart with helmfile.

pcapriotti:
Sorry, I wasn't aware of the `Makefile` target. I guess I'll remove the argument, since the current script doesn't take any, and just renders all charts.

> I think it's useful to be able to template a single chart, rather than "everything". How is this script intended to be used? Some docs would be nice for usage and purpose.

I've been using it by piping the output of the script to `jq` (via some yaml -> json conversion script). Or one could use `yq` directly. This works well enough for me. Not sure what docs are needed, the script just works by itself without any arguments.

hack/bin/helm-template.sh
jschaul: Ok, thanks for the explanations!

jschaul:
@pcapriotti  This PR still does not work (as written in my previous comment) after the latest changes, with or without the makefile target:

```
./hack/bin/helm-template.sh
/home/user/Documents/git/wire-server/hack/bin/selfsigned-kubernetes.sh: line 29: FEDERATION_DOMAIN_BASE: you must provide a FEDERATION_DOMAIN_BASE env variable
```

So I cannot approve this yet, please set the necessary environment variable in the script matching the domain.

pcapriotti:
> @pcapriotti This PR still does not work (as written in my previous comment) after the latest changes, with or without the makefile target:
> 
> ```
> ./hack/bin/helm-template.sh
> /home/user/Documents/git/wire-server/hack/bin/selfsigned-kubernetes.sh: line 29: FEDERATION_DOMAIN_BASE: you must provide a FEDERATION_DOMAIN_BASE env variable
> ```
> 
> So I cannot approve this yet, please set the necessary environment variable in the script matching the domain.

You're right, I hadn't tested this correctly. Should be fixed now.

pcapriotti:
I'm not sure how to make this usable for everyone, so I'll just move into my own collection of scripts and close the PR. Feel free to reopen if you want to brig this back.

PR #2135, opened 2022-02-16
Sqservices 1157 2 mf for generating SCIM token errors and renaming

battermann:


libs/wire-api/src/Wire/API/Routes/Public/Spar.hs
```diff
@@ -203,7 +204,10 @@ sparResponseURI (Just tid) =
 
 type APIScim =
   OmitDocs :> "v2" :> ScimSiteAPI SparTag
-    :<|> "auth-tokens" :> APIScimToken
+    :<|> "auth-tokens"
+      :> CanThrow PasswordAuthenticationFailed
```
fisx, line 208: that was missing before, right?  interesting.  looking forward to polysemizing `CanThrow` away...  :)

fisx: [for another PR] that `OmitDocs` above btw may make the `CanThrow` pretty pointless, since it makes sure this doesn't make it into the swagger docs.  we should probably remove `OmitDocs` entirely and have all the scim docs online as well.  the reason we didn't was that there is a standard for it available where the information exists, and we didn't want to write all the schema code.

fisx (approved):

fisx:
erm, but why don't the golden tests fail here?

PR #2136, opened 2022-02-16
correct order of modules

battermann:


fisx (approved):

fisx:
yes, we should definitely get rid of hpack, too!  :)

PR #2140, opened 2022-02-16
[Polysemy] Introduce the row type variable in Brig

mdimjasevic:
This is the first PR in a series of PRs that introduce Polysemy to Brig. It introduces the row type variable to several monads in the service.

## Checklist


services/brig/src/Brig/Data/Client.hs
```diff
@@ -399,8 +399,7 @@ withOptLock u c ma = go (10 :: Int)
         execDyn' e m conv cmd = recovering policy handlers (const run)
           where
             run = execCatch e cmd >>= either handleErr (return . conv)
-            handlers = httpHandlers --TODO: Figure out why it needs a constraint Typeable effs
-            -- handlers = httpHandlers ++ [const $ EL.handler_ AWS._ConditionalCheckFailedException (pure True)]
+            handlers = httpHandlers ++ [const $ Handler $ pure . const True . preview (AWS._ConditionalCheckFailedException @SomeException)]
```
pcapriotti, line 402: This doesn't look right. The prism doesn't actually get used, so this will catch any exception, rather than the one identified by the prism.

I'm not sure why `handler_` requires a typeable constraint (probably a limitation of `Control.Monad.Catch`), but that means this code should run in `IO` (or some simple monad stack over `IO`), and later be lifted to the `AppIO`/polysemy monad.

Or, for the time being, you can add a `Typeable` instance to `AppIO`.

pcapriotti (changes requested):

services/brig/src/Brig/Data/Client.hs
mdimjasevic: Thanks for spotting this. I'll take a look at how to fix this.

We didn't want to introduce the `Typeable` constraint on `AppIO` as that would propagate the constraint further up the call stack, and this doesn't look like the right thing to do.

fisx: Except for the noise, adding `Typeable` constraints is trivial and easy, and the reason it needs to be added explicitly is rather technical.  I wouldn't spend any effort trying to avoid it.

mdimjasevic: (Sorry, I keep on forgetting not to rebase once a PR is under review.)

I took Paolo's suggestion and turned the helper function to run in `IO` instead of running in a constrained monad `m`. See https://github.com/wireapp/wire-server/pull/2140/commits/1abfd3c9c8c713f7d264dc37f64c106e0d43d0f0 for details.

@fisx , @pcapriotti , is that OK now?

pcapriotti (approved):
Looks good!

PR #2147, opened 2022-02-21
Add MLS public keys to clients

pcapriotti:
This PR introduces changes to the client API needed to support validation of MLS key packages. See https://wearezeta.atlassian.net/wiki/spaces/CORE/pages/553418860/MLS+implementation+in+wire-server#Changes-to-the-client-API for details.

The client JSON object now has an additional field `mls_public_keys`, containing an object mapping signature schemes to public keys, e.g.
```
{
  ...
  "mls_public_keys": { "ed25519": "GY+t1EQu0Zsm0r/zrm6zz9UpjPcAPyT5i8L1iaY3ypM=" }
  ...
}
```
At the moment, `ed25519` is the only supported signature scheme, corresponding to MLS ciphersuite 1.

When creating a new client with `POST /clients`, the field `mls_public_keys` can be set, and the corresponding public keys are bound to the device identity on the backend, and will be used to veriy uploaded key packages with a matching signature scheme.

When updating a client with `PUT /clients/:client`, the field `mls_public_keys` can also be set, with a similar effect. If a given signature scheme already has a public key set for that device, the request will fail.

Tracked by https://wearezeta.atlassian.net/browse/FS-472.

## Checklist


services/brig/src/Brig/Data/Client.hs
```diff
@@ -247,6 +266,31 @@ claimPrekey u c =
       ind <- liftIO $ randomRIO (0, length pks' - 1)
       return $ atMay pks' ind
 
+lookupMLSPublicKey :: UserId -> ClientId -> SignatureSchemeTag -> AppIO r (Maybe LByteString)
```
mdimjasevic, line 269: Where possible, I'd prefer these data functions to run in a `MonadClient m` rather than `AppIO r` (with a `MonadReader` constraint where needed).

services/brig/src/Brig/Data/MLS/KeyPackage.hs
```diff
@@ -43,30 +46,45 @@ insertKeyPackages uid cid kps = retry x5 . batch $ do
     q :: PrepQuery W (UserId, ClientId, KeyPackageData, KeyPackageRef) ()
     q = "INSERT INTO mls_key_packages (user, client, data, ref) VALUES (?, ?, ?, ?)"
 
-claimKeyPackage :: UserId -> ClientId -> MaybeT (AppIO r) KeyPackageData
-claimKeyPackage u c = MaybeT $ do
+claimKeyPackage :: Local UserId -> ClientId -> MaybeT (AppIO r) (KeyPackageRef, KeyPackageData)
```
mdimjasevic, line 49: If possible, it'd be great if you could replace `AppIO r` with `m` for `MonadClient m, MonadReader Env m`.

pcapriotti:
I'm not sure why you prefer the more general type signatures. It makes it tricker to use CPS functions like `withMVar`, and since we're polysemising these things anyway, I would imagine having the concrete types in there makes it easier to turn this into an interpretation function without having to use any higher-order stuff.

mdimjasevic:
> I'm not sure why you prefer the more general type signatures. It makes it tricker to use CPS functions like `withMVar`, and since we're polysemising these things anyway, I would imagine having the concrete types in there makes it easier to turn this into an interpretation function without having to use any higher-order stuff.

This was just a suggestion. Of course, if you have good reasons not to have more general type signatures, that is fine. I'm not requesting to change them.

mdimjasevic:
Will the CI status change with this test message?

PR #2150, opened 2022-02-22
Support Creating an MLS Conversation

mdimjasevic:
This is per https://wearezeta.atlassian.net/browse/FS-432.

The PR extends the existing endpoint `POST /conversations` by:

- adding an optional request body field `protocol`, which can be "proteus" or "mls",
- adding an optional response body field 'group_id', which when this is an MLS conversation contains an MLS group ID associated with this conversation.

Remarks for the reviewer:

- Commits are not nicely organised so I do not recommend reviewing on a per-commit basis.
- Do we need the group ID in the `conversation` table given that there is a new table mapping group IDs to qualified conversation IDs?
- I am not sure why the difference for the Cassandra schema `.cql` file is so big. The schema migration is given in the `V61_MLSConversation` module.
- The `onConversationCreated` handler in the federation API does not need any special handling of MLS conversations like checking a group ID. This is so because a new MLS conversation always starts out with a single local member (the conversation creator) and therefore no remote backends are contacted at this phase.

## Checklist


libs/wire-api/src/Wire/API/MLS/GroupId.hs
```diff
@@ -0,0 +1,96 @@
...
+
+instance ToSchema GroupId where
+  schema =
+    GroupId
+      <$> unGroupId
+      .= named "GroupId" (Base64ByteString .= fmap fromBase64ByteString base64Schema)
+
+-- | Parse a 'GroupId' byte sequence in the network order as a qualified
+-- conversation ID.
+parseConvId :: GroupId -> A.Parser (Qualified ConvId)
```
pcapriotti, line 60: Could we not use `binary` here too for consistency with the rest of the MLS code?

services/galley/src/Galley/Cassandra/Conversation.hs
```diff
@@ -52,17 +52,20 @@ import qualified UnliftIO
...
+        ProtocolProteus -> Nothing
+        ProtocolMLS -> Just . convIdToGroupId . qUntagged . qualifyAs loc $ conv
+  retry x5 . batch $ do
+    setType BatchLogged
+    setConsistency LocalQuorum
+    addPrepQuery
+      Cql.insertConv
+      (conv, ty, usr, Cql.Set (toList acc), Cql.Set (toList arole), fmap fromRange name, mtid, mtimer, recpt, protocol, groupId)
+    for_ mtid $ \tid -> addPrepQuery Cql.insertTeamConv (tid, conv)
+    for_ groupId $ \gid -> addPrepQuery Cql.insertGroupId (gid, conv, tDomain loc)
```
pcapriotti, line 68: What's the point of `decodeConvId` if we're storing the group ID mapping for local conversations?

mdimjasevic: I dropped `decodeConvId` compared to the last time you review this PR.

libs/types-common/src/Data/Domain.hs
```diff
@@ -76,10 +78,17 @@ domainText :: Domain -> Text
...
 
 mkDomain :: Text -> Either String Domain
-mkDomain = Atto.parseOnly (domainParser <* Atto.endOfInput) . Text.E.encodeUtf8
+mkDomain = Atto.parseOnly (domainParser Nothing <* Atto.endOfInput) . Text.E.encodeUtf8
+
+-- | In MLS, a 'GroupId' can be up to 256 bytes. In the chosen encoding, the
+-- first 16 bytes correspond to the conversation UUID, where the rest
+-- corresponds to the domain. This means that the domain can be at most 240
+-- bytes.
+mkMLSDomain :: Text -> Either String Domain
```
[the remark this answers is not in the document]
mdimjasevic: I reverted all changes to the `Data.Domain` module.

docs/reference/cassandra-schema.cql
```diff
@@ -311,17 +347,14 @@ CREATE TABLE galley_test.user (
...
-    user_remote_id uuid,
-    conversation_role text,
-    PRIMARY KEY (conv, user_remote_domain, user_remote_id)
-) WITH CLUSTERING ORDER BY (user_remote_domain ASC, user_remote_id ASC)
-    AND bloom_filter_fp_chance = 0.1
+CREATE TABLE galley_test.group_id_conv_id (
+    group_id blob PRIMARY KEY,
+    conv_id uuid,
+    domain text
+) WITH bloom_filter_fp_chance = 0.01
```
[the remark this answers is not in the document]
jschaul: I tried to reproduce this locally by compiling your branch, doing a db-migrate, then a make git-add-cassandra-schema; I get the same ordering you get (i.e. the one with the big diff here). There's also https://issues.apache.org/jira/browse/CASSANDRA-14534 which indicates there is no guarantee for ordering of these objects.
I'd just go with what you have, i.e. the bigger diff. The actual changes can be seen in the schema migrations, this file can be seen as a "state-of-now" snapshot, rather than a useful too to see changes over time, for that we have the schema migrations.

services/galley/src/Galley/Cassandra/Conversation.hs
```diff
@@ -78,8 +84,18 @@ createConversation (NewConversation ty usr acc arole name mtid mtimer recpt user
...
+        convGroupId = groupId
       }
+  where
+    randomGroupId :: MonadIO m => m GroupId
+    randomGroupId = do
+      g <- newStdGen
+      let (len, g2) = genWord8 g
+          bs = fromShort . fst $ genShortByteString 256 g2
+      -- The length can be at most 256 bytes
+      pure . GroupId $ Blake2.hash (fromIntegral len) mempty bs
```
pcapriotti, line 98: What's the point of hashing here?

pcapriotti (approved):
Looks good to merge once the issues below are resolved.

libs/wire-api/src/Wire/API/Conversation.hs
```diff
@@ -621,7 +651,9 @@ data NewConv = NewConv
     newConvMessageTimer :: Maybe Milliseconds,
     newConvReceiptMode :: Maybe ReceiptMode,
     -- | Every member except for the creator will have this role
-    newConvUsersRole :: RoleName
+    newConvUsersRole :: RoleName,
+    -- | The protocol of the conversation. It can be Proteus or MLS (1.0).
+    newConvProtocol :: Protocol
```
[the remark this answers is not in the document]
mdimjasevic: This is the type used as the request body of the conversation creation endpoint. If the group ID was part of this type, that would mean the API user would set the group ID. Therefore, it is not part of the type. Does that make sense?

services/galley/src/Galley/Cassandra/Conversation.hs
mdimjasevic: Should I skip that step, i.e., should I construct the group ID like `pure . GroupId $ bs`, where `bs` is defined on line 96 above?

mdimjasevic: Should I skip that step, i.e., should I construct the group ID like `pure . GroupId $ bs`, where `bs` is defined on line 96 above?

mdimjasevic: Should I skip that step, i.e., should I construct the group ID like `pure . GroupId $ bs`, where `bs` is defined on line 96 above?

libs/wire-api/src/Wire/API/Conversation.hs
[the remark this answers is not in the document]
pcapriotti: Yes, I misread the diff.

PR #2152, opened 2022-02-23
Set login cookies with a versioned access path

pcapriotti:
## Checklist

### DIFF
--- a/deploy/services-demo/conf/nginz/nginx.conf
+++ b/deploy/services-demo/conf/nginz/nginx.conf
@@ -194,7 +194,7 @@ http {
         proxy_pass http://brig;
     }
 
-    location /register {
+    location ~* (/v[0-9]+)?/register {
       include common_response_no_zauth.conf;
       proxy_pass http://brig;
     }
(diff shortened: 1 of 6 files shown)


jschaul (commented):
The lifetime of the api version evolution on backends and clients is separate from the lifecycle of cookies on a user's device.

Currently, as long as a user uses their device at least once every X (default on wire production = 56 days), they continually refresh their cookies and access tokens and are never prompted for a login.

Say a user's device currently has a cookie with `http path` `/v3/access`. Now, a new version of backend and client is released that understands `v4` and they both agree to use that. That means the client will make a call to `/v4/access` but the (existing, valid) cookie won't be sent along, leading to the user getting logged out and having to log in again.

This means users will get logged out each time a new version starts being supported by both clients and backends. That's a serious usability deterioration, in a tradeoff for... what exactly? What do we gain by versioning /access?

pcapriotti:
Abandoned in favour of keeping `/access` available at its unversioned path.

fisx:
> Abandoned in favour of keeping /access available at its unversioned path.

good solution!  maybe you can leave a comment in the code somewhere, where future me will find it when he tries the same mistake?

PR #2102, opened 2022-02-04
MLS key packages

pcapriotti:
Implementation of MLS key package API.

## Checklist

 - [ ] The **PR Title** explains the impact of the change.
 - [ ] The **PR description** provides context as to why the change should occur and what the code contributes to that effect. This could also be a link to a JIRA ticket or a Github issue, if there is one.
 - [ ] If HTTP endpoint paths have been added or renamed, the **endpoint / config-flag checklist** (see Wire-employee only backend [wiki page](https://github.com/zinfra/backend-wiki/wiki/Checklists)) has been followed.
 - [ ] If a cassandra schema migration has been added, I ran **`make git-add-cassandra-schema`** to update the cassandra schema documentation.
 - [ ] **changelog.d** contains the following bits of information ([details](https://github.com/wireapp/wire-server/blob/develop/docs/developer/changelog.md)):
   - [ ] A file with the changelog entry in one or more suitable sub-sections. The sub-sections are marked by directories inside `changelog.d`.
   - [ ] If new config options introduced: added usage description under docs/reference/config-options.md
   - [ ] If new config options introduced: recommended measures to be taken by on-premise instance operators.
   - [ ] If a cassandra schema migration is backwards incompatible (see also [these docs](https://github.com/wireapp/wire-server/blob/develop/docs/developer/cassandra-interaction.md#cassandra-schema-migrations)), measures to be taken by on-premise instance operators are explained.
   - [ ] If a data migration (not schema migration) introduced: measures to be taken by on-premise instance operators.
   - [ ] If public end-points have been changed or added: does nginz need un upgrade?
   - [ ] If internal end-points have been added or changed: which services have to be deployed in a specific order?

PR #2154, opened 2022-02-24
Add crypto-cli to the deps docker image

pcapriotti:
The crypto-cli tool is going to be used in integration test images for testing MLS functionality. See for example https://github.com/wireapp/wire-server/pull/2102.

## Checklist

### DIFF
--- a/build/ubuntu/Dockerfile.deps
+++ b/build/ubuntu/Dockerfile.deps
@@ -3,17 +3,26 @@ FROM ubuntu:20.04 as cryptobox-builder
 # compile cryptobox-c
 RUN export DEBIAN_FRONTEND=noninteractive && \
     apt-get update && \
-    apt-get install -y cargo file libsodium-dev git pkg-config && \
+    apt-get install -y cargo file libsodium-dev git pkg-config make && \
     cd /tmp && \
     git clone https://github.com/wireapp/cryptobox-c.git && \
     cd cryptobox-c && \
     export SODIUM_USE_PKG_CONFIG=1 && \
     cargo build --release
 
+# compile core-crypto cli tool
+RUN cd /tmp && \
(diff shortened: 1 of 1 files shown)


build/ubuntu/Dockerfile.deps
```diff
@@ -3,18 +3,27 @@ FROM ubuntu:20.04 as cryptobox-builder
...
+  cd core-crypto/cli && \
+  cargo build --release
+
 # Minimal dependencies for ubuntu-compiled, dynamically linked wire-server Haskell services
 FROM ubuntu:20.04
 
 COPY --from=cryptobox-builder /tmp/cryptobox-c/target/release/libcryptobox.so /usr/lib
 
+# FUTUREWORK: only copy crypto-cli executable if we are building an integration test image
+COPY --from=cryptobox-builder /tmp/core-crypto/target/release/crypto-cli /usr/bin
```
fisx, line 25: Do you want to `rm -rf /tmp/core-crypto` here?  Or are you confident this happens before the image is done?

fisx (approved):
One question, otherwise LGTM!

build/ubuntu/Dockerfile.deps
pcapriotti: My docker-fu isn't great, but my understanding is that the second `FROM ubuntu:20.04` creates a new stage, and the stuff from the old image is discarded, unless explicitly copied.

This seems to confirm it:
```
▶ docker run --rm -it quay.io/wire/ubuntu20-deps
root@518ee8a4b142:/# which crypto-cli
/usr/bin/crypto-cli
root@518ee8a4b142:/# ls /tmp/core-crypto
ls: cannot access '/tmp/core-crypto': No such file or directory
```

PR #2157, opened 2022-02-24
Fix version bounds.

fisx:
## Checklist

### DIFF
--- a/changelog.d/5-internal/bump-aeson-amazonka
+++ b/changelog.d/5-internal/bump-aeson-amazonka
@@ -1,1 +1,1 @@
-Bump aeson to v2.0.3.0 and update amazonka fork from upstream repository.
+Bump aeson to v2.0.3.0 and update amazonka fork from upstream repository.  (#2153, #2157)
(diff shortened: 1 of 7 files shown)


mdimjasevic (commented):
Why putting an upper bound on `aeson`?

fisx:
> Why putting an upper bound on `aeson`?

same reason we always put upper version bounds.  semantic versioning rules about legal breaking changes.

pcapriotti:
Probably a stupid question, but couldn't we just get rid of all the version bounds and simply rely on cabal.freeze?

mdimjasevic:
> Probably a stupid question, but couldn't we just get rid of all the version bounds and simply rely on cabal.freeze?

@fisx , I'm curious about the same: can't we address the concern via the freeze file?

fisx:
not a stupid question, this just goes against the idea of how packages work as self-contained entities, especially library packages.  one could argue that we don't make those, and they are not self-contained for many other reasons besides the inconsistently applied version bounds in the cabal files.  i just had the urge to clean up a little bit.

i think this discussion is becoming much bigger than this PR, though.  :)

pcapriotti (approved):

PR #2164, opened 2022-02-28
provide a default cabal project file default through direnv

jschaul:
Aiming to provide a sane default for everyone using direnv and compiling with cabal. Direnv shouldn't be in use by CI.

Allows writing custom options in cabal.project.local-custom; otherwise, if that file does not exist, uses the `-default` one.

### DIFF
--- a/.envrc
+++ b/.envrc
@@ -1,5 +1,11 @@
 env=$(nix-build --no-out-link "$PWD/direnv.nix")
 PATH_add "${env}/bin"
 
+if [[ -f cabal.project.local-custom ]]; then
+    cp cabal.project.local-custom cabal.project.local
+else
+    cp cabal.project.local-default cabal.project.local
+fi
+
 # allow local .envrc overrides
 [[ -f .envrc.local ]] && source_env .envrc.local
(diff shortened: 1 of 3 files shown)


smatting (approved):
I think this would work well!

akshaymankar (commented):
I think it is better to use the `hack/bin/cabal-project-local-template.sh` for this, instead of `cabal.project.local-default`. And, if you really want it regenerated on every direnv reload, you can add it to your `.envrc.local` file with whatever options you like. I have manually tuned all packages for maximum CPU utilization with different `-j` flags, I could move my config to `cabal.project.local-custom`, but honestly this is getting a bit too complex now. That said, I don't oppose this, if it makes your life easier. So not requesting changes.

smatting:
The advantage of this PR over customer .envr.local is that newcomers will get sane defaults

pcapriotti:
Why add yet another file that we have to keep in sync, when there's already a script that is able to generate it and that needs to be used anyway to tweak it? Instead of this, we can maybe add something to `.envrc` that warns you when `cabal.project.local` is missing.

jschaul:
Obsolete.

PR #2166, opened 2022-03-01
Add qualified broadcast endpoint

pcapriotti:
This adds a qualified version of the message broadcast endpoint, with the same API as the one for normal messages. This makes it possible for the clients to uniformly use the qualified API for all possible message types.

Tracked by https://wearezeta.atlassian.net/browse/SQCORE-1257.

## Checklist

### DIFF
--- a/deploy/services-demo/conf/nginz/nginx.conf
+++ b/deploy/services-demo/conf/nginz/nginx.conf
@@ -337,7 +337,7 @@ http {
     }
 
     location /broadcast {
-      include common_response_no_zauth.conf;
+      include common_response_with_zauth.conf;
       proxy_pass http://galley;
     }
 
--- a/deploy/services-demo/conf/nginz/nginx.conf
+++ b/deploy/services-demo/conf/nginz/nginx.conf
@@ -336,8 +336,8 @@ http {
         proxy_pass http://galley;
     }
 
-    location /broadcast/otr/messages {
-      include common_response_with_zauth.conf;
+    location /broadcast {
+      include common_response_no_zauth.conf;
       proxy_pass http://galley;
     }
 
--- a/services/galley/src/Galley/API/Update.hs
+++ b/services/galley/src/Galley/API/Update.hs
@@ -52,6 +52,7 @@ module Galley.API.Update
     -- * Talking
     postProteusMessage,
     postOtrMessageUnqualified,
+    postProteusBroadcast,
     postOtrBroadcastUnqualified,
     isTypingUnqualified,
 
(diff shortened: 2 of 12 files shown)


services/galley/src/Galley/API/Update.hs
```diff
@@ -1114,6 +1115,32 @@ postProteusMessage sender zcon conv msg = runLocalInput sender $ do
...
+       TeamStore,
+       TinyLog
+     ]
+    r =>
+  Local UserId ->
+  ConnId ->
+  QualifiedNewOtrMessage ->
+  Sem r (PostOtrResponse MessageSendingStatus)
+postProteusBroadcast sender zcon msg =
+  runLocalInput sender $
```
mdimjasevic, line 1141: I don't understand why this works. `postBroadcast` does not have a member `Input (Local ())` in its constraints, so where does it come from?

services/galley/test/integration/API/Util.hs
```diff
@@ -1964,10 +2019,13 @@ assertMismatchQualified failedToSend missing redundant deleted = do
 
   assertExpected "deleted" deleted (fmap mssDeletedClients . responseJsonMaybe)
 
-otrRecipients :: [(UserId, [(ClientId, Text)])] -> OtrRecipients
-otrRecipients = OtrRecipients . UserClientMap . buildMap
-  where
-    buildMap = fmap Map.fromList . Map.fromList
+otrRecipients :: [(UserId, ClientId, Text)] -> OtrRecipients
```
mdimjasevic, line 2022: I forgot the slight differences between these types of user ID to ClientId to something mapping, but I remember our conclusion is that they're not quite the same. For example, the prior type that you removed on line 1967 allows for a UserId to map to an empty list, whereas the type you introduced on line 2022 (and in other places as well) does not allow it. Are you sure this change is OK?

services/galley/test/integration/API/Teams.hs
```diff
@@ -1613,8 +1624,8 @@ testUpdateTeamStatus = do
         const 403 === statusCode
         const "invalid-team-status-update" === (Error.label . responseJsonUnsafeWithMsg "error label")
 
-postCryptoBroadcastMessageJson :: TestM ()
-postCryptoBroadcastMessageJson = do
+postCryptoBroadcastMessageJson :: Broadcast -> TestM ()
```
mdimjasevic, line 1627: Maybe you could rename this function now that the content type is the input.

services/galley/test/integration/API/Teams.hs
```diff
@@ -1663,13 +1674,12 @@ postCryptoBroadcastMessageJson = do
         void . liftIO $
           WS.assertMatch t wsA2 (wsAssertOtr (q (selfConv alice)) (q alice) ac ac2 (toBase64Text "ciphertext0"))
 
-postCryptoBroadcastMessageJsonFilteredTooLargeTeam :: TestM ()
-postCryptoBroadcastMessageJsonFilteredTooLargeTeam = do
+postCryptoBroadcastMessageJsonFilteredTooLargeTeam :: Broadcast -> TestM ()
```
mdimjasevic, line 1677: ditto.

deploy/services-demo/conf/nginz/nginx.conf
```diff
@@ -336,8 +336,8 @@ http {
         proxy_pass http://galley;
     }
 
-    location /broadcast/otr/messages {
-      include common_response_with_zauth.conf;
+    location /broadcast {
+      include common_response_no_zauth.conf;
```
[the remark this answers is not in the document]
pcapriotti: Oh, my mistake, it's a leftover from some manual tests that I ran. Fixed in 9d315e5cac89fae089d40108c447c4b245f3a330.

PR #2167, opened 2022-03-01
Strip versions from ibis paths

pcapriotti:
This makes it possible to access external API endpoints not implemented in wire-server with a version prefix.

## Checklist


mdimjasevic (approved):
I'm not too familiar with the templating language, but I assume you've checked the outcome in some way.

PR #2175, opened 2022-03-03
Forward local MLS Welcome messages

pcapriotti:
Forwarding of MLS welcome messages to local users only.

 - [x] Add `/mls/welcome` endpoint stub
 - [x] Add wrapper to hold raw MLS objects
 - [x] Unit tests
 - [x] Forward welcome messages to local users
 - [x] Integration tests

Validation of welcome message might not be possible at this stage, since with the handshake processing logic still missing, we wouldn't have the correct database state in tests.

Tracked by https://wearezeta.atlassian.net/browse/FS-436.

## Checklist


libs/wire-api/src/Wire/API/MLS/Extension.hs
```diff
@@ -0,0 +1,144 @@
...
+  _ == _ = False
+
+instance Show SomeExtension where
+  show (SomeExtension SCapabilitiesExtensionTag caps) = show caps
+  show (SomeExtension SLifetimeExtensionTag lt) = show lt
+
+decodeExtension :: Extension -> Either Text (Maybe SomeExtension)
+decodeExtension e = do
+  case toMLSEnum' (extType e) of
+    Left MLSEnumUnkonwn -> pure Nothing
```
mdimjasevic, line 107: A typo in `Unkonwn`.

libs/wire-api/src/Wire/API/MLS/Serialisation.hs
```diff
@@ -121,10 +128,51 @@ decodeMLS' = decodeMLS . LBS.fromStrict
...
+-- | An MLS value together with its serialisation.
+--
+-- This can be used whenever we need to parse an object, but at the same time
+-- retain the original serialised bytes (e.g. for signature verification, or to
+-- forward them verbatim).
+data RawMLS a = RawMLS
+  { rmRaw :: ByteString,
+    rmValue :: a
+  }
+  deriving stock (Eq, Show, Functor, Foldable, Traversable)
```
mdimjasevic, line 152: What does it mean to have a functor instance for this type? When we map over and change `rmValue`, `rmRaw` stays the same, but that `rmRaw` has nothing to do with the new `rmValue`. Is that OK?

libs/wire-api/src/Wire/API/MLS/Extension.hs
pcapriotti: Done in 5c301984a49c8f4227e77914fa45a3c4661c0c84.

libs/wire-api/src/Wire/API/MLS/Serialisation.hs
pcapriotti: Well, technically it is a valid functor and comonad, but you're right that it might be slightly misleading. I guess there is an unwritten assumption here that the `rmRaw` field is somehow the serialised form of the corresponding `rmValue`, but of course that's not enforced in any sense. To avoid any confusion, I removed the `Comonad` instance in 3b8a62e61d3d7c947d42b341d4330f4a1aa8fafe.

services/galley/src/Galley/API/MLS.hs
```diff
@@ -34,15 +40,18 @@ import Wire.API.MLS.Welcome
...
+       Error UnknownWelcomeRecipient,
+       Input UTCTime
      ]
     r =>
   Local UserId ->
   RawMLS Welcome ->
   Sem r ()
-postMLSWelcome _ wel = do
+postMLSWelcome lusr wel = do
+  now <- input
```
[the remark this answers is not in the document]
pcapriotti: Doesn't really make a difference, but sure: d9a3042e6b1365f5ed7bdb8007a353b601842854.

libs/wire-api/test/golden/Test/Wire/API/Golden/Generated.hs
```diff
@@ -645,7 +645,9 @@ tests =
             (Test.Wire.API.Golden.Generated.Event_user.testObject_Event_user_11, "testObject_Event_user_11.json"),
             (Test.Wire.API.Golden.Generated.Event_user.testObject_Event_user_12, "testObject_Event_user_12.json"),
             (Test.Wire.API.Golden.Generated.Event_user.testObject_Event_user_13, "testObject_Event_user_13.json"),
-            (Test.Wire.API.Golden.Generated.Event_user.testObject_Event_user_14, "testObject_Event_user_14.json")
+            (Test.Wire.API.Golden.Generated.Event_user.testObject_Event_user_14, "testObject_Event_user_14.json"),
```
mdimjasevic, line 648: Wow, we dropped this test! Thanks for noticing!

mdimjasevic (approved):
Looks good!

PR #2177, opened 2022-03-04
Redundant constraints [skip ci]

fisx:
Probably best to read commit-by-commit; 331f8bc2a4608a731b8eba5db3fe9913f058833b contains only the effects of rerunning hpack and can probably be ignored entirely.

## Checklist


fisx:
never mind, i remembered why we don't this.  `HasCallStack` and "but i want to restrict this more, it makes more sense then!"

PR #2178, opened 2022-03-07
Tweak wording in reference docs

fisx:
## Checklist

### DIFF
--- a/docs/reference/config-options.md
+++ b/docs/reference/config-options.md
@@ -396,7 +396,7 @@ Additionally if `setSftListAllServers` is set to `enabled` (disabled by default)
 
 #### setDefaultLocale (deprecated / ignored)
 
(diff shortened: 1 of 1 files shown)


battermann (approved):

pcapriotti (approved):
Looks good.

One minor suggestion: instead of "it is the default", I would maybe write "the default is chosen".

docs/reference/config-options.md
```diff
@@ -410,7 +410,7 @@ optSettings:
 
 #### `setDefaultUserLocale`
 
-This option is the default user locale to be used if it is not set in the user profile. This can be the case if the users are provisioned by SCIM e.g. This option determines which language to use for email communication. If not set the default is the value that is configured for `setDefaultTemplateLocale`.
+This option determines which language to use for email communication. It is the default if not set in the user profile, or if no user profile exists (eg., if user is being provisioned via SCIM or manual team invitation via the team management app). If not set, `setDefaultTemplateLocale` is used instead.
```
fisx, line 413: @pcapriotti are you referring to this?

```suggestion
This option determines which language to use for email communication. It is the default value if none is given in the user profile, or if no user profile exists (eg., if user is being provisioned via SCIM or manual team invitation via the team management app). If not set, `setDefaultTemplateLocale` is used instead.
```

To me, "the default is chosen" would be more confusing than the original wording.  What do you think of this suggestion?

PR #2179, opened 2022-03-07
fix setDefaultUserLocale mapping

battermann:
## Checklist

 - [x] The **PR description** provides context as to why the change should occur and what the code contributes to that effect.


fisx (approved):
Even if this would work with missing `.setDefaultUserLocale`, I can't see how this change could hurt!  :+1:

PR #2174, opened 2022-03-02
Fix spar user provisioning corner case

fisx:
https://wearezeta.atlassian.net/browse/CS-742

With the tests in this PR, but without the changes, I get this error:

```
Failures:

  test-integration/Test/Spar/Scim/UserSpec.hs:264:7:
  1) WireIdPAPIV1.Spar.Scim.User, Create with TM invitation; then re-provision with SCIM, True
       "<?xml version=\"1.0\" encoding=\"UTF-8\"?><!DOCTYPE html PUBLIC \"-//W3C//DTD XHTML 1.1//EN\" \"http://www.w3.org/TR/xhtml11/DTD/xhtml11.dtd\"><html xml:lang=\"en\" xmlns=\"http://www.w3.org/1999/xhtml\"><head>  <title>wire:sso:error:not-found</title>   <script type=\"text/javascript\">       const receiverOrigin = '*';       window.opener.postMessage({\"payload\":{\"errors\":[\"Could not find SAML credentials, and auto-provisioning is disabled.\"],\"label\":\"forbidden\"},\"type\":\"AUTH_ERROR\"}, receiverOrigin);   </script></head></html>" does not contain "<title>wire:sso:success</title>"

  To rerun use: --match "/WireIdPAPIV1/Spar.Scim.User/Create with TM invitation; then re-provision with SCIM/True/"

  test-integration/Test/Spar/Scim/UserSpec.hs:264:7:
  2) WireIdPAPIV2.Spar.Scim.User, Create with TM invitation; then re-provision with SCIM, True
  [...]

Randomized with seed 55186091

Finished in 212.9946 seconds
527 examples, 2 failures, 67 pending
```

The changes in the later commits fix this.  The problem was that under certain conditions, spar did not create SAML credentials during SCIM provisioning.

## Checklist

 - [x] The **PR Title** explains the impact of the change.
 - [x] The **PR description** provides context as to why the change should occur and what the code contributes to that effect. This could also be a link to a JIRA ticket or a Github issue, if there is one.
 - [x] **changelog.d** contains the following bits of information ([details](https://github.com/wireapp/wire-server/blob/develop/docs/developer/changelog.md)):
   - [x] A file with the changelog entry in one or more suitable sub-sections. The sub-sections are marked by directories inside `changelog.d`.

PR #2181, opened 2022-03-07
Release 2022-03-07

zebot:
# [2022-03-07]

## Release notes

* For wire.com operators: make sure that nginz is deployed (#2166)

## API changes

* Add qualified broadcast endpoint (#2166)

* Always create spar credentials during SCIM provisioning when applicable (#2174)

## Internal changes

* Add tests for additional information returned by `GET /api-version` (#2159)

* Clean up `Base64ByteString` implementation (#2170)

* The `Event` record type does not contain a `type` field anymore (#2160)

* Add MLS message types and corresponding deserialisers (#2145)

* Servantify `POST /register` and `POST /i/users` endpoints (#2121)


akshaymankar (changes requested):
Looks like you removed a commit and force-pushed to the release branch. If we merge this into master, it will make master diverge from develop in an annoying way where same work exists in two separate commits on master and develop. Can you please instead revert the commit on develop and then create a release PR?

PR #2183, opened 2022-03-07
Merge master

fisx:
This also recovers #2162, which has been omitted in the last release.


akshaymankar (commented):
Looks like we lost the changelog for #2162, please add that back.

fisx:
> Looks like we lost the changelog for #2162, please add that back.

Can you please look again?  It doesn't show in the diff because it hasn't been removed, so it's still there!

PR #2187, opened 2022-03-08
[Polysemy] Remove MonadClient from AppT in Brig

mdimjasevic:
The PR generalises Brig by introducing parameterised monads wherever possible in Brig and by removing the `MonadClient` constraint from Brig's main monad `AppT`. This will make it easier to put the `Sem` monad instead of `m` in `AppT` in an upcoming PR. A new function `wrapClient` is introduced that can be used as a crutch instead of the `MonadClient` instance while polysemisation of Brig is in progress.

## Checklist


services/brig/src/Brig/API/Internal.hs
```diff
@@ -435,14 +435,14 @@ changeAccountStatusH (usr ::: req) = do
...
-  status <- lift $ API.lookupStatus usr
+  status <- lift $ wrapClient $ API.lookupStatus usr
   return $ case status of
     Just s -> json $ AccountStatusResp s
     Nothing -> setStatus status404 empty
 
 getConnectionsStatusUnqualified :: ConnectionsStatusRequest -> Maybe Relation -> (Handler r) [ConnectionStatus]
 getConnectionsStatusUnqualified ConnectionsStatusRequest {csrFrom, csrTo} flt = lift $ do
-  r <- maybe (API.lookupConnectionStatus' csrFrom) (API.lookupConnectionStatus csrFrom) csrTo
+  r <- maybe (wrapClient $ API.lookupConnectionStatus' csrFrom) (wrapClient . API.lookupConnectionStatus csrFrom) csrTo
```
isovector, line 445: maybe this instead?

```suggestion
  r <- wrapClient $ maybe (API.lookupConnectionStatus' csrFrom) (API.lookupConnectionStatus csrFrom) csrTo
```

services/brig/src/Brig/Data/Connection.hs
```diff
@@ -274,23 +281,27 @@ countConnections u r = do
...
 
-deleteConnections :: UserId -> (AppIO r) ()
+deleteConnections :: UserId -> AppIO r ()
 deleteConnections u = do
-  runConduit $
-    paginateC contactsSelect (paramsP LocalQuorum (Identity u) 100) x1
-      .| C.mapM_ (pooledMapConcurrentlyN_ 16 delete)
-  retry x1 . write connectionClear $ params LocalQuorum (Identity u)
-  retry x1 . write remoteConnectionClear $ params LocalQuorum (Identity u)
+  e <- ask
```
isovector, line 289: unnecessary?

services/brig/src/Brig/Email.hs
```diff
@@ -49,7 +49,7 @@ import Imports
 import Network.Mail.Mime
 
 -------------------------------------------------------------------------------
-sendMail :: Mail -> (AppIO r) ()
+sendMail :: (MonadIO f, MonadReader Env f) => Mail -> f ()
```
isovector, line 52:
```suggestion
sendMail :: (MonadIO m, MonadReader Env m) => Mail -> m ()
```

services/brig/src/Brig/IO/Intra.hs
```diff
@@ -254,14 +255,17 @@ dispatchNotifications orig conn e = case e of
...
 notifyUserDeletionRemotes deleted = do
-  runConduit $
-    Data.lookupRemoteConnectedUsersC deleted (fromInteger (natVal (Proxy @UserDeletedNotificationMaxConnections)))
-      .| C.mapM_ fanoutNotifications
+  e <- ask
+  fmap
+    wrapClient
+    runConduit
+    $ Data.lookupRemoteConnectedUsersC deleted (fromInteger (natVal (Proxy @UserDeletedNotificationMaxConnections)))
+      .| C.mapM_ (liftIO . runAppT e . fanoutNotifications)
```
isovector, line 268: Same question ; why is there a `runAppT` now?

services/brig/src/Brig/User/Auth.hs
```diff
@@ -78,78 +80,88 @@ data Access u = Access
...
+  mapExceptT wrapClient (Data.authenticate uid pw) `catchE` \case
+    AuthInvalidUser -> mapExceptT wrapClient $ loginFailed mLimitFailedLogins uid
+    AuthInvalidCredentials -> mapExceptT wrapClient $ loginFailed mLimitFailedLogins uid
     AuthSuspended -> throwE LoginSuspended
     AuthEphemeral -> throwE LoginEphemeral
     AuthPendingInvitation -> throwE LoginPendingActivation
   newAccess @ZAuth.User @ZAuth.Access uid typ label
 login (SmsLogin phone code label) typ = do
-  uid <- resolveLoginId (LoginByPhone phone)
+  uid <- mapExceptT wrapClient $ resolveLoginId (LoginByPhone phone)
```
isovector, line 140: Maybe we should add a helper function `wrapClientE = mapExceptT wrapClient`?

isovector (approved):
looks great!

services/brig/src/Brig/Email.hs
mdimjasevic: Sure, I applied this suggestion. While your suggestion works because values and types live in different name spaces in Haskell, I tend to avoid name ambiguity as I find it easier to read code when one name can be only one thing in the given context.

services/brig/src/Brig/Provider/API.hs
```diff
@@ -767,34 +784,37 @@ data UpdateServiceWhitelistResp
...
+          wrapClient
+          runConduit
+          $ User.lookupServiceUsersForTeam pid sid tid
             .| C.mapM_
-              ( pooledMapConcurrentlyN_
-                  16
-                  ( \(bid, cid) ->
-                      deleteBot uid (Just con) bid cid
-                  )
+              ( liftIO . runAppT e
```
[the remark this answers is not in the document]
mdimjasevic: That's a good idea. In an upcoming commit I've introduced `runAppIOLifted` and put it to use in places like above.

services/brig/src/Brig/User/Auth.hs
mdimjasevic: Thanks! I just added it in an upcoming commit.

PR #2192, opened 2022-03-10
Map claimed key packages to clients

pcapriotti:
This PR implements a mapping of key package ref to qualified clients. The corresponding cassandra table also contains fields to hold the qualified conversation ID of where the key package is being used. These fields are not set for the moment, but they will play a role when validating incoming handshake messages.

This also adds an internal brig endpoint to dereference a key package. This is only used in tests for the moment, but it will later be used by galley to resolve a key package ref to a user and client ID.

Part of https://wearezeta.atlassian.net/browse/FS-436.

## Checklist


mdimjasevic (approved):
Looks good!

mdimjasevic:
The CI gave a green light, but there's a failure in communicating this here in the status part of the page.

PR #2196, opened 2022-03-10
Start coding style guide.

fisx:
https://wearezeta.atlassian.net/browse/WPB-7271

I'm starting this in the hope that people will like it.  It's a bit like ormolu: annoying at times, but it will free so much energy that currently goes into discussions about things that have been agreed on at some point in the past.

The rules I'm starting this with have been agreed on a long time ago, and I'm not sure what the current mood is, so I'm seeking feedback on two questions:

- is a style guide a tolerable idea?
- is any of the two rules i'm starting with acceptable?

I'm also curious if you're also eager to add something, but I'm perfectly happy to be the only one restricting our freedom in these ways.  :)

### DIFF
--- a/docs/src/developer/developer/coding-style.md
+++ b/docs/src/developer/developer/coding-style.md
@@ -13,7 +13,9 @@ doing things, whoever can point to a rule in this document wins the
 argument.
 
 Where feasible and if there is time, rules should be enforced
-automatically.
+automatically.  What we can enforce in code does not need to be
+covered here.  The style guide is for things we don't (yet) enforce in
+other ways.
 
 
 ### Integer literals
--- a/docs/src/developer/developer/coding-style.md
+++ b/docs/src/developer/developer/coding-style.md
@@ -48,7 +48,8 @@ also go through the C&D history and pick details on that, and other decisions.
 
 All modules must have explicit export lists.  Reasons:
 
-- This is good documentation;
+- It helps the compiler with optimizations across module boundaries;
+- it is good documentation;
 - it helps with identifying dead code (if it's not exported and not used, we get a compiler warning); and
 - prevents us from using functions that aren't part of a module's interface by accident: if we *want* to use them, we can always add them, but then we are transparent about changing the interface.
 
--- a/docs/src/developer/developer/coding-style.md
+++ b/docs/src/developer/developer/coding-style.md
@@ -38,7 +38,7 @@ Instead of the standard `Prelude`, we use the module `Imports` from
 
 All libraries that are or want to be on hackage also cannot depend on `/libs/import`, namely:
 
-- `/libs/http2-manager`-
+- `/libs/http2-manager`
 - `/libs/hscim`
 
 `string-conversions` is forbidden for production code, since some of
--- a/docs/src/developer/developer/coding-style.md
+++ b/docs/src/developer/developer/coding-style.md
@@ -15,12 +15,14 @@ argument.
 Where feasible and if there is time, rules should be enforced
 automatically.
 
+
 ### Integer literals
 
 Integer literals are spelled `50_000` (not `50 * 1000` and not
 `50000`).  (NB: ghc allows to write things like `40_0000`, which makes
 sense eg. in Chinese.  This is illegal in this code base for obvious
 reasons.)
+
 
 ### Libraries and dependencies
 
@@ -47,3 +49,11 @@ All modules must have explicit export lists.  Reasons:
 - This is good documentation;
 - it helps with identifying dead code (if it's not exported and not used, we get a compiler warning); and
 - prevents us from using functions that aren't part of a module's interface by accident: if we *want* to use them, we can always add them, but then we are transparent about changing the interface.
+
+
+### TODO
+
+Import (relevant parts of) our internal C&D logs:
+
+- https://wearezeta.atlassian.net/wiki/spaces/BAC/pages/824148311/Archive+up+to+2023-09-27
+- https://wearezeta.atlassian.net/wiki/spaces/BAC/pages/384928204/Cuss+Discuss+How+Stuff+Works+meeting+minutes
--- a/docs/src/developer/developer/coding-style.md
+++ b/docs/src/developer/developer/coding-style.md
@@ -36,6 +36,11 @@ Instead of the standard `Prelude`, we use the module `Imports` from
 - `/libs/imports` (there is some ongoing debate whether that exception should be removed, though.)
 - `/libs/wire-api` (compiling wire-api takes longer, and there was a small majority in favor of testing serialization here, not (only) in golden tests.)
 
+All libraries that are or want to be on hackage also cannot depend on `/libs/import`, namely:
+
+- `/libs/http2-manager`-
+- `/libs/hscim`
+
 `string-conversions` is forbidden for production code, since some of
 the transformations throw impure exceptions (notably `ByteString` to
 `Text`).  It's ok to use it in test code.
--- a/docs/src/developer/developer/coding-style.md
+++ b/docs/src/developer/developer/coding-style.md
@@ -1,0 +1,49 @@
+
+# The wire-server Coding Style Guide
+
+This document contains a collection of rules as they emerge from
+day-to-day discussions.  If you are unhappy with anything, open a PR
+and trigger a discussion.  Coding style PRs of course should be
+reviewed by more or less the entire team if feasible.
+
+These rules are by no means enforced consistently at the time of
+writing them down, but they are offer a way to avoid debate when
+reviewing current pull requests: if there is more than one way of
+doing things, whoever can point to a rule in this document wins the
+argument.
+
+Where feasible and if there is time, rules should be enforced
+automatically.
+
+### Integer literals
+
+Integer literals are spelled `50_000` (not `50 * 1000` and not
+`50000`).  (NB: ghc allows to write things like `40_0000`, which makes
+sense eg. in Chinese.  This is illegal in this code base for obvious
+reasons.)
+
+### Libraries and dependencies
+
+Instead of the standard `Prelude`, we use the module `Imports` from
+`/libs/imports/`.  (Exception: see below.)
+
+`/integration` has some restrictions on what it may depend on:
+
+- `/libs/imports` (there is some ongoing debate whether that exception should be removed, though.)
+- `/libs/wire-api` (compiling wire-api takes longer, and there was a small majority in favor of testing serialization here, not (only) in golden tests.)
+
+`string-conversions` is forbidden for production code, since some of
+the transformations throw impure exceptions (notably `ByteString` to
+`Text`).  It's ok to use it in test code.
+
+
+also go through the C&D history and pick details on that, and other decisions.
+
+
+### Module boundaries
+
+All modules must have explicit export lists.  Reasons:
+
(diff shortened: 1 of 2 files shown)


mdimjasevic:
I don't have a strong opinion on the idea of having a coding style guide. So far I had almost no discussions with teammates about coding style, let alone disputes. For me it is not solving a problem that I have. On the other hand, if the guide is introduced, this means I will be expected to stick to it, meaning I will need more time and energy to write code. I agree this is a local view and if the guide is overall beneficial for the majority working on `wire-server`, then I will not oppose it.

If you will keep on working on the guide, I'd suggest to add a paragraph or two on the motivation and benefits of each rule. For example, why explicit or qualified import lists? What are its pros and cons?

pcapriotti:
> 1. All modules must have explicit export lists.
> 
> 2. All imports must be either explicit (`import Data.Id (UserId)`) or qualified (`import qualified Data.Id`).

I like 1, but I would like to argue against explicit import lists. The way I see it, it just adds a lot of extra work, noise and merge conflicts, for exactly zero benefits. What is the argument for explicit import lists, in fact? And what is the logic behind the prescription you suggest?

I would instead propose something in the opposite direction, along the lines of:

- Modules export lists should be designed so that they can be imported unqualified most of the times. Try to avoid creating name clashes on purpose, and make exported names understandable without a prefix.
- All imports should be implicit, except for modules that are intended to imported qualified. In case of clashes, prefer hiding to qualifying, and qualifying to explicit imports. Reserve explicit imports for modules that export many conflicting names and are awkward to use qualified (e.g. `Control.Lens`).

fisx:
Ok, no consensus in reach there.  :)

I think I may be ok with adjusting to your rules @pcapriotti once I have HLS up and running.  Currently I rely on the import lists to understand where names are defined, that's (probably?) where our disagreement comes from?

Having said that, I really dislike hiding, and I think I will keep disliking it.  But I also dislike bike shedding.  Dilemma!

pcapriotti:
> Ok, no consensus in reach there. :)
> 
> I think I may be ok with adjusting to your rules @pcapriotti once I have HLS up and running. Currently I rely on the import lists to understand where names are defined, that's (probably?) where our disagreement comes from?
> 
> Having said that, I really dislike hiding, and I think I will keep disliking it. But I also dislike bike shedding. Dilemma!

I don't use HLS either. I normally use hoogle to find definitions, and it works pretty well and fast (I have a labelled bookmark in firefox, so I can simply type `z weirdFunction` and get its location immediately).

I also don't like `hiding` much, so I try to use it sparingly. I think it's mostly useful when there are clashes in our own modules, so this can simply be fixed by not creating clashes!

If we were using some other language, I wouldn't be opposed to having more explicit import lists. The problems with haskell that make this basically unworkable are, IMHO:

 - importing qualified without an alias is unusable (and it's hard to keep the aliases consistent)
 - record fields and ADT constructors are in the global scope (so you end up having to import a trillion names just to use a couple of record types)
 - modules can't be nested, and imports cannot be scoped

I think the only solution is to accept that haskell's module system is a joke, and not try to extract too much mileage out of it.

smatting:
I don't mind having a rule to make all imports explicit or qualified. Since I use HLS I can ignore these huge import blocks of text anyways when reading. I would make exception though, e.g. `Imports`.

stephen-smith:
> What is the argument for explicit import lists, in fact?

It prevents breakage when a symbol is added to a module.  If you do unqualified imports with no import list, the PVP (https://pvp.haskell.org/#dependencies-in-cabal) specifically calls out this type of breakage.

I have a mild preference for explicit import lists, especially for packages that are outside our organization.  I find it makes it easier to determine where an unfamiliar symbol is coming from.  HLS is fine when it works, which is why the preference is mild these days, but it's still nice to be able navigate the source with noting but text tools.

stephen-smith (approved):

pcapriotti:
> > What is the argument for explicit import lists, in fact?
> 
> It prevents breakage when a symbol is added to a module. If you do unqualified imports with no import list, the PVP (https://pvp.haskell.org/#dependencies-in-cabal) specifically calls out this type of breakage.'

There's not much value in this. We are pinning dependencies, so such a breakage will only happen when we change the pins, and will result in a compile-time error which is trivial to fix.

> I have a mild preference for explicit import lists, especially for packages that are outside our organization. I find it makes it easier to determine where an unfamiliar symbol is coming from. HLS is fine when it works, which is why the preference is mild these days, but it's still nice to be able navigate the source with noting but text tools.

I don't use HLS, and I don't buy this argument. In fact, it seems to me that people using fancy tools like HLS are more likely to be the ones arguing for explicit imports, because they are such a pain to manage manually. And to get this benefit you really have to use them consistently (including forbidding `Foo(..)` style imports), which becomes unreasonable pretty quickly.

We can navigate the source with grep if we stick to non-clashing names for identifiers. But even without this convention, grep works reasonably well. There's also hoogle (both the public instance and our internal one), which works great most of the times, as I described above.

elland:
Is this something we want to revisit?

docs/src/developer/developer/coding-style.md
```diff
@@ -0,0 +1,49 @@
...
+also go through the C&D history and pick details on that, and other decisions.
+
+
+### Module boundaries
+
+All modules must have explicit export lists.  Reasons:
+
+- This is good documentation;
+- it helps with identifying dead code (if it's not exported and not used, we get a compiler warning); and
+- prevents us from using functions that aren't part of a module's interface by accident: if we *want* to use them, we can always add them, but then we are transparent about changing the interface.
```
MangoIV, line 49: it also helps the compiler with optimizations across module boundaries. :)

docs/src/developer/developer/coding-style.md
```diff
@@ -0,0 +1,49 @@
...
+
+Where feasible and if there is time, rules should be enforced
+automatically.
+
+### Integer literals
+
+Integer literals are spelled `50_000` (not `50 * 1000` and not
+`50000`).  (NB: ghc allows to write things like `40_0000`, which makes
+sense eg. in Chinese.  This is illegal in this code base for obvious
+reasons.)
```
akshaymankar, line 26:
```suggestion
sense eg. in Chinese.  However, this codebase requires underscores
after every 3 digits.)
```

Might not be very obvious a Chinese person (or a South Asian person)

docs/src/developer/developer/coding-style.md
```diff
@@ -0,0 +1,49 @@
...
+
+- `/libs/imports` (there is some ongoing debate whether that exception should be removed, though.)
+- `/libs/wire-api` (compiling wire-api takes longer, and there was a small majority in favor of testing serialization here, not (only) in golden tests.)
+
+`string-conversions` is forbidden for production code, since some of
+the transformations throw impure exceptions (notably `ByteString` to
+`Text`).  It's ok to use it in test code.
+
+
+also go through the C&D history and pick details on that, and other decisions.
```
akshaymankar, line 44:
```suggestion
```

This is not accessible publically, anyway the reasoning is not very detailed in the C&D notes.

docs/src/developer/developer/coding-style.md
```diff
@@ -0,0 +1,61 @@
...
+### Libraries and dependencies
+
+Instead of the standard `Prelude`, we use the module `Imports` from
+`/libs/imports/`.  (Exception: see below.)
+
+`/integration` has some restrictions on what it may depend on:
+
+- `/libs/imports` (there is some ongoing debate whether that exception should be removed, though.)
+- `/libs/wire-api` (compiling wire-api takes longer, and there was a small majority in favor of testing serialization here, not (only) in golden tests.)
+
```
akshaymankar, line 38: missing: hscim and http2-manager. These libraries are or want to be on hackage, it doesn't make sense to depend on imports.

docs/src/developer/developer/coding-style.md
```diff
@@ -0,0 +1,61 @@
...
+also go through the C&D history and pick details on that, and other decisions.
+
+
+### Module boundaries
+
+All modules must have explicit export lists.  Reasons:
+
+- This is good documentation;
+- it helps with identifying dead code (if it's not exported and not used, we get a compiler warning); and
+- prevents us from using functions that aren't part of a module's interface by accident: if we *want* to use them, we can always add them, but then we are transparent about changing the interface.
```
akshaymankar, line 53: IMO this causes too much chore work and the reasons are very weak:
1. The docs are going to be there anyway
2. We should probs just use weeder to do this
3. What does prevention bring anyway?

> it also helps the compiler with optimizations across module boundaries. :)

Sure, but the cost is wasting developer time. I'm willing to bet that these optimizations are not as useful as my time.

docs/src/developer/developer/coding-style.md
akshaymankar: I tried to find the name of this millions, billions, trillions numeral system, but I couldn't.

docs/src/developer/developer/coding-style.md
fisx: thanks!  5db31f5015732882525acb31c5a0108e3fb8cdab

docs/src/developer/developer/coding-style.md
fisx: ...  unless we want to publish imports? :)

yeah, i'll add these exceptions.  we can always change it.

docs/src/developer/developer/coding-style.md
fisx: sorry, this was a TODO that i forgot to remove.

docs/src/developer/developer/coding-style.md
fisx: ok, i'll remove it.  just non-controversial stuff for now.  (i *really* hope that set is non-empty...)

docs/src/developer/developer/coding-style.md
akshaymankar: It will increase deps of these libraries so it won't be a good idea.

docs/src/developer/developer/coding-style.md
```diff
@@ -0,0 +1,54 @@
...
+### Libraries and dependencies
+
+Instead of the standard `Prelude`, we use the module `Imports` from
+`/libs/imports/`.  (Exception: see below.)
+
+`/integration` has some restrictions on what it may depend on:
+
+- `/libs/imports` (there is some ongoing debate whether that exception should be removed, though.)
+- `/libs/wire-api` (compiling wire-api takes longer, and there was a small majority in favor of testing serialization here, not (only) in golden tests.)
+
```
smatting, line 38: let's migrate the README of integrations (or at least the guidlines) here and ask for review? the not compiling `wire-api` is still up for debate I think

smatting:
Let's add a section where we can optionally sign or names? Someting like

> We've read this style guide and are making an effort to follow and and maintain it.
> Signed: @smatting, ...

docs/src/developer/developer/coding-style.md
fisx: > the not compiling `wire-api` is still up for debate I think

it's currently the law, even if some are opposed to the rule.  i would leave it in, and take it out *after* we've changed our mind.

fisx: > let's migrate the README of integrations (or at least the guidlines) here and ask for review?

yes, but not in the initial version?

PR #2197, opened 2022-03-10
Clean up brig internal routing tree.

fisx:
This is a very small change, and even keeps the order of the routes unchanged.  But the new tree structure does not keep most of the account API and some other routes under the `ejpdAPI` node, which didn't make any sense.


fisx:
stale, closed.

PR #2199, opened 2022-03-13
More test labels [skip ci]

fisx:
https://wearezeta.atlassian.net/browse/FS-488


fisx:
i think these changes have been applied in another PR by now...

PR #2205, opened 2022-03-14
SQSERVICES-1300-be-lock-2nd-factor-password-challenge-team-feature-as-default

battermann:
https://wearezeta.atlassian.net/browse/SQSERVICES-1300

## Checklist


services/galley/test/integration/API/Teams/Feature.hs
```diff
@@ -396,7 +396,23 @@ testSimpleFlagWithLockStatus ::
   Public.TeamFeatureStatusValue ->
   Public.LockStatusValue ->
   TestM ()
-testSimpleFlagWithLockStatus defaultStatus defaultLockStatus = do
+testSimpleFlagWithLockStatus = testSimpleFlagWithLockStatus' @a True
+
+testSimpleFlagWithLockStatus' ::
```
fisx, line 401: So you removed the permission in `/libs/galley-types/src/Galley/Types/Teams.hs`, and this is why you have to introduce this special case for the tests?

Why don't we keep the permission, and regulate that no team is allowed to enable this feature by simply not unlocking it?  That would make these changes here unnecessary, no?

fisx (changes requested):

fisx (approved):

PR #2207, opened 2022-03-14
SQSERVICES-1375-block-bot-api-if-2-fa-is-enabled

battermann:
https://wearezeta.atlassian.net/browse/SQSERVICES-1375

## Checklist

### DIFF
--- a/services/brig/src/Brig/IO/Intra.hs
+++ b/services/brig/src/Brig/IO/Intra.hs
@@ -1008,13 +1008,13 @@ getVerificationCodeEnabled tid = do
         . expect2xx
 
 getTeamFeatureStatusSndFactorPasswordChallenge :: Maybe UserId -> (AppIO r) TeamFeatureStatusNoConfig
-getTeamFeatureStatusSndFactorPasswordChallenge = \case
-  Just u -> do
-    resp <- galleyRequest GET (paths ["i", "feature-configs", toByteString' (knownTeamFeatureName @'TeamFeatureSndFactorPasswordChallenge)] . queryItem "user_id" (toByteString' u))
-    pure $ responseJsonUnsafe resp
-  Nothing -> do
-    resp <- galleyRequest GET (paths ["i", "feature-configs", toByteString' (knownTeamFeatureName @'TeamFeatureSndFactorPasswordChallenge)])
-    pure $ responseJsonUnsafe resp
+getTeamFeatureStatusSndFactorPasswordChallenge mbUserId =
+  responseJsonUnsafe
+    <$> galleyRequest
+      GET
+      ( paths ["i", "feature-configs", toByteString' (knownTeamFeatureName @'TeamFeatureSndFactorPasswordChallenge)]
+          . maybe id (queryItem "user_id" . toByteString') mbUserId
+      )
 
 -- | Calls 'Galley.API.updateTeamStatusH'.
 changeTeamStatus :: TeamId -> Team.TeamStatus -> Maybe Currency.Alpha -> (AppIO r) ()
--- a/services/brig/src/Brig/IO/Intra.hs
+++ b/services/brig/src/Brig/IO/Intra.hs
@@ -58,7 +58,7 @@ module Brig.IO.Intra
     changeTeamStatus,
     getTeamSearchVisibility,
     getVerificationCodeEnabled,
-    getTeamFeatureStatusSndFactorPasswordChallengeForSite,
+    getTeamFeatureStatusSndFactorPasswordChallenge,
 
     -- * Legalhold
     guardLegalhold,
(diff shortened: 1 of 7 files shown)


services/brig/src/Brig/IO/Intra.hs
```diff
@@ -1006,6 +1007,19 @@ getVerificationCodeEnabled tid = do
       paths ["i", "teams", toByteString' tid, "features", toByteString' TeamFeatureSndFactorPasswordChallenge]
         . expect2xx
 
+getVerificationCodeEnabledForServer :: (AppIO r) Bool
```
fisx, line 1010:
```suggestion
getVerificationCodeEnabledForServer :: AppIO r Bool
```

fisx (commented):
I would prefer using the team end-point (`GET /i/teams/:tid/features/sndFactorPasswordChallenge`).  You could easily get the team id from the user id or bot id in the handlers; ie. `checkAllowed` would be something like `:: UserId -> Handler r ()` and be called first thing inside the handler, not from outside of it.

The only arguable downside I can see is that the block is buried inside the handler logic rather than in being visible in the routing table.  But the big advantage is that the block will be effective for *all* situations in which the feature is enabled.

services/brig/src/Brig/Provider/API.hs
```diff
@@ -1001,6 +1041,11 @@ botDeleteSelf bid cid = do
 --------------------------------------------------------------------------------
 -- Utilities
 
+checkAllowed :: Maybe UserId -> (Handler r) ()
+checkAllowed mbUserId = do
+  enabled <- lift $ (==) Feature.TeamFeatureEnabled . Feature.tfwoStatus <$> RPC.getTeamFeatureStatusSndFactorPasswordChallenge mbUserId
+  when enabled $ throwStd accessDenied
```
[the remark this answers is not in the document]
battermann: I think this is correct, when the feature is enabled we want to block the bot API.

services/galley/src/Galley/API/Teams/Features.hs
```diff
@@ -730,6 +731,24 @@ getSndFactorPasswordChallengeInternal = \case
     getCfgDefault :: Sem r (Public.TeamFeatureStatus 'Public.WithLockStatus 'Public.TeamFeatureSndFactorPasswordChallenge)
     getCfgDefault = input <&> view (optSettings . setFeatureFlags . flagTeamFeatureSndFactorPasswordChallengeStatus . unDefaults)
 
+continueOnSndFactorPasswordChallengeDisabledOrAccessDenied ::
```
fisx, line 734:
```suggestion
-- | If second factor auth is enabled, make sure that end-points that don't support it, but should, are blocked completely.  (This is a workaround until we have 2FA for those end-points as well.)
--
-- This function exists to resolve a cyclic dependency.
guardSecondFactorDisabled ::
```

services/brig/src/Brig/Provider/API.hs
```diff
@@ -1001,6 +1041,11 @@ botDeleteSelf bid cid = do
 --------------------------------------------------------------------------------
 -- Utilities
 
+checkAllowed :: Maybe UserId -> (Handler r) ()
```
fisx, line 1044:
```suggestion
-- | If second factor auth is enabled, make sure that end-points that don't support it, but should, are blocked completely.  (This is a workaround until we have 2FA for those end-points as well.)
guardSecondFactorDisabled :: Maybe UserId -> Handler r ()
```

this is 3 suggestions, some or all of which you may want to ignore:
- comment
- rename
- remove parens in type signature

services/galley/src/Galley/API/Internal.hs
```diff
@@ -158,6 +158,20 @@ type InternalAPI =
                       :> ReqBody '[Servant.JSON] UpsertOne2OneConversationRequest
                       :> Post '[Servant.JSON] UpsertOne2OneConversationResponse
                   )
+           :<|> Named
+                  "feature-config-snd-factor-password-challenge"
+                  ( Summary "Get the server wide feature config for the 2nd factor password challenge feature."
```
fisx, line 163:
```suggestion
                  -- FUTUREWORK: Introduce `/i/feature-configs` and drop this one again.  The internal end-poins has the
                  -- same handler as the public one, plus optional user id in the query.  Maybe require `DoAuth` to disable
                  -- access control only on the internal end-point, not on the public one.  (This may also be a good oppportunity
                  -- to make `AllFeatureConfigs` more type-safe.)
                  ( Summary "Get feature config for the 2nd factor password challenge feature (for user/team; if n/a fall back to site config)."
```

fisx (approved):
LGTM!  (I don't need to re-approve the upcoming integration test if you and concourse agree that it's good.)

PR #2209, opened 2022-03-15
[WIP] Add Helm charts for coturn

sysvinit:
This PR adds Helm charts to allow deployment of our patched version of coturn as an alternative TURN server to restund.

(Work in progress, will update description later.)

## Checklist


charts/coturn/templates/_helpers.yaml
```diff
@@ -0,0 +1,45 @@
...
+{{- else }}
+{{- printf "%s-%s" .Release.Name $name | trunc 63 | trimSuffix "-" }}
+{{- end }}
+{{- end }}
+{{- end }}
+
+{{- define "coturn.selectorLabels" -}}
+app.kubernetes.io/name: {{ include "coturn.name" . }}
+app.kubernetes.io/instance: {{ .Release.Name }}
+{{- end }}
```
apurvaWire, line 45: Please add these labels which are important for wire-server components.
```
wireService: coturn
heritage: {{ .Release.Service }}
```

charts/coturn/templates/configmap-coturn-conf-template.yaml
```diff
@@ -0,0 +1,68 @@
...
+    denied-peer-ip=240.0.0.0-255.255.255.255
+    denied-peer-ip=::1
+    denied-peer-ip=64:ff9b::-64:ff9b::ffff:ffff
+    denied-peer-ip=::ffff:0.0.0.0-::ffff:255.255.255.255
+    denied-peer-ip=100::-100::ffff:ffff:ffff:ffff
+    denied-peer-ip=2001::-2001:1ff:ffff:ffff:ffff:ffff:ffff:ffff
+    denied-peer-ip=2002::-2002:ffff:ffff:ffff:ffff:ffff:ffff:ffff
+    # fc00::/7 is reserved for ipv6 ula, but fc00::/8 is not assigned at present.
+    denied-peer-ip=fc00::-fcff:ffff:ffff:ffff:ffff:ffff:ffff:ffff
+    denied-peer-ip=fe80::-febf:ffff:ffff:ffff:ffff:ffff:ffff:ffff
```
apurvaWire, line 66: 1. To avoid writing duplicate code 
2. If these config changes happen frequently

I would suggest something like this
```
{{- range .Values..deniedPeerIP }}
  denied-peer-ip= {{ . }}
 {{- end }}
```

And in the `values.yaml` file you just define a list or a map like
```
deniedPeerIP: ["192.88.99.0-192.88.99.255", "192.88.99.0-192.88.99.255" ........]
```

charts/coturn/values.yaml
```diff
@@ -0,0 +1,26 @@
...
+podSecurityContext:
+  fsGroup: 31338
+
+securityContext:
+  # Pick a high number that is unlikely to conflict with the host
+  # https://kubesec.io/basics/containers-securitycontext-runasuser/
+  runAsUser: 31338
+
+coturnTurnListenPort: 3478
+coturnMetricsListenPort: 9641
```
apurvaWire, line 26: I would highly recommend to add the minimum amount of resources in terms of CPU and MEM usage. 
```
resources:
  requests:
    cpu: "300m"
    memory: "500Mi"
  limits:
    cpu: "600m"
    memory: "1Gi"
```

apurvaWire (commented):
Thanks @sysvinit lgtm but I added some comments.

charts/coturn/templates/_helpers.yaml
sysvinit: A few things:
- This chart is based on the `restund` chart, so a lot of the boilerplate (including `_helpers.yaml` in particular) has just been copied verbatim.
- We don't ship calling services as part of the catch-all `wire-server` chart at present, as not all on-premise users have calling enabled in their environment.

This chart is intended to be relatively standalone, like restund is already, so I'm not sure this labelling is required.

charts/coturn/templates/configmap-coturn-conf-template.yaml
sysvinit: This is intended to be a list of known-invalid IP ranges, as these are all reserved as special-use by various RFC's. This list isn't expected to change in the future.

(There's definitely some scope for allowing operators to specify their own additional site-specific allowed and denied IP address lists, in a similar manner to how you suggested, however that's a feature I'd prefer to implement separately later on.)

charts/coturn/values.yaml
sysvinit: I don't really have any good ideas about sensible resource request sizes and limits to set for this service, unfortunately.

charts/coturn/templates/_helpers.yaml
apurvaWire: Again, was not aware about the background info. It was just a suggestion but we can totally skip it if it is not valid.

charts/coturn/Chart.yaml
```diff
@@ -0,0 +1,14 @@
+apiVersion: v2
```
jschaul, line 1: which kubernetes versions does this work with, and did you test the helm chart on? (we have various versions in use, customers might have various versions in use)

Makefile
```diff
@@ -8,7 +8,7 @@ DOCKER_TAG            ?= $(USER)
 # default helm chart version must be 0.0.42 for local development (because 42 is the answer to the universe and everything)
 HELM_SEMVER           ?= 0.0.42
 # The list of helm charts needed on internal kubernetes testing environments
-CHARTS_INTEGRATION    := wire-server databases-ephemeral fake-aws nginx-ingress-controller nginx-ingress-services wire-server-metrics fluent-bit kibana sftd restund
+CHARTS_INTEGRATION    := wire-server databases-ephemeral fake-aws nginx-ingress-controller nginx-ingress-services wire-server-metrics fluent-bit kibana sftd restund coturn
```
jschaul, line 11: I would add this also to CHARTS_RELEASE on line 17 so that the chart gets published on our helm chart repository, making it easier to install it elsewhere. Right now, it would only be available from a local checkout, and wouldn't get uploaded anywhere.

charts/coturn/templates/configmap-coturn-conf-template.yaml
```diff
@@ -0,0 +1,68 @@
...
+
+    ## logs
+    log-file=stdout
+    verbose
+
+    ## access control settings.
+    # note that this does not include rfc1918 ipv4 or ipv6 ula address
+    # space. these ranges may be valid client ip addresses in some private
+    # network environments. however, the other ranges listed here are intended
+    # for special use and should not be used.
```
jschaul, line 45: If this is a list of "well known" ranges, can you link to the respective RFCs or other documents where those IP ranges are listed in a comment?

charts/coturn/templates/configmap-coturn-conf-template.yaml
```diff
@@ -0,0 +1,68 @@
...
+    no-dtls
+    no-tcp-relay
+
+    ## don't turn on coturn's cli.
+    no-cli
+
+    ## authentication setup
+    # FUTUREWORK: enable support for multiple secrets?
+    zrest
+    static-auth-secret=__COTURN_SECRET__
```
jschaul, line 21: Hm, why these identifiers which you replace with `sed` on startup if you could use a variable taken from a kubernetes secret instead? Changes on a new deploy could override the configmap again with these placeholders leading to weird behaviour; as you need to ensure those sed commands always run to completion first.

Probably the existing approach here works; but it's hacky, especially since the secret is available, and you could just use `{{ .Values.secrets.zrestSecret | quote }}` here and entirely avoid the creation of the `charts/coturn/templates/secret.yaml` file

charts/coturn/templates/secret.yaml
```diff
@@ -0,0 +1,12 @@
...
+kind: Secret
+metadata:
+  name: coturn
+  labels:
+    chart: {{ .Chart.Name }}-{{ .Chart.Version | replace "+" "_" }}
+    release: "{{ .Release.Name }}"
+    heritage: "{{ .Release.Service }}"
+type: Opaque
+data:
+  zrest_secret.txt: {{ .Values.secrets.zrestSecret | b64enc | quote }}
```
jschaul, line 11: Are you sure you wish to base64encode the secret one more time? Is something decoding that secret somewhere?

charts/coturn/values.yaml
```diff
@@ -0,0 +1,26 @@
...
+# to `hostNetwork`. If this number is higher than the amount of nodes that can
+# be used for scheduling (Also see `nodeSelector`) pods will remain in a
+# pending state untill you add more capacity.
+replicaCount: 1
+
+image:
+  repository: quay.io/wire/coturn
+  pullPolicy: IfNotPresent
+  # overwrite the tag here, otherwise `appVersion` of the chart will be used
+  tag: ""
```
jschaul, line 11:
```suggestion
  # `appVersion` of the chart will be used as tag
```

Could you remove the tag here entirely, and always use the appVersion in Chart.yaml?

charts/coturn/templates/statefulset.yaml
```diff
@@ -0,0 +1,128 @@
...
+              # In the cloud, this setting is available to indicate the true IP address
+              addr=$(kubectl get node $HOSTNAME -ojsonpath='{.status.addresses[?(@.type=="ExternalIP")].address}')
+              # On on-prem we allow people to set  "wire.com/external-ip" to override this
+              if [ -z "$addr" ]; then
+                addr=$(kubectl get node $HOSTNAME -ojsonpath='{.metadata.annotations.wire\.com/external-ip}')
+              fi
+              echo -n "$addr" | tee /dev/stderr > /external-ip/ip
+      containers:
+        - name: {{ .Chart.Name }}
+          image: "{{ .Values.image.repository}}:{{ .Values.image.tag | default .Chart.AppVersion }}"
```
jschaul, line 65:
```suggestion
          image: "{{ .Values.image.repository}}:{{ .Chart.appVersion }}"
```

(also see review comment in values.yaml)

jschaul (approved):
Looks good enough as a first version, see some inline comments for suggestions.

charts/coturn/templates/configmap-coturn-conf-template.yaml
jschaul, line 21:
```suggestion
    static-auth-secret={{ .Values.secrets.zrestSecret | quote }}
```

charts/coturn/Chart.yaml
```diff
@@ -0,0 +1,14 @@
...
+name: coturn
+description: coturn - a STUN and TURN server
+type: application
+
+# This is the chart version. This version number should be incremented each time you make changes
+# to the chart and its templates, including the app version.
+# Versions are expected to follow Semantic Versioning (https://semver.org/)
+version: 0.0.1
```
jschaul, line 9:
```suggestion
version: 0.0.42
```

(to be in inline with the other charts's versioning scheme - same version for all charts set by CI)

charts/coturn/Chart.yaml
sysvinit: I've tested with Kubernetes version 1.19.7. I've got no idea what the range of compatible versions is, this was taken from the chart for restund.

Makefile
sysvinit: Good catch, I'd forgotten about that. I've pushed a commit which should add coturn to this list now.

charts/coturn/templates/configmap-coturn-conf-template.yaml
sysvinit: I've amended this comment with a link to the appropriate IANA registries which list all the special-purpose addresses.

charts/coturn/templates/configmap-coturn-conf-template.yaml
sysvinit: This is more cargo-culting from the restund chart. That chart stores the zrest authentication secret in a Kubernetes Secret, loads it into an environment variable in the container command, and then substitutes it into the config file using `envsubst`. The choice of sed here is because the coturn container image I've been working with doesn't contain `envsubst`, and I'm a bit reluctant to add that to the container image, as that would diverge from upstream.

I'd be happy to swap this for a different solution, if you have any suggestions.

charts/coturn/templates/secret.yaml
sysvinit: The restund and brig charts both perform similar base64 encoding. This does eventually get decoded somewhere, as in my test environment I've been using during development, the content of the `zrest_secret.txt` file live on the cluster matches the value in the configuration file I feed into Helm. Is it possible that Kubernetes expects file content for Secrets to be base64-encoded when specified like this?

charts/coturn/values.yaml
sysvinit: Is it possible to override `appVersion` at deployment time? My understanding is that exposing this handle for overriding the container image tag is useful for deploying different versions of images to test environments (e.g. testing a new version of coturn before burning the version number into the Helm chart which gets shipped to the public).

charts/coturn/templates/statefulset.yaml
sysvinit: See my response above.

charts/coturn/templates/configmap-coturn-conf-template.yaml
sysvinit: Oh, something else worth mentioning here is that I don't know how Kubernetes handles access controls or visibility for ConfigMaps and Secrets. Embedding the zrest auth secret into the ConfigMap when it's rendered might mean the secret is more visible within the cluster at runtime than if it were embedded in a Secret.

The brig chart stores various sensitive values (including the TURN zrest auth secret) in a separate Secret object, which attached to the pod separately from the ConfigMap with the main configuration file.

sysvinit: See my comments above regarding substituting values into the coturn config file.

charts/coturn/Chart.yaml
sysvinit: I've pushed a commit with this change just now, thanks for pointing it out.

PR #2213, opened 2022-03-16
SQSERVICES-1028-be-add-sso-scim-attributes-to-user-profile-for-team-admins

battermann:
https://wearezeta.atlassian.net/browse/SQSERVICES-1028

## Checklist

### DIFF
--- a/libs/wire-api/src/Wire/API/User/Search.hs
+++ b/libs/wire-api/src/Wire/API/User/Search.hs
@@ -167,6 +167,7 @@ instance ToSchema Contact where
 --------------------------------------------------------------------------------
 -- TeamContact
 
+-- | Related to `UserSSOId`, but more straight-forward because it does not take SCIM externalId into account.
 data Sso = Sso
   { ssoIssuer :: Text,
     ssoNameId :: Text
@@ -205,6 +206,14 @@ data TeamContact = TeamContact
   deriving stock (Eq, Show, Generic)
   deriving (Arbitrary) via (GenericUniform TeamContact)
 
+modelSso :: Doc.Model
+modelSso = Doc.defineModel "Sso" $ do
+  Doc.description "Single Sign-On"
+  Doc.property "issuer" Doc.string' $
+    Doc.description "Issuer"
+  Doc.property "nameid" Doc.string' $
+    Doc.description "Name ID"
+
 modelTeamContact :: Doc.Model
 modelTeamContact = Doc.defineModel "TeamContact" $ do
   Doc.description "Contact discovered through search"
@@ -225,6 +234,9 @@ modelTeamContact = Doc.defineModel "TeamContact" $ do
     Doc.optional
   Doc.property "scim_external_id" Doc.string' $ do
     Doc.description "SCIM external ID"
+    Doc.optional
+  Doc.property "sso" (Doc.ref modelSso) $ do
+    Doc.description "Single Sign-On"
     Doc.optional
 
 instance ToJSON TeamContact where
--- a/libs/wire-api/src/Wire/API/User/Search.hs
+++ b/libs/wire-api/src/Wire/API/User/Search.hs
@@ -23,6 +23,7 @@ module Wire.API.User.Search
     Contact (..),
     TeamContact (..),
     RoleFilter (..),
+    Sso (..),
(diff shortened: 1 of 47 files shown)


libs/wire-api/src/Wire/API/User/Search.hs
```diff
@@ -200,6 +223,9 @@ modelTeamContact = Doc.defineModel "TeamContact" $ do
   Doc.property "email" Doc.string' $ do
     Doc.description "Email address"
     Doc.optional
+  Doc.property "scim_external_id" Doc.string' $ do
```
fisx, line 226: what about `sso`?

libs/wire-api/src/Wire/API/User/Search.hs
```diff
@@ -166,6 +167,26 @@ instance ToSchema Contact where
 --------------------------------------------------------------------------------
 -- TeamContact
 
+data Sso = Sso
```
fisx, line 170:
```suggestion
-- | Related to `UserSSOId`, but more straight-forward because it does not take SCIM externalId into account.
data Sso = Sso
```

... or something to that end?

services/brig/src/Brig/User/Search/Index.hs
```diff
@@ -586,7 +588,38 @@ indexMapping =
                   mpIndex = True,
                   mpAnalyzer = Nothing,
                   mpFields = mempty
-                }
+                },
+            "scim_external_id"
+              .= MappingProperty
+                { mpType = MPKeyword,
+                  mpStore = False,
+                  mpIndex = False,
```
fisx, line 596: On second thought, can you double-check if this is something we need for implementing the current designs?  Would be annoying if we had to make another PR, another migration...  (sorry for the back & forth, but it's better not to index things we don't need indices for, ES takes a lot of resources).

services/brig/src/Brig/User/Search/Index.hs
```diff
@@ -827,9 +862,17 @@ reindexRowToIndexUser
...
-        Just $ (cs . toLazyByteString . serializeURIRef) uri
+        Just $ fromUri uri
       idpUrl (UserScimExternalId _) = Nothing
 
+      fromUri :: URI -> Text
+      fromUri = cs . toLazyByteString . serializeURIRef
+
+      sso :: UserSSOId -> Maybe Sso
+      sso userSsoId = do
+        (issuer, nameid) <- User.ssoIssuerAndNameId userSsoId
```
fisx, line 873: [optional] Why the detour through the tuple?  I would have named `Sso` `SsoOnly` (or something else more specific) and typed `User.ssoIssuerAndNameId :: UserSSOId -> Maybe SsoOnly`.

services/brig/src/Brig/User/Search/Index/Types.hs
```diff
@@ -60,7 +61,9 @@ data IndexUser = IndexUser
     _iuManagedBy :: Maybe ManagedBy,
     _iuCreatedAt :: Maybe UTCTime,
     _iuRole :: Maybe Role,
-    _iuSearchVisibilityInbound :: Maybe SearchVisibilityInbound
+    _iuSearchVisibilityInbound :: Maybe SearchVisibilityInbound,
+    _iuScimExternalId :: Maybe Text,
+    _iuSso :: Maybe Sso
```
fisx, line 66: This may be correct (I don't know ES), but are you sure if a field is not indexed, it should still appear here and not just in `UserDoc`?  If so, why?

fisx (approved):

services/brig/src/Brig/User/Search/Index.hs
battermann: Yes, but this would make a cyclic dependency, therefore the tuple.

services/brig/src/Brig/User/Search/Index/Types.hs
battermann: AFAIK `UserDoc` is a subset of the index.

services/brig/src/Brig/User/Search/Index.hs
battermann: Not sure, I understand the concern. The field is explicitly not indexed.

PR #2220, opened 2022-03-21
Sqservices 1028 be add sso SCIM attributes to user profile for team admins unvalidated email

battermann:
as a team admin
when I search for team contacts in team management
I want to see un-validated emails in the user profiles that are returned by the search

https://wearezeta.atlassian.net/browse/SQSERVICES-1028

## Checklist


libs/wire-api/src/Wire/API/User/Search.hs
```diff
@@ -200,7 +200,8 @@ data TeamContact = TeamContact
     teamContactSAMLIdp :: Maybe Text,
     teamContactRole :: Maybe Role,
     teamContactScimExternalId :: Maybe Text,
-    teamContactSso :: Maybe ContactSso
+    teamContactSso :: Maybe ContactSso,
+    teamContactEmailUnvalidated :: Maybe Email
```
fisx, line 204: Is this type only visible to admins?  If not, hiding the unvalidated email should be enforced:

```
data TeamContact f = TeamContact
  { ...
    teamContactEmailUnvalidated :: f Email
  }

...
userSearch :: ... -> m (TeamContact (Const ()))
...
adminSearch :: ... -> m (TeamContact Id)
...
```

services/brig/schema/src/V70_UserEmailUnvalidated.hs
```diff
@@ -0,0 +1,36 @@
...
+import Cassandra.Schema
+import Imports
+import Text.RawString.QQ
+
+migration :: Migration
+migration =
+  Migration 70 "Add email_unvalidated to user table" $
+    schema'
+      [r| ALTER TABLE user ADD (
+          email_unvalidated text
```
fisx, line 34: given that changing this column and changing other columns usually don't coincided (or do they?), wouldn't it be more straight-forward and more readable to have new table?

services/brig/src/Brig/API/User.hs
```diff
@@ -582,8 +582,10 @@ changeSelfEmail u email allowScim = do
   changeEmail u email allowScim !>> Error.changeEmailError >>= \case
     ChangeEmailIdempotent ->
       pure ChangeEmailResponseIdempotent
-    ChangeEmailNeedsActivation (usr, adata, en) -> do
-      lift $ sendOutEmail usr adata en
+    ChangeEmailNeedsActivation (usr, adata, en) -> lift $ do
+      sendOutEmail usr adata en
+      wrapClient $ Data.updateEmailUnvalidated u email
+      Intra.onUserEvent u Nothing (UserUpdated ((emptyUserUpdatedData u) {eupEmailUnvalidated = Just email}))
```
fisx, line 588: since we're not doing this before, I don't think we should do it now.  that's probably another reason to store this column in a separate table.  my (not very reliable) intuition also says that may be easier on cassandra resources.

libs/brig-types/src/Brig/Types/User/Event.hs
```diff
@@ -78,7 +78,8 @@ data UserUpdatedData = UserUpdatedData
     eupLocale :: !(Maybe Locale),
     eupManagedBy :: !(Maybe ManagedBy),
     eupSSOId :: !(Maybe UserSSOId),
-    eupSSOIdRemoved :: Bool
+    eupSSOIdRemoved :: Bool,
+    eupEmailUnvalidated :: !(Maybe Email)
```
fisx, line 82: Here and in general you need to be careful not to leak unvalidated emails.  It's fine to show them to admins, but nobody else must see them.

services/brig/src/Brig/API/User.hs
fisx: ah, I think this is also how reIndex is called?  sorry, i really should have more context.  you need to decide what you're doing, but what I think is required here is that (a) the change in unvalidated email does NOT trigger an event being sent to any other users (except, probably, team management, come to think of it); (b) re-index is called somewhere.

fisx (commented):
also good...  mostly.  :)

feel free to second-guess me if you can make @smatting agree with you!

services/brig/schema/src/V70_UserEmailUnvalidated.hs
battermann: there is one big advantage when having it in a single table, and that is we only have to query C* once when creating/updating the ES index.
if you insist, I'm also ok with having a new table.

smatting (changes requested):
Can you please bumpt `expectedMigrationVersion` to trigger a reindexing.
Also update `oldMapping` in the brig integration tests

smatting (approved):

PR #2222, opened 2022-03-21
nginx-ingress-services chart: allow for custom challenge solvers

smatting:
## Checklist

https://wearezeta.atlassian.net/browse/FS-513


charts/nginx-ingress-services/templates/_helpers.tpl
```diff
@@ -38,6 +38,16 @@ at the same time being used for distinct scenarios
 {{- join "-" $nameParts -}}
 {{- end -}}
 
+{{/*
+Generate the secrate name in a conistent way, since it's referred to in multiple places, while
```
flokli, line 42:
```suggestion
Generate the secret name in a conistent way, since it's referred to in multiple places, while
```

charts/nginx-ingress-services/templates/_helpers.tpl
```diff
@@ -38,6 +38,16 @@ at the same time being used for distinct scenarios
...
+at the same time being used for distinct scenarios
+*/}}
+{{- define "nginx-ingress-services.getAdditionalSolversSecret" -}}
```
flokli, line 45:
```suggestion
{{- define "nginx-ingress-services.getAdditionalSolversSecretName" -}}
```

This will only return a string

charts/nginx-ingress-services/templates/additional_solvers_secret.yaml
```diff
@@ -0,0 +1,13 @@
+{{- if .Values.secrets.certManager.additionalSolversSecret }}
+apiVersion: v1
+kind: Secret
+metadata:
+  name: {{ include "nginx-ingress-services.getAdditionalSolversSecret" . | quote }}
```
flokli, line 5:
```suggestion
  name: {{ include "nginx-ingress-services.getAdditionalSolversSecretName" . | quote }}
```

see above

charts/nginx-ingress-services/values.yaml
```diff
@@ -40,6 +40,9 @@ certManager:
   #  (1) required to be set in values.yaml
   #  (2) used by Letsencrypt to send expiration notices and alike
   certmasterEmail:
+  # can be used to use custom solvers, such as dns01
+  # you may use secrets.certManager.additionalSolvers to store secrets
+  additionalSolvers:
```
flokli, line 45: This doesn't explain what structure / kind of resource(s) we expect here.

Plural suggest its multiple. List or map, `Issuer`/`ClusterIssuer`?

charts/nginx-ingress-services/values.yaml
```diff
@@ -40,6 +40,10 @@ certManager:
   #  (1) required to be set in values.yaml
   #  (2) used by Letsencrypt to send expiration notices and alike
   certmasterEmail:
+  # List of acme.cert-manager.io/v1.ACMEChallengeSolver to be used instead of a
+  # http01 nginx based solver. You can use secrets.certManager.customSolvers
+  # to store secrets required by the solvers.
```
flokli, line 45: This is only one secret, and you're supposed to specify the contents below `.data` of it. We should probably clarify this in the comment here.

flokli:
I really think it's more intuitive to let the user describe their own issuer in its entirety, rather than messing with additional solvers and their priorities.

charts/nginx-ingress-services/values.yaml
smatting: Done

smatting:
> I really think it's more intuitive to let the user describe their own issuer in its entirety, rather than messing with additional solvers and their priorities.

This would be a bigger change, which could potentially break existing deployments.

PR #2223, opened 2022-03-22
Weeder

jschaul:
Try out https://hackage.haskell.org/package/weeder. Could be added to CI to remind us where dead code lives; e.g. during a concourse session.

## Checklist

 - [x] The **PR Title** explains the impact of the change.
 - [ ] The **PR description** provides context as to why the change should occur and what the code contributes to that effect. This could also be a link to a JIRA ticket or a Github issue, if there is one.
 - [ ] **changelog.d** contains the following bits of information ([details](https://github.com/wireapp/wire-server/blob/develop/docs/developer/changelog.md)):
   - [ ] A file with the changelog entry in one or more suitable sub-sections. The sub-sections are marked by directories inside `changelog.d`.

PR #2224, opened 2022-03-22
delete some dead code in galley

jschaul:
Noticed some example dead code from running weeder, see https://github.com/wireapp/wire-server/pull/2223

## Checklist


pcapriotti (approved):
Nice!

PR #2227, opened 2022-03-23
restund chart: Add dtls support

smatting:
## Checklist

https://wearezeta.atlassian.net/browse/FS-513


smatting:
Commit https://github.com/wireapp/wire-server/pull/2227/commits/fc56813fe0be9ef3228a4ea070ec5ae6c1b11aa5 is to meant to be reverted after charts have been published to wire-develop

PR #2233, opened 2022-03-24
[Polysemy] Remove MonadUnliftIO for Brig's AppT

mdimjasevic:
As part of effort to introduce Polysemy to Brig, this PR removes the `MonadUnliftIO` instance for the App monad in Brig. It introduces no change in the behavior of Brig.

## Checklist

### DIFF
--- a/services/brig/src/Brig/API/User.hs
+++ b/services/brig/src/Brig/API/User.hs
@@ -1151,7 +1151,14 @@ lookupPasswordResetCode emailOrPhone = do
       c <- wrapClient $ Data.lookupPasswordResetCode u
       return $ (k,) <$> c
 
-deleteUserNoVerify :: UserId -> (AppIO r) ()
+deleteUserNoVerify ::
+  ( MonadReader Env m,
+    MonadIO m,
+    MonadLogger m,
+    MonadThrow m
+  ) =>
+  UserId ->
+  m ()
 deleteUserNoVerify uid = do
   queue <- view internalEvents
   Queue.enqueue queue (Internal.DeleteUser uid)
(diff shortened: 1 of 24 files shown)


services/brig/src/Brig/Queue/Stomp.hs
```diff
@@ -132,6 +132,8 @@ enqueue b q m =
 -- configured properly, after failing on the same message several times the
 -- message will go into the Dead Letter Queue where it can be analyzed
 -- manually.
+--
+-- FUTUREWORK: This probably deserves a Polysemy action
```
mdimjasevic, line 136: @isovector , I changed a TODO note here to a futurework note. We won't be introducing effects in this PR, hence I think a futurework note is more appropriate (we tend to avoid merging TODO notes).

services/brig/src/Brig/API/User.hs
```diff
@@ -1166,43 +1209,74 @@ lookupProfile self other =
...
-  (AppIO r) [UserProfile]
+  m [UserProfile]
 lookupLocalProfiles requestingUser others = do
-  users <- wrapClient (Data.lookupUsers NoPendingInvitations others) >>= mapM userGC
+  users <- (Data.lookupUsers NoPendingInvitations others) >>= mapM userGC
   css <- case requestingUser of
-    Just localReqUser -> toMap <$> wrapClient (Data.lookupConnectionStatus (map userId users) [localReqUser])
-    Nothing -> mempty
+    Just localReqUser -> toMap <$> Data.lookupConnectionStatus (map userId users) [localReqUser]
+    Nothing -> pure $ toMap [] -- mempty
```
isovector, line 1279: I had to change this in order to get the type to derive properly, but we should change it back.

services/brig/src/Brig/API/User.hs
```diff
@@ -1236,10 +1310,28 @@ lookupLocalProfiles requestingUser others = do
...
+getLegalHoldStatus' ::
+  ( MonadLogger f,
+    MonadReader Env f,
+    MonadIO f,
+    MonadMask f,
+    MonadHttp f,
+    HasRequestId f
+  ) =>
+  User ->
+  f UserLegalHoldStatus
```
isovector, line 1334: `m` instead of `f`

isovector (approved):
LGTM

PR #2238, opened 2022-03-28
Fix docs: real-world domain used in example

smatting:


docs/reference/config-options.md
```diff
@@ -371,7 +371,7 @@ Configuring SFT load balancing can be done in two (mutually exclusive) settings:
 ```
 # [brig.yaml]
 sft:
-  sftBaseDomain: sft.wire.example.yourcloud.comk
+  sftBaseDomain: sft.wire.example.com
```
mdimjasevic, line 374: Isn't "example.com" also an actual domain? In some other context I remember we agreed not to use it.

smatting: Example.com may be used in examples https://www.plothost.com/kb/example-com-domain/#:~:text=Example.com%20(example.org,for%20Internet%20Assigned%20Numbers%20Authority.

PR #2244, opened 2022-03-30
Servantify brig properties endpoints

pcapriotti:
This PR servantifies brig's properties endpoints.

A couple of things worth noting:

 - One of the servantified endpoints (namely `GET /properties`) was used to test `/i/metrics`. This role has now been taken by `/onboarding/v3` (arbitrarily).
 - Since property values are validated in a way that requires knowledge of the settings, it cannot simply be done at the level of the Servant type classes. Therefore, I introduced a `RawPropertyValue` wrapper around a `ByteString`, which leaves it to the handler to do the actual parsing and validation.

As a side note, it is a bit strange that we are parsing property values at all, since we just convert them back to bytestrings immediately to save them in the database, and never look inside again. In principle, we could make them completely opaque bytestrings and avoid having to deal with all of these complications. However, it is hard to do without changing the API somehow, since these values are used inside other JSON objects.

## Checklist


services/brig/src/Brig/API/Public.hs
```diff
@@ -486,57 +432,56 @@ apiDocs =
...
-listPropertyKeysAndValuesH :: UserId ::: JSON -> (Handler r) Response
-listPropertyKeysAndValuesH (u ::: _) = do
-  keysAndVals <- lift $ wrapClient (API.lookupPropertyKeysAndValues u)
-  pure $ json (keysAndVals :: Public.PropertyKeysAndValues)
+listPropertyKeysAndValues :: UserId -> Handler r Public.PropertyKeysAndValues
+listPropertyKeysAndValues u = do
+  keysAndVals <- fmap Map.fromList . lift $ wrapClient (API.lookupPropertyKeysAndValues u)
+  fmap Public.PropertyKeysAndValues . hoistEither $
+    traverse
+      ( first (const (StdError internalServerError))
```
akshaymankar, line 481: Bit ugly to explicitly throw a 500, but I guess this is what was happening earlier when Cql parsing failed? Did you check if the logging is similar?

mdimjasevic: Hey, I'll take a look.

akshaymankar: > This can only happen if the database is corrupted, so 500 seems appropriate. What's your reservation about it?

5xx errors cause alarms, so I expect us to try to avoid them whenever possible. But in this case I guess it is fair to throw this, which looked like we're throwing 500 on an "expected" error. As I write this I realize my reservation wasn't very well founded, so please ignore me on this.

> Since it feels slightly dangerous to show the details of the parse error to the client in this case, I went with just explicitly logging the error when this happens, and still return the opaque 500 response. Does that sound reasonable? Do you have other suggestions?

Sounds good to me.

> The information is all there, but it seems the error gets logged twice. I suspect this is due to the polysemy refactoring, probably something around the Final IO effect ends up duplicating some logging action, or something along these lines.

Sounds a like a separate issue, I will mark this comment as resolved, please track it separately.

libs/wire-api/src/Wire/API/Routes/Public/Brig.hs
```diff
@@ -699,6 +700,76 @@ type ConnectionAPI =
...
+-- Properties API -----------------------------------------------------
+
+-- get "/properties-values" (continue listPropertyKeysAndValuesH) $
+--   zauthUserId
+--     .&. accept "application" "json"
+-- document "GET" "listPropertyKeysAndValues" $ do
+--   Doc.summary "List all properties with key and value."
+--   Doc.returns (Doc.ref Public.modelPropertyDictionary)
+--   Doc.response 200 "Object with properties as attributes." Doc.end
+
```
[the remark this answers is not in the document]
pcapriotti: That was a leftover from the refactoring. Fixed now.

libs/wire-api/src/Wire/API/Properties.hs
```diff
@@ -28,36 +29,79 @@ module Wire.API.Properties
...
 import Data.Text.Ascii
 import Imports
+import Servant
 import Wire.API.Arbitrary (Arbitrary)
 
-newtype PropertyKeysAndValues = PropertyKeysAndValues [(PropertyKey, PropertyValue)]
-  deriving stock (Eq, Show, Generic)
-  deriving newtype (Hashable)
+newtype PropertyKeysAndValues = PropertyKeysAndValues (Map PropertyKey PropertyValue)
+  deriving newtype (ToJSON)
```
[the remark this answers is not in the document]
akshaymankar: I explored the code a little and found that it is just tested as a JSON value, good enough I guess.

akshaymankar (approved):
:rocket:

libs/wire-api/src/Wire/API/Properties.hs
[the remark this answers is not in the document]
pcapriotti: *(I was sure I had replied to this comment, but I guess not. Maybe I closed the window without clicking the button :cry:)*

Do you mean if there are tests for the serialisation itself? There aren't, but it's just a newtype instance, so I'd say there's no need.

The endpoint returning those things is indeed tested: https://github.com/wireapp/wire-server/blob/develop/services/brig/test/integration/API/User/Property.hs#L100.

services/brig/src/Brig/API/Public.hs
mdimjasevic: > Here is the log output before:
> 
> ```
> [brig] E, request=N/A, cql-io: protocol error: parse error: response body reading: Failed reading: Failed to read property value: Error in $: Failed reading: not a valid json value at 'garbagebaz' Empty call stack
> [brig] D, code=500, label=server-error, request=N/A, "Server Error"
> ```
> 
> and here it is now:
> 
> ```
> [brig] E, request=N/A, Failed to parse a stored property value, raw_value="baz", parse_error=Error in $: Failed reading: not a valid json value at 'garbagebaz'
> [brig] D, code=500, label=internal-server-error, request=N/A, "Internal Server Error"
> [brig] E, request=N/A, Error {code = Status {statusCode = 500, statusMessage = "Internal Server Error"}, label = "internal-server-error", message = "Internal Server Error", errorData = Nothing}
> [brig] D, code=500, label=internal-server-error, request=N/A, "Internal Server Error"
> ```
> 
> The information is all there, but it seems the error gets logged twice. I suspect this is due to the polysemy refactoring, probably something around the `Final IO` effect ends up duplicating some logging action, or something along these lines.
> 
> @mdimjasevic Could you maybe take a look at some point?


What exactly is logged twice? There are four lines in that second snippet and I guess the lines correspond to two pairs of successive errors. The first error line (marked with E) is not the same as the second error line, so it doesn't seem strange to me that each is paired with its own debug line (marked with D). It could be I am missing something so please provide more context.

PR #2247, opened 2022-03-30
MLS commit processing

pcapriotti:
This PR implements processing of incoming MLS commit messages and associated `Add` proposals. The overall behaviour is as follows:

 - [x] the commit message is decoded
 - [x] the conversation ID is obtained from the group ID and the conversation is fetched from the database
 - [x] the epoch of the message is checked against the epoch of the conversation
 - [x] `Add` proposals are processed in turn
 - [x] for each `Add`, the corresponding key package is turned into a ref, then dereferenced by making a call to Brig
 - [x] all the added clients are grouped by user
 - [x] for each added user, it is checked whether all its currently known clients are being added by making another call to Brig
 - [x] all the users are added to the conversation (and events are sent out)
 - [x] the list of clients in the conversation is updated on the database
 - [x] the conversation epoch is incremented

What is current missing:

 - better commit message and proposal validation
 - forwarding of the processed commit message (TBD in #2253 or followup PRs)
 - processing of proposal refs
 - processing of any other proposal type other than add
 - any sort of federated behaviour (for both remote members and remote conversations)
 - bunching requests to brig for performance
 - locking the conversation data during the processing of a commit

**Note**: this PR contains a variation of the commit in #2260. Simply rebase it away before merging or rebasing for a conflict-free experience.

Tracked by https://wearezeta.atlassian.net/browse/FS-434.

## Checklist


fisx:
@pcapriotti a083cfae827f7876c9b1a4724d918ce28d05fa85 seems safe to me, and I don't think I need ancient knowledge about wire authentication black magic to decide that: it does not change behavior on the public API, and the change on the internal API seems fine.  Two further thoughts:

1. Maybe it would be good to somehow rename the new `Brig.API.Client.addClient` to `addClientUnsafe` (as it allows to set a bad policy), and keep `addClient` with the same interface and behavior as before?
2. Have you double-checked that we're not calling the internal end-point from some handler in galley in production, and rely on the authentication taking place in brig there?  Maybe you should mark this end-point as "only for testing" somehow?

I have not looked at the wider picture in this PR.  Let me know if I should.

pcapriotti:
Thanks @fisx, good points. I'll rename the function and check if the internal handler is used by something else. The rest of the PR is unrelated, this was only so that I could create multiple clients in tests without having to deal with the authentication stuff.

fisx:
> The rest of the PR is unrelated, this was only so that I could create multiple clients in tests without having to deal with the authentication stuff.

then you *could* have made this a separate PR...  :)  but maybe that's too much overhead.

pcapriotti:
> then you _could_ have made this a separate PR... :) but maybe that's too much overhead.

Sure, I can do that, it's not that much overhead.

services/galley/src/Galley/Cassandra/Conversation.hs
```diff
@@ -294,6 +299,7 @@ interpretConversationStoreToCassandra = interpret $ \case
   CreateConversationId -> Id <$> embed nextRandom
   CreateConversation loc nc -> embedClient $ createConversation loc nc
   GetConversation cid -> embedClient $ getConversation cid
+  GetConversationByGroupId gid -> embedClient $ getConversationByGroupId gid
```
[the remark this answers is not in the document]
pcapriotti: Thanks, I didn't see it :blush:. I'll remove it.

libs/wire-api/src/Wire/API/Error/Galley.hs
```diff
@@ -113,6 +142,18 @@ type instance MapError 'ConvAccessDenied = 'StaticError 403 "access-denied" "Con
 
 type instance MapError 'MLSNonEmptyMemberList = 'StaticError 400 "non-empty-member-list" "Attempting to add group members outside MLS"
 
+type instance MapError 'DuplicateMLSPublicKey = 'StaticError 400 "mls-duplicate-public-key" "MLS public key for the given signature scheme already exists"
+
+type instance MapError 'KeyPackageRefNotFound = 'StaticError 404 "mls-key-package-ref-not-found" "A referenced key packages could not be mapped to a known client"
```
mdimjasevic: A mistake in plurality: packages -> package.

libs/wire-api/src/Wire/API/Error/Galley.hs
```diff
@@ -113,6 +142,18 @@ type instance MapError 'ConvAccessDenied = 'StaticError 403 "access-denied" "Con
...
+
+type instance MapError 'UnsupportedMLSMessage = 'StaticError 422 "mls-unsupported-message" "Attempted to send a message with an unsupported combination of content type instance MapError ' and wire format"
```
mdimjasevic: The single quote within the message seems like something is off. Have you forgotten to finish the message?

libs/wire-api/src/Wire/API/Error/Galley.hs
```diff
@@ -113,6 +142,18 @@ type instance MapError 'ConvAccessDenied = 'StaticError 403 "access-denied" "Con
...
+
+type instance MapError 'ProposalNotFound = 'StaticError 404 "mls-proposal-not-found" "A proposal referenced in a commit message could not be found"
+
+type instance MapError 'UnsupportedProposal = 'StaticError 422 "mls-unsupported-proposal" "Unsupported proposal type instance MapError '"
```
mdimjasevic: ditto.

libs/wire-api/src/Wire/API/Error/Galley.hs
pcapriotti: This and the other mistakes below were fixed in later commits.

libs/wire-api/src/Wire/API/Error/Galley.hs
```diff
@@ -225,3 +264,21 @@ instance Member (Error DynError) r => ServerEffect (Error TeamFeatureError) r wh
...
+  { pfInner :: Wai.Error
+  }
+
+type instance ErrorEffect ProposalFailure = Error ProposalFailure
+
+-- Proposal failures are only reported generically in Swagger
+instance IsSwaggerError ProposalFailure where
+  addToSwagger = S.allOperations . S.description %~ Just . (<> desc) . fold
+    where
+      desc = "\n\n!!PROPOSAL FAILURE!!" -- TODO
```
mdimjasevic: Have you addressed the TODO in a later commit?

pcapriotti: Yes.

services/galley/src/Galley/Cassandra/Queries.hs
```diff
@@ -360,6 +360,11 @@ rmMemberClient c =
   let t = LT.fromStrict (client c)
    in QueryString $ "update clients set clients = clients - {'" <> t <> "'} where user = ?"
 
+-- MLS Clients --------------------------------------------------------------
+
+addMLSClients :: PrepQuery W (ConvId, UserId, C.Set ClientId) ()
+addMLSClients = "update member set mls_clients = mls_clients + ? where conv = ? and user = ?"
```
mdimjasevic, line 366: Doesn't the order of types in the tuple in the signature matter, i.e., shouldn't you use corresponding values in the same order? Here it looks like you're doing a + operation (which I guess is set union in Cassandra) between a Set of client IDs and a conversation ID.

mdimjasevic (commented):
This is a partial review. The rest is coming soon.

services/galley/src/Galley/API/MLS/Message.hs
```diff
@@ -18,16 +18,195 @@
...
+  ProposalOrRef ->
+  Sem r ProposalAction
+applyProposalRef (Ref _) = throwS @'ProposalNotFound
+applyProposalRef (Inline p) = applyProposal p
+
+instance Semigroup ProposalAction where
+  ProposalAction add1 <> ProposalAction add2 = ProposalAction (add1 <> add2)
+
+instance Monoid ProposalAction where
+  mempty = ProposalAction mempty
```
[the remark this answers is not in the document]
pcapriotti: Sure, now it's a `newtype derive`, but we'll add removals later, so might as well write the boilerplate now. For a record with many monoid fields, one can use `generic-monoid` or some such to derive instances via Generics, but for 2 or 3 fields, I'm not sure it's worth it.

services/galley/src/Galley/Cassandra/Queries.hs
pcapriotti: Also fixed later.

services/galley/src/Galley/API/MLS/Message.hs
```diff
@@ -18,16 +18,195 @@
...
-postMLSMessage _lusr _con smsg = case rmValue smsg of
-  SomeMessage _tag _msg -> undefined
+  Sem r [Event]
+postMLSMessage lusr con smsg = case rmValue smsg of
+  SomeMessage SMLSPlainText msg -> case msgTBS (msgPayload msg) of
+    CommitMessage c ->
+      processCommit lusr con (rmRaw smsg) (msgGroupId msg) c
+    ApplicationMessage _ -> throwS @'UnsupportedMLSMessage
+    _ -> pure mempty -- FUTUREWORK: handle other message types
+  _ -> pure mempty -- FUTUREWORK: handle encrypted messages
```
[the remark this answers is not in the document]
pcapriotti: Fixed in f21d52ae9376707fa76375955aa5090ce6581624.

services/galley/src/Galley/API/MLS/Message.hs
```diff
@@ -18,16 +18,195 @@
...
+  )
+
+data ProposalAction = ProposalAction
+  { paAdd :: [Qualified (UserId, ClientId)]
+  }
+
+processCommit ::
+  ( HasProposalEffects r,
+    Member (ErrorS 'ProposalNotFound) r,
+    Member (ErrorS 'ConvNotFound) r,
```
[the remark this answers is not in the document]
pcapriotti: It was actually redundant in `HasProposalEffects r`. Fixed in f21d52ae9376707fa76375955aa5090ce6581624.

services/galley/src/Galley/API/MLS/Message.hs
```diff
@@ -18,16 +18,195 @@
...
+    Member LegalHoldStore r,
+    Member MemberStore r,
+    Member TeamStore r
+  ) =>
+  Local UserId ->
+  ConnId ->
+  Local ConvId ->
+  ProposalAction ->
+  Sem r [Event]
+executeProposalAction lusr con lcnv action =
```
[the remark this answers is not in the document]
pcapriotti: Yes.

services/galley/test/integration/API/MLS.hs
```diff
@@ -197,17 +248,17 @@ aliceInvitesBob SetupOptions {..} = withSystemTempDirectory "mls" $ \tmp -> do
...
         then do
           conv <-
             responseJsonError =<< postConvQualified (qUnqualified alice) defNewMLSConv
               <!! const 201 === statusCode
           liftIO $ case cnvProtocol conv of
-            ProtocolMLS mlsData -> pure (cnvmlsGroupId mlsData)
+            ProtocolMLS mlsData -> pure (cnvmlsGroupId mlsData, cnvQualifiedId conv)
             p -> assertFailure $ "Expected MLS conversation, got protocol: " <> show (protocolTag p)
-        else pure "test_group"
+        else pure ("test_group", error "No conversation created")
```
mdimjasevic: How about using `assertFailure` instead of returning this pair?

mdimjasevic (commented):
Here is another partial review.

services/galley/test/integration/API/MLS.hs
pcapriotti: This is not a failure, though. We could return a `Maybe` on the second component, but whatever, this is just a test code, what does it matter if we crash with error vs get an assertion failure? It can only happen if there's a bug in the test itself.

mdimjasevic (approved):
This was a lot of work! It looks good.

services/galley/test/integration/API/MLS.hs
mdimjasevic: Agreed it's not a big deal. It can for sure stay this way.

services/galley/src/Galley/Cassandra/Queries.hs
```diff
@@ -193,6 +193,9 @@ updateTeamStatus = "update team set status = ? where team = ?"
 selectConv :: PrepQuery R (Identity ConvId) (ConvType, UserId, Maybe (C.Set Access), Maybe AccessRoleLegacy, Maybe (C.Set AccessRoleV2), Maybe Text, Maybe TeamId, Maybe Bool, Maybe Milliseconds, Maybe ReceiptMode, Maybe ProtocolTag, Maybe GroupId, Maybe Epoch)
 selectConv = "select type, creator, access, access_role, access_roles_v2, name, team, deleted, message_timer, receipt_mode, protocol, group_id, epoch from conversation where conv = ?"
 
+selectConvByGroupId :: PrepQuery R (Identity GroupId) (ConvId, Domain)
+selectConvByGroupId = "select conv_id, domain from group_id_conv_id where group_id = ?"
+
```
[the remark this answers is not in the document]
mdimjasevic: A fair point! I'm not objecting to having TODOs or similar notes in intermediate commits.

On the other hand, an alternative means I would have to read the whole PR first, and then go back in the commit history (which gets rewritten) to find where I had a question.

I suppose there's no ideal outcome in situations like this one.

pcapriotti:
Thanks @mdimjasevic! I'll merge now and improve those tests in a separate PR.

pcapriotti:
I addressed your comments in #2271.

PR #2248, opened 2022-03-30
Follow up on #2207 - Release Notes

battermann:


supersven (approved):
Looks good to me. :+1: 

I think it would be nice to mention in the commit message *why* something has been changed. And, in the changelog *what* has been changed.

Are you sure that no authentication is needed for this endpoint? (Just to double check that this is the expected outcome.)

PR #2253, opened 2022-03-31
[FS-207] MLS Message Propagation

mdimjasevic:
This implements local message validation and forwarding.

The PR is part of https://wearezeta.atlassian.net/browse/FS-207.

## Checklist

### DIFF
--- a/libs/wire-api/src/Wire/API/MLS/Message.hs
+++ b/libs/wire-api/src/Wire/API/MLS/Message.hs
@@ -25,6 +25,7 @@ module Wire.API.MLS.Message
     SWireFormatTag (..),
     SomeMessage (..),
     ContentType (..),
+    mpTag,
     MessagePayload (..),
     MessagePayloadTBS (..),
     Sender (..),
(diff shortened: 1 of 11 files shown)


libs/wire-api/src/Wire/API/MLS/Message.hs
```diff
@@ -160,6 +161,12 @@ data ContentType
 instance ParseMLS ContentType where
   parseMLS = parseMLSEnum @Word8 "content type"
 
+-- | Identify the content type of a message payload to be signed.
+mpTag :: MessagePayloadTBS -> ContentType
+mpTag (ApplicationMessage _) = ApplicationMessageTag
+mpTag (ProposalMessage _) = ProposalMessageTag
+mpTag (CommitMessage _) = CommitMessageTag
+
```
pcapriotti, line 169: Is this still used?

pcapriotti (approved):

libs/wire-api/src/Wire/API/MLS/Message.hs
mdimjasevic: In the refactored code it wasn't used. You just merged it so I guess we can remove it in some other PR.

pcapriotti: Yeah, I'm working on a follow up PR anyway, it's quicker to just merge as is rather than waiting for CI again and having to deal with stacked PRs in the meantime.

PR #2257, opened 2022-04-01
github workflows: use environment

smatting:
Pushing to cachix failed silently because the credentials were not provided. This PR makes use of a github environment that contain the auth token and signing key required.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,6 +6,7 @@ on:
 jobs:
   build-dev-env:
    name: Build dev env
+   environment: cachix # for secrets
    strategy:
      matrix:
        os:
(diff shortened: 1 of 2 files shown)


flokli:
From https://docs.github.com/en/actions/deployment/targeting-different-environments/using-environments-for-deployment#about-environments:

> Environments are used to describe a general deployment target like production, staging, or development. When a GitHub Actions workflow deploys to an environment, the environment is displayed on the main page of the repository. For more information about viewing deployments to environments, see "[Viewing deployment history](https://docs.github.com/en/developers/overview/viewing-deployment-history)."

As soon as an environment is set, the `subject` claim in the github actions token contains the environment name, and not the branch name.

I agree we should make sure the cachix secrets are available for builds hitting develop/master, but `environment` doesn't sound like the right way to do that.

PR #2258, opened 2022-04-01
Migrate wire-docs into wire-server

smatting:
This PR imports the [`wire-docs`](https://github.com/wireapp/wire-docs) repository into this repository. The PR was created like this:

1. Copy the contents of `wire-docs` into `docs/` from commit https://github.com/wireapp/wire-docs/commit/a360443306e80e2b9a865b9994e300efb6f15e0a with the exception of hidden files. This overwrites `docs/README.md`.
2. Remove `texlive` from nix environment, because it is a huge (>2GB) dependency. We should explore other option if we want to create PDFs from our docs, such as https://rst2pdf.org/
3. Disable `sphinx-multiversion` (for now). We're planning to explore versioning options.
4. Adjust the ci script
5. Restore the the `docs/README.md` overwritten in step 1 as `docs/README.original.md`

## Checklist


hack/bin/entrypoint.py
```diff
@@ -0,0 +1,88 @@
+#!/usr/bin/env python3
```
jschaul, line 1: Did you mean to commit these files? They look like they have nothing to do with wire-docs

hack/bin/wire
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env bash
+
+DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+TOP_LEVEL="$(cd "$DIR/../.." && pwd)"
+
+BIN="$TOP_LEVEL"/hack/bin
+
+export ENTRY_NAME="wire"
+export ENTRY_DIRS="$BIN"
+export ENTRY_IGNORE_REGEX="wire|entrypoint.py"
```
jschaul, line 10: same here, looks like an accidental file commit?

jschaul (changes requested):
Please keep the git history intact; see e.g. how this was done in https://github.com/zinfra/cailleach/pull/953 (discussion; e.g. ask @sysvinit ) or, alternatively, see https://github.com/zinfra/khan/pull/9#issuecomment-405694148

arthurwolf:
One issue if we are going to be merging wire-docs into wire-server, is there are a bunch of open issues and pull requests in the wire-docs repo. 
What happens to these?
Do we say we need to finish/close all of them before doing this?
Or do we just merge, and then any "last" changes to wire-docs will need to be manually applied to wire-server as an extra PR each time?

hack/bin/entrypoint.py
smatting: No, it seems I accidentaly staged them. Thanks for checking

smatting:
Since there aren't that many, I would duplicate the issues and PR in wire-server.
For coordination I would note the git commit in wire-docs were the snapshot was taken.

hack/bin/entrypoint.py
smatting: I recreated the PR from scratch

battermann (approved):

rohan-wire:
Hi, 
Two comments:
1) I think this was a step in the wrong direction in that this further discourages any reasonable, cohesive documentation for SFT, bots, STUN/TURN code that does not live in wire-server. We have one system. Our customers shouldn't care how we organize our code, they want documentation of the system. The system includes more than what is in wire-server, so IMO it should be elsewhere. 
2) we were using the PDF documentation for BSI. I will find another solution, but found this a bit of a rug pull. I didn't see this moved discussed or announced on Engineering , but perhaps I missed it. 
Thanks, -r

smatting:
> 1. I think this was a step in the wrong direction in that this further discourages any reasonable, cohesive documentation for SFT, bots, STUN/TURN code that does not live in wire-server.

Having the docs live here shouldn't discourage anyone from making changes compared to wire-docs: PRs that change docs don't trigger any CI checks, so the manual-only checks for merging apply.

> We have one system. Our customers shouldn't care how we organize our code, they want documentation of the system. The system includes more than what is in wire-server, so IMO it should be elsewhere.

To customers there isn't any visible change, docs are still at docs.wire.com. Where the docs live is indeed a bit arbitrary. There are at least two reasons to move it to wire docs:

- consolidate the doc.wire.com documentation with the existing docs in wire-server (e.g. docs/refernce)
- make it easier to keep the changelogs in wire docs up-to-date

Merging the docs here has been long consensus in the backend chapter at least:
https://wearezeta.atlassian.net/browse/BE-515
 
> 2. we were using the PDF documentation for BSI. I will find another solution, but found this a bit of a rug pull. I didn't see this moved discussed or announced on Engineering , but perhaps I missed it.

I'm sorry, I wasn't sure if there was a need for it: there was no build automation for it. I'll add it back. We could add build automation for the pdf and publish it, I created a PR for this: https://github.com/wireapp/wire-server/pull/2321

arthurwolf:
> Having the docs live here shouldn't discourage anyone from making changes compared to wire-docs: PRs that change docs don't trigger any CI checks, so the manual-only checks for merging apply (same as wire-docs).

In wire-docs we had a wire-docs specific template when creating a new PR/issue with a specific checklist etc. Now when creating a docs PR, we have a wire-server-specific template. Any way this can be fixed/improved?

battermann:
currently we create a changelog with the latest release notes when creating a new release PR
I think `/docs/src/release-notes.rst` should be updated within that PR as well, as long as we haven't stream-lined those, right?

smatting:
> In wire-docs we had a wire-docs specific template when creating a new PR/issue with a specific checklist etc. Now when creating a docs PR, we have a wire-server-specific template. Any way this can be fixed/improved?

Yes, it's the `.github/pull_request_template.md` file in this repo. We can add a section here.

PR #2260, opened 2022-04-04
Skip reauth when adding clients internally

pcapriotti:
This PR changes the internal Brig endpoint for adding new clients to make it skip the check for the need to re-authenticate. This is useful for integration tests that need to set up a number of clients.

Answering comments by @fisx on a similar commit in #2247:

> Maybe it would be good to somehow rename the new `Brig.API.Client.addClient` to `addClientUnsafe` (as it allows to set a bad policy), and keep `addClient` with the same interface and behavior as before?

I have done that, but called the unsafe function `addClientWithReAuthPolicy` instead, because "unsafe" usually means something else in Haskell.

> Have you double-checked that we're not calling the internal end-point from some handler in galley in production, and rely on the authentication taking place in brig there? Maybe you should mark this end-point as "only for testing" somehow?

It seems it is only called from galley in [`addLegalHoldClientToUser`](https://github.com/wireapp/wire-server/blob/develop/services/galley/src/Galley/Intra/Client.hs#L120). I'm now handling the 403 error from Brig and throwing an `AuthenticationError` in Galley. I have also added a query parameter as discussed, to make sure that only tests are taking advantage of the ability to skip the reauth check.

## Checklist

### DIFF
--- a/changelog.d/5-internal/brig-client-reauth
+++ b/changelog.d/5-internal/brig-client-reauth
@@ -1,0 +1,1 @@
+Add `skip_reauth` param to internal API for creating clients. This is intended to be used in test.
--- a/services/brig/src/Brig/API/Client.hs
+++ b/services/brig/src/Brig/API/Client.hs
@@ -19,6 +19,7 @@
 module Brig.API.Client
   ( -- * Clients
     addClient,
+    addClientWithReAuthPolicy,
     updateClient,
     rmClient,
     pubClient,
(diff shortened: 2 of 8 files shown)


fisx (commented):
as discussed on the phone. let me know if you need more input.

mdimjasevic (commented):
This looks good to me, but where can this be seen used in integration tests?

pcapriotti:
> This looks good to me, but where can this be seen used in integration tests?

https://github.com/wireapp/wire-server/pull/2247/commits/2c917417fc06d035194d17100218facce3e29543.

mdimjasevic:
> 
> [2c91741](https://github.com/wireapp/wire-server/commit/2c917417fc06d035194d17100218facce3e29543).

Thanks for the pointer! Can you please add the test change to this PR? I think that would make sense as the change is implemented by this PR.

pcapriotti:
> > [2c91741](https://github.com/wireapp/wire-server/commit/2c917417fc06d035194d17100218facce3e29543).
> 
> Thanks for the pointer! Can you please add the test change to this PR? I think that would make sense as the change is implemented by this PR.

That test doesn't exist in this branch.

mdimjasevic (approved):

PR #2264, opened 2022-04-04
[Polysemy] Replace the Base Monad in Brig's AppT with Polysemy Sem

mdimjasevic:
This change replaces the underlying monad `m` in Brig's `AppT`:

```haskell
newtype AppT r m a = AppT { unAppT :: ReaderT Env m a }
```

with the `Sem r` monad from Polysemy:

```haskell
newtype AppT r m a = AppT { unAppT :: Member (Final IO) r => ReaderT Env (Sem r) a }
```
This obsoletes the `m` type variable so it is dropped to get to the final version of `AppT`:

```haskell
newtype AppT r a = AppT { unAppT :: Member (Final IO) r => ReaderT Env (Sem r) a }
```

In terms of Brig behavior, this PR is a no-op.

## Checklist

schema documentation.


services/brig/src/Brig/App.hs
```diff
@@ -549,45 +586,46 @@ instance Cas.MonadClient HttpClientIO where
...
   liftIO . runResourceT $ transResourceT (runAppT e) ma
 
-forkAppIO ::
+forkAppT ::
   (MonadIO m, MonadUnliftIO m, MonadReader Env m) =>
   Maybe UserId ->
   m a ->
   m ()
-forkAppIO u ma = do
+forkAppT u ma = do
```
isovector, line 621: Do we still use this function?

mdimjasevic: Thanks! I changed it to `fork` in commit b28740f42.

PR #2266, opened 2022-04-05
Servantify Galley Teams [1/3]

stephen-smith:
## Checklist

JIRA: https://wearezeta.atlassian.net/browse/SQSERVICES-1080

No endpoints were supposed to be added or renamed, but they are now specified in servant type instead of wai-routes paths, so this should be a point of review.

   - [ ] If public end-points have been changed or added: does nginz need an upgrade?

### DIFF
--- a/services/galley/src/Galley/API/Internal.hs
+++ b/services/galley/src/Galley/API/Internal.hs
@@ -210,6 +210,9 @@ type ITeamsAPIBase =
            ( CanThrow 'NoBindingTeam :> CanThrow 'NotAOneMemberTeam :> CanThrow 'DeleteQueueFull
                :> DeleteAccepted '[Servant.JSON] NoContent
            )
+    :<|> "name" :> IGetTeamNameAPI
+
+type IGetTeamNameAPI = Named "get-team-name" (Get '[Servant.JSON] TeamName)
 
 type ITeamEffects = ErrorS 'TeamNotFound : GalleyEffects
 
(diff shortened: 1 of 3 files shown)


services/galley/src/Galley/API/Internal.hs
```diff
@@ -243,10 +272,22 @@ internalAPI =
...
+iTeamsAPI :: API ITeamsAPI GalleyEffects
+iTeamsAPI = mkAPI $ \tid -> hoistAPIHandler id (base tid)
+  where
+    base :: TeamId -> API ITeamsAPIBase ITeamEffects
+    base tid =
+      mkNamedAPI @"get-team" (Teams.getTeamInternalH tid)
+        <@> mkNamedAPI @"create-binding-team" (Teams.createBindingTeamH tid)
+        <@> mkNamedAPI @"delete-binding-team-with-one-member" (Teams.internalDeleteBindingTeamWithOneMemberH tid)
+        <@> hoistAPI @IGetTeamNameAPI id (mkNamedAPI @"get-team-name" $ Teams.getTeamNameInternalH tid)
+        <@> hoistAPI @IUpdateTeamStatusAPI id (mkNamedAPI @"update-team-status" $ Teams.updateTeamStatusH tid)
```
smatting, line 289: Using `hoistAPI` is probably not necessary here if the path routes (e.g. `"name"`) are nested inside `Named` combinator.

smatting (approved):

services/galley/src/Galley/API/Internal.hs
stephen-smith: True, and I have a commit that swizzles that around in my local development environment that will be included in the next round.

PR #2267, opened 2022-04-06
Suspend non-existing user should not create DB entry and return 404

battermann:


services/brig/src/Brig/Data/User.hs
```diff
@@ -341,7 +341,11 @@ deleteServiceUser pid sid bid = do
       \WHERE provider = ? AND service = ? AND team = ? AND user = ?"
 
 updateStatus :: MonadClient m => UserId -> AccountStatus -> m ()
-updateStatus u s = retry x5 $ write userStatusUpdate (params LocalQuorum (s, u))
+updateStatus u s =
+  whenM (userExists u) $ retry x5 $ write userStatusUpdate (params LocalQuorum (s, u))
+  where
+    userExists :: MonadClient m => UserId -> m Bool
+    userExists uid = isJust <$> retry x1 (query1 idSelect (params LocalQuorum (Identity uid)))
```
akshaymankar, line 348: Can you please make this throw a 404 when the user doesn't exist? Otherwise someone using the backoffice might think they did what was necessary when in reality it was a typo.

akshaymankar (changes requested):

akshaymankar (approved):
Looks good, can you please add a changelog?

PR #2273, opened 2022-04-11
[helm-charts] Allow filtering cassandra nodes by datacenter

akshaymankar:
## Checklist

### DIFF
--- a/charts/brig/templates/configmap.yaml
+++ b/charts/brig/templates/configmap.yaml
@@ -26,6 +26,9 @@ data:
         host: {{ .cassandra.host }}
         port: 9042
       keyspace: brig
+      {{- if hasKey .cassandra "filterNodesByDatacentre" }}
+      filterNodesByDatacentre: {{ .cassandra.filterNodesByDatacentre }}
+      {{- end }}
 
     elasticsearch:
       url: http://{{ .elasticsearch.host }}:{{ .elasticsearch.port }}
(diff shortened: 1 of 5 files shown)


jschaul (approved):

charts/brig/templates/configmap.yaml
```diff
@@ -26,6 +26,9 @@ data:
         host: {{ .cassandra.host }}
         port: 9042
       keyspace: brig
+      {{- if hasKey .cassandra "filterNodesByDatacentre" }}
```
jschaul, line 29: This syntax is wrong

```
  Error: template: wire-server/charts/spar/templates/deployment.yaml:27:31: executing "wire-server/charts/spar/templates/deployment.yaml" at <include (print .Template.BasePath "/configmap.yaml") .>: error calling include: template: wire-server/charts/spar/templates/configmap.yaml:23:20: executing "wire-server/charts/spar/templates/configmap.yaml" at <.cassandra>: wrong type for value; expected map[string]interface {}; got interface {}
  ```
  
  try with `helm template` to get the syntax right?

jschaul (changes requested):

charts/brig/templates/configmap.yaml
jschaul: I pushed a commit to fix the issue.

jschaul (approved):

charts/brig/templates/configmap.yaml
akshaymankar: Too much copy-pasta for me!

PR #2274, opened 2022-04-11
Document error handling and logging

pcapriotti:
This PR adds a bit of developer documentation regarding how errors are propagated from a service, through the middleware, and to the client.

It also removes some unnecessary duplication of error logs and simplifies the way errors are thrown in the main `Handler → Servant.Handler` natural transformation.

## Checklist

### DIFF
--- a/services/brig/src/Brig/API/Handler.hs
+++ b/services/brig/src/Brig/API/Handler.hs
@@ -93,7 +93,7 @@ toServantHandler env action = do
 
     handleWaiErrors logger reqId =
       \case
-        -- throw in IO so that the `catchError` middleware can turn this error
+        -- throw in IO so that the `catchErrors` middleware can turn this error
         -- into a response and log accordingly
         StdError werr -> liftIO $ throwIO werr
         RichError werr body headers -> do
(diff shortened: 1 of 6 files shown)


services/brig/src/Brig/API/Handler.hs
```diff
@@ -94,12 +93,13 @@ toServantHandler env action = do
 
     handleWaiErrors logger reqId =
       \case
-        StdError werr -> do
-          Server.logError' logger (Just reqId) werr
-          Servant.throwError $
-            Servant.ServerError (mkCode werr) (mkPhrase (WaiError.code werr)) (Aeson.encode werr) [(hContentType, renderHeader (Servant.contentType (Proxy @Servant.JSON)))]
+        -- throw in IO so that the `catchError` middleware can turn this error
```
mdimjasevic, line 96:
```suggestion
        -- throw in IO so that the `catchErrors` middleware can turn this error
```

mdimjasevic (approved):
Looks good! Just note minor remark about the name of a middleware function.

services/brig/src/Brig/API/Handler.hs
pcapriotti: Fixed.

mdimjasevic (approved):
It's interesting to see how code can be deleted and still have the same behavior!

PR #2275, opened 2022-04-11
Sqservices 1233 stern backoffice fix team deletion

battermann:
https://wearezeta.atlassian.net/browse/SQSERVICES-1233

Fix team deletion in backoffice/stern.
Added support for deleting teams with none or more than one member.

## Checklist


battermann:
here is a preview of the stern UI:
Screenshot from 2022-04-11 22-58-57

tools/stern/src/Stern/Intra.hs
```diff
@@ -324,6 +326,21 @@ deleteBindingTeam tid = do
           . expect2xx
       )
 
+-- | Caution! This will permanently delete all team members!
```
fisx, line 329:
```suggestion
-- | Caution! This may permanently delete all team members!
```

services/galley/src/Galley/API/Teams.hs
```diff
@@ -393,25 +392,31 @@ deleteTeam zusr zcon tid body = do
...
+  case tdTeam <$> mbTeamData of
+    Nothing -> throwS @'TeamNotFound
+    Just team | team ^. teamBinding /= Binding -> throwS @'NoBindingTeam
+    Just team -> do
+      mems <- E.getTeamMembersWithLimit tid (unsafeRange 2)
+      case mems ^. teamMembers of
+        [mem] -> queueTeamDeletion tid (mem ^. userId) Nothing
+        -- if the team has more than one member (and deletion is forced) or no members we use the team creator's userId for deletion events
+        xs | null xs || force -> queueTeamDeletion tid (team ^. teamCreator) Nothing
+        _ -> throwS @'NotAOneMemberTeam
```
fisx, line 419: It looks like you've rewritten some things that didn't change the behavior of the function.  This makes the commit hard to review and is a potential source of errors.  As a general rule, if you have to do rewrites, please commit them separately.

fisx: After having read it, this looks sound.  Maybe there wasn't anything obvious to rewrite without changing semantics separately?  Except maybe the `TeamNotFound` error?

tools/stern/src/Stern/API.hs
```diff
@@ -581,6 +590,10 @@ deleteTeam (givenTid ::: email) = do
     handleNoTeam = ifNothing (mkError status404 "no-binding-team" "No such binding team")
     wrongMemberCount = mkError status403 "wrong-member-count" "Only teams with 1 user can be deleted"
     bindingTeamMismatch = mkError status404 "binding-team-mismatch" "Binding team mismatch"
+deleteTeam (tid ::: True ::: _) = do
+  void $ Intra.getTeamData tid -- throws 404 if team does not exist
```
fisx, line 594:
```suggestion
```

(`deleteBindingTeamForce` does that, too, no?)

tools/stern/src/Stern/Intra.hs
```diff
@@ -324,6 +326,21 @@ deleteBindingTeam tid = do
...
+deleteBindingTeamForce tid = do
+  info $ msg "Deleting team with force flag"
+  g <- view galley
+  void . catchRpcErrors $
+    rpc'
+      "galley"
+      g
+      ( method DELETE
+          . paths ["/i/teams", toByteString' tid]
+          . queryItem' "force" Nothing
```
fisx, line 340: `Nothing`?  Why doesn't this default to `False`?  (We do want `True` for teams with >1 member, right?)

fisx (commented):
Looks good, just a few questions, hopefully easy to resolve.

services/galley/src/Galley/API/Teams.hs
```diff
@@ -693,13 +698,15 @@ internalDeleteBindingTeamWithOneMemberH ::
...
+       ErrorS 'TeamNotFound,
        Queue DeleteItem,
        TeamStore
      ]
     r =>
   TeamId ->
+  Bool ->
   Sem r NoContent
-internalDeleteBindingTeamWithOneMemberH = (NoContent <$) . internalDeleteBindingTeamWithOneMember
+internalDeleteBindingTeamWithOneMemberH tid force = internalDeleteBindingTeam tid force $> NoContent
```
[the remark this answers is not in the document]
battermann: why?

services/galley/src/Galley/Cassandra/Team.hs
```diff
@@ -282,7 +282,7 @@ teamMembersCollectedWithPagination lh tid = do
   where
     collectTeamMembersPaginated acc mems = do
       tMembers <- mapM (newTeamMember' lh tid) (result mems)
-      if (null $ result mems)
+      if hasMore mems
```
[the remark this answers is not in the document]
battermann: yes, this was a bug, which the integration test from this PR revealed
it didn't pop up before because teams always had 1 < n < 2000 members

tools/stern/src/Stern/Intra.hs
battermann: `queryItem'` has an optional parameter, and if empty it will just add the query parameter without a value like this:

```
/api/path?force
```

on the API side, existing parameter is interpreted as `true` and non-exisiting as `false`.
it's servants `QueryFlag`

tools/stern/src/Stern/API.hs
battermann: unfortunately errors from the internal API are not propagated to stern.
instead stern responds with 500
`deleteBindingTeamForce` responds with 404 but this will be a 500 in stern

services/galley/src/Galley/Cassandra/Team.hs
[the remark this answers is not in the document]
fisx: The we should probably add that to the release notes?

tools/stern/src/Stern/API.hs
fisx: Ok.  I would have preferred to fix that (and re-throw the error) rather than adding this awkward workaround, but it's not essential.

tools/stern/src/Stern/Intra.hs
fisx: Ok.  (I still think having an explicit value is easier to read, though.  I don't see any value in introducing complicated implicit semantics here.)

services/galley/src/Galley/API/Teams.hs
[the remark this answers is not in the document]
fisx: Sorry, I was trolling.  :)

The reason I was trolling is that I think your question should be: "why didn't I change the `<$` into a `$>` even though they are isomorophic?"  Violates the principle of separating refactorings from smallest change that accomplishes desired new behavior.

But I'm heavily bike shedding now, sorry.

fisx (approved):

services/galley/src/Galley/Cassandra/Team.hs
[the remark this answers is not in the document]
battermann: not sure, it was neither reported nor observed before, but it also doesn't hurt to add it

tools/stern/src/Stern/API.hs
battermann: me too, but I thought I stick to the convention for now

services/galley/src/Galley/Cassandra/Team.hs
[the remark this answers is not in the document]
battermann: actually I don't think this counts as a bug because the code was working under given preconditions when it was run

fisx (approved):

PR #2277, opened 2022-04-11
Servantify Galley Teams [2/2]

stephen-smith:
JIRA:

* https://wearezeta.atlassian.net/browse/SQSERVICES-1074 (search visibiliy)
* https://wearezeta.atlassian.net/browse/SQSERVICES-1078

## Checklist

### DIFF
--- a/changelog.d/5-internal/SQSERVICES-1078-servantify-galley_teams-2
+++ b/changelog.d/5-internal/SQSERVICES-1078-servantify-galley_teams-2
@@ -1,0 +1,1 @@
+Migrate 7 more galley/teams endpoints to Servant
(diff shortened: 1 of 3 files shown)


pcapriotti (approved):
Looks good!

One thing I'd like to improve is the use of `NoContent`, which is a bit of a hack in servant, and it contaminates the handler code with servant-specific nonsense. We have created a `MultiVerb` combinator to solve this and other limitations of servant, and here it works pretty well (in its `MultiVerb1` incarnation, which is for a single response).

I've taken the liberty of adding a commit to this PR which removes the uses of `NoContent`. As you can see, a lot of meaningless glue code around the handler just goes away.

fisx (approved):

stephen-smith:
> With the removal of wai-route, we probably want to remove all the wrappers.

Some I've been able to drop and call the underlying function directly.  Some have status code manipulation or `Maybe` handling, and I've been leaving that in the ...H function.  I haven't looked at the `NoContent` elimination, but that should let me remove a few more ...H functions that were just converting `()` to `NoContent`.

PR #2278, opened 2022-04-12
Proteus actions on MLS conversations should fail

pcapriotti:
Make sure that clients cannot send a proteus message to an MLS conversation, or add and remove users directly using the `/conversations/.../members` endpoints. Add corresponding tests.

Tracked by https://wearezeta.atlassian.net/browse/FS-533.

## Checklist

### DIFF
--- a/services/galley/src/Galley/API/Util.hs
+++ b/services/galley/src/Galley/API/Util.hs
@@ -498,6 +498,15 @@ getConversationAndCheckMembership uid lcnv = do
       lcnv
   pure conv
 
+getConversationWithError ::
+  ( Member ConversationStore r,
+    Member (ErrorS 'ConvNotFound) r
+  ) =>
+  Local ConvId ->
+  Sem r Data.Conversation
+getConversationWithError lcnv =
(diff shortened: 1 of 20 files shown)


services/galley/src/Galley/API/Util.hs
```diff
@@ -498,17 +498,23 @@ getConversationAndCheckMembership uid lcnv = do
...
   forall e uid mem r.
   (Members '[ConversationStore, ErrorS 'ConvNotFound, ErrorS e] r, IsConvMemberId uid mem) =>
   uid ->
   Local ConvId ->
   Sem r (Data.Conversation, mem)
 getConversationAndMemberWithError usr lcnv = do
-  c <- getConversation (tUnqualified lcnv) >>= noteS @'ConvNotFound
-  when (DataTypes.isConvDeleted c) $ do
-    deleteConversation (tUnqualified lcnv)
-    throwS @'ConvNotFound
```
mdimjasevic, line 511: Why is this check no more?

mdimjasevic (approved):
Nice!

PR #2284, opened 2022-04-12
Servantify Galley Misc. 3

stephen-smith:
https://wearezeta.atlassian.net/browse/SQSERVICES-1074

## Checklist

### DIFF
--- a/services/galley/src/Galley/API/LegalHold.hs
+++ b/services/galley/src/Galley/API/LegalHold.hs
@@ -26,9 +26,7 @@ module Galley.API.LegalHold
     approveDeviceH,
     disableForUserH,
     isLegalHoldEnabledForTeam,
-    setTeamLegalholdWhitelistedH,
     unsetTeamLegalholdWhitelistedH,
-    getTeamLegalholdWhitelistedH,
   )
 where
 
@@ -66,8 +64,7 @@ import qualified Galley.External.LegalHoldService as LHService
 import Galley.Types (LocalMember, lmConvRoleName, lmId)
 import Galley.Types.Teams as Team
 import Imports
-import Network.HTTP.Types (status200, status404)
-import Network.HTTP.Types.Status (status201, status204)
+import Network.HTTP.Types.Status (status200, status201, status204)
 import Network.Wai
 import Network.Wai.Predicate hiding (Error, or, result, setStatus, _3)
 import Network.Wai.Utilities as Wai hiding (Error)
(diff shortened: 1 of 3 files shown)


pcapriotti:
> Currently, this is on top of another PR which will be squash merged after CI completes. At that point this will be rebased, but going ahead and opening a draft for review -- though it may be best to look at just the final commit than the whole PR, for now.

You can change the base branch for a PR on github. I have now set the base for this PR to the branch of the previous one to make it easier to review.

services/galley/src/Galley/API/Error.hs
```diff
@@ -154,7 +154,8 @@ legalHoldServiceBadResponse :: Wai.Error
 legalHoldServiceBadResponse = Wai.mkError status400 "legalhold-status-bad" "legal hold service: invalid response"
 
 legalHoldNotEnabled :: Wai.Error
-legalHoldNotEnabled = Wai.mkError status403 "legalhold-not-enabled" "legal hold is not enabled for this team"
+legalHoldNotEnabled =
+  Wai.mkError status404 "legalhold-not-enabled" "legal hold is not enabled for this team"
```
pcapriotti, line 158: I don't think we should be changing this status code, because this error is used somewhere else as well.

services/galley/src/Galley/API/LegalHold.hs
```diff
@@ -840,32 +838,22 @@ blockNonConsentingConnections uid = do
...
+  (Member LegalHoldStore r, Member (Error LegalHoldError) r) =>
+  TeamId ->
+  Sem r ()
 getTeamLegalholdWhitelistedH tid = do
   lhEnabled <- LegalHoldData.isTeamLegalholdWhitelisted tid
-  pure $
-    if lhEnabled
-      then setStatus status200 empty
-      else setStatus status404 empty
+  unless lhEnabled $ throw LegalHoldNotEnabled -- NB: Should switch to ErrorS / MapError
```
pcapriotti, line 856: Instead of throwing an exception here, we can simply use `MultiVerb` with two responses.

services/galley/src/Galley/API/Error.hs
stephen-smith: I checked and the Galley version wasn't used anywhere else, though Brig has an identical definition that is used.

services/galley/src/Galley/API/LegalHold.hs
stephen-smith: That eliminates another ...H function. :)

services/galley/src/Galley/API/Error.hs
stephen-smith: But, with the switch to MultiVerb, we put the 404 info elsewhere, so this change disappears.

pcapriotti (approved):
Looks good!

PR #2286, opened 2022-04-13
SQSERVICES-1014-backend-servantify-galley-team-feature-flags-api

battermann:
https://wearezeta.atlassian.net/browse/SQSERVICES-1014

## Checklist

### DIFF
--- a/libs/wire-api/src/Wire/API/Routes/Public/Galley.hs
+++ b/libs/wire-api/src/Wire/API/Routes/Public/Galley.hs
@@ -54,6 +54,7 @@ import Wire.API.Team
 import Wire.API.Team.Conversation
 import Wire.API.Team.Feature
 import Wire.API.Team.Permission (Perm (..))
+import Wire.API.Team.SearchVisibility (TeamSearchVisibilityView)
 
 instance AsHeaders '[ConvId] Conversation Conversation where
   toHeaders c = (I (qUnqualified (cnvQualifiedId c)) :* Nil, c)
@@ -1050,6 +1051,8 @@ type FeatureAPI =
     :<|> FeatureStatusPut '() 'TeamFeatureSearchVisibility
     :<|> FeatureStatusDeprecatedGet 'WithoutLockStatus 'TeamFeatureSearchVisibility
     :<|> FeatureStatusDeprecatedPut 'TeamFeatureSearchVisibility
+    :<|> SearchVisibilityGet
+    :<|> SearchVisibilitySet
     :<|> FeatureStatusGet 'TeamFeatureValidateSAMLEmails
     :<|> FeatureStatusDeprecatedGet 'WithoutLockStatus 'TeamFeatureValidateSAMLEmails
     :<|> FeatureStatusGet 'TeamFeatureDigitalSignatures
@@ -1176,6 +1179,34 @@ type AllFeatureConfigsGet =
         :> CanThrow 'TeamNotFound
         :> "feature-configs"
         :> Get '[Servant.JSON] AllFeatureConfigs
+    )
+
+type SearchVisibilityGet =
+  Named
+    "get-search-visibility"
+    ( Summary "Shows the value for search visibility"
+        :> CanThrow 'NotATeamMember
+        :> CanThrow OperationDenied
+        :> ZUser
+        :> "teams"
+        :> Capture "tid" TeamId
+        :> "search-visibility"
+        :> Get '[Servant.JSON] TeamSearchVisibilityView
+    )
+
+type SearchVisibilitySet =
+  Named
+    "set-search-visibility"
+    ( Summary "Sets the search visibility for the whole team"
+        :> CanThrow 'NotATeamMember
+        :> CanThrow OperationDenied
+        :> CanThrow 'TeamSearchVisibilityNotEnabled
+        :> ZUser
+        :> "teams"
+        :> Capture "tid" TeamId
+        :> "search-visibility"
+        :> ReqBody '[JSON] TeamSearchVisibilityView
+        :> MultiVerb 'PUT '[JSON] '[RespondEmpty 204 "Search visibility set"] ()
     )
 
 type MLSMessagingAPI =
--- a/libs/wire-api/src/Wire/API/Routes/Public/Galley.hs
+++ b/libs/wire-api/src/Wire/API/Routes/Public/Galley.hs
@@ -1192,11 +1192,11 @@ type AllFeaturesGet =
         :> CanThrow 'NotATeamMember
         :> CanThrow OperationDenied
         :> CanThrow 'TeamNotFound
-        :> ZUser
+        :> ZLocalUser
         :> "teams"
         :> Capture "tid" TeamId
         :> "features"
-        :> Get '[Servant.JSON] AllFeatureConfigs
+        :> Get '[JSON] AllFeatureConfigs
     )
 
 type SearchVisibilityGet =
@@ -1205,11 +1205,11 @@ type SearchVisibilityGet =
     ( Summary "Shows the value for search visibility"
         :> CanThrow 'NotATeamMember
         :> CanThrow OperationDenied
-        :> ZUser
+        :> ZLocalUser
         :> "teams"
         :> Capture "tid" TeamId
         :> "search-visibility"
-        :> Get '[Servant.JSON] TeamSearchVisibilityView
+        :> Get '[JSON] TeamSearchVisibilityView
     )
 
 type SearchVisibilitySet =
@@ -1219,7 +1219,7 @@ type SearchVisibilitySet =
         :> CanThrow 'NotATeamMember
         :> CanThrow OperationDenied
         :> CanThrow 'TeamSearchVisibilityNotEnabled
-        :> ZUser
+        :> ZLocalUser
         :> "teams"
         :> Capture "tid" TeamId
         :> "search-visibility"
--- a/libs/wire-api/src/Wire/API/Routes/Public/Galley.hs
+++ b/libs/wire-api/src/Wire/API/Routes/Public/Galley.hs
@@ -22,6 +22,7 @@ module Wire.API.Routes.Public.Galley where
 
 import qualified Data.Code as Code
 import Data.CommaSeparatedList
+import Data.Domain (Domain)
 import Data.Id (ConvId, TeamId, UserId)
 import Data.Qualified (Qualified (..))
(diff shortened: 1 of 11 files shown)


libs/wire-api/src/Wire/API/Routes/Public/Galley.hs
```diff
@@ -1178,6 +1185,48 @@ type AllFeatureConfigsGet =
...
     )
 
+type AllFeaturesGet =
+  Named
+    "get-all-features"
+    ( Summary "Shows the configuration status of every team feature"
+        :> CanThrow 'NotATeamMember
+        :> CanThrow OperationDenied
+        :> CanThrow 'TeamNotFound
+        :> ZUser
```
mdimjasevic, line 1195: I'd suggest to make this `ZLocalUser` instead of `ZUser`. That way we provide more information to the endpoint and have a local value in scope should it be needed.

libs/wire-api/src/Wire/API/Routes/Public/Galley.hs
```diff
@@ -1178,6 +1185,48 @@ type AllFeatureConfigsGet =
...
+        :> Get '[Servant.JSON] AllFeatureConfigs
+    )
+
+type SearchVisibilityGet =
+  Named
+    "get-search-visibility"
+    ( Summary "Shows the value for search visibility"
+        :> CanThrow 'NotATeamMember
+        :> CanThrow OperationDenied
+        :> ZUser
```
mdimjasevic, line 1208: ditto.

libs/wire-api/src/Wire/API/Routes/Public/Galley.hs
```diff
@@ -1178,6 +1185,48 @@ type AllFeatureConfigsGet =
...
+    )
+
+type SearchVisibilitySet =
+  Named
+    "set-search-visibility"
+    ( Summary "Sets the search visibility for the whole team"
+        :> CanThrow 'NotATeamMember
+        :> CanThrow OperationDenied
+        :> CanThrow 'TeamSearchVisibilityNotEnabled
+        :> ZUser
```
mdimjasevic, line 1222: ditto.

libs/wire-api/src/Wire/API/Routes/Public/Galley.hs
```diff
@@ -1178,6 +1185,48 @@ type AllFeatureConfigsGet =
...
+    "get-all-features"
+    ( Summary "Shows the configuration status of every team feature"
+        :> CanThrow 'NotATeamMember
+        :> CanThrow OperationDenied
+        :> CanThrow 'TeamNotFound
+        :> ZUser
+        :> "teams"
+        :> Capture "tid" TeamId
+        :> "features"
+        :> Get '[Servant.JSON] AllFeatureConfigs
```
mdimjasevic, line 1199: You can also use just `JSON` instead of `Servant.JSON`.

libs/wire-api/src/Wire/API/Routes/Public/Galley.hs
```diff
@@ -1178,6 +1185,48 @@ type AllFeatureConfigsGet =
...
+  Named
+    "get-search-visibility"
+    ( Summary "Shows the value for search visibility"
+        :> CanThrow 'NotATeamMember
+        :> CanThrow OperationDenied
+        :> ZUser
+        :> "teams"
+        :> Capture "tid" TeamId
+        :> "search-visibility"
+        :> Get '[Servant.JSON] TeamSearchVisibilityView
```
mdimjasevic, line 1212: You can also use just `JSON` instead of `Servant.JSON`.

libs/wire-api/src/Wire/API/Routes/Public/Galley.hs
```diff
@@ -1208,6 +1257,17 @@ type MLSMessagingAPI =
...
 type MLSAPI = LiftNamed (ZLocalUser :> "mls" :> MLSMessagingAPI)
 
+type CustomBackendAPI =
+  Named
+    "get-custom-backend-by-domain"
+    ( Summary "Shows information about custom backends related to a given email domain"
+        :> CanThrow 'CustomBackendNotFound
+        :> "custom-backend"
+        :> "by-domain"
+        :> Capture "domain" Domain
```
mdimjasevic, line 1267: This seems like a non-trivial piece of information that was dropped:
```suggestion
        :> Capture' '[Description "URL-encoded email domain"] "domain" Domain
```

services/galley/src/Galley/API/Teams/Features.hs
```diff
@@ -297,9 +277,9 @@ getAllFeatures ::
     r =>
   UserId ->
   TeamId ->
-  Sem r Aeson.Value
+  Sem r Aeson.Object
 getAllFeatures uid tid = do
-  Aeson.object
+  KeyMap.fromList
```
mdimjasevic, line 282: Why is there a need for this change?

libs/wire-api/src/Wire/API/Team/SearchVisibility.hs
```diff
@@ -63,38 +66,40 @@ data TeamSearchVisibility
   | SearchVisibilityNoNameOutsideTeam
   deriving stock (Eq, Show, Ord, Enum, Bounded, Generic)
   deriving (Arbitrary) via (GenericUniform TeamSearchVisibility)
+  deriving (ToJSON, FromJSON, S.ToSchema) via (Schema TeamSearchVisibility)
 
 typeSearchVisibility :: Doc.DataType
 typeSearchVisibility =
   Doc.string . Doc.enum $
-    cs . encode <$> [(minBound :: TeamSearchVisibility) ..]
+    cs . A.encode <$> [(minBound :: TeamSearchVisibility) ..]
```
mdimjasevic, line 74: Is this still needed?

services/galley/test/integration/API/CustomBackend.hs
```diff
@@ -52,9 +51,11 @@ getByDomainInvalidDomain :: TestM ()
 getByDomainInvalidDomain = do
   galley <- view tsGalley
   -- contains invalid character '+'
+  -- this used to respond with '400 bad request'
+  -- but after servantification it returns '404 not found'
+  -- because the domain parameter is invalid and therefore the endpoint is invalid, too
   get (galley . path "/custom-backend/by-domain/invalid%2Bdomain") !!! do
-    const 400 === statusCode
-    const ("client-error" :: ByteString) =~= (cs . fold . responseBody)
+    const 404 === statusCode
```
mdimjasevic, line 58: I can't see what in this PR changes the status code. From what I can see, when an endpoint throws `'CustomBackendNotFound`, this has always mapped to a 404 as it can be seen in `libs/wire-api/src/Wire/API/Error/Galley.hs`:

```haskell
type instance MapError 'CustomBackendNotFound = 'StaticError 404 "custom-backend-not-found" "Custom backend not found"
```

Can you point me to the place where your PR changes the status code?

mdimjasevic (changes requested):
Overall this is good. I'd just ask you to include that domain's description in an endpoint's path as explained inlined. Others are minor comments.

services/galley/src/Galley/API/Teams/Features.hs
battermann: because the old endpoint returned just some JSON and it didn't matter if this was an object or a value, but now we return `AllFeatureConfigs` which is defined like this:

```
newtype AllFeatureConfigs = AllFeatureConfigs {_allFeatureConfigs :: Aeson.Object}
```

services/galley/test/integration/API/CustomBackend.hs
battermann: that is a good question, my assumption was that in the wai-routes definition [here](https://github.com/wireapp/wire-server/pull/2286/files#diff-f577bd929063ed2e5a0fc6ded4e6ecce3948249ef3812dfddcf77c3db91b6398L394):

```
  get "/custom-backend/by-domain/:domain" (continueE CustomBackend.getCustomBackendByDomainH) $
    capture "domain"
```
`domain` is treated as user input and therefore the response is 400
whereas servant treats it as a path and if it is not parsable the response is 404

services/galley/src/Galley/API/Public/Servant.hs
```diff
@@ -141,8 +142,8 @@ servantSitemap =
               setTeamSearchVisibilityAvailableInternal
               . DoAuth
           )
-        <@> mkNamedAPI @"get-search-visibility" getSearchVisibility
-        <@> mkNamedAPI @"set-search-visibility" setSearchVisibility
+        <@> mkNamedAPI @"get-search-visibility" (getSearchVisibility . tUnqualified)
```
mdimjasevic, line 145: Please change the type of `getSearchVisibility` to take a `Local UserId` instead of a `UserId` now that you used `ZLocalUser` instead of `ZUser`.

services/galley/src/Galley/API/Public/Servant.hs
```diff
@@ -224,7 +225,7 @@ servantSitemap =
               . DoAuth
           )
         <@> mkNamedAPI @"get-all-feature-configs" getAllFeatureConfigs
-        <@> mkNamedAPI @"get-all-features" (\uid tid -> AllFeatureConfigs <$> getAllFeatures uid tid)
+        <@> mkNamedAPI @"get-all-features" (\luid tid -> AllFeatureConfigs <$> getAllFeatures (tUnqualified luid) tid)
```
mdimjasevic, line 228: ditto.

mdimjasevic (approved):
This is good!

The reason I proposed to use `ZLocalUser` instead of `ZUser` is so that a `Local` value is available in endpoint handlers. By using `tUnqualified` in `Galley.API.Public.Servant` you threw that additional information away and then there's no point in using `ZLocalUser` in specifying the API.

`ZLocalUser` is in my view a preferred way of obtaining the user ID of the request maker, though `ZUser` would work just fine as well. The benefit is, as stated earlier, that `ZLocalUser` provides you with a `Local` value, which can be useful in handlers (though the specific handlers in this PR maybe don't benefit right now, but some future refactorings might).

services/galley/test/integration/API/CustomBackend.hs
mdimjasevic: Oh, OK, so it's how Servant handles the issue vs. how wai-routes handles it.

mdimjasevic (approved):

PR #2287, opened 2022-04-13
Add `skip_own` param to MLS kp claiming endpoint

pcapriotti:
Clients need to claim key packages for other clients of the same user, but never their own. This adds a `skip_own` query parameter to the key package claiming endpoint, which can be used to prevent wasting a key package for the requesting client.

Tracked by https://wearezeta.atlassian.net/browse/FS-527.

## Checklist

### DIFF
--- a/services/brig/src/Brig/API/MLS/KeyPackages.hs
+++ b/services/brig/src/Brig/API/MLS/KeyPackages.hs
@@ -34,7 +34,6 @@ import Control.Monad.Trans.Maybe
 import Data.Id
 import Data.Qualified
 import qualified Data.Set as Set
-import Debug.Trace
 import Imports
 import Wire.API.Federation.Error
 import Wire.API.MLS.Credential
@@ -60,7 +59,6 @@ claimLocalKeyPackages :: Local UserId -> Maybe ClientId -> Local UserId -> Handl
 claimLocalKeyPackages lusr skipOwn target = do
   -- skip own client when the target is the requesting user itself
   let own = guard (lusr == target) *> skipOwn
-  traceM $ "own: " <> show own
   clients <- map clientId <$> wrapClientE (Data.lookupClients (tUnqualified target))
   withExceptT clientError $
     wrapHttpClientE $ guardLegalhold (ProtectedUser (tUnqualified lusr)) (mkUserClients [(tUnqualified target, clients)])
--- a/services/brig/src/Brig/API/MLS/KeyPackages.hs
+++ b/services/brig/src/Brig/API/MLS/KeyPackages.hs
@@ -34,6 +34,7 @@ import Control.Monad.Trans.Maybe
 import Data.Id
 import Data.Qualified
 import qualified Data.Set as Set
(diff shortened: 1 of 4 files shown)


services/brig/test/integration/API/MLS.hs
```diff
@@ -106,6 +107,54 @@ testKeyPackageClaim brig = do
...
+    bundle <-
+      responseJsonError
+        =<< post
+          ( brig
+              . paths ["mls", "key-packages", "claim", toByteString' (qDomain u), toByteString' (qUnqualified u)]
+              . queryItem "skip_own" (toByteString' c1)
+              . zUser (qUnqualified u')
+          )
+          <!! const 200 === statusCode
+    liftIO $ Set.map (\e -> (kpbeUser e, kpbeClient e)) (kpbEntries bundle) @?= Set.fromList [(u, c1), (u, c2)]
```
mdimjasevic, line 156: Shouldn't entries for `c2` be absent from the response? I would expect so because there were 3 key packages uploaded for `c2` and they were all claimed in a block with line numbers 121-132.

mdimjasevic (approved):

services/brig/src/Brig/API/MLS/KeyPackages.hs
```diff
@@ -47,24 +48,29 @@ uploadKeyPackages lusr cid (kpuKeyPackages -> kps) = do
...
     (\_ -> throwStd federationNotImplemented)
+    target
 
-claimLocalKeyPackages :: Local UserId -> Local UserId -> Handler r KeyPackageBundle
-claimLocalKeyPackages lusr target = do
+claimLocalKeyPackages :: Local UserId -> Maybe ClientId -> Local UserId -> Handler r KeyPackageBundle
+claimLocalKeyPackages lusr skipOwn target = do
+  -- skip own client when the target is the requesting user itself
+  let own = guard (lusr == target) *> skipOwn
+  traceM $ "own: " <> show own
```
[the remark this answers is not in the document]
mdimjasevic: Please remove `traceM` before merging.

services/brig/test/integration/API/MLS.hs
pcapriotti: Why were they all claimed? Only *one* key package per client is claimed by each endpoint call.

mdimjasevic: You're right! My mistake.

Maybe another useful assertion in the last block would be on the count of available key packages per client after `u'` claimed a key package.

pcapriotti: That's covered by the test above, though, isn't it?

mdimjasevic: Are you referring to the block in lines 134-142? If so, I thought the outcome is different when `u'` claims key packages compared to `u` claiming it.

pcapriotti: Alright, I added the assertion.

PR #2288, opened 2022-04-14
[Polysemy] Introduce an Effect for Password Resetting

mdimjasevic:
The PR introduces the first Polysemy-based effects to Brig. One effect is for a code store. Combined with `Now` (PR #2292), it is  combined into a higher-level effect for password resetting.

In some places an effect is immediately interpreted at the use site; this is to avoid introducing embeddings of various constrained monads. As more effects will be introduced in subsequent PRs, it will be possible to bubble up these interpretations and eventually have them only at the very top level when running the application.

## Checklist


services/brig/src/Brig/App.hs
```diff
@@ -575,6 +580,15 @@ newtype HttpClientIO a = HttpClientIO
       MonadIndexIO
     )
 
+type BrigEffects =
```
isovector, line 583: I'd suggest calling this sometihng like `BrigCanonicalEffects` to highlight that other choices are possible

mdimjasevic: Done in bba1c0998.

services/brig/src/Brig/App.hs
```diff
@@ -575,6 +580,15 @@ newtype HttpClientIO a = HttpClientIO
...
+  '[ CodeStore,
+     Embed Cas.Client, -- FUTUREWORK: remove this effect once MonadClient
+     -- constraints are removed from application code
+     BrigTime,
```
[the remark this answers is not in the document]
mdimjasevic: Done in 7216817ed.

services/brig/src/Brig/Sem/BrigTime.hs
```diff
@@ -0,0 +1,27 @@
...
+-- You should have received a copy of the GNU Affero General Public License along
+-- with this program. If not, see <https://www.gnu.org/licenses/>.
+{-# LANGUAGE TemplateHaskell #-}
+
+module Brig.Sem.BrigTime where
+
+import Data.Time (UTCTime)
+import Polysemy
+
+data BrigTime m a where
```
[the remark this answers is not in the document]
mdimjasevic: Thanks for pointing this out!

I just made a PR that moves the `Now` effect out of Spar and into a library: https://github.com/wireapp/wire-server/pull/2292

I'll rebase this PR on top of that one so my apologies for rewriting the Git history.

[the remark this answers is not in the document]
mdimjasevic: > I'll rebase this PR on top of that one so my apologies for rewriting the Git history.

Done. @isovector , if you could review again, I'd appreciate that a lot!

services/brig/src/Brig/App.hs
```diff
@@ -576,6 +583,16 @@ newtype HttpClientIO a = HttpClientIO
       MonadIndexIO
     )
 
+type BrigCanonicalEffects =
+  '[ PasswordResetStore,
+     Now,
+     CodeStore,
+     Embed Cas.Client, -- FUTUREWORK: remove this effect once MonadClient
```
isovector, line 590: I'm not sure this FUTUREWORK is correct; I think you'll want to keep this forever.

services/brig/src/Brig/App.hs
```diff
@@ -609,10 +626,19 @@ instance MonadIndexIO (AppT r) => MonadIndexIO (ExceptT err (AppT r)) where
...
+runAppT e (AppT ma) =
+  runFinal
+    . embedToFinal
+    . interpretClientToIO (_casClient e)
+    . codeStoreToCassandra @Cas.Client
+    . brigTimeToMonadIO @IO (_currentTime e)
+    . passwordResetStoreToCodeStore
+    $ runReaderT ma e
+
+runAppResourceT ::
```
isovector, line 639: This function is getting more dangerous as we add more effects. Right now we're fine, but any sort of resumptive effects will silently get the wrong results when being pushed through this.

services/brig/src/Brig/Sem/CodeStore.hs
```diff
@@ -0,0 +1,44 @@
...
+import Imports
+import Polysemy
+
+data CodeStore m a where
+  MkPasswordResetKey :: UserId -> CodeStore m PasswordResetKey
+  GenerateEmailCode :: CodeStore m PasswordResetCode
+  GeneratePhoneCode :: CodeStore m PasswordResetCode
+  CodeSelect ::
+    PasswordResetKey ->
+    CodeStore m (Maybe (PasswordResetCode, UserId, Maybe Int32, Maybe UTCTime))
```
isovector, line 33: Please build a record for this output type and return that

services/brig/src/Brig/Sem/Now/MonadIO.hs
```diff
@@ -0,0 +1,41 @@
...
+-- | An interpreter of the 'Now' effect to MonadIO.
+brigTimeToMonadIO ::
+  forall m r a.
+  (MonadIO m, Member (Embed m) r) =>
+  IO UTCTime ->
+  Sem (Now ': r) a ->
+  Sem r a
+brigTimeToMonadIO ioTime = interpret now
+  where
+    -- If this helper was inlined, GHC wouldn't be able to figure out
```
isovector, line 35: I think it should if you enable `{-# OPTIONS_GHC -fplugin=Polysemy.Plugin #-}`

services/brig/src/Brig/Sem/Now/MonadIO.hs
```diff
@@ -0,0 +1,41 @@
...
+module Brig.Sem.Now.MonadIO (brigTimeToMonadIO) where
+
+import Data.Time
+import Imports
+import Polysemy
+import Wire.Sem.FromUTC
+import Wire.Sem.Now
+
+-- | An interpreter of the 'Now' effect to MonadIO.
+brigTimeToMonadIO ::
```
isovector, line 27: Is this interpretation different from the one already in Spar.Sem.Now.IO?

services/brig/src/Brig/Sem/PasswordResetStore/CodeStore.hs
```diff
@@ -0,0 +1,86 @@
...
+import Data.Time
+import Imports hiding (lookup)
+import Polysemy
+import Wire.Sem.Now
+import qualified Wire.Sem.Now as Now
+
+passwordResetStoreToCodeStore ::
+  forall r a.
+  Members '[CodeStore, Now] r =>
+  Sem (PasswordResetStore ': r) a ->
```
isovector, line 36: I really like this effect/interpretation!

services/brig/test/integration/API/User/Util.hs
```diff
@@ -188,15 +190,23 @@ initiateEmailUpdateNoSend brig email uid =
...
   let qry = queryItem "email" (toByteString' email)
   r <- get $ brig . path "/i/users/password-reset-code" . qry
   let lbs = fromMaybe "" $ responseBody r
   let Just pwcode = PasswordResetCode . Ascii.unsafeFromText <$> (lbs ^? key "code" . _String)
-  ident <- PasswordResetIdentityKey <$> mkPasswordResetKey uid
+  ident <- PasswordResetIdentityKey <$> sem (mkPasswordResetKey uid)
   let complete = CompletePasswordReset ident pwcode newpw
   return complete
+  where
+    sem = liftIO . runFinal @IO . interpretClientToIO cs . codeStoreToCassandra @DB.Client
```
isovector, line 209: Can you use `runAppT` here instead?

isovector (approved):
LGTM; minor nits

services/brig/src/Brig/Sem/Now/MonadIO.hs
mdimjasevic: Ignoring for a moment that the one you had in Spar runs in IO and this one runs in `m` for `MonadIO m`, this interpretation takes as input an IO action of type `IO UTCTime` from which it extracts the current time. After more thought in the afternoon, I realized that this interpretation is more general than the one in Spar/PR #2292 and the later can be implemented in terms of the former; I'll do that in PR #2292 and just reuse that more general interpreter in Brig.

services/brig/src/Brig/Sem/Now/MonadIO.hs
mdimjasevic: You'll have to show me in the next pairing session how to make use of the plugin.

changelog.d/5-internal/polysemy-brig
```diff
@@ -0,0 +1 @@
+Introduce first effects to Brig: code store and time
```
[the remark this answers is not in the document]
mdimjasevic: Thanks for spotting it! Fixed in 14771b263.

services/brig/src/Brig/App.hs
mdimjasevic: Are you referring to things like throwing and catching, e.g, like in `wrapMonadClient` in one of your first Spar PRs? https://github.com/wireapp/wire-server/pull/1775/files#diff-718f7e7fa3d155d592d1551bd8fb924edf14990399056f4696b90ca27933dde2R215

services/brig/test/integration/API/User/Util.hs
mdimjasevic: I cannot, right? `runAppT` expects a `PasswordReset` effect at the beginning of the effect row, yet this use site is only about the lower-level code store effect.

services/brig/src/Brig/App.hs
mdimjasevic: You're right: code store is interpreted to `Embed m` for `MonadClient m`, e.g., `Cas.Client`.

Fixed in e090dd9b8c8ade4736a4e21111861ba6204d55d8.

services/brig/src/Brig/Sem/CodeStore.hs
mdimjasevic: Done in 3190c082f7133ae1a66554e44e69fc904ca502ba.

services/brig/src/Brig/Sem/Now/MonadIO.hs
mdimjasevic: As I pointed out in PR #2292, I moved this function into the library as it is not Brig-specific. You can take a look there to see a subtle difference compared to how it used to be in Spar.Sem.Now.IO.

services/brig/src/Brig/App.hs
isovector: An effect like `Error e` or `State s` will get the wrong result if it's pushed through `runAppResourceT`

services/brig/test/integration/API/User/Util.hs
isovector: Fair enough

PR #2291, opened 2022-04-15
Remove TemplateHaskell as a Global Default Extension

mdimjasevic:
`TemplateHaskell` as a language extension can lead to unnecessary recompilations. To aid GHC in figuring out what needs recompilation, this PR removes `TemplateHaskell` as a global default extension and moves it to modules that do need it. Hopefully this reduces the compilation time by a noticeable amount of time.

The one-line change that triggered the rest is the removal of `TemplateHaskell` from `package-defaults.yaml`.

## Checklist

### DIFF
--- a/changelog.d/5-internal/th
+++ b/changelog.d/5-internal/th
@@ -1,0 +1,1 @@
+Remove TemplateHaskell as a global default extension
--- a/libs/api-bot/api-bot.cabal
+++ b/libs/api-bot/api-bot.cabal
@@ -4,7 +4,7 @@ cabal-version: 1.12
 --
 -- see: https://github.com/sol/hpack
 --
--- hash: 208b12ff1a7bb406c0e9b006b8badb293d9f3fb39750fa4df03a5267cf5d7bfd
+-- hash: 4010a19af0b65ce213a7afb528aa57a5ae3dba5fd870c09d28781afae11d2c4b
 
 name:           api-bot
 version:        0.4.2
@@ -68,7 +68,6 @@ library
       RankNTypes
       ScopedTypeVariables
       StandaloneDeriving
-      TemplateHaskell
       TupleSections
       TypeApplications
       TypeFamilies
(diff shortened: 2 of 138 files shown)


fisx (approved):
Did you take the time before and after?  Either way, as long as it doesn't slow down compilation I'm in favor.  :)

mdimjasevic:
> Did you take the time before and after? Either way, as long as it doesn't slow down compilation I'm in favor. :)

I didn't. At first and without too much thinking, I thought I'd see some time savings in the CI runs for this PR. However, I think it's the same time as before because everything is compiled from scratch. Now I am expecting time savings while developing as that's when we edit already compiled modules and recompile them.

@isovector brought this up in one of our pairing sessions. I don't have a good reference for this, though I tried looking it up. To some extent these are relevant:

- https://well-typed.com/blog/2022/04/hls-performance/
- https://www.parsonsmatt.org/2021/07/12/template_haskell_performance_tips.html

mdimjasevic:
And this cannot slow down the compilation as here we're making the reach of TemplateHaskell strictly smaller, suggesting GHC knows more about what needs to be recompiled.

mdimjasevic:
I'll proceed with merging. In the unlikely case of us having a proof this slows down the compilation, it can easily be reverted.

fisx:
```
git co template-haskell-per-module
make clean
time make
git revert --no-commit a776512b0035eabf0f941ea31aa0e4d00e40fa45
make clean
time make
```

=>

```
real	8m33.984s
real	8m39.574s
```

i don't think that's significant, so all good!

isovector:
The performance improvements here are going to be in incremental builds. GHC maintains interface files which hash their exports; subsequent files only need to get rebuilt if one of their imports gets changed. TemplateHaskell messes with this machinery, where it doesn't check which imports are used by *splices;* thus, if any dependency of a module with TH changes, the TH module will rebuild, and invalidate the build artifacts of everything downstream.

PR #2292, opened 2022-04-15
Move the Now Effect from Spar into a Library

mdimjasevic:
The `Now` effect introduced to Spar during polysemisation can be turned into a more general and more useful effect that can be used in other services. This PR does that by moving the effect into the polysemy-wire-zoo library.

This is going to be useful in polysemisation of Brig as it also needs a time effect.

TODOs:

- [x] Move `brigTimeToMonadIO` from PR #2288 to here and rename it. It is not specific to Brig and `nowToMonadIO` can be implemented in terms of it.

## Checklist


libs/polysemy-wire-zoo/src/Wire/Sem/FromUTC.hs
```diff
@@ -15,16 +15,23 @@
...
-nowToIO :: Member (Embed IO) r => Sem (Now ': r) a -> Sem r a
-nowToIO = interpret $ \case
-  Get -> embed @IO getNowIO
+-- | The class is helpful in interpreters for the 'Now' effect. It makes it
+-- possible to use the interpreters for any time type so long as it implements
+-- this single-method class.
+class FromUTC a where
+  fromUTCTime :: UTCTime -> a
+
+-- An orphan instance
```
isovector, line 31: These are not orphans!

libs/polysemy-wire-zoo/src/Wire/Sem/Now.hs
```diff
@@ -27,10 +27,10 @@ where
 import Imports
 import Polysemy
 import Polysemy.Check (deriveGenericK)
-import qualified SAML2.WebSSO as SAML
+import Wire.Sem.FromUTC
 
 data Now m a where
-  Get :: Now m SAML.Time
+  Get :: FromUTC a => Now m a
```
isovector, line 33: I don't love this change; is there a reason to prefer this over calling `fromUtcTime` at the call site?

libs/polysemy-wire-zoo/src/Wire/Sem/Now/MonadIO.hs
```diff
@@ -0,0 +1,53 @@
...
+import Imports
+import Polysemy
+import Wire.Sem.FromUTC
+import Wire.Sem.Now
+
+-- | An interpreter of the 'Now' effect to MonadIO via a custom IO action that
+-- provides the current time.
+nowToMonadIOAction ::
+  forall m r a.
+  (MonadIO m, Member (Embed m) r) =>
```
isovector, line 35: What's the value of doing this in `MonadIO` rather than in `IO` directly? `MonadIO` doesn't make much sense in the land of effects.

libs/polysemy-wire-zoo/src/Wire/Sem/Now.hs
mdimjasevic: My goal here was to make the `Get` action work not only for the `Time` type from that SAML module, but also for `UTCTime`.

If I remove this constraint then I don't know how to make it work. In the interpreter I just can't get types line up. Sure, `fromUTCTime` has to be called, but then there needs to be an instance of `FromUTC` for the type variable `x`, and `x` is completely unconstrained there and I don't see how to make it have this constraint.

I'm open to suggestions, though!

libs/polysemy-wire-zoo/src/Wire/Sem/Now/MonadIO.hs
mdimjasevic: Oh, OK, I didn't know that. I was just trying to make it as polymorphic as possible. I'll revert this back to IO.

libs/polysemy-wire-zoo/src/Wire/Sem/FromUTC.hs
mdimjasevic: You're right! Silly me. I'll delete those comments.

libs/polysemy-wire-zoo/src/Wire/Sem/Now.hs
isovector: i'd suggest `Get :: Now a UtcTime` and then change client code to turn that into the right type

mdimjasevic: Is what I did in commit https://github.com/wireapp/wire-server/pull/2292/commits/c6079896c1db25b741d737a743a194d0fe6e7c72 what you had in mind?

isovector (approved):

PR #2293, opened 2022-04-19
bulkpush on by default in gundeck's helm chart

jschaul:
We've been using this setting ourselves for months as can be seen in https://github.com/zinfra/cailleach/blob/master/environments/hegemony/roles/gundeck/vars/main.yml#L11-L12 - there's no reason this should not be the default these days.

## Checklist


supersven (approved):
Looks good to me :+1:

PR #2297, opened 2022-04-20
API versioning

pcapriotti:
Implement API versioning system. See `docs/legacy/developer/api-versioning.md` for details.

Tracked by:
 - https://wearezeta.atlassian.net/browse/FS-127 (federation API versioning)
 - https://wearezeta.atlassian.net/browse/SQCORE-1209 (add versions to client API routing tables)
 - https://wearezeta.atlassian.net/browse/FS-63 (version negotiation)

## Checklist


changelog.d/1-api-changes/api-versioning
```diff
@@ -0,0 +1,30 @@
+Start version 2 of the public API. Main changes:
+
+ - Asset endpoints have lost their `v3` and `v4` suffixes. So for example
+   `/assets/v3` has been replaced by `/access`.
```
fisx, line 4: I hope this is a typo?  :)

```suggestion
   `/assets/v3` has been replaced by `/assets`.
```

Also, isn't that something that could have gone into a separate PR?

docs/legacy/developer/api-versioning.md
```diff
@@ -1,464 +1,279 @@
...
+Stable versions are all supported versions that are not development.
 
-In the following, we will refer to HTTP clients as "clients", no
-matter whether it is an app or a backend talking to another backend
-(federation); and to HTTP servers as "server", no matter which API is
-it serving (federation or app).
+Stable versions map to well-defined and fixed API *contracts*, which will not
+change over time. That is to say, if two different instances of the backend
+both list version `3` in `supported`, a client can assume that they will accept
+the same exact requests when version `3` is used.
```
fisx, line 33:
```suggestion
the same exact requests when version `3` is used, and handle them in exactly the same way.
```

docs/legacy/developer/api-versioning.md
```diff
@@ -1,464 +1,279 @@
...
 
+On the other hand, development versions give no such guarantees. When making
+requests using a development version on some backend, the client needs to be
+aware of the corresponding API contract used by that specific backend.
 
-# Versions and servant routing tables
+Of course, development versions are useful while building a new API, but are
+not suitable for production. Backends deployed to production environments
+should disable development versions (and not advertise them in `/api-version`).
```
fisx, line 41:
```suggestion
should disable development versions (and not advertise them in `/api-version`).

On the other hand, production clients must disregard all development versions
as unsupported.
```

This would be easier if development versions would not also be listed as versions.  I'm actually in favor of this.  I think it's a lot less error-prone.

docs/legacy/developer/api-versioning.md
```diff
@@ -1,464 +1,279 @@
...
-`Versions version :>` in the routing type instead of `"v1"`.
+Before making a request to the server, the client needs to have negotiated
+which version to use. The recommended flow is as follows:
 
-2 seems a lot less work to write, read, and understand.
+ - query `GET /api-version` and retrieve the set of supported and development
+   versions;
+ - decide whether using a development version is unacceptable, and
+   if so, take it out of consideration;
+ - use the latest (i.e. numerically larger) version that the client supports;
```
fisx, line 199:
```suggestion
 - use the latest (i.e. numerically largest) version that the client supports;
```

fisx (commented):
I like the outcome of this process!  I think it's both pragmatic and reasonably elegant.  I have a few nit-picks, and haven't read enough of this PR to approve it, but I'm generally happy.

changelog.d/1-api-changes/api-versioning
pcapriotti: It could've, but I thought it would be useful to start using the new system immediately. It's kind of like adding tests for X in the same PR where you implement X itself. In fact, it is a bit worrying that we are not making significant use of it for the federation API. I've played around with it in a branch, though, by making some random changes to the API, and it seemed to be working reasonably well there, too.

docs/legacy/developer/api-versioning.md
pcapriotti: I went back and forth with this, and in the end I think I prefer the current system, because it places no burden on clients. Clients can just start using V2 now, without any special logic, and as long as we disable it in production, it won't affect anything.

Also, it makes it easier to test, because we only need to make sure that the backend has a development version enabled, and not that the client has turned out development API access as well.

But I can be convinced otherwise. This is not too problematic to change later on, anyway.

pcapriotti: I have added the following paragraph:
> Similarly, clients that are meant for production use can decide to ignore
development versions on their backend. This is not strictly necessary, but it
can be used as a safeguard against mistakes in deployment.

fisx: > I went back and forth with this, and in the end I think I prefer the current system, because it places no burden on clients. Clients can just start using V2 now, without any special logic, and as long as we disable it in production, it won't affect anything.
> 
> Also, it makes it easier to test, because we only need to make sure that the backend has a development version enabled, and not that the client has turned out development API access as well.
> 
> But I can be convinced otherwise. This is not too problematic to change later on, anyway.

I'm ok with everything except the last sentence.  :)  So, yes, do it your way.

services/cargohold/src/CargoHold/API/Public.hs
```diff
@@ -57,6 +58,12 @@ servantSitemap =
     providerAPI = uploadAssetV3 @tag :<|> downloadAssetV3 @tag :<|> deleteAssetV3 @tag
     legacyAPI = legacyDownloadPlain :<|> legacyDownloadPlain :<|> legacyDownloadOtr
     qualifiedAPI = downloadAssetV4 :<|> deleteAssetV4
+    mainAPI =
+      renewTokenV3
+        :<|> deleteTokenV3
+        :<|> uploadAssetV3 @'UserPrincipalTag
+        :<|> downloadAssetV4
+        :<|> deleteAssetV4
```
mdimjasevic, line 66: While at it, how about also renaming these handlers to remove old-style version suffices?

libs/wire-api/src/Wire/API/Routes/Version.hs
```diff
@@ -14,63 +16,78 @@
...
-  Aeson.Success v -> pure v
+instance FromHttpApiData Version where
+  parseHeader = first Text.pack . Aeson.eitherDecode . LBS.fromStrict
+  parseUrlPiece = parseHeader . Text.encodeUtf8
 
 supportedVersions :: [Version]
 supportedVersions = [minBound .. maxBound]
 
+developmentVersions :: [Version]
+developmentVersions = [V2]
```
mdimjasevic, line 80: Would it make more sense to put `[maxBound]` for the value?

mdimjasevic (approved):
Very nice work!

Minor comments are left inlined. I am not sure why GitHub shows the PR as introducing 32,648 new lines. Is it really that many? I suppose it's the exported Swagger versioned docs that account for the vast majority of the lines.

mdimjasevic:
A suggestion (and this can easily be left for a follow-up PR) is to have a test API with endpoints that use both the `Until` and the `From` combinators (I guess they can be used to form a version range). I'm saying a test API because we don't have an endpoint in a client API or a federation API that uses both combinators.

services/cargohold/src/CargoHold/API/Public.hs
pcapriotti: I thought about it, but they're also used by the actual endpoints with v3 and v4 suffixes, so it kind of makes sense to leave it like this for now. We can normalise the naming once API v1 gets removed from the code.

libs/wire-api/src/Wire/API/Routes/Public/Galley.hs
```diff
@@ -275,15 +276,26 @@ type ConversationAPI =
                :> Get '[Servant.JSON] (ConversationList Conversation)
            )
     :<|> Named
-           "list-conversations"
+           "list-conversations-v1"
```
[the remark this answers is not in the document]
pcapriotti: That v1 refers to the API version, not the suffix.

pcapriotti:
> A suggestion (and this can easily be left for a follow-up PR) is to have a test API with endpoints that use both the `Until` and the `From` combinators (I guess they can be used to form a version range). I'm saying a test API because we don't have an endpoint in a client API or a federation API that uses both combinators.

Sure, this can be done, but what would we test exactly?

mdimjasevic:
> 
> Sure, this can be done, but what would we test exactly?

We could test a number of things: for `Until vN` we could test if an endpoint is available at version `vN`, then the same for `From`, that for `vN < vM` and an enpoint with `From vN` and `Until vM` we can use the endpoint at any version within the range, and that for `vP > vR` and an endpoint with `From vP` and `Until vR` we do not actually have the endpoint (assuming this combination type-checks).

smatting (approved):

PR #2306, opened 2022-04-22
[Polysemy] Move the Logger Effect to polysemy-wire-zoo

mdimjasevic:
The PR generalises and moves the Logger effect from Spar into polysemy-wire-zoo.

## Checklist


pcapriotti:
There is already a logger effect in polysemy-wire-zoo, which we use in galley and federator. Would it make sense to merge the two?

mdimjasevic:
> There is already a logger effect in polysemy-wire-zoo, which we use in galley and federator. Would it make sense to merge the two?

It crossed my mind and I was lazy to look at that. But you're right, this is a good opportunity to do that so I'll do it.

libs/polysemy-wire-zoo/src/Wire/Sem/Logger/TinyLog.hs
```diff
@@ -0,0 +1,63 @@
...
+      . Log.log tinylog (toLevel lvl)
+      $ Log.field "request" (unRequestId r) Log.~~ msg
+
+stringLoggerToTinyLog :: Member (Logger (Log.Msg -> Log.Msg)) r => Sem (Logger String ': r) a -> Sem r a
+stringLoggerToTinyLog = mapLogger @String Log.msg
+
+discardLogs :: Sem (Logger msg ': r) a -> Sem r a
+discardLogs = interpret $ \(Log _ _) -> pure ()
+
+discardTinyLogs :: Sem (Logger (Log.Msg -> Log.Msg) ': r) a -> Sem r a
```
isovector, line 62: Why do you need this function?

services/federator/test/unit/Test/Federator/Response.hs
```diff
@@ -58,7 +62,12 @@ testValidationError =
 testServerError :: TestTree
 testServerError =
   testCase "server errors should be converted to wai error responses" $ do
-    resp <- runM . TinyLog.discardLogs . runWaiError @ServerError $ throw InvalidRoute
+    resp <-
+      runM
+        . Log.discardTinyLogs
+        . runWaiError @ValidationError
```
isovector, line 68: This `@ValidationError` is new. Intentional?

isovector (changes requested):
Looks good, but I'd suggest dropping `discardTinyLogs`

libs/polysemy-wire-zoo/src/Wire/Sem/Logger/TinyLog.hs
mdimjasevic: This is a help in type inference. If I don't use it, then GHC reports overlapping instances. An alternative is to use `discardLogs @(Log.Msg -> Log.Msg)` instead of `discardTinyLogs`.

services/federator/test/unit/Test/Federator/Response.hs
mdimjasevic: Thanks for spotting! This was not intentional.

libs/polysemy-wire-zoo/src/Wire/Sem/Logger/TinyLog.hs
isovector: That's surprising to me. Does the polysemy plugin help?

services/federator/test/unit/Test/Federator/Response.hs
mdimjasevic: Removed in b4b3ae05a.

libs/polysemy-wire-zoo/src/Wire/Sem/Logger/TinyLog.hs
mdimjasevic: > That's surprising to me. Does the polysemy plugin help?

It doesn't seem to. I tried this: 1) put the plugin as a GHC option in the cabal file and make its package a dependency, 2) use `discardLogs` instead of `discardTinyLogs` where I'm hoping the plugin would help, 3) observe I get the aforementioned error on overlapping instances, e.g.:

```
src/Federator/MockServer.hs:93:15: error:
    * Overlapping instances for Member
                                  (Wire.Sem.Logger.Logger
                                     (System.Logger.Message.Msg -> System.Logger.Message.Msg))
                                  '[Wire.Sem.Logger.Logger msg0, Embed IO]
        arising from a use of `runWaiErrors'
      Matching instances:
        two instances involving out-of-scope types
        (use -fprint-potential-instances to see them all)
      (The choice depends on the instantiation of `msg0'
       To pick the first instance above, use IncoherentInstances
       when compiling the other instance declarations)
    * In the second argument of `(.)', namely
        `runWaiErrors @'[ValidationError, ServerError, MockException]'
      In the second argument of `(.)', namely
        `discardLogs
           . runWaiErrors @'[ValidationError, ServerError, MockException]'
      In the expression:
        runM
          . discardLogs
              . runWaiErrors @'[ValidationError, ServerError, MockException]
   |
93 |             . runWaiErrors
   |               ^^^^^^^^^^^^...
cabal: Failed to build federator-1.0.0 (which is required by brig-2.0).
```

mdimjasevic:
> There is already a logger effect in polysemy-wire-zoo, which we use in galley and federator. Would it make sense to merge the two?

Done! I made the one that has been in Galley and Federator just a special case of the one I generalised from Spar.

PR #2315, opened 2022-04-25
[Polysemy] Refactor Password Resetting Effects

mdimjasevic:
This patch improves effects around password resetting. It prevents the `CodeStore` effect from bubbling up in function constraints by introducing a `PasswordResetStore` action that just maps to the corresponding `CodeStore` action.

The `CodeStore` effect has been split into DB query actions and those that generate values (in IO) needed for password resetting; the new effect is `PasswordResetSupply`.

@isovector , does this look good to you? In some places `PasswordResetStore` shows up in `Members` constraints along with `PasswordResetSupply`, even though the former is interpreted in terms of the later (but `PasswordResetSupply` actions are not used exclusively for the interpretation of the store effect). One alternative for avoiding having them both show up in `Members` constraints would be to introduce a new effect `PasswordReset` that is interpreted in terms of `PasswordResetStore` and `PasswordResetSupply` where both are needed.

## Checklist


services/brig/src/Brig/Data/Activation.hs
```diff
@@ -135,6 +137,17 @@ activateKey k c u = verifyCode k c >>= pickUser >>= activate
         updateEmailAndDeleteEmailUnvalidated :: MonadClient m => UserId -> Email -> m ()
         updateEmailAndDeleteEmailUnvalidated u' email =
           updateEmail u' email <* deleteEmailUnvalidated u'
+        deleteCode :: UserId -> m ()
+        deleteCode uId =
+          runM
+            -- FUTUREWORK: use the DeletePasswordResetCode action instead of CodeDelete
+            ( codeStoreToCassandra @m
+                ( embed @m
+                    ( liftIO @m (runM (passwordResetSupplyToIO @'[Embed IO] (E.mkPasswordResetKey uId)))
```
isovector, line 146: This looks wrong to me. It should be something like

```
runM . codeStoreToCassandra . passwordResetSupplyToIO $ E.mkPasswordResetKey uId >>= E.codeDelete
```

isovector: But also this is a dangerous thing to do in general and would be better to push these constraints up to the type signature instead.

isovector (changes requested):
suspicious use of runM

services/brig/src/Brig/Data/Activation.hs
mdimjasevic: > But also this is a dangerous thing to do in general and would be better to push these constraints up to the type signature instead.

The function doesn't yet run in `Sem`, hence this soon-to-be-done-away-with running of effects. I will push the constraints up to the type signature in one of the follow-up PRs.

mdimjasevic: @isovector , in case you find time, let me know what you think about what I wrote above.

isovector: I had another look at this, and I think I understand the issue you're running into. This function doesn't yet run in `Sem`, but you want to run `Sem` actions inside of it?

The solution here is to port this function into `Sem`. To do the `runM` dance currently in this PR is extremely dangerous. By using `runM`, you are committing to a particular interpretation of your effects, which may or *may not* be the same choices that are made globally. That means, eg, maybe a test wants to run against an in memory database, but this function has committed to using Cassandra. In the best case, your tests will fail, in the worst, you'll corrupt the database invariants.

I'd recommend using this technique even in the short-term as a crutch, because the compiler can't help if we forget to take the crutch out. It'll just be a sitting time bomb, waiting to waste someone's day when they run afoul of the presumed interpretation.

isovector: A better solution here for the interim is to give (orphan) `Member (Embed Client) r => MonadClient (Sem r)` and similar `MonadReader` instances to `Sem`. Then you don't need to play the interpretation game, and you can still interop mtl code with polysemy before having the whole thing ported.

services/brig/src/Brig/App.hs
```diff
@@ -493,6 +494,21 @@ instance MonadReader Env (AppT r) where
...
   local f (AppT m) = AppT $ local f m
 
+instance Member (Input Env) r => MonadReader Env (Sem r) where
+  ask = input
+  local f sem = do
+    env <- input
+    runInputConst (f env) . raise @(Input Env) $ sem
+
+instance
+  ( Members '[Final IO, Embed IO, Embed Cas.Client] r,
```
isovector, line 509: Why do you need IO here?

mdimjasevic: I'm deep into compilation issues so I can't answer this off the bat. I think something needed it.

I'm thinking of closing this PR. First I've started cleaning up per your suggestion, but then I got into a rabbit hole that looked a lot like what I was already doing in a PR that builds on top of this, but that has more things polysemized.

PR #2316, opened 2022-04-25
Fix versioned metrics

pcapriotti:
This makes sure that every other middleware sees the rewritten (unversioned) path. In particular, the prometheus middleware will now only see paths it knows about, which prevents it from reporting "N/A" as the path.

## Checklist

### DIFF
--- a/changelog.d/5-internal/fix-versioned-metrics
+++ b/changelog.d/5-internal/fix-versioned-metrics
@@ -1,0 +1,5 @@
+Apply `versionMiddleware` last. This makes sure that every other middleware sees
+the rewritten (unversioned) path. In particular, the prometheus middleware will
+now only see paths it knows about, which prevents it from reporting "N/A" as the
+path.
+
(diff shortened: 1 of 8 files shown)


smatting (approved):

smatting:
I hope that another side-effect of building without optimization (sorry I forgot this) is that cabal can use use caches from the `ubuntu20-builder` builder now, which also builds without optimization.

jschaul:
> I hope that another side-effect of building without optimization (sorry I forgot this) is that cabal can use use caches from the ubuntu20-builder builder now, which also builds without optimization.

Don't we need to build images that are production ready, with optimizations? Those charts/images we release should have optimizations enabled, no?
For CI I agree it doesn't matter, but perhaps in some refactoring this was turned off at some point?

smatting:
> > I hope that another side-effect of building without optimization (sorry I forgot this) is that cabal can use use caches from the ubuntu20-builder builder now, which also builds without optimization.
> 
> Don't we need to build images that are production ready, with optimizations? Those charts/images we release should have optimizations enabled, no? For CI I agree it doesn't matter, but perhaps in some refactoring this was turned off at some point?

Sorry, my comment was meant for another PR: https://github.com/zinfra/cailleach/pull/1027

Yes, production is build with optimization.

PR #2318, opened 2022-04-25
Add instructions to use cachix when using nix+direnv

fisx:
## Checklist

### DIFF
--- a/docs/developer/dependencies.md
+++ b/docs/developer/dependencies.md
@@ -30,7 +30,12 @@ You can then run all the builds, and the native dependencies will be automatical
    * When successfully installed and hooked, direnv should ask you to `direnv allow`
      the current `.envrc` when you cd to this repository.
      See the [Installation documentation](https://direnv.net/docs/installation.html) for further details.
-3. [Optional, but very helpful] install [cachix](https://app.cachix.org/cache/wire-server).
+3. Optionally, enable the [wire-server Cachix cache](https://app.cachix.org/cache/wire-server).
+   * The Cachix page includes instructions for using the `cachix` command-line
+     tool for enabling the wire-server cache.
+   * Alternatively, you can enable the cache manually by adding the
+     cache's public key (listed on the Cachix page) and the substituter URL
+     (https://wire-server.cachix.org) to your `nix.conf` file.
 
 ### Fedora:
 
--- a/docs/developer/dependencies.md
+++ b/docs/developer/dependencies.md
@@ -30,6 +30,7 @@ You can then run all the builds, and the native dependencies will be automatical
    * When successfully installed and hooked, direnv should ask you to `direnv allow`
      the current `.envrc` when you cd to this repository.
      See the [Installation documentation](https://direnv.net/docs/installation.html) for further details.
+3. [Optional, but very helpful] install [cachix](https://app.cachix.org/cache/wire-server).
 
 ### Fedora:
 
--- a/docs/developer/dependencies.md
+++ b/docs/developer/dependencies.md
@@ -20,7 +20,6 @@ You can then run all the builds, and the native dependencies will be automatical
    * MacOS users with a recent Mac might need to follow [these
    instructions](https://nixos.org/nix/manual/#sect-macos-installation)
    * Debian users can use their distro's `nix` package, and should remember
-   
    to add their user to the `nix-users` group in /etc/group, and re-start
    their login session.
 2. Install [Direnv](https://direnv.net/).


fisx:
if you approve this, please merge right away.

sysvinit (changes requested):
I think more detailed instructions might be more appropriate, so we assume less prior knowledge about Nix from the reader. In particular, I think something closer to what we have in the README for cailleach might be more appropriate -- the Cachix information page has various instructions about installing external components, but to download from the cache all you need is a few lines in your `nix.conf` file.

fisx:
> I think more detailed instructions might be more appropriate, so we assume less prior knowledge about Nix from the reader. In particular, I think something closer to what we have in the README for cailleach might be more appropriate -- the Cachix information page has various instructions about installing external components, but to download from the cache all you need is a few lines in your `nix.conf` file.

I was about to add something like:

```
   * either follow the steps in the above link or the more detailed instructions [here](https://github.com/zinfra/cailleach/#development-setup).
```

But then realized that cailleach is a private repo.  Are you sure thought that you want me to complicate the instructions further?  The steps under the link worked fine for me, so either I was lucky, or that's enough to get things working for people who don't care enough to find their own sources of information.

Can I change your mind here?

sysvinit:
> Are you sure thought that you want me to complicate the instructions further? The steps under the link worked fine for me, so either I was lucky, or that's enough to get things working for people who don't care enough to find their own sources of information.

I'm not sure along which axis additional complexity would be introduced here. I think there's some value in having slightly more detailed documentation, even if it makes it slightly longer, given that (I assume) the target audience of these docs is not guaranteed to be *a priori* familiar with these tools. Moreover, I think our public dev setup docs should match the internal docs, for consistency -- the instructions are more likely to be accurate if they match what we're using for day-to-day development.

(There's a related topic here regarding concurrently maintaining two different sets of setup documentation -- maybe we should get rid of the internal docs and make the public docs authoritative, if they have identical content.)

fisx:
> > Are you sure thought that you want me to complicate the instructions further? The steps under the link worked fine for me, so either I was lucky, or that's enough to get things working for people who don't care enough to find their own sources of information.
> 
> I'm not sure along which axis additional complexity would be introduced here. I think there's some value in having slightly more detailed documentation, even if it makes it slightly longer, given that (I assume) the target audience of these docs is not guaranteed to be _a priori_ familiar with these tools. Moreover, I think our public dev setup docs should match the internal docs, for consistency -- the instructions are more likely to be accurate if they match what we're using for day-to-day development.
> 
> (There's a related topic here regarding concurrently maintaining two different sets of setup documentation -- maybe we should get rid of the internal docs and make the public docs authoritative, if they have identical content.)

Two things: (1) I followed the instructions behind the link, and it worked immediately.  So the level of technical familiarity expected of the target audience should not be an issue.  (2) I was hoping to make a small improvement with little effort, and it's getting out of hand.  What can we do to close this issue without me having to understand the cailleach docs *and* the docs I've followed, propose a consolidated documentation that we can use in both places, and start a debate about all the things that are wrong with my proposal?  What about I merge this and link it from cailleach so we know there are two different ways of accomplishing the same thing?  (From what I understand the thing we're doing in cailleach is the manual way of doing what the cachix cli can accomplish automatically?  Again, I haven't checked, and don't have time to.)

tl;dr: if this is at least an incremental improvement and not making things strictly worse, can we please not try to improve it further?

stephen-smith:
> I followed the instructions behind the link, and it worked immediately.

As another point of reference, I already had Nix installed, and step 2 "worked", reporting no errors, but step 3 just failed with a command not found error.

fisx:
> > I followed the instructions behind the link, and it worked immediately.
> 
> As another point of reference, I already had Nix installed, and step 2 "worked", reporting no errors, but step 3 just failed with a command not found error.

Huh?  That's weird, step 2 should have made that impossible?

But I suppose that's just nix being a power tool that you need to hold right.  Still, I don't think we should include all the nix documentation potentially required for inexperienced users to catch up on to every place where we mention nix.

fisx:
@sysvinit I'm not trying to be difficult, I just don't understand what you want me to do, and why.  Feel free to make a suggestion in form of a commit, and I'll let you know what I think.  :)

sysvinit:
> tl;dr: if this is at least an incremental improvement and not making things strictly worse, can we please not try to improve it further?

"The perfect is the enemy of the good". I've expanded the bullet point about the Cachix cache to acknowledge that there's more than one way to do things; pointing out that Cachix exists and might be of use to others is better than omitting it entirely.

sysvinit (approved):

flokli:
This needs a rebase, as docs were moved around. I created an alternative PR in https://github.com/wireapp/wire-server/pull/2349.

PR #2327, opened 2022-04-27
update deps of libzauth

jschaul:
Upgrade version of libzauth dependencies, notably sodiumoxide bindings to libsodium, and fix resulting errors and warnings.

libzauth is used as part of the nginz module for authorization.

This PR was triggered by dependabot warnings.

## Checklist


libs/libzauth/libzauth-c/Cargo.lock
```diff
@@ -1,71 +1,71 @@
...
 [[package]]
 name = "serde"
-version = "1.0.66"
+version = "1.0.136"
 source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "ce31e24b01e1e524df96f1c2fdd054405f8d7376249a5110886fb4b658484789"
 
 [[package]]
 name = "sodiumoxide"
 version = "0.0.16"
```
stephen-smith, line 47: https://github.com/wireapp/wire-server/security/dependabot/1 says we need to update this to 0.2.5 or better...

mdimjasevic: I just updated it to sodiumoxide 0.2.7 in https://github.com/wireapp/wire-server/pull/2327/commits/bd82723e17d504372c51b01bde1797db276f1cb3. I also added to the repo `Cargo.lock` under `libzauth` (note we've been discussing `libzauth-c/Cargo.lock` above). I have no clue if it should be tracked; if not, I'll stop tracking it.

mdimjasevic:
@jschaul, did you say yesterday we might have to update `sodiumoxide` in some other place too?

jschaul:
We're currently using libsodium 1.0.16 [here](https://github.com/zinfra/cailleach/blob/master/wire-server-private/ci/dockerfiles/ubuntu14-nginz-deps/Dockerfile#L4), there's libsodium 1.0.18 out there, @comawill do we need to /should we update this also?

comawill (changes requested):
Good work, on updating this library and the dependencies.

Yes, we definitely should update the used libsodium, too.

Right now the build process does not execute the tests, we provide in the library, `cargo test` should be executed when building the debian files.

libs/libzauth/libzauth-c/src/lib.rs
```diff
@@ -177,7 +176,6 @@ pub extern fn zauth_token_delete(t: *mut ZauthToken) {
 }
 
 #[repr(C)]
-#[no_mangle]
 #[derive(Clone, Copy, Debug)]
```
jschaul, line 179: @comawill this is the place I'm least sure about my change: the no_mangle at the enum level led to warnings; but I'm not sure if removing them is okay to do here.

stephen-smith: I think it's fine.  I don't think enum names or enum value names are part of the C ABI.

libs/libzauth/libzauth-c/Makefile
```diff
@@ -43,7 +43,11 @@ uninstall:
 	rm -f $(PREFIX_INSTALL)/lib/libzauth.$(LIB_TYPE)
 	rm -f $(PREFIX_INSTALL)/lib/pkgconfig/libzauth.pc
 
-dist: build-release
+.PHONY: test
+test:
+	cd ../libzauth && cargo test
```
pcapriotti, line 48: Maybe add `--release` here?

libs/libzauth/libzauth-c/src/lib.rs
pcapriotti: I think `no_mangle` on structs and enums doesn't do anything, so it is safe to remove.

pcapriotti (approved):
LGTM

libs/libzauth/libzauth-c/Makefile
jschaul: Done, thanks.

libs/libzauth/libzauth-c/src/lib.rs
jschaul: thanks for your reviews!

PR #2328, opened 2022-04-28
Move old /docs to /docs/legacy (leaving references)

fisx:
We've recently merged github.com/wireapp/wire-docs into github.com/wire-server/docs, which already had some stuff in it.  This PR moves all the old stuff that was already there into `/docs/legacy` to make it more transparent what's what.

With `ln -s legacy/reference` (as in d89746f59607a730d1d4257c416058cba7cd55cf), links like [this one](https://github.com/wireapp/wire-server/blob/d89746f59607a730d1d4257c416058cba7cd55cf/docs/reference/config-options.md) won't work any more.

By linking all files individually, we would get github pages with all the information the user needs to follow the symlink, but the UX is not very nice.

So I went with keeping the files and leaving breadcrumbs inside the files.

## Checklist


smatting:
What is the long-term plan for the wire-server docs? If we want to merge them into the docs why not include reference/ and developer/ into the sphinx doc instead of moving them into a "legacy" subdir?

fisx:
> What is the long-term plan for the wire-server docs? If we want to merge them with the rest of the wire docs why not include reference/ and developer/ instead of moving them into a "legacy" subdir?

it's a smaller change.  incremental improvements are easier than "make everything perfect".  we thought about moving it to docs.wire.com, but it's not very well organized as it is, and we neither want to spend time cleaning it up nor expose it on docs.wire.com in its current form.

does that make sense?

smatting (approved):
@fisx yes makes sense to maybe not publish everything at once in its current form. Let's commit to creating PRs that move sections out of legacy/ to docs.wire.com.

PR #2329, opened 2022-04-28
[DOC] SQSERVICES-1500 Document 2nd Factor Password Challenge Team Feature

battermann:
https://wearezeta.atlassian.net/browse/SQSERVICES-1500

### DIFF
--- a/changelog.d/4-docs/pr-2329
+++ b/changelog.d/4-docs/pr-2329
@@ -1,0 +1,1 @@
+Documentation for the 2nd factor password challenge feature
--- a/docs/src/how-to/install/index.rst
+++ b/docs/src/how-to/install/index.rst
@@ -17,6 +17,7 @@ Installing wire-server
    (production) How to monitor wire-server <monitoring.rst>
    (production) How to see centralized logs for wire-server <logging.rst>
    (production) Other configuration options <configuration-options.rst>
+   Feature settings <team-feature-settings.md>
    sft
    restund
    configure-federation
--- a/docs/src/how-to/install/team-feature-settings.md
+++ b/docs/src/how-to/install/team-feature-settings.md
@@ -1,5 +1,9 @@
 # Feature settings
 
+Features can be enabled or disabled on a team level or server wide.
+
+When a feature's lock status is `unlocked` its settings can be overridden on a team level by team admins.
+
 ## 2nd factor password challenge
 
 By default Wire enforces a 2nd factor authentication for certain user operations like e.g. activating an account, changing email or password, or deleting an account.
@@ -20,5 +24,7 @@ galley:
         sndFactorPasswordChallenge:
           defaults:
             status: enabled
-            lockStatus: locked # note that the lockstatus is required and should be locked
+            lockStatus: locked
 ```
+
+Note that the lock status is required but has no effect, as it is currently not supported for team admins to enable or disable `sndFactorPasswordChallenge`. We recommend to set the lock status to `locked`.
--- a/docs/src/how-to/install/team-feature-settings.md
+++ b/docs/src/how-to/install/team-feature-settings.md
@@ -1,0 +1,24 @@
+# Feature settings
+
+## 2nd factor password challenge
+
+By default Wire enforces a 2nd factor authentication for certain user operations like e.g. activating an account, changing email or password, or deleting an account.
+
+If the `sndFactorPasswordChallenge` feature is enabled, a 6 digit verification code will be send per email to authenticate for additional user operations like e.g. for login, adding a new client, generating SCIM tokens, or deleting a team.
+
+Usually the default is what you want. If you explicitly want to enable additional password challenges, add the following to your Helm overrides in `values/wire-server/values.yaml`:
+
+```yaml
+galley:
+  # ...
+  config:
+    # ...
+    settings:
+      # ...
+      featureFlags:
+        # ...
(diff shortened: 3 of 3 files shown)


fisx (approved):

jschaul (commented):
A general comment:

Before, the config-options file in wire-server/docs took the perspective of a wire-server developer, essentially, and giving examples like:

```
.. code:: yaml
  # brig.yaml
  optSettings:
```

So the above snippet has a comment stating `brig.yaml`. What's `brig.yaml` exactly? It's the configuration file for the brig binary that, if you compile brig yourself, and run it yourself, with `./dist/brig -c path/to/brig.yaml`, will start the brig process (which may fail as it needs databases and other things to be reachable).

This is all fine for a wire-server developer / Wire employee.

But this is not the kind of configuration information anyone reading docs.wire.com under a section "how to install Wire" can make use of. For that audience, it only makes sense to document configuration options *at the helm chart value override level*, and not at the *process' configuration file level*. 

Now, for this particular example of the federation domain, that is already documented at the *helm chart override level*, here: https://docs.wire.com/how-to/install/configure-federation.html#configure-helm-charts-federator-and-ingress-and-webapp-subcharts or here: https://github.com/wireapp/wire-server/blob/develop/docs/src/how-to/install/configure-federation.rst#configure-helm-charts-federator-and-ingress-and-webapp-subcharts

As such, the federation domain text in the file you're adding here does not make a lot of sense on its own - it's decoupled from the rest of federation configuration, and it talks at a configuration level no reader can really modify - nobody actually runs `brig -c path/to/config.yaml` other than (perhaps) Wire backend engineers once per year.

A similar idea holds true for much of the rest of this document - some config options have useful descriptions and comments but may not be modifyable as such in an actual deployment - since e.g. the way to override this reuires scoping them under some additional yaml headers, or since the helm value level configs are named differently to the ones that the haskell processes take.

So in general I think this is very worthwhile work, but it's not sufficient to just move and link docs, the content has to be adpated and partially re-written, and partially deleted (as for the federation things, as those are already covered elsewhere).

Unless you have ideas of structuring docs differently, I would generally write the configuration options in the "helm value override level" perspective, i.e. the way you would structure a `myvalues.yaml` which one could use with `helm (template|install) ./charts/wire-server -f myvalues.yaml`.
Separate notes for backend engineer documentation could also be useful, but that should be very clearly marked and put into a part of a documentation that has lots of warnings about this being only useful for developers of the team.

Thanks for working on this!

jschaul (changes requested):
Requesting changes to avoid this being merged too early before addressing my comments.

battermann:
> A general comment:
> 
> Before, the config-options file in wire-server/docs took the perspective of a wire-server developer, essentially, and giving examples like:
> 
> ```
> .. code:: yaml
>   # brig.yaml
>   optSettings:
>     setFederationDomain: example.com
> ```
> 
> So the above snippet has a comment stating `brig.yaml`. What's `brig.yaml` exactly? It's the configuration file for the brig binary that, if you compile brig yourself, and run it yourself, with `./dist/brig -c path/to/brig.yaml`, will start the brig process (which may fail as it needs databases and other things to be reachable).
> 
> This is all fine for a wire-server developer / Wire employee.
> 
> But this is not the kind of configuration information anyone reading docs.wire.com under a section "how to install Wire" can make use of. For that audience, it only makes sense to document configuration options _at the helm chart value override level_, and not at the _process' configuration file level_.
> 
> Now, for this particular example of the federation domain, that is already documented at the _helm chart override level_, here: https://docs.wire.com/how-to/install/configure-federation.html#configure-helm-charts-federator-and-ingress-and-webapp-subcharts or here: https://github.com/wireapp/wire-server/blob/develop/docs/src/how-to/install/configure-federation.rst#configure-helm-charts-federator-and-ingress-and-webapp-subcharts
> 
> As such, the federation domain text in the file you're adding here does not make a lot of sense on its own - it's decoupled from the rest of federation configuration, and it talks at a configuration level no reader can really modify - nobody actually runs `brig -c path/to/config.yaml` other than (perhaps) Wire backend engineers once per year.
> 
> A similar idea holds true for much of the rest of this document - some config options have useful descriptions and comments but may not be modifyable as such in an actual deployment - since e.g. the way to override this reuires scoping them under some additional yaml headers, or since the helm value level configs are named differently to the ones that the haskell processes take.
> 
> So in general I think this is very worthwhile work, but it's not sufficient to just move and link docs, the content has to be adpated and partially re-written, and partially deleted (as for the federation things, as those are already covered elsewhere).
> 
> Unless you have ideas of structuring docs differently, I would generally write the configuration options in the "helm value override level" perspective, i.e. the way you would structure a `myvalues.yaml` which one could use with `helm (template|install) ./charts/wire-server -f myvalues.yaml`. Separate notes for backend engineer documentation could also be useful, but that should be very clearly marked and put into a part of a documentation that has lots of warnings about this being only useful for developers of the team.
> 
> Thanks for working on this!

Yes, I completely agree with this.

The reason for the PR is that we need to document how to enable the 2nd factor password challenge team feature, and we thought we could migrate the complete options-config docs at the same time. But as I see now, that this is not a good idea.

Will create a separate PR.

battermann:
rendered:
image

jschaul (approved):

jschaul (approved):

comawill (approved):

PR #2331, opened 2022-04-28
Generate/write nix derivations for all the haskell code

akshaymankar:
Major changes:
1. Instead of pinning things in `cabal.project`, we pin them in `nix/haskell-pins.nix`. This allows us to share built artefacts among the team and helps us avoid problems of compiling things with newer C libraries because cabal doesn't invalidate the built artefacts in the cabal-store.
2. Images are built using nixpkgs' `dockerTools.streamLayeredImage`. This allows us to build minimal images without needing a docker daemon to be running. This also helps us cache most of our images in the nix cache.
3. Every time any cabal file is changed or a new package is added, we must run `make regen-local-nix-derivations`. This will update various nix derivations we have for our project. This is protected by `make check-local-nix-derivations` which runs as a dependency of `make lint-all` in CI. 

Usual development workflow shouldn't change if you were already using nix provided development environment. In case you face any problems deleting one or all of these directories from the root of the project might fix your issues:
- `.env`
- `dist-newstyle`


hack/bin/generate-nix-overrides.sh
```diff
@@ -0,0 +1,48 @@
+#!/usr/bin/env bash
```
smatting, line 1: We could try to replace this with a script that reads cabal.project instead, right?

smatting (approved):

smatting:
> The build.nix at the root of the repo

This seems to be missing from the PR

akshaymankar:
> > The build.nix at the root of the repo
> 
> This seems to be missing from the PR

Added it now.

hack/bin/generate-nix-overrides.sh
akshaymankar: We could, but parsing cabal.project doesn't seem like fun. If this works well with `shellFor`, we could just get rid of those `source-package-repository` sections from cabal.project.

build.nix
```diff
@@ -0,0 +1,104 @@
...
+    executables = [
+      "brig"
+      "cannon"
+      "cargohold"
+      "federator"
+      "galley"
+      "gundeck"
+      "proxy"
+      "spar"
+    ];
```
akshaymankar, line 21: TODO: Extend these to the real list of executables.

build.nix
```diff
@@ -0,0 +1,130 @@
...
+              lens-datetime = hlib.markUnbroken (hlib.doJailbreak hsuper.lens-datetime);
+              wai-predicates = hlib.markUnbroken hsuper.wai-predicates;
+              bytestring-arbitrary = hlib.markUnbroken (hlib.doJailbreak hsuper.bytestring-arbitrary);
+              invertible = hlib.markUnbroken hsuper.invertible;
+              polysemy-check = hlib.markUnbroken (hlib.doJailbreak hsuper.polysemy-check);
+              swagger = hlib.doJailbreak externalOverrides.swagger;
+              multihash = hlib.doJailbreak externalOverrides.multihash;
+              wire-message-proto-lens = hlib.addBuildTool localOverrides.wire-message-proto-lens pkgs.protobuf;
+              types-common-journal = hlib.addBuildTool localOverrides.types-common-journal pkgs.protobuf;
+              hashable = hsuper.hashable_1_4_0_2;
```
smatting, line 49: where do the versions overriden here come from? did you arrive them by trail-and-error?

akshaymankar: Yeah, whenever I saw an error saying, hey I want a newer `<package>`, I checked in  `haskell.packages.ghc8107` whether there was some `<package>_<version>`, if I found it, I just used it.

build.nix
```diff
@@ -0,0 +1,130 @@
...
+
+    staticExecs =
+      let nested = attrsets.mapAttrs (hPkgName: execNames:
+            attrsets.genAttrs execNames (extractExec hPkgName)
+          ) executablesMap;
+          unnested = pkgs.lib.lists.foldr (x: y: x // y) {} (attrsets.attrValues nested);
+      in unnested;
+
+    images = attrsets.mapAttrs (execName: drv:
+      pkgs.dockerTools.buildImage {
```
akshaymankar, line 98: Could use `buildLayeredImage` with `maxLayers`

hack/bin/generate-local-nix-overrides.sh
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+SCRIPT_DIR=$(cd -- "$(dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
+ROOT_DIR=$(cd -- "$SCRIPT_DIR/../../" &> /dev/null && pwd)
+cabalFiles=$(find "$ROOT_DIR" -name '*.cabal' \
+                 | grep -v dist-newstyle)
```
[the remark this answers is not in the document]
stephen-smith: Might have to use `-not -path` instead, since I think `dist-newstyle` doesn't appear in the final path component ("name").  Alternatively, you could type `\( -type d -name dist-newstyle -prune \) -o ...` to avoid walking those directories at all, if you are just worried about directories with that specific name and not something like "indist-newstyler.cabal".

stephen-smith:
How much of this is generated vs. written?  I naturally recoil at treating anything generated as source code and checking it in, though I certainly understand it's the path of least resistence in some cases.

akshaymankar:
> How much of this is generated vs. written? I naturally recoil at treating anything generated as source code and checking it in, though I certainly understand it's the path of least resistence in some cases.

All the new files in the PR except `pins.yaml`, `build.nix` and the 3 shell scripts in `hack/` are generated. For all the `default.nix` files, we could use `callCabal2nix` function, but a file with list of those calls to the function for each package would either need to be generated or manually written. For all the pinned things, I think it is possible to write something with `callCabal2nix` and `fetchGit`, but it will be a little effort to do this. That said, I think this will also make things harder to debug, so I would still prefer just committing these files and have CI ensure that these files are not out of date.

pcapriotti:
A few questions, sorry if they are dumb, but I still struggle with nix things.

- Why are you generating all these nix files? Can't cabal2nix or whatever just read the cabal file dynamically and create the corresponding derivation?
- Are we dropping stackage now? Is this necessary? Couldn't we be using the same pins as before and avoid the trouble of finding a consistent choice of versions? This might become an issue in the future.
- Relatedly, why do we need the hackage portion of `pins.yaml`? Can't we get versions from the existing yaml files (i.e. `cabal.project.freeze`).

hack/bin/generate-local-nix-overrides.sh
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env bash
```
flokli, line 1: Can we put this in a less hacky location, and add a toplevel comment to all shellscripts, explaining what they do and how /when they're supossed to be called?

build.nix
```diff
@@ -0,0 +1,130 @@
+let pkgs = import ./nix;
```
flokli, line 1: Can this be moved into the `nix/` subfolder? Maybe `nix/wire-server.nix`? I wanted to do a bit of cleanup in `nix/` too, and pull in some of the `docs/` folder and the plumbing there.

Essentially have `nix/default.nix` return more than just `pkgs`, but an attrset, which could contain `nixpkgs`, and a `wire-server` sub-namespace.

dev-packages.nix
```diff
@@ -203,5 +203,6 @@ in
   # which sets LD_LIBRARY_PATH and others correctly.
   cabal-wrapper
```
flokli, line 204: Do we still need `cabal-wrapper` after this, and all the `compile-deps` in there? I'd assume it'd become less, as we only need to compile `wire-server` itself (and can use the `shellFor` stuff for that?)

libs/api-bot/default.nix
```diff
@@ -0,0 +1,25 @@
+{ mkDerivation, aeson, ansi-terminal, api-client, async, attoparsec
```
jschaul, line 1: Can all of these files have a comment stating
1. that this file is generated
2. how this file was generated
3. how to regenerate exactly this file one more time, and when this becomes necessary.

This could also be a comment linking to a longer description on why these files are needed, who calls them, how to regenerate them.

hack/bin/generate-nix-overrides.sh
```diff
@@ -0,0 +1,48 @@
...
+    else (
+      .subdirs[] | "\($basecmd) --subpath \(.)"
+    )
+    end
+EOF
+            )
+
+hackageCabal2Nix='.hackagePins[] | "cabal2nix cabal://\(.package)-\(.version) --no-check --no-haddock"'
+
+(yq -r "$gitCabal2Nix" "$pinsFile" && \
```
jschaul, line 37: Please add comments on what these commands in this file actually do, this is not easily decipherable.

smatting:
I rebased on latest `develop` and updated `pins.yaml` to track cabal.project in the rebased commit.

Currently build failures:
```
Configuring partial-isomorphisms-0.2.3.0...
  $, called at libraries/Cabal/Cabal/Distribution/Simple/Configure.hs:1024:20 in Cabal-3.2.1.0:Distribution.Simple.Configure
  configureFinalizedPackage, called at libraries/Cabal/Cabal/Distribution/Simple/Configure.hs:477:12 in Cabal-3.2.1.0:Distribution.Simple.Configure
  configure, called at libraries/Cabal/Cabal/Distribution/Simple.hs:625:20 in Cabal-3.2.1.0:Distribution.Simple
  confHook, called at libraries/Cabal/Cabal/Distribution/Simple/UserHooks.hs:65:5 in Cabal-3.2.1.0:Distribution.Simple.UserHooks
  configureAction, called at libraries/Cabal/Cabal/Distribution/Simple.hs:180:19 in Cabal-3.2.1.0:Distribution.Simple
  defaultMainHelper, called at libraries/Cabal/Cabal/Distribution/Simple.hs:116:27 in Cabal-3.2.1.0:Distribution.Simple
  defaultMain, called at Setup.hs:2:8 in main:Main
Setup: Encountered missing or private dependencies:
template-haskell >=2.17
```

.envrc
```diff
@@ -1,26 +1,35 @@
...
-# store path as a cache key. We only use this to build the dev environment. This
-# will fail building in case anything inside the devEnv is reaching outside
-# the nix/ subfolder
-store_path=$(nix-store --add ./nix/)
+# or any of the `default.nix` files change. We do this by adding all these files
+# to the nix store and using the store paths as a cache key.
+
+store_paths=$(find . -name default.nix  | grep -v '^./nix' | grep -v '^./dist-newstyle' | xargs nix-store --add ./nix)
+
+# store_paths=$(nix-store --add ./nix/)
```
flokli, line 10:
```suggestion
```

.envrc
```diff
@@ -1,26 +1,35 @@
...
-if [[ ! -d ./.env || ! -f "$layout_dir/nix-rebuild" || "$store_path" != $(< "$layout_dir/nix-rebuild" ) ]]; then
 
-  # this builds "devEnv" defined in ./nix/default.nix and outputs to ./.env
-  nix-build "$store_path" -A devEnv --out-link ./.env
+if [[ ! -d "$env_dir" || ! -f "$layout_dir/nix-rebuild" || "$store_paths" != $(< "$layout_dir/nix-rebuild" ) ]]; then
+
+  [[ -d "$env_dir" ]] || mkdir -p "$env_dir"
+
+  # this builds "wireServer.devShell" defined in ./nix/default.nix and outputs to ./.env
+  nix print-dev-env --impure --expr "(import ./nix).wireServer.devShell" > "${env_dir}/dev-shell.sh"
```
flokli, line 20: This fails for me:

```
❯ direnv allow
direnv: loading ~/dev/wire/wire-server/.envrc                                                            
./.envrc:20: ./.env/dev-shell.sh: Read-only file system
this derivation will be built:
  /nix/store/4acqazyg0sd2q2si8k9pdy5gdkkmz6xl-profile-env.drv
this path will be fetched (0.01 MiB download, 0.04 MiB unpacked):
  /nix/store/xyksslzlz3m0grffb2qy4jqgg8k7pkvb-stdenv-linux
copying path '/nix/store/xyksslzlz3m0grffb2qy4jqgg8k7pkvb-stdenv-linux' from 'https://cache.nixos.org'...
building '/nix/store/4acqazyg0sd2q2si8k9pdy5gdkkmz6xl-profile-env.drv'...
error: creating symlink from '/home/flokli/dev/wire/wire-server/.env/profile-env.tmp-1020538-1297609385' to '/nix/store/8sdg1skjhc74x379gx3hld656rg5kjy6-profile-env': Read-only file system
./.envrc:27: ./.env/dev-shell.sh: No such file or directory
direnv: export +LC_ALL
```

akshaymankar: Ah, you probably have the old `.env` lying around. I added some code to delete the symlink if it exists.

.envrc
```diff
@@ -1,26 +1,37 @@
...
```
sysvinit, line 22: I'm using nix 2.3 (from Debian Stable's nix package), which doesn't have a `print-dev-env`. Are you able to implement this bit of the script without using this nix feature, or do we need to set a minimum bound on the version of nix which we require in `wire-server`?

akshaymankar: Hmm, I copied this out of direnv's `use_nix` function. Let me see if they have some old ways of doing the same thing.

hack/bin/generate-nix-overrides.sh
akshaymankar: I deleted the file, no more `yq` abuse :smile:

libs/api-bot/default.nix
akshaymankar: All these files have comments now.

dev-packages.nix
akshaymankar: We don't, its gone.

build.nix
akshaymankar: Moved it to `nix/wire-server.nix`. The `devEnv` is not `wireServer.devShell` (because it is a nix-shell now).

fisx:
i checkout this branch out on a fresh clone.  `direnv allow` succeeded after 40 minutes.  `make c` had this to say:

```
Configuring wire-message-proto-lens-0.1.0...
[ 9 of 16] Compiling Test.Schema.Util ( test/Test/Schema/Util.hs, /home/mf/tmp/wire-server/dist-newstyle/build/x86_64-linux/ghc-8.10.7/hscim-0.3.6/t/spec/build/spec/spec-tmp/Test/Schema/Util.o, /home/mf/tmp/wire-server/dist-newstyle/build/x86_64-linux/ghc-8.10.7/hscim-0.3.6/t/spec/build/spec/spec-tmp/Test/Schema/Util.dyn_o )
[10 of 16] Compiling Test.Schema.UserSpec ( test/Test/Schema/UserSpec.hs, /home/mf/tmp/wire-server/dist-newstyle/build/x86_64-linux/ghc-8.10.7/hscim-0.3.6/t/spec/build/spec/spec-tmp/Test/Schema/UserSpec.o, /home/mf/tmp/wire-server/dist-newstyle/build/x86_64-linux/ghc-8.10.7/hscim-0.3.6/t/spec/build/spec/spec-tmp/Test/Schema/UserSpec.dyn_o )
[11 of 16] Compiling Test.Schema.ResourceSpec ( test/Test/Schema/ResourceSpec.hs, /home/mf/tmp/wire-server/dist-newstyle/build/x86_64-linux/ghc-8.10.7/hscim-0.3.6/t/spec/build/spec/spec-tmp/Test/Schema/ResourceSpec.o, /home/mf/tmp/wire-server/dist-newstyle/build/x86_64-linux/ghc-8.10.7/hscim-0.3.6/t/spec/build/spec/spec-tmp/Test/Schema/ResourceSpec.dyn_o )
[12 of 16] Compiling Test.Schema.PatchOpSpec ( test/Test/Schema/PatchOpSpec.hs, /home/mf/tmp/wire-server/dist-newstyle/build/x86_64-linux/ghc-8.10.7/hscim-0.3.6/t/spec/build/spec/spec-tmp/Test/Schema/PatchOpSpec.o, /home/mf/tmp/wire-server/dist-newstyle/build/x86_64-linux/ghc-8.10.7/hscim-0.3.6/t/spec/build/spec/spec-tmp/Test/Schema/PatchOpSpec.dyn_o )
setup: filepath wildcard 'generic-message-proto/proto/otr.proto' does not
match any files.

[13 of 16] Compiling Test.Schema.MetaSchemaSpec ( test/Test/Schema/MetaSchemaSpec.hs, /home/mf/tmp/wire-server/dist-newstyle/build/x86_64-linux/ghc-8.10.7/hscim-0.3.6/t/spec/build/spec/spec-tmp/Test/Schema/MetaSchemaSpec.o, /home/mf/tmp/wire-server/dist-newstyle/build/x86_64-linux/ghc-8.10.7/hscim-0.3.6/t/spec/build/spec/spec-tmp/Test/Schema/MetaSchemaSpec.dyn_o )
[14 of 16] Compiling Test.Schema.GroupSpec ( test/Test/Schema/GroupSpec.hs, /home/mf/tmp/wire-server/dist-newstyle/build/x86_64-linux/ghc-8.10.7/hscim-0.3.6/t/spec/build/spec/spec-tmp/Test/Schema/GroupSpec.o, /home/mf/tmp/wire-server/dist-newstyle/build/x86_64-linux/ghc-8.10.7/hscim-0.3.6/t/spec/build/spec/spec-tmp/Test/Schema/GroupSpec.dyn_o )
[15 of 16] Compiling Test.Schema.AuthenticationSchemeSpec ( test/Test/Schema/AuthenticationSchemeSpec.hs, /home/mf/tmp/wire-server/dist-newstyle/build/x86_64-linux/ghc-8.10.7/hscim-0.3.6/t/spec/build/spec/spec-tmp/Test/Schema/AuthenticationSchemeSpec.o, /home/mf/tmp/wire-server/dist-newstyle/build/x86_64-linux/ghc-8.10.7/hscim-0.3.6/t/spec/build/spec/spec-tmp/Test/Schema/AuthenticationSchemeSpec.dyn_o )
[16 of 16] Compiling Main             ( test/Spec.hs, /home/mf/tmp/wire-server/dist-newstyle/build/x86_64-linux/ghc-8.10.7/hscim-0.3.6/t/spec/build/spec/spec-tmp/Main.o, /home/mf/tmp/wire-server/dist-newstyle/build/x86_64-linux/ghc-8.10.7/hscim-0.3.6/t/spec/build/spec/spec-tmp/Main.dyn_o )
Linking /home/mf/tmp/wire-server/dist-newstyle/build/x86_64-linux/ghc-8.10.7/hscim-0.3.6/t/spec/build/spec/spec ...
[26 of 26] Compiling Wire.Swagger     ( src/Wire/Swagger.hs, /home/mf/tmp/wire-server/dist-newstyle/build/x86_64-linux/ghc-8.10.7/types-common-0.16.0/build/Wire/Swagger.o, /home/mf/tmp/wire-server/dist-newstyle/build/x86_64-linux/ghc-8.10.7/types-common-0.16.0/build/Wire/Swagger.dyn_o )
cabal: Failed to build wire-message-proto-lens-0.1.0 (which is required by
test:wire-api-tests from wire-api-0.1.0, test:wire-api-golden-tests from
wire-api-0.1.0 and others).
```

smatting:
@fisx You need to fetch submodules in wire-serer too, `git submodule update --init --recursive`

fisx:
there was a disk space issue.  with enough disk space, `make full-clean c ci` still fails in the same place.

fisx:
> git submodule update --init --recursive

that sounds helpful!  is there a way to make this automatic?  if not i'll add this trick to the PSA we already give i `make c` fails.

smatting:
> there was a disk space issue. with enough disk space, `make full-clean c ci` still fails in the same place.

What is the eror message? It should be different when you've pulled the submodules. For pulling submodules automatically I think https://stackoverflow.com/a/49427199

fisx:
> > there was a disk space issue. with enough disk space, `make full-clean c ci` still fails in the same place.
> 
> What is the eror message? It should be different when you've pulled the submodules. For pulling submodules automatically I think https://stackoverflow.com/a/49427199

trying with submodules now, hopeful this will work.

fisx:
> > git submodule update --init --recursive
> 
> that sounds helpful! is there a way to make this automatic? if not i'll add this trick to the PSA we already give i `make c` fails.

https://github.com/wireapp/wire-server/pull/2761/commits/c9acc5aa574c017973f130734c6334ba92a27300

smatting:
> direnv allow succeeded after 40 minutes

I'm wondering why it took so long. We don't build the devShell on ci I think, however all the `wireServerPackages` should be in the cache, right @akshaymankar ?

fisx:
> > direnv allow succeeded after 40 minutes
> 
> I'm wondering why it took so long. We don't build the devShell on ci I think, however all the `wireServerPackages` should be in the cache, right @akshaymankar ?

I'm not sure and I lost the terminal and no logs, but I don't remember any building.  Maybe the nix-cache server is just slow?  Haven't done the numbers to decide how plausible that is.  I'm on a good uplink, though.

fisx:
another weird thing: every shell command output is not enriched with this direnv noise:

```
~/tmp/wire-server$ ls
cabal.project	      changelog.d   charts  dist-newstyle  hack      libs     Makefile	package-defaults.yaml  SECURITY.md  shell.nix  tools
cassandra-schema.cql  CHANGELOG.md  deploy  docs	   hie.yaml  LICENSE  nix	README.md	       services     snapshots
direnv: loading ~/tmp/wire-server/.envrc
direnv: loading ~/tmp/wire-server/.env/.env/dev-shell.sh
direnv: loading ~/tmp/wire-server/.env/profile-env/.env/profile-env/.profile
./.envrc:40: warning: setlocale: LC_ALL: cannot change locale (en_US.UTF-8)
direnv: export +AR +AS +CC +CONFIG_SHELL +CXX +DETERMINISTIC_BUILD +HOST_PATH +IN_NIX_SHELL +LC_ALL +LD +LOCALE_ARCHIVE +NIX_BINTOOLS +NIX_BINTOOLS_WRAPPER_TARGET_HOST_x86_64_unknown_linux_gnu +NIX_BUILD_CORES +NIX_BUILD_TOP +NIX_CC +NIX_CC_WRAPPER_TARGET_HOST_x86_64_unknown_linux_gnu +NIX_CFLAGS_COMPILE +NIX_ENFORCE_NO_NATIVE +NIX_GHC +NIX_GHCPKG +NIX_GHC_DOCDIR +NIX_GHC_LIBDIR +NIX_HARDENING_ENABLE +NIX_INDENT_MAKE +NIX_LDFLAGS +NIX_PATH +NIX_PKG_CONFIG_WRAPPER_TARGET_HOST_x86_64_unknown_linux_gnu +NIX_STORE +NM +OBJCOPY +OBJDUMP +PKG_CONFIG +PKG_CONFIG_PATH +PYTHONHASHSEED +PYTHONNOUSERSITE +PYTHONPATH +RANLIB +READELF +SIZE +SOURCE_DATE_EPOCH +STRINGS +STRIP +XDG_DATA_DIRS +_PYTHON_HOST_PLATFORM +_PYTHON_SYSCONFIGDATA_NAME +buildInputs +builder +cmakeFlags +configureFlags +depsBuildBuild +depsBuildBuildPropagated +depsBuildTarget +depsBuildTargetPropagated +depsHostHost +depsHostHostPropagated +depsTargetTarget +depsTargetTargetPropagated +doCheck +doInstallCheck +installPhase +mesonFlags +name +nativeBuildInputs +out +outputs +patches +phases +propagatedBuildInputs +propagatedNativeBuildInputs +shell +shellHook +stdenv +strictDeps +system ~PATH ~WINDOWID
```

can this be turned off?

fisx:
```
The build process was killed (i.e. SIGKILL). The
typical reason for this is that there is not enough memory available (e.g. the
OS killed a process using lots of memory).
```

trying again, with less system noise.

*sigh*

akshaymankar:
> I'm wondering why it took so long. We don't build the devShell on ci I think, however all the `wireServerPackages` should be in the cache, right @akshaymankar ?

@fisx do you have https://wire-server.cachix.org/ in your nix substituters? 

> another weird thing: every shell command output is not enriched with this direnv noise:

This happened to me once, I deleted `.direnv` in wire-server and it went away. I will look deeper into direnv to see if this step is unnecessary.

fisx:
> ```
> The build process was killed (i.e. SIGKILL). The
> typical reason for this is that there is not enough memory available (e.g. the
> OS killed a process using lots of memory).
> ```
> 
> trying again, with less system noise.
> 
> _sigh_

forgot to copy cabal.profile.local into the fresh clone.  i'm not going to miss a single trap...  :)

akshaymankar:
> forgot to copy cabal.profile.local into the fresh clone. i'm not going to miss a single trap... :)

Oh yeah, compiling with `-O2` requires a lot of memory right :(

fisx (approved):
https://github.com/wireapp/wire-server/pull/2331#issuecomment-1274506170 fixed my last problem.  works fine for me!

hack/bin/generate-local-nix-packages.sh
```diff
@@ -0,0 +1,34 @@
...
+
+warningFile=$(mktemp)
+cat >"$warningFile" <<EOF
+# WARNING: GENERATED FILE, DO NOT EDIT.
+# This file is generated by running hack/bin/generate-local-nix-packages.sh and
+# must be regenerated whenever local packages are added or removed, or
+# dependencies are added or removed.
+EOF
+
+# cabalFiles="/home/stefan/repos/wire-server/libs/wire-api/wire-api.cabal"
```
flokli, line 18: 🐈

hack/bin/upload-hoogle-image.sh
```diff
@@ -0,0 +1,40 @@
...
+fi
+
+tmp_link_store=$(mktemp -d)
+# wireServer.hoogleImage outputs an executable which prints the image on stdout
+# when executed. This is done so we don't store large images in nix-store and
+# hence in the cache. The image is a product of other store paths which should
+# already be cached. More information:
+# https://nixos.org/manual/nixpkgs/unstable/#ssec-pkgs-dockerTools-streamLayeredImage
+image_stream_file="$tmp_link_store/image_stream"
+nix -v --show-trace -L build -f "$ROOT_DIR/nix" "wireServer.hoogleImage" -o "$image_stream_file"
```
flokli, line 34: Do we still need the `--show-trace` here?

hack/bin/upload-images.sh
```diff
@@ -0,0 +1,47 @@
...
+# This script builds all the images in wireServer.images attribute of
+# $ROOT_DIR/nix/default.nix and uploads them to the docker registry using the
+# repository name specified in the image derivation and tag specified by
+# environment variable "$DOCKER_TAG".
+#
+# If $DOCKER_USER and $DOCKER_PASSWORD are provided, the script will use them to
+# upload the images.
+#
+# This script is intended to be run by CI/CD pipelines.
+
```
flokli, line 12: This seems to do pretty much the same as upload-hooge-image. Can the scripts be unified, and expect the attributes / attribute to build as args?

nix/haskell-overrides/HaskellNet-SSL.nix
```diff
@@ -0,0 +1,23 @@
+{ mkDerivation, base, bytestring, connection, data-default
```
flokli, line 1: How are these files created? I assume they're slightly modified variants of the autogenerated ones?

Could we instead use the same overriding mechanism from nixpkgs, like doJailbreak and friends, slightly modifying the autogenerated files, rather than vendoring them in, as we do here?

In any case, that workflow of hasmell-overrides probably needs a small readme in this folder.

nix/haskell-pins.nix
```diff
@@ -0,0 +1,201 @@
...
+        url = "https://github.com/wireapp/tasty";
+        rev = "394943c7672e5ad269e5587528b7678caf3b0720";
+        sha256 = "sha256-rKvWGCLJUyKPmzeYaTj5J0VAExXYIpwJ5J2lJTcuXJI=";
+      };
+      packages = {
+        tasty-hunit = "hunit";
+      };
+    };
+  };
+  # TODO: This depends on <some-hardcoded-list-of-sha56sum-for-each-package-version-combo>, make it not depend on it.
```
flokli, line 175: Missing todo

nix/manual-overrides.nix
```diff
@@ -0,0 +1,63 @@
+{ libsodium, protobuf, snappy, hlib, mls-test-cli }:
+# TODO: make sure the things marked unbroken are actually broken.
```
flokli, line 2: Missing todo

nix/wire-server.nix
```diff
@@ -0,0 +1,271 @@
...
+      inherit pkgs;
+      name = "quay.io/wire/wire-server-ci";
+      maxLayers = 2;
+      # We don't need to push the "latest" tag, every step in CI should depend
+      # deterministically on a specific image.
+      tag = null;
+      bundleNixpkgs = false;
+      extraPkgs = commonTools ++ [pkgs.cachix];
+      nixConf = {
+        experimental-features = "nix-command flakes";
```
flokli, line 230: Do we still need this, or is a nix without flakes sufficient?

I probably still can build wire-server without enabling flakes, so i don't se why CI would need to...

nix/sources.json
```diff
@@ -1,4 +1,16 @@
 {
+    "nix": {
```
flokli, line 2: Why do we pin nix here?

nix/wire-server.nix
```diff
@@ -0,0 +1,271 @@
...
+        src = lib.cleanSourceWith {
+          src = old.src;
+          filter = path: type:
+            let baseName = baseNameOf (toString path);
+            in baseName != "dist";
+        };
+      });
+
+    gitignoreSource =
+      let gitignoreSrc = pkgs.fetchFromGitHub {
```
flokli, line 15: There should be a gitignore source vendored into nixpkgs, which we can probably use instead.

nix/wire-server.nix
```diff
@@ -0,0 +1,271 @@
...
+    };
+
+    localPackages = {enableOptimization, enableDocs, enableTests}: hsuper: hself:
+      # The default packages are expected to have optimizations and docs turned
+      # on.
+      let defaultPkgs = import ./local-haskell-packages.nix {
+            inherit gitignoreSource;
+          } hsuper hself;
+
+          # TODO: Remove
```
flokli, line 56: Open todo

nix/wire-server.nix
sysvinit: I'm running Nix 2.3 on my laptop, which predates the introduction of flakes, and Akshay's gone to some effort to ensure that the nix-ified build still works for me. I would think that we don't need the flakes feature flag here, unless there are changes in recent Nix so that features which were previously independent of flakes now depend on them being enabled.

nix/haskell-overrides/HaskellNet-SSL.nix
flokli: @akshaymankar says this needs to go away entirely, and be replaced by callCabal2Nix.

nix/wire-server.nix
```diff
@@ -0,0 +1,271 @@
...
+      pkgs.kubectl
+      pkgs.ormolu
+      pkgs.shellcheck
+      (hlib.justStaticExecutables pkgs.haskellPackages.cabal-fmt)
+    ] ++ pkgs.lib.optionals pkgs.stdenv.isLinux [
+      pkgs.skopeo
+    ];
+
+    sources = import ./sources.nix;
+    ciImage = import "${sources.nix}/docker.nix" {
```
flokli, line 220: We can use `${pkgs.nix.src}/docker.nix` here, and drop the niv pin.

nix/wire-server.nix
flokli: See `doc/functions/nix-gitignore.section.md` in nixpkgs

flokli: If we don't want to specify the path to the root .gitignore when calling `gitignoreSource`, we might also be able to use `nix-gitignore.gitignoreSourcePure` with a small hardcoded list of paths to ignore, that impure local haskell builds create in the workdir.

hack/bin/upload-hoogle-image.sh
akshaymankar: It shows easier to debug error in CI if we ever mess with our nix builds, so I would say that let's keep it.

nix/wire-server.nix
akshaymankar: `nix-gitignore.gitignoreSource` just fails if there are no `.gitignore` files in the package directory, the one we have right now traverses parent directories to figure out the gitignore. I don't want to duplicate the list by using `.gitignoreSourcePure` because then someone will forget to add things here and be surprised. So, I think its not so bad to rely on `hercules-ci/gitignore.nix`.

hack/bin/generate-local-nix-overrides.sh
akshaymankar: What a less hack location?

hack/bin/generate-local-nix-overrides.sh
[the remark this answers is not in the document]
akshaymankar: `\( -type d -name dist-newstyle -prune \) -o ...` doesn't seem to work and `-not -path` requires wildcards anyway. I think I will just leave it like this and hope nobody creates `indist-newstyler.cabal` or anything like that!

nix/wire-server.nix
akshaymankar: Then we'd have to write code to figure out relative path of the repo_root from every package. Why is using the gitignore from hercules-ci so bad?

flokli: Hmmh, I assumed we have like one expression for each wire-server component, and these files are committed into the repository, so the path is known and won't change. I wasn't aware we codegen here as well?

akshaymankar: The expression for each haskell component is generated, this changes everytime there is a change in the cabal file.

PR #2332, opened 2022-04-28
Prevent duplicated rendering of FEATURE_ENABLE_PAYMENT

supersven:
If the `FEATURE_ENABLE_PAYMENT` feature flag is set by `envVars`, do not render it again. Otherwise, we might end up in confusing cases like: The flag was set to `true`, but it's additionally rendered with a default to `false`.

I gave the `envVars` entry priority because it's used many times in *cailleach* and my gut feeling tells me that it would be confusing to override an explicitly set flag with an opposite value.

Example from k8s:

```yaml
kubectl get pod -n wire team-settings-598ddf7956-r6jfn -o yaml

...
spec:
  containers:
  - env:
    - name: NODE_PORT
      value: "8080"
    - name: APP_BASE
      value: https://teams.secusmart-column-1a.wire.link/
    - name: BACKEND_REST
      value: https://nginz-https.secusmart-column-1a.wire.link
    - name: BACKEND_WS
      value: wss://nginz-ssl.secusmart-column-1a.wire.link
    - name: FEATURE_ENABLE_PAYMENT
      value: "false"
    - name: FEATURE_ENABLE_PAYMENT
      value: "false"
...
```

## Checklist

### DIFF
--- a/changelog.d/3-bug-fixes/fix-duplicated-enable-payment-flag-in-helm-charts
+++ b/changelog.d/3-bug-fixes/fix-duplicated-enable-payment-flag-in-helm-charts
@@ -1,0 +1,4 @@
+When `config.enablePayment` and `FEATURE_ENABLE_PAYMENT` (`envVars`) were set,
+the team-settings feature flag `FEATURE_ENABLE_PAYMENT` was rendered two times.
+The new behavior is to give the `envVars` entry priority. I.e. when it's set,
+it's used instead of the `config.enablePayment` value.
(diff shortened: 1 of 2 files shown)


jschaul (approved):

fisx:
This makes sense to me.  Of course it also raises the question if we really need `envVars` in the first place.  Do I understand correctly that this is posix/shell env, and what's under `.Values.config` is from `values.yaml`?  I would strongly suggest we drop env variable handling, or at least deprecate it.  But I may be missing an important use case.

fisx (approved):

supersven:
> .

Hey @fisx ,

In my PIT branch of reality `envVars` shows up like this: https://github.com/zinfra/cailleach/blob/master/environments/secusmart-column-1a/helm_vars/wire-server/values.yaml.gotmpl#L207

It's both `values.yaml` stuff. I think `envVars` is called as it is, because it relates to the environment variables for the K8s pod of `team-settings`. So, it's env variables inside the deployed cluster (not on your local machine).

supersven:
Tested successfully.

PR #2338, opened 2022-04-29
SQSERVICES-1506-doc-1-5

battermann:
https://wearezeta.atlassian.net/browse/SQSERVICES-1506

rendered:

image

### DIFF
--- a/docs/src/how-to/install/web-app-settings.md
+++ b/docs/src/how-to/install/web-app-settings.md
@@ -44,3 +44,17 @@ webapp:
     # ...
     FEATURE_ENABLE_MEDIA_EMBEDS: "false"
 ```
+
+## Enable extra entropy
+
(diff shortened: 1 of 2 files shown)


docs/src/how-to/install/web-app-settings.md
```diff
@@ -44,3 +44,17 @@ webapp:
     # ...
     FEATURE_ENABLE_MEDIA_EMBEDS: "false"
 ```
+
+## Enable extra entropy
+
+The Wire desktop application uses system-dependent pseudo-random number generators (PRNG) as an internal entropy source when generating cryptographic keys. In certain cases it may be desired to enable externally generated entropy derived from mouse movement.
```
fisx, line 50: You probably copied all this from somewhere?  Should the source be replaced with a link?

Also: I think it's inaccurate to describe the source of entropy as *P*RNG.  These usually take truly random system events and the PRNG only runs over the data provided from those to provide more bits when needed.   What about this:

```suggestion
The Wire desktop application uses system-dependent source of random bits as an internal entropy source when generating cryptographic keys. In certain cases it may be desired to enable externally generated entropy derived from mouse movement.
```

fisx (approved):

PR #2340, opened 2022-04-29
SQSERVICES-1504-doc-1-3

battermann:
https://wearezeta.atlassian.net/browse/SQSERVICES-1504

render (outdated):

image

### DIFF
--- a/docs/src/how-to/associate/custom-backend-for-desktop-client.rst
+++ b/docs/src/how-to/associate/custom-backend-for-desktop-client.rst
@@ -60,6 +60,10 @@ Create a file named ``init.json`` and set ``customWebAppURL`` and ``proxyServerU
 
 The ``env`` setting must be set to ``CUSTOM`` for this to work.
 
+.. note:: 
+  
+    Consult your site admin to learn what goes into these settings. The value of ``customWebAppURL`` can be found `here <https://github.com/wireapp/wire-server/blob/e6aa50913cdcfde1200114787baabf7896394a2f/charts/webapp/templates/deployment.yaml#L40-L41>`_ or `resp. here <https://github.com/wireapp/wire-server/blob/e6aa50913cdcfde1200114787baabf7896394a2f/charts/webapp/values.yaml#L26>`_.  The value of ``proxyServerURL`` is your browser proxy.  It depends on the configuration of the network your client is running in.
+
 Windows
 ~~~~~~~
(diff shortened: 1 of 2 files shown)


docs/src/how-to/associate/custom-backend-for-desktop-client.rst
```diff
@@ -35,8 +38,45 @@ To create the application
...
 
 - Open a Terminal
 - Start the application with the command line arguments: `--env {URL}`, where `{URL}` is the URL of your webapp as noted down above
 
+Using an initialization file
+----------------------------
+
+By providing an initialization file the instance connection parameters plus proxy settings for the Wire desktop application can be pre-configured. This requires Wire version >= 3.27.
+
+Create a file named ``init.json`` and set ``customWebAppURL`` and ``proxyServerURL`` e.g. as follows:
```
supersven, line 51: I think it would be good to exactly describe what ``customWebAppURL`` and ``proxyServerURL`` are. How would it get them as a user or administrator?

supersven (approved):
Thanks a lot for writing this up! Yesterday, Paul had questions about exactly this topic ... So, it'll be useful :smile: 

Only two smaller remarks.

supersven:
> I am not completely sure on this documentation. I put it together from ticket description and comments. Also, I found nothing on how to configure this for linux. Please review this carefully and feel free to push updates @mythsunwind

Ups, I saw your comment too late. In this case please ignore my approval. I've only checked if the text is reasonable, not that the described procedure really works.

docs/src/how-to/associate/custom-backend-for-desktop-client.rst
fisx: | Consult your site admin to learn what goes into these settings.  The value of `customWebAppURL` can be found [here](https://github.com/wireapp/wire-server/blob/e6aa50913cdcfde1200114787baabf7896394a2f/charts/webapp/templates/deployment.yaml#L40-L41) resp. [here](https://github.com/wireapp/wire-server/blob/e6aa50913cdcfde1200114787baabf7896394a2f/charts/webapp/values.yaml#L26).  The value of `proxyServerURL` is your browser proxy.  It depends on the configuration of the network your client is running in.

fisx (approved):
I added a few details, but I'm sure this is correct!  :)

PR #2341, opened 2022-04-29
buildah-docker: use cabal instead of stack

smatting:
We recently removed `stack` from the builder images. This PR changes the buildah-based build scripts to use cabal instead of stack.

I manually tested `make buildah-docker`

## Checklist

### DIFF
--- a/hack/bin/buildah-compile.sh
+++ b/hack/bin/buildah-compile.sh
@@ -2,7 +2,7 @@
 
 # This compiles wire-server inside an ubuntu-based container based on quay.io/wire/ubuntu20-builder.
 # the tool 'buildah' is used to mount some folders in, and to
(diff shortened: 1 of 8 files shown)


hack/bin/buildah-compile.sh
```diff
@@ -2,7 +2,7 @@
 
 # This compiles wire-server inside an ubuntu-based container based on quay.io/wire/ubuntu20-builder.
 # the tool 'buildah' is used to mount some folders in, and to
-# keep the stack caches of .stack and .stack-work (renamed to avoid conflicts) for the next compilation
+# keep the stack caches of /.root/.cabal and dist-newstyle (renamed to avoid conflicts) for the next compilation
```
stephen-smith, line 5: Delete "stack" in "the stack caches".

smatting: Thanks

PR #2344, opened 2022-05-02
fixup: SQSERVICES-1507-doc-1-11

comawill:
This PR attempts to improve the feature flag documentation from #2336 on when CBR is already used and when not.


battermann (approved):
thanks!

PR #2345, opened 2022-05-02
fixup: SQSERVICES-1506-doc-1-5

comawill:
This PR intends to adjust one detail from #2338, as this flag only has an effect on windows users.

### DIFF
--- a/docs/src/how-to/install/web-app-settings.md
+++ b/docs/src/how-to/install/web-app-settings.md
@@ -47,7 +47,7 @@ webapp:
 
 ## Enable extra entropy (only on Windows)
 
-The Wire desktop application uses system-dependent source of random bits as an internal entropy source when generating cryptographic keys. In certain cases it may be desired to enable externally generated entropy derived from mouse movement. This option only affects only Windows users.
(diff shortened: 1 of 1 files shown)


docs/src/how-to/install/web-app-settings.md
```diff
@@ -45,9 +45,9 @@ webapp:
     FEATURE_ENABLE_MEDIA_EMBEDS: "false"
 ```
 
-## Enable extra entropy
+## Enable extra entropy (only on Windows)
 
-The Wire desktop application uses system-dependent source of random bits as an internal entropy source when generating cryptographic keys. In certain cases it may be desired to enable externally generated entropy derived from mouse movement.
+The Wire desktop application uses system-dependent source of random bits as an internal entropy source when generating cryptographic keys. In certain cases it may be desired to enable externally generated entropy derived from mouse movement. This option only affects only Windows users.
```
battermann, line 50:
```suggestion
The Wire desktop application uses system-dependent source of random bits as an internal entropy source when generating cryptographic keys. In certain cases it may be desired to enable externally generated entropy derived from mouse movement. This option only affects Windows users.
```

battermann (changes requested):

battermann (approved):

PR #2350, opened 2022-05-03
Drop snappy support from bonanza

pcapriotti:
Also get rid of the snappy and snappy-framing dependencies.

## Checklist


jschaul (approved):
In a few months, we can drop all of bonanza, as we don't use it in the kubernetes world. 

I tried to find out if snappy is in use anywhere, I don't think so. If this is merged, CI should do a bump commit like [this](https://github.com/zinfra/cailleach/commit/587ba2b5453c5412309d824cafa3482617ba9c10), and in theory CI should then run the logserver playbook from [here](https://bridge-ie-concourse.zinfra.io/teams/main/pipelines/hegemony/jobs/logserver/builds/484).

Feel free to merge, if something about our logging breaks on staging due to this we may need to revert.

PR #2361, opened 2022-05-05
Add ShellCheck linter target

supersven:
It lints all shell scripts that are not excluded, i.e. new ones and old ones without issues. Excluded are those that currently have issues. They should be fixed on a step-by-step basis. This conservative/careful approach is necessary due to shell scripts having neither a type checker nor tests. "Surprises" should be avoided, especially as they might
appear in stressful situations.

This PR supersedes https://github.com/wireapp/wire-server/pull/2310 . In that PR I learned that even simple looking quotes may lead to awful errors.

https://wearezeta.atlassian.net/browse/SQPIT-279

## Checklist

 - [X] If this PR changes development workflow or dependencies, they have been A) automated and B) documented under docs/developer/. All efforts have been taken to minimize development setup breakage or slowdown for co-workers.
schema`** to update the cassandra schema documentation.
 - [X] **changelog.d** contains the following bits of information ([details](https://github.com/wireapp/wire-server/blob/develop/docs/developer/changelog.md)):

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -128,6 +128,52 @@ add-license:
 	@echo ""
 	@echo "you might want to run 'make formatf' now to make sure ormolu is happy"
 
+# lint all shell scripts with ShellCheck
+# FUTUREWORK: Fix issues of the explicitly (no globbing) excluded files.
+SHELL_FILES_TO_LINT=$(shell find \
+	-not -path "./dist-newstyle/*" \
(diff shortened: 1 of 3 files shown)


akshaymankar (approved):
We should create issues to fix the ignored scripts if we really want to do the `FUTUREWORK`.
Approving since this is still an improvement.

stephen-smith (approved):

PR #2368, opened 2022-05-05
[FS-507] Support for Remote Welcome Messages

mdimjasevic:
The PR adds support for remote welcome messages. So far only local welcome messages have been supported.

## Checklist

### DIFF
--- a/libs/types-common/src/Data/Json/Util.hs
+++ b/libs/types-common/src/Data/Json/Util.hs
@@ -40,7 +40,6 @@ module Data.Json.Util
     -- * Base64
     Base64ByteString (..),
     base64Schema,
-    toBase64ByteString,
     Base64ByteStringL (..),
     base64SchemaL,
     fromBase64TextLenient,
@@ -211,9 +210,6 @@ base64SchemaN = toBase64Text .= parsedText "Base64ByteString" fromBase64Text
 
 base64Schema :: ValueSchema SwaggerDoc ByteString
 base64Schema = unnamed base64SchemaN
-
-toBase64ByteString :: ByteString -> Base64ByteString
-toBase64ByteString = Base64ByteString . B64.encode
 
 --------------------------------------------------------------------------------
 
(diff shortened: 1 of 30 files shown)


services/galley/src/Galley/API/MLS/Welcome.hs
```diff
@@ -102,5 +108,17 @@ sendLocalWelcomes con now rawWelcome lclients = do
...
+  Remote [(UserId, ClientId)] ->
+  Sem r ()
+sendRemoteWelcomes rawWelcome rClients = do
+  let req =
+        MLSWelcomeRequest
+          { mwrRawWelcome = Base64ByteString rawWelcome,
+            mwrRecipients = MLSWelcomeRecipient <$> tUnqualified rClients
+          }
+  let rpc = fedClient @'Galley @"mls-welcome" req
+  -- TODO(md): do error handling
```
smatting, line 123: @mdimjasevic what errors should we handle here?

mdimjasevic: Now that I think about it, I don't see any left. We handle errors in dereferencing a key package reference even before getting into this function, so I think there's no errors left. So feel free to drop that note unless you see any error cases we've forgotten to handle.

libs/types-common/src/Data/Json/Util.hs
```diff
@@ -212,9 +211,6 @@ base64SchemaN = toBase64Text .= parsedText "Base64ByteString" fromBase64Text
 base64Schema :: ValueSchema SwaggerDoc ByteString
 base64Schema = unnamed base64SchemaN
 
-toBase64ByteString :: ByteString -> Base64ByteString
```
smatting: So apparently it was being used :)

mdimjasevic: Oh, my bad. I used it in a test, but now we know it was a wrong thing to do. I replaced its usage with the `Base64ByteString` data constructor.

stephen-smith (approved):

PR #2374, opened 2022-05-09
SQSERVICES-377 fix email verification when external id is updated via SCIM

battermann:
https://wearezeta.atlassian.net/browse/SQSERVICES-377

## Checklist

 - [x] The **PR description** provides context as to why the change should occur and what the code contributes to that effect. This could also be a link to a JIRA ticket or a Github issue, if there is one.schema documentation.

### DIFF
--- a/libs/wire-api/src/Wire/API/User/Scim.hs
+++ b/libs/wire-api/src/Wire/API/User/Scim.hs
@@ -329,6 +329,8 @@ data ValidScimUser = ValidScimUser
   }
   deriving (Eq, Show)
 
+-- | Note that a 'SAML.UserRef' may contain an email. Even though it is possible to construct a 'ValidExternalId' from such a 'UserRef' with 'UrefOnly',
+-- this does not represent a valid 'ValidExternalId'. So in case of a 'UrefOnly', we can assume that the 'UserRef' does not contain an email.
 data ValidExternalId
   = EmailAndUref Email SAML.UserRef
   | UrefOnly SAML.UserRef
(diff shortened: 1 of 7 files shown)


libs/wire-api/src/Wire/API/User/Scim.hs
```diff
@@ -329,6 +329,8 @@ data ValidScimUser = ValidScimUser
   }
   deriving (Eq, Show)
 
+-- | Note that a 'SAML.UserRef' may contain an email. Even though it is possible to construct a 'ValidExternalId' from such a 'UserRef' with 'UrefOnly',
+-- this does not represent a valid 'ValidExternalId'. So in case of a 'UrefOnly', we can assume that the 'UserRef' does not contain an email.
```
stephen-smith, line 333: Is there as smart constructor that enforces this somewhere?  It doesn't seem like an assumption that is likely to "bite" us in the future, but it also shouldn't be expensive to have the code check to ensure we aren't recording a `UNameIDEmail` in a `URefOnly` (losslessly up-converting to using `EmailAndUref` instead).

supersven (approved):
LGTM :+1:

libs/wire-api/src/Wire/API/User/Scim.hs
fisx: I would aim for a tag at the `NameID` type instead that guarantees the existence of an email address at the type level.  A smart constructor for `UserRef` would work, too, but my hunch is that it'll leave more room for error.

Not sure we'll have the time for either, though.  If you want to try touching saml2-web-sso knock yourself out!  :)

PR #2381, opened 2022-05-10
galley: refactor withSettingsOverrides

smatting:
Refactors `withSettingsOverrides` analogous to the refactorings `withSettingsOverrides` of  `cargohold` and `gundeck`:

This PR makes `withSettingsOverrides` spawn galley in a proper http server instead of relying on `SessionT`.

Using `SessionT` has the common pitfall that *any* http calls get interpreted as calls to the application, e.g.

without this PR:
```haskell
brig <- view tsBrig
galley <- view tsGalley
withSettingsOverrides opts $ do
   callBrig brig -- this is a lie. The MonadHttp call is interepred as a SessionT call to galley (host ignored), not brig! Most likely this results in `404` that is hard to debug
   callGalley galley -- another lie, same as above
```

with this PR (real http server):
```haskell
brig <- view tsBrig
withSettingsOverrides updateOpts $ do
   galleyWithOpts <- view tsGalley
   callBrig brig -- the MonadHttp call is interpreted as a real HTTP call to brig
   callGalley galleyWithOpts -- the MonadHttp call is interpreted as a real HTTP call to `galleyWithOpts` instance
```

After applying PR we could also write helper functions in `TestM` and get the galley endpoint from the environment. E.g. `callGalley :: TestM ()`

```haskell
callGalley -- calls the unmodified galley
withSettingsOverrides updateOpts $ do
   callGalley -- calls the galley spawned with `updateOpts`
```

### Effect on CI runs
What is the overhead of spawning a http server? Lets run `time make ci package=galley`

- without this PR: `8:39.72`
- with this this PR applied: `7:43.75 total`

this is surprising... On CI the run times are the same, so I don't think this PR affects the test run duration

## Checklist

### DIFF
--- a/services/galley/test/integration/API.hs
+++ b/services/galley/test/integration/API.hs
@@ -2024,7 +2024,7 @@ postConvQualifiedFederationNotEnabled = do
   alice <- randomUser
   bob <- flip Qualified (Domain "some-remote-backend.example.com") <$> randomId
   connectWithRemoteUser alice bob
-  let federatorNotConfigured = \opts -> opts & optFederator .~ Nothing
+  let federatorNotConfigured = optFederator .~ Nothing
   withSettingsOverrides federatorNotConfigured $ do
     g <- view tsGalley
     postConvHelper g alice [bob] !!! do
@@ -2586,7 +2586,7 @@ testAddRemoteMemberFederationDisabled = do
 
   -- federator endpoint not configured is equivalent to federation being disabled
   -- This is the case on staging/production in May 2021.
-  let federatorNotConfigured = \opts -> opts & optFederator .~ Nothing
+  let federatorNotConfigured = optFederator .~ Nothing
   withSettingsOverrides federatorNotConfigured $
     postQualifiedMembers alice (remoteBob :| []) convId !!! do
       const 400 === statusCode
@@ -2607,7 +2607,7 @@ testAddRemoteMemberFederationUnavailable = do
   -- available (i.e. no service listing on that IP/port) can happen due to a
   -- misconfiguration of federator. That should give a 500.
   -- Port 1 should always be wrong hopefully.
-  let federatorUnavailable = \opts -> opts & optFederator ?~ Endpoint "127.0.0.1" 1
+  let federatorUnavailable = optFederator ?~ Endpoint "127.0.0.1" 1
   withSettingsOverrides federatorUnavailable $
     postQualifiedMembers alice (remoteBob :| []) convId !!! do
       const 500 === statusCode
--- a/services/galley/test/integration/API.hs
+++ b/services/galley/test/integration/API.hs
@@ -67,7 +67,7 @@ import Data.Time.Clock (getCurrentTime)
 import Federator.Discovery (DiscoveryFailure (..))
 import Federator.MockServer (FederatedRequest (..), MockException (..))
 import Galley.API.Mapping
-import Galley.Options (Opts, optFederator)
+import Galley.Options (optFederator)
 import Galley.Types hiding (LocalMember (..))
 import Galley.Types.Conversations.Intra
 import Galley.Types.Conversations.Members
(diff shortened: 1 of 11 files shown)


services/galley/test/integration/API.hs
```diff
@@ -2585,10 +2584,9 @@ testAddRemoteMemberFederationDisabled = do
   convId <- decodeConvId <$> postConv alice [] (Just "remote gossip") [] Nothing Nothing
   connectWithRemoteUser alice remoteBob
 
-  opts <- view tsGConf
   -- federator endpoint not configured is equivalent to federation being disabled
   -- This is the case on staging/production in May 2021.
-  let federatorNotConfigured :: Opts = opts & optFederator .~ Nothing
+  let federatorNotConfigured = \opts -> opts & optFederator .~ Nothing
```
pcapriotti, line 2589:
```suggestion
  let federatorNotConfigured = optFederator .~ Nothing
```

services/galley/test/integration/API.hs
```diff
@@ -2605,12 +2603,11 @@ testAddRemoteMemberFederationUnavailable = do
   convId <- decodeConvId <$> postConv alice [] (Just "remote gossip") [] Nothing Nothing
   connectWithRemoteUser alice remoteBob
 
-  opts <- view tsGConf
   -- federator endpoint being configured in brig and/or galley, but not being
   -- available (i.e. no service listing on that IP/port) can happen due to a
   -- misconfiguration of federator. That should give a 500.
   -- Port 1 should always be wrong hopefully.
-  let federatorUnavailable :: Opts = opts & optFederator ?~ Endpoint "127.0.0.1" 1
+  let federatorUnavailable = \opts -> opts & optFederator ?~ Endpoint "127.0.0.1" 1
```
pcapriotti, line 2610:
```suggestion
  let federatorUnavailable = optFederator ?~ Endpoint "127.0.0.1" 1
```

services/galley/test/integration/API/Util.hs
```diff
@@ -2338,14 +2327,21 @@ postSSOUser name hasEmail ssoid teamid = do
...
--- | This allows you to run requests against a galley instantiated using the given options.
---   Note that ONLY 'galley' calls should occur within the provided action, calls to other
---   services will fail.
-withSettingsOverrides :: (HasGalley m, MonadIO m, MonadMask m) => Opts.Opts -> SessionT m a -> m a
-withSettingsOverrides opts action = do
+withSettingsOverrides :: (Opts.Opts -> Opts.Opts) -> TestM a -> TestM a
+withSettingsOverrides f action = do
+  ts :: TestSetup <- ask
+  let opts = f (ts ^. tsGConf)
   (galleyApp, _, finalizer) <- liftIO $ Run.mkApp opts
```
pcapriotti, line 2334: Since the mock server runs in `Codensity IO` anyway, it might be nicer to turn `mkApp` into a `Codensity IO` action as well, instead of returning a finaliser explicitly (which is error-prone. But I guess it's fine for now.

pcapriotti (approved):
Really nice! I've been meaning to do this for some time. The same could be done in brig, right? Minor suggestions and comments follow.

services/galley/test/integration/API/Util.hs
smatting: I gave this a shot int https://github.com/wireapp/wire-server/pull/2381/commits/f6edc73ed9115822f495f73bf73ee0a53fc8617e I'm note sure if https://github.com/wireapp/wire-server/pull/2381/commits/92a5935a3856967ebf5fbab3bbe6635324267083 makes sense. Can you please have look again?

pcapriotti (commented):
Looks good, but I would change a few things here and there. I've pushed my suggestions to another branch [here](https://github.com/wireapp/wire-server/commit/b6ca760317c07d2a6eb44b852a7f77a3ee024fa1).

PR #2382, opened 2022-05-11
Adds automated linting

elland:
In the interest of keeping things automated, consistent and avoiding repetition, we want to automate linting. 

As it can be a contentious topic, this PR will serve as both a battleground for bikeshedding the linter rules as well as the de-facto linting process, whereas we incrementally apply the newly defined rules to the whole codebase.

## Checklist


tools/stern/src/Stern/API.hs
```diff
@@ -480,7 +479,7 @@ routes = do
   document "GET" "getConsentLog" $ do
     summary "Fetch the consent log given an email address of a non-user"
     notes "Relevant only internally at Wire"
-    Doc.parameter Doc.Query "email" Doc.string' $ do
```
fisx, line 483: i like redundant `do`.  also it guides ormolu in the right direction sometimes, but i forgot where.  can we ignore this as well?

fisx (approved):
i have a change request, but i'll approve anyway in the hope that you'll follow it anyway before merging.  if you disagree i'll find out how strongly i feel about this.  :)

tools/stern/src/Stern/Intra.hs
```diff
@@ -832,8 +832,8 @@ getUserNotifications uid = do
     fetchAll xs start = do
       userNotificationList <- fetchBatch start
       let batch = view queuedNotifications userNotificationList
-      if (not . null) batch && (view queuedHasMore userNotificationList)
-        then fetchAll (batch ++ xs) (Just . (view queuedNotificationId) $ last batch)
+      if (not . null) batch && view queuedHasMore userNotificationList
```
smatting, line 835: I'm always surprised that there is no alternative to `not . null`

smatting (changes requested):
+1 for keeping redundant `$` and `do`. All the other changes look like an improvement to me.
Maybe apply hlint to more code to see more examples?

tools/stern/src/Stern/Intra.hs
elland: Indeed. Also sounds like that's a hard naming problem. Any ideas?

stephen-smith: `isJust . nonEmpty` ? ;)

`isJust . uncons` ?

elland: How about `isNotEmpty = not . null`?

services/spar/src/Spar/API.hs
```diff
@@ -1,6 +1,7 @@
-{-# LANGUAGE GeneralizedNewtypeDeriving #-}
 {-# LANGUAGE RecordWildCards #-}
 {-# OPTIONS_GHC -fplugin=Polysemy.Plugin #-}
+{-# OPTIONS_GHC -Wno-unrecognised-pragmas #-}
+{-# HLINT ignore "Use $>" #-}
```
elland, line 4: For this particular case, it makes (IMO) the code harder to read, with added parens and less uniform formatting.

```haskell
authreqPrecheck msucc merr idpid =
  ( validateAuthreqParams msucc merr
      *> getIdPConfig idpid
  )
    $> NoContent
```

services/spar/src/Spar/API.hs
```diff
@@ -235,7 +236,7 @@ authreqPrecheck ::
 authreqPrecheck msucc merr idpid =
   validateAuthreqParams msucc merr
     *> getIdPConfig idpid
-    *> return NoContent
+    *> pure NoContent
```
elland, line 239: I did this by hand, but I'd like to add this as an hlint rule. Thoughts?

services/spar/src/Spar/Sem/ScimTokenStore/Mem.hs
```diff
@@ -38,4 +38,4 @@ scimTokenStoreToMem = (runState mempty .) $
     Lookup st -> gets $ M.lookup st
     LookupByTeam tid -> gets $ filter ((== tid) . stiTeam) . M.elems
     Delete tid stid -> modify $ M.filter $ \sti -> not $ stiTeam sti == tid && stiId sti == stid
-    DeleteByTeam tid -> modify $ M.filter (not . (== tid) . stiTeam)
+    DeleteByTeam tid -> modify $ M.filter ((/= tid) . stiTeam)
```
elland, line 41: Can this ever hit a `NaN` case?

services/spar/src/Spar/Scim.hs
```diff
@@ -184,5 +184,5 @@ server ::
 server conf =
   ScimSite
     { config = toServant $ Scim.configServer conf,
-      users = \authData -> toServant (Scim.userServer @tag authData)
+      users = toServant . Scim.userServer @tag
```
jschaul, line 187: I personally find the explicit arguments and anonymous functions sometimes make for more readable code. I.e. I would find it cool if writing the line `\authData -> ...` would remain allowed to write without CI barking at me in the future.
I don't mind there being code of the old version of the diff or code in the new version of the diff existing; I'm just not sure it's a good idea to always force the latter writing style. Just because hlint found out that one argument doesn't have to be explicitly written and the code still compiles (in a similar way, one could add a linter which deletes all comments - code would still compile without comments also, and there'd be fewer lines in the text file), it doesn't mean it's a useful thing to do in caring for the readability of the code for your coworkers.

There's more indirection involved when reading the code when you need to keep looking at types of only partially applied functions. Perhaps that's just me though.

Not sure what others think here?

jschaul (commented):
Generally looks like good improvements. See in-line comments.

services/spar/src/Spar/Scim.hs
smatting: Ah missted that. Yes, I agree with you. I also think this point-free style (the "point" is the argument) that hlint sometimes suggests is harder to read sometimes, extreme example: try entering `\x -> f (g x) x` to `http://pointfree.io/`. I wouldn't want to enforce it.

pcapriotti: Wars have been fought over this stuff. There's the whole idea of [concatenative languages](https://en.wikipedia.org/wiki/Concatenative_programming_language), where all you get is pointfree style, since there are no variable bindings at all.

Personally, I agree with the general sentiment expressed by the above comments, but I suspect it's mostly a matter of what people are used to, and most people are not used to pointfree style, especially when taken to an extreme.

Sometimes, though, the fact that an expression doesn't refactor well to pointfree is a sign of something messy going on, like leakage in an abstraction, or something like that, and it can be worthwhile to reorganise things to make the expression more amenable to this kind of "concatenative" style.

But I agree that it shouldn't be a rule that we just enforce all the time.

services/spar/src/Spar/Scim/Types.hs
```diff
@@ -96,4 +87,4 @@ normalizeLikeStored usr =
     tweakExtra = ScimUserExtra . RichInfo . normalizeRichInfoAssocList . unRichInfo . view sueRichInfo
 
     tweakActive :: Maybe Scim.ScimBool -> Maybe Scim.ScimBool
-    tweakActive = fmap Scim.ScimBool . maybe (Just True) Just . fmap Scim.unScimBool
+    tweakActive = fmap Scim.ScimBool . maybe (Just True) (Just . Scim.unScimBool)
```
pcapriotti, line 90: This doesn't look like an improvement, it's much harder to understand this code now. How about changing it to `Just . Scim.ScimBool . fromMaybe True . fmap Scim.unScimBool`?

services/spar/src/Spar/Sem/IdPConfigStore/Spec.hs
```diff
@@ -55,8 +55,8 @@ propsForInterpreter interpreter extract labeler lower = do
     prop "getConfig/storeConfig" $ prop_getStore (Just $ show . (() <$) . extract) lower
     prop "getConfig/getConfig" $ prop_getGet (Just $ show . ((() <$) *** (() <$)) . extract) lower
     prop "setReplacedBy/clearReplacedBy" $ prop_setClear labeler lower
-    prop "setReplacedBy/getReplacedBy" $ prop_setGet (Just $ show . (fmap (() <$)) . extract) lower
-    prop "setReplacedBy/setReplacedBy" $ prop_setSet (Just $ show . (fmap (() <$)) . extract) lower
+    prop "setReplacedBy/getReplacedBy" $ prop_setGet (Just $ show . fmap (() <$) . extract) lower
+    prop "setReplacedBy/setReplacedBy" $ prop_setSet (Just $ show . fmap (() <$) . extract) lower
```
pcapriotti, line 59: Isn't `(() <$)` the same as `void`?. I'm surprised hlint doesn't report this.

services/spar/src/Spar/Sem/ScimTokenStore/Mem.hs
pcapriotti: What do you mean? There are no floating points involved.

pcapriotti (approved):

services/spar/src/Spar/Sem/ScimTokenStore/Mem.hs
elland: Just wanted to be sure, b/c of the `/=` usage.

services/spar/src/Spar/Sem/IdPConfigStore/Spec.hs
elland: Good point. Adding as a custom rule.

jschaul (commented):
So can we have a way to ignore enforcement of point-free style in .hlint?

Also @elland the changes here don't pass ormolu (linters step on CI); which suggests that your editor isn't set up to run ormolu on file save. You can manually run `make format`, but I'd suggest to hook ormolu into your editor setup. That's documented under /docs/developer somewhere.

services/spar/src/Spar/Scim.hs
elland: `Avoid lambdas` has been disabled and left it to the programmer to decide.

services/spar/src/Spar/Scim/Types.hs
elland: Isn't that the same as…
```haskell
Just . Scim.ScimBool . maybe True Scim.unScimBool
```

.hlint.yaml
```diff
@@ -0,0 +1,9 @@
+# We need quasi quotes support.
+- arguments: [ -XQuasiQuotes, --color ]
+# Used to enforce ormolu styling. Can be revisited if we change formatters.
+- ignore: { name: Redundant $ }
+- ignore: { name: Redundant do }
+- ignore: { name: Avoid lambda }
```
jschaul, line 6:
```suggestion
# Left for the programmer to decide. See discussion at https://github.com/wireapp/wire-server/pull/2382#discussion_r871194424
- ignore: { name: Avoid lambda }
```

jschaul (approved):

services/spar/src/Spar/Scim/Types.hs
pcapriotti: Yes, that's better!

jschaul:
Could we get the previously discussed and approved changeset merged, and all follow-up changes made in a separate PR, please? It's very hard to review an ever-growing PR again and again.

elland:
@jschaul I separated them per commits, so only the latest commit needs review, but I can break it down into PRs as well.

smatting (approved):

PR #2387, opened 2022-05-11
Attempt to fix flakyness of path traversal test

pcapriotti:
Instead of generating `String` values, which might be invalid for the file system encoding in some locales, generate bytestrings directy and encode them as paths.

I hope this fixes a flakyness in the path traversal tests reported by @jschaul.

## Checklist

### DIFF
--- a/services/federator/test/unit/Test/Federator/Monitor.hs
+++ b/services/federator/test/unit/Test/Federator/Monitor.hs
@@ -21,6 +21,8 @@ import Control.Concurrent.Chan
 import Control.Exception (bracket)
 import Control.Lens (view)
 import Control.Monad.Trans.Cont
+import qualified Data.ByteString as BS
+import qualified Data.ByteString.Char8 as B8
 import qualified Data.Set as Set
 import Data.X509 (CertificateChain (..))
 import Federator.Env (TLSSettings (..), creds)
@@ -395,15 +397,26 @@ testMergeWatchedPaths =
          in mergedPaths == origPaths
     ]
 
-newtype Path = Path {getPath :: FilePath}
+newtype Path = Path {getRawPath :: ByteString}
+
+getPath :: Path -> IO FilePath
+getPath = fromRawPath . getRawPath
+
+getAbsolutePath :: Path -> IO FilePath
+getAbsolutePath p = do
+  path <- getPath p
+  makeAbsolute ("/" <> path)
 
 instance Show Path where
-  show = show . getPath
+  show = show . getRawPath
 
 instance Arbitrary Path where
-  arbitrary = Path . intercalate "/" <$> listOf (listOf1 ch)
+  arbitrary =
+    Path . B8.intercalate "/"
+      <$> listOf (BS.pack <$> listOf1 ch)
     where
-      ch = arbitrary `suchThat` (/= '/')
+      ch :: Gen Word8
+      ch = arbitrary `suchThat` (/= fromIntegral (ord '/'))
 
 trivialResolve :: FilePath -> IO (Maybe FilePath)
 trivialResolve _ = pure Nothing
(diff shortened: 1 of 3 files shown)


services/federator/test/unit/Test/Federator/Monitor.hs
```diff
@@ -395,15 +397,26 @@ testMergeWatchedPaths =
...
 
 instance Arbitrary Path where
-  arbitrary = Path . intercalate "/" <$> listOf (listOf1 ch)
+  arbitrary =
+    Path . B8.intercalate "/"
+      <$> listOf (BS.pack <$> listOf1 ch)
     where
-      ch = arbitrary `suchThat` (/= '/')
+      ch :: Gen Word8
+      ch = arbitrary `suchThat` (/= fromIntegral (ord '/'))
```
stephen-smith, line 419: Probably need to exclude `0` / ASCII `NUL`, too right?

pcapriotti: I'm not 100% sure, but I don't think so. It seems the roundtripping encoding used by GHC can handle arbitrary bytes. A 0 byte should just become a zero character, which is a valid code point, and roundtrip back to a 0 byte.

stephen-smith: ?  Then, what does Char(s)? does `0x30 :: Word8` map to and vice-versa?  Also, whatever GHC roundtripping is happening should handle `0x2F` / `'/'`, but we still exclude it...

If this is going through some other form of filtering to make sure the Characters are valid in path components, why do we need to exclude something that maps to `/`?  If this is the only such filtering / processing, then we need to exclude stuff that outputs ASCII `NUL`.

It didn't look like there was any additional processing before this was used to access to the filesystem, though...

stephen-smith: > `0x30 :: Word8` is mapped to unicode code point `\u0030`, i.e. `0`. Not sure why that would be special.

Then, shouldn't `0 :: Word8` be mapped to `'\u0000' :: Char` (aka ASCII NUL), which isn't a valid character in a file path, under most encodings.

stephen-smith (approved):

services/federator/test/unit/Test/Federator/Monitor.hs
pcapriotti: > > Isn't that what I wrote two comments ago?
> 
> You wrote "A 0 byte should just become a zero character" which I interpreted as `0 :: Word8` maps to `'0' :: Char`. 

Oh, I see the confusion now. Sorry about that.

> I'm fine with leaving it as is if the test stays reliable. I think the filtering of `'/'` paired with the `useAsCString` just had me in the mindset that we needed the byte string to be a valid path component on Linux / POSIX, as `'/'` and NUL are the two banned bytes. (`':'` is also problematic in places, because of PATH-like things.)

Alright, thanks. Let's leave it for now then. I'll still wait for confirmation from @jschaul before merging, since I think he's the only one who's actually witnessed a failure so far.

jschaul (approved):
I have problems myself reproducing this, as it doesn't always happen, and annoyingly I am unable to get the tests to fail when only running the tests in question.

Both outputs below are from `develop`, not this branch: The top pane is output from running `make` (without further arguments), which eventually runs all tests including the federator ones. This *sometimes* exhibits the problem, and *sometimes* not. The bottom pane is a manual run of the federator tests. I ran the bottom command (with and without the quickcheck seed) 100 times in a loop and it never failed. I'm not sure if there's something that increases the likelihood for this to fail when I run all unit tests instead of just the federator ones.

traversal

But I ran `make` a few times on this branch, and I still get some failures in other packages (`hscim`, `types-common`) sometimes; but not in federator anymore. So this change seems to be useful.

I'll post the offending remaining sometimes-failing tests elsewhere to be addressed in another PR.

PR #2388, opened 2022-05-12
Workaround for HLS not finding files while running template haskell

akshaymankar:
https://github.com/haskell/haskell-language-server/issues/481

## Checklist

 - [x] No Changelog


pcapriotti (approved):

elland (commented):
Not a fan of the solution, but I'm also not aware of a better one… 😰

PR #2391, opened 2022-05-12
[FS-507] Ensure Users Connected Before Exchanging Welcome Messages

mdimjasevic:
Neither a local nor a federated exchange of welcome messages ensured that users are connected before exchanging welcome messages. This can lead to spamming with welcome messages as well as exhausting all of key packages a user has uploaded. The PR ensures that users are connected before sending welcome messages.

## Checklist

### DIFF
--- a/changelog.d/2-features/mls
+++ b/changelog.d/2-features/mls
@@ -1,3 +1,4 @@
 MLS implementation progress:
 
  - remote key package claim is now supported (#2353)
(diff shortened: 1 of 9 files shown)


mdimjasevic:
After a discussion with @pcapriotti , I'm closing this PR. The request sender of a remote welcome message request can put anything in the sender ID field, and we have no way of checking on the remote end it is their ID.
Connection checks are done anyway during an add proposal.

PR #2392, opened 2022-05-12
show package name for which tests run

jschaul:
This change leads to the following new behaviour:

* show package name for the unit tests that run
* stop executing other packages' unit tests if there's a failure

doesn't warrant a changelog I think.

### DIFF
--- a/hack/bin/cabal-run-all-tests.sh
+++ b/hack/bin/cabal-run-all-tests.sh
@@ -5,12 +5,12 @@ set -euo pipefail
 DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 TOP_LEVEL="$(cd "$DIR/../.." && pwd)"
 
-packages=$(find "$TOP_LEVEL" -name '*.cabal' |
+mapfile -t packages < <(find "$TOP_LEVEL" -name '*.cabal' |
     grep -v dist-newstyle |
     xargs -n 1 dirname |
     xargs -n 1 basename)
 
-for p in $packages; do
+for p in "${packages[@]}"; do
     echo "==== Testing $p..."
     "$DIR/cabal-run-tests.sh" "$p"
 done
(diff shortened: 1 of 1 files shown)


hack/bin/cabal-run-all-tests.sh
```diff
@@ -2,11 +2,16 @@
...
+DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 TOP_LEVEL="$(cd "$DIR/../.." && pwd)"
 
-find "$TOP_LEVEL" -name '*.cabal' |
+packages=$(find "$TOP_LEVEL" -name '*.cabal' |
     grep -v dist-newstyle |
     xargs -n 1 dirname |
     xargs -n 1 basename |
-    xargs -n 1 "$DIR/cabal-run-tests.sh"
+    xargs -n 1 echo)
```
stephen-smith, line 12: Why the `echo`?

hack/bin/cabal-run-all-tests.sh
```diff
@@ -2,11 +2,16 @@
...
+
+for p in $packages; do
```
stephen-smith, line 14: Maybe the workaround[1] isn't worth it, but this will cause problems if there's ever an IFS character (e.g. SPC) in a package name.

[1] Have xargs invoke a /bin/sh subshell and use $1 there instead of $p.

akshaymankar: I think using xargs makes it not possible to stop the execution on first test failure.

stephen-smith: You can definately stop a GNU xargs run:

> If any invocation of the command exits with a status of 255, xargs will stop immediately without reading any further input.

and that's standard behavior, too:

> This sequence shall be repeated until one of the following occurs: [...] An invocation of a constructed command line returns an exit status of 255.

jschaul: I don't get the discussion; haskell package names never contain spaces, so this is a silly objection. But if that's what it takes to get a positive review on this PR that has been hanging for two weeks; then; sure, see https://github.com/wireapp/wire-server/pull/2392/commits/5e3333eb5fed5c05494f492d660215c2be3dc1f4.

stephen-smith (commented):
Sorry, it was not my intent to hold things up or require any changes.

PR #2393, opened 2022-05-13
Fix locale variables in Nix and .envrc

mdimjasevic:
Some wire-server tests would fail if the locale variables were not set properly. This PR moves setting the LANG variable into `.envrc` (and also puts `LC_ALL` there). Furthermore, there is no need to set these variables any more in `tools/ormolu.sh`.

The C locale is Debian/Ubuntu-specific; it has been removed from the Ormolu script.

See https://sourceware.org/legacy-ml/libc-alpha/2015-02/msg00247.html.

## Checklist

### DIFF
--- a/.envrc
+++ b/.envrc
@@ -24,3 +24,7 @@ PATH_add "./.env/bin"
 
 # allow local .envrc overrides
 [[ -f .envrc.local ]] && source_env .envrc.local
+
+# Locale
+export LC_ALL=en_US.UTF-8
+export LANG=en_US.UTF-8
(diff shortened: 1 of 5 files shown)


pcapriotti (approved):

mdimjasevic:
@jschaul , hopefully this fixes some test failures you have been observing locally on your machine.

mdimjasevic:
@jschaul , I just tested this locally and it looks like all the problems you report in https://wearezeta.atlassian.net/browse/SQSERVICES-1552 and https://wearezeta.atlassian.net/browse/SQSERVICES-1553 seem to be addressed by this PR.

jschaul:
Thank you for the fix! Works for me.

PR #2398, opened 2022-05-16
[FS-507] Improve a Connection Check Utility

mdimjasevic:
So far a connection checking (or belonging to a team) is done only for local users. This PR extends the utility to also cover remote users.

## Checklist


pcapriotti:
I'm not sure I like this change. The way things are right now, `ensureConnectedOrSameTeam` is called on the "local" code paths, while `ensureConnectedToRemotes` is called on the "remote" paths. So I'd rather keep them separated, instead of making the former call the latter.

mdimjasevic:
> I'm not sure I like this change. The way things are right now, `ensureConnectedOrSameTeam` is called on the "local" code paths, while `ensureConnectedToRemotes` is called on the "remote" paths. So I'd rather keep them separated, instead of making the former call the latter.

If it's just a naming concern, then I'd rename `ensureConnectedOrSameTeam` to `ensureConnectedToLocalsOrSameTeam` (or alike) for consistency and to allow a general function that can check both local and remote users.

pcapriotti:
It's not just naming. Either change the code that calls this function, so that there's no separate check for remotes, or leave things as they are. With this change, one of the two functions does both local and remote checks, while the other does only remote, and they are both used. This is not good.

mdimjasevic:
In a discussion with Paolo we came to a conclusion this might have some performance penalty. Given that it doesn't improve the code much in other regards, I am abandoning this PR.

PR #2400, opened 2022-05-17
SQSERVICES-1519 IdP for SAML that was overridden by update cannot be added again

battermann:
https://wearezeta.atlassian.net/browse/SQSERVICES-1519

## Old behavior

When an IdP was updated, the entry in the table `issuer_idp_v2` did not get overridden, but instead a new entry was made. Because the old entry still existed, adding the old IdP again failed.

## Changes

In this PR, on IdP update, old issuers in `issuer_idp_v2` will be deleted. Not sure if this is the correct fix, as those old issuers may serve some other purpose that I am not aware of?

## Checklist

### DIFF
--- a/changelog.d/3-bug-fixes/pr-2400
+++ b/changelog.d/3-bug-fixes/pr-2400
@@ -1,0 +1,1 @@
+An IdP for SAML that was overridden can be added again
--- a/changelog.d/3-bug-fixes/pr-2400
+++ b/changelog.d/3-bug-fixes/pr-2400
@@ -1,1 +1,1 @@
-An IdP for SAML that was overridden can be added again
+When an IdP issuer (aka entity ID) is updated, the old issuer was still marked as "in use".
--- a/services/spar/src/Spar/API.hs
+++ b/services/spar/src/Spar/API.hs
@@ -683,12 +683,14 @@ idpUpdateXML ::
 idpUpdateXML zusr raw idpmeta idpid = withDebugLog "idpUpdate" (Just . show . (^. SAML.idpId)) $ do
   (teamid, idp) <- validateIdPUpdate zusr idpmeta idpid
   GalleyAccess.assertSSOEnabled teamid
-  forM_ (idp ^. SAML.idpExtraInfo . wiOldIssuers) IdPConfigStore.deleteIssuer
   IdPRawMetadataStore.store (idp ^. SAML.idpId) raw
   -- (if raw metadata is stored and then spar goes out, raw metadata won't match the
-  -- structured idp config.  since this will lead to a 5xx response, the client is epected to
+  -- structured idp config.  since this will lead to a 5xx response, the client is expected to
   -- try again, which would clean up cassandra state.)
   storeIdPConfig idp
+  -- if the IdP issuer is updated, the old issuer must be removed explicitly.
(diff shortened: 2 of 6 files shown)


battermann:
When an IdP is updated, the entry in the table `issuer_idp_v2` does not get overridden, but instead a new entry is made. Because the old entry still exists, adding the old IdP again fails.

In this PR on IdP update, old issuers in `issuer_idp_v2` will be deleted. Not sure if this is the correct fix, as those old issuers may serve some other purpose that I am not aware of?

services/spar/src/Spar/API.hs
```diff
@@ -683,6 +683,7 @@ idpUpdateXML ::
 idpUpdateXML zusr raw idpmeta idpid = withDebugLog "idpUpdate" (Just . show . (^. SAML.idpId)) $ do
   (teamid, idp) <- validateIdPUpdate zusr idpmeta idpid
   GalleyAccess.assertSSOEnabled teamid
+  forM_ (idp ^. SAML.idpExtraInfo . wiOldIssuers) IdPConfigStore.deleteIssuer
   IdPRawMetadataStore.store (idp ^. SAML.idpId) raw
```
fisx, line 687: if spar crashes at the right moment, this will cause the old idp to be deleted and the new not to be introduced, which is not fixable over the rest api (let alone the UI).

I don't see a good way around this, though.  It would be really nice to have transactions and postgres and everything. but i would like to have the comment below reflect this.  (could you maybe also fix my typo there while you're at it ("epected")?)

changelog.d/3-bug-fixes/pr-2400
```diff
@@ -0,0 +1 @@
+An IdP for SAML that was overridden can be added again
```
fisx, line 1:
```suggestion
When an IdP issuer (aka entity ID) is updated, the old issuer was still marked as "in use".
```

services/spar/test-integration/Test/Spar/APISpec.hs
```diff
@@ -908,6 +908,20 @@ specCRUDIdentityProvider = do
           `shouldRespondWith` ((== 200) . statusCode)
         callIdpGet (env ^. teSpar) (Just owner) idpid
           `shouldRespondWith` ((== idpmeta') . view idpMetadata)
+      it "updates IdP metadata and creates a new IdP with the first metadata" $ do
+        env <- ask
+        (owner, _) <- call $ createUserWithTeam (env ^. teBrig) (env ^. teGalley)
```
fisx, line 913:
```suggestion
        (owner, _) <- createUserWithTeam' (env ^. teBrig) (env ^. teGalley)
```

(unless i'm missing something, which is very possible.)

fisx (approved):

services/spar/test-integration/Test/Spar/APISpec.hs
battermann: The function `createUserWithTeam'` does not exist in spar.

services/spar/src/Spar/API.hs
battermann: I moved the deletion to the end which leaves a less inconsistent state, in the case of a failure. And I added a comment.

PR #2401, opened 2022-05-17
chart/nginx-ingress-services: upgrade resources and allow fine-grained configurability

jschaul:
charts/nginx-ingress-services: 

* Allow more fine-grained control over what services are installed. (e.g. allow webapp/fakeS3 to not be installed, and federator dns name to not be set if not using federator)
* Upgrade Certificate/Issuer resources: If using [cert-manager](https://github.com/cert-manager/cert-manager), you need to have least version 1.0.0 installed (0.15.X will no longer work).

Related to https://github.com/zinfra/cailleach/pull/1079

## Checklist

### DIFF
--- a/charts/nginx-ingress-services/templates/issuer.yaml
+++ b/charts/nginx-ingress-services/templates/issuer.yaml
@@ -1,5 +1,4 @@
-{{- if .Values.tls.createIssuer -}}
-{{- if and .Values.tls.enabled .Values.tls.useCertManager -}}
+{{- if and .Values.tls.enabled .Values.tls.useCertManager .Values.tls.createIssuer -}}
 apiVersion: cert-manager.io/v1
 {{- if or (eq .Values.tls.issuer.kind "Issuer") (eq .Values.tls.issuer.kind "ClusterIssuer") }}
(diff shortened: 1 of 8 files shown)


charts/nginx-ingress-services/templates/issuer.yaml
```diff
@@ -1,3 +1,4 @@
+{{- if .Values.tls.createIssuer -}}
 {{- if and .Values.tls.enabled .Values.tls.useCertManager -}}
```
akshaymankar, line 2:
```suggestion
{{- if and .Values.tls.enabled .Values.tls.useCertManager .Values.tls.createIssuer -}}
```

charts/nginx-ingress-services/templates/issuer.yaml
```diff
@@ -24,3 +25,4 @@ spec:
             class: nginx
 {{- end }}
 {{- end -}}
+{{- end -}}
```
akshaymankar, line 28:
```suggestion
```

akshaymankar (approved):

PR #2412, opened 2022-05-18
[FS-507] Improve the Federated Welcome Message Request

mdimjasevic:
An earlier PR #2368 introduced a federated welcome message request `"mls-welcome"` that beside a raw welcome message also included the list of recipients. This is not only redundant, but it also opens ground for abuse by allowing to include recipients that are not intended to receive the welcome message. This PR fixes the issue by updating the federated request to carry only the raw welcome message. The endpoint handler decodes the welcome message and also checks for key package references.

## Checklist

### DIFF
--- a/changelog.d/4-docs/feature-schema-fix
+++ b/changelog.d/4-docs/feature-schema-fix
@@ -1,0 +1,1 @@
+Feature configs should have different swagger schema names
--- a/docs/reference/provisioning/scim-token.md
+++ b/docs/reference/provisioning/scim-token.md
@@ -1,1 +1,1 @@
-file has moved [here](../legacy/reference/provisioning/scim-token.md)
+file has moved [here](../../legacy/reference/provisioning/scim-token.md)
--- a/docs/reference/provisioning/scim-via-curl.md
+++ b/docs/reference/provisioning/scim-via-curl.md
@@ -1,1 +1,1 @@
-file has moved [here](../legacy/reference/provisioning/scim-via-curl.md)
+file has moved [here](../../legacy/reference/provisioning/scim-via-curl.md)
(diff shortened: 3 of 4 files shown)


services/galley/src/Galley/API/Federation.hs
```diff
@@ -529,26 +533,58 @@ instance
...
+        . decodeMLS'
+        $ rawWelcome
+    -- Extract only recipients local to this backend
+    rcpts <-
+      fmap fold $
+        foldQualified loc ((: []) . tUnqualified) (const mempty)
+          <$$> welcomeRecipients welcome
+    void $
+      runMessagePush loc Nothing $
+        foldMap (uncurry $ mkPush rawWelcome loc now) rcpts
```
pcapriotti, line 562: Why is this copy-pasted from `Galley.API.MLS.Welcome` instead of simply using `sendLocalWelcomes`?

services/galley/src/Galley/API/Federation.hs
```diff
@@ -529,26 +533,58 @@ instance
...
+    runProtocolError =
+      fmap (mapLeft (\(Tagged msg) -> F.MLSWelcomeResponseErrorDecodingFailed msg))
+        . runError
+
+    runDerefError ::
+      Sem (ErrorS 'MLSKeyPackageRefNotFound ': r) (Either F.MLSWelcomeResponseError ()) ->
+      Sem r (Either F.MLSWelcomeResponseError ())
+    runDerefError =
+      fmap (join . mapLeft (const F.MLSWelcomeResponseErrorRefNotFound))
+        . runError
```
pcapriotti, line 590: Could you condense this error handling into just one composition chain? See for example https://github.com/wireapp/wire-server/pull/2378/files#diff-124e9b9cb198ffe2963c75b0fd6233a4c89d690eaa0eb4c76dc80bf75fa17d8eR546-R551.

services/galley/src/Galley/API/Federation.hs
mdimjasevic: We've refactored this yesterday while pairing to reuse `sendLocalWelcomes`.

services/galley/src/Galley/API/Federation.hs
mdimjasevic: This was taken care of yesterday as well.

pcapriotti (approved):

PR #2413, opened 2022-05-18
charts/*: drop wireService label, use app= instead, add servicemonitor support

flokli:
This aligns labels a bit more with how they look like in other
deployments. In some cases, we were already setting the `app` label,
too.

There's one possible regression:
The wire-server-metrics helm chart previously configured kube-prometheus-stack to automatically scrape everything with a wireService label at port http,
path /i/metrics.

This custom configuration has been removed, and instead each chart provides the option to create `ServiceMonitor` resources, which add wire services to metric scraping that way.

## Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -360,8 +360,8 @@ kube-integration-teardown-sans-federation:
 
 .PHONY: kube-restart-%
 kube-restart-%:
-	kubectl delete pod -n $(NAMESPACE) -l wireService=$(*)
-	kubectl delete pod -n $(NAMESPACE)-fed2 -l wireService=$(*)
+	kubectl delete pod -n $(NAMESPACE) -l app=$(*)
+	kubectl delete pod -n $(NAMESPACE)-fed2 -l app=$(*)
 
 .PHONY: latest-tag
 latest-tag:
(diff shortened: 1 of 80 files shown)


flokli:
> The wire-server-metrics chart is used by the federation environments. So, if you do change this flow, please don't forget to update the federation environments.

Are we fine with just stopping to instantiate the `wire-server-metrics` chart there?

akshaymankar:
> Are we fine with just stopping to instantiate the `wire-server-metrics` chart there?

We added metrics there because it was hard for us to figure out what was wrong when things didn't work. So, I would say maybe this is still needed, but best ask people working on federation these days. /cc @smatting @pcapriotti @mdimjasevic @stephen-smith

smatting:
> So, I would say maybe this is still needed, but best ask people working on federation these days. /cc @smatting @pcapriotti @mdimjasevic @stephen-smith

I don't think we have used these metrics for any debugging yet. @supersven have you made use of them while debugging calling? If not I think it's fine to remove it.

> Once there are ServiceProbes, a default kube-prometheus stack, or grafana-agent should be able to scrape these metrics
Would this also be eventually available for the federation environments/

supersven:
> I don't think we have used these metrics for any debugging yet. @supersven have you made use of them while debugging calling? If not I think it's fine to remove it, wdyt?

Nope. Debugging calling is currently more looking at logs and reasoning about Helm charts.

flokli:
I guess we can just keep this PR open until it also adds ServiceProbes, which provides an alternative.

flokli:
I went through all services in `/services`, and checked if they expose metrics at `/i/metrics`.

For those that did, I created the necessary helm file to create a `ServiceMonitor` resource, which marks the endpoint for scraping to a prometheus-operator (or something else looking for these CRs, like `grafana-agent-operator`).

As the CRDs don't come shipped with Kubernetes out of the box, but they're usually installed while installing a monitoring operator, it's opt-in and disabled by default.

flokli:
Hmm, it seems it's not possible to modify the `spec.selector.matchLabels` field of a `Deployment` or `StatefulSet` resource, so this might require deleting and recreating these resources.

flokli:
For posterity: release notes to document the upgrade were added in https://github.com/wireapp/wire-server/pull/2472.

PR #2417, opened 2022-05-19
Added TTL query field to features API.

elland:
## Checklist

### DIFF
--- a/changelog.d/1-api-changes/added-ttl-param-to-back-office-feature-api
+++ b/changelog.d/1-api-changes/added-ttl-param-to-back-office-feature-api
@@ -1,1 +1,1 @@
-The back-office team feature API now accenpts a TTL parameter (in days), so features can be activated for a limited period.
+The back-office (aka stern) team feature API now accenpts an optional TTL parameter (in days), so features can be activated for a limited period.
--- a/changelog.d/1-api-changes/added-ttl-param-to-back-office-feature-api
+++ b/changelog.d/1-api-changes/added-ttl-param-to-back-office-feature-api
@@ -1,0 +1,1 @@
+The back-office team feature API now accenpts a TTL parameter (in days), so features can be activated for a limited period.
--- a/libs/wire-api/src/Wire/API/Routes/Public/Galley.hs
+++ b/libs/wire-api/src/Wire/API/Routes/Public/Galley.hs
@@ -1176,6 +1176,7 @@ type FeatureStatusBasePut errs featureName =
     :> "features"
     :> KnownTeamFeatureNameSymbol featureName
     :> ReqBody '[Servant.JSON] (TeamFeatureStatus 'WithoutLockStatus featureName)
+    :> QueryParam "ttl" TeamFeatureTTLValue
     :> Put '[Servant.JSON] (TeamFeatureStatus 'WithoutLockStatus featureName)
 
 -- | A type for a GET endpoint for a feature with a deprecated path
--- a/libs/wire-api/src/Wire/API/Team/Feature.hs
+++ b/libs/wire-api/src/Wire/API/Team/Feature.hs
@@ -298,6 +298,8 @@ instance FromByteString TeamFeatureTTLValue where
 instance Cass.Cql TeamFeatureTTLValue where
   ctype = Cass.Tagged Cass.IntColumn
 
+  -- Passing TTL = 0 to Cassandra removes the TTL.
+  -- It does not instantly revert back.
   fromCql (Cass.CqlInt 0) = pure TeamFeatureTTLUnlimited
   fromCql (Cass.CqlInt n) = pure . TeamFeatureTTLSeconds . fromIntegral $ n
   fromCql _ = Left "fromCql: TTLValue: CqlInt expected"
--- a/libs/wire-api/src/Wire/API/Team/Feature.hs
+++ b/libs/wire-api/src/Wire/API/Team/Feature.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE DeriveAnyClass #-}
 {-# LANGUAGE GeneralizedNewtypeDeriving #-}
 {-# LANGUAGE StrictData #-}
 
@@ -25,6 +26,7 @@ module Wire.API.Team.Feature
     TeamFeatureSelfDeletingMessagesConfig (..),
     TeamFeatureClassifiedDomainsConfig (..),
     TeamFeatureStatusValue (..),
+    TeamFeatureTTLValue (..),
     FeatureHasNoConfig,
     EnforceAppLock (..),
     KnownTeamFeatureName (..),
@@ -42,6 +44,7 @@ module Wire.API.Team.Feature
     -- * Swagger
     typeTeamFeatureName,
     typeTeamFeatureStatusValue,
+    typeTeamFeatureTTLValue,
     modelTeamFeatureStatusNoConfig,
     modelTeamFeatureStatusWithConfig,
     modelTeamFeatureAppLockConfig,
@@ -66,10 +69,11 @@ import qualified Data.Swagger as S
 import qualified Data.Swagger.Build.Api as Doc
 import qualified Data.Text as T
 import qualified Data.Text.Encoding as T
+import qualified Data.Text.Lazy as TL
 import Deriving.Aeson
 import GHC.TypeLits (Symbol)
 import Imports
-import Servant (FromHttpApiData (..))
+import Servant (FromHttpApiData (..), Proxy (..), ToHttpApiData (..))
 import Test.QuickCheck.Arbitrary (arbitrary)
 import Wire.API.Arbitrary (Arbitrary, GenericUniform (..))
 
@@ -256,6 +260,52 @@ instance HasDeprecatedFeatureName 'TeamFeatureDigitalSignatures where
 
 typeTeamFeatureName :: Doc.DataType
 typeTeamFeatureName = Doc.string . Doc.enum $ cs . toByteString' <$> [(minBound :: TeamFeatureName) ..]
+
+----------------------------------------------------------------------
+-- TeamFeatureTTLValue
+
+-- Using Word to avoid dealing with negative numbers.
+-- Ideally we would also not support zero.
+-- Currently a TTL=0 is ignored on the cassandra side.
+data TeamFeatureTTLValue
+  = TeamFeatureTTLDays Word
+  | TeamFeatureTTLUnlimited
+  deriving stock (Eq, Show, Generic)
+
+deriving instance ToJSON TeamFeatureTTLValue
+
+deriving instance FromJSON TeamFeatureTTLValue
+
+instance ToHttpApiData TeamFeatureTTLValue where
+  toQueryParam a = case a of
+    TeamFeatureTTLDays d -> T.pack . show $ d
+    TeamFeatureTTLUnlimited -> "unlimited"
+
+instance FromHttpApiData TeamFeatureTTLValue where
+  parseQueryParam = \case
+    "unlimited" -> Right TeamFeatureTTLUnlimited
+    d -> case readEither . T.unpack $ d of
+      Right d' -> Right . TeamFeatureTTLDays $ d'
+      Left e -> Left . T.pack $ e <> " oops."
+
+instance S.ToParamSchema TeamFeatureTTLValue where
+  toParamSchema _ = S.toParamSchema (Proxy @Int)
+
+instance ToByteString TeamFeatureTTLValue where
+  builder TeamFeatureTTLUnlimited = "unlimited"
+  builder (TeamFeatureTTLDays d) = (builder . TL.pack . show) d
+
+instance FromByteString TeamFeatureTTLValue where
+  parser =
+    Parser.takeByteString >>= \b ->
+      case T.decodeUtf8' b of
+        Right "unlimited" -> pure TeamFeatureTTLUnlimited
+        Right d -> pure . TeamFeatureTTLDays . read . T.unpack $ d
+        Left e -> fail $ "Invalid TeamFeatureTTLDays: " <> show e
+
+typeTeamFeatureTTLValue :: Doc.DataType
+typeTeamFeatureTTLValue =
+  Doc.int64'
 
 ----------------------------------------------------------------------
 -- TeamFeatureStatusValue
--- a/libs/wire-api/src/Wire/API/Team/Feature.hs
+++ b/libs/wire-api/src/Wire/API/Team/Feature.hs
@@ -1,4 +1,3 @@
-{-# LANGUAGE DeriveAnyClass #-}
 {-# LANGUAGE GeneralizedNewtypeDeriving #-}
 {-# LANGUAGE StrictData #-}
 
@@ -272,10 +271,6 @@ data TeamFeatureTTLValue
   | TeamFeatureTTLUnlimited
   deriving stock (Eq, Show, Generic)
   deriving (Arbitrary) via (GenericUniform TeamFeatureTTLValue)
-
-deriving instance ToJSON TeamFeatureTTLValue
-
-deriving instance FromJSON TeamFeatureTTLValue
 
 instance ToHttpApiData TeamFeatureTTLValue where
   toQueryParam = T.decodeUtf8 . toByteString'
--- a/libs/wire-api/src/Wire/API/Team/Feature.hs
+++ b/libs/wire-api/src/Wire/API/Team/Feature.hs
@@ -280,11 +280,7 @@ instance ToHttpApiData TeamFeatureTTLValue where
   toQueryParam = T.decodeUtf8 . toByteString'
 
 instance FromHttpApiData TeamFeatureTTLValue where
-  parseQueryParam = \case
-    "unlimited" -> Right TeamFeatureTTLUnlimited
-    d -> case readEither . T.unpack $ d of
-      Right d' -> Right . TeamFeatureTTLSeconds $ d'
-      Left e -> Left . T.pack $ e <> " oops."
+  parseQueryParam = maybeToEither invalidTTLErrorString . fromByteString . T.encodeUtf8
 
 instance S.ToParamSchema TeamFeatureTTLValue where
   toParamSchema _ = S.toParamSchema (Proxy @Int)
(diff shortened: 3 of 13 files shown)


libs/wire-api/src/Wire/API/Team/Feature.hs
```diff
@@ -267,6 +271,52 @@ instance HasDeprecatedFeatureName 'TeamFeatureDigitalSignatures where
...
 typeTeamFeatureName = Doc.string . Doc.enum $ cs . toByteString' <$> [(minBound :: TeamFeatureName) ..]
 
+----------------------------------------------------------------------
+-- TeamFeatureTTLValue
+
+-- Using Word to avoid dealing with negative numbers.
+-- Ideally we would also not support zero.
+-- Currently a TTL=0 is ignored on the cassandra side.
+data TeamFeatureTTLValue
+  = TeamFeatureTTLDays Word
```
elland, line 281: Do we have any natural implementation in use? Is it worth going for more type-level constriction here or documenting externally that `TTL=0` is ignored?

libs/wire-api/src/Wire/API/Team/Feature.hs
```diff
@@ -267,6 +271,52 @@ instance HasDeprecatedFeatureName 'TeamFeatureDigitalSignatures where
...
+  toQueryParam a = case a of
+    TeamFeatureTTLDays d -> T.pack . show $ d
+    TeamFeatureTTLUnlimited -> "unlimited"
+
+instance FromHttpApiData TeamFeatureTTLValue where
+  parseQueryParam = \case
+    "unlimited" -> Right TeamFeatureTTLUnlimited
+    d -> case readEither . T.unpack $ d of
+      Right d' -> Right . TeamFeatureTTLDays $ d'
+      Left e -> Left . T.pack $ e <> " oops."
```
elland, line 299: Errors are currently not surfaced, everything turns into a generic 500. Maybe we should address that. :)

services/galley/src/Galley/Cassandra/TeamFeatures.hs
```diff
@@ -96,14 +96,24 @@ setFeatureStatusNoConfig ::
...
   TeamFeatureStatus 'WithoutLockStatus a ->
+  Maybe TeamFeatureTTLValue ->
   m (TeamFeatureStatus 'WithoutLockStatus a)
-setFeatureStatusNoConfig _ tid status = do
+setFeatureStatusNoConfig _ tid status ttl = do
   let flag = tfwoStatus status
   retry x5 $ write insert (params LocalQuorum (tid, flag))
   pure status
   where
+    minutes = (* 60)
```
elland, line 106: Do we already have a way to handle this kind of stuff in the codebase?

elland (commented):
Left some open questions.

services/galley/src/Galley/Cassandra/TeamFeatures.hs
```diff
@@ -96,14 +96,24 @@ setFeatureStatusNoConfig ::
...
+    hours = minutes . (* 60)
+    days = hours . (* 24)
```
battermann, line 108: These 3 function read nicely, at the same time it's not so clear what they really do. `days` converts days to seconds, right? What about `fromDays` or `toSeconds`?

services/galley/src/Galley/API/Teams/Features.hs
```diff
@@ -545,7 +569,7 @@ setFileSharingInternal ::
     Member (Input Opts) r
   ) =>
   FeatureSetter 'TeamFeatureFileSharing r
-setFileSharingInternal = Tagged $ \tid status -> do
+setFileSharingInternal = Tagged $ \tid status ttl -> do
```
battermann, line 572: Maybe it's not very important, but something to consider, when the TTL expires, and the feature is disabled again, this will not trigger a push event. We should check if this can cause any problems, when clients do not get this info via push event right away.

battermann (commented):
Looking good
An integration would be nice
Does it make sense to use days for the TTL in stern and seconds in the galley API? In this case we can have an integration test with a very short TTL.

services/galley/src/Galley/Cassandra/TeamFeatures.hs
elland: Yeah, the naming is not great, I just copied it over from https://hackage.haskell.org/package/timespan-0.4.0.0/docs/src/Data.Time.TimeSpan.html#days :D

services/galley/src/Galley/API/Teams/Features.hs
elland: For this specific feature or any feature?

services/galley/src/Galley/Cassandra/TeamFeatures.hs
```diff
@@ -96,14 +96,24 @@ setFeatureStatusNoConfig ::
...
+
     insert :: PrepQuery W (TeamId, TeamFeatureStatusValue) ()
-    insert = fromString $ "insert into team_features (team_id, " <> statusCol @a <> ") values (?, ?)"
+    insert =
+      fromString $
+        "insert into team_features (team_id, " <> statusCol @a <> ") values (?, ?)"
+          <> case ttl of
+            Just (TeamFeatureTTLDays d) | d > 0 -> " using ttl " <> show (days d)
+            _ -> mempty
```
battermann, line 116:
```suggestion
            _ -> "USING TTL 0"
```

services/galley/src/Galley/Cassandra/TeamFeatures.hs
battermann: I am not aware

libs/wire-api/src/Wire/API/Team/Feature.hs
fisx: but this `e` should result in a 400 generated by servant, no?

in general: yes, if you find an easy way to pass on error objects from galley intact to the stern "UI", please do so.  if it's hard for any reason, please open a ticket.

libs/wire-api/src/Wire/API/Team/Feature.hs
```diff
@@ -267,6 +271,52 @@ instance HasDeprecatedFeatureName 'TeamFeatureDigitalSignatures where
...
+instance ToByteString TeamFeatureTTLValue where
+  builder TeamFeatureTTLUnlimited = "unlimited"
+  builder (TeamFeatureTTLDays d) = (builder . TL.pack . show) d
+
+instance FromByteString TeamFeatureTTLValue where
+  parser =
+    Parser.takeByteString >>= \b ->
+      case T.decodeUtf8' b of
+        Right "unlimited" -> pure TeamFeatureTTLUnlimited
+        Right d -> pure . TeamFeatureTTLDays . read . T.unpack $ d
```
fisx: Use `readEither` here, too, and `fail` if `Left`?

libs/wire-api/src/Wire/API/Team/Feature.hs
```diff
@@ -267,6 +271,52 @@ instance HasDeprecatedFeatureName 'TeamFeatureDigitalSignatures where
...
+        Left e -> fail $ "Invalid TeamFeatureTTLDays: " <> show e
+
+typeTeamFeatureTTLValue :: Doc.DataType
```
fisx, line 337: this is old swagger-1.2 code, but it looks like we still need this?

nix/default.nix
```diff
@@ -54,6 +54,7 @@ let
   # for cabal, as setting it in direnv can interfere with programs in the host
   # system, especially for non-NixOS users.
   cabal-wrapper = pkgs.writeShellScriptBin "cabal" ''
+    export NIX_BUILD_SHELL="${pkgs.bash}/bin/bash"
```
fisx, line 57: separate PR?  it's ok if you're lazy and somebody approves, just couldn't keep myself from raising an eyebrow at least.  :)

services/galley/src/Galley/API/Teams/Features.hs
```diff
@@ -604,7 +628,7 @@ setConferenceCallingInternal ::
   Members '[GundeckAccess, TeamFeatureStore, TeamStore, P.TinyLog] r =>
   FeatureSetter 'TeamFeatureConferenceCalling r
 setConferenceCallingInternal =
-  setFeatureStatusNoConfig @'TeamFeatureConferenceCalling $ \_status _tid -> pure ()
+  setFeatureStatusNoConfig @'TeamFeatureConferenceCalling $ \_status _tid _ttl -> pure ()
```
fisx, line 631: i like `_ttl` slightly better than `_`.  :)

services/galley/src/Galley/Cassandra/TeamFeatures.hs
fisx: it's local names, so names don't need to be awfully long.

services/galley/src/Galley/Cassandra/TeamFeatures.hs
fisx: i found https://github.com/wireapp/wire-server/blob/4c2dc58c75ce25da228f6e9dc1069a82e628fc26/services/brig/src/Brig/Unique.hs#L68, but it's not used exactly exhaustively...

services/galley/src/Galley/API/Internal.hs
```diff
@@ -435,11 +435,14 @@ featureAPI =
     <@> mkNamedAPI getTeamSearchVisibilityInboundInternalMulti
   where
     fs g s = fsGet g <@> fsSet s
+    fsDeprecated g s = fsGet g <@> fsSetD s
```
fisx, line 438: what's this about?

fisx (commented):
looks good so far!

services/galley/src/Galley/API/Internal.hs
fisx: ah, now i see the difference between the two definitions.  :)  if you don't want to add a comment, maybe it'll help future me if you change `fsSetD` to `fsSetDeprecated`?

changelog.d/1-api-changes/added-ttl-param-to-back-office-feature-api
```diff
@@ -0,0 +1 @@
+The back-office team feature API now accenpts a TTL parameter (in days), so features can be activated for a limited period.
```
fisx, line 1:
```suggestion
The back-office (aka stern) team feature API now accenpts an optional TTL parameter (in days), so features can be activated for a limited period.
```

libs/wire-api/src/Wire/API/Team/Feature.hs
```diff
@@ -257,6 +264,72 @@ instance HasDeprecatedFeatureName 'TeamFeatureDigitalSignatures where
...
+
+-- Using Word to avoid dealing with negative numbers.
+-- Ideally we would also not support zero.
+-- Currently a TTL=0 is ignored on the cassandra side.
+data TeamFeatureTTLValue
+  = TeamFeatureTTLSeconds Word
+  | TeamFeatureTTLUnlimited
+  deriving stock (Eq, Show, Generic)
+
+deriving instance ToJSON TeamFeatureTTLValue
```
fisx, line 278: do you even need these?  (just wondering, i don't mind them being there just in case.)

libs/wire-api/src/Wire/API/Team/Feature.hs
```diff
@@ -257,6 +264,72 @@ instance HasDeprecatedFeatureName 'TeamFeatureDigitalSignatures where
...
+    "unlimited" -> Right TeamFeatureTTLUnlimited
+    d -> case readEither . T.unpack $ d of
+      Right d' -> Right . TeamFeatureTTLSeconds $ d'
+      Left e -> Left . T.pack $ e <> " oops."
+
+instance S.ToParamSchema TeamFeatureTTLValue where
+  toParamSchema _ = S.toParamSchema (Proxy @Int)
+
+instance ToByteString TeamFeatureTTLValue where
+  builder TeamFeatureTTLUnlimited = "unlimited"
```
fisx, line 298: I sometimes like to implement `FromHttpApiData` in terms of `FromByteString` or vice versa.  Makes for more coherent error handling, for once.  You could also factor out a function that is called in both instances.  But I don't insist.

fisx (commented):
please sync with the people working on https://github.com/wireapp/wire-server/pull/2435 before merging, i'm sure there will be lots of ugly conflicts...

libs/wire-api/src/Wire/API/Routes/Public/Galley.hs
```diff
@@ -1176,6 +1176,7 @@ type FeatureStatusBasePut errs featureName =
     :> "features"
     :> KnownTeamFeatureNameSymbol featureName
     :> ReqBody '[Servant.JSON] (TeamFeatureStatus 'WithoutLockStatus featureName)
+    :> QueryParam "ttl" TeamFeatureTTLValue
```
fisx, line 1179: no time to check, but is this optional (with default "unlimited"?  if not please change it, to make sure ibis won't break over this change!

elland: Should the default be `unlimited` or 8 weeks?

libs/wire-api/src/Wire/API/Team/Feature.hs
elland: `ToBytestring` is used to serialise it for Bilge. `ToHttpApiData` on the other hand, doesn't seem necessary. 👍

libs/wire-api/src/Wire/API/Team/Feature.hs
elland: Removing one of them anyway, so no need to 🏌️‍♂️it. :)

services/galley/test/integration/API/Teams/Feature.hs
```diff
@@ -313,14 +328,139 @@ testSimpleFlag ::
...
+        -- TTL should be NULL inside cassandra
+        cassState <- view tsCass
+        liftIO $ do
+          storedTTL <- maybe Nothing runIdentity <$> Cql.runClient cassState (Cql.query1 select $ params LocalQuorum (Identity tid))
+          storedTTL `shouldBe` Nothing
+
+      half = 500000
+      seconds = 1000000
+
+  -- assertFlagForbidden $ Util.getTeamFeatureFlag feature nonMember tid
```
fisx, line 389: uncomment

tools/stern/src/Stern/API.hs
```diff
@@ -380,7 +380,7 @@ routes = do
     capture "tid"
       .&. capture "feature"
       .&. param "status"
-      .&. def (Public.teamFeatureTTLDays $ 8 * 7) (query "ttl") -- default is 8 weeks
+      .&. def Public.TeamFeatureTTLUnlimited (query "ttl") -- default is unlimited
```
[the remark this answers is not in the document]
elland: Good catch! Would be nice to have tests through Stern.

libs/wire-api/src/Wire/API/Team/Feature.hs
```diff
@@ -1,4 +1,3 @@
-{-# LANGUAGE DeriveAnyClass #-}
```
fisx: I remember that we don't want this extension, but I don't remember why.  I might be wrong.

But the json instances created weren't very nice:

```
*Wire.API.Team.Feature Data.Aeson> encode TeamFeatureTTLUnlimited
"{\"tag\":\"TeamFeatureTTLUnlimited\"}"
*Wire.API.Team.Feature Data.Aeson> encode (TeamFeatureTTLSeconds 3)
"{\"contents\":3,\"tag\":\"TeamFeatureTTLSeconds\"}"
```

Which is a nice case for adding golden tests for new types as well as keeping the old ones, come to think of it.  :)

libs/wire-api/src/Wire/API/Team/Feature.hs
```diff
@@ -298,6 +298,7 @@ instance FromByteString TeamFeatureTTLValue where
 instance Cass.Cql TeamFeatureTTLValue where
   ctype = Cass.Tagged Cass.IntColumn
 
+  fromCql (Cass.CqlInt 0) = pure TeamFeatureTTLUnlimited
```
fisx, line 301: would this have been really bad (unlimited enablement would have been disabled immediately again), or no problem (0 means "indefinite" in cassandra as well)?  either way, it should be fixed, and we should have roundtrip tests for this, and there should be an error if a cql instance does not have a test, and...

tools/stern/src/Stern/Intra.hs
```diff
@@ -551,20 +551,29 @@ setTeamFeatureFlagNoConfig ::
...
   case statusCode resp of
     200 -> pure ()
     404 -> throwE (mkError status404 "bad-upstream" "team doesnt exist")
     _ -> throwE $ responseJsonUnsafe resp
+  where
+    convert = Just . toByteString' . fromDays
+    fromMinutes = (* 60)
+    fromHours = fromMinutes . (* 60)
+    fromDays = \case
+      Public.TeamFeatureTTLSeconds s -> Public.TeamFeatureTTLSeconds . fromHours . (* 24) $ s
```
fisx, line 575: This still looks wrong.  We're parsing something that we call seconds, and then we interpret it as days.  Introduce a type `TeamFeatureTTLDays` here?  I know it's annoying, so I'll leave it to you to decide whether it's worth it.

fisx (approved):
Found a few more details.  Please take a look at my changes and comments.  No further review needed if you're happy.

tools/stern/src/Stern/Intra.hs
elland: I went back and forth on this a few times but never got to a point I was happy with. We end up with either a "mirror" type, that does Unlimited/Days and converts to Unlimited/Seconds, or an extra, used only once constructor for Days that we can't disambiguate on parsing when receiving it via the API call.

I'm open to suggestions, but maybe this would be easier if/when we servantify stern?

fisx: maybe the solution would be something like this:

```haskell
data TTL = TTLSecs Int | TTLUnlimited

parse "3s" = TTLSecs 3
parse "1d" = TTLSecs $ 1 * 60 * 60 * 24
...
```

anyway, yes, let's keep it as is for now.

fisx (approved):

PR #2418, opened 2022-05-19
SQSERVICES-1561 add lock status to all features as hard wired default plus optionall server yaml and or db default unlocked

battermann:
## Checklist

### DIFF
--- a/libs/galley-types/src/Galley/Types/Teams.hs
+++ b/libs/galley-types/src/Galley/Types/Teams.hs
@@ -38,7 +38,6 @@ module Galley.Types.Teams
     flagTeamFeatureSndFactorPasswordChallengeStatus,
     flagTeamFeatureSearchVisibilityInbound,
     Defaults (..),
-    unDefaults,
     FeatureSSO (..),
     FeatureLegalHold (..),
     FeatureTeamSearchVisibility (..),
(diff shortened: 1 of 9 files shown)


libs/galley-types/src/Galley/Types/Teams.hs
```diff
@@ -220,15 +219,15 @@ data FeatureFlags = FeatureFlags
   { _flagSSO :: !FeatureSSO,
```
fisx, line 219: I think there is some redundancy between this type and `AllFeatureFlags` now, right?  Can we move this one to `wire-api` and use it instead of the other one?

I also want to make the type polymorphic in `WithLockStatus`.  Not sure we need the flexibility, but it would be nice to express more clearly that the value is the same throughout all the record fields.  Not sure it's a good idea, though.  If you do it for my reason, please explain in haddocks.

libs/galley-types/test/unit/Test/Galley/Types.hs
```diff
@@ -90,6 +91,8 @@ instance Arbitrary FeatureFlags where
       <*> arbitrary
       <*> arbitrary
       <*> arbitrary
-      <*> arbitrary
+      -- the default lock status is implicitly added on deserialization and ignored on serialization, therefore we need to fix it to the default here
+      -- we will be able to remove this once the lock status is explicitly included in the config
+      <*> (arbitrary <&> \status -> status {Public.tfwoapsLockStatus = Public.Unlocked})
```
fisx, line 96: :+1:

services/galley/src/Galley/API/Teams/Features.hs
```diff
@@ -367,9 +367,9 @@ getValidateSAMLEmailsInternal ::
   ( Member TeamFeatureStore r,
     Member (Input Opts) r
   ) =>
-  FeatureGetter 'WithoutLockStatus 'TeamFeatureValidateSAMLEmails r
+  FeatureGetter 'WithLockStatus 'TeamFeatureValidateSAMLEmails r
 getValidateSAMLEmailsInternal =
-  Tagged $ getFeatureStatusWithDefaultConfig @'TeamFeatureValidateSAMLEmails flagsTeamFeatureValidateSAMLEmailsStatus . mbTeam
+  Tagged $ getFeatureStatusWithDefaultConfigAndImplicitLockStatus @'TeamFeatureValidateSAMLEmails flagsTeamFeatureValidateSAMLEmailsStatus . mbTeam
```
fisx, line 372: I don't understand.  I thought the point was to get rid of the concern where the lock status is coming from, and always have one at this point.  Can we somehow get rid of `getFeatureStatusWithDefaultConfig`?  If so, I would use that name, but I don't feel strongly about it.  Long names are helpful!  :)

libs/galley-types/src/Galley/Types/Teams.hs
battermann: That would be nice indeed, and it would work for SSO and SearchVilisibility, but AFAICT not easily for LegalHold as this config has more states than just enabled and disabled. Maybe we can convert it to a team features with config, though. I will try.

services/galley/src/Galley/API/Teams/Features.hs
battermann: The question is if we want to update the DB schema with the lock status for all feature even when we do not want to support changing it. If that is the case, then I agree, otherwise we need to handle the case separately.

battermann:
this is obsolete with https://github.com/wireapp/wire-server/pull/2435

PR #2420, opened 2022-05-23
Linted brig.

elland:
Mostly `s/return/pure`, but a few interesting suggestions hidden in between.


jschaul (approved):
The return->pure change is really annoying to review, as it's one big commit. Personally I don't see the benefit of replacing return with pure, but I don't want to bikeshed too much. This changeset is fine, feel free to merge.

PR #2422, opened 2022-05-24
Save environment to file when entering direnv

pcapriotti:
This can be useful for scripts that have to run within the direnv environment, to avoid having to evaluate .envrc every time the script is run, which can be noticeably slow.

## Checklist

### DIFF
--- a/.envrc
+++ b/.envrc
@@ -28,3 +28,6 @@ PATH_add "./.env/bin"
 # Locale
 export LC_ALL=en_US.UTF-8
 export LANG=en_US.UTF-8
+
+# save environment variables
+env > .vars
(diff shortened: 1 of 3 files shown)


smatting (approved):

flokli:
I'm not sure I like this. This introduces yet another loose file that might get out of sync. Where do we run scripts "with the direnv environment", but without already being in direnv?

pcapriotti:
I'm using this to run `ormolu` from `vim`:

```
if [[ -f ".vars" ]]; then
  eval <<< .vars
fi
```
This takes around 10ms. For comparison, running
```
eval "$(direnv stdlib)"
source .envrc 2>/dev/null
```
Takes around 80ms.

But nevermind, I can just put this in `.envrc.local`. Closing.

PR #2445, opened 2022-06-01
Email localization when provisioning users with SCIM.

elland:
We now account for the preferred language field in the SCIM spec, and fallback to it in case of locale being undefined.

Currently untested.

## Checklist

### DIFF
--- a/services/spar/src/Spar/Scim/User.hs
+++ b/services/spar/src/Spar/Scim/User.hs
@@ -271,7 +271,7 @@ validateScimUser' errloc midp richInfoLimit user = do
     either err pure $ Brig.mkUserName (Scim.displayName user) veid
   richInfo <- validateRichInfo (Scim.extra user ^. ST.sueRichInfo)
   let active = Scim.active user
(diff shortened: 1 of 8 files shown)


services/spar/src/Spar/Scim/User.hs
```diff
@@ -271,7 +271,8 @@ validateScimUser' errloc midp richInfoLimit user = do
     either err pure $ Brig.mkUserName (Scim.displayName user) veid
   richInfo <- validateRichInfo (Scim.extra user ^. ST.sueRichInfo)
   let active = Scim.active user
-  pure $ ST.ValidScimUser veid handl uname richInfo (maybe True Scim.unScimBool active)
+  lang <- maybe (error "TODO: scim error") pure $ mapM parseLanguage $ Scim.preferredLanguage user
```
stephen-smith, line 274: How hard to go ahead and address this TODO in this PR?

fisx: yes definitely, that's just a hole to be filled before the PR is merged.  i suppose the PR should not be in "ready for review" state yet...  :)

stephen-smith (approved):

PR #2450, opened 2022-06-02
[Polysemy] Effects Around the User Query Effect

mdimjasevic:
## Checklist


libs/brig-types/src/Brig/Types/Common.hs
```diff
@@ -111,3 +121,74 @@ instance FromJSON ExcludedPrefix where
...
+  showsPrec _ = shows . emailKeyUniq
+
+instance Eq EmailKey where
+  (EmailKey k _) == (EmailKey k' _) = k == k'
+
+-- | Turn an 'Email' into an 'EmailKey'.
+--
+-- The following transformations are performed:
+--
+--   * Both local and domain parts are forced to lowercase to make
```
stephen-smith, line 144: While it isn't common, the local part is "domain dependent" and can be case-sensitive.  We probably shouldn't case fold it by default, though we can if we know the domain does handle the local part case-insensitively.

Case-folding by default will make for some very hard to trace issues if the domain does use case-sensitive local parts.

libs/brig-types/src/Brig/Types/Common.hs
```diff
@@ -111,3 +121,74 @@ instance FromJSON ExcludedPrefix where
...
+--     e-mail addresses fully case-insensitive.
+--   * "+" suffixes on the local part are stripped unless the domain
```
stephen-smith, line 146: This is also far from universal.  If we are modeling off of GMail we would also need to strip all `.` characters.

mdimjasevic: This is not new code added by me in this PR; rather, it's old code I moved around.

libs/brig-types/src/Brig/Types/Common.hs
mdimjasevic: Just like the rest in this module, this is not new code added by me in the PR; rather, it's just being reorganised and moved from other modules. That said, I don't think this PR should be about addressing your concern and I believe you should open a ticket and bring it to the team's attention.

PR #2458, opened 2022-06-08
Lint remaining services

elland:
And that's the lest of /services. Next, tools and libs!

## Checklist

### DIFF
--- a/services/spar/migrate-data/src/Spar/DataMigration/V1_ExternalIds.hs
+++ b/services/spar/migrate-data/src/Spar/DataMigration/V1_ExternalIds.hs
@@ -186,11 +186,14 @@ sink = go
                 logDebug ("No team for user " <> show uid <> " from extid " <> show extid)
           NewExternalId (tid, extid, uid) ->
             lift $
-              dryRun <$> askMigEnv >>= \case
-                DryRun -> pure ()
-                NoDryRun ->
-                  runSpar $
-                    write insert (params LocalQuorum (tid, extid, uid))
+              askMigEnv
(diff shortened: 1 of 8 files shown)


services/spar/migrate-data/src/Spar/DataMigration/V1_ExternalIds.hs
```diff
@@ -186,11 +186,14 @@ sink = go
...
-                  runSpar $
-                    write insert (params LocalQuorum (tid, extid, uid))
+              askMigEnv
+                >>= ( \case
+                        DryRun -> pure ()
+                        NoDryRun ->
+                          runSpar $
+                            write insert (params LocalQuorum (tid, extid, uid))
+                    )
+                  . dryRun
```
fisx, line 196: I strongly disagree.  what does hlint say if we put brackets around `dryRun <$> askMigEnv` in the original code?

fisx (approved):
see comment, i'd like to see this changed into something not more obfuscating, but less.  but i'll leave it to the author to second-guess me.

services/spar/migrate-data/src/Spar/DataMigration/V1_ExternalIds.hs
elland: Yeah, I'm not sold on this either… A few suggestions seem to be counter-productive when we have code like this.

elland: Same, redundant `<$>`. How about extracting the case match?

PR #2459, opened 2022-06-08
Lint remaining tools

elland:
Almost done!

## Checklist

### DIFF
--- a/tools/bonanza/src/Bonanza/Parser/Time.hs
+++ b/tools/bonanza/src/Bonanza/Parser/Time.hs
@@ -41,13 +41,13 @@ tai64N = do
   nano <- take 8 >>= int
   let t = posixSecondsToUTCTime $ posix secs nano
       l = (-1) * fromIntegral (leapSeconds t)
-  return $ addUTCTime l t
+  pure $ addUTCTime l t
   where
     int :: ByteString -> Parser Integer
     int bs =
       either
         (const $ fail "not a hexadecimal number")
-        return
+        pure
         (parseOnly hexadecimal bs)
     posix :: Integer -> Integer -> POSIXTime
     posix secs nano =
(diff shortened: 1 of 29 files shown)


tools/bonanza/src/Bonanza/Parser/Time.hs
```diff
@@ -88,20 +89,20 @@ commonLogDate = do
...
-    month "Apr" = return 4
-    month "May" = return 5
-    month "Jun" = return 6
-    month "Jul" = return 7
-    month "Aug" = return 8
-    month "Sep" = return 9
-    month "Oct" = return 10
-    month "Nov" = return 11
-    month "Dec" = return 12
+    month "Jan" = pure 1
```
elland, line 94: itching to simplify this function. 😅

fisx: how can it be simpler than this?

tools/bonanza/test/unit/Test/Bonanza/Arbitrary.hs
```diff
@@ -478,8 +477,7 @@ mkSockInput df rec msg =
           mconcat
             [ "|",
               BC.intercalate ","
-                . map (\(k, v) -> BC.intercalate "=" [k, v])
-                . map (encodeUtf8 *** encodeUtf8)
+                . map ((\(k, v) -> BC.intercalate "=" [k, v]) . (encodeUtf8 *** encodeUtf8))
```
fisx, line 480: i disagree with this change, but i let hlint overrule me.  :)

fisx (approved):

tools/bonanza/test/unit/Test/Bonanza/Arbitrary.hs
elland: About the line break, or about removing `map`? We can restore the line break, I think that's more on ormolu.

tools/bonanza/src/Bonanza/Parser/Time.hs
elland: I'm not sure if simpler, but one idea would be using a sum type and extracting the `pure` to remove repetition. Just idle thoughts, not worth pursuing, I think.

PR #2471, opened 2022-06-10
fix(docs): use if-block to support older version of sphinx when adding the logo

comawill:
It looks like the sphinx version in our docker container is older than the one being used to generate the documentation with sphinx.

Prior to this change the logo was only added correctly locally but not on the data ending up on docs.wire.com

### DIFF
--- a/docs/src/_templates/layout.html
+++ b/docs/src/_templates/layout.html
@@ -2,7 +2,11 @@
 
 {% block sidebartitle %}
     <a href="{{ pathto(master_doc) }}"> 
+    {%- if logo %}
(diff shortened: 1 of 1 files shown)


Veki301:
LGTM, except I can't give you a passing review becfause of permissions because i'm external contractor :)

fisx:
@Veki301 and @comawill, could you two agree on who's right?  https://github.com/wireapp/wire-server/pull/2452/files  :)

fisx:
ah, i see.  this is not a full revert.  i suppose you already agree, then.  thanks!

fisx (approved):

PR #2479, opened 2022-06-14
Make all feature configs overridable

battermann:
All feature configs like guest links e.g. can now be overridden in the helm configuration, so that they can be disabled/enabled and configured server wide.

## Checklist


smatting (approved):
I checked again

- that configmap can be read by galley
- that the default in values in the values.yaml correspond to the hardcoded fallback default values

PR #2486, opened 2022-06-15
Fix handling of creator client in MLS conversations

pcapriotti:
When an MLS conversation is created, it implicitly contains the creator client as a member. This was not reflected in the database. This PR fixes the problem as follows:

 - it requires a `creator_client` field when a conversation is created, and uses it to record the creator client as a member of the conversation
 - it associates the key package ref of the creator to that client when the first commit is processed.

This required changing the conversation creation endpoint `POST /conversations`.  The change is only visible in v2.

## Checklist


stefanwire (approved):

flokli:
This accidentially reverted the submodule bump in `nginx-module-vts`.

flokli:
PR to undo in https://github.com/wireapp/wire-server/pull/2566.

PR #2489, opened 2022-06-15
[docs] release notes for 4.12 4.13 4.14

battermann:


arthurwolf:
Does this "make docs" compile without errors? A screenshot would be nice. If you don't have the time tell me and I'll do it myself.

arthurwolf:
this compiles with a warning, but I don't think the warning is caused by *this* PR, so I'm approving this, and I'll need to look into the warning later

arthurwolf (approved):
approved

Screenshot 2022-06-15 at 15-15-19 Release notes of helm charts — Wire 0 0 4 documentation

battermann:
Yes, compiles with a warning that is caused by missing wire logo, and it is not caused by this PR
the additions were so big that I thought a screenshot would not help a lot, but I will add one next time anyway

PR #2492, opened 2022-06-16
SQSERVICES-1529 Swagger docs and endpoint deprecation for API version V2

battermann:
https://wearezeta.atlassian.net/browse/SQSERVICES-1529

The API changes in this PR were also added here: https://wearezeta.atlassian.net/wiki/spaces/ENGINEERIN/pages/603095166/API+changes+v1+v2

## Checklist

   - [x] A file with the changelog entry in one or more suitable sub-sections. The sub-sections are marked by directories inside `changelog.d`.r?

### DIFF
--- a/libs/wire-api/src/Wire/API/Routes/Public/Galley.hs
+++ b/libs/wire-api/src/Wire/API/Routes/Public/Galley.hs
@@ -1120,16 +1120,16 @@ type FeatureAPI =
     :<|> FeatureStatusPut '() SndFactorPasswordChallengeConfig
     :<|> AllFeatureConfigsUserGet
     :<|> AllFeatureConfigsTeamGet
-    :<|> FeatureConfigDeprecatedGet "This endpoint was removed in iOS in version 3.101. It is not used by team management, or webapp as of June 2022" LegalholdConfig
+    :<|> FeatureConfigDeprecatedGet "This endpoint was removed in iOS in version 3.101. It is not used by team management, or webapp, and is potentially used by the old Android client as of June 2022" LegalholdConfig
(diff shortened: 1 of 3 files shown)


libs/wire-api/src/Wire/API/Routes/Public/Galley.hs
```diff
@@ -1179,9 +1179,16 @@ type FeatureStatusBasePutPublic errs featureConfig =
...
 -- | A type for a GET endpoint for a feature with a deprecated path
-type FeatureStatusBaseDeprecatedGet featureConfig =
+type FeatureStatusBaseDeprecatedGet desc featureConfig =
   ( Summary
       (AppendSymbol "[deprecated] Get config for " (FeatureSymbol featureConfig))
+      -- :> Until 'V2
+      :> Description
+           ( "Deprecated. Please use `GET /teams/:tid/features/"
+               `AppendSymbol` FeatureSymbol featureConfig
+               `AppendSymbol` "` instead."
```
fisx, line 1189:
```suggestion
               `AppendSymbol` "` instead.\n"
```

fisx (approved):
Looks good to merge to me!

The only two clients of internal end-points outside wire-server/services are ibis and stern.  We could check these, document the findings, and that would allow us to remove the deprecated internal end-points soon(ish).

@dkovacevic what internal end-points are you using for feature configuration

battermann:
@fisx by adding `Until 'V2` to the endpoint types they will no longer be published to the current swagger UI. I think it is good to keep the documentation anyway and I also added everything here: https://wearezeta.atlassian.net/wiki/spaces/ENGINEERIN/pages/603095166/API+changes+v1+v2

PR #2495, opened 2022-06-20
Use New ConversationId for MLS Welcome Message

stephen-smith:
https://wearezeta.atlassian.net/browse/FS-532

## Checklist

### DIFF
--- a/services/galley/src/Galley/API/Federation.hs
+++ b/services/galley/src/Galley/API/Federation.hs
@@ -705,7 +705,6 @@ mlsSendWelcome ::
   Members
     '[ BrigAccess,
        Error InternalError,
-       ErrorS 'MLSKeyPackageRefNotFound,
        GundeckAccess,
        Input (Local ()),
        Input UTCTime
@@ -720,10 +719,13 @@ mlsSendWelcome _origDomain (fromBase64ByteString . F.unMLSWelcomeRequest -> rawW
   welcome <- either (throw . InternalErrorWithDescription . LT.fromStrict) pure $ decodeMLS' rawWelcome
   -- Extract only recipients local to this backend
   rcpts <-
-    traverse
-      ( resolveMember
-          . gsNewMember
-      )
+    fmap catMaybes
+      . traverse
+        ( fmap hush
+            . runError @(Tagged 'MLSKeyPackageRefNotFound ())
+            . resolveMember
+            . gsNewMember
+        )
       $ welSecrets welcome
   let lrcpts = qualifyAs loc $ fst $ partitionQualified loc $ map (uncurry qTrip) rcpts
   sendLocalWelcomes Nothing now rawWelcome lrcpts
--- a/services/galley/src/Galley/API/Federation.hs
+++ b/services/galley/src/Galley/API/Federation.hs
@@ -33,13 +33,11 @@ import Data.Qualified
 import Data.Range (Range (fromRange))
 import qualified Data.Set as Set
 import Data.Singletons (SingI (..), demote, sing)
-import Data.Tagged
 import qualified Data.Text.Lazy as LT
 import Data.Time.Clock
 import Galley.API.Action
 import Galley.API.Error
 import Galley.API.MLS.GroupInfo
-import Galley.API.MLS.KeyPackage
 import Galley.API.MLS.Message
 import Galley.API.MLS.Removal
 import Galley.API.MLS.Welcome
@@ -67,8 +65,7 @@ import Polysemy.Internal.Kind (Append)
 import Polysemy.Resource
 import Polysemy.TinyLog
 import qualified Polysemy.TinyLog as P
-import Servant (ServerT)
-import Servant.API
+import Servant
 import qualified System.Logger.Class as Log
 import Wire.API.Connection
 import Wire.API.Conversation hiding (Member)
@@ -85,7 +82,6 @@ import Wire.API.Federation.API.Galley
 import qualified Wire.API.Federation.API.Galley as F
 import Wire.API.Federation.Error
 import Wire.API.MLS.CommitBundle
-import Wire.API.MLS.Credential
 import Wire.API.MLS.Message
 import Wire.API.MLS.PublicGroupState
 import Wire.API.MLS.Serialisation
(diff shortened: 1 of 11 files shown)


services/galley/src/Galley/Intra/Client.hs
```diff
@@ -183,6 +187,28 @@ getClientByKeyPackageRef ref = do
...
+  r <-
+    call Brig $
+      method GET
+        . paths ["i", "mls", "key-packages", toByteString' $ toUrlPiece ref, "conversation"]
+        . expectStatus (flip elem [200, 404])
+  if statusCode (responseStatus r) == 200
+    then Just <$> parseResponse (ErrorCall . LText.unpack) r
+    else pure Nothing
+
+-- | Calls 'Brig.API.Internal.getConvIdByKeyPackageRef'.
```
mdimjasevic, line 202: The copy-pasta left-over: update the reference so it's _put_ instead of _get_. While at it, can you also please fix a typo in the `Wire.API.Routes.Internal.Brig` module where on line 210: there's this?
```haskell
  RespondEmpty 204 "Converstaion associated"                                                                                                                       
```

services/galley/src/Galley/Intra/Client.hs
```diff
@@ -183,6 +187,28 @@ getClientByKeyPackageRef ref = do
...
+getConvIdByKeyPackageRef :: KeyPackageRef -> App (Maybe (Qualified ConvId))
+getConvIdByKeyPackageRef ref = do
+  r <-
+    call Brig $
+      method GET
+        . paths ["i", "mls", "key-packages", toByteString' $ toUrlPiece ref, "conversation"]
+        . expectStatus (flip elem [200, 404])
+  if statusCode (responseStatus r) == 200
+    then Just <$> parseResponse (ErrorCall . LText.unpack) r
```
mdimjasevic, line 199: Why `ErrorCall` here? In other analogous places we've used `mkError status502 "server-error"` instead.

services/galley/src/Galley/API/MLS/Welcome.hs
```diff
@@ -78,33 +87,34 @@ welcomeRecipients ::
...
 welcomeRecipients =
-  traverse
-    ( fmap cidQualifiedClient
-        . derefKeyPackage
-        . gsNewMember
-    )
-    . welSecrets
+  traverse (either throw pure <=< resolveMember . gsNewMember) . welSecrets
+
+resolveMember :: Member BrigAccess r => KeyPackageRef -> Sem r (Either (Tagged 'MLSKeyPackageRefNotFound ()) (Qualified (UserId, ClientId), Qualified ConvId))
```
mdimjasevic, line 94: Can you please change the type of the function such that the value on the left of the return type is instead an error effect?

services/galley/src/Galley/API/MLS/Welcome.hs
```diff
@@ -78,33 +87,34 @@ welcomeRecipients ::
...
+resolveMember kpref = runError $ do
+  clientIdentity <- derefKeyPackage kpref
+  -- TODO Generate new + Put conversaion id?
+  let defConv = Qualified (selfConv (ciUser clientIdentity)) (ciDomain clientIdentity)
+  convId <- fromMaybe defConv <$> getConvIdByKeyPackageRef kpref
+  pure (cidQualifiedClient clientIdentity, convId)
```
mdimjasevic, line 100: This seems odd (and has a TODO note). For a given key package ref, the `mls_key_package_refs` table in Brig has the corresponding qualified conversation ID, doesn't it?

services/galley/src/Galley/API/Federation.hs
```diff
@@ -626,15 +622,18 @@ mlsSendWelcome _origDomain (fromBase64ByteString . F.unMLSWelcomeRequest -> rawW
...
+            . resolveMember
             . gsNewMember
         )
         $ welSecrets welcome
-  let lrcpts = qualifyAs loc $ fst $ partitionQualified loc rcpts
+  let lrcpts = qualifyAs loc $ fst $ partitionQualified loc $ map (uncurry qTrip) rcpts
   sendLocalWelcomes Nothing now rawWelcome lrcpts
   pure EmptyResponse
+  where
+    qTrip qc conv = qc {qUnqualified = (u, c, conv)}
```
mdimjasevic, line 634: You define this function twice in the module, so I'd suggest to make it top-level and module-local.

mdimjasevic (commented):
Can you please walk me through the PR in a pairing session? I don't understand everything you did and as you said, there are test case failures in CI so we could take a look together.

services/galley/src/Galley/API/MLS/Welcome.hs
stephen-smith: Not necessarily.  The conversation id fields are optional for that table.  After claiming a key package ref, a conversation id can be added to the claim, but it's not required in order to claim.  That's why the relevant brig internal servant endpoint returns a Maybe.

services/galley/src/Galley/API/Federation.hs
stephen-smith: Okay.  It really needs a better name, too.

services/galley/src/Galley/Intra/Client.hs
stephen-smith: Looks like a common typo; fixed 5-6 other occurrences, too.

stephen-smith:
@pcapriotti Could you give this a look over to see if I'm even on the right track?

> The client posts its Welcome messages in the context of a specific qualified conversation ID. (This would require an endpoint change.)

-- https://wearezeta.atlassian.net/wiki/spaces/~377173175/pages/605291411/Backend+MLS+Encrypted+Commit+and+Proposal+handling

Should I roll that endpoint change into this effort FS-532 ?  @rohan-wire

It seems like that would address some of the defaulting issues in my resolveMember, since I'd have the conversation id "in hand".

pcapriotti:
> @pcapriotti Could you give this a look over to see if I'm even on the right track?

Yes, it all seems correct, from what I can see. I guess you're mostly missing integration tests.

> > The client posts its Welcome messages in the context of a specific qualified conversation ID. (This would require an endpoint change.)
> 
> -- https://wearezeta.atlassian.net/wiki/spaces/~377173175/pages/605291411/Backend+MLS+Encrypted+Commit+and+Proposal+handling
> 
> Should I roll that endpoint change into this effort FS-532 ? @rohan-wire
> 
> It seems like that would address some of the defaulting issues in my resolveMember, since I'd have the conversation id "in hand".

That proposal is outdated. Please keep referring to https://wearezeta.atlassian.net/wiki/spaces/CORE/pages/553418860/MLS+implementation+in+wire-server, I've been trying to keep it up to date with the outcome of recent discussions. That endpoint change was discussed some time ago, but AFAIK we haven't reached a consensus.

fisx:
> That proposal is outdated.

I updated the outdated document rather radically.  Please let me know @pcapriotti or @rohan-wire if I should change it back, or change it differently.

PR #2497, opened 2022-06-20
Fixed flakiness of a test by adding a clean-up step at the end.

elland:
As the code comment hints at, we have a very narrow window between a user being created and the account being suspended inside the test environment. For a good reason too. We don't want to run a test that takes too long to check that account suspension works. But a suspended account can no longer be updated, rendering this particular test flaky, especially on subsequent test suite runs.

This should address that issue but immediately cleaning up after itself.

## Checklist

### DIFF
--- a/services/brig/test/integration/API/User/Account.hs
+++ b/services/brig/test/integration/API/User/Account.hs
@@ -875,6 +875,12 @@ testEmailUpdate brig aws = do
   -- we want to use a non-trusted domain in order to verify profile changes
   flip initiateUpdateAndActivate uid =<< mkEmailRandomLocalSuffix "test@example.com"
   flip initiateUpdateAndActivate uid =<< mkEmailRandomLocalSuffix "test@example.com"
+
+  -- adding a clean-up step seems to avoid the subsequent failures.
+  -- If subsequent runs start failint, it's possible that the aggressive setting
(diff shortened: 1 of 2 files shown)


services/brig/test/integration/API/User/Account.hs
```diff
@@ -875,6 +875,12 @@ testEmailUpdate brig aws = do
   -- we want to use a non-trusted domain in order to verify profile changes
   flip initiateUpdateAndActivate uid =<< mkEmailRandomLocalSuffix "test@example.com"
   flip initiateUpdateAndActivate uid =<< mkEmailRandomLocalSuffix "test@example.com"
+
+  -- adding a clean-up step seems to avoid the subsequent failures.
+  -- If subsequent runs start failint, it's possible that the aggressive setting
```
battermann, line 880:
```suggestion
  -- If subsequent runs start failing, it's possible that the aggressive setting
```

battermann (approved):
looks good
it's a really small change, but I think we still want a changelog entry

mdimjasevic (approved):

mdimjasevic:
Today I noticed the same test being flaky so thank you for taking care of it, @elland !

fisx (approved):

PR #2501, opened 2022-06-21
[FS-705] Fix Repeated MLS Public Keys

mdimjasevic:
This PR fixes a bug where all clients, including non-MLS clients, would be associated with the same MLS public key in the response to `GET /clients`.

Tracked by https://wearezeta.atlassian.net/browse/FS-705.

## Checklist

### DIFF
--- a/services/brig/test/integration/API/User/Client.hs
+++ b/services/brig/test/integration/API/User/Client.hs
@@ -308,10 +308,15 @@ testListClients brig = do
   let (pk1, lk1) = (somePrekeys !! 0, (someLastPrekeys !! 0))
   let (pk2, lk2) = (somePrekeys !! 1, (someLastPrekeys !! 1))
   let (pk3, lk3) = (somePrekeys !! 2, (someLastPrekeys !! 2))
-  c1 <- responseJsonMaybe <$> addClient brig uid (defNewClient PermanentClientType [pk1] lk1)
-  c2 <- responseJsonMaybe <$> addClient brig uid (defNewClient PermanentClientType [pk2] lk2)
-  c3 <- responseJsonMaybe <$> addClient brig uid (defNewClient TemporaryClientType [pk3] lk3)
-  let cs = sortBy (compare `on` clientId) $ catMaybes [c1, c2, c3]
+  c1 <- responseJsonError =<< addClient brig uid (defNewClient PermanentClientType [pk1] lk1)
+  c2 <- responseJsonError =<< addClient brig uid (defNewClient PermanentClientType [pk2] lk2)
+  c3 <- responseJsonError =<< addClient brig uid (defNewClient TemporaryClientType [pk3] lk3)
+
+  let pks = Map.fromList [(Ed25519, "random")]
+  void $ putClient brig uid (clientId c1) pks
+  let c1' = c1 {clientMLSPublicKeys = pks}
+  let cs = sortBy (compare `on` clientId) [c1', c2, c3]
+
   get
     ( brig
         . path "clients"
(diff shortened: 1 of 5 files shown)


mdimjasevic:
I just added a related test that checks that in presence of a client with MLS public keys, a non-MLS client still has no MLS public keys when looked up via `GET /clients/:client`.

pcapriotti (approved):

services/brig/test/integration/API/User/Client.hs
```diff
@@ -308,10 +309,15 @@ testListClients brig = do
   let (pk1, lk1) = (somePrekeys !! 0, (someLastPrekeys !! 0))
```
elland, line 309: hlint will soon complain about not using `head` here. Whilst the consistent argument is a solid one, I don't think we want to disable the rule globally.

services/brig/test/integration/API/User/Client.hs
```diff
@@ -308,10 +309,15 @@ testListClients brig = do
...
-  c1 <- responseJsonMaybe <$> addClient brig uid (defNewClient PermanentClientType [pk1] lk1)
-  c2 <- responseJsonMaybe <$> addClient brig uid (defNewClient PermanentClientType [pk2] lk2)
-  c3 <- responseJsonMaybe <$> addClient brig uid (defNewClient TemporaryClientType [pk3] lk3)
-  let cs = sortBy (compare `on` clientId) $ catMaybes [c1, c2, c3]
+  c1 <- responseJsonError =<< addClient brig uid (defNewClient PermanentClientType [pk1] lk1)
+  c2 <- responseJsonError =<< addClient brig uid (defNewClient PermanentClientType [pk2] lk2)
+  c3 <- responseJsonError =<< addClient brig uid (defNewClient TemporaryClientType [pk3] lk3)
+
+  let pks = Map.fromList [(Ed25519, "random")]
+  void $ putClient brig uid (clientId c1) pks
```
elland, line 317: Could you clarify this for me? From what I understood, this test checks that listing the clients work, but here we're updating something.

services/brig/test/integration/API/User/Client.hs
```diff
@@ -321,6 +327,25 @@ testListClients brig = do
       const 200 === statusCode
       const (Just cs) === responseJsonMaybe
 
+testMLSClient :: Brig -> Http ()
+testMLSClient brig = do
+  uid <- userId <$> randomUser brig
+  let (pk1, lk1) = (somePrekeys !! 0, (someLastPrekeys !! 0))
```
elland, line 333: Same issue with the `head` thing. 😅

elland (commented):
I didn't expect I'd make that many comments, or I would've made a batch review. Sorry about that.

services/brig/test/integration/API/User/Client.hs
mdimjasevic: I'm in favor of not applying that hlint rule in tests. In tests we don't have to strive to have total functions and `head` instead of `(!!)`.

services/brig/test/integration/API/User/Client.hs
elland: Ah, I thought that was what the new test did, but this one as well?

services/brig/test/integration/API/User/Client.hs
elland: Fair point. I'll look into disabling the rule for test targets. 👍

services/brig/test/integration/API/User/Client.hs
mdimjasevic: The new test `testMLSClient` in essence does the same, but for a different endpoint: `GET /clients/:client`. I thought I'd add that test while at it (and actually my hunch was that the test would fail, but it turns out we got looking up one client right).

PR #2512, opened 2022-06-23
retry Redis connection in case of network errors

stefanwire:
The plain `hedis` client discards the initial connection data and only retains a list of Redis cluster node IPs. When none of these IPs is valid anymore, for instance, due to Redis cluster or (Kubernetes) node updates, `hedis` immediately looses all retained connections to Redis without any option to reconnect. In this patch, we wrap the `hedis` client and retry connecting with the initially provided connection data in case of network errors. Also, the wrapper makes sure that
* if reconnecting fails, it is retried with exponential back-off, and
* only one thread reconnects while all other threads are blocked and are immediately unblocked as soon as the new connection is established.

## Checklist

### DIFF
--- a/services/gundeck/src/Gundeck/Env.hs
+++ b/services/gundeck/src/Gundeck/Env.hs
@@ -35,6 +35,7 @@ import Data.Time.Clock.POSIX
 import qualified Database.Redis as Redis
 import qualified Gundeck.Aws as Aws
 import Gundeck.Options as Opt
+import qualified Gundeck.Redis as Redis
 import Gundeck.ThreadBudget
 import Imports
 import Network.HTTP.Client (responseTimeoutMicro)
@@ -42,6 +43,7 @@ import Network.HTTP.Client.TLS (tlsManagerSettings)
 import qualified System.Logger as Log
 import qualified System.Logger.Extended as Logger
 import Util.Options
+import Control.Retry (exponentialBackoff)
 
 data Env = Env
   { _reqId :: !RequestId,
@@ -50,8 +52,8 @@ data Env = Env
     _applog :: !Logger.Logger,
     _manager :: !Manager,
     _cstate :: !ClientState,
-    _rstate :: !Redis.Connection,
-    _rstateAdditionalWrite :: !(Maybe Redis.Connection),
+    _rstate :: !Redis.RobustConnection,
+    _rstateAdditionalWrite :: !(Maybe Redis.RobustConnection),
     _awsEnv :: !Aws.Env,
     _time :: !(IO Milliseconds),
     _threadBudgetState :: !(Maybe ThreadBudgetState)
@@ -113,7 +115,7 @@ reqIdMsg :: RequestId -> Logger.Msg -> Logger.Msg
 reqIdMsg = ("request" Logger..=) . unRequestId
 {-# INLINE reqIdMsg #-}
 
-createRedisPool :: Logger.Logger -> RedisEndpoint -> ByteString -> IO Redis.Connection
+createRedisPool :: Logger.Logger -> RedisEndpoint -> ByteString -> IO Redis.RobustConnection
 createRedisPool l endpoint identifier = do
   let redisConnInfo =
         Redis.defaultConnectInfo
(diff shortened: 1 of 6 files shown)


services/gundeck/src/Gundeck/Env.hs
```diff
@@ -127,28 +129,9 @@ createRedisPool l endpoint identifier = do
     Log.msg (Log.val $ "starting connection to " <> identifier <> "...")
       . Log.field "connectionMode" (show $ endpoint ^. rConnectionMode)
       . Log.field "connInfo" (show redisConnInfo)
+  let connectWithRetry = Redis.connectRobust l (exponentialBackoff 50000)
```
stefanwire, line 132: To the reviewer: I wonder whether what currently is `exponentialBackoff 50000` should be more flexible and I can imagine two ways to add flexibility,
* `50000` is a magic number and should perhaps rather come from the environment variables, however, the choice of the constant might not be mattering much enough, since more importantly the back-off would always be exponential;
* `exponentialBackoff` is just one of many retry policies in the [retry](https://hackage.haskell.org/package/retry) package and perhaps retries should be limited in their number or by the time they require and/or even
  * combine this with environment configuration.

services/gundeck/src/Gundeck/Run.hs
```diff
@@ -66,7 +67,8 @@ run o = do
     Async.cancel lst
     Async.cancel wCollectAuth
     forM_ wtbs Async.cancel
-    Redis.disconnect (e ^. rstate)
+    Redis.disconnect . (^. Redis.rrConnection) =<< takeMVar (e ^. rstate)
+    maybe (pure ()) ((=<<) (Redis.disconnect . (^. Redis.rrConnection)) . takeMVar) (e ^. rstateAdditionalWrite)
```
stefanwire, line 71: To the reviewer: the second Redis connection was never disconnected upon service shutdown in the past. I added disconnecting now, so if this is undesirable for some reason, I need to know.

battermann (commented):
Looks pretty good to me. Because I am not so familiar with redis, this is just a comment review.

flokli:
Are there any plans on how to fix this in Hedis in general, instead of working around this downstream in every application using the library?

services/gundeck/src/Gundeck/Env.hs
```diff
@@ -127,28 +129,9 @@ createRedisPool l endpoint identifier = do
...
```
akshaymankar, line 132:
```suggestion
  let connectWithRetry = Redis.connectRobust l (capDelay 1000000 (exponentialBackoff 50000))
```

Waiting more than a second to retry is not very nice and retrying every second shouldn't cause any problems either.

services/gundeck/src/Gundeck/Redis.hs
```diff
@@ -0,0 +1,125 @@
...
+  reconnectRedis robustConnection
+  pure robustConnection
+  where
+    reconnectRedis robustConnection = do
+      conn <- connectLowLevel
+
+      Log.info l $ Log.msg (Log.val "lazy connection established, running ping...")
+      -- TODO With ping, we only verify that a single node is running as opposed
+      -- to verifying that all nodes of the cluster are up and running. It
+      -- remains unclear how cluster health can be verified in hedis.
```
akshaymankar, line 79:
```suggestion
      -- FUTUREWORK: With ping, we only verify that a single node is running as opposed
      -- to verifying that all nodes of the cluster are up and running. It
      -- remains unclear how cluster health can be verified in hedis.
```

I think it is ok if we don't do this now.

akshaymankar (commented):
Looks good so far, I will test it in some cluster before approving.

services/gundeck/src/Gundeck/Redis.hs
```diff
@@ -0,0 +1,127 @@
...
+--
+-- Blocks on connection errors as long as the connection is not reestablished.
+-- Without externally enforcing timeouts, this may lead to leaking threads.
+runRobust :: RobustConnection -> Redis a -> IO a
+runRobust mvar action = do
+  robustConnection <- readMVar mvar
+  catches
+    (runRedis (_rrConnection robustConnection) action)
+    [ Handler (\(_ :: ConnectionLostException) -> reconnectRetry robustConnection), -- Redis connection lost during request
+      Handler (\(_ :: IOException) -> reconnectRetry robustConnection) -- Redis unreachable
```
akshaymankar, line 115: Can you also add logs here? Without this it is a little confusing why the lazy connection starts restarting. Also, it would be nice to know which type of errors are causing a connection restart.

akshaymankar (approved):
It works :tada: 
There is a small window where the messages just don't go. But I guess, this is what was happening even with the old library when connecting to redis in master mode.

PR #2514, opened 2022-06-24
Remove non binding teams

battermann:
https://wearezeta.atlassian.net/browse/SQSERVICES-1619

Remove deprecated non-binding teams code.

## Checklist


failing_teams_api_tests.txt
```diff
@@ -0,0 +1,3 @@
+-p '/Teams API/&&/add team conversation with role/'
+-p '/Teams API/&&/add team member to conversation without connection/'
+-p '/Teams API/&&/delete team conversation/'
```
fisx, line 3: what's with this file?

tools/stern/src/Stern/Intra.hs
```diff
@@ -404,8 +404,7 @@ getUserBindingTeam u = do
   pure $
     listToMaybe $
       fmap (view teamId) $
-        filter ((== Binding) . view teamBinding) $
-          teams ^. teamListTeams
+        teams ^. teamListTeams
```
fisx, line 407: This makes me think that perhaps you should filter the galley cql queries for the `binding` field.  We don't *expect* there to be any non-binding teams in the database, but for robustness we might as well make sure we don't suddenly treat find them and treat them as binding teams.

services/brig/src/Brig/API/Error.hs
```diff
@@ -366,8 +366,9 @@ authTokenUnsupported = Wai.mkError status403 "invalid-credentials" "Unsupported
 insufficientTeamPermissions :: Wai.Error
 insufficientTeamPermissions = errorToWai @'E.InsufficientTeamPermissions
 
-noBindingTeam :: Wai.Error
-noBindingTeam = Wai.mkError status403 "no-binding-team" "Operation allowed only on binding teams"
+-- FUTUREWORK: should this status be 404?
+teamNotFound :: Wai.Error
+teamNotFound = Wai.mkError status403 "no-binding-team" "Operation allowed only on binding teams"
```
fisx, line 371: Your FUTUREWORK is correct:

```suggestion
teamNotFound :: Wai.Error
teamNotFound = notFound "Team not found."
```

You probably have to also fix some `CanThrow`s if you do this, right?

libs/wire-api/src/Wire/API/Routes/Public/Galley.hs
```diff
@@ -930,6 +930,7 @@ type TeamAPI =
     :<|> Named
            "get-teams"
            ( Summary "Get teams (deprecated); use `GET /teams/:tid`"
+               :> Until 'V2
```
fisx, line 933: beautiful!  :-)

services/galley/test/integration/API/Teams.hs
```diff
@@ -123,14 +121,11 @@ tests s =
...
       test s "enable/disable Custom Search Visibility" testEnableTeamSearchVisibilityPerTeam,
-      test s "create 1-1 conversation between non-binding team members (fail)" testCreateOne2OneFailNonBindingTeamMembers,
+      test s "create 1-1 conversation between non-team-members (fail)" testCreateOne2OneFailForNonTeamMembers,
       test s "create 1-1 conversation between binding team members" (testCreateOne2OneWithMembers RoleMember),
       test s "create 1-1 conversation between binding team members as partner" (testCreateOne2OneWithMembers RoleExternalPartner),
-      test s "add new team member" testAddTeamMember,
       test s "poll team-level event queue" testTeamQueue,
-      test s "add new team member binding teams" testAddTeamMemberCheckBound,
       test s "add new team member internal" testAddTeamMemberInternal,
-      test s "remove aka delete team member" testRemoveNonBindingTeamMember,
```
fisx, line 133: I think these should all stay, and use binding teams instead.  (Or have you moved them elsewhere?)

services/galley/test/integration/API/Teams.hs
```diff
@@ -187,20 +182,12 @@ timeout = 3 # Second
...
-    assertQueueEmpty
-    liftIO $ do
-      assertEqual "owner" owner (team ^. teamCreator)
-      eventChecks <- WS.awaitMatch timeout wsOwner $ \notif -> do
-        ntfTransient notif @?= False
-        let e = List1.head (WS.unpackPayload notif)
-        e ^. eventTeam @?= tid
-        e ^. eventData @?= EdTeamCreate team
-      void $ WS.assertSuccess eventChecks
+  tid <- Util.createBindingTeamInternal "foo" owner
```
fisx, line 186: you should keep the code that checks for the team creation events.  (unless those events are only triggered for non-binding teams?)

services/galley/test/integration/API/Teams/Feature.hs
```diff
@@ -264,47 +246,42 @@ testSearchVisibility = do
         g <- view tsGalley
         Util.putTeamSearchVisibilityAvailableInternal g teamid val
 
-  assertFlagForbidden $ Util.getTeamFeatureFlag @Public.SearchVisibilityAvailableConfig nonMember tid
+  (owner, tid, member : _) <- Util.createBindingTeamWithNMembers 1
```
fisx, line 249:
```suggestion
  (owner, tid, [member]) <- Util.createBindingTeamWithNMembers 1
```

this is a super-nit-pick, but in tests i prefer incomplete patterns over over-general patterns that silently fail if something doesn't work as expected.

services/brig/src/Brig/API/Error.hs
battermann: It's only used in the internal POST /i/legalhold-login` route. So there is no swagger.

services/galley/test/integration/API/Teams.hs
battermann: No, because "Add Team Member" was only implemented for non-binding teams.
And there is a separate test for removing binding team members.

fisx (approved):

tools/stern/src/Stern/Intra.hs
fisx: https://github.com/wireapp/wire-server/pull/2514/commits/fbee436ab2f362f01c8d1604ed76778b87bea3a2

did i forget anything?

fisx:
One more thing: There are now a lot of names containing "binding", which doesn't make sense any more now.  I wonder if there should be another PR cleaning those up as well?

battermann:
> One more thing: There are now a lot of names containing "binding", which doesn't make sense any more now. I wonder if there should be another PR cleaning those up as well?

I think we should do this in another PR.

services/galley/test/integration/API/Teams.hs
battermann: We check for the actived event, there is no creation event. And we also assert that the queue is empty.

services/brig/test/integration/API/User/Auth.hs
```diff
@@ -586,7 +586,7 @@ testRegularUserLegalHoldLogin brig = do
   uid <- userId <$> randomUser brig
   -- fail if user is not a team user
   legalHoldLogin brig (LegalHoldLogin uid (Just defPassword) Nothing) PersistentCookie !!! do
-    const 403 === statusCode
+    const 404 === statusCode
```
fisx, line 589: @dkovacevic do you check the specifics of 4xx when logging in LH devices?  we're changing the status code for an error condition from 403 to 404 here, any objections?

PR #2526, opened 2022-06-30
[SQSERVICES-1067] Address user account consistency issue when SCIM provisioning crashes

elland:
This should prevent issues when creating a user fails midway, and we end up with dangling data that needs to be manually deleted in order to enable user creation.

As part of this effort, we're taking steps to break down `createUser` into multiple, use-case specific functions, which take more specific arguments to what it actually needs.

I've left a few checkmarks below unchecked, but I'll be getting those done soon, it shouldn't affect review.

Special attention should be paid to the new `NewUserSpar` record and its usage as well as a change in the `createSAML` functions. Also, this adds a new internal endpoint, `POST i/users/spar`, to disambiguate from the (hopefully soon to be legacy) `i/users`.

## Checklist


libs/wire-api/src/Wire/API/User.hs
```diff
@@ -40,11 +40,16 @@ module Wire.API.User
...
     RegisterError (..),
+    -- CreateUserSparError (..),
     RegisterSuccess (..),
     RegisterResponses,
     RegisterInternalResponses,
     NewUser (..),
     emptyNewUser,
+    NewUserSpar (..),
+    newUserFromSpar,
+    urefToExternalId,
```
elland, line 51: Organising code.

libs/wire-api/src/Wire/API/User/RichInfo.hs
```diff
@@ -288,22 +301,19 @@ modelRichField = Doc.defineModel "RichField" $ do
...
-instance FromJSON RichField where
-  parseJSON = withObject "RichField" $ \o -> do
-    RichField
-      <$> (CI.mk <$> o .: "type")
-      <*> o .: "value"
+--   -- NB: "name" would be a better name for 'richFieldType', but "type" is used because we
+--   -- also have "type" in SCIM; and the reason we use "type" for SCIM is that @{"type": ...,
+--   -- "value": ...}@ is how all other SCIM payloads are formatted, so it's quite possible
+--   -- that some provisioning agent would support "type" but not "name".
+instance ToSchema RichField where
```
elland, line 308: Less is more :)

services/spar/src/Spar/Scim/User.hs
```diff
@@ -463,8 +463,8 @@ createValidScimUser tokeninfo@ScimTokenInfo {stiTeam} vsu@(ST.ValidScimUser veid
           -- up.  We should consider making setUserHandle part of createUser and
           -- making it transactional.  If the user redoes the POST A new standalone
           -- user will be created.}
-          BrigAccess.setHandle buid handl
-          BrigAccess.setRichInfo buid richInfo
+          -- BrigAccess.setHandle buid handl
```
elland, line 466: Still needs to be handled for the noSAML use case.

services/spar/test/Test/Spar/DataSpec.hs
```diff
@@ -40,12 +41,14 @@ check :: HasCallStack => Int -> Env -> String -> Either TTLError (TTL "authresp"
 check testnumber env (parsetm -> endOfLife) expectttl =
   it (show testnumber) $ mkTTLAssertions env endOfLife `shouldBe` expectttl
 
+parsetm :: HasCallStack => String -> UTCTime
+parsetm = fromJust . parseTimeM True defaultTimeLocale "%Y-%m-%dT%H:%M:%S%QZ"
+
+{-# HLINT ignore "Eta reduce" #-} -- For clarity
```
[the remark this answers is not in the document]
elland: Yeah, it's certainly contentions.

libs/wire-api/src/Wire/API/User.hs
```diff
@@ -588,6 +596,103 @@ instance (res ~ RegisterInternalResponses) => AsUnion res (Either RegisterError
...
+    newUserSparRichInfo :: Maybe RichInfo
+  }
+  deriving stock (Eq, Show, Generic)
+  deriving (ToJSON, FromJSON, S.ToSchema) via (Schema NewUserSpar)
+
+instance ToSchema NewUserSpar where
+  schema =
+    object "NewUserSpar" $
+      NewUserSpar
+        <$> newUserSparUUID .= field "newUserSparUUID" genericToSchema
```
supersven, line 669: I'm wondering if the `newUserSpar` prefix is needed for each field name of the rendered / unrendered JSON?

services/brig/src/Brig/API/Internal.hs
```diff
@@ -124,7 +123,9 @@ mlsAPI =
     :<|> mapKeyPackageRefsInternal
 
 accountAPI :: ServerT BrigIRoutes.AccountAPI (Handler r)
-accountAPI = Named @"createUserNoVerify" createUserNoVerify
+accountAPI =
+  Named @"createUserNoVerify" createUserNoVerify
+    :<|> Named @"createUserNoVerifySpar" createUserNoVerifySpar
```
supersven, line 128: When will this new endpoint be called? Is there a test for it?

supersven (changes requested):

services/brig/src/Brig/API/User.hs
```diff
@@ -213,6 +215,73 @@ verifyUniquenessAndCheckBlacklist uk = do
...
+  -- Set up feature flags
+  let uid = userId (accountUser account)
+  lift $ initAccountFeatureConfig uid
+
+  -- Set handle
+  updateHandle' uid handle'
+
+  pure $! CreateUserResult account Nothing Nothing (Just userTeam)
+  where
+    lmap f =
```
mdimjasevic, line 253: There's [mapLeft](https://hackage.haskell.org/package/either-5.0.2/docs/Data-Either-Combinators.html#v:mapLeft), or alternatively you could use [first](https://hackage.haskell.org/package/base-4.14.1.0/docs/Data-Bifunctor.html#v:first).

services/spar/src/Spar/Scim/User.hs
```diff
@@ -446,21 +446,17 @@ createValidScimUser tokeninfo@ScimTokenInfo {stiTeam} vsu@(ST.ValidScimUser veid
...
                     -- to `createValidScimUserSpar`?
                     uid <- Id <$> Random.uuid
-                    BrigAccess.createSAML uref uid stiTeam name ManagedByScim
+                    BrigAccess.createSAML uref uid stiTeam name ManagedByScim (Just handl) (Just richInfo)
               )
-              ( \email ->
-                  BrigAccess.createNoSAML email stiTeam name language
+              ( \email -> do
+                  buid <- BrigAccess.createNoSAML email stiTeam name language
+                  BrigAccess.setHandle buid handl -- TODO: possibly do the same one req as we do for saml?
```
mdimjasevic, line 453: An unhandled TODO note.

mdimjasevic (commented):
There are several things I would change and you can find comments inlined.

I wasn't able to grasp the logic behind your proposed change so it would be great if you can walk me through the PR in a call.

services/brig/src/Brig/API/User.hs
```diff
@@ -213,6 +215,73 @@ verifyUniquenessAndCheckBlacklist uk = do
...
+      let h = \case
+            Left x ->
+              Left (f x)
+            Right a ->
+              Right a
+          {-# INLINE h #-}
```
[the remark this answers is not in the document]
elland: I just copied it over as is, but it should be fine to remove this.

libs/wire-api/src/Wire/API/User.hs
elland: Since this is an internal API used only between services we plan on merging into one soon®️, I didn't think necessary to massage the data presentation more than necessary to make it work.

services/brig/src/Brig/API/Internal.hs
elland: This is called when creating users via `createBrigUserSAML`, from `createValidScimUser` in `Spar/Scim/User.hs` and is tested through those integration tests, as well as through `createSamlUserWithId` from `Spar/App.hs`.

libs/wire-api/src/Wire/API/User/Identity.hs
```diff
@@ -58,11 +58,22 @@ import Data.Aeson (FromJSON (..), ToJSON (..))
...
+      SwaggerDoc,
+      ObjectSchema,
+      (.=),
+      genericToSchema,
+      maybe_,
+      optField,
+      parsedText,
+      withParser,
+      HasDoc(doc),
+      Schema(Schema) )
```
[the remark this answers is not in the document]
elland: Oh I think this was HLS being too proactive.

libs/wire-api/src/Wire/API/User.hs
```diff
@@ -588,6 +596,103 @@ instance (res ~ RegisterInternalResponses) => AsUnion res (Either RegisterError
   toUnion = eitherToUnion (toUnion @RegisterErrorResponses) (Z . I)
   fromUnion = eitherFromUnion (fromUnion @RegisterErrorResponses) (unI . unZ)
 
+urefToExternalId :: SAML.UserRef -> Maybe Text
```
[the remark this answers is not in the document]
elland: Same. Since I just moved this code from somewhere else, I didn't want to make this PR even larger than it already was by refactoring this + all the call sites.

services/brig/src/Brig/API/User.hs
```diff
@@ -250,14 +251,8 @@ createUserSpar new = do
...
-    lmap f =
-      let h = \case
-            Left x ->
-              Left (f x)
-            Right a ->
-              Right a
-          {-# INLINE h #-}
-       in ExceptT . fmap h . runExceptT
+    lmap :: Functor m => (a -> b) -> ExceptT a m v -> ExceptT b m v
+    lmap f = ExceptT . fmap (mapLeft f) . runExceptT
```
mdimjasevic, line 255: This looks like [withExceptT](https://hoogle.zinfra.io/file/root/.ghcup/ghc/8.10.7/share/doc/ghc-8.10.7/html/libraries/transformers-0.5.6.2/Control-Monad-Trans-Except.html#v:withExceptT).

mdimjasevic (approved):
Just take care of remaining TODOs and that's it! Thanks for the walk-through during the call!

services/brig/src/Brig/API/User.hs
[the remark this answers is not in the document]
fisx: (code changed completely, so this question doesn't apply to this PR any more.)

PR #2543, opened 2022-07-06
Sqservices 1548 be csv export from tm created on date missing in some instances

battermann:
https://wearezeta.atlassian.net/browse/SQSERVICES-1548

## Checklist

### DIFF
--- a/services/spar/test-integration/Test/Spar/Scim/UserSpec.hs
+++ b/services/spar/test-integration/Test/Spar/Scim/UserSpec.hs
@@ -2067,6 +2067,8 @@ testDeletedUsersFreeExternalIdNoIdp = do
       (runSpar $ ScimExternalIdStore.lookup tid email)
       (== Nothing)
 
+-- | CSV download of team members is mainly tested here: 'API.Teams.testListTeamMembersCsv'.
+-- the additional CSV download test here is specifically focused on SCIM users
 specSCIMManaged :: SpecWith TestEnv
 specSCIMManaged = do
   describe "SCIM-managed users" $ do
--- a/libs/wire-api/test/unit/Test/Wire/API/Roundtrip/Aeson.hs
+++ b/libs/wire-api/test/unit/Test/Wire/API/Roundtrip/Aeson.hs
@@ -225,6 +225,7 @@ tests =
       testRoundTrip @(User.LimitedQualifiedUserIdList 20),
       testRoundTrip @User.UserProfile,
       testRoundTrip @User.User,
+      testRoundTrip @User.UserSet,
       testRoundTrip @User.SelfProfile,
       testRoundTrip @User.InvitationCode,
       testRoundTrip @User.BindingNewTeamUser,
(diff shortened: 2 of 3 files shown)


libs/wire-api/src/Wire/API/Routes/Public/Spar.hs
```diff
@@ -124,6 +125,7 @@ type APIINTERNAL =
   "status" :> Get '[JSON] NoContent
     :<|> "teams" :> Capture "team" TeamId :> DeleteNoContent
     :<|> "sso" :> "settings" :> ReqBody '[JSON] SsoSettings :> Put '[JSON] NoContent
+    :<|> "scim" :> "userinfos" :> ReqBody '[JSON] UserSet :> Post '[JSON] ScimUserInfos
```
fisx, line 128: Unrelated to this PR, but why is `APIINTERNAL` all-capsed (probably a question i should ask myself), and why is it defined in `.../API/Public.hs`?  Oh well.

libs/wire-api/src/Wire/API/User.hs
```diff
@@ -892,6 +895,53 @@ instance ToSchema BindingNewTeamUser where
         <$> bnuTeam .= bindingNewTeamObjectSchema
         <*> bnuCurrency .= maybe_ (optField "currency" genericToSchema)
 
+--------------------------------------------------------------------------------
+-- SCIM User Info
+
+data ScimUserInfo = ScimUserInfo
+  { suiUserId :: UserId,
+    suiCreatedOn :: Maybe UTCTimeMillis
```
fisx, line 903: How long would it take to add "createdby" to the csv file, while you're at it?  Seems like a good opportunity to avoid potential follow-up tickets, but I'm not sure.

services/spar/test-integration/Test/Spar/Scim/UserSpec.hs
```diff
@@ -2112,6 +2113,27 @@ specSCIMManaged = do
           updateProfileBrig brig uid uupd !!! do
             (fmap Wai.label . responseJsonEither @Wai.Error) === const (Right "managed-by-scim")
             statusCode === const 403
+    it "created_on should be filled in CSV export" $ do
```
fisx, line 2116: There is a test `testListTeamMembersCsv` in galley that (I think) could be easily duplicated into a variant that generates the members via scim, not team invitations.  That test should fail before the changes in this PR, and succeed after.

If you don't want to move this test there and avoid code redundancy, please cross-reference the two tests, that's just as good IMO.

fisx (approved):
Perfect, except for the missing round-trip tests.

libs/wire-api/src/Wire/API/User.hs
```diff
@@ -23,6 +23,9 @@ module Wire.API.User
   ( UserIdList (..),
     QualifiedUserIdList (..),
     LimitedQualifiedUserIdList (..),
+    ScimUserInfo (..),
+    ScimUserInfos (..),
+    UserSet (..),
```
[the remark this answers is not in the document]
battermann: done

PR #2544, opened 2022-07-06
Add inbucket Helm chart

supersven:
[Inbucket](https://www.inbucket.org/) is a SMTP server that displays received eMails in a webapp and provides them via an REST API. This is useful for testing.

The `inbucket` chart in wire-server provides defaults for the parent chart ([`inbucket/inbucket`](https://artifacthub.io/packages/helm/inbucket/inbucket)) to simplify it's usage in Wire test environments. To use it in our `cailleach` systems, only a `host` value has to be defined (despite of including this chart, of course.).
The parent Helm chart uses an older image, I'm overriding it to keep the count of known vulnerabilities down.

See http://inbucket.sven-test.wire.link/monitor as an example.

The corresponding ticket is: https://wearezeta.atlassian.net/browse/SQPIT-1034

I've also added a `make` target `charts-serve-all`, that makes all (not only those needed by integration tests) Helm charts available on a local HTTP server.

## Checklist

schema`** to update the cassandra schema documentation.

### DIFF
--- a/charts/inbucket/templates/ingress.yaml
+++ b/charts/inbucket/templates/ingress.yaml
@@ -10,7 +10,7 @@ metadata:
     helm.sh/chart: {{ include "inbucket.chart" . }}
 spec:
   rules:
-    - host: {{ .Values.host | quote }}
+    - host: {{ required "must specify host" .Values.host | quote }}
       http:
         paths:
           - path: /


charts/inbucket/templates/ingress.yaml
```diff
@@ -0,0 +1,19 @@
...
+  name: "inbucket"
+  namespace: {{ .Release.Namespace }}
+  labels:
+    app.kubernetes.io/name: {{ include "inbucket.name" . }}
+    app.kubernetes.io/instance: {{ .Release.Name }}
+    app.kubernetes.io/managed-by: {{ .Release.Service }}
+    helm.sh/chart: {{ include "inbucket.chart" . }}
+spec:
+  rules:
+    - host: {{ .Values.host | quote }}
```
jschaul, line 13:
```suggestion
    - host: {{ required "must specify host" .Values.host | quote }}
```

jschaul (approved):

charts/inbucket/templates/ingress.yaml
supersven: That looks like a good improvement! :+1: 

I think I (almost) accomplished the same by defining `host: ` to be empty in the `values.yaml`. However, this one is more "expressive"; tells better what's expected.

PR #2553, opened 2022-07-12
coturn multi-tenancy support

sysvinit:
This change updates the coturn Helm chart to add support for passing multiple authentication secrets to coturn. This makes it possible to operate a multi-tenant TURN service without needing to share secrets between consumers.

coturn already supports multiple auth secrets, so this change just adds the necessary configuration machinery to the Helm chart.

## Checklist

### DIFF
--- a/charts/coturn/templates/secret.yaml
+++ b/charts/coturn/templates/secret.yaml
@@ -1,3 +1,8 @@
+{{- if or (not .Values.secrets) (not .Values.secrets.zrestSecrets) }}
+{{- fail "Secrets are not defined" }}
+{{- else if eq (len .Values.secrets.zrestSecrets) 0 }}
+{{- fail "At least one authentication secret must be defined" }}
+{{- else }}
 apiVersion: v1
 kind: Secret
 metadata:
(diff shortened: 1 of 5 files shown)


charts/coturn/templates/secret.yaml
```diff
@@ -7,6 +12,8 @@ metadata:
     release: "{{ .Release.Name }}"
     heritage: "{{ .Release.Service }}"
 type: Opaque
-data:
-  zrest_secret.txt: {{ .Values.secrets.zrestSecret | b64enc | quote }}
-
+stringData:
+  zrest_secret.txt: |
+    {{- range .Values.secrets.zrestSecrets }}{{ . | nindent 4 }}
```
supersven, line 17: Could you please confirm that `b64enc` (base64 encoding) isn't needed for values of the list?

(It looks a bit suspicious, but might be fine.)

supersven (approved):
Looks good :+1: 

Just one question...

charts/coturn/templates/secret.yaml
sysvinit: Yes, I'm purposefully using `stringData` here instead of `data` -- see [this API spec](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.24/#secret-v1-core).

supersven: Didn't know that one :smiley_cat:

PR #2562, opened 2022-07-14
fix typo of feature flag everywhere. Is this an API breaking change?

jschaul:
It looks like settings for the feature flag in helm chart never had any effects, as the parser looked for the version with captital `E`, not lowercase `e`.

Would one of you reviewers be able to amend/push-to/merge/close-unmerged this PR as needed? I don't have any context on what this is about; we just did diffing on configurations and saw this difference between old (hegemony) and new (helm based) prod configs.

### DIFF
--- a/libs/wire-api/src/Wire/API/Team/Feature.hs
+++ b/libs/wire-api/src/Wire/API/Team/Feature.hs
@@ -517,7 +517,7 @@ instance ToSchema ValidateSAMLEmailsConfig where
   schema = object "ValidateSAMLEmailsConfig" objectSchema
 
 instance IsFeatureConfig ValidateSAMLEmailsConfig where
-  type FeatureSymbol ValidateSAMLEmailsConfig = "validateSAMLemails"
+  type FeatureSymbol ValidateSAMLEmailsConfig = "validateSAMLEmails"
   defFeatureStatus = WithStatus FeatureStatusEnabled LockStatusUnlocked ValidateSAMLEmailsConfig
   objectSchema = pure ValidateSAMLEmailsConfig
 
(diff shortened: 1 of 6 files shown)


elland:
Nice catch. Sadly now there's a conflict.

services/spar/src/Spar/Intra/Galley.hs
```diff
@@ -84,7 +84,7 @@ assertSSOEnabled tid = do
 
 isEmailValidationEnabledTeam :: (HasCallStack, MonadSparToGalley m) => TeamId -> m Bool
 isEmailValidationEnabledTeam tid = do
-  resp <- call $ method GET . paths ["i", "teams", toByteString' tid, "features", "validateSAMLemails"]
+  resp <- call $ method GET . paths ["i", "teams", toByteString' tid, "features", "validateSAMLEmails"]
```
battermann, line 87:
```suggestion
  resp <- call $ method GET . paths ["i", "teams", toByteString' tid, "features", featureNameBS @ValidateSAMLEmailsConfig]
```

libs/wire-api/src/Wire/API/Team/Feature.hs
```diff
@@ -517,7 +517,7 @@ instance ToSchema ValidateSAMLEmailsConfig where
   schema = object "ValidateSAMLEmailsConfig" objectSchema
 
 instance IsFeatureConfig ValidateSAMLEmailsConfig where
-  type FeatureSymbol ValidateSAMLEmailsConfig = "validateSAMLemails"
+  type FeatureSymbol ValidateSAMLEmailsConfig = "validateSAMLEmails"
```
battermann, line 520: This could be a breaking change because URL should be case-sensitive, however in practice they might not be.

battermann (commented):
Do we need to make any changes in the cailleach repo, too?

jschaul:
The helm chart part was fixed in #2563, see https://github.com/wireapp/wire-server/pull/2563/files#diff-252a8cec589e29d129e79ecfe2e881b1eba870106c2b1cffa618dc19b19dd54eR82-R84

The Haskell code is and remains inconsistent, but perhaps we can live with that. Someone working normally on this code can decide to follow up on this; or not.

PR #2576, opened 2022-07-22
refresh federator certificates for unit tests

stefanwire:
## Checklist


stefanwire:
The new certificates will expire on Aug 21, 2022 with a TTL of just one month. Should the TTL be extended to a much larger duration given that it's all about unit tests?

mdimjasevic (approved):

PR #2581, opened 2022-07-26
do rate limiting exemptions by proxy_protocol_ip

jschaul:
Previous place we did this was in https://github.com/zinfra/cailleach/pull/726/commits/6f10bf407915a2f07e849007f181a723660eac9f as part of https://github.com/zinfra/cailleach/pull/726

We are in general lacking tests for these issues. Anyone has ideas on how to test this to avoid regressions?

### DIFF
--- a/changelog.d/3-bug-fixes/nginz-rate-limit-exemptions
+++ b/changelog.d/3-bug-fixes/nginz-rate-limit-exemptions
@@ -1,0 +1,1 @@
+charts/nginz: rate limit exemptions should be by comparing with proxy_protocol_addr
--- a/charts/nginz/templates/conf/_nginx.conf.tpl
+++ b/charts/nginz/templates/conf/_nginx.conf.tpl
@@ -93,7 +93,7 @@ http {
   # Rate Limiting Exemptions
   #
 
-  geo $rate_limit {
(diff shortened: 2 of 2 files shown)


sysvinit:
There are some sharp edges in here because clients' true source addresses are stored in different variables depending on whether the connection uses proxy protocol or `X-Forwarded-For`, and from memory nginx makes it very difficult to write configuration which is generic over both cases.

This change is the right thing to do (so that the chart matches the config we've been using internally hitherto), and I'll approve it, but consolidating (and defanging) the rest of the nginx config here is something we should consider as future work.

sysvinit (approved):

jschaul:
> There are some sharp edges in here because clients' true source addresses are stored in different variables depending on whether the connection uses proxy protocol or `X-Forwarded-For`, and from memory nginx makes it very difficult to write configuration which is generic over both cases.
> 
> This change is the right thing to do (so that the chart matches the config we've been using internally hitherto), and I'll approve it, but consolidating (and defanging) the rest of the nginx config here is something we should consider as future work.

That makes sense. https://wearezeta.atlassian.net/browse/SQPIT-1149

PR #2582, opened 2022-07-27
[FS-794] MLS: Delete Expired Key Packages

mdimjasevic:
When looking up an MLS key package for a client, all of its expired key packages are deleted first from the database. This affects the claim endpoint and the key-package count endpoint (in the sense of filtering out expired key packages).

My first attempt was to use Cassandra's `using ttl` clause, but that failed because a maximum value of 20 years is supported for  expiring records in Cassandra, yet key packages can last longer than that.

Tracked by https://wearezeta.atlassian.net/browse/FS-794.

TODO:

- [x] Merge a PR https://github.com/wireapp/mls-test-cli/pull/4 to `mls-test-cli` that incorporates the option for setting the key package lifetime and update the reference to the tool in the Nix configuration.

## Checklist

### DIFF
--- a/changelog.d/2-features/expired-key-packages
+++ b/changelog.d/2-features/expired-key-packages
@@ -1,0 +1,1 @@
+Expired MLS key packages are deleted from the database
--- a/libs/wire-api/wire-api.cabal
+++ b/libs/wire-api/wire-api.cabal
@@ -478,9 +478,9 @@ test-suite wire-api-golden-tests
       Test.Wire.API.Golden.Generated.VerifyDeleteUser_user
       Test.Wire.API.Golden.Generated.ViewLegalHoldService_team
       Test.Wire.API.Golden.Generated.ViewLegalHoldServiceInfo_team
-      Test.Wire.API.Golden.Generated.WithStatusPatch_team
       Test.Wire.API.Golden.Generated.WithStatus_team
       Test.Wire.API.Golden.Generated.WithStatusNoLock_team
+      Test.Wire.API.Golden.Generated.WithStatusPatch_team
       Test.Wire.API.Golden.Generated.Wrapped_20_22some_5fint_22_20Int_user
       Test.Wire.API.Golden.Manual
       Test.Wire.API.Golden.Manual.ClientCapability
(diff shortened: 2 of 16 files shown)


pcapriotti:
Maybe it's fine, but having to parse all of a user's key packages every time one of them is claimed seems like it might result in some performance issues. Can you reconsider using Cassandra's TTL? Capping the lifetime to 20 years shouldn't be an issue.

mdimjasevic:
> Maybe it's fine, but having to parse all of a user's key packages every time one of them is claimed seems like it might result in some performance issues. Can you reconsider using Cassandra's TTL? Capping the lifetime to 20 years shouldn't be an issue.

I just skimmed over the MLS specification and I can't find anything on the upper bound for lifetime. Sure, 20 years sounds like a lot, but I wasn't sure if it would be fine to introduce this Cassandra-specific constraint in the implementation.

Should I go ahead with Cassandra's TTL?

pcapriotti:
I mean, what's the probability that all of this will be relevant in 2042? :)

mdimjasevic:
> Maybe it's fine, but having to parse all of a user's key packages every time one of them is claimed seems like it might result in some performance issues. Can you reconsider using Cassandra's TTL? Capping the lifetime to 20 years shouldn't be an issue.

I was eager to rebase and delete the version based on Cassandra's TTL. Silly me. Sure, I can reconstruct it. Still, I don't think parsing would be the bottleneck, rather the database access would be, but still I don't expect this to cause any user-observable slowdowns.

I'm inclined to keep this as is and not rewrite it to the Cassandra's TTL approach unless you see it as a must.

pcapriotti (approved):
Alright, let's go with this then, although I'm slightly uncomfortable with the repeated parsing. There is also the option of storing the expiry timestamp in cassandra and check that instead of reparsing the keypackage, although this would essentially be reimplementing the TTL feature.

mdimjasevic:
> Alright, let's go with this then, although I'm slightly uncomfortable with the repeated parsing. There is also the option of storing the expiry timestamp in cassandra and check that instead of reparsing the keypackage, although this would essentially be reimplementing the TTL feature.

Ok, thanks!

I thought I'd add an expiration timestamp column to the table, but came to the same conclusion. Also, I believe it'd require a data migration aside a schema migration, and this made me not so excited about the option.

PR #2588, opened 2022-07-29
Allow remove splash screen

fisx:
https://wearezeta.atlassian.net/browse/SQSERVICES-759

## Checklist

 - [x] If HTTP endpoint paths have been added or renamed, or feature configs have changed, the **endpoint / config-flag checklist** (see Wire-employee only backend [wiki page](https://github.com/zinfra/backend-wiki/wiki/Checklists)) has been followed.


services/galley/src/Galley/API/Teams.hs
```diff
@@ -336,6 +337,29 @@ updateTeamH zusr zcon tid updateData = do
...
+  ConnId ->
+  TeamId ->
+  Sem r ()
+updateTeamDeleteSplashScreenH zusr zcon tid = do
+  zusrMembership <- E.getTeamMember tid zusr
+  void $ permissionCheckS SSetTeamData zusrMembership
+  E.deleteTeamSplashScreen tid
+  now <- input
+  memList <- getTeamMembersForFanout tid
+  let e = newEvent tid now (EdTeamUpdate newTeamUpdateData)
```
fisx, line 359: I'm not very happy about this: deleting the splash screen is not the same as changing nothing.  :)  Maybe add a new constructor to `TeamUpdateData`?

services/galley/src/Galley/API/Teams.hs
```diff
@@ -336,6 +337,29 @@ updateTeamH zusr zcon tid updateData = do
...
+     ]
+    r =>
+  UserId ->
+  ConnId ->
+  TeamId ->
+  Sem r ()
+updateTeamDeleteSplashScreenH zusr zcon tid = do
+  zusrMembership <- E.getTeamMember tid zusr
+  void $ permissionCheckS SSetTeamData zusrMembership
+  E.deleteTeamSplashScreen tid
```
fisx, line 356: I'm not convinced this is the right approach, but the alternative of changing `TeamUpdateData` to accomodate both leaving the splash screen intact (`Nothing`) and setting it to `Nothing` (???) doesn't appeal to me either.  We could change the `Maybe` to something like:

```haskell
data ChangeSplashScreen = Keep | Delete | Set AssetKey
```

But then how would that translate into json?  like this?:

```
{ ... 
  "splash_screen": null | {"delete"} | {"set": "key"},
  ...
}
```

fisx: I mean:

```
  "splash_screen": null | {"delete": true} | {"set": "key"},
```

fisx: I ended up doing what Patryk suggested earlier, namely do the same thing we're already doing for the Icon (aka Logo) [here](https://github.com/wireapp/wire-server/blob/5b0a276051990c4ff51eeff77798ab15715cd059/libs/wire-api/src/Wire/API/Team.hs#L247).

fisx:
Golden tests are still failing, but it's ready for review!

fisx:
ok, this failure is also legit, fix coming up.

pcapriotti (approved):
LGTM.

PR #2564, opened 2022-07-14
[SQServices-1627] Check remaining time of an activated a feature in back office

elland:
The changes to the feature endpoint is backward compatible and shouldn't require any specific deploy order. 

## Checklist

 - [x] The **PR Title** explains the impact of the change.
 - [x] The **PR description** provides context as to why the change should occur and what the code contributes to that effect. This could also be a link to a JIRA ticket or a Github issue, if there is one.
 - [x] **changelog.d** contains the following bits of information ([details](https://github.com/wireapp/wire-server/blob/develop/docs/developer/changelog.md)):
 - [x] If public end-points have been changed or added: does nginz need un upgrade?
 - [x] If internal end-points have been added or changed: which services have to be deployed in a specific order?

PR #2592, opened 2022-08-03
Updated Schema Profuctor readme with another example

elland:
## Checklist

### DIFF
--- a/libs/schema-profunctor/README.md
+++ b/libs/schema-profunctor/README.md
@@ -211,6 +211,12 @@ a highly customised instance.
 Here's an example of such a type. We can have either Seconds or Unlimited,
 with a caveat: 0 ~ Unlimited.
 
+(This examples has problems: it (a) breaks the law that `decode
+. encode == Just . id`, and (b) allows you to construct a value
+`FeatureTTLSeconds 0` to mean unlimited, which will probably break
+your application code.  So consider this a suggestion what
+schema-profunctor *allows* you to do, not what you *should* be doing.)
+
 ``` haskell
 data FeatureTTL
   = FeatureTTLSeconds Word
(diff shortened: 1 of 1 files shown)


fisx (changes requested):
please change as discussed in https://github.com/wireapp/wire-server/pull/2564

pcapriotti (commented):
I'm not sure this example requires `mkSchema`. Couldn't you use [`tag`](https://wearezeta.atlassian.net/wiki/spaces/CORE/pages/553418860/MLS+implementation+in+wire-server#Backend-side-removal) and `withParser`?

libs/schema-profunctor/README.md
```diff
@@ -203,6 +203,50 @@ Finally, we add a name to the schema using the `named`
...
+              else A.parseFail "Expected ''unlimited' or '0'."
+
+      parseSeconds :: A.Value -> A.Parser FeatureTTL
+      parseSeconds = A.withScientific "FeatureTTL" $
+        \s -> case toBoundedInteger s of
+          Just 0 -> pure FeatureTTLUnlimited
+          Just i -> pure . FeatureTTLSeconds $ i
+          Nothing -> A.parseFail "Expected an integer."
+
+      fromTTL :: FeatureTTL -> Maybe A.Value
```
fisx, line 244: Btw @pcapriotti why is this a `Maybe`?

pcapriotti: Good question. In order to support modular serialisation of sum types, serialisers can "fail" by returning `Nothing`. So for example, consider the serialiser side of how we build a schema for `A + B` from schemas `sa`, `sb` for `A` and `B` respectively. We first turn the serialiser of `sa` into a serialiser for `A + B` that simply fails in the `B` case, and similarly for `sb`. Then finally we merge them together using the appropriate `Monoid` instance for functions into `Maybe`. [This](https://github.com/wireapp/wire-server/blob/develop/libs/schema-profunctor/README.md#sum-types) has a bit more detail on the user-facing API of this mechanism.

libs/schema-profunctor/README.md
```diff
@@ -203,6 +203,50 @@ Finally, we add a name to the schema using the `named`
...
+        A.withText "FeatureTTL" $
+          \t ->
+            if t == "unlimited" || t == "0"
+              then pure FeatureTTLUnlimited
+              else A.parseFail "Expected ''unlimited' or '0'."
+
+      parseSeconds :: A.Value -> A.Parser FeatureTTL
+      parseSeconds = A.withScientific "FeatureTTL" $
+        \s -> case toBoundedInteger s of
+          Just 0 -> pure FeatureTTLUnlimited
```
fisx, line 240:
```suggestion
          Just 0 -> pure FeatureTTLUnlimited  -- this makes `decode . encode /= id`, but you get the idea.
```

fisx (approved):

PR #2596, opened 2022-08-05
Drop hpack

fisx:
We've discussed this in the past, with the following alternatives being tossed around:

- keep both package.yaml and cabal files
- keep only package.yaml files
- keep only cabal files

I am not completely certain, but I think the last had the least resistance from everybody, so I'll raise this in the standup on Monday, and if I get at least one approval and no objections, I'll merge this.


elland (approved):

smatting (approved):
Nice! I'd like to propose `cabal-fmt` as a replacement: https://github.com/wireapp/wire-server/pull/2601

PR #2601, opened 2022-08-09
Add cabal-fmt

smatting:
This PR adds `cabal-fmt` to the development environment as well as a Makefile target:

```
make cabal-fmt package=wire-api
```
which will reformat the corresponding `.cabal` file in-place. This PR also adds formatting to `make c` target.

If we add suitable stanzas like this (see [blog post](http://oleg.fi/gists/posts/2019-08-11-cabal-fmt.html#extra-expand-exposed-modules-and-other-modules))

```
-- cabal-fmt: expand src
```

then `cabal-fmt` will auto-fill modules like hpack does.

We can use this after we drop hpack in https://github.com/wireapp/wire-server/pull/2596

## Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -61,7 +61,7 @@ endif
 
 # Usage: make c package=brig test=1
 .PHONY: c
-c:
+c: cabal-fmt
 	cabal build $(WIRE_CABAL_BUILD_OPTIONS) $(package)
 ifeq ($(test), 1)
 	./hack/bin/cabal-run-tests.sh $(package) $(testargs)
--- a/Makefile
+++ b/Makefile
@@ -75,6 +75,10 @@ endif
 ci: c
 	./hack/bin/cabal-run-integration.sh $(package)
 
+.PHONY: cabal-fmt
+cabal-fmt:
+	./hack/bin/cabal-fmt.sh $(package)
+
 # Use ghcid to watch a particular package.
 # pass target=package:name to specify which target is watched.
 .PHONY: ghcid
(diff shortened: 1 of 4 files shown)


elland (approved):

nix/default.nix
```diff
@@ -107,6 +107,7 @@ let
 
     # For cabal-migration
     pkgs.haskellPackages.cabal-plan
+    pkgs.haskellPackages.cabal-fmt
```
fisx, line 110: this doesn't really belong under that comment two lines above, but close enough.

Makefile
```diff
@@ -75,6 +75,10 @@ endif
 ci: c
 	./hack/bin/cabal-run-integration.sh $(package)
 
+.PHONY: cabal-fmt
+cabal-fmt:
+	./hack/bin/cabal-fmt.sh $(package)
```
fisx, line 80: I wonder if there is a more automatic way to do this?  Maybe add this goal as a dependency of `default`?

fisx (approved):
also nice!  :)

nix/default.nix
smatting: Solved it by removning the comment. `cabal-plan` is not just used for migration but for our regular workflows

Makefile
smatting: Great idea. I've added the formatting to the `c` target as a dependency. Also the script will format all cabal files if invoked with `"all"`, e.g. when running `make c` without a `package=<>`

smatting: Unfortunately I can't add this to default `make install` target, because that runs on CI, which doesn't have `cabal-fmt` available in its environment

PR #2603, opened 2022-08-09
Reformat all cabal files with cabal-fmt

smatting:
This PR reformats all .cabal files in the project with `cabal-fmt`.
This is a followup to #2601

For the following packages `cabal-fmt` stanzas have been added to the cabal file (see [blog post](http://oleg.fi/gists/posts/2019-08-11-cabal-fmt.html#extra-expand-exposed-modules-and-other-modules)), which enables auto-fill of modules:

* brig
* galley
* spar
* wire-api
* wire-api-federation

A caveat of using `cabal-fmt` auto-fill functionality is that it also fills in `Main` modules, which creates problems:
-  Having main modules in `other-modules` creates warnings
```
Warning: Enabling workaround for Main module 'Main' listed in 'other-modules'
```
which is probably benign. For executables a solution is to move the main module outside of source directory. See `executable spar-schema` in `spar.cabal` in this PR for an example. 
- When `Spec.hs` modules from hspec get auto-filled in `other-modules` this breaks building, because filename and module name (which is always `Main` for hspec) doesn't match up. A workaround is renaming these modules to `Main.hs` or not to use cabal-fmt's autofill functionality

## Checklist


smatting:
@supersven I'm merging this early to avoid confusion for the other PRs. If you have suggestions I'll fix them in a follow-up PR

supersven:
> @supersven I'm merging this early to avoid confusion for the other PRs. If you have suggestions I can add them in a follow-up PR

Looks good to me :+1:

PR #2597, opened 2022-08-08
Serialisation of remove proposals

pcapriotti:
This PR adds serialisation of signed remove proposals.

https://wearezeta.atlassian.net/browse/FS-823

## Checklist

 - [x] The **PR Title** explains the impact of the change.
 - [x] The **PR description** provides context as to why the change should occur and what the code contributes to that effect. This could also be a link to a JIRA ticket or a Github issue, if there is one.
 - [ ] **changelog.d** contains the following bits of information ([details](https://github.com/wireapp/wire-server/blob/develop/docs/developer/changelog.md)):
   - [ ] A file with the changelog entry in one or more suitable sub-sections. The sub-sections are marked by directories inside `changelog.d`.
   - [ ] If new config options introduced: added usage description under docs/reference/config-options.md
   - [ ] If new config options introduced: recommended measures to be taken by on-premise instance operators.

PR #2604, opened 2022-08-10
update clients' key package refs upon committing

stefanwire:
https://wearezeta.atlassian.net/browse/FS-878

The logic has been manually tested. Automated tests will be added/integrated by FS-797.

## Checklist

### DIFF
--- a/services/brig/src/Brig/Data/MLS/KeyPackage.hs
+++ b/services/brig/src/Brig/Data/MLS/KeyPackage.hs
@@ -212,16 +212,17 @@ updateKeyPackageRef prevRef newRef =
 --------------------------------------------------------------------------------
 -- Utilities
 
-backupKeyPackageMeta :: MonadClient m => KeyPackageRef -> MaybeT m (ClientId, ConvId, Domain, Domain, UserId)
+backupKeyPackageMeta :: MonadClient m => KeyPackageRef -> MaybeT m (ClientId, Qualified ConvId, Qualified UserId)
 backupKeyPackageMeta ref = do
-  MaybeT . retry x1 $ query1 q (params LocalQuorum (Identity ref))
+  (clientId, convId, convDomain, userDomain, userId) <- MaybeT . retry x1 $ query1 q (params LocalQuorum (Identity ref))
+  pure (clientId, Qualified convId convDomain, Qualified userId userDomain)
   where
     q :: PrepQuery R (Identity KeyPackageRef) (ClientId, ConvId, Domain, Domain, UserId)
     q = "SELECT client, conv, conv_domain, domain, user FROM mls_key_package_refs WHERE ref = ?"
 
-restoreKeyPackageMeta :: MonadClient m => KeyPackageRef -> (ClientId, ConvId, Domain, Domain, UserId) -> m ()
-restoreKeyPackageMeta ref (client, conv, convDomain, domain, user) = do
-  write q (params LocalQuorum (ref, client, conv, convDomain, domain, user))
+restoreKeyPackageMeta :: MonadClient m => KeyPackageRef -> (ClientId, Qualified ConvId, Qualified UserId) -> m ()
+restoreKeyPackageMeta ref (clientId, convId, userId) = do
+  write q (params LocalQuorum (ref, clientId, qUnqualified convId, qDomain convId, qDomain userId, qUnqualified userId))
   where
     q :: PrepQuery W (KeyPackageRef, ClientId, ConvId, Domain, Domain, UserId) ()
     q = "INSERT INTO mls_key_package_refs (ref, client, conv, conv_domain, domain, user) VALUES (?, ?, ?, ?, ?, ?)"
--- a/services/brig/src/Brig/Data/MLS/KeyPackage.hs
+++ b/services/brig/src/Brig/Data/MLS/KeyPackage.hs
@@ -24,6 +24,7 @@ module Brig.Data.MLS.KeyPackage
     keyPackageRefConvId,
     keyPackageRefSetConvId,
     addKeyPackageRef,
+    updateKeyPackageRef,
   )
 where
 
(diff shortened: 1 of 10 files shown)


services/brig/src/Brig/Data/MLS/KeyPackage.hs
```diff
@@ -194,9 +195,47 @@ addKeyPackageRef ref nkpr = do
...
+updateKeyPackageRef :: MonadClient m => KeyPackageRef -> KeyPackageRef -> m ()
+updateKeyPackageRef prevRef newRef =
+  void . runMaybeT $
+    backupKeyPackageMeta prevRef
+      >>= ($) lift . ((>> deleteKeyPackage prevRef) . restoreKeyPackageMeta newRef)
+
 --------------------------------------------------------------------------------
 -- Utilities
 
+backupKeyPackageMeta :: MonadClient m => KeyPackageRef -> MaybeT m (ClientId, ConvId, Domain, Domain, UserId)
```
mdimjasevic, line 215: Can you regroup the return type so `Qualified` is used where appropriate? Seeing two occurrences of one and the same type (`Domain` in this case) makes me uneasy as it's really easy to confuse values of the same type.

services/brig/src/Brig/Data/MLS/KeyPackage.hs
```diff
@@ -194,9 +195,47 @@ addKeyPackageRef ref nkpr = do
...
+backupKeyPackageMeta ref = do
+  MaybeT . retry x1 $ query1 q (params LocalQuorum (Identity ref))
+  where
+    q :: PrepQuery R (Identity KeyPackageRef) (ClientId, ConvId, Domain, Domain, UserId)
+    q = "SELECT client, conv, conv_domain, domain, user FROM mls_key_package_refs WHERE ref = ?"
+
+restoreKeyPackageMeta :: MonadClient m => KeyPackageRef -> (ClientId, ConvId, Domain, Domain, UserId) -> m ()
```
mdimjasevic, line 222: I have the same note about the two occurrences of `Domain` in the tuple in the type signature as above.

services/brig/src/Brig/Data/MLS/KeyPackage.hs
```diff
@@ -194,9 +195,47 @@ addKeyPackageRef ref nkpr = do
...
+-- previous one.
+--
+-- FUTUREWORK: this function has to be extended if a table mapping (client,
+-- conversation) to key package ref is added, for instance, when implementing
+-- external delete proposals.
+updateKeyPackageRef :: MonadClient m => KeyPackageRef -> KeyPackageRef -> m ()
+updateKeyPackageRef prevRef newRef =
+  void . runMaybeT $
+    backupKeyPackageMeta prevRef
+      >>= ($) lift . ((>> deleteKeyPackage prevRef) . restoreKeyPackageMeta newRef)
```
mdimjasevic, line 210: I suppose you got this right, but wouldn't it be easier to read if it was written in the `do` notation?

services/galley/src/Galley/Effects/BrigAccess.hs
```diff
@@ -128,6 +129,7 @@ data BrigAccess m a where
   GetClientByKeyPackageRef :: KeyPackageRef -> BrigAccess m (Maybe ClientIdentity)
   GetLocalMLSClients :: Local UserId -> SignatureSchemeTag -> BrigAccess m (Set ClientId)
   AddKeyPackageRef :: KeyPackageRef -> Qualified UserId -> ClientId -> Qualified ConvId -> BrigAccess m ()
+  UpdateKeyPackageRef :: KeyPackageRef -> KeyPackageRef -> BrigAccess m ()
```
mdimjasevic, line 132: I know we need two key package refs here, but I don't like the fact they're of the same type, which makes it easy to confuse them. Would it make sense or would it be an overkill to introduce a wrapper like this?

```haskell
data Age = New | Old
data Aged t (a :: Age) where
  NewAged :: t -> Aged t 'New
  OldAged :: t -> Aged t 'Old
```

Then this effect action would look like this (and similarly for other places where there are two key package refs in the signature):
```haskell
  UpdateKeyPackageRef :: Aged KeyPackageRef 'Old -> Aged KeyPackageRef 'New -> BrigAccess m ()
```
Use sites would then mark it explicit which one is new, and which one is old:

```haskell
  ...
  updateKeyPackageRef (OldAged prevRef) (NewAged newRef)
```

services/galley/src/Galley/API/MLS/Message.hs
```diff
@@ -360,9 +360,11 @@ processCommit qusr con lconv epoch sender commit = do
         -- client (the creator)
         case (sender, first (toList . lmMLSClients) self) of
           (MemberSender ref, Left [creatorClient]) -> do
+            -- use update path as sender reference and if not existing fall back to sender
+            let senderRef = maybe ref (fromMaybe ref . kpRef' . upLeaf) $ cPath commit
```
mdimjasevic: Cannot this be collapsed into one `maybe` instead of a combination of `fromMaybe` and `maybe`? The fact you're using `ref` in two default cases suggests it is possible.

mdimjasevic (commented):
I think you're pretty much there. I left a few comments inlined.

mdimjasevic:
Can you also add tests? It would be good to have one for the case when there's no key package in the update path and then one for the case when there is a key package.

services/galley/src/Galley/API/MLS/Message.hs
mdimjasevic: Disregard this comment; I see you changed it in a subsequent commit.

services/brig/src/Brig/Data/MLS/KeyPackage.hs
stefanwire: I like pointfree, but that's not the point. I converted the expression to two kind-of statements in a `do` block.

stefanwire:
> Can you also add tests? It would be good to have one for the case when there's no key package in the update path and then one for the case when there is a key package.

I'm not sure whether we can do this using the `mls-test-cli`. There's always an update path in a commit from what I see. 🤔

mdimjasevic:
> I'm not sure whether we can do this using the `mls-test-cli`. There's always an update path in a commit from what I see. thinking

We're not there yet (or at least I haven't thought this through), but it would be nice if we didn't have to write an integration test for everything we do. With Galley polysemized, I think we ought to start thinking how to make testing easy without needing to reach for another integration test. For this particular scenario, perhaps it would be too much to construct by hand a binary representation of a commit so we can `parseMLS` it; with Paolo's serialisation PR #2597 , hopefully this will be a breeze.

[Not every commit](https://github.com/mlswg/mls-protocol/blob/838f0e15951b49af438bad501f3bd787798c2f4d/draft-ietf-mls-protocol.md#commit) has an update path - it is an optional field.

stefanwire:
> We're not there yet (or at least I haven't thought this through), but it would be nice if we didn't have to write an integration test for everything we do. With Galley polysemized, I think we ought to start thinking how to make testing easy without needing to reach for another integration test. For this particular scenario, perhaps it would be too much to construct by hand a binary representation of a commit so we can `parseMLS` it; with Paolo's serialisation PR #2597 , hopefully this will be a breeze.

Thanks for the pointer!

> [Not every commit](https://github.com/mlswg/mls-protocol/blob/838f0e15951b49af438bad501f3bd787798c2f4d/draft-ietf-mls-protocol.md#commit) has an update path - it is an optional field.

True and already acknowledged in code. 👍

services/galley/src/Galley/Effects/BrigAccess.hs
stefanwire: Done, but using a type with two fields. See [12c9dbe](https://github.com/wireapp/wire-server/pull/2604/commits/12c9dbe4b5864d355ad264943a209b212f68adc8).

services/brig/src/Brig/Data/MLS/KeyPackage.hs
stefanwire: Done. See [ea33455](https://github.com/wireapp/wire-server/pull/2604/commits/ea334556b017d02bd9643cd89caeb041e78ccce0).

services/brig/src/Brig/Data/MLS/KeyPackage.hs
stefanwire: Done. See [ea33455](https://github.com/wireapp/wire-server/pull/2604/commits/ea334556b017d02bd9643cd89caeb041e78ccce0).

mdimjasevic:
I really think the PR should come with tests. One option is to see if OpenMLS supports explicitly leaving out/including an update path. Another one, which I've been doing in another PR, is to construct the payload directly in Haskell; sure, that's a bit tedious to do by hand until we get Paolo's serialisation PR merged, but it should be doable. And you can always cherry-pick from Paolo's branch.

mdimjasevic (approved):
Not happy with this, but letting this to be merged without any tests.

On a more general note, testing changes to the application code ought to be easy. We should figure out how to do that easily with Polysemy.

PR #2609, opened 2022-08-11
Tweak nix env, cabal-fmt

fisx:
See commits for details.

## Checklist


fisx:
Thanks @smatting for the nix fix!

Would it be better to have a changelog entry for this?  I am leaning towards "no".

smatting (approved):
Nice. I always add a changelog entry because I assume otherwise the release manager has to come up with one. But if that's not the case I think it's fine to leave it out.

PR #2612, opened 2022-08-11
Bump mls-test-cli

smatting:
Updated mls-test-cli dependency, and the pin nixpkgs itself. Also removed the nixpkgs pin for building the docs.

This required some fiddling with rust dependencies (in particular the `ring` crate), since it seems to require the presence of a `.git` directory in order to build successfully. Hopefully this kludge can be removed eventually once the rust libraries mature.

## Checklist

### DIFF
--- a/nix/pkgs/mls_test_cli/default.nix
+++ b/nix/pkgs/mls_test_cli/default.nix
@@ -9,15 +9,18 @@
 
 rustPlatform.buildRustPackage rec {
   name = "mls-test-cli-${version}";
-  version = "0.2.0";
+  version = "0.3.0";
   nativeBuildInputs = [ pkg-config perl ];
   buildInputs = [ libsodium ];
   src = fetchFromGitHub {
     owner = "wireapp";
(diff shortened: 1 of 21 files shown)


nix/pkgs/mls_test_cli/default.nix
```diff
@@ -1,6 +1,7 @@
 { fetchFromGitHub
 , lib
 , libsodium
+, nasm
```
smatting, line 4: this should be removed, it's unnecessary

pcapriotti:
The problem seems to be that the patch is applied on the wrong directory (`source`, instead of the root where the vendor directory resides). Adding `..` to the patch path would fix this, but then the patch is rejected:
```
Ignoring potentially dangerous file name ../mls-test-cli-0.3.0-vendor.tar.gz/ring/.git/placeholder
```

pcapriotti:
Using a different hook (`cargoDepsHook`) instead of patch seems to have worked.

stephen-smith (approved):

PR #2605, opened 2022-08-10
SQSERVICES-1388-email-localization-when-provisioning-users-with-scim-2

battermann:
https://wearezeta.atlassian.net/browse/SQSERVICES-1388

The `preferredLanguage` field from SCIM now maps to the user locale in BRIG and will be set and updated on post SCIM user and on update SCIM user using SAML.

When the `preferredLangauge` is not set in SCIM it will fall back to the `defaultUserLocale` defined in brig. This also works for updates.

## Checklist

 - [x] The **PR Title** explains the impact of the change.
 - [x] The **PR description** provides context as to why the change should occur and what the code contributes to that effect. This could also be a link to a JIRA ticket or a Github issue, if there is one.
 - [x] **changelog.d** contains the following bits of information ([details](https://github.com/wireapp/wire-server/blob/develop/docs/developer/changelog.md)):
   - [x] A file with the changelog entry in one or more suitable sub-sections. The sub-sections are marked by directories inside `changelog.d`.

PR #2590, opened 2022-08-01
Brig Polysemy: Blacklist effect (cloned from Sandy's PR)

mdimjasevic:
(Cloned from Sandy's PR #2578.)

This PR exposes two new effects into Brig:

- `BlacklistStore`
- `BlacklistPhonePrefixStore`

previously housed in `Brig.Data.Blackist`.

In addition, it separates out the `CanonicalInterpreter` machinery into its own module, which is necessary for avoiding cyclical dependencies when doing large-scale effect work.

## Checklist

 - [x] The **PR Title** explains the impact of the change.
 - [x] The **PR description** provides context as to why the change should occur and what the code contributes to that effect. This could also be a link to a JIRA ticket or a Github issue, if there is one.
 - [x] **changelog.d** contains the following bits of information ([details](https://github.com/wireapp/wire-server/blob/develop/docs/developer/changelog.md)):
   - [x] A file with the changelog entry in one or more suitable sub-sections. The sub-sections are marked by directories inside `changelog.d`.

PR #2580, opened 2022-07-26
SQCORE-1166: Port Brig SearchAPI and UserRichInfo endpoints to Servant (CI copy)

mdimjasevic:
(This is a copy of @isovector 's PR #2577 so it runs in the CI.)

This PR finishes the work on SQCORE-1166, porting miscellaneous brig endpoints to servant.

## Checklist

 - [x] The **PR Title** explains the impact of the change.
 - [x] The **PR description** provides context as to why the change should occur and what the code contributes to that effect. This could also be a link to a JIRA ticket or a Github issue, if there is one.
 - [x] **changelog.d** contains the following bits of information ([details](https://github.com/wireapp/wire-server/blob/develop/docs/developer/changelog.md)):
   - [x] A file with the changelog entry in one or more suitable sub-sections. The sub-sections are marked by directories inside `changelog.d`.

PR #2579, opened 2022-07-25
charts/nginz: allow larger client requests.

jschaul:
Fix an issue for larger client requests on e.g. /list-users and /list-conversations, which were giving 413 errors for some users. 
See https://wearezeta.atlassian.net/browse/SQPIT-1044 
Allow client requests of 256k by default (was 64k).

## Checklist

 - [x] The **PR Title** explains the impact of the change.
 - [x] The **PR description** provides context as to why the change should occur and what the code contributes to that effect. This could also be a link to a JIRA ticket or a Github issue, if there is one.
 - [x] **changelog.d** contains the following bits of information ([details](https://github.com/wireapp/wire-server/blob/develop/docs/developer/changelog.md)):
   - [x] A file with the changelog entry in one or more suitable sub-sections. The sub-sections are marked by directories inside `changelog.d`.

PR #2614, opened 2022-08-16
Release 2022-08-16 - (expected chart version 4.22.0)

zebot:
# [2022-08-16] (Chart Release 4.22.0)

## API changes

* Drop the deprecated member removal endpoint (#2593)

## Features

* charts/cannon: Ensure HSTS headers are set for all endpoints (#2574)

* Expired MLS key packages are deleted from the database (#2582)

* Add support for MLS Remove proposals (#2561)

* Human readable names for SAML IdPs (#2565)

* The `preferredLanguage` field from SCIM now maps to the user locale in BRIG and will be set and updated on post SCIM user and on update SCIM user using SAML. (#2605)

* For TLS1.2, by default, remove ECDHE-ECDSA-AES128-GCM-SHA256 and ECDHE-RSA-AES128-GCM-SHA256 ciphers for ingress traffic. (#2528)

* Allow deleting existing splash screens in `PUT /teams/:tid (see also PR#2474 in Release 4.18.0) (#2588)
* Backoffice: Fix an issue where in some deployments ibis/galeb (Wire Cloud internal services) are unreachable from backoffice if deployed in a different namespace. (#2610)
* Fix an issue for larger client requests on e.g. /list-users and /list-conversations, which were giving 413 errors for some users. Allow client requests of 256k by default (was 64k). (#2579)
## Internal changes
* Add shellcheck, libstdc++ to nix env; handle emacs auto-save files better (#2609)
* Allow features to be set with HTTP method PATCH. This reflects a prior behavior
  that is used by Ibis. Additionally, it's more consistent when all setters can be
  called with PUT and PATCH. As this will fix calls by Ibis, the deployment order
  doesn't matter. (#2575)
* Brig Polysemization: introduce BlacklistStore and BlacklistPhonePrefixStore effects (#2590)
* Add cabal-fmt development tool (#2601)
* Reformat all cabal files with cabal-fmt (#2603)
* Delete tools: bonanza and makedeb (#2600)
* No more package.yaml / hpack, and stick with cabal files as the single (and only) source of truth (#2596)
* Port Brig SearchAPI and UserRichInfo endpoints to Servant (#2580)
* Added TTL data to stern feature flag GET endpoint. (#2564)
* Prepare removing deprecated non-binding teams (no more used in integration tests) (#2514, #2607)
* Add internal endpoint in Brig to update clients' key package refs in DB upon committing.
  Brig should be deployed before Galley. (#2604)
* Improved the resilience of provisioning new users via SAML by combining two persistence calls into one, preventing a creation failure from locking a user handle with no corresponding user. (#2526)
## Federation changes
* Fix TBS field in MLS Message type (#2599)


CHANGELOG.md
```diff
@@ -1,3 +1,75 @@
...
+* Port Brig SearchAPI and UserRichInfo endpoints to Servant (#2580)
+
+* Added TTL data to stern feature flag GET endpoint. (#2564)
+
+* Prepare removing deprecated non-binding teams (no more used in integration tests) (#2514, #2607)
+
+* Add internal endpoint in Brig to update clients' key package refs in DB upon committing.
+  Brig should be deployed before Galley. (#2604)
+
+* Improved the resilience of provisioning new users via SAML by combining two persistence calls into one, preventing a creation failure from locking a user handle with no corresponding user. (#PR_NOT_FOUND)
```
akshaymankar, line 64: Please find the PR number for this and remind the committer that they need to write this number in the changelog.d file if they don't want to use squash merges.

akshaymankar (changes requested):

CHANGELOG.md
supersven: I've added the missing PR numbers. Sorry for the delay, I attended to a meeting (Deep Backlock Grooming).

jschaul (approved):

CHANGELOG.md
jschaul: Thanks, looks good!

PR #2618, opened 2022-08-16
Add additional checks on incoming MLS messages

stefanwire:
https://wearezeta.atlassian.net/browse/FS-797

Removed the test `"user tries to remove themselves from conversation"`. In the test, the group creator was creating a _remove_ commit which was then send by Bob to the backend. It is not working anymore, since now it is made sure by the backend that Bob creates all his commits himself. Switching to Bob creating the _remove_ commit for his clients won't work either, since OpenMLS would forbid to create a commit in which the user even removes one of his/her own clients. Thus, the test was beyond repairable. However, there might be the option to craft a _remove_ commit for Bob without OpenMLS later on when crafting MLS messages is available.

## Checklist

### DIFF
--- a/services/galley/src/Galley/API/MLS/Message.hs
+++ b/services/galley/src/Galley/API/MLS/Message.hs
@@ -94,6 +94,7 @@ type MLSMessageStaticErrors =
      ErrorS 'MLSUnsupportedProposal,
      ErrorS 'MLSCommitMissingReferences,
      ErrorS 'MLSSelfRemovalNotAllowed,
+     ErrorS 'MLSClientSenderUserMismatch,
      ErrorS 'MLSGroupConversationMismatch
    ]
 
@@ -110,6 +111,7 @@ postMLSMessageFromLocalUserV1 ::
          ErrorS 'MLSSelfRemovalNotAllowed,
          ErrorS 'MLSStaleMessage,
          ErrorS 'MLSUnsupportedMessage,
+         ErrorS 'MLSClientSenderUserMismatch,
          ErrorS 'MLSGroupConversationMismatch,
          ErrorS 'MissingLegalholdConsent,
          Input (Local ()),
@@ -142,6 +144,7 @@ postMLSMessageFromLocalUser ::
          ErrorS 'MLSSelfRemovalNotAllowed,
          ErrorS 'MLSStaleMessage,
          ErrorS 'MLSUnsupportedMessage,
+         ErrorS 'MLSClientSenderUserMismatch,
          ErrorS 'MLSGroupConversationMismatch,
          ErrorS 'MissingLegalholdConsent,
          Input (Local ()),
@@ -175,6 +178,7 @@ postMLSMessage ::
          ErrorS 'MissingLegalholdConsent,
          ErrorS 'MLSCommitMissingReferences,
          ErrorS 'MLSSelfRemovalNotAllowed,
+         ErrorS 'MLSClientSenderUserMismatch,
          ErrorS 'MLSGroupConversationMismatch,
          Resource,
          TinyLog,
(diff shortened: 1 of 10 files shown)


services/galley/src/Galley/API/MLS/Message.hs
```diff
@@ -179,19 +179,45 @@ postMLSMessage ::
...
+    when (maybe False (qUnqualified qcnv /=) convId) $ throwS @'MLSUnsupportedMessage
+    unless (msgEpoch msg == Epoch 0) $
+      flip unless (throwS @'MLSUnsupportedMessage) =<< isUserSender qusr smsg
     foldQualified
       loc
       (postMLSMessageToLocalConv qusr con smsg)
       (postMLSMessageToRemoteConv loc qusr con smsg)
       qcnv
 
+isUserSender ::
```
elland, line 199: I'm unsure of what exactly is being tested here. We only care about the equalities for one specific use case, and default to `True` otherwise?

stefanwire: It's checking that the MLS client who created the message belongs to the user who is the sender of the REST request, identified by HTTP header. This is only relevant in an ongoing conversation. The check should be skipped in case of 
* `SMLSCipherText`: encrypted messages in which we don't have access to the sending client's key package, 
* `PreconfiguredSender`: messages sent by the backend, and
* `NewMemberSender`: external add proposals which propose fresh key packages for new clients and thus the validity of the key package cannot be known at the time of this check.

External add proposals are validated at a later stage of message processing and so will be messages sent by the backend. The latter is not in place yet, I think. There's not much the backend can do with/about encrypted messages other than passing them forward to the clients (and perhaps query the clients for news).

elland: Thanks, that's much clearer. Would it be good to add this summary as a comment on top of the function? Or point to the spec/some doc that clarifies this?

stefanwire: Yep, realised after writing my reply that a comment in code could do no harm to this function.

services/galley/src/Galley/API/MLS/Message.hs
```diff
@@ -179,19 +191,57 @@ postMLSMessage ::
...
+      -- skip message sent by backend
+      PreconfiguredSender _ -> pure True
+      -- skip external add proposal
+      NewMemberSender -> pure True
+      MemberSender ref -> do
+        ClientIdentity {ciUser, ciDomain} <- derefKeyPackage ref
+        pure $
+          ciUser == qUnqualified qusr
+            && ciDomain == qDomain qusr
+
```
pcapriotti, line 244: How about this?

```suggestion
        ci <- derefKeyPackage ref
        pure $ fmap fst (cidQualifiedClient ci) == qusr
```

services/galley/src/Galley/Cassandra/Conversation/Members.hs
```diff
@@ -321,6 +321,15 @@ filterRemoteConvMembers users (qUntagged -> Qualified conv dom) =
...
         $ query Cql.selectRemoteConvMembers (params LocalQuorum (user, dom, conv))
 
+lookupLocalMemberRemoteConv ::
+  UserId ->
+  Remote ConvId ->
+  Client [UserId]
+lookupLocalMemberRemoteConv userId (qUntagged -> Qualified conv dom) =
+  fmap (map runIdentity)
+    . retry x5
+    $ query Cql.selectRemoteConvMembers (params LocalQuorum (userId, dom, conv))
```
pcapriotti, line 331: This should probably use `query1`.

pcapriotti (approved):
Looks good. Minor comments below.

services/galley/src/Galley/Cassandra/Conversation/Members.hs
stefanwire: Nice catch! Fix done in https://github.com/wireapp/wire-server/pull/2618/commits/69c7365cb38a7b6b23f7ebfa819c6967b2ebbc6f.

PR #2620, opened 2022-08-16
Add TLS support to coturn Helm chart

sysvinit:
TODO(sysvinit): chart README, changelog.

## Checklist

### DIFF
--- a/charts/coturn/templates/certificate.yaml
+++ b/charts/coturn/templates/certificate.yaml
@@ -1,0 +1,55 @@
+{{- if .Values.tls.enabled }}
+{{- if .Values.tls.certManager.enabled }}
+apiVersion: cert-manager.io/v1
+kind: Certificate
+metadata:
+  name: coturn-letsencrypt-cert
+  namespace: {{ .Release.Namespace }}
+  labels:
+    chart: {{ .Chart.Name }}-{{ .Chart.Version | replace "+" "_" }}
+    release: "{{ .Release.Name }}"
+    heritage: "{{ .Release.Service }}"
+spec:
+  issuerRef:
+    name: coturn-letsencrypt-issuer
+    kind: {{ .Values.tls.certManager.issuer.kind }}
+  usages:
+    - server auth
+  duration: 2160h # 90 days, ignored by LE
+  renewBefore: 360h # 15 days
+  isCA: false
+  secretName: coturn-tls
+
+  privateKey:
+    algorithm: ECDSA
+    size: 384
+    encoding: PKCS1
+    rotationPolicy: Always
+
+  {{- if eq (len .Values.tls.certManager.dnsNames) 0 }}
+  {{- fail "List of DNS names for LetsEncrypt certificate is empty" }}
+  {{- else }}
+  dnsNames:
+  {{- with .Values.tls.certManager.dnsNames }}
+    - {{ . | quote }}
+  {{- end }}
+  {{- end }}
+{{- else }}
+apiVersion: v1
+kind: Secret
+metadata:
+  name: coturn-tls
+  labels:
+    chart: {{ .Chart.Name }}-{{ .Chart.Version | replace "+" "_" }}
+    release: "{{ .Release.Name }}"
+    heritage: "{{ .Release.Service }}"
+type: kubernetes.io/tls
+data:
+  {{- if and .Values.secrets.tls.cert .Values.secrets.tls.key }}
+  tls.crt: {{ .Values.secrets.tls.cert | b64enc | quote }}
+  tls.key: {{ .Values.secrets.tls.key | b64enc | quote }}
+  {{- else }}
+  {{- fail "TLS enabled, but cert and key are not provided in secrets.tls.key and secrets.tls.cert" }}
+  {{- end }}
+{{- end }}
+{{- end }}
--- a/charts/coturn/templates/issuer.yaml
+++ b/charts/coturn/templates/issuer.yaml
@@ -1,0 +1,31 @@
+{{- if and .Values.tls.enabled .Values.tls.certManager.enabled .Values.tls.certManager.issuer.create }}
+apiVersion: cert-manager.io/v1
+{{- if or (eq .Values.tls.certManager.issuer.kind "Issuer") (eq .Values.tls.certManager.issuer.kind "ClusterIssuer") }}
+kind: {{ .Values.tls.certManager.issuer.kind }}
+{{- else }}
+{{- fail "Unexpected TLS issuer kind; must be either Issuer or ClusterIssuer" }}
+{{- end }}
+
+metadata:
+  name: coturn-letsencrypt-issuer
+  {{- if eq .Values.tls.certManager.issuer.kind "Issuer" }}
+  namespace: {{ .Release.Namespace }}
+  {{- end }}
+  labels:
+    chart: {{ .Chart.Name}}-{{ .Chart.Version | replace "+" "_" }}
+    release: {{ .Release.Name | quote }}
+    heritage: {{ .Release.Service | quote }}
+
+spec:
+  acme:
+    server: 'https://acme-v02.api.letsencrypt.org/directory'
+    {{- if eq .Values.tls.certManager.issuer.certmasterEmail "" }}
+    {{- fail "LetsEncrypt account email address is empty" }}
+    {{- else }}
+    email: {{ .Values.tls.certManager.issuer.certmasterEmail | quote }}
+    {{- end }}
+    privateKeySecretRef:
+      name: coturn-letsencrypt-issuer-account-key
+    solvers:
+{{ toYaml .Values.tls.certManager.issuer.solverConfiguration | indent 6 }}
+{{- end }}
--- a/charts/coturn/values.yaml
+++ b/charts/coturn/values.yaml
@@ -24,6 +24,22 @@ securityContext:
 
 coturnTurnListenPort: 3478
 coturnMetricsListenPort: 9641
+coturnTurnTlsListenPort: 5349
+
+tls:
+  enabled: false
+  certManager:
+    enabled: false
+    dnsNames: []
+    issuer:
+      create: false
+      kind:
+      certmasterEmail:
+      solverConfiguration:
+  reloaderImage:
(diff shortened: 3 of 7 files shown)


charts/coturn/README.md
```diff
@@ -16,3 +17,12 @@ therefore recommended to run coturn on a separate Kubernetes cluster
...
 [this](https://docs.wire.com/understand/restund.html#network) page.
+
+coturn can optionally be configured to expose a TLS control
+port. The TLS private key and certificate to use for this port can
+be supplied manually in `secrets.tls.cert` and `secrets.tls.key`, or
+[cert-manager](https://cert-manager.io) can be used to provision certificates
+automatically. If cert-manager is in use, and no Issuer or ClusterIssuer is
+provided in the cluster already, then this chart can create an issuer using
+Let's Encrypt, though the configuration of ACME challenge solvers must be
+provided by the operator manually.
```
flokli, line 28: IMHO, this should be out of scope for our coturn chart.

Deploying and configuring cert-manager according to the environment is something that the cluster operator should do - we shouldn't create `Issuer` in the helm chart. Can you drop the end of the paragraph, and drop the creation of the `Issuer` resource?

charts/coturn/values.yaml
```diff
@@ -24,6 +24,23 @@ securityContext:
...
+  enabled: false
+  certManager:
+    enabled: false
+    dnsNames: []
+    issuer:
+      create: false
+      name:
+      kind:
+      certmasterEmail:
+      solverConfiguration:
```
flokli, line 39:
```suggestion
    issuerName: # name of the [Cluster]Issuer cert-manager should use
    issuerKind: ClusterIssuer # can be set to `Issuer` too
```

If we don't create the issuer anymore in the helm chart, this can become much simpler, as we only refer to an existing [Cluster]Issuer, rather than creating one (and having to provide values for all the bits and pieces that need to be configured)

charts/coturn/values.yaml
```diff
@@ -24,6 +24,23 @@ securityContext:
...
+  reloaderImage:
+    repository: quay.io/wire/config-reloader-sidecar
+    pullPolicy: IfNotPresent
+    tag: 1aa6cbbf2ce3a5182ec47e3579bbcb8f47e22fdc
```
flokli, line 43: This could have a comment where it's coming from, and what code it's using.

charts/coturn/templates/issuer.yaml
```diff
@@ -0,0 +1,31 @@
+{{- if and .Values.tls.enabled .Values.tls.certManager.enabled .Values.tls.certManager.issuer.create }}
```
flokli, line 1: Please drop this file, for the reasons described in the other comments.

charts/coturn/templates/certificate.yaml
```diff
@@ -0,0 +1,55 @@
...
+  name: coturn-auto-certificate
+  namespace: {{ .Release.Namespace }}
+  labels:
+    chart: {{ .Chart.Name }}-{{ .Chart.Version | replace "+" "_" }}
+    release: "{{ .Release.Name }}"
+    heritage: "{{ .Release.Service }}"
+spec:
+  issuerRef:
+    name: {{ .Values.tls.certManager.issuer.name }}
+    kind: {{ .Values.tls.certManager.issuer.kind }}
```
flokli, line 15:
```suggestion
    name: {{ .Values.tls.certManager.issuerName }}
    kind: {{ .Values.tls.certManager.issuerKind }}
```

charts/coturn/templates/certificate.yaml
```diff
@@ -0,0 +1,55 @@
...
+  usages:
+    - server auth
+  duration: 2160h # 90 days, ignored by LE
+  renewBefore: 360h # 15 days
+  isCA: false
+  secretName: coturn-tls
```
flokli, line 21: This should definitely become a value (with `coturn-dns` as a default), so people not using cert-manager can specify the path to a secret resource containing a cert.

charts/coturn/templates/certificate.yaml
```diff
@@ -0,0 +1,55 @@
...
+  {{- if eq (len .Values.tls.certManager.dnsNames) 0 }}
+  {{- fail "List of DNS names for automatically provisioned certificate is empty" }}
+  {{- else }}
+  dnsNames:
+  {{- with .Values.tls.certManager.dnsNames }}
+    - {{ . | quote }}
+  {{- end }}
+  {{- end }}
+{{- else }}
+apiVersion: v1
```
flokli, line 38: Please drop the else branch entirely. If the user doesn't want to use cert-manager, they should refer to a secret that's externally created (and updated).

The current state of things requires people to re-run helm whenever they rotate the cert manually, which is not very practical.

flokli (commented):
Two things:

 - creating a cert-manager Issuer should be out of scope for this helm chart
 - in case cert-manager is not used, the secret should be provided externally, not managed by Helm (and it's name passed as a helm value)

There's some more detailed comments about resulting simplifications.

charts/coturn/README.md
```diff
@@ -1,10 +1,11 @@
 **Warning**: this chart is currently considered alpha. Use at your own risk!
 
-This chart deploys [coturn](https://github.com/coturn/coturn), a STUN and TURN
+This chart deploys [coturn](https://github.com/wireapp/coturn), a STUN and TURN
```
flokli, line 3: IMHO, we should still link to upstream here, and maybe additional to our branch with the changes on top.

We intend to upstream all this one way or another, right?

sysvinit:
> Two things:
> 
>     * creating a cert-manager Issuer should be out of scope for this helm chart
> 
>     * in case cert-manager is not used, the secret should be provided externally, not managed by Helm (and it's name passed as a helm value)

I was just using the other Helm charts as prior art -- in particular, the `nginx-ingress-services` chart has `cert-manager` [integration](https://github.com/wireapp/wire-server/blob/3fbb76a73b49d67a80e002171620f5d63abb2413/charts/nginx-ingress-services/values.yaml#L37-L43) (including support for creating the Issuer, which is where I got the idea in the first place), [as does](https://github.com/wireapp/wire-server/blob/3fbb76a73b49d67a80e002171620f5d63abb2413/charts/restund/templates/secret-or-certificate.yaml#L7-L25) the `restund` chart. So, I'm just following what we have in the tree already -- is what we're doing already the wrong way to be doing this?

charts/coturn/README.md
sysvinit: > IMHO, we should still link to upstream here, and maybe additional to our branch with the changes on top.

That's maybe not a bad idea, I could be convinced of that.

> We intend to upstream all this one way or another, right?

I'm not sure that all of our changes are upstreamable (especially some of the authentication changes), and I don't otherwise have the time just now to run all our changes through the pipeline of proposing them to upstream.

flokli: I'm not saying we should upstream this now ;-)

I only think we should primarily link to upstream, and /then/ to our local branch with changes on top to convey this is some changes on top of upstream, but not its own project.

sysvinit: Fixed in [d7a8a72](https://github.com/wireapp/wire-server/pull/2620/commits/d7a8a72b122733da98a8a23b2b6403f489ba6b02).

charts/coturn/values.yaml
sysvinit: Fixed in [8d5f083](https://github.com/wireapp/wire-server/pull/2620/commits/8d5f0833018c14bf079967b3d18dc0dcac556832)

charts/coturn/templates/certificate.yaml
sysvinit: What do you mean, "this should become a value"? I don't understand what you're referring to. Are you suggesting that the certificate parameters should be exposed in the chart configuration `values.yaml`?

Is it worth even referencing cert-manager resources in this chart at all, if all we need is the name of a `Secret` containing the certificates, irrespective of whether it's managed manually by the operator or buy cert-manager?

flokli: Sorry, helm naming schemes are terrible. The idea was that instead of hardcoding `coturn-tls`, this parameter will become a helm value, or "variable".

And you're right, how these secrets are populated (either triggered by some cert-manager machinery via `Certificate` resources, or manually placed and rotated by an operator) should be entirely of scope for the helm chart. You just give it the name of the secret it tries to use, and that's it.

charts/coturn/README.md
sysvinit: Fixed in [3cae51b](https://github.com/wireapp/wire-server/pull/2620/commits/3cae51b9080c3c316e57fa93fe396418c348f4c9)

charts/coturn/values.yaml
sysvinit: Fixed in [3cae51b](https://github.com/wireapp/wire-server/pull/2620/commits/3cae51b9080c3c316e57fa93fe396418c348f4c9)

charts/coturn/templates/issuer.yaml
sysvinit: Done in [3cae51b](https://github.com/wireapp/wire-server/pull/2620/commits/3cae51b9080c3c316e57fa93fe396418c348f4c9)

charts/coturn/templates/certificate.yaml
sysvinit: Fixed in [3cae51b](https://github.com/wireapp/wire-server/pull/2620/commits/3cae51b9080c3c316e57fa93fe396418c348f4c9)

charts/coturn/templates/certificate.yaml
sysvinit: Done in [3cae51b](https://github.com/wireapp/wire-server/pull/2620/commits/3cae51b9080c3c316e57fa93fe396418c348f4c9)

charts/coturn/templates/certificate.yaml
sysvinit: Done in [3cae51b](https://github.com/wireapp/wire-server/pull/2620/commits/3cae51b9080c3c316e57fa93fe396418c348f4c9)

flokli (approved):

PR #2627, opened 2022-08-17
[WIP] building: check dynamic library dependencies are sane using ldd in

jschaul:
compile and docker building phase.

port ldd script over from hegemony and check for shared library in dynamic executable in build phase

In the past, we checked that binaries have no missing dependencies at
deploy time. This could happen when adding a compile-time dependency in
dockerfiles for builder and prebuilder but forgetting the runtime in the
dockerfile for the end container. This now does this check 

* [x] at build time during compilation, though usually compilation would fail, so this is perhaps not so useful 
* [ ] TODO at runtime-container build time

(rather than at deploy time. The earlier the better)

(see also https://github.com/zinfra/cailleach/pull/1317)

## Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -46,6 +46,7 @@ ifeq ($(WIRE_BUILD_WITH_CABAL), 1)
 	cabal build all
 	./hack/bin/cabal-run-all-tests.sh
 	./hack/bin/cabal-install-artefacts.sh all
+	./hack/bin/check_ldd.sh
 else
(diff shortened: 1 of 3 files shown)


flokli:
With the move to using Nix to compile all the binaries, we shouldn't need this anymore, right?

jschaul:
> With the move to using Nix to compile all the binaries, we shouldn't need this anymore, right?

I suppose this should be a non-issue now, closing.

PR #2628, opened 2022-08-18
[FS-721] Forward All Default Proposals

mdimjasevic:
The PR adapts the Wire server such that it simply forwards all MLS default proposal types, including those that it currently does not support.

Tracked by https://wearezeta.atlassian.net/browse/FS-721.

## Checklist

### DIFF
--- a/libs/wire-api/src/Wire/API/MLS/Message.hs
+++ b/libs/wire-api/src/Wire/API/MLS/Message.hs
@@ -39,6 +39,8 @@ module Wire.API.MLS.Message
     KnownFormatTag (..),
     verifyMessageSignature,
     mkRemoveProposalMessage,
+    mkAppAckProposalMessageFailable,
+    mkAppAckProposalMessage,
   )
 where
 
@@ -358,6 +360,39 @@ mkRemoveProposalMessage priv pub gid epoch ref = maybeCryptoError $ do
               tbsMsgAuthData = mempty,
               tbsMsgSender = PreconfiguredSender 0,
               tbsMsgPayload = ProposalMessage (mkRemoveProposal ref)
+            }
+  let sig = BA.convert $ sign priv pub (rmRaw tbs)
+  pure (Message tbs (MessageExtraFields sig Nothing Nothing))
+
+mkAppAckProposalMessage ::
+  SecretKey ->
+  PublicKey ->
+  GroupId ->
+  Epoch ->
+  KeyPackageRef ->
+  [MessageRange] ->
+  Maybe (Message 'MLSPlainText)
+mkAppAckProposalMessage priv pub gid epoch ref mrs =
+  maybeCryptoError $ mkAppAckProposalMessageFailable priv pub gid epoch ref mrs
+
+mkAppAckProposalMessageFailable ::
+  SecretKey ->
+  PublicKey ->
+  GroupId ->
+  Epoch ->
+  KeyPackageRef ->
+  [MessageRange] ->
+  CryptoFailable (Message 'MLSPlainText)
+mkAppAckProposalMessageFailable priv pub gid epoch ref mrs = do
+  let tbs =
+        mkRawMLS $
+          MessageTBS
+            { tbsMsgFormat = KnownFormatTag,
+              tbsMsgGroupId = gid,
+              tbsMsgEpoch = epoch,
+              tbsMsgAuthData = mempty,
+              tbsMsgSender = MemberSender ref,
+              tbsMsgPayload = ProposalMessage (mkAppAckProposal mrs)
             }
   let sig = BA.convert $ sign priv pub (rmRaw tbs)
   pure (Message tbs (MessageExtraFields sig Nothing Nothing))
--- a/libs/wire-api/src/Wire/API/MLS/Message.hs
+++ b/libs/wire-api/src/Wire/API/MLS/Message.hs
@@ -39,8 +39,6 @@ module Wire.API.MLS.Message
     KnownFormatTag (..),
     verifyMessageSignature,
     mkRemoveProposalMessage,
-    mkAppAckProposalMessageFailable,
-    mkAppAckProposalMessage,
   )
 where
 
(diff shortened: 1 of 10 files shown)


libs/wire-api/src/Wire/API/MLS/Message.hs
```diff
@@ -361,3 +363,36 @@ mkRemoveProposalMessage priv pub gid epoch ref = maybeCryptoError $ do
             }
   let sig = BA.convert $ sign priv pub (rmRaw tbs)
   pure (Message tbs (MessageExtraFields sig Nothing Nothing))
+
+mkAppAckProposalMessage ::
```
pcapriotti, line 367: This isn't used, is it?

libs/wire-api/src/Wire/API/MLS/Message.hs
```diff
@@ -361,3 +363,36 @@ mkRemoveProposalMessage priv pub gid epoch ref = maybeCryptoError $ do
...
+  PublicKey ->
+  GroupId ->
+  Epoch ->
+  KeyPackageRef ->
+  [MessageRange] ->
+  Maybe (Message 'MLSPlainText)
+mkAppAckProposalMessage priv pub gid epoch ref mrs =
+  maybeCryptoError $ mkAppAckProposalMessageFailable priv pub gid epoch ref mrs
+
+mkAppAckProposalMessageFailable ::
```
pcapriotti, line 378: This is only used for testing, couldn't it just be moved to the tests?

libs/wire-api/src/Wire/API/MLS/Serialisation.hs
```diff
@@ -81,6 +82,16 @@ parseMLSVector getItem = do
...
       (:) x <$> (if pos < endPos then go endPos else pure [])
 
+serialiseMLSVector ::
+  forall w a.
+  (Binary w, Integral w) =>
+  (a -> Put) ->
+  [a] ->
+  Put
+serialiseMLSVector p as = do
+  put @w . fromIntegral . length $ as
```
pcapriotti, line 92: This is not correct. Vectors begin with the whole size of their binary representation, not the number of items. See https://www.rfc-editor.org/rfc/rfc5246#section-4.3.

services/galley/test/integration/API/MLS.hs
```diff
@@ -1421,3 +1430,46 @@ testPublicKeys = do
           (unMLSPublicKeys keys)
       )
       @?= [Ed25519]
+
+propUnsupported :: TestM ()
```
pcapriotti, line 1434: This is a somewhat unconventional test. Maybe add a couple of lines of comment explaining what it is we're doing here, like the fact that we are manually reading mls-test-cli's store and extracting the private key.

libs/wire-api/src/Wire/API/MLS/Message.hs
smatting: Maybe this function and `mkRemoveProposalMessage` can be factored to use a common function that signs a `MessageTBS` to `(Message 'MLSPlainText)`.
Both functions use `maybeCryptoError` together with a `pure` which is just [CryptoPassed](https://hoogle.zinfra.io/file/root/.cabal/store/ghc-8.10.7/cryptonite-0.29-28e7ebd201e4a5f09d307c6367c2ccfcc28816639f78d8c6cbc11e17376be535/share/doc/html/src/Crypto.Error.Types.html#CryptoPassed). I think in both functions nothing is acutally checked.

libs/wire-api/src/Wire/API/MLS/Message.hs
mdimjasevic: True. I removed it in [39d7c31](https://github.com/wireapp/wire-server/pull/2628/commits/39d7c3171b399a4f6aa0752429035ba85d6f0aa9).

libs/wire-api/src/Wire/API/MLS/Serialisation.hs
mdimjasevic: Thanks for spotting! Fixed in [cc6548a](https://github.com/wireapp/wire-server/pull/2628/commits/cc6548a051b3ba565478cf6bdbd8674da9f33a6d).

libs/wire-api/src/Wire/API/MLS/Serialisation.hs
```diff
@@ -81,6 +82,19 @@ parseMLSVector getItem = do
...
+  (Binary w, Integral w) =>
+  (a -> Put) ->
+  [a] ->
+  Put
+serialiseMLSVector p as = do
+  -- every element is long the same number of bytes when serialised. If there is
+  -- no single element in the vector, the whole length is 0 anyways.
+  let elLen = maybe 0 (LBS.length . runPut . p) . listToMaybe $ as
+  put @w . fromIntegral $ elLen * fromIntegral (length as)
+  traverse_ p as
```
pcapriotti, line 96: This doesn't look right to me. Why is every element the same length?

To be fair, it's not easy to do this efficiently. As a compromise, we can do the following:
 - serialise the whole vector ignoring the size
 - convert to lazy bytestring
 - put count
 - put bytestring

To get good performance we need imperative access to the buffer, which is probably possible, but requires the unsafe API. Maybe the above is good enough.

pcapriotti (changes requested):
Looks good except for the vector serialisation. See comment below.

libs/wire-api/src/Wire/API/MLS/Serialisation.hs
mdimjasevic: Good points! I obviously overlooked at this.

I changed the implementation and round-trip tests for `serialiseMLSVector` pass. I am not too familiar with how efficient the conversion from a lazy bytestring to a strict bytestring is, but that's what I did.

libs/wire-api/src/Wire/API/MLS/Serialisation.hs
```diff
@@ -81,6 +83,17 @@ parseMLSVector getItem = do
...
+serialiseMLSVector ::
+  forall w a.
+  (Binary w, Integral w) =>
+  (a -> Put) ->
+  [a] ->
+  Put
+serialiseMLSVector p =
+  serialiseMLSBytes @w . LBS.toStrict . toLazyByteString
+    . execPut
+    . traverse_ p
```
pcapriotti, line 95: This looks correct, but we can avoid the bytestring conversion if we create a version of `serialiseMLSBytes` that takes a lazy bytestring and uses `putLazyByteString`.

pcapriotti (approved):

libs/wire-api/src/Wire/API/MLS/Serialisation.hs
pcapriotti: Lazy bytestrings are essentially lists of strict bytestrings (represent their concatenation). Converting from lazy to strict amounts to actually performing the concatenation (i.e. basically allocating a big buffer, copying, and likely leaving the original bytestrings to the GC). So it's basically a bunch of extra copies and allocations, good to avoid if possible.

libs/wire-api/src/Wire/API/MLS/Serialisation.hs
mdimjasevic: Thanks again! I've implemented that in https://github.com/wireapp/wire-server/pull/2628/commits/8ce0da0ae3091f8fd667b2ed124a509b3b381097.

PR #2629, opened 2022-08-18
Fix broken hls-hlint-plugin in nix env

smatting:
This PR applies this suggestion https://github.com/NixOS/nixpkgs/issues/168064#issuecomment-1146691884
in a wire's nixpkgs fork. Since the problem has been fixed in https://github.com/haskell/haskell-language-server/pull/2854 we can switch to nixos/nixpkgs once a new HLS version is released an available in nixpkgs.

## Checklist

### DIFF
--- a/changelog.d/5-internal/hls-hlint-plugin-workaround
+++ b/changelog.d/5-internal/hls-hlint-plugin-workaround
@@ -1,0 +1,1 @@
+Fix broken hls-hlint-plugin in nix env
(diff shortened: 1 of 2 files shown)


flokli:
Nixpkgs PR: https://github.com/NixOS/nixpkgs/pull/187284

smatting:
I tested again: it works! Squiggly lines

PR #2631, opened 2022-08-19
Improve MLS client checks

pcapriotti:
This PR changes the way MLS-capable clients are defined to exclude those that have a public key set, but no key packages. Furthermore, adding a client to an MLS conversation that already includes the user results in the client check being skipped.

Tracked by https://wearezeta.atlassian.net/browse/FS-853.

## Checklist

### DIFF
--- a/changelog.d/0-release-notes/get-mls-clients
+++ b/changelog.d/0-release-notes/get-mls-clients
@@ -1,1 +1,1 @@
-The `GET mls/clients` has been changed, and it now returns a pair of lists, instead of one
+The internal endpoint `GET i/mls/clients` has been changed, and it now returns a pair of lists, instead of one
--- a/changelog.d/0-release-notes/get-mls-clients
+++ b/changelog.d/0-release-notes/get-mls-clients
@@ -1,0 +1,1 @@
+The `GET mls/clients` has been changed, and it now returns a pair of lists, instead of one
--- a/changelog.d/0-release-notes/get-mls-clients
+++ b/changelog.d/0-release-notes/get-mls-clients
@@ -1,1 +1,1 @@
-The internal endpoint `GET i/mls/clients` has been changed, and it now returns a pair of lists, instead of one
+The internal endpoint `GET i/mls/clients` has been changed, and it now returns a list of `ClientInfo` instead of a list of `ClientId`.
--- a/libs/wire-api-federation/src/Wire/API/Federation/API/Brig.hs
+++ b/libs/wire-api-federation/src/Wire/API/Federation/API/Brig.hs
@@ -68,7 +68,7 @@ type BrigApi =
     -- (handles can be up to 256 chars currently)
     :<|> FedEndpoint "search-users" SearchRequest SearchResponse
     :<|> FedEndpoint "get-user-clients" GetUserClients (UserMap (Set PubClient))
-    :<|> FedEndpoint "get-mls-clients" MLSClientsRequest (Set ClientId)
+    :<|> FedEndpoint "get-mls-clients" MLSClientsRequest (Set ClientId, Set ClientId)
     :<|> FedEndpoint "send-connection-action" NewConnectionRequest NewConnectionResponse
     :<|> FedEndpoint "on-user-deleted-connections" UserDeletedConnectionsNotification EmptyResponse
     :<|> FedEndpoint "claim-key-packages" ClaimKeyPackageRequest (Maybe KeyPackageBundle)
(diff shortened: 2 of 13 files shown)


libs/wire-api-federation/src/Wire/API/Federation/API/Brig.hs
```diff
@@ -68,7 +68,7 @@ type BrigApi =
     -- (handles can be up to 256 chars currently)
     :<|> FedEndpoint "search-users" SearchRequest SearchResponse
     :<|> FedEndpoint "get-user-clients" GetUserClients (UserMap (Set PubClient))
-    :<|> FedEndpoint "get-mls-clients" MLSClientsRequest (Set ClientId)
+    :<|> FedEndpoint "get-mls-clients" MLSClientsRequest (Set ClientId, Set ClientId)
```
mdimjasevic, line 71: This is a breaking federation API change. Should it be versioned, i.e., should it go into a new version 2?

services/galley/src/Galley/API/MLS/Message.hs
```diff
@@ -768,10 +781,14 @@ getMLSClients ::
   Local x ->
   Qualified UserId ->
   SignatureSchemeTag ->
-  Sem r (Set ClientId)
+  Sem r (Set ClientId, Set ClientId)
```
mdimjasevic, line 784: I keep on being confused and worried about pair types with the same type in each component, i.e., `(a, a)`. While reviewing this PR I already went a few times back and forth to check what is assigned to what component of the pair, and I find it to be using time, while it shouldn't. Do you have a suggestion on how to make them separate types so to avoid confusing the components easily, yet such that is still easy to use?

services/galley/test/integration/API/MLS.hs
```diff
@@ -406,9 +406,21 @@ testAddUserWithProteusClients = do
 
 testAddUserPartial :: TestM ()
 testAddUserPartial = do
-  (creator, commit) <- withSystemTempDirectory "mls" $ \tmp -> do
+  (creator, commit) <- withSystemTempDirectory "mls" $ \_tmp -> do
+    let tmp = "/tmp/mls"
```
mdimjasevic, line 410: This manual assignment of `tmp` looks like a leftover from debugging.

changelog.d/0-release-notes/get-mls-clients
```diff
@@ -0,0 +1 @@
+The `GET mls/clients` has been changed, and it now returns a pair of lists, instead of one
```
mdimjasevic, line 1: Isn't this actually an internal endpoint, i.e., `GET /i/mls/clients`?

mdimjasevic (approved):
Overall, this looks good, but there are a few small things I suggest changing; comments are inlined.

One thing that was confusing for me is the distinction between MLS clients and all clients in this PR. I believe both of them in the context of this PR refer to MLS clients (and not Proteus clients), but instead when you say "MLS clients" you actually mean MLS clients with at least one key package available, while by all clients you mean all MLS clients, including those that don't have a key package available. Is that right?

libs/wire-api-federation/src/Wire/API/Federation/API/Brig.hs
pcapriotti: Fortunately, we don't yet have a stable federation API version.

services/galley/src/Galley/API/MLS/Message.hs
pcapriotti: It's two subsets, ordered by inclusion. I thought that would be the most intuitive way to represent this. But you're probably right, we need to create a new type.

pcapriotti: I turned it into a single set of `ClientInfo`, where `ClientInfo` is essentially `(ClientId, Bool)`.

services/galley/test/integration/API/MLS.hs
pcapriotti: Fixed

changelog.d/0-release-notes/get-mls-clients
pcapriotti: Fixed

pcapriotti:
> Overall, this looks good, but there are a few small things I suggest changing; comments are inlined.
> 
> One thing that was confusing for me is the distinction between MLS clients and all clients in this PR. I believe both of them in the context of this PR refer to MLS clients (and not Proteus clients), but instead when you say "MLS clients" you actually mean MLS clients with at least one key package available, while by all clients you mean all MLS clients, including those that don't have a key package available. Is that right?

- "MLS clients" means, like you wrote, clients with at least 1 key package
- "all clients" literally means all, including ones that are not MLS capable.

services/galley/src/Galley/API/MLS/Message.hs
```diff
@@ -615,15 +615,19 @@ executeProposalAction qusr con lconv action = do
   foldQualified lconv (\_ -> pure ()) (\_ -> throwS @'MLSUnsupportedProposal) qusr
 
   -- check that all clients of each user are added to the conversation
```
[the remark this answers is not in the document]
pcapriotti: Reworded.

PR #2633, opened 2022-08-19
SQSERVICES-1645-backend-servantify-brig-auth-api

battermann:
## Checklist

### DIFF
--- a/libs/wire-api/test/unit/Test/Wire/API/User/Auth.hs
+++ b/libs/wire-api/test/unit/Test/Wire/API/User/Auth.hs
@@ -1,0 +1,39 @@
+-- This file is part of the Wire Server implementation.
+--
+-- Copyright (C) 2022 Wire Swiss GmbH <opensource@wire.com>
+--
+-- This program is free software: you can redistribute it and/or modify it under
+-- the terms of the GNU Affero General Public License as published by the Free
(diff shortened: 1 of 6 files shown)


libs/wire-api/test/unit/Test/Wire/API/User/Auth.hs
```diff
@@ -0,0 +1,39 @@
...
+loginIdHappyCase = testCase "LoginId parser: valid email" $ do
+  let actual :: Maybe LoginId = Aeson.decode "{\"email\":\"foo@bar.com\"}"
+  let expected = Just $ LoginByEmail (Email {emailLocal = "foo", emailDomain = "bar.com"})
+  assertEqual "should succeed" expected actual
+
+loginIdFailFast :: T.TestTree
+loginIdFailFast = testCase "LoginId parser: invalid email, valid phone" $ do
+  let actual :: Maybe LoginId = Aeson.decode "{\"email\":\"invalid-email\",\"phone\":\"+123456789\"}"
+  let expected = Nothing
+  assertEqual "should fail if any provided login id is invalid" expected actual
```
fisx, line 39: nice tests!   now you could also do the same for a broken phone number and correct email, to make sure we're actually running all parsers before picking a valid value.  (not crazy important, but it's easy and nice.)

fisx (approved):
looks good!

PR #2637, opened 2022-08-22
Make deletions via SCIM more stable

supersven:
This story gave me some headache: I tried to make things better without adding new fields/tables or expensive queries / indexes. And, had to keep in mind that data, e.g. email addresses or handles may be reused by another account.

Changes:
- `brig` accounts are now deleted in a synchronous call (`/i/users/:uid/ensure-deleted`). This way the SCIM caller should at least get an error when something in `brig` went wrong.
- The endpoint can be re-called to ensure that a brig account is deleted. If there is still dangling data around, it re-runs the deletion.
- Some data may be re-used; e.g. handles or email addresses. These are deleted first. We cannot access them (without expensive queries / indexes) after the user *tombstone* has been written. So, the *tombstone* is written after they were deleted. (It has been like that before, however it's an important detail to keep / understand.) 
- `spar` depends on user data from `brig` (`veid`) to delete it's user data. To ensure this is available when needed, `spar` data is deleted before `brig` data.

All this could have been much nicer without the query restrictions of Cassandra. :thinking: 

## Checklist

 - [X] ~~If HTTP endpoint paths have been added or renamed, or feature configs have changed, the **endpoint / config-flag checklist** (see Wire-employee only backend [wiki page](https://github.com/zinfra/backend-wiki/wiki/Checklists)) has been followed.~~ (Discussed in `Backend` channel.)


libs/wire-api/src/Wire/API/User.hs
```diff
@@ -1356,6 +1358,41 @@ instance FromJSON DeletionCodeTimeout where
   parseJSON = A.withObject "DeletionCodeTimeout" $ \o ->
     DeletionCodeTimeout <$> o A..: "expires_in"
 
+data EnsureAccountDeletedResult = NoUser | AccountAlreadyDeleted | AccountDeleted
+  deriving (Eq, Show)
+  deriving (S.ToSchema) via (Schema EnsureAccountDeletedResult)
+
+instance ToSchema EnsureAccountDeletedResult where
```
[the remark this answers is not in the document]
supersven: After discussing this in `backend-core`, I'd like to have a JSON object with a field as response (not just the plain JSON string value).

I just don't know how to get it from a value with e.g. `ToSchema`. Any hints?

services/brig/src/Brig/Data/UserKey.hs
```diff
@@ -164,6 +165,13 @@ deleteKey k = do
   retry x5 $ write deleteHashed (params LocalQuorum (Identity hk))
   retry x5 $ write keyDelete (params LocalQuorum (Identity $ keyText k))
 
+deleteKeyForUser :: (MonadClient m, MonadReader Env m) => UserId -> UserKey -> m ()
+deleteKeyForUser uid k = do
+  mbKeyUid <- lookupKey k
+  case mbKeyUid of
+    Just keyUid | keyUid == uid -> deleteKey k
+    _ -> pure ()
```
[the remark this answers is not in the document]
supersven: Yes, this is unfortunately *delete it for this user, iff it's still around*.

Depending on where a previous deletion attempt was aborted, the key may be around or not. So, the whole deletion process has to continue in both cases. I saw no value in returning a *success*/*fail* when the outer process would ignore the result anyways.

services/brig/test/integration/API/User/Account.hs
```diff
@@ -1632,18 +1640,104 @@ testTooManyMembersForLegalhold opts brig = do
...
+      ( brig
+          . paths ["/i/users", toByteString' uid, "ensure-deleted"]
+      )
+      !!! do
+        const 200 === statusCode
+        const (Right AccountAlreadyDeleted) === responseJsonEither
+
+testEnsureAccountDeletedWithNoUser :: Brig -> Http ()
+testEnsureAccountDeletedWithNoUser brig =
+  case parseIdFromText "19166173-49eb-49bc-962f-72c95f27a428" of
```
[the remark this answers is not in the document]
supersven: Ah, I think I see what you mean: I could get rid of this `case` expression with predictable outcome, when I just send the UUID string. :bulb:

libs/wire-api/src/Wire/API/User.hs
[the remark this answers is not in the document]
elland: @pcapriotti maybe?

services/brig/test/integration/API/User/Account.hs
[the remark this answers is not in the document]
elland: But also, if we can send garbled strings, reduces the infinitesimal chance of accidentally hitting a real UUID, but that's way too out there :D

fisx (commented):
I've made two commits simplifying things, but I think there is more:
- can't we just modify the old `Delete` effect instead of removing it and adding a different one?  semantics is still the same, no?
- similarly, can we modify the internal delete end-point in brig that was used before, rather than adding a new one?

while i was changing this, i found another instance where users are deleted on brig (namely when the idp is deleted and all the accounts in it are deleted with it).  have you intentionally not updated that?  why?

supersven:
> can't we just modify the old Delete effect instead of removing it and adding a different one? semantics is still the same, no?

@fisx , yes, your two commits regarding this seem to be perfectly reasonable. :+1: 

> similarly, can we modify the internal delete end-point in brig that was used before, rather than adding a new one?

This question follows from your changes in `spar` (previously it was used by the function you deleted). It sounds like a good idea. :+1:
As there's no response body in `delete "/i/users/:uid"`, this change should be fine. (Operationally, the call would become synchronous instead of asynchronous. However, I cannot think of any issues regarding this.)

Speaking of *synchronous vs. asynchronous*: Haven you seen that the prior behavior was asynchronous (with an event) and this is now synchronous? Is that fine with you? 

> while i was changing this, i found another instance where users are deleted on brig (namely when the idp is deleted and all the accounts in it are deleted with it). have you intentionally not updated that? why?

I think I wasn't brave enough and a bit overwhelmed by the implicit relationships of side-effects. So, I kept my change set non invasive. Next time, I should be braver... :smile:

libs/wire-api/src/Wire/API/User.hs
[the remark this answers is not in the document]
supersven: I've agreed with @fisx that I just add to / change the old internal deletion endpoint (`wai-routes`). So, this instance isn't needed, yet. (Probably, later, when we'll servantify the endpoint.)

services/brig/src/Brig/API/Internal.hs
```diff
@@ -475,14 +474,20 @@ createUserNoVerifySpar uData =
 
 deleteUserNoVerifyH :: UserId -> (Handler r) Response
 deleteUserNoVerifyH uid = do
```
elland, line 476: What is the meaning of "no verify" in this case? Seems like we're doing some verification. Also should it make sense to retry internally a few times until we get "already deleted"?

services/brig/src/Brig/API/Internal.hs
```diff
@@ -475,14 +474,20 @@ createUserNoVerifySpar uData =
...
 
-deleteUserNoVerify :: UserId -> (Handler r) ()
-deleteUserNoVerify uid = do
-  void $
-    lift (wrapClient $ API.lookupAccount uid)
-      >>= ifNothing (errorToWai @'E.UserNotFound)
-  lift $ API.deleteUserNoVerify uid
+--  setStatus status202 empty <$ deleteUserNoVerify uid
+
+--deleteUserNoVerify :: UserId -> (Handler r) ()
```
elland: Marked for deletion? :)

services/brig/src/Brig/API/Internal.hs
supersven: I haven't invented the name (only recycled :recycle: :wink:). However, to me it means that there is no verification that you're allowed to do so. I.e. when you call this with a valid `UserId`, the user will be deleted without any additional checks/constraints.

supersven: As I've already got the code to check if a user has been deleted completely, a retry mechanism sounds quite reasonable. :thinking:

supersven:
@elland , @fisx , Thanks a lot for your feedback!

Unfortunately, I had to change the behavior a bit: I realized that being able to delete a (potentially partially deleted) user without comparing it's brig user's teamId with the teamId of the request token is dangerous: A malicious user could delete users of other teams. (https://github.com/wireapp/wire-server/pull/2637/commits/e2a6cb7f8c66c6d2de80dc53ce2cc792d30e9b05)

services/brig/src/Brig/API/Internal.hs
fisx: > What is the meaning of "no verify" in this case? Seems like we're doing some verification.

you're right, it would be an improvement to rename this: `s/NoVerify/NoAuth/` and/or add a comment that this is the handler without the caller authentication/authorization part.

>  Also should it make sense to retry internally a few times until we get "already deleted"?

This handler queues the deletion for async procesing; IIRC (please double-check), the entry will be removed from the queue at the end of the deletion, so retry would merely trigger if the queue is somehow broken.  I don't think that's needed.

supersven: @fisx I'll do the renaming. However, the processing isn't asynchronous anymore. I changed that so synchronous to get a result. (To be able to tell the SCIM user if their request failed.)

supersven:
> @elland , @fisx , Thanks a lot for your feedback!
> 
> Unfortunately, I had to change the behavior a bit: I realized that being able to delete a (potentially partially deleted) user without comparing it's brig user's teamId with the teamId of the request token is dangerous: A malicious user could delete users of other teams. ([e2a6cb7](https://github.com/wireapp/wire-server/commit/e2a6cb7f8c66c6d2de80dc53ce2cc792d30e9b05))

Sorry, I think I panicked a bit too early: The ability to delete an *already deleted* or *non-existent* user should be harmless. :thinking: 
(I'm in touch with @fisx to discuss this.)

fisx:
> > @elland , @fisx , Thanks a lot for your feedback!
> > Unfortunately, I had to change the behavior a bit: I realized that being able to delete a (potentially partially deleted) user without comparing it's brig user's teamId with the teamId of the request token is dangerous: A malicious user could delete users of other teams. ([e2a6cb7](https://github.com/wireapp/wire-server/commit/e2a6cb7f8c66c6d2de80dc53ce2cc792d30e9b05))
> 
> Sorry, I think I panicked a bit too early: The ability to delete an _already deleted_ or _non-existent_ user should be harmless. thinking (I'm in touch with @fisx to discuss this.)



i agree, reverting e2a6cb7 should be safe!

libs/wire-api/src/Wire/API/User.hs
```diff
@@ -1356,6 +1358,31 @@ instance FromJSON DeletionCodeTimeout where
...
+      toTag NoUser = "no-user"
+      toTag AccountAlreadyDeleted = "already-deleted"
+      toTag AccountDeleted = "deleted"
+
+instance FromJSON DeleteUserResult where
+  parseJSON (A.Object o) = do
+    tagString <- o A..: "tag"
+    case fromTag tagString of
+      Just t -> pure t
+      Nothing -> A.parseFail $ "Unknown tag: " ++ tagString
```
fisx, line 1377:
```suggestion
    maybe (A.parseFail $ "Unknown tag: " ++ tagString) pure (fromTag tagString)
```

I'm not going to tell you which version I like better, just wanted to see for myself.  :)

services/brig/src/Brig/API/Internal.hs
```diff
@@ -472,16 +472,13 @@ createUserNoVerifySpar uData =
...
-deleteUserNoVerify uid = do
-  void $
-    lift (wrapClient $ API.lookupAccount uid)
-      >>= ifNothing (errorToWai @'E.UserNotFound)
-  lift $ API.deleteUserNoVerify uid
+deleteUserNoAuthH :: UserId -> (Handler r) Response
+deleteUserNoAuthH uid = do
+  r <- lift $ wrapHttp $ API.ensureAccountDeleted uid
+  pure $ case r of
+    NoUser -> setStatus status404 $ json r
```
fisx, line 515: Is this different from `errorToWai @'E.UserNotFound`?  I think it is, and I `r` is rendered into a top-level string, correct?  If so, does any of the call sides care?   And can we not have top-level string bodies?

To be more specific, can we keep `errorToWai @'E.UserNotFound` in the `NoUser` case and respond with empty both otherwise, and distinguish the cases based on the status code on the call side?

services/brig/src/Brig/API/User.hs
```diff
@@ -1230,10 +1232,57 @@ verifyDeleteUser d = do
   for_ account $ lift . wrapHttpClient . deleteAccount
   lift . wrapClient $ Code.delete key Code.AccountDeletion
 
--- | Internal deletion without validation.  Called via @delete /i/user/:uid@, or indirectly
--- via deleting self.
--- Team owners can be deleted if the team is not orphaned, i.e. there is at least one
--- other owner left.
+-- | Check if `deleteAccount` succeeded and run it again if needed
```
fisx, line 1235:
```suggestion
-- | Check if `deleteAccount` succeeded and run it again if needed.
```

services/brig/src/Brig/API/User.hs
```diff
@@ -1230,10 +1232,57 @@ verifyDeleteUser d = do
...
+  case mbAcc of
+    Nothing -> pure NoUser
+    Just acc -> do
+      probs <- Data.lookupPropertyKeysAndValues uid
+
+      let accIsDeleted = accountStatus acc == Deleted
+      clients <- Data.lookupClients uid
+
+      localUid <- qualifyLocal uid
+      conCount <- countConnections localUid [minBound .. maxBound]
```
fisx, line 1264:
```suggestion
      conCount <- countConnections localUid [(minBound @Relation) .. maxBound]
```

Both for human readers that don't have a mouse-over type checker handy, and for a future in which we suddenly change this signature to replace `Relation` with `Int`, and wonder why some end-points start crashing brig.  :-)

services/brig/src/Brig/API/User.hs
```diff
@@ -1230,10 +1232,57 @@ verifyDeleteUser d = do
...
+      cookies <- listCookies uid []
+
+      if (not . null) probs
```
fisx, line 1267:
```suggestion
      if notNull probs
```

https://hoogle.zinfra.io/file/root/.cabal/store/ghc-8.10.7/extra-1.7.9-8f69c8c825ffedcd23560ad4ef0789aad992568ab4532f16f7edce65b84795fc/share/doc/html/Data-List-Extra.html#v:notNull

(now that i'm writing this, i wonder if `notNull` should even exist...)

services/brig/src/Brig/Data/UserKey.hs
```diff
@@ -164,6 +165,13 @@ deleteKey k = do
   retry x5 $ write deleteHashed (params LocalQuorum (Identity hk))
   retry x5 $ write keyDelete (params LocalQuorum (Identity $ keyText k))
 
+deleteKeyForUser :: (MonadClient m, MonadReader Env m) => UserId -> UserKey -> m ()
+deleteKeyForUser uid k = do
```
fisx, line 169: what problem does this solve?

does it create a new problem in that now, if the key doesn't map onto the uid in half-deleted users any more, we can't clean up the rest?

services/spar/src/Spar/Scim/User.hs
```diff
@@ -698,11 +700,20 @@ deleteScimUser tokeninfo@ScimTokenInfo {stiTeam, stiIdP} uid =
     )
     (const id)
     $ do
-      mbBrigUser <- lift (Brig.getBrigUser Brig.WithPendingInvitations uid)
-      case mbBrigUser of
+      -- `getBrigUser` does not include deleted users. However, these
```
fisx, line 703:
```suggestion
      -- `getBrigUser` does not include deleted users. This is fine: these
```

(Is that what you mean?)

services/spar/src/Spar/Scim/User.hs
```diff
@@ -712,21 +723,45 @@ deleteScimUser tokeninfo@ScimTokenInfo {stiTeam, stiIdP} uid =
...
-                  (SAMLUserStore.delete uid)
-                  (ScimExternalIdStore.delete stiTeam)
-                  veid
-
-          lift $ ScimUserTimesStore.delete uid
-          lift $ BrigAccess.delete uid
+          -- This deletion needs data from the non-deleted User in brig. So,
+          -- execute it first, then delete the user in brig. Unfortunately, this
+          -- dependency prevents us from cleaning up users with deleted accounts
+          -- in brig here in spar.
```
fisx, line 729:
```suggestion
          -- dependency prevents us from cleaning up the spar fragments of users
          -- that have been deleted in brig.  Deleting scim-managed users in brig
          -- (via the TM app) is blocked, though, so there is no legal way to enter
          -- that situation.
```

Is this what you mean?

services/spar/src/Spar/Sem/BrigAccess/Http.hs
```diff
@@ -54,7 +55,7 @@ brigAccessToHttp mgr req =
       SetLocale itlu l -> Intra.setBrigUserLocale itlu l
       GetRichInfo itlu -> Intra.getBrigUserRichInfo itlu
       CheckHandleAvailable h -> Intra.checkHandleAvailable h
-      Delete itlu -> Intra.deleteBrigUser itlu
+      DeleteUser itlu -> Itra.deleteBrigUserInternal itlu
```
fisx, line 58:
```suggestion
      DeleteUser itlu -> Intra.deleteBrigUserInternal itlu
```

services/spar/src/Spar/Sem/BrigAccess/Http.hs
```diff
@@ -26,6 +26,7 @@ import Polysemy
 import Polysemy.Error (Error)
 import Spar.Error (SparError)
 import qualified Spar.Intra.Brig as Intra
+import qualified Spar.Intra.Brig as Itra
```
fisx, line 29:
```suggestion
```

hehe, stupid ide...  :-)

fisx (commented):
I've read everything except the tests now.

services/brig/src/Brig/API/User.hs
supersven: Oh, yes, that's reasonable. :+1:

services/spar/src/Spar/Sem/BrigAccess/Http.hs
supersven: Indeed: I made the `Itra` typo and the IDE fixed it... :rofl:

services/brig/src/Brig/API/User.hs
```diff
@@ -1264,7 +1264,7 @@ ensureAccountDeleted uid = do
       conCount <- countConnections localUid [(minBound @Relation) .. maxBound]
       cookies <- listCookies uid []
 
-      if (not . null) probs
+      if notNull probs
```
fisx, line 1267: there are two more below, now you have to make it consistent.  this is me, promising to be less nit-picky now...  m-)

services/brig/src/Brig/Data/UserKey.hs
supersven: The questions that led to this is: What if a user is partially deleted, their email (or phone number) was deleted correctly and is now used by another user (new account with other `UserId`)?

I think it wouldn't stop another deletion attempt of a partially deleted user, because it doesn't fail (just returns `()`).

fisx: that makes perfect sense, thanks!  could you add that verbatim as a haddock for future me?

services/brig/src/Brig/API/Internal.hs
supersven: The answer to most of your questions is: *Yes* :smile_cat: 

Will update the code accordingly.

libs/wire-api/src/Wire/API/User.hs
supersven: Well, I think your solution is more abstract and elegant, while mine is a bit easier to grasp (Go and ask some Java/Scala/Kotlin developers... :wink: ).

As the aeson instances aren't needed anymore (endpoint results are now only HTTP status codes), this debate will likely never be answered... :smile_cat:

services/brig/src/Brig/API/User.hs
supersven: Well, consistency is a good thing. (I've just overseen the other `not.null`s due to the small excerpt in the comment window.) :smile_cat:

services/brig/test/integration/API/User/Account.hs
```diff
@@ -1632,18 +1633,91 @@ testTooManyMembersForLegalhold opts brig = do
...
+    objectProp =
+      object
+        [ "key.1" .= ("val1" :: Text),
+          "key.2" .= ("val2" :: Text)
+        ]
+
 -- helpers
 
 setHandleAndDeleteUser :: Brig -> Cannon -> User -> [UserId] -> AWS.Env -> (UserId -> HttpT IO ()) -> Http ()
 setHandleAndDeleteUser brig cannon u others aws execDelete = do
```
fisx, line 1705: What an odd set of things to group into one function...?  But ok!

services/brig/test/integration/API/User/Account.hs
```diff
@@ -1667,7 +1741,7 @@ setHandleAndDeleteUser brig cannon u others aws execDelete = do
   forM_ others $ \usr -> do
     get (brig . paths ["users", toByteString' uid] . zUser usr) !!! assertDeletedProfilePublic
     Search.assertCan'tFind brig usr quid (fromName (userDisplayName u))
-    Search.assertCan'tFind brig usr quid hdl
+    Search.assertCan'tFind brig usr quid (fromHandle hdl)
```
fisx, line 1744: :heart:

services/brig/test/integration/API/User/Account.hs
```diff
@@ -1632,18 +1633,91 @@ testTooManyMembersForLegalhold opts brig = do
...
+testDeleteUserWithCompletelyDeletedUser brig cannon aws = do
+  u <- randomUser brig
+  liftIO $ Util.assertUserJournalQueue "user activate testDeleteUserWithCompletelyDeletedUser" aws (userActivateJournaled u)
+  setHandleAndDeleteUser brig cannon u [] aws $
+    \uid -> deleteUserInternal uid brig !!! const 202 === statusCode
+  do
+    let uid = userId u
+    deleteUserInternal uid brig
+      !!! do
+        const 200 === statusCode
```
fisx, line 1646:
```suggestion
  deleteUserInternal (userId u) brig !!! const 200 === statusCode
```

(and similarly below)

(question of taste, feel free to ignore)

services/brig/test/integration/API/User/Account.hs
```diff
@@ -1632,18 +1633,91 @@ testTooManyMembersForLegalhold opts brig = do
...
+testDeleteUserWithNotDeletedUser brig cannon aws = do
+  u <- randomUser brig
+  liftIO $ Util.assertUserJournalQueue "user activate testDeleteUserWithNotDeletedUser" aws (userActivateJournaled u)
+  do
+    setHandleAndDeleteUser brig cannon u [] aws $
+      ( \uid' ->
+          deleteUserInternal uid' brig
+            !!! do
+              const 202 === statusCode
+      )
```
fisx, line 1665:
```suggestion
  setHandleAndDeleteUser brig cannon u [] aws $
      \uid' -> deleteUserInternal uid' brig !!! const 202 === statusCode
```

services/brig/test/integration/API/User/Account.hs
```diff
@@ -1632,18 +1633,91 @@ testTooManyMembersForLegalhold opts brig = do
...
+  -- First set a unique handle (to verify freeing of the handle)
+  hdl <- randomHandle
+  let update = RequestBodyLBS . encode $ HandleUpdate hdl
+  put (brig . path "/self/handle" . contentJson . zUser uid . zConn "c" . body update)
+    !!! const 200 === statusCode
+
+  deleteUserInternal uid brig !!! const 202 === statusCode
+  liftIO $ Util.assertUserJournalQueue "user deletion testDeleteUserWithDanglingProperty" aws (userDeleteJournaled uid)
+
+  setProperty brig (userId u) "foo" objectProp
```
fisx, line 1682: interesting, these work with invalid `UserId`s?  ok, good for writing this test i guess...

services/spar/test/Test/Spar/Scim/UserSpec.hs
```diff
@@ -0,0 +1,189 @@
...
+        accountUser =
+          user
+            { userDisplayName = Name "Some User",
+              userAccentId = defaultAccentId,
+              userPict = noPict,
+              userAssets = [],
+              userHandle = parseHandle "some-handle",
+              userIdentity = (Just . EmailIdentity . fromJust . parseEmail) "someone@wire.com",
+              userTeam = Just $ stiTeam tokenInfo
+            }
```
fisx, line 188: why not just `user`?  why do you update all these fields?  i can't find anything in this module that expects these to be anything in particular.

services/spar/test/Test/Spar/Scim/UserSpec.hs
```diff
@@ -0,0 +1,189 @@
...
+import Test.Hspec
+import Test.QuickCheck
+import Web.Scim.Schema.Error
+import Wire.API.User
+import qualified Wire.API.User.Identity
+import Wire.API.User.Scim
+import Wire.Sem.Logger.TinyLog (discardTinyLogs)
+
+spec :: Spec
+spec = describe "deleteScimUser" $ do
```
fisx, line 33: This is a nice case study on how to write tests based on polysemy, and it's perfectly mergeable and useful.

In the long run, I wonder how we can make this more convenient.  If we want to replace arbitrary interpreters out of a composition of 5-15 of them, things will become awkward.  Organising interpreters in trees (ie., building composite effects from more basic ones) would probably help?  Or what about a record that contains all the default interpreters and that is passed declared somewhere in the test suite and a function that builds a composite interpreter from all the record fields; and then we can replace individual fields in the record using lenses before passing it to that function?

@isovector iff you are curious, feel free to form an opinion while you're waiting for the new contract.  (no pressure :))

fisx (approved):

services/brig/test/integration/API/User/Account.hs
supersven: Hm, I think most of it was an unconscious decision (i.e. `ormulu`).

services/brig/test/integration/API/User/Account.hs
supersven: Auto-formatter... :smile:

services/brig/test/integration/API/User/Account.hs
supersven: :smile:

services/spar/test/Test/Spar/Scim/UserSpec.hs
supersven: I thought it would be good to provide a valid / sound / half-way realistic test object: It's easier to extend or add tests when the test data isn't too artificial.

services/brig/test/integration/API/User/Account.hs
supersven: Well, it's sole purpose is to reduce code duplication... :shrug:
Nice side-effect is, that some assumptions are checked for (almost) all deletion tests, that likely wouldn't be if one had to duplicate the assertions.

(Though, the `And` in the function name is almost always an indicator that concepts are mixed that shouldn't be... :thinking: )

services/spar/test/Test/Spar/Scim/UserSpec.hs
isovector: Good question. Interpreters are a DAG that we're unfortunate enough to need to encode as a list. The hard part here is tracking effect dependencies, but some sort of staged "compilation" could be work. I'll put some thought into it.

PR #2639, opened 2022-08-22
We only need one rule for `make clean`.

fisx:
- simplify and consolidate make rule names
- drop hint in case of compiler error

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -52,7 +52,7 @@ endif
 
 # Clean
 .PHONY: clean
-clean: delete-cache-on-linker-errors
+clean:
 	rm -rf ~/.cache/hie-bios
 ifdef CABAL_DIR
 	rm -rf $(CABAL_DIR)/store
--- a/Makefile
+++ b/Makefile
@@ -74,9 +74,6 @@ clean-hint:
 	@echo -e ">>> try 'make clean' and run your command again."
 	@echo -e ">>> see https://github.com/wireapp/wire-server/blob/develop/docs/developer/building.md#linker-errors-while-compiling\n\n\n"
 
-.PHONY: delete-cache-on-linker-errors
-delete-cache-on-linker-errors: clean
-
 .PHONY: cabal.project.local
 cabal.project.local:
 	echo "optimization: False" > ./cabal.project.local
--- a/Makefile
+++ b/Makefile
@@ -51,14 +51,17 @@ else
 endif
 
 # Clean
-.PHONY: clean
-clean:
+.PHONY: full-clean
+full-clean: clean
 	rm -rf ~/.cache/hie-bios
 ifdef CABAL_DIR
 	rm -rf $(CABAL_DIR)/store
 else
 	rm -rf ~/.cabal/store
 endif
+
+.PHONY: clean
+clean:
 ifeq ($(WIRE_BUILD_WITH_CABAL), 1)
 	cabal clean
 else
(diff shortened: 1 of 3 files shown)


fisx:
feel free to merge it this is green from ci and reviewer!

elland (approved):

Makefile
```diff
@@ -50,15 +50,32 @@ else
 	stack install --pedantic --test --bench --no-run-benchmarks --local-bin-path=dist
 endif
 
-.PHONY: delete-cache-on-linker-errors
-delete-cache-on-linker-errors:
+# Clean
+.PHONY: clean
+clean: delete-cache-on-linker-errors
```
supersven, line 55: So, `clean` depends on `delete-cache-on-linker-errors` and the other way round. Does this work? Looks like an endless loop... :thinking:

Makefile
```diff
@@ -50,15 +50,32 @@ else
...
+	-rm -f .metadata
+
+.PHONY: clean-hint
+clean-hint:
+	@echo -e "\n\n\n>>> PSA: if you get errors that are hard to explain,"
+	@echo -e ">>> try 'make clean' and run your command again."
+	@echo -e ">>> see https://github.com/wireapp/wire-server/blob/develop/docs/developer/building.md#linker-errors-while-compiling\n\n\n"
+
+.PHONY: delete-cache-on-linker-errors
+delete-cache-on-linker-errors: clean
```
supersven, line 78: Wouldn't it be better to remove this target and replace it's occurrences in any docs with `clean`?

(`delete-cache-on-linker-errors` is pretty new. So, there shouldn't be much that depends on it. And, IMHO the name is a bit misleading.)

Makefile
```diff
@@ -77,7 +94,7 @@ endif
 # Usage: make c package=brig test=1
 .PHONY: c
 c: cabal-fmt
-	cabal build $(WIRE_CABAL_BUILD_OPTIONS) $(package)
+	cabal build $(WIRE_CABAL_BUILD_OPTIONS) $(package) || ( make clean-hint; false )
```
supersven, line 97: Why do you show the `clean-hint` in this specific target? :thinking:

supersven (changes requested):

Makefile
fisx: because i don't understand the structure of this makefile :)?  which targets would you like me to add it to?

Makefile
fisx: ah, no, oversight, i was moving things around too violently in between and lost perspective.  will fix it.

Makefile
fisx: ok, i'll do that.

fisx: 9df39b14ae41226cbceeb36a00fa2247ec47423d

Makefile
fisx: 7b8d18634

Makefile
supersven: Ups, my mistake, sorry: `c` should be fine. (`c` is *build everything with cabal*)

supersven (approved):
Now, it looks good to me :+1: 

Thanks for taking the effort to clean this up.

akshaymankar (changes requested):
This is not a good idea, this will delete all the cache in CI and CI will take more than an hour to compile things.

pcapriotti:
Also, `make clean` in a single repository shouldn't affect the global or per-user state of cabal.

elland:
@fisx `make nuke`? `make purge`?

akshaymankar (approved):
make :broom:

PR #2640, opened 2022-08-22
Fix cql-io bug where restarting C* cluster could cause downtime

akshaymankar:
Upstream changes in https://gitlab.com/twittner/cql-io/-/merge_requests/20

JIRA: https://wearezeta.atlassian.net/browse/SQCORE-1345

## Checklist


akshaymankar:
Please approve this PR only if you approve of the upstream MR.

cabal.project.freeze
```diff
@@ -1153,6 +1152,8 @@ constraints: any.AC-Angle ==1.0,
              any.hourglass ==0.2.12,
              any.hourglass-orphans ==0.1.0.0,
              any.hp2pretty ==0.10,
+             any.hpack ==0.34.5,
+             any.hpack-dhall ==0.5.3,
```
fisx, line 1156: why is this added here?  don't mind either way, just surprised.

fisx (approved):

cabal.project.freeze
akshaymankar: I just ran the script in `tools/convert-to-cabal` to keep `stack.yaml` and  `cabal.project` in sync. I was confused too, but then thought maybe someone added some dependency and forgot to run that script.

jschaul (commented):
Is it possible to create a test for this behaviour before/after; or is that not so easy? If not, would it be possible to detail the manual test procedure steps either here in wire-server in some docs; or in the MR upstream on how you tested the change? (internal jira comments can't be read by cql-io maintainers or anyone else interested)

jschaul (approved):

akshaymankar:
> Is it possible to create a test for this behaviour before/after; or is that not so easy? If not, would it be possible to detail the manual test procedure steps either here in wire-server in some docs; or in the MR upstream on how you tested the change? (internal jira comments can't be read by cql-io maintainers or anyone else interested)

I think it is quite hard to create an automated test for this since it requires talking to kubernetes. I wrote about my method of testing in the upstream MR: https://gitlab.com/twittner/cql-io/-/merge_requests/20#note_1074472850

PR #2641, opened 2022-08-22
Sqservices 1721 acme be new nonce endpoint

battermann:
https://wearezeta.atlassian.net/browse/SQSERVICES-1721

## Checklist

 - [x] If HTTP endpoint paths have been added or renamed, or feature configs have changed, the **endpoint / config-flag checklist** (see Wire-employee only backend [wiki page](https://github.com/zinfra/backend-wiki/wiki/Checklists)) has been followed.
 - [x] If configuration options have been added or removed, check [docs/developer/adding-config-options](https://github.com/wireapp/wire-server/blob/develop/docs/legacy/developer/adding-config-options.md) and follow the steps there.

### DIFF
--- a/changelog.d/0-release-notes/pr-2641
+++ b/changelog.d/0-release-notes/pr-2641
@@ -1,0 +1,1 @@
+Deploy nginz before brig.
--- a/changelog.d/0-release-notes/pr-2641
+++ b/changelog.d/0-release-notes/pr-2641
@@ -1,1 +1,0 @@
-Deploy nginz before brig.
--- a/changelog.d/2-features/pr-2641
+++ b/changelog.d/2-features/pr-2641
@@ -1,1 +1,1 @@
-New endpoints `HEAD` and `GET /nonce/clients` to request new nonces for clients.
+New endpoints `HEAD` and `GET /nonce/clients` to request new nonces for client certificate requests (coming up soon).
--- a/changelog.d/2-features/pr-2641
+++ b/changelog.d/2-features/pr-2641
@@ -1,0 +1,1 @@
+New endpoints `HEAD` and `GET /nonce/clients` to request new nonces for clients.
--- a/hack/helm_vars/wire-server/values.yaml.gotmpl
+++ b/hack/helm_vars/wire-server/values.yaml.gotmpl
@@ -82,6 +82,7 @@ brig:
         - domain: federation-test-helper.{{ .Release.Namespace }}.svc.cluster.local
           search_policy: full_search
       set2FACodeGenerationDelaySecs: 5
+      setNonceTtlSecs: 5
     aws:
       sesEndpoint: http://fake-aws-ses:4569
       sqsEndpoint: http://fake-aws-sqs:4568
--- a/hack/helm_vars/wire-server/values.yaml.gotmpl
+++ b/hack/helm_vars/wire-server/values.yaml.gotmpl
@@ -82,7 +82,7 @@ brig:
         - domain: federation-test-helper.{{ .Release.Namespace }}.svc.cluster.local
           search_policy: full_search
       set2FACodeGenerationDelaySecs: 5
-      setNonceTtlSecs: 5
+      setNonceTtlSecs: 300
     aws:
       sesEndpoint: http://fake-aws-ses:4569
       sqsEndpoint: http://fake-aws-sqs:4568
--- a/libs/types-common/src/Data/Nonce.hs
+++ b/libs/types-common/src/Data/Nonce.hs
@@ -22,16 +22,13 @@ module Data.Nonce where
 
 import Cassandra hiding (Value)
 import qualified Data.Aeson as A
-import Data.Bifunctor (Bifunctor (first))
 import qualified Data.ByteString.Base64.URL as Base64
-import Data.ByteString.Builder (toLazyByteString)
 import Data.ByteString.Conversion
 import Data.ByteString.Lazy (fromStrict, toStrict)
 import Data.Proxy (Proxy (Proxy))
 import Data.String.Conversions (cs)
 import qualified Data.Swagger as S
 import Data.Swagger.ParamSchema
-import Data.Text (pack)
 import Data.Text.Encoding (encodeUtf8)
 import Data.UUID as UUID (UUID, fromByteString, toByteString)
 import Data.UUID.V4 (nextRandom)
@@ -59,11 +56,11 @@ instance ToParamSchema Nonce where
   toParamSchema _ = toParamSchema (Proxy @Text)
 
 instance ToHttpApiData Nonce where
-  toQueryParam = cs . toLazyByteString . builder
+  toQueryParam = cs . toByteString'
 
 instance FromHttpApiData Nonce where
   parseQueryParam s =
-    first pack $ runParser parser (encodeUtf8 s)
+    maybe (Left "Invalid Nonce") Right $ fromByteString' (cs . encodeUtf8 $ s)
 
 randomNonce :: (Functor m, MonadIO m) => m Nonce
 randomNonce = Nonce <$> liftIO nextRandom
--- a/libs/types-common/src/Data/Nonce.hs
+++ b/libs/types-common/src/Data/Nonce.hs
@@ -29,7 +29,6 @@ import Data.Proxy (Proxy (Proxy))
 import Data.String.Conversions (cs)
 import qualified Data.Swagger as S
 import Data.Swagger.ParamSchema
-import Data.Text.Encoding (encodeUtf8)
 import Data.UUID as UUID (UUID, fromByteString, toByteString)
 import Data.UUID.V4 (nextRandom)
 import Imports
@@ -59,14 +58,13 @@ instance ToHttpApiData Nonce where
   toQueryParam = cs . toByteString'
 
 instance FromHttpApiData Nonce where
-  parseQueryParam s =
-    maybe (Left "Invalid Nonce") Right $ fromByteString' (cs . encodeUtf8 $ s)
+  parseQueryParam = maybe (Left "Invalid Nonce") Right . fromByteString' . cs
 
 randomNonce :: (Functor m, MonadIO m) => m Nonce
 randomNonce = Nonce <$> liftIO nextRandom
 
 isValidBase64UrlEncodedUUID :: ByteString -> Bool
-isValidBase64UrlEncodedUUID = isRight . runParser (parser @Nonce)
+isValidBase64UrlEncodedUUID = isJust . fromByteString' @Nonce . cs
 
 instance Cql Nonce where
   ctype = Tagged UuidColumn
--- a/libs/types-common/src/Data/Nonce.hs
+++ b/libs/types-common/src/Data/Nonce.hs
@@ -1,0 +1,43 @@
+{-# LANGUAGE DataKinds #-}
+{-# LANGUAGE GeneralizedNewtypeDeriving #-}
+
+-- This file is part of the Wire Server implementation.
+--
+-- Copyright (C) 2022 Wire Swiss GmbH <opensource@wire.com>
+--
+-- This program is free software: you can redistribute it and/or modify it under
+-- the terms of the GNU Affero General Public License as published by the Free
+-- Software Foundation, either version 3 of the License, or (at your option) any
+-- later version.
+--
+-- This program is distributed in the hope that it will be useful, but WITHOUT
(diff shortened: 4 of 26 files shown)


changelog.d/2-features/pr-2641
```diff
@@ -0,0 +1 @@
+New endpoints `HEAD` and `GET /nonce/clients` to request new nonces for clients.
```
fisx, line 1:
```suggestion
New endpoints `HEAD` and `GET /nonce/clients` to request new nonces for client certificate requests (coming up soon).
```

fisx (commented):
You removed an item from the check list in the PR description that is relevant here.  Maybe you want to check if that's green, and if you missed anything else?

Meanwhile I'll keep reviewing the rest of the code.

hack/helm_vars/wire-server/values.yaml.gotmpl
```diff
@@ -82,6 +82,7 @@ brig:
         - domain: federation-test-helper.{{ .Release.Namespace }}.svc.cluster.local
           search_policy: full_search
       set2FACodeGenerationDelaySecs: 5
+      setNonceTtlSecs: 5
```
fisx, line 85: So if the customers don't change this, on-prem setup will be 5 seconds, not 5 minutes?  I'd prefer 5 minutes.

libs/types-common/src/Data/Nonce.hs
```diff
@@ -0,0 +1,74 @@
...
+instance ToByteString Nonce where
+  builder = builder . Base64.encode . toStrict . UUID.toByteString . unNonce
+
+instance FromByteString Nonce where
+  parser = do
+    a <- parser
+    maybe
+      (fail "invalid base64url encoded uuidv4")
+      (pure . Nonce)
+      (either (const Nothing) (UUID.fromByteString . fromStrict) (Base64.decode a))
```
fisx, line 53:
```suggestion
  parser = do
    a :: ByteString <- parser
    maybe
      (fail "invalid base64url encoded uuidv4")
      (pure . Nonce)
      ((UUID.fromByteString . fromStrict <=< either (const Nothing) Just . Base64.decode) a)
```

hum.  your variant is fine.  i was hoping i could do better, but it looks like you already nailed it.  :)  ignore me!

libs/types-common/src/Data/Nonce.hs
```diff
@@ -0,0 +1,74 @@
...
+
+instance ToParamSchema Nonce where
+  toParamSchema _ = toParamSchema (Proxy @Text)
+
+instance ToHttpApiData Nonce where
+  toQueryParam nonce = cs (toByteString' nonce)
+
+instance FromHttpApiData Nonce where
+  parseQueryParam s =
+    first pack $ runParser parser (encodeUtf8 s)
```
fisx, line 63:
```suggestion
instance ToHttpApiData Nonce where
  toQueryParam = cs . toByteString'

instance FromHttpApiData Nonce where
  parseQueryParam = fmap cs . fromByteString'
```

I haven't compiled this, but I think your's could be more symmetrical.

services/brig/test/integration/API/User/Util.hs
```diff
@@ -551,3 +551,25 @@ setTeamFeatureLockStatus galley tid status =
...
+getNonce = nonce get
+
+headNonce ::
+  (MonadIO m, MonadHttp m) =>
+  Brig ->
+  UserId ->
+  m ResponseLBS
+headNonce = nonce Bilge.head
+
+nonce :: ((Request -> c) -> t) -> (Request -> c) -> UserId -> t
```
fisx, line 569: This is fine, just FYI I think there is also a combinator `method` that lets you change the method in the function sequence that also contains `paths`, `zUser`, ...

libs/types-common/src/Data/Nonce.hs
```diff
@@ -0,0 +1,74 @@
...
+import qualified Data.Swagger as S
+import Data.Swagger.ParamSchema
+import Data.Text (pack)
+import Data.Text.Encoding (encodeUtf8)
+import Data.UUID as UUID (UUID, fromByteString, toByteString)
+import Data.UUID.V4 (nextRandom)
+import Imports
+import Servant (FromHttpApiData (..), ToHttpApiData (..))
+
+newtype Nonce = Nonce {unNonce :: UUID}
```
fisx, line 40: add a roundtrip encode-decode test somewhere?  and even a golden test, maybe?  it can all be dropped somewhere in the existing test suite, i'm ok with a preference of easy over well-organized.

libs/wire-api/src/Wire/API/Routes/Public/Brig.hs
```diff
@@ -533,6 +539,43 @@ type UserClientAPI =
...
+
+type NonceAPI =
+  -- be aware that the order matters, if get was first, then head requests would be routed to the get handler
+  NewNonce "head-nonce" 'HEAD 200
+    :<|> NewNonce "get-nonce" 'GET 204
+
+type NewNonce name method statusCode =
+  Named
+    name
+    ( Summary "Get a new nonce for a client, specified in the response header `Replay-Nonce` as a uuidv4 in base64url encoding"
```
fisx, line 552:
```suggestion
    ( Summary "Get a new nonce for a client CSR, specified in the response header `Replay-Nonce` as a uuidv4 in base64url encoding"
```

libs/wire-api/src/Wire/API/Routes/Public/Brig.hs
```diff
@@ -533,6 +539,43 @@ type UserClientAPI =
...
+
+instance ToParamSchema NonceHeader where
+  toParamSchema _ = toParamSchema (Proxy @Text)
+
+instance ToHttpApiData NonceHeader where
+  toQueryParam nonce = cs (toByteString' nonce)
+
+instance FromHttpApiData NonceHeader where
+  parseQueryParam s =
+    first pack $ runParser parser (encodeUtf8 s)
```
fisx, line 574:
```suggestion
  deriving newtype (FromByteString, ToByteString, ToParamSchema, ToHttpApiData, FromHttpApiData)
```

...  because Nonce already has these.  If I'm missing anything important here, please add a source comment.

services/brig/src/Brig/API/Public.hs
```diff
@@ -248,6 +250,8 @@ servantSitemap = userAPI :<|> selfAPI :<|> accountAPI :<|> clientAPI :<|> prekey
         :<|> Named @"get-client" getClient
         :<|> Named @"get-client-capabilities" getClientCapabilities
         :<|> Named @"get-client-prekeys" getClientPrekeys
+        :<|> Named @"head-nonce" newNonce
+        :<|> Named @"get-nonce" newNonce
```
fisx, line 254: Huh, they are exactly the same?  In that case why not drop the `GET`?

services/brig/src/Brig/Data/Nonce.hs
```diff
@@ -0,0 +1,52 @@
...
+-- ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+-- FOR A PARTICULAR PURPOSE. See the GNU Affero General Public License for more
+-- details.
+--
+-- You should have received a copy of the GNU Affero General Public License along
+-- with this program. If not, see <https://www.gnu.org/licenses/>.
+
+module Brig.Data.Nonce
+  ( addNonce,
+    invalidateNonce,
```
fisx, line 20:
```suggestion
  ( insertNonce,
    deleteNonce,
```

?

services/brig/src/Brig/Data/Nonce.hs
```diff
@@ -0,0 +1,52 @@
...
+  Nonce ->
+  m ()
+addNonce ttl nonce = retry x5 . write insertNonce $ params LocalQuorum (Identity nonce)
+  where
+    insertNonce :: PrepQuery W (Identity Nonce) ()
+    insertNonce = fromString $ "INSERT INTO nonce (nonce) VALUES (?)" <> renderTtl
+    renderTtl :: String
+    renderTtl
+      | ttl > 0 = " USING TTL " <> show ttl
+      | otherwise = " USING TTL null"
```
fisx, line 43: I'm not sure `null` works, but I'm sure `0` works.  So the only thing you want to handle here are negative numbers, and you can avoid those by using `Word64`, which fits 6 years in seconds.  Then you can use a simple `show` or `toByteString'` instead of this render function.

services/brig/src/Brig/Options.hs
```diff
@@ -589,7 +589,10 @@ data Settings = Settings
     setEnableDevelopmentVersions :: Maybe Bool,
     -- | Minimum delay in seconds between consecutive attempts to generate a new verification code.
     -- use `set2FACodeGenerationDelaySecs` as the getter function which always provides a default value
-    set2FACodeGenerationDelaySecsInternal :: !(Maybe Int)
+    set2FACodeGenerationDelaySecsInternal :: !(Maybe Int),
+    -- | The time-to-live of a nonce in seconds.
+    -- use `set2FANonceTtlSecs` as the getter function which always provides a default value
```
fisx, line 594: comment needs more editing.

services/brig/test/integration/API/User/Client.hs
```diff
@@ -946,6 +948,18 @@ testPreKeyRace brig = do
   liftIO $ assertEqual "duplicate prekeys" (length regular) (length (nub regular))
   deleteClient brig uid (clientId c) (Just defPasswordText) !!! const 200 === statusCode
 
+testNewNonce :: Brig -> Http ()
+testNewNonce brig = do
```
fisx, line 952: [optional] if you need more nit-picks, you can compare the two uuids and fail if they are the same.  :)

changelog.d/0-release-notes/pr-2641
```diff
@@ -0,0 +1 @@
+Deploy nginz before brig.
```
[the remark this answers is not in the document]
battermann: Interesting, I just realized that this is not a problem at after all, because the new endpoint becomes available once both services are updated. So order does not matter.

libs/types-common/src/Data/Nonce.hs
battermann: Good catch, I think it is even better if we use the `ToByteString` and `FromByteString` instances like this:

```haskell
instance ToByteString Nonce where
  builder = builder . Base64.encode . toStrict . UUID.toByteString . unNonce

instance FromByteString Nonce where
  parser = do
    a <- parser
    maybe
      (fail "invalid base64url encoded uuidv4")
      (pure . Nonce)
      (either (const Nothing) (UUID.fromByteString . fromStrict) (Base64.decode a))

instance ToHttpApiData Nonce where
  toQueryParam = cs . toLazyByteString . builder

instance FromHttpApiData Nonce where
  parseQueryParam s =
    first pack $ runParser parser (encodeUtf8 s)
```

battermann: WDYT @fisx?

services/brig/src/Brig/API/Public.hs
battermann: ACME RFC suggests HEAD plus optional GET, it was easy to add, so I did.

libs/wire-api/src/Wire/API/Routes/Public/Brig.hs
battermann: We need instances of these to decode and encode base64url.

services/brig/src/Brig/Options.hs
battermann: Can explain what how this needs to be edited?

libs/types-common/src/Data/Nonce.hs
fisx: erm...  you've copied the exact code that was in the branch when i reviewed it.  was that intentional?

services/brig/src/Brig/API/Public.hs
```diff
@@ -621,7 +621,7 @@ newNonce :: UserId -> (Handler r) Nonce
 newNonce _ = do
   ttl <- setNonceTtlSecs <$> view settings
   nonce <- randomNonce
-  lift $ wrapClient $ Nonce.addNonce ttl nonce
+  lift $ wrapClient $ Nonce.insertNonce ttl nonce
```
fisx, line 624: interesting, `delete` is not called anywhere.  have you tested it manually?  (not that important, it'll happen soon, might as well be in a separate PR.)

fisx (approved):
LGTM!  Except for the CI problem (haven't looked what it's about) and the remaining unresolved comments (which we discussed offline), this is ready.

PR #2622, opened 2022-08-16
Publish developer docs

fisx:
https://wearezeta.atlassian.net/browse/SQSERVICES-1724

## Checklist

 - [x] Make sure you've incorporated all changes introduced between a0efd6a0e61c56e7283ce389c6d720dbd0888cdf and develop.
 - [x] The **PR Title** explains the impact of the change.
 - [x] The **PR description** provides context as to why the change should occur and what the code contributes to that effect. This could also be a link to a JIRA ticket or a Github issue, if there is one.
 - [x] **changelog.d** contains the following bits of information ([details](https://github.com/wireapp/wire-server/blob/develop/docs/developer/changelog.md)):
   - [x] A file with the changelog entry in one or more suitable sub-sections. The sub-sections are marked by directories inside `changelog.d`.

PR #2646, opened 2022-08-23
Create PR guidelines and simplify checklist

pcapriotti:
This PR simplifies the PR checklist and moves most of the content into a dedicated document containing guidelines to be followed by PR authors.

## Checklist

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,12 +1,9 @@
 ## Checklist
 
- - [ ] The **PR Title** explains the impact of the change.
- - [ ] The **PR description** provides context as to why the change should occur and what the code contributes to that effect. This could also be a link to a JIRA ticket or a Github issue, if there is one.
- - [ ] If this PR changes development workflow or dependencies, they have been A) automated and B) documented under docs/developer/. All efforts have been taken to minimize development setup breakage or slowdown for co-workers.
- - [ ] If HTTP endpoint paths have been added or renamed, check [docs/developer/adding-api-endpoints](https://github.com/wireapp/wire-server/blob/develop/docs/legacy/developer/adding-api-endpoints.md) and follow the steps there.
  - [ ] If configuration options have been added or removed, check [docs/developer/adding-config-options](https://github.com/wireapp/wire-server/blob/develop/docs/legacy/developer/adding-config-options.md) and follow the steps there.
  - [ ] If a cassandra schema migration has been added, I ran **`make git-add-cassandra-schema`** to update the cassandra schema documentation.
-- [ ] I swear that if I have changed internal end-points, I do not implicitly rely on deployment ordering (brig needing to be deployed before galley), i.e. I have **not** introduced a new internal endpoint in brig and already make use of if in galley, as I'm aware old deployed galleys would throw 500s until all new version have been rolled out and I do not want a few minutes of downtime. Instead, I have thought how to merge brig an galley codebases.
+
+ - [ ] Create a new entry in **changelog.d**.
  - [ ] I updated **changelog.d** subsections with one or more entries with the following bits of information ([details](https://github.com/wireapp/wire-server/blob/develop/docs/developer/changelog.md)):
    - [ ] If a cassandra schema migration is backwards incompatible (see also [these docs](https://github.com/wireapp/wire-server/blob/develop/docs/developer/cassandra-interaction.md#cassandra-schema-migrations)), measures to be taken by on-premise instance operators are explained.
    - [ ] If a data migration (not schema migration) introduced: measures to be taken by on-premise instance operators.
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,9 +1,5 @@
 ## Checklist
 
- - [ ] If configuration options have been added or removed, check [docs/developer/adding-config-options](https://github.com/wireapp/wire-server/blob/develop/docs/legacy/developer/adding-config-options.md) and follow the steps there.
- - [ ] If a cassandra schema migration has been added, I ran **`make git-add-cassandra-schema`** to update the cassandra schema documentation.
-
- - [ ] Create a new entry in **changelog.d**.
- - [ ] I updated **changelog.d** subsections with one or more entries with the following bits of information ([details](https://github.com/wireapp/wire-server/blob/develop/docs/developer/changelog.md)):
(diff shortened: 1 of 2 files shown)


.github/pull_request_template.md
```diff
@@ -1,15 +1,5 @@
...
-   - [ ] A file with the changelog entry in one or more suitable sub-sections. The sub-sections are marked by directories inside `changelog.d`.
-   - [ ] If new config options introduced: added usage description under docs/reference/config-options.md
-   - [ ] If new config options introduced: recommended measures to be taken by on-premise instance operators.
-   - [ ] If a cassandra schema migration is backwards incompatible (see also [these docs](https://github.com/wireapp/wire-server/blob/develop/docs/developer/cassandra-interaction.md#cassandra-schema-migrations)), measures to be taken by on-premise instance operators are explained.
-   - [ ] If a data migration (not schema migration) introduced: measures to be taken by on-premise instance operators.
-   - [ ] If public end-points have been changed or added: does nginz need un upgrade?
-   - [ ] If internal end-points have been added or changed: which services have to be deployed in a specific order?
+ - [ ] Add a new entry in a appropriate subdirectory of `changelog.d`
```
fisx: this is also covered by the next point, but i guess it's good for me that i'll have it here because i forget every single time, still.  :)

fisx (approved):
i would prefer to leave empty files pointing to the new place rather than fully deleting them, but i'll leave it to you.

otherwise i like these changes!

docs/developer/pr-guidelines.md
```diff
@@ -0,0 +1,102 @@
...
+
+See `docs/legacy/developer/changelog.md` for more information.
+
+## Schema migrations
+
+If a cassandra schema migration has been added then
+
+ - [ ] Run **`make git-add-cassandra-schema`** to update the cassandra schema documentation
+
+## Incompatible schema migrations and data migrations
```
flokli, line 23:
```suggestion
### Incompatible schema migrations and data migrations
```

This is a subsection of "Schema migrations", no?

.github/pull_request_template.md
pcapriotti: True, but we decided to add it anyway because it's needed by essentially all PRs. The idea is that the PR author would take relevant checklists from the guidelines document and add them to the PR description, and this one item will basically always be there anyway.

pcapriotti:
> i would prefer to leave empty files pointing to the new place rather than fully deleting them, but i'll leave it to you.

Those files where created very recently. I don't think anyone is going to be looking for them.

docs/developer/pr-guidelines.md
```diff
@@ -0,0 +1,102 @@
...
+
+The recommended way to use this document is to copy the relevant checklists below into the PR description, when appropriate, and make sure they are all checked before the PR is merged.
+
+## Changelog entries
+
+Every PR should add a new file in the appropriate subdirectory of `changelog.d`, containing just the text of the corresponding changelog entry. There is no need to explicitly write a PR number, because the `mk-changelog.sh` script (used on release) will add it automatically at the end. The name of the file does not matter, but it should be unique to avoid unnecessary conflicts (e.g. use the branch name).
+
+It is still possible to write the PR number manually if so desired, which is useful in case the entry is shared by multiple PRs. In that case, the script would leave the PR number reference intact, as long as it is at the very end of the entry, with no period afterwards and in brackets. It is also possible to use the pattern `##` to refer to the current PR number. This will be replaced throughout.
```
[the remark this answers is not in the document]
pcapriotti: I clarified how the PR number should be written, but I'm not quite sure what the other thing you're referring to is. If you can write a paragraph about that, I'll add it.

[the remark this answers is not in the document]
flokli: https://github.com/wireapp/wire-server/pull/2476 for example was merged by "Merge", not "Squash", and the PR number was added in the PR writing the changelog entry.

Maybe after

> It is still possible to write the PR number manually if so desired, which is useful in case the entry is shared by multiple PRs

add a

> , or if the PR is merged with a merge commit, rather than a squash.

I'd move the part about `##` either before or in a new paragraph, because it doesn't make sense in the "non-squash case".

fisx (commented):
#2622 has landed, please move `docs/developer/pr-guidelines.md` to `docs/src/developer/developer/pr-guidelines.md` and add an entry in `docs/src/developer/index.rst`.

pcapriotti:
> #2622 has landed, please move `docs/developer/pr-guidelines.md` to `docs/src/developer/developer/pr-guidelines.md` and add an entry in `docs/src/developer/index.rst`.

Not sure what I should add to `docs/src/developer/index.rst`.

smatting:
I just checked: it gets indexed automatically, there is a `:glob:` option with a `*` wildcard in the `.toctree`

.github/pull_request_template.md
```diff
@@ -1,12 +1,5 @@
...
- - [ ] If HTTP endpoint paths have been added or renamed, check [docs/developer/adding-api-endpoints](https://github.com/wireapp/wire-server/blob/develop/docs/legacy/developer/adding-api-endpoints.md) and follow the steps there.
- - [ ] If configuration options have been added or removed, check [docs/developer/adding-config-options](https://github.com/wireapp/wire-server/blob/develop/docs/legacy/developer/adding-config-options.md) and follow the steps there.
- - [ ] If a cassandra schema migration has been added, I ran **`make git-add-cassandra-schema`** to update the cassandra schema documentation.
-- [ ] I swear that if I have changed internal end-points, I do not implicitly rely on deployment ordering (brig needing to be deployed before galley), i.e. I have **not** introduced a new internal endpoint in brig and already make use of if in galley, as I'm aware old deployed galleys would throw 500s until all new version have been rolled out and I do not want a few minutes of downtime. Instead, I have thought how to merge brig an galley codebases.
```
smatting, line 9: This might be worth dicussing, because it has been removed. If we were to keep it in this would mean that we couldn't introduce or change any internal endpoints anymore (in any non-backwards compatible way).

Even such a strategy as making the change in release X and then only start using it in relase X+1 doesn't work, because customers might skip releases when upgrading.

I think there is nothing but to accept 500s during the deployment process.

/cc @jschaul

smatting (approved):

.github/pull_request_template.md
pcapriotti: The only real solution would be to start using a versioning system for internal APIs. It seems a bit impractical, but it could be done, potentially.

docs/developer/pr-guidelines.md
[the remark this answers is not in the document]
pcapriotti: Thanks, done.

PR #2636, opened 2022-08-21
Brig Polysemy: Port UserPendingActivationStore to polysemy

isovector:
This PR moves `Brig.Data.UserPendingActivation` into a polysemy effect `Brig.Sem.UserPendingActivationStore`.

## Checklist

 - [x] The **PR Title** explains the impact of the change.
 - [x] The **PR description** provides context as to why the change should occur and what the code contributes to that effect. This could also be a link to a JIRA ticket or a Github issue, if there is one.
 - [x] **changelog.d** contains the following bits of information ([details](https://github.com/wireapp/wire-server/blob/develop/docs/developer/changelog.md)):
   - [x] A file with the changelog entry in one or more suitable sub-sections. The sub-sections are marked by directories inside `changelog.d`.

PR #2648, opened 2022-08-23
Move paging effect into polysemy-wire-zoo

isovector:
In #2636, @pcapriotti suggested I use the existing paging machinery in Galley, rather than roll my own version. This PR moves the necessary machinery into `polysemy-wire-zoo` so it can be reused by other services.

## Checklist

 - [x] I updated **changelog.d** subsections with one or more entries with the following bits of information ([details](https://github.com/wireapp/wire-server/blob/develop/docs/developer/changelog.md)):


isovector:
@pcapriotti please take a look

pcapriotti (approved):

fisx (approved):

PR #2652, opened 2022-08-25
[FS-736] ACME Create DPoP Access Token

battermann:
https://wearezeta.atlassian.net/browse/FS-736

Since this is just the "skeleton" implementation for this feature, and the rust library this feature depends on is not finished either, documentation, exhaustive tests, and release notes will be in a separate PR.

## Checklist

 - [x] If HTTP endpoint paths have been added or renamed, check [docs/developer/adding-api-endpoints](https://github.com/wireapp/wire-server/blob/develop/docs/legacy/developer/adding-api-endpoints.md) and follow the steps there.
 - [ ] If configuration options have been added or removed, check [docs/developer/adding-config-options](https://github.com/wireapp/wire-server/blob/develop/docs/legacy/developer/adding-config-options.md) and follow the steps there.
- [x] I swear that if I have changed internal end-points, I do not implicitly rely on deployment ordering (brig needing to be deployed before galley), i.e. I have **not** introduced a new internal endpoint in brig and already make use of if in galley, as I'm aware old deployed galleys would throw 500s until all new version have been rolled out and I do not want a few minutes of downtime. Instead, I have thought how to merge brig an galley codebases.
 - [x] I updated **changelog.d** subsections with one or more entries with the following bits of information ([details](https://github.com/wireapp/wire-server/blob/develop/docs/developer/changelog.md))


libs/wire-api/src/Wire/API/Routes/Public/Brig.hs
```diff
@@ -535,33 +536,56 @@ type UserClientAPI =
...
+           "create-access-token"
+           ( Summary "Create a JWT DPoP access token"
+               :> Description
+                    ( "Create an JWT DPoP access token for the client CSR, given a JWT DPoP proof, specified in the `DPoP` header. \
+                      \The access token will be returned as JWT DPoP token in the `DPoP` header."
+                    )
+               :> QualifiedCaptureUserId "uid"
+               :> "clients"
+               :> CaptureClientId "client"
+               :> "access-token"
```
fisx, line 552:
```suggestion
               :> Capture "cid" ClientId  -- (unless there is a dedicated combinator for that like for uid above?)
               :> "access-token"
```

libs/wire-api/src/Wire/API/Routes/Public/Brig.hs
```diff
@@ -535,33 +536,56 @@ type UserClientAPI =
...
 type NewNonce name method statusCode =
   Named
     name
-    ( Summary "Get a new nonce for a client CSR, specified in the response header `Replay-Nonce` as a uuidv4 in base64url encoding"
+    ( Summary "Get a new nonce for a client CSR"
+        :> Description "Get a new nonce for a client CSR, specified in the response header `Replay-Nonce` as a uuidv4 in base64url encoding."
         :> ZUser
         :> "nonce"
         :> "clients"
         :> MultiVerb1
```
fisx, line 572:
```suggestion
        :> Capture "cid" ClientId  -- (unless there is a dedicated combinator for that like for uid above?)
        :> MultiVerb1
```

this allows us to pass the `ClientId` to the `newNonce` function that needs it as of https://github.com/wireapp/wire-server/pull/2655

libs/wire-api/src/Wire/API/Routes/Public/Brig.hs
```diff
@@ -535,33 +536,56 @@ type UserClientAPI =
...
 newtype NonceHeader = NonceHeader Nonce
   deriving (Eq, Show)
   deriving newtype (FromByteString, ToByteString, ToParamSchema, ToHttpApiData, FromHttpApiData)
 
-instance AsHeaders '[NonceHeader, Text] () Nonce where
-  fromHeaders (I (NonceHeader nonce) :* (_ :* Nil), _) = nonce
-  toHeaders nonce = (I (NonceHeader nonce) :* (I "no-store" :* Nil), ())
+instance AsHeaders '[NonceHeader, CacheControl] () (Nonce, CacheControl) where
+  toHeaders (n, cc) = (I (NonceHeader n) :* (I cc :* Nil), ())
+  fromHeaders (I (NonceHeader n) :* (I cc :* Nil), ()) = (n, cc)
```
fisx, line 588:
```suggestion
  fromHeaders (I (NonceHeader n) :* (I cc :* Nil), ()) = (n, cc)
  toHeaders (n, cc) = (I (NonceHeader n) :* (I cc :* Nil), ())
```

(makes for an easier-to-read diff, maybe?)

libs/wire-api/src/Wire/API/User/Client/DPoPAccessToken.hs
```diff
@@ -0,0 +1,117 @@
...
+instance ToHttpApiData CacheControl where
+  toQueryParam = cs . toByteString'
+
+instance FromHttpApiData CacheControl where
+  parseQueryParam = maybe (Left "Invalid CacheControl") Right . fromByteString' . cs
+
+instance ToParamSchema CacheControl where
+  toParamSchema _ = toParamSchema (Proxy @Text)
+
+instance AsHeaders '[CacheControl] DPoPAccessTokenResponse (DPoPAccessTokenResponse, CacheControl) where
```
fisx, line 115: 1. why is this defined here, but the corresp. `NonceHeader` instance is `API.Public`, see above?
2. does this mean we respond with a header here?  but we need to respond with a json body, is this code left-over from earlier plans?
3. `CacheControl` seems too general to be defined here.  can we either use untyped strings or in something like `libs/servant-utilities`?  (i'm also ok with abusing `wai-utilities` for this)

services/brig/test/integration/API/User/Client.hs
```diff
@@ -962,6 +965,29 @@ testNewNonce brig = do
...
+        <> "{"
+        <> "    \"jti\": \"7535d380-673e-4219-8410-b8df679c306e\","
+        <> "    \"iat\": 1653455836315,"
+        <> "    \"htm\": \"POST\","
+        <> "    \"htu\": \"https://wire.example.com/client/token\","
+        <> "    \"nonce\": \"OZ-5WL9vR5CxaaXoMWpcZQ\","
+        <> "    \"chal\": \"okAJ33Ym/XS2qmmhhh7aWSbBlYy4Ttm1EysqW8I/9ng\","
+        <> "    \"sub\": \"URI:wireapp:SvPfLlwBQi-6oddVRrkqpw/04c7@example.com\","
+        <> "    \"exp\": 1661231836315"
+        <> "}    "
```
fisx, line 989: this is outdated, right?  we can probably get one from the test suite of the rust library that we're calling here?

services/brig/src/Brig/API/Public.hs
```diff
@@ -617,12 +619,19 @@ getRichInfo self user = do
 getClientPrekeys :: UserId -> ClientId -> (Handler r) [Public.PrekeyId]
 getClientPrekeys usr clt = lift (wrapClient $ API.lookupPrekeyIds usr clt)
 
-newNonce :: UserId -> (Handler r) Nonce
+newNonce :: UserId -> (Handler r) (Nonce, CacheControl)
```
fisx, line 622: why?  why not hard-wire the cache-control into the end-point?

services/brig/src/Brig/API/Error.hs
```diff
@@ -171,6 +171,9 @@ clientError ClientMissingLegalholdConsent = StdError (errorToWai @'E.MissingLega
 clientError ClientCodeAuthenticationFailed = StdError verificationCodeAuthFailed
 clientError ClientCodeAuthenticationRequired = StdError verificationCodeRequired
 
+certEnrollmentError :: CertEnrollmentError -> Error
+certEnrollmentError CertEnrollmentError = StdError $ Wai.mkError status400 "generic-cer-enrollment-error" "generic error" --todo(leif): implement all possible errors
```
fisx, line 175:
```suggestion
certEnrollmentError CertEnrollmentError = StdError $ Wai.mkError status400 "generic-cert-enrollment-error" "generic error" --todo(leif): implement all possible errors
```

libs/wire-api/src/Wire/API/Routes/Public/Brig.hs
battermann: I don't understand, `CaptureClientId` is defined like this:

```haskell
type CaptureClientId name = Capture' '[Description "ClientId"] name ClientId
```

services/brig/src/Brig/API/Public.hs
battermann: I like this because this is more explicit.

libs/jwt-tools/src/Jwt/Tools.hs
```diff
@@ -1,4 +1,5 @@
 {-# LANGUAGE ForeignFunctionInterface #-}
+{-# OPTIONS_GHC -fno-warn-unused-top-binds #-}
```
elland, line 2: Is this really necessary here or is this one of those HLS accidents? :D

libs/jwt-tools/src/Jwt/Tools.hs
```diff
@@ -69,35 +111,46 @@ generateDpopToken dpopProof cid nonce uri method maxSkewSecs maxExpiration now b
...
+          Just r -> do
+            mErr <- getError r
+            print mErr
+            mToken <- getToken r
+            pure $ toResult mErr mToken
 
-  ExceptT $ liftIO $ bracket getToken freeDpopAccessToken mkAccessToken
+  let free = maybe (pure ()) free_dpop_access_token
+
+  ExceptT $ liftIO $ bracket before free mkAccessToken
```
elland, line 147: 👍

libs/jwt-tools/src/Jwt/Tools.hs
```diff
@@ -0,0 +1,188 @@
...
+    toResult (Just 9) _ = Left BackendNonceMismatchError
+    toResult (Just 10) _ = Left HtuMismatchError
+    toResult (Just 11) _ = Left HtmMismatchError
+    toResult (Just 12) _ = Left MissingJtiError
+    toResult (Just 13) _ = Left MissingChallengeError
+    toResult (Just 14) _ = Left MissingIatError
+    toResult (Just 15) _ = Left IatError
+    toResult (Just 16) _ = Left MissingExpError
+    toResult (Just 17) _ = Left ExpMismatchError
+    toResult (Just 18) _ = Left ExpError
```
elland, line 171: How about automating that using bounded Enums?

libs/jwt-tools/src/Jwt/Tools.hs
```diff
@@ -0,0 +1,188 @@
...
+    -- this should also not happen, but apparently something went wrong
+    toResult _ _ = Left FfiError
+
+    toCStr :: forall a m. (ToByteString a, MonadIO m) => a -> m CString
+    toCStr = liftIO . newCString . cs . toByteString'
+
+    methodToBS :: StdMethod -> ByteString
+    methodToBS = \case
+      GET -> "GET"
```
elland, line 180: Isn't this `encode . show`?

libs/jwt-tools/src/Jwt/Tools.hs
```diff
@@ -0,0 +1,188 @@
...
+import Numeric (readHex)
+import Wire.API.MLS.Credential
+import Wire.API.MLS.Epoch (Epoch (..))
+import Wire.API.User.Client.DPoPAccessToken
+
+data JwtResponse
+
+foreign import ccall unsafe "generate_dpop_access_token"
+  generate_dpop_access_token ::
+    CString ->
```
elland, line 45: I wonder if it's worth adding comments or type aliases to these arguments for legibility?

libs/jwt-tools/src/Jwt/Tools.hs
battermann: Show instances are meant to produce syntactically correct Haskell expressions, and that is not what we want here, even though it happens to give us the same result. I'd rather have a `ToByteString` instance, which does not exist for `StdMethod`. So we would have to wrap it into a `newtype`, but then I think a function is just as good, here.

libs/jwt-tools/src/Jwt/Tools.hs
```diff
@@ -0,0 +1,188 @@
...
+          (epochNumber maxExpiration)
+          (epochNumber now)
+          backendPubkeyBundleCStr
+
+  let mkAccessToken response = do
+        case response of
+          Nothing -> pure $ Left FfiError
+          Just r -> do
+            mErr <- getError r
+            print mErr
```
elland, line 140: Leftover? :)

battermann: :100:

battermann: nice because we can use applicative composition now: `toResult <$> getError r <*> getToken r`

services/brig/test/integration/API/User/Util.hs
```diff
@@ -575,3 +576,12 @@ nonce m brig uid cid =
         . paths ["clients", toByteString' cid, "nonce"]
         . zUser uid
     )
+
+createAccessToken :: (MonadIO m, MonadHttp m, HasCallStack) => Brig -> UserId -> ClientId -> Proof -> m ResponseLBS
+createAccessToken brig uid cid proof =
+  post
+    ( brig
+        . paths ["clients", toByteString' cid, "access-token"]
```
elland, line 584: Tangential:

Considering we have lots of RPC calls between services, would it make sense to start documenting these kind of calls to make it easier to jump from here to wherever this call is handled inside Brig? Probably not as important inside test files.

elland:
Quick info: when running `make ci` on my VPS, I got a flaky failure on `user/auth/login/throttle/`. It seems to be very dependent on exact timing. If the process is not quick enough to run the login attempts, we don't get the throttle error.

libs/types-common/src/Data/Nonce.hs
```diff
@@ -45,7 +45,14 @@ import Test.QuickCheck.Instances.UUID ()
...
   deriving (Eq, Show)
-  deriving newtype (A.FromJSON, A.ToJSON, S.ToSchema, Arbitrary)
+  deriving newtype (Arbitrary)
+  deriving (A.FromJSON, A.ToJSON, S.ToSchema) via (Schema Nonce)
+
+instance ToSchema Nonce where
+  schema = (cs . toByteString') .= parsedText "Nonce" p
+    where
+      p :: Text -> Either String Nonce
+      p = maybe (Left "Invalid Nonce") Right . fromByteString' . cs
```
fisx, line 55: double-check if there already is a roundtrip test?

libs/wire-api/src/Wire/API/User/Client/DPoPAccessToken.hs
```diff
@@ -0,0 +1,127 @@
...
+--
+-- This program is distributed in the hope that it will be useful, but WITHOUT
+-- ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+-- FOR A PARTICULAR PURPOSE. See the GNU Affero General Public License for more
+-- details.
+--
+-- You should have received a copy of the GNU Affero General Public License along
+-- with this program. If not, see <https://www.gnu.org/licenses/>.
+
+module Wire.API.User.Client.DPoPAccessToken where
```
fisx, line 21: jwt-tools should provide a high-level, but generic interface to the rust code.  i think your goal was to make the interface as low-level as possible

example: `DPoPTokenGenerationError` isn't strictly needed to import the rust code, but it makes using the rust code more convenient without loss of generality.

also, don't import wire-api in jwt-tools.  :-)

fisx: on the other hand, we probably don't want to talk about the servant api inside jwt-tools, and that may account for almost everything in this module.  or you'd have to wrap everything in newtypes to avoid orphans.

nix/pkgs/rusty_jwt_tools/default.nix
```diff
@@ -0,0 +1,21 @@
...
+  src = fetchFromGitHub {
+    owner = "wireapp";
+    repo = "rusty-jwt-tools";
+    rev = "6370cd556f03f6834d0b8043615ffaf0044ef1fa";
+    sha256 = "sha256-vnTvKITie4pu+ISIl/RdYPfb/yWCdCI9eHl1KcZb050=";
+  };
+  cargoBuildFeatures = "haskell";
+  cargoSha256 = "sha256-gR9XKcJslCcXo3EnD3MweWm+pQr4/EAEuFyOq3/l97g=";
+  cargoPatches = [
+    # a patch file to add/update Cargo.lock in the source code
```
fisx, line 18: link to instructions how to re-generated cargo.lock.patch?

libs/wire-api/src/Wire/API/Routes/Public/Brig.hs
fisx: doesn't make sense to me either.  either my suggestions was made on an outdated version of the code, or i was not paying much attention to what i wrote?  let's ignore me!

services/brig/src/Brig/Options.hs
```diff
@@ -627,6 +636,18 @@ defaultNonceTtlSecs = NonceTtlSecs $ 5 * 60 -- 5 minutes
...
 setNonceTtlSecs = fromMaybe defaultNonceTtlSecs . setNonceTtlSecsInternal
 
+defaultDpopMaxSkewSecs :: Word16
+defaultDpopMaxSkewSecs = 1
+
+setDpopMaxSkewSecs :: Settings -> Word16
+setDpopMaxSkewSecs = fromMaybe defaultDpopMaxSkewSecs . setDpopMaxSkewSecsInternal
+
+defaultDpopTokenExpirationTimeSecs :: Word64
+defaultDpopTokenExpirationTimeSecs = 5 * 60 -- 5 minutes
```
fisx, line 646: i suppose seconds is fine here, but the default should be 30 seconds (or less?).  it's an automatic process, no human reaction times in the picture.

services/brig/src/Brig/Effects/JwtTools.hs
```diff
@@ -0,0 +1,70 @@
...
+    JwtTools m (Either Jwt.DPoPTokenGenerationError DPoPAccessToken)
+
+makeSem ''JwtTools
+
+interpretJwtToolsStub :: Members '[Embed IO] r => Sem (JwtTools ': r) a -> Sem r a
+interpretJwtToolsStub = interpret $ \GenerateDPoPAccessToken {} -> do
+  pure $ Right $ DPoPAccessToken "eyJ0eXAiOiJKV1QiLA0KICJhbGciOiJIUzI1NiJ9.eyJpc3MiOiJqb2UiLA0KICJleiOjEzMDA4MTkzODAsDQogImh0dHA6Ly9leGFtcGxlLmNvbS9pc19yb290Ijp0cnVlfQ.dBjftJeZ4CVP-mB92K27uhbUJU1p1r_wW1gFWFOEjXk"
+
+interpretJwtTools :: Members '[Embed IO] r => Sem (JwtTools ': r) a -> Sem r a
+interpretJwtTools = interpret $ \(GenerateDPoPAccessToken pr ci n uri method skew ex now pem) -> do
```
fisx, line 52: i feel there are a lot of levels until we finally get to the rust function.  maybe they are all necessary or reasonable, but can you maybe try and make this one more trivial?  just a hunch.

services/brig/test/integration/API/User/Client.hs
```diff
@@ -107,7 +108,13 @@ tests _cl _at opts p db b c g =
       test p "get /clients/:client - 200" $ testMLSClient b,
       test p "post /clients - 200 multiple temporary" $ testAddMultipleTemporary b g c,
       test p "client/prekeys/race" $ testPreKeyRace b,
-      test p "get/head nonce/clients" $ testNewNonce b
+      test p "get/head nonce/clients" $ testNewNonce b,
+      testGroup
+        "post /clients/:cid/access-token"
+        [ test p "success" $ testCreateAccessToken b,
+          test p "proof missing" $ testCreateAccessTokenMissingProof b,
+          test p "no nonce" $ testCreateAccessTokenNoNonce b
```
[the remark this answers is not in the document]
battermann: I agree, the test cases here are errors that occur before FFI, so I think they are very valid.

libs/wire-api/src/Wire/API/User/Client/DPoPAccessToken.hs
```diff
@@ -0,0 +1,90 @@
...
+  schema =
+    enum @Text "AccessTokenType" $
+      mconcat
+        [ element "DPoP" DPoP
+        ]
+
+data DPoPAccessTokenResponse = DPoPAccessTokenResponse
+  { datrToken :: DPoPAccessToken,
+    datrType :: AccessTokenType,
+    datrExpiresIn :: Word64
```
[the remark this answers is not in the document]
battermann: It comes from the requirements:

> A successful response would contain the access-token in a JSON body. No refresh token is required or desirable in this application.

```
HTTP/1.1 200 OK
Content-Type: application/json
Cache-Control: no-store

{
  "access_token": "HUoU502gM_gtz1PPTH6UT-poymg-wcvZkBC0BP5b8NXyZApeGwBSbqH1CM5z1Lu3.PWNeVtS7eJNZrv1DYn8-7UCqNKfD6ab1SLuS7PXHblfMfv7KBFM0V0EMAjt3kWaLBA6r-pIZ5YSPlqdFu-Feo0LSdlF5LW6NOUyyBfxjw1PhxxDdpikYw3IvAVCUtqPB-oj5YgCE66JY4NjQG2Foyv9TSO4kwyf80cx0bkM3fcXBTrL8KxZWL3IzHLRGg-z6k3px14LjHb57FM5jVpWEkbsr2zfYOJRyQR-kFNeY_3Cpoz5Vws0KswbMg4_4-uMopYsw75srDvqZ9e9Eb8fRgi01ZbZkNZWDgZ7dSC0J14_8c2D9kNP7H_pKWnYzykXjE1rcx1XJ6mP3nxqgRIqFSSYRLuaLO2j57zdpwwPf.hD66MO8CNqFUUEp76TPXSIsAhQIcoh5K_JO2qOOFSO4",
  "token_type": "DPoP",
  "expires_in": 360
}
```

services/brig/src/Brig/Options.hs
```diff
@@ -627,6 +636,18 @@ defaultNonceTtlSecs = NonceTtlSecs $ 5 * 60 -- 5 minutes
 setNonceTtlSecs :: Settings -> NonceTtlSecs
 setNonceTtlSecs = fromMaybe defaultNonceTtlSecs . setNonceTtlSecsInternal
 
+defaultDpopMaxSkewSecs :: Word16
+defaultDpopMaxSkewSecs = 1
```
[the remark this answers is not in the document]
battermann: It comes from the requirements, the rust lib also expects seconds.

libs/jwt-tools/src/Jwt/Tools.hs
```diff
@@ -0,0 +1,290 @@
...
+  ptr <- generate_dpop_access_token dpopProof user client domain nonce uri method maxSkewSecs expiration now backendKeys
+  if ptr /= nullPtr
+    then pure $ Just ptr
+    else pure Nothing
+
+getError :: Ptr JwtResponse -> IO (Maybe Word8)
+getError ptr = do
+  let errorPtr = get_error ptr
+  if errorPtr /= nullPtr
+    then Just . fromIntegral <$> peek errorPtr
```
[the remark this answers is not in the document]
battermann: I really don't think the type annotations are necessary, but I can add them if you insist.

```haskell
  let errorPtr :: Ptr CUChar = get_error ptr
  if errorPtr /= nullPtr
    then Just . (fromIntegral :: CUChar -> Word8) <$> peek errorPtr
    else pure Nothing
  ```

File: libs/wire-api/src/Wire/API/User/Client/DPoPAccessToken.hs [hunk 24]
```diff
@@ -0,0 +1,90 @@
...
+
+newtype Proof = Proof {unProof :: ByteString}
+  deriving (Eq, Show, Generic)
+  deriving newtype (FromByteString, ToByteString)
+
+instance ToHttpApiData Proof where
+  toQueryParam = cs . toByteString'
+
+instance FromHttpApiData Proof where
+  parseQueryParam = maybe (Left "Invalid Proof") Right . fromByteString' . cs
```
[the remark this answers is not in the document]
battermann: None of this works because `ByteString` has no `FromHttpApiData` and `ToHttpApiData` instances.

services/brig/src/Brig/API/Client.hs
```diff
@@ -438,3 +453,41 @@ removeLegalHoldClient uid = do
...
   forM_ legalHoldClients (execDelete uid Nothing)
   wrapHttpClient $ Intra.onUserEvent uid Nothing (UserLegalHoldDisabled uid)
+
+createAccessToken ::
+  (Member JwtTools r, Member Now r, Member PublicKeyBundle r) =>
+  UserId ->
+  ClientId ->
+  StdMethod ->
+  Link ->
+  Maybe Proof ->
```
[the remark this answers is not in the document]
battermann: :100: 
The `Header` combinator defines an optional header, but we can use `Header' '[Required, Strict]` and make it mandatory.

services/brig/src/Brig/API/Client.hs
```diff
@@ -438,3 +453,41 @@ removeLegalHoldClient uid = do
...
+    nonce <- ExceptT $ note NonceNotFound <$> wrapClient (Nonce.lookupAndDeleteNonce uid (cs $ toByteString cid))
+    httpsUrl <- do
+      let urlBs = "https://" <> toByteString' domain <> "/" <> cs (toUrlPiece link)
+      maybe (throwE MisconfiguredRequestUrl) pure $ fromByteString $ urlBs
+    maxSkewSeconds <- Opt.setDpopMaxSkewSecs <$> view settings
+    expiresIn <- Opt.setDpopTokenExpirationTimeSecs <$> view settings
+    now <- fromUTCTime <$> lift (liftSem Now.get)
+    let expiresAt = now & addToEpoch expiresIn
+    pathToKeys <- ExceptT $ note KeyBundleError . Opt.setPublicKeyBundle <$> view settings
+    pubKeyBundle <- ExceptT $ note KeyBundleError <$> liftSem (PublicKeyBundle.get pathToKeys)
```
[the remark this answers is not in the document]
battermann: Well, I can't get it to work, probably because of the `Sem` effect from `PublicKeyBundle.get`.

And for the other places, `domain` and `expiresIn` are used again later, so that doesn't work either.

services/brig/src/Brig/API/Types.hs
```diff
@@ -204,6 +205,13 @@ data AccountStatusError
 data VerificationCodeThrottledError
   = VerificationCodeThrottled RetryAfter
 
+data CertEnrollmentError
```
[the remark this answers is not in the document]
battermann: idk

services/brig/src/Brig/Effects/JwtTools.hs
```diff
@@ -0,0 +1,70 @@
...
+    Epoch ->
+    -- | PEM format concatenated private key and public key of the Wire backend
+    PEMKeys ->
+    JwtTools m (Either Jwt.DPoPTokenGenerationError DPoPAccessToken)
+
+makeSem ''JwtTools
+
+interpretJwtToolsStub :: Members '[Embed IO] r => Sem (JwtTools ': r) a -> Sem r a
+interpretJwtToolsStub = interpret $ \GenerateDPoPAccessToken {} -> do
+  pure $ Right $ DPoPAccessToken "eyJ0eXAiOiJKV1QiLA0KICJhbGciOiJIUzI1NiJ9.eyJpc3MiOiJqb2UiLA0KICJleiOjEzMDA4MTkzODAsDQogImh0dHA6Ly9leGFtcGxlLmNvbS9pc19yb290Ijp0cnVlfQ.dBjftJeZ4CVP-mB92K27uhbUJU1p1r_wW1gFWFOEjXk"
```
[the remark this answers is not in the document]
battermann: It's the in memory interpreter for Sem. Not used ATM, true. We can reintroduce it when necessary.

services/brig/src/Brig/Effects/JwtTools.hs
battermann: I don't see how, honestly.

fisx:
re [3226c9c](https://github.com/wireapp/wire-server/pull/2652/commits/3226c9cfd7deb9709a27a0090ac06bb4971b1781): this isn't done, but it should show you what i want to do differently:

- there should be one haskell enum type that maps all the error codes that can be returned from rust.
- (what i haven't done yet) there should probably also some `data SuperError = RustError DPoPTokenGenerationError | OtherSuperError | YetanotherSuperError`, where `DPoPTokenGenerationError` is fully isomorphic to the rust number range (no collapsing of other errors into `FfiError`).
- `{from,to}Enum` should be used instead of manual enumeration.

I don't insist on this change, but it revealed to me that the error types in haskell do not map to the errors in rust very cleanly, so at least that part I would like to see fixed.

libs/jwt-tools/src/Data/Jwt/Tools.hs
```diff
@@ -198,7 +198,10 @@ generateDpopToken dpopProof uid cid domain nonce uri method maxSkewSecs maxExpir
     toResult _ _ = Left FfiError
 
     toCStr :: forall a m. (ToByteString a, MonadIO m) => a -> m CString
-    toCStr = liftIO . newCString . cs . toByteString'
+    toCStr = liftIO . newCString . toStr
+      where
+        toStr :: a -> String
+        toStr = cs . toByteString'
```
fisx, line 204: `newCString` takes a `String`?!  that's horrible.  :')

but ok, then this is fine i suppose...  you could have also left things inline and written `(cs @ByteString @String)` instead of `cs`.

libs/wire-api/src/Wire/API/Routes/Public/Brig.hs
```diff
@@ -101,7 +101,7 @@ instance AsUnion DeleteSelfResponses (Maybe Timeout) where
   toUnion Nothing = Z (I ())
   fromUnion (Z (I ())) = Nothing
   fromUnion (S (Z (I (DeletionCodeTimeout t)))) = Just t
-  fromUnion (S (S x)) = case x of {}
+  fromUnion (S (S x)) = case x of
```
fisx: Is this because `x :: Void`?  can't you just skip this case, then?  (it's nice to have it there explicitly, just trying to understand.)

fisx (commented):
I've reverted my last commit to make this compile again.  please cherry-pick it if you decide it's useful.

Also, more comments.  :)

libs/wire-api/src/Wire/API/Routes/Public/Brig.hs
battermann: Annoyingly, this change was made by my code formatter, but it does not match the formatting that is expected on CI. I reverted the change.

fisx (approved):
I added two more commits and merged develop.  Please take another look, but otherwise LGTM!

libs/wire-api/src/Wire/API/Asset.hs
```diff
@@ -396,10 +396,6 @@ instance S.ToParamSchema (AssetLocation r) where
       & S.type_ ?~ S.SwaggerString
       & S.format ?~ "url"
 
-instance AsHeaders '[AssetLocation r] Asset (Asset, AssetLocation r) where
-  toHeaders (asset, loc) = (I loc :* Nil, asset)
-  fromHeaders (I loc :* Nil, asset) = (asset, loc)
```
jschaul, line 401: What's the motivation for this removal that appears unrelated to this PR?

battermann: In this PR a generic version of this is introduced so this instance is obsolete.

PR #2655, opened 2022-08-26
Make nonces lookuppable by client ids

fisx:
https://wearezeta.atlassian.net/browse/FS-736 (the part where we need to change the way we store nonces)

## Checklist

 - [x] If HTTP endpoint paths have been added or renamed, check [docs/developer/adding-api-endpoints](https://github.com/wireapp/wire-server/blob/develop/docs/legacy/developer/adding-api-endpoints.md) and follow the steps there.
 - [x] If configuration options have been added or removed, check [docs/developer/adding-config-options](https://github.com/wireapp/wire-server/blob/develop/docs/legacy/developer/adding-config-options.md) and follow the steps there.
- [x] I swear that if I have changed internal end-points, I do not implicitly rely on deployment ordering (brig needing to be deployed before galley), i.e. I have **not** introduced a new internal endpoint in brig and already make use of if in galley, as I'm aware old deployed galleys would throw 500s until all new version have been rolled out and I do not want a few minutes of downtime. Instead, I have thought how to merge brig an galley codebases.
 - [x] I updated **changelog.d** subsections with one or more entries with the following bits of information ([details](https://github.com/wireapp/wire-server/blob/develop/docs/developer/changelog.md)):

### DIFF
--- a/cassandra-schema.cql
+++ b/cassandra-schema.cql
@@ -694,23 +694,6 @@ CREATE TABLE brig_test.team_invitation_info (
     AND read_repair_chance = 0.0
     AND speculative_retry = '99PERCENTILE';
 
-CREATE TABLE brig_test.client_nonce (
-    nonce uuid PRIMARY KEY
-) WITH bloom_filter_fp_chance = 0.01
-    AND caching = {'keys': 'ALL', 'rows_per_partition': 'NONE'}
-    AND comment = ''
-    AND compaction = {'class': 'org.apache.cassandra.db.compaction.SizeTieredCompactionStrategy', 'max_threshold': '32', 'min_threshold': '4'}
-    AND compression = {'chunk_length_in_kb': '64', 'class': 'org.apache.cassandra.io.compress.LZ4Compressor'}
-    AND crc_check_chance = 1.0
-    AND dclocal_read_repair_chance = 0.1
-    AND default_time_to_live = 300
-    AND gc_grace_seconds = 864000
-    AND max_index_interval = 2048
-    AND memtable_flush_period_in_ms = 0
-    AND min_index_interval = 128
-    AND read_repair_chance = 0.0
-    AND speculative_retry = '99PERCENTILE';
-
 CREATE TABLE brig_test.rich_info (
     user uuid PRIMARY KEY,
     json blob
@@ -1102,6 +1085,24 @@ CREATE TABLE brig_test.invitee_info (
     AND crc_check_chance = 1.0
     AND dclocal_read_repair_chance = 0.1
     AND default_time_to_live = 0
+    AND gc_grace_seconds = 864000
+    AND max_index_interval = 2048
+    AND memtable_flush_period_in_ms = 0
+    AND min_index_interval = 128
+    AND read_repair_chance = 0.0
+    AND speculative_retry = '99PERCENTILE';
+
+CREATE TABLE brig_test.nonce (
+    key text PRIMARY KEY,
+    nonce uuid
+) WITH bloom_filter_fp_chance = 0.01
+    AND caching = {'keys': 'ALL', 'rows_per_partition': 'NONE'}
+    AND comment = ''
+    AND compaction = {'class': 'org.apache.cassandra.db.compaction.SizeTieredCompactionStrategy', 'max_threshold': '32', 'min_threshold': '4'}
+    AND compression = {'chunk_length_in_kb': '64', 'class': 'org.apache.cassandra.io.compress.LZ4Compressor'}
+    AND crc_check_chance = 1.0
+    AND dclocal_read_repair_chance = 0.1
+    AND default_time_to_live = 300
     AND gc_grace_seconds = 864000
     AND max_index_interval = 2048
     AND memtable_flush_period_in_ms = 0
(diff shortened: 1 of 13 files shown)


services/brig/src/Brig/Data/Nonce.hs
```diff
@@ -31,18 +31,25 @@ import Imports
...
+  Text ->
   Nonce ->
   m ()
-insertNonce ttl nonce = retry x5 . write insert $ params LocalQuorum (Identity nonce)
+insertNonce ttl key nonce = retry x5 . write insert $ params LocalQuorum (key, nonce)
   where
-    insert :: PrepQuery W (Identity Nonce) ()
-    insert = fromString $ "INSERT INTO client_nonce (nonce) VALUES (?) USING TTL " <> show ttl
+    insert :: PrepQuery W (Text, Nonce) ()
+    insert = fromString $ "INSERT INTO nonce (key, nonce) VALUES (?) USING TTL " <> show ttl
```
battermann, line 40:
```suggestion
    insert = fromString $ "INSERT INTO nonce (key, nonce) VALUES (?, ?) USING TTL " <> show ttl
```

fisx:
This PR does not solve the problem that [we need to change the API to somehow give us the `ClientId`](https://github.com/wireapp/wire-server/blob/bf7efbe7486c87c5b374911b130a449d8f1ba431/services/brig/src/Brig/API/Public.hs#L626) (I though it's derivable from the `Z-Conn` header, but I changed my mind about that).

After `TASTY_PATTERN='/nonce/' make ci package=brig` has passed:

```
cqlsh> select * from brig_test.nonce ;

 key | nonce
-----+--------------------------------------
 123 | e002192e-7447-4a8d-9c26-f6edd7563128

(1 rows)
cqlsh> select * from brig_test.nonce ;

 key | nonce
-----+-------

(0 rows)
```

fisx:
no need to discuss the destructive cassandra update, as (a) the `client_nonce` has never been released, and (b) it cannot be used for anything useful in production yet.

fisx:
If you want to make sure we don't store negative TTLs:

```
$ git diff
diff --git a/services/brig/src/Brig/Data/Nonce.hs b/services/brig/src/Brig/Data/Nonce.hs
index ade768e4e..d11fd8372 100644
--- a/services/brig/src/Brig/Data/Nonce.hs
+++ b/services/brig/src/Brig/Data/Nonce.hs
@@ -30,11 +30,11 @@ import Imports
 
 insertNonce ::
   (MonadClient m, MonadReader Brig.App.Env m) =>
-  Int32 ->
+  Word32 ->
   Text ->
   Nonce ->
   m ()
-insertNonce ttl key nonce = retry x5 . write insert $ params LocalQuorum (key, nonce, ttl)
+insertNonce ttl key nonce = retry x5 . write insert $ params LocalQuorum (key, nonce, fromIntegral ttl)
   where
     insert :: PrepQuery W (Text, Nonce, Int32) ()
     insert = "INSERT INTO nonce (key, nonce) VALUES (?, ?) USING TTL ?"
```

(`Word*` doesn't have Cql instances, or at least I couldn't find any.)

battermann (approved):

libs/wire-api/src/Wire/API/Routes/Public/Brig.hs
```diff
@@ -547,8 +547,9 @@ type NewNonce name method statusCode =
     name
     ( Summary "Get a new nonce for a client CSR, specified in the response header `Replay-Nonce` as a uuidv4 in base64url encoding"
         :> ZUser
-        :> "nonce"
         :> "clients"
+        :> CaptureClientId "client"
+        :> "nonce"
```
fisx, line 552: :+1:

cassandra-schema.cql
```diff
@@ -1093,9 +1093,12 @@ CREATE TABLE brig_test.invitee_info (
     AND speculative_retry = '99PERCENTILE';
 
 CREATE TABLE brig_test.nonce (
-    key text PRIMARY KEY,
-    nonce uuid
-) WITH bloom_filter_fp_chance = 0.01
+    user uuid,
```
fisx, line 1096: Good catch, thanks!

I was going to say that we wanted to avoid typing the key, but I can't think of a hypothetical use case where asking for a `UserId` doesn't make sense.  So, good, I think.

libs/types-common/src/Data/Nonce.hs
```diff
@@ -70,3 +77,24 @@ instance Cql Nonce where
...
   toCql = toCql . unNonce
   fromCql v = Nonce <$> fromCql v
+
+newtype NonceTtlSecs = NonceTtlSecs {unNonceTtlSecs :: Word32}
+  deriving (Eq, Show, Generic)
+  deriving (A.FromJSON, A.ToJSON, S.ToSchema) via (Schema NonceTtlSecs)
+
+-- | Convert 'Word32' to 'Int32', with clipping if it doesn't fit.
+word32ToInt32 :: Word32 -> Int32
+word32ToInt32 = fromIntegral . min (fromIntegral (maxBound @Int32))
```
fisx, line 87: i know it's an impractical corner case in *this* context, but hum...  can we use `Range` to make sure we don't run into this overflow?

PR #2657, opened 2022-08-29
[FS-672] Improve swagger docs for Client fields.

stephen-smith:
[FS-672](https://wearezeta.atlassian.net/browse/FS-672)

## Checklist

 - [x] I updated **changelog.d** subsections with one or more entries.

### DIFF
--- a/libs/wire-api/src/Wire/API/User/Client.hs
+++ b/libs/wire-api/src/Wire/API/User/Client.hs
@@ -79,7 +79,7 @@ where
 
 import qualified Cassandra as Cql
 import Control.Applicative
-import Control.Lens (over, view, (?~), (^.))
+import Control.Lens hiding (element, enum, set, (#), (.=))
 import Data.Aeson (FromJSON (..), ToJSON (..))
 import qualified Data.Aeson as A
 import qualified Data.Aeson.Key as Key
@@ -96,6 +96,7 @@ import Data.Qualified
 import Data.Schema
 import qualified Data.Semigroup as Semigroup
 import qualified Data.Set as Set
+import Data.Swagger hiding (Schema, ToSchema, schema)
 import qualified Data.Swagger as Swagger
 import qualified Data.Swagger.Build.Api as Doc
 import qualified Data.Text.Encoding as Text.E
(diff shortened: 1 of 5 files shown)


pcapriotti:
It would be good to also improve the example. Right now it reads:
```
    "mls_public_keys": {
      "additionalProp1": "string",
      "additionalProp2": "string",
      "additionalProp3": "string"
    },
```
which is not very useful.

libs/wire-api/src/Wire/API/User/Client.hs
```diff
@@ -477,6 +477,22 @@ data Client = Client
...
+             . (example ?~ toJSON (Map.fromList $ map (,exampleValue) keys))
+         )
+    & named "MLSPublicKeys"
+  where
+    keys :: [SignatureSchemeTag]
+    keys = [minBound .. maxBound]
+    exampleValue = fromMaybe (toJSON ("base64==" :: Text)) (valueSchema ^. doc . example)
+    valueSchema = base64Schema
+    mapSchema :: ValueSchema SwaggerDoc MLSPublicKeys
+    mapSchema = map_ valueSchema
```
fisx, line 494:
```suggestion

    exampleValue :: A.Value
    exampleValue = fromMaybe (toJSON ("base64==" :: Text)) (valueSchema ^. doc . example)

    mapSchema :: ValueSchema SwaggerDoc MLSPublicKeys
    mapSchema = map_ base64Schema
```

(I think these are questions of taste.)

fisx (commented):
when building this, running `http://localhost:8080/swagger-ui/`, and opening http://localhost:8080/swagger-ui/, I see this:

2022-09-01-120754_1920x1080_scrot

`MLSPublicKeys` shouldn't be empty, should it?

stephen-smith:
> `MLSPublicKeys` shouldn't be empty, should it?

It doesn't have any fields, so yes, I think it is supposed to be empty in that display.

As best as I can tell, that's normal for map/dict types in swagger.

The example should have the only valid key and an example value, and there should also be a description if you expand the name.

image

image

pcapriotti (approved):
LGTM

fisx (approved):

PR #2663, opened 2022-08-31
Added linting scripts.

elland:
As our next step, I've added entries to the makefile to allow for quick linting changes as you work. Next step would be to add this to either git hooks and or CI.

Also updated some file changes to be linted and updated rules to remove some annoyances.

Regarding checkpoint 3 below, is there anything I should document about it?

## Checklist

### DIFF
--- a/libs/wire-api/src/Wire/API/User/IdentityProvider.hs
+++ b/libs/wire-api/src/Wire/API/User/IdentityProvider.hs
@@ -190,7 +190,7 @@ instance ToSchema IdPMetadataInfo where
           & properties .~ properties_
           & minProperties ?~ 1
           & maxProperties ?~ 1
-          & type_ .~ Just SwaggerObject
+          & type_ ?~ SwaggerObject
     where
       properties_ :: InsOrdHashMap Text (Referenced Schema)
       properties_ =
--- a/tools/hlint.sh
+++ b/tools/hlint.sh
@@ -14,7 +14,7 @@ while getopts ':f:m:' opt
               echo "WARNING: not linting tests."
             elif [ "$f" = "changeset" ]; then
               files=$(git diff --name-only HEAD | grep \.hs\$)
-              echo "WARNING: linting all test files with changes.  this may lead to some hard to fix warnings/errors, it is safe to ignore these!"
+              echo "WARNING: linting test files with changes. This may lead to some hard to fix warnings/errors, it is safe to ignore those!"
             else
               usage
             fi
--- a/tools/hlint.sh
+++ b/tools/hlint.sh
@@ -11,8 +11,10 @@ while getopts ':f:m:' opt
          f) f=${OPTARG}
             if [ "$f" = "all" ]; then
               files=$(find libs/ services/ -not -path "*/test/*" -name "*.hs")
+              echo "WARNING: not linting tests."
             elif [ "$f" = "changeset" ]; then
               files=$(git diff --name-only HEAD | grep \.hs\$)
+              echo "WARNING: linting all test files with changes.  this may lead to some hard to fix warnings/errors, it is safe to ignore these!"
             else
               usage
             fi
--- a/tools/hlint.sh
+++ b/tools/hlint.sh
@@ -1,0 +1,49 @@
+#!/usr/bin/env bash
+
+usage() { echo "Usage: $0 -f [all, changeset] -m [check, inplace]" 1>&2; exit 1; }
+
+files=''
+check=true
+
+while getopts ':f:m:' opt
+ do
+     case $opt in
+         f) f=${OPTARG}
+            if [ "$f" = "all" ]; then
+              files=$(find libs/ services/ -not -path "*/test/*" | grep \.hs\$)
+            elif [ "$f" = "changeset" ]; then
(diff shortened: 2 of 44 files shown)


tools/hlint.sh
```diff
@@ -0,0 +1,49 @@
...
+            ;;
+         *) usage;;
+     esac
+done
+
+if [ -z "${f}" ] || [ -z "${m}" ]; then
+    usage
+fi
+
+count=$(echo "$files" | wc -l | xargs echo)
```
stephen-smith, line 37: * Why `xargs echo` at the end?
* This is count too high if we have a filename containing a NL.
* This will not match the number of loop iterations below if we have a filename contains SP or TAB.

tools/hlint.sh
```diff
@@ -0,0 +1,49 @@
...
+
+files=''
+check=true
+
+while getopts ':f:m:' opt
+ do
+     case $opt in
+         f) f=${OPTARG}
+            if [ "$f" = "all" ]; then
+              files=$(find libs/ services/ -not -path "*/test/*" | grep \.hs\$)
```
stephen-smith, line 13: `find` can do the `grep` for you; `-name '*.hs'` should work.

tools/hlint.sh
```diff
@@ -0,0 +1,49 @@
...
+
+if [ -z "${f}" ] || [ -z "${m}" ]; then
+    usage
+fi
+
+count=$(echo "$files" | wc -l | xargs echo)
+
+echo "Analysing $count file(s)…"
+
+for f in $files
```
stephen-smith, line 41: For filenames containing SP, TAB, or NL (at least), this will iterate multiple times for a single file.

Haskell modules can't have any "funky" characters in them, so that should be fine to commit/merge; just an FYI.

stephen-smith (approved):

tools/hlint.sh
elland: Thanks,  I'm not that versed in bash script! :)

tools/hlint.sh
```diff
@@ -0,0 +1,49 @@
...
+check=true
+
+while getopts ':f:m:' opt
+ do
+     case $opt in
+         f) f=${OPTARG}
+            if [ "$f" = "all" ]; then
+              files=$(find libs/ services/ -not -path "*/test/*" -name "*.hs")
+            elif [ "$f" = "changeset" ]; then
+              files=$(git diff --name-only HEAD | grep \.hs\$)
```
akshaymankar, line 15: Running `-f all` doesn't lint tests, but running `-f changeset` runs tests. Is this correct?
Aside from that, why not lint tests?

fisx: > Aside from that, why not lint tests?

I'd very much like to have tests linted, but separate PR maybe?  Get it to work first?

stephen-smith (approved):

tools/hlint.sh
elland: The reason behind ignoring tests is that we currently have certain amount of "unsafe code" that's perfectly safe inside their test environments. Most notably we use index accessors a lot when dealing with prekeys in tests. 

Should I add back test folders to the all flag or remove it from changeset?

I left it like this bc when going over the whole Codebase I didn't want to have to rework all those test cases or add a bunch of ignore rules. But I did want my own tests to be linted. 

I should've brought that up in the chat :)

fisx: as i said, this can be done separately.

the question that remains is "what's with the difference between changeset and all"?

elland: I thought I answered that for changeset if I am touching tests I wanted them linted. I can exclude the tests there as well or add test lint as it's own extra flag.

libs/wire-api/src/Wire/API/User/IdentityProvider.hs
```diff
@@ -190,7 +190,7 @@ instance ToSchema IdPMetadataInfo where
           & properties .~ properties_
           & minProperties ?~ 1
           & maxProperties ?~ 1
-          & type_ .~ Just SwaggerObject
+          & type_ ?~ SwaggerObject
```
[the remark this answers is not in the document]
fisx: I'm wrong, all good!

```
let x :: (Maybe Char, Int) = (Nothing, 2) in (x, x & _1 ?~ '3', x & _1 .~ Just '3')
==> ((Nothing,2),(Just '3',2),(Just '3',2))
```

tools/hlint.sh
fisx: ah, sorry.  got to the point where i understand the keyword now!  i'm just concerned that if you add a test in a file that is hard to lint, you'll get all those warnings, but then again you can ignore them.

maybe explain it better in the script?  let me try!

fisx:
```suggestion
              files=$(find libs/ services/ -not -path "*/test/*" -name "*.hs")
              echo "WARNING: not linting tests."
            elif [ "$f" = "changeset" ]; then
              files=$(git diff --name-only HEAD | grep \.hs\$)
              echo "WARNING: linting all test files with changes.  this may lead to some hard to fix warnings/errors, it is safe to ignore these!"
```

fisx (approved):

PR #2671, opened 2022-09-05
Fix wire-server metrics-dashboards queries

amitsagtani97:
Update the prometheus queries for the message-stats and services dashboard in wire-server-metrics

Some of the metrics were not working, replaced them with the one used in the staging/prod grafana metrics.


amitsagtani97:
Turns out the issue was with the service monitor configuration and the same dashboard worked as expected without updating the queries.
Also, the wire-server-metrics support is dropped in recent updates.

Closing this pr.

PR #2674, opened 2022-09-05
Deleting clients creates MLS remove proposals

smatting:
## Checklist

### DIFF
--- a/services/galley/src/Galley/API/Clients.hs
+++ b/services/galley/src/Galley/API/Clients.hs
@@ -22,6 +22,7 @@ module Galley.API.Clients
   )
 where
 
+import Data.Either.Combinators (whenLeft)
 import Data.Hex
 import Data.Id
 import Data.Proxy
@@ -44,10 +45,10 @@ import Galley.Env
 import Galley.Types.Clients (clientIds, fromUserClients)
 import Imports
 import Network.Wai
-import Network.Wai.Predicate hiding (setStatus)
-import Network.Wai.Utilities
+import Network.Wai.Predicate hiding (Error, setStatus)
+import Network.Wai.Utilities hiding (Error)
 import Polysemy
-import qualified Polysemy.Error as Poly
+import Polysemy.Error
 import Polysemy.Input
 import qualified Polysemy.TinyLog as P
 import qualified System.Logger as Log
@@ -99,7 +100,7 @@ rmClientH ::
          ListItems p1 ConvId,
          ListItems p1 (Remote ConvId),
          MemberStore,
-         Poly.Error InternalError,
+         Error InternalError,
          ProposalStore,
          P.TinyLog
        ]
@@ -119,8 +120,7 @@ rmClientH (usr ::: cid) = do
     rpc = fedClient @'Galley @"on-client-removed"
     goConvs :: Range 1 1000 Int32 -> ConvIdsPage -> Local UserId -> Sem r ()
     goConvs range page lusr = do
-      loc :: Local () <- input
-      let (localConvs, remoteConvs) = partitionQualified loc (mtpResults page)
+      let (localConvs, remoteConvs) = partitionQualified lusr (mtpResults page)
       for_ localConvs $ \convId -> do
         mConv <- getConversation convId
         for_ mConv $ \conv -> do
@@ -134,22 +134,20 @@ rmClientH (usr ::: cid) = do
         goConvs range newCids lusr
 
     removeRemoteMLSClients :: Range 1 1000 [Remote ConvId] -> Sem r ()
-    removeRemoteMLSClients cids = do
-      for_ (bucketRemote (fromRange cids)) $ \remoteConvs -> do
+    removeRemoteMLSClients convIds = do
+      for_ (bucketRemote (fromRange convIds)) $ \remoteConvs -> do
         runFederatedEither remoteConvs (rpc (ClientRemovedRequest usr cid (tUnqualified remoteConvs)))
           >>= logAndIgnoreError "Error in onConversationUpdated call" usr
 
-    logAndIgnoreError message usr' res = do
-      case res of
-        Left federationError ->
-          P.err
-            ( Log.msg
-                ( "Federation error while notifying remote backends of a client deletion (Galley). "
-                    <> message
-                    <> " "
-                    <> show federationError
-                )
-                . Log.field "user" (show usr')
-                . Log.field "client" (hex . T.unpack . client $ cid)
-            )
-        Right _ -> pure ()
+    logAndIgnoreError message usr' res =
+      whenLeft res $ \federationError ->
+        P.err
+          ( Log.msg
+              ( "Federation error while notifying remote backends of a client deletion (Galley). "
+                  <> message
+                  <> " "
+                  <> show federationError
+              )
+              . Log.field "user" (show usr')
+              . Log.field "client" (hex . T.unpack . client $ cid)
+          )
--- a/services/galley/src/Galley/API/Clients.hs
+++ b/services/galley/src/Galley/API/Clients.hs
@@ -22,16 +22,40 @@ module Galley.API.Clients
   )
 where
 
+import Data.Hex
 import Data.Id
+import Data.Proxy
+import Data.Qualified
+import Data.Range
+import Data.String.Conversions
+import qualified Data.Text as T
+import Data.Time
+import Galley.API.Error
+import Galley.API.MLS.Removal
+import qualified Galley.API.Query as Query
+import Galley.API.Util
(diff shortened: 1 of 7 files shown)


services/galley/test/integration/API/MLS.hs
```diff
@@ -1630,3 +1632,81 @@ testBackendRemoveProposalLocalConvRemoteLeaver = do
         for_ bobClients $ \(_, ref) ->
           WS.assertMatch_ (5 # WS.Second) wsA $
             wsAssertBackendRemoveProposal bob qcnv ref
+
+-- for_ kprefs $ \kp ->
+--   WS.assertMatch_ (5 # WS.Second) wsA $ \notification ->
+--     void $
+--       wsAssertBackendRemoveProposal (pUserId bob) conversation kp notification
```
mdimjasevic: I suppose this has been removed in subsequent commits, but this is just a reminder in case it's not the case.

services/galley/test/integration/API/MLS.hs
```diff
@@ -1630,3 +1632,81 @@ testBackendRemoveProposalLocalConvRemoteLeaver = do
...
+testBackendRemoveProposalLocalConvLocalClient :: TestM ()
+testBackendRemoveProposalLocalConvLocalClient = do
+  [alice, bob] <- createAndConnectUsers [Nothing, Nothing]
+
+  runMLSTest $ do
+    [alice1, bob1, bob2] <- traverse createMLSClient [alice, bob, bob]
+    traverse_ uploadNewKeyPackage [bob1, bob2]
+    (_, qcnv) <- setupMLSGroup alice1
+    void $ createAddCommit alice1 [bob] >>= sendAndConsumeCommit
+    [(bobA, kpBobA), (bobB, _kpBobB)] <- getClientsFromGroupState alice1 bob
```
mdimjasevic, line 1615: Just for readability purposes, I propose to add assertions `bobA == bob1` and `bobB == bob2`.

services/galley/test/integration/API/MLS.hs
```diff
@@ -1630,3 +1632,81 @@ testBackendRemoveProposalLocalConvRemoteLeaver = do
...
+
+    let mock req = case frRPC req of
+          "on-conversation-updated" -> pure (Aeson.encode ())
+          "on-new-remote-conversation" -> pure (Aeson.encode EmptyResponse)
+          "get-mls-clients" ->
+            pure
+              . Aeson.encode
+              . Set.fromList
+              . map (flip ClientInfo True)
+              . toList
```
mdimjasevic, line 1659: `toList` on a list is an identity function, right?

services/galley/test/integration/API/MLS.hs
```diff
@@ -752,6 +743,7 @@ testRemoteAppMessage = do
           "on-conversation-updated" -> pure (Aeson.encode ())
           "on-new-remote-conversation" -> pure (Aeson.encode EmptyResponse)
           "on-mls-message-sent" -> pure (Aeson.encode EmptyResponse)
+          "mls-welcome" -> pure (Aeson.encode EmptyResponse)
```
mdimjasevic, line 722: Why is this needed in existing test cases? I'd assume the endpoint isn't called, otherwise the catch-all clause would trigger and we'd get an error.

That said, is is needed in the tests introduced by this PR?

mdimjasevic (commented):
Overall this looks good, but I have some questions and comments that are inlined.

services/galley/src/Galley/API/Clients.hs
```diff
@@ -61,9 +85,71 @@ addClientH (usr ::: clt) = do
...
+        goConvs range newCids lusr
+
+    removeRemoteMLSClients :: Range 1 1000 [Remote ConvId] -> Sem r ()
+    removeRemoteMLSClients cids = do
+      for_ (bucketRemote (fromRange cids)) $ \remoteConvs -> do
+        runFederatedEither remoteConvs (rpc (ClientRemovedRequest usr cid (tUnqualified remoteConvs)))
+          >>= logAndIgnoreError "Error in onConversationUpdated call" usr
+
+    logAndIgnoreError message usr' res = do
+      case res of
```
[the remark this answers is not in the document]
smatting: ah nice, didn't know about this one

services/galley/src/Galley/API/Clients.hs
```diff
@@ -61,9 +85,71 @@ addClientH (usr ::: clt) = do
...
+        let nextState = mtpPagingState page
+            nextQuery = GetPaginatedConversationIds (Just nextState) range
+        newCids <- Query.conversationIdsPageFrom lusr nextQuery
+        goConvs range newCids lusr
+
+    removeRemoteMLSClients :: Range 1 1000 [Remote ConvId] -> Sem r ()
+    removeRemoteMLSClients cids = do
+      for_ (bucketRemote (fromRange cids)) $ \remoteConvs -> do
+        runFederatedEither remoteConvs (rpc (ClientRemovedRequest usr cid (tUnqualified remoteConvs)))
+          >>= logAndIgnoreError "Error in onConversationUpdated call" usr
```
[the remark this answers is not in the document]
smatting: Yes, it is. `rpc` is bound to `fedClient @'Galley @"on-client-removed"`. The reason we define this name outside is because creating the servant client `rpc` is costly apparently

services/galley/test/integration/API/MLS.hs
smatting: Right, it was a set before I rebased this PR

services/galley/test/integration/API/MLS.hs
smatting: Errors get thrown actually when it's not implemented, however the tests still pass.

services/galley/src/Galley/API/Clients.hs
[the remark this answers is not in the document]
mdimjasevic: I am not sure I follow. I am asking if instead of:
```haskell
  logAndIgnoreError "Error in onConversationUpdated call" usr
```
it should have been:
```haskell
  logAndIgnoreError "Error in onClientRemoved call" usr
```
instead.

services/galley/test/integration/API/MLS.hs
smatting: This assumption can't be made. I went with a solution involving `find`

smatting:
@mdimjasevic Thanks for the review! I followed all your suggetions in https://github.com/wireapp/wire-server/pull/2674/commits/85575f030ce666b614aebe6da49846b59077f281

services/galley/test/integration/API/MLS.hs
mdimjasevic: I am not sure I follow. I made this change locally on top of the PR:
```diff
diff --git a/services/galley/test/integration/API/MLS.hs b/services/galley/test/integration/API/MLS.hs
index 0927979dc..d562d1f2c 100644
--- a/services/galley/test/integration/API/MLS.hs
+++ b/services/galley/test/integration/API/MLS.hs
@@ -1613,6 +1613,8 @@ testBackendRemoveProposalLocalConvLocalClient = do
     (_, qcnv) <- setupMLSGroup alice1
     void $ createAddCommit alice1 [bob] >>= sendAndConsumeCommit
     [(bobA, kpBobA), (bobB, _kpBobB)] <- getClientsFromGroupState alice1 bob
+    liftIO $ assertEqual "bob1 /= bobA" bob1 bobA
+    liftIO $ assertEqual "bob2 /= bobB" bob2 bobB
```
and the test passes. These are assumptions I read the test with, and these are just codified versions of the assumptions.

pcapriotti (approved):
LGTM

services/galley/test/integration/API/MLS.hs
smatting: What I mean is that the order of `getClientsFromGroupState` is not guaranteed to be the order of generation. It might work, but I wouldn't trust it. For the test it's arbitraty whether `bob1` or `bob2` is deleted. But in any case with the latest changes I can pick `bob1` and have a less confusing test

mdimjasevic: > What I mean is that the order of `getClientsFromGroupState` is not guaranteed to be the order of generation. It might work, but I wouldn't trust it. For the test it's arbitraty whether `bob1` or `bob2` is deleted. But in any case with the latest changes I can pick `bob1` and have a less confusing test

Oh, thanks for clarifying this! I wish we had `Map ClientIdentity KeyPackageRef` in the return type of `getClientsFromGroupState` instead of `[(ClientIdentity, KeyPackageRef)]`, which would clearly tell there's no order to expect among the pairs of associated values.

PR #2678, opened 2022-09-07
Refactor MLS test framework

smatting:
This PR introduces a new system for simulating clients in MLS tests. Every client is maintained in a separate directory, and is kept up to date with MLS messages sent to a test group.

## Checklist


services/galley/test/integration/API/Util.hs
```diff
@@ -2792,10 +2807,60 @@ wsAssertBackendRemoveProposal fromUser convId kpref n = do
...
+    getMLSMessageData (EdMLSMessage bs) = bs
+    getMLSMessageData d = error ("Excepected EdMLSMessage, but got " <> show d)
+
+wsAssertAddProposal ::
+  HasCallStack =>
+  Qualified UserId ->
+  Qualified ConvId ->
+  Notification ->
+  IO ByteString
+wsAssertAddProposal fromUser convId n = do
```
mdimjasevic, line 2824: Given this one is MLS-specific, maybe a better place for it would be the `MLS.Util` module.

mdimjasevic (approved):

services/galley/test/integration/API/Util.hs
pcapriotti: I think it's fine to have all these event assertions there.

PR #2679, opened 2022-09-07
removed/replaced todo comments

battermann:
Replaced some "todo" comments with something more meaningful.

No changelog necessary, IMO.

## Checklist

### DIFF
--- a/services/spar/src/Spar/Sem/IdPConfigStore/Mem.hs
+++ b/services/spar/src/Spar/Sem/IdPConfigStore/Mem.hs
@@ -46,7 +46,9 @@ idPToMem = evState . evEff
     evEff = reinterpret @_ @(State TypedState) $ \case
       InsertConfig iw ->
         modify' (insertConfig iw)
-      NewHandle _tid -> pure $ IdPHandle "IdP 1" --todo(leif): generate a new handle
+      NewHandle _tid ->
+        -- Same handle for all IdPs is good enough, for now
+        pure $ IdPHandle "IdP 1"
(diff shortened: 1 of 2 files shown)


smatting (approved):

services/spar/src/Spar/Sem/IdPConfigStore/Mem.hs
```diff
@@ -46,7 +46,9 @@ idPToMem = evState . evEff
     evEff = reinterpret @_ @(State TypedState) $ \case
       InsertConfig iw ->
         modify' (insertConfig iw)
-      NewHandle _tid -> pure $ IdPHandle "IdP 1" --todo(leif): generate a new handle
+      NewHandle _tid ->
+        -- Same handle for all IdPs is good enough, for now
```
elland, line 50: Can we elaborate on this? Why is a hardcoded handle fine?

battermann: Because this is just a label shown to the team admin in the UI. IMO uniqueness is irrelevant for tests,

fisx: :+1: we'll see right away if this is not enough any more in the future, and then we can build something more sophisticated.

fisx (approved):

fisx:
> No changelog necessary, IMO.

it would not have been more work to write the changelog entry instead of this line.  and it wouldn't have prompted me to make things worse!  :-)

anyway, fine.

PR #2682, opened 2022-09-08
Add alias to GET /conversations/{cnv} endpoint for LH devices

smatting:
This PR adds an alias for the endpoint
```
GET     /conversations/{cnv}
```
at
```
GET     /legalhold/conversations/{cnv}
```
and permits legalhold devices access to it.

We use an alias instead of allowing `GET     /conversations/{cnv}` because the nginz `acl` files are not expressive enough to match only this endpoint (and not e.g. `POST /conversations/join`)

Checklist:

 - [x] Update nginz config in helm: `charts/nginz/values.yaml`
 - [x] Update nginz config in the demo: `deploy/services-demo/conf/nginz/nginx.conf`


battermann (approved):

jschaul (commented):
If the intention is only to give legalhold access to one thing (read some information), wouldn't it be better to create another (alias) endpoint `/legalhold/conversations/{cnv}` and whitelist only that? That would restrict what legalhold can do.

smatting:
> If the intention is only to give legalhold access to one thing (read some information), wouldn't it be better to create another (alias) endpoint `/legalhold/conversations/{cnv}` and whitelist only that? That would restrict what legalhold can do.

Good idea! @dkovacevic, would that work for you?

jschaul:
> > If the intention is only to give legalhold access to one thing (read some information), wouldn't it be better to create another (alias) endpoint `/legalhold/conversations/{cnv}` and whitelist only that? That would restrict what legalhold can do.
> 
> Good idea! 

Thanks for creating the alias!

PR #2688, opened 2022-09-14
Add commit bundle support

smatting:
- wire-api: Add types and test (wip)
- Add endpoint in galley

## Checklist

 - [x] Update nginz config in helm: `charts/nginz/values.yaml`
 - [x] Update nginz config in the demo: `deploy/services-demo/conf/nginz/nginx.conf`


libs/wire-api/test/unit/Test/Wire/API/MLS.hs
```diff
@@ -214,6 +216,49 @@ testRemoveProposalMessageSignature = withSystemTempDirectory "mls" $ \tmp -> do
       )
       Nothing
 
+testParseGroupInfoBundle :: IO ()
+testParseGroupInfoBundle = withSystemTempDirectory "mls" $ \tmp -> do
+  qcid <- do
+    let c = newClientId 0x3ae58155
+    usr <- flip Qualified (Domain "example.com") <$> (Id <$> UUID.nextRandom)
+    pure (userClientQid usr c)
+  void . liftIO $ spawn (cli qcid tmp ["init", qcid]) Nothing
```
elland, line 225: Since #2678 was merged, I think we can use that instead, right?

elland: Should we extract those into their own internal test lib so we can use them across the codebase? @pcapriotti ?

pcapriotti: I'm not to keen on doing this. It's not that much code, anyway, and the unit test setup is much simpler anyway, since we only need to extract one particular message, and not keep track of the state of several clients.

stefanwire (approved):

PR #2689, opened 2022-09-14
stop deleting clients when no prekeys exist for these clients

stefanwire:
https://wearezeta.atlassian.net/browse/FS-966

## Checklist

### DIFF
--- a/services/brig/src/Brig/API/Client.hs
+++ b/services/brig/src/Brig/API/Client.hs
@@ -388,6 +388,11 @@ execDelete u con c = do
 -- not exist, since there must be no client without prekeys,
 -- thus repairing any inconsistencies related to distributed
 -- (and possibly duplicated) client data.
+--
+-- FUTUREWORK: As of 2022-09-14, we see a lot of clients without prekeys. Also,
+-- this function does not delete clients in Brig, but only in Galley and
+-- Gundeck. To debug the issue properly, we stop deleting clients and log
+-- occurances instead in which we would otherwise delete the client.
 noPrekeys ::
   ( MonadReader Env m,
     MonadMask m,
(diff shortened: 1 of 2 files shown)


jschaul (commented):
I'm not convinced the approach of pushing code to production then looking at logs is fruitful in the backend's case; the feedback loop is too long and the risk of damage done of leaving *more* state inconsistency for more users is high. If this state can be achieved with current code, then it's possible to write an integration test showing clients without prekeys existing either in brig, or galley, or both. Then the production code could be changed.

The above is my gut feeling; I've not been involved with this issue before and don't have prior context; but it sounds like a bug regarding clients was introduced accidentally at some point in the recent past; Perhaps a review of code changes here would be more fruitful.

services/brig/src/Brig/API/Client.hs
```diff
@@ -404,7 +409,8 @@ noPrekeys u c = do
     field "user" (toByteString u)
       ~~ field "client" (toByteString c)
       ~~ msg (val "No prekey found. Ensuring client does not exist.")
-  Intra.rmClient u c
+  -- Intra.rmClient u c
```
elland, line 412: I assume at some point we need to actually delete the clients correctly everywhere or is that not the case? Also would that require some script to prune the required Casandras?

stefanwire:
The discussion continues in https://wearezeta.atlassian.net/browse/FS-966.

jschaul (changes requested):
"requesting changes" here to block this PR from being merged until the discussion in FS-966 has lead to a consensus.

smatting:
Closing this. See details in https://wearezeta.atlassian.net/browse/FS-966

PR #2690, opened 2022-09-15
Added hlint pr check makefile cmd.

elland:
Sometimes we create smaller commits during the course of work, and don't remember or take the time to wait for the linter between those steps. Sometimes it's useful / helpful to be able to run hlint over all haskell files touched by a PR as a last ste, or as a CI step.

### DIFF
--- a/tools/hlint.sh
+++ b/tools/hlint.sh
@@ -13,7 +13,7 @@ while getopts ':f:m:' opt
               files=$(find libs/ services/ -not -path "*/test/*" -name "*.hs")
               echo "WARNING: not linting tests."
             elif [ "$f" = "pr" ]; then
-              files=$(git diff --name-only develop | grep \.hs\$)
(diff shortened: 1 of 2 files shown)


tools/hlint.sh
```diff
@@ -12,6 +12,9 @@ while getopts ':f:m:' opt
             if [ "$f" = "all" ]; then
               files=$(find libs/ services/ -not -path "*/test/*" -name "*.hs")
               echo "WARNING: not linting tests."
+            elif [ "$f" = "pr" ]; then
+              files=$(git diff --name-only develop | grep \.hs\$)
```
smatting, line 16:
```suggestion
              files=$(git diff --name-only origin/develop... | grep \.hs\$)
```
then you don't to update your local develop branch. a `git fetch` would be enough

fisx (approved):

PR #2693, opened 2022-09-16
[SQSERVICE- 1509] 2FA in the context of sso

battermann:
https://wearezeta.atlassian.net/browse/SQSERVICES-1509

## Checklist


fisx (approved):
Good idea!  This is an obvious improvement over the current situation in which sso+2fa just doesn't work.  Let's merge this and think about supporting 2fa+sso elsewhere!

Did you run the test without the fix and reproduce the problem?

battermann:
> Did you run the test without the fix and reproduce the problem?
yes

PR #2695, opened 2022-09-16
Make sure integration tests use most recent API version

pcapriotti:
## Checklist


services/brig/test/integration/Util.hs
```diff
@@ -135,6 +136,38 @@ type Spar = Request -> Request
 
 data FedClient (comp :: Component) = FedClient HTTP.Manager Endpoint
 
+-- | Note: Apply this function last when composing (Request -> Request) functions
+apiVersion :: ByteString -> Request -> Request
```
mdimjasevic, line 140: A nit-pick: why not make the first argument a specific version, e.g., `V1` instead of a `ByteString`? When there will be a need to change use sites of `apiVersion`, we can grep for `apiVersion` in the project, but I thought it'd be even nicer if we used versions instead of their string representations.

services/galley/test/integration/API.hs
```diff
@@ -455,8 +454,9 @@ postCryptoMessageVerifyRejectMissingClientAndRepondMissingPrekeysJson = do
       assertMismatchWithMessage (Just "client mismatch") [(eve, Set.singleton ec)] [] []
   let x = responseJsonUnsafeWithMsg "ClientMismatch" r1
   -- Fetch all missing clients prekeys
+  b <- view tsUnversionedBrig
   r2 <-
-    post (b . zUser alice . path "/users/prekeys" . json (missingClients x))
+    post (b . zUser alice . path "v1/users/prekeys" . json (missingClients x))
```
mdimjasevic, line 459: This looks strange: using `b` as an unversioned brig and then followed by a path prefixed with "v1". Is that a mistake?

services/galley/test/integration/API.hs
```diff
@@ -485,8 +484,9 @@ postCryptoMessageVerifyRejectMissingClientAndRepondMissingPrekeysProto = do
   pure r1
     !!! assertMismatchWithMessage (Just "client mismatch") [(eve, Set.singleton ec)] [] []
   -- Fetch all missing clients prekeys
+  b <- view tsUnversionedBrig
   r2 <-
-    post (b . zUser alice . path "/users/prekeys" . json (missingClients x))
+    post (b . zUser alice . path "v1/users/prekeys" . json (missingClients x))
```
mdimjasevic, line 489: ditto

services/galley/test/integration/API/Util.hs
```diff
@@ -938,12 +955,16 @@ getConvs u r s = do
...
+  UserId ->
+  ListConversations ->
+  m ResponseLBS
 listConvs u req = do
-  g <- viewGalley
+  g <- view tsUnversionedGalley
   post $
     g
-      . path "/conversations/list/v2"
+      . path "/v1/conversations/list/v2"
```
mdimjasevic, line 967: Unversioned in addition to being at v1? Is there no function `apiVersion` in Galley tests? I think it'd be worth it creating a small test package that has utilities like `apiVersion` instead of copy-pasting it several times.

services/galley/test/integration/API/Util.hs
```diff
@@ -997,45 +1018,44 @@ listRemoteConvs remoteDomain uid = do
...
   ConvId ->
   m ResponseLBS
 postQualifiedMembers zusr invitees conv = do
-  g <- viewGalley
+  g <- view tsUnversionedGalley
   let invite = InviteQualified invitees roleNameWireAdmin
   post $
     g
-      . paths ["conversations", toByteString' conv, "members", "v2"]
+      . paths ["v1", "conversations", toByteString' conv, "members", "v2"]
```
mdimjasevic, line 1031: unversioned + v1 with the v2 suffix? This is a rather strange combination.

services/galley/test/integration/API/Util.hs
```diff
@@ -1819,13 +1837,13 @@ connectUsersWith ::
...
     connectTo v = do
-      b <- view tsBrig
+      b <- view tsUnversionedBrig
       r1 <-
         post
           ( b
               . zUser u
               . zConn "conn"
-              . path "/connections"
+              . paths ["v1", "connections"]
```
mdimjasevic, line 1846: unversioned + v1?

services/galley/test/integration/API/Util.hs
```diff
@@ -1862,10 +1880,10 @@ connectWithRemoteUser self other = do
 -- | A copy of 'postConnection' from Brig integration tests.
 postConnection :: UserId -> UserId -> TestM ResponseLBS
 postConnection from to = do
-  brig <- view tsBrig
+  brig <- view tsUnversionedBrig
   post $
     brig
-      . path "/connections"
+      . paths ["v1", "connections"]
```
mdimjasevic, line 1886: ditto

mdimjasevic (approved):
Overall this looks good.

I have concerns with using an unversioned service and then prefixing manually "v1". Also, I found a combination of unversioned + v1 + the v2 suffix.

The Util module in Galley tests has lots of unversioned + v1. Could this be changed to something like `apiVersion` in other services?

services/galley/test/integration/API.hs
pcapriotti: Why would it be a mistake? I don't understand why you think this and the other cases below are strange.

mdimjasevic: Tests pass and that suggests it works, but I find it strange to read: it is unversioned and at version v1. Maybe if this was hidden away in a utility function (such as `apiVersion` that takes a version as input), then I wouldn't have problems with it.

pcapriotti: Sure, we can add an `apiVersion`, but I don't quite understand what's strange about it. To make a versioned path, you start with the unversioned one and you add a version prefix to it. It's just the normal thing to do :shrug:.

services/galley/test/integration/API/Util.hs
pcapriotti: Right, that's why we're removing those v2 suffixes everywhere.

pcapriotti:
I'll merge for now. We can think about changing `apiVersion` to take a version value later. It's not that important. As for those unversioned + v1 calls, I hope the confusion is resolved. I prefer it like this, instead of the hackish approach that we were forced to take in brig, because there's no string manipulation involved: the path is cleanly constructed starting from the unversioned one, instead of having to remove the existing prefix first.

services/galley/test/integration/API.hs
mdimjasevic: I agree, this is how it is achieved under the hood, i.e., these are the steps to construct a versioned path. I was just trying to say I'd prefer it put in a utility function (where we care about details of how to construct a valid versioned path) compared to exposing this detail at every use site to avoid always reading it's unversioned first and then we make it versioned.

PR #2700, opened 2022-09-19
[SQSERVICES-1643] Servantify brig account API 2 - `GET /activate`

battermann:

## Checklist

### DIFF
--- a/libs/wire-api/src/Wire/API/User/Activation.hs
+++ b/libs/wire-api/src/Wire/API/User/Activation.hs
@@ -182,7 +182,7 @@ instance ToSchema ActivationResponse where
     Schema.objectWithDocModifier "ActivationResponse" (description ?~ "Response body of a successful activation request") $
       ActivationResponse
         <$> activatedIdentity Schema..= userIdentityObjectSchema
-        <*> activatedFirst Schema..= Schema.fieldWithDocModifier "first" (description ?~ "Whether this is the first successful activation (i.e. account activation).") Schema.schema
+        <*> activatedFirst Schema..= (fromMaybe False <$> Schema.optFieldWithDocModifier "first" (description ?~ "Whether this is the first successful activation (i.e. account activation).") Schema.schema)
 
 modelActivationResponse :: Doc.Model
 modelActivationResponse = Doc.defineModel "ActivationResponse" $ do
--- a/libs/wire-api/src/Wire/API/User/Activation.hs
+++ b/libs/wire-api/src/Wire/API/User/Activation.hs
@@ -40,14 +40,19 @@ module Wire.API.User.Activation
   )
 where
 
+import Control.Lens ((?~))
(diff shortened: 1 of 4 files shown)


libs/wire-api/src/Wire/API/User/Activation.hs
```diff
@@ -158,6 +175,14 @@ data ActivationResponse = ActivationResponse
...
   deriving stock (Eq, Show, Generic)
   deriving (Arbitrary) via (GenericUniform ActivationResponse)
+  deriving (ToJSON, FromJSON, S.ToSchema) via Schema ActivationResponse
+
+instance ToSchema ActivationResponse where
+  schema =
+    Schema.objectWithDocModifier "ActivationResponse" (description ?~ "Response body of a successful activation request") $
+      ActivationResponse
+        <$> activatedIdentity Schema..= userIdentityObjectSchema
+        <*> activatedFirst Schema..= (fromMaybe False <$> Schema.optFieldWithDocModifier "first" (description ?~ "Whether this is the first successful activation (i.e. account activation).") Schema.schema)
```
fisx, line 185:
```suggestion
        <*> activatedFirst Schema..= (fromMaybe False <$> Schema.optFieldWithDocModifier "first" (description ?~ "Whether this is the first successful activation of this account.") Schema.schema)
```

(Not sure I understand, though.  How often can one account be activated successfully?)

services/brig/src/Brig/API/Public.hs
```diff
@@ -1020,15 +997,15 @@ respFromActivationRespWithStatus = \case
 activateKeyH :: JSON ::: JsonRequest Public.Activate -> (Handler r) Response
 activateKeyH (_ ::: req) = do
   activationRequest <- parseJsonBody req
-  respFromActivationRespWithStatus <$> activate activationRequest
+  respFromActivationRespWithStatus <$> activate' activationRequest
 
-activateH :: Public.ActivationKey ::: Public.ActivationCode -> (Handler r) Response
-activateH (k ::: c) = do
+activate :: Public.ActivationKey -> Public.ActivationCode -> (Handler r) ActivationRespWithStatus
+activate k c = do
```
fisx, line 1003: (`functionNameH` traditionally means "`functionName`, but there is a servant handler with that name, and this is just the wrapper with the wai-route plumbing in it; will be removed once we switch to servant".  So it's probably/usually a good idea to merge the two, but yes, in this case I'm not sure how to do that off the top of my head.)

fisx (approved):

libs/wire-api/src/Wire/API/User/Activation.hs
battermann: This is from the original swagger documentation. IMO it should not be changed in this PR. Also it is not obvious to me either how this should be phrased correctly.

PR #2704, opened 2022-09-19
[SQSERVICES-1643] Servantify brig account API 6- `POST password-reset/complete`

battermann:

## Checklist

### DIFF
--- a/libs/wire-api/src/Wire/API/Routes/Public/Brig.hs
+++ b/libs/wire-api/src/Wire/API/Routes/Public/Brig.hs
@@ -55,7 +55,7 @@ import Wire.API.User.Activation
 import Wire.API.User.Client
 import Wire.API.User.Client.Prekey
 import Wire.API.User.Handle
-import Wire.API.User.Password (NewPasswordReset)
+import Wire.API.User.Password (CompletePasswordReset, NewPasswordReset)
 import Wire.API.User.RichInfo (RichInfoAssocList)
 import Wire.API.User.Search (Contact, RoleFilter, SearchResult, TeamContact, TeamUserSearchSortBy, TeamUserSearchSortOrder)
 import Wire.API.UserMap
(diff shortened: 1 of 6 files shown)


libs/wire-api/src/Wire/API/Routes/Public/Brig.hs
```diff
@@ -468,6 +468,15 @@ type AccountAPI =
...
            )
+    :<|> Named
+           "post-password-reset-complete"
+           ( Summary "Complete a password reset."
+               :> CanThrow 'InvalidPwResetCode
+               :> "password-reset"
+               :> "complete"
+               :> ReqBody '[JSON] CompletePasswordReset
+               :> MultiVerb 'POST '[JSON] '[RespondEmpty 200 "Password reset successful."] ()
+           )
```
akshaymankar, line 479: I didn't understand that you meant `CanThrow` not being type checked in brig when you mentioned it in standup. IIRC, I was using `MultiVerb` and `FromUnion`, `ToUnion` classes to ensure things were type checked.

libs/wire-api/src/Wire/API/Error/Brig.hs
```diff
@@ -194,3 +195,5 @@ type instance
 type instance MapError 'InvalidPwResetKey = 'StaticError 400 "invalid-key" "Invalid email or mobile number for password reset."
 
 type instance MapError 'DuplicatePwResetCode = 'StaticError 409 "code-exists" "A password reset is already in progress."
+
+type instance MapError 'InvalidPwResetCode = 'StaticError 400 "invalid-code" "Invalid password reset code."
```
akshaymankar, line 199: This is currently only used for swagger, I think we should also use this to create the actually error being thrown, you can use `errorToWai`. There are examples in the `Brig.API.Error` module.

akshaymankar (changes requested):

akshaymankar (approved):

PR #2705, opened 2022-09-19
[SQSERVICES-1643] Servantify brig account API 7 - `POST /password-reset/:key`

battermann:

## Checklist

### DIFF
--- a/libs/wire-api/src/Wire/API/Routes/Public/Brig.hs
+++ b/libs/wire-api/src/Wire/API/Routes/Public/Brig.hs
@@ -96,7 +96,7 @@ instance AsUnion DeleteSelfResponses (Maybe Timeout) where
   toUnion Nothing = Z (I ())
   fromUnion (Z (I ())) = Nothing
   fromUnion (S (Z (I (DeletionCodeTimeout t)))) = Just t
-  fromUnion (S (S x)) = case x of {}
+  fromUnion (S (S x)) = case x of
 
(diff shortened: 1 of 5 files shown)


libs/wire-api/src/Wire/API/Routes/Public/Brig.hs
```diff
@@ -476,6 +476,19 @@ type AccountAPI =
...
+    :<|> Named
+           "post-password-reset-key-deprecated"
+           ( Summary "Complete a password reset."
+               :> CanThrow 'PasswordResetInProgress
+               :> CanThrow 'InvalidPasswordResetKey
+               :> CanThrow 'InvalidPasswordResetCode
+               :> CanThrow 'ResetPasswordMustDiffer
+               :> Description "DEPRECATED: Use 'POST /password-reset/complete'."
+               :> "password-reset"
+               :> QueryParam' '[Required, Strict, Description "An opaque key for a pending password reset."] "key" PasswordResetKey
```
fisx, line 488: `Caption`, not `QueryParam`.

Now I'm concerned about there not be any tests that cover this change, though.  Maybe better leave this one alone until we remove it?

fisx (changes requested):
I like the changes in `Wire.API.User.Password`!  But the end-point either needs to be fixed or rolled back.  Your call!

fisx (approved):

PR #2706, opened 2022-09-19
[FS-917] Added an optional sender (ClientId) to MLS message events.

elland:
Client side asked for the sender id as part of the top level payload, but this raised a feel questions since it's a breaking change that will require clients to adapt how they are currently deserialising the JSON payload, for this reason we've left it as a draft, until we've reached a consensus on how to move forward with this.

## Checklist


elland:
We can close this as a #wontfix.

PR #2709, opened 2022-09-20
Nit pick

fisx:
since we're not doing "you have to write a change log entry" any more...  :-)

this is a semantics-preserving local refactoring.  some change in the past made this code look very silly, and i fixed that in passing.


elland:
Reran your lint step, to undo the false positive. 👍

elland (approved):

PR #2712, opened 2022-09-20
Use hlint with no summary instead of grep.

elland:
Turns out hlint already provided a way to silence the annoying no hints output.
On top of that, I added a quick early break with a better error message in case of grep failures b/c no files were found.

### DIFF
--- a/tools/hlint.sh
+++ b/tools/hlint.sh
@@ -35,6 +35,14 @@ while getopts ':f:m:' opt
      esac
 done
 
+ret=$?
+if [ $ret -ne 0 ]; then
+  echo
+else
+  echo "No files to check"
+  exit 1
+fi
+
 if [ -z "${f}" ] || [ -z "${m}" ]; then
     usage
 fi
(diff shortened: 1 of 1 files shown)


tools/hlint.sh
```diff
@@ -35,6 +35,14 @@ while getopts ':f:m:' opt
      esac
 done
 
+ret=$?
```
elland, line 38: If someone who actually knows shellscript has a better suggestion, please let me know :)

akshaymankar: Usually fighting bash to get nice error messages is not very fruitful and leaves error handling to the script writer. So I prefer it if we just add `set -euo pipefail` to all the scripts. You still have the option to error check and print nice errors, you just have to use `if ! <cmd>` instead of `if [ $? -ne 0 ]`. This way we don't have to worry about every line that is added to the bash script failing and error getting lost.

akshaymankar (approved):

Issue #2660, opened 2022-08-30
nginz: Keep /usr/include/zauth.h compatible with other projects


kiufta:
nginz needs libzauth with installs [this file](https://github.com/wireapp/wire-server/blob/master/libs/libzauth/libzauth-c/src/zauth.h) into _any place you want_.

Nevertheless it would great to have only one _/usr/include/zauth.h_ file which should be compatible with zeromq's [file at the same place](https://github.com/zeromq/czmq/blob/master/include/zauth.h)

PR #2714, opened 2022-09-20
Improve libzauth ACL syntax

pcapriotti:
This PR replaces the prefix-tree matcher used in libzauth for matching ACL paths with a simple regex-based matcher, which constructs a single regular expression containing all possible paths. This makes it trivial to accept user-provided regular expressions in the ACL language itself.

## Checklist


elland:
Ignore if unrelated, but would this be a good place to also deal with #2660?

pcapriotti:
> Ignore if unrelated, but would this be a good place to also deal with #2660?

I don't think so, it's pretty much unrelated, as this PR only affects the rust library and the change is mostly internal and backwards compatible. Moving installed files around requires some extra care.

PR #2715, opened 2022-09-20
Better signal handling for proxy, stern

fisx:
https://wearezeta.atlassian.net/browse/SQPIT-1431

## Checklist

### DIFF
--- a/changelog.d/3-bug-fixes/sqpit-1431
+++ b/changelog.d/3-bug-fixes/sqpit-1431
@@ -1,0 +1,1 @@
+Less surprising handling of SIGINT, SIGTERM for proxy, stern
--- a/changelog.d/3-bug-fixes/sqpit-1431
+++ b/changelog.d/3-bug-fixes/sqpit-1431
@@ -1,1 +1,1 @@
-Less surprising handling of SIGINT, SIGTERM for proxy, stern
+Less surprising handling of SIGINT, SIGTERM for proxy, stern. Increase grace period for shutdown from 5s to 30s for all services. 
--- a/services/cannon/src/Cannon/Run.hs
+++ b/services/cannon/src/Cannon/Run.hs
@@ -41,17 +41,13 @@ import Data.Text (pack, strip)
 import Data.Text.Encoding (encodeUtf8)
 import Imports hiding (head)
 import qualified Network.Wai as Wai
-import Network.Wai.Handler.Warp hiding (run)
 import qualified Network.Wai.Middleware.Gzip as Gzip
 import Network.Wai.Utilities.Server
 import Servant
 import qualified System.IO.Strict as Strict
 import qualified System.Logger.Class as LC
 import qualified System.Logger.Extended as L
-import System.Posix.Signals
-import qualified System.Posix.Signals as Signals
 import System.Random.MWC (createSystemRandom)
-import UnliftIO.Concurrent (myThreadId, throwTo)
 import qualified Wire.API.Routes.Internal.Cannon as Internal
 import Wire.API.Routes.Public.Cannon
 import Wire.API.Routes.Version.Wai
@@ -87,10 +83,8 @@ run o = do
       server =
         hoistServer (Proxy @PublicAPI) (runCannonToServant e) publicAPIServer
           :<|> hoistServer (Proxy @Internal.API) (runCannonToServant e) internalServer
-  tid <- myThreadId
-  void $ installHandler sigTERM (signalHandler (env e) tid) Nothing
-  void $ installHandler sigINT (signalHandler (env e) tid) Nothing
-  runSettings s app `finally` do
+  runSettingsWithShutdown s app 5 `finally` do
+    runWS (env e) drain
     Async.cancel refreshMetricsThread
     L.close (applog e)
   where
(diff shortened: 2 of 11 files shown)


akshaymankar:
Can you please try `SIGTERM`? `SIGNIT` is actually handled correctly, as the issue says.

fisx:
> Can you please try `SIGTERM`? `SIGNIT` is actually handled correctly by default, as the issue says.

Doesn't change anything; the handler still doesn't get caught.

Anyway that's besides the point, no?  I'm installing a `SIGINT` handler that doesn't get called if I send a `SIGINT`.  I must be holding it wrong, somehow.

fisx:
ah, i see: brig is calling `runSettingsWithShutdown`, and cannon isn't.  that function installs its own signal handlers and thus removes the ones i've just installed.  so it looks like brig is fine?  which services throw the 502 / which services are they talking to at the time.  I'm going to check the others.

akshaymankar:
> ah, i see: brig is calling `runSettingsWithShutdown`, and cannon isn't. that function installs its own signal handlers and thus removes the ones i've just installed. so it looks like brig is fine? which services throw the 502 / which services are they talking to at the time. I'm going to check the others.

Nice catch! We don't actually have logs from last deployment to prod because they're gone. So, I guess we can either spam staging with a lot of requests and do a deployment and find out or, we could just go through all the services and see which ones don't use `runSettingsWithShutdown`.

fisx:
> or, we could just go through all the services and see which ones don't use `runSettingsWithShutdown`.

that was easy!  there is only proxy who calls `runSettings` directly.  patch coming up!

fisx:
we should also add an hlint rule that forbids us to use `runSettings` anywhere...  :-)

services/cannon/src/Cannon/Run.hs
```diff
@@ -87,10 +83,8 @@ run o = do
       server =
         hoistServer (Proxy @PublicAPI) (runCannonToServant e) publicAPIServer
           :<|> hoistServer (Proxy @Internal.API) (runCannonToServant e) internalServer
-  tid <- myThreadId
-  void $ installHandler sigTERM (signalHandler (env e) tid) Nothing
-  void $ installHandler sigINT (signalHandler (env e) tid) Nothing
-  runSettings s app `finally` do
+  runSettingsWithShutdown s app 5 `finally` do
```
akshaymankar, line 86: I am not sure about the impact of this 5 second wait, so I will run a couple of experiments to see that the draining happens correctly. Earlier we were also not closing the listen socket, so impact of that is another mystery to me.
I am being super careful because this functionality is important and we don't have nice ways of automatically testing it.

elland:
After running hlint with the changes from #2718 it found a possible spot where this is also not used, in Federator.Response.

akshaymankar:
I also suspect that the reason for 502s is that we have 5 second grace period for requests which are already running, maybe we can bump it to 30s. Specially in cargohold where the uploads could take some time to finish.

fisx:
> After running hlint with the changes from #2718 it found a possible spot where this is also not used, in Federator.Response.

unrelated.

fisx:
> I also suspect that the reason for 502s is that we have 5 second grace period for requests which are already running, maybe we can bump it to 30s. Specially in cargohold where the uploads could take some time to finish.

I'll do that :+1:

services/cannon/src/Cannon/Run.hs
fisx: 67946bd37a4354cfff512b35d761d6fdf8433360 (also 0f264d2eaec2b709d86820620f892f169157d286)

fisx: > [67946bd](https://github.com/wireapp/wire-server/commit/67946bd37a4354cfff512b35d761d6fdf8433360) (also [0f264d2](https://github.com/wireapp/wire-server/commit/0f264d2eaec2b709d86820620f892f169157d286))

no, that's different commits.  but the github UI is messing with me, so i'll just leave you to the actually commit history and the "files changed" tab.  :)

tools/stern/src/Stern/API.hs
```diff
@@ -78,7 +78,7 @@ start :: Opts -> IO ()
 start o = do
   e <- newEnv o
   s <- Server.newSettings (server e)
-  runSettings s (pipeline e)
+  Server.runSettingsWithShutdown s (pipeline e) Nothing `finally` pure () -- nothing to clean up?
```
akshaymankar, line 81:
```suggestion
  Server.runSettingsWithShutdown s (pipeline e) Nothing
```

Nit: Why write that `finally`? Stern probably has nothing important running, so it makes sense that it doesn't have anything to cleanup.

changelog.d/3-bug-fixes/sqpit-1431
```diff
@@ -0,0 +1 @@
+Less surprising handling of SIGINT, SIGTERM for proxy, stern
```
akshaymankar, line 1:
```suggestion
Less surprising handling of SIGINT, SIGTERM for proxy, stern. Increase grace period for shutdown from 5s to 30s for all services. 
```

akshaymankar (approved):
Looks good. Please update the changelog though!

tools/stern/src/Stern/API.hs
fisx: wasn't sure about this, just thought sometimes it's nice to be explicit about doing nothing.  makes it clear that it's not an oversight, but intentional.  maybe here it's just noise, though.

PR #2717, opened 2022-09-20
[FS-918] Verify signatures of incoming MLS message

elland:
We attempt here to verify signature of incoming messages when we are able to, to add some level of sanity checks.

### DIFF
--- a/services/galley/src/Galley/API/MLS/Message.hs
+++ b/services/galley/src/Galley/API/MLS/Message.hs
@@ -14,7 +14,6 @@
 --
 -- You should have received a copy of the GNU Affero General Public License along
 -- with this program. If not, see <https://www.gnu.org/licenses/>.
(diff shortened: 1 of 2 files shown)


services/galley/src/Galley/API/MLS/Message.hs
```diff
@@ -868,14 +874,31 @@ checkExternalProposalSignature ::
...
+  _ -> pure ()
+  where
+    throwUnsupported = throwS @'MLSUnsupportedProposal
+
+    f ci = do
+      UserClientsFull clientsFull <- lookupClientsFull [ciUser ci]
+      for_ clientsFull $ \clients' ->
+        maybe throwUnsupported f2 (find (\c -> clientId c == ciClient ci) clients')
+
+    f2 client =
```
elland, line 890: TODO(elland): clean this up.

PR #2718, opened 2022-09-21
Add new custom hlint rule for runSetting.

elland:
Also applies hlint again to the whole codebase (excluding tests), as we had some drift between finalising hlint and new PRs being merged without being linted / having CI catch those cases.

I also disabled the pipefail from the script, as that would short-circuit the linter on first issue found. Hopefully that doesn't mess with CI.

PS: This will fail CI linters phase until #2715 has been merged.

### DIFF
--- a/tools/hlint.sh
+++ b/tools/hlint.sh
@@ -1,7 +1,5 @@
 #!/usr/bin/env bash
 
-set -euo pipefail
-
 usage() { echo "Usage: $0 -f [all, changeset] -m [check, inplace]" 1>&2; exit 1; }
 
 files=''
--- a/.hlint.yaml
+++ b/.hlint.yaml
@@ -15,3 +15,6 @@
 # custom rules:
 - hint: { lhs: (() <$), rhs: void }
 - hint: { lhs: return, rhs: pure }
+## We want the latter to properly handle signals.
+- error: { name: Use shutdown, lhs: runSettings, rhs: runSettingsWithShutdown }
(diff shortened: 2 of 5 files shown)


tools/hlint.sh
```diff
@@ -1,7 +1,5 @@
 #!/usr/bin/env bash
 
-set -euo pipefail
-
```
akshaymankar, line 4: Why remove this?

akshaymankar (commented):
Does checking only files changed by PR mean that adding new rules won't run over everything in CI? Maybe we should reconsider this approach and just run hlint on everything.

tools/hlint.sh
elland: From the PR description:

> I also disalbed the pipefail from the script, as that would short-circuit the linter on first issue found. Hopefully that doesn't mess with CI.

Does that make sense?

elland:
@akshaymankar We might want to do that indeed, at least on CI. For this case, or when I touch the rules, I do run on everything (except tests), but that's a fraught process that requires people to remember to do manual steps, and as you mentioned somewhere else, linting runs parallel to the other CI steps, so it taking 5x or more longer wouldn't impact us, as it would remain much faster than integration tests. I'll bring this up to the rest of the team.

tools/hlint.sh
akshaymankar: Ah, sorry I didn't read the description! I am sure there is way we can find, but I don't have much time to spend on this, so I will take back my objection.

akshaymankar (approved):

PR #2721, opened 2022-09-22
[FS-923] Store Per-conversation GroupInfo Structure and Expose It

mdimjasevic:
Tracked by https://wearezeta.atlassian.net/browse/FS-923.

## Checklist


services/galley/src/Galley/Cassandra/Instances.hs
```diff
@@ -200,13 +199,14 @@ instance Cql GroupId where
...
 
-  toCql = CqlBlob . runPut . serialiseMLS
-
-  fromCql (CqlBlob b) = mapLeft T.unpack $ decodeMLS @GroupInfoBundle b
-  fromCql _ = Left "GroupInfoBundle: blob expected"
+  toCql = CqlBlob . LBS.fromStrict . rmRaw
+  fromCql (CqlBlob b) =
+    mapLeft T.unpack $
+      decodeMLS @(RawMLS PublicGroupStateTBS) b
+  fromCql _ = Left "PublicGroupStateTBS: blob expected"
```
mdimjasevic, line 209: @pcapriotti , does this meet your expectations? The part we cannot avoid (if we want to have the `RawMLS PublicGroupStateTBS` type here) is decoding, can we? An alternative is to have a `ByteString` (or a newtype around it) instead that would hold just the raw bytes, but then I don't know how this will play out once we start supporting the delta and by-reference kinds of public group states.

pcapriotti: I'd rather just read a bytestring from the database. For the moment, we don't need to parse it, and if we eventually do, we're probably not going to serialise it as a blob anyway (but I'm not sure).

Also, why is it `PublicGroupStateTBS`? Shouldn't it be the whole `PublicGroupState`?

mdimjasevic: Ok, the bytestring variant sounds reasonable. I'll do that.

Regarding `PublicGroupStateTBS` vs. `PublicGroupState`, I've put the TBS one assuming that for these purposes we don't need the signature of the state. As I type this, it sounds like it'd be useful to have a signature as well, wouldn't it? I can put it for there. Should I use `PublicGroupState` instead?

pcapriotti: How were you planning to reconstruct the signature when serving it to the clients?

cassandra-schema.cql
```diff
@@ -458,6 +458,7 @@ CREATE TABLE galley_test.conversation (
     deleted boolean,
     epoch bigint,
     group_id blob,
+    group_info_bundle blob,
```
pcapriotti, line 461: I guess this needs to be regenerated?

libs/wire-api/src/Wire/API/Routes/Public/Galley.hs
```diff
@@ -219,6 +220,25 @@ type ConversationAPI =
                :> "roles"
                :> Get '[Servant.JSON] ConversationRolesList
            )
+    :<|> Named
+           "get-group-info"
+           ( Summary "Get MLS group information"
+               :> CanThrow 'ConvNotFound
+               :> CanThrow 'ConvAccessDenied
```
smatting, line 227:
```suggestion
```
I think it's no lopnger needed

services/galley/test/integration/API/MLS/Util.hs
```diff
@@ -890,3 +893,59 @@ receiveNewRemoteConv conv gid = do
...
+  ) =>
+  UserId ->
+  Qualified ConvId ->
+  m ResponseLBS
+getGroupInfo sender qcnv = do
+  galley <- viewGalley
+  get
+    ( galley
+        . paths
+          [ "v2",
```
pcapriotti, line 943: If you rebase, you won't need the v2 prefix.

pcapriotti (approved):
LGTM

services/galley/src/Galley/API/Util.hs
```diff
@@ -494,8 +494,11 @@ getMember ::
 getMember p u = noteS @e . find ((u ==) . p)
 
 getConversationAndCheckMembership ::
-  Members '[ConversationStore, ErrorS 'ConvNotFound, ErrorS 'ConvAccessDenied] r =>
-  UserId ->
+  forall uid mem r.
+  ( Members '[ConversationStore, ErrorS 'ConvNotFound, ErrorS 'ConvAccessDenied] r,
+    IsConvMemberId uid mem
+  ) =>
+  uid ->
```
smatting, line 501: I think this is a leftover from a different approach, we can remove

cassandra-schema.cql
mdimjasevic: Yes, fixed in a recent commit.

services/galley/test/integration/API/MLS/Util.hs
mdimjasevic: Thanks! Fixed.

services/galley/test/integration/API/MLS/Util.hs
```diff
@@ -890,3 +893,59 @@ receiveNewRemoteConv conv gid = do
...
+                  { cjUsers = pure joiner,
+                    cjRole = roleNameWireMember
+                  }
+          }
+  void $
+    runFedClient
+      @"on-conversation-updated"
+      client
+      (qDomain conv)
+      cu
```
smatting, line 928: where do we use this

smatting (approved):

PR #2722, opened 2022-09-22
Start linting test files

elland:
Added test files to the lint script, they'll no longer be skipped. We can now add these checks to CI and get warnings when it fails.

I've also added an option to bail out of linting if we hit any issues.

The changelog is sadly massive, let me know if it would make sense to do it step by step or if you're fine reviewing it like this.


stephen-smith:
Are the hie.yaml changes necessary?  They look to me like something that hie-boot or implicit-hie or something else could dynamically generate.  And, perhaps more relevantly, something that, if we make this change, developers will have to update in future commits.

elland:
@stephen-smith that was updated with implicit-hie, but in any case shouldn't be part of this PR, reverting it 👍

services/brig/test/integration/API/User/Handles.hs
```diff
@@ -335,7 +335,8 @@ assertCanFind brig from target = do
   let targetHandle = fromMaybe (error "Impossible") (userHandle target)
   get (apiVersion "v1" . brig . path "/users" . queryItem "handles" (toByteString' targetHandle) . zUser (userId from)) !!! do
     const 200 === statusCode
-    const (userHandle target) === (>>= (listToMaybe >=> profileHandle)) . responseJsonMaybe
+    const (userHandle target) === ((listToMaybe >=> profileHandle) <=< responseJsonMaybe)
```
fisx, line 338:
```suggestion
    const (userHandle target) === (responseJsonMaybe >=> listToMaybe >=> profileHandle)
```

(not tested)

fisx (approved):
It would have been good to have two PRs: one with the changes to the linting machinery, and one with the mechanical output of `make hlint-inplace-all`.  But since one is so tiny it wasn't a big deal to review it in one gulp.

fisx:
let's hope this settles the topic of hlint once and for all!  :)

services/brig/test/integration/API/User/Handles.hs
elland: I'm now surprised the linter didn't suggest this itself… I'll test it in a bit.

PR #2725, opened 2022-09-23
Update nginz and cannon ACLs to match api-versioned paths

pcapriotti:
This PR updates the ACL files to include api-versioned paths (paths prefixed by `v2/`, etc).

This PR makes use of the new regex syntax in libzauth that was introduced in https://github.com/wireapp/wire-server/pull/2714

## Checklist

### DIFF
--- a/charts/cannon/conf/static/zauth.acl
+++ b/charts/cannon/conf/static/zauth.acl
@@ -1,13 +1,1 @@
-a (blacklist (path "(/v[0-9]+)?/provider(/.*)?")
-             (path "(/v[0-9]+)?/bot(/.*)?")
-             (path "(/v[0-9]+)?/i/.*"))
-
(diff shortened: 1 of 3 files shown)


charts/cannon/conf/static/zauth.acl
```diff
@@ -1,17 +1 @@
...
-
-p (whitelist (path "/provider")
-             (path "/provider/**"))
-
-# LegalHold Access Tokens
-la (whitelist (path "/notifications")
-              (path "/assets/v3/**")
-              (path "/users")
-              (path "/users/**"))
+a (whitelist (regex "/await"))
```
smatting, line 1: Since cannon only have the `/await` route there is no need to include other rules

smatting (approved):

PR #2739, opened 2022-09-29
json schemas for event notifications from haskell types

fisx:
https://wearezeta.atlassian.net/browse/FS-1008

our swagger2 docs do not contain event types, since they are contained in an untyped json body in a the notification payload.  this PR partially fills this gap by creating the event type schemas under a new public end-point.  it falls short of creating schemas that are correct and complete, but it contains a list of tasks that would get us there [in the code](https://github.com/wireapp/wire-server/blob/9a050941e2a492ff93d25a61f184eade485fdb89/services/brig/src/Brig/API/Public/Swagger.hs#L48-L72).

```
$ curl http://localhost:8082/api/event-notification-schemas | jq . | tee /tmp/schema.json
```

[schema.json.zip](https://github.com/wireapp/wire-server/files/9864165/schema.json.zip)

## Checklist

### DIFF
--- a/libs/wire-api/src/Wire/API/Event/Team.hs
+++ b/libs/wire-api/src/Wire/API/Event/Team.hs
@@ -80,7 +80,7 @@ instance ToSchema Event where
         <*> _eventData .= fieldWithDocModifier "data" (description ?~ dataFieldDesc) schema
         <* eventType .= field "version" schema
     where
-      dataFieldDesc = "TODO: this part of the docs is lying; we're working on it!"
+      dataFieldDesc = "FUTUREWORK: this part of the docs is lying; we're working on it!"
 
 instance S.ToSchema Event where
   declareNamedSchema = schemaToSwagger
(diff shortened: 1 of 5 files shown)


libs/wire-api/src/Wire/API/Event/Team.hs
```diff
@@ -68,67 +71,36 @@ data Event = Event
...
+instance ToSchema Event where
+  schema =
+    object "Event" $
+      Event
+        <$> _eventTeam .= field "team" schema
+        <*> _eventTime .= field "time" utcTimeSchema
+        <*> _eventData .= fieldWithDocModifier "data" (description ?~ dataFieldDesc) schema
+        <* eventType .= field "version" schema
+    where
+      dataFieldDesc = "TODO: this part of the docs is lying; we're working on it!"
```
stephen-smith, line 83: TODO or FUTUREWORK ?

fisx:
When using the save-hook for ormolizing a module, I get the new behavior, but when I undo that and run `make formatf` (to make ormolu changes a separate, clean commit), nothing changes.

Other than that (and I suspect that's an independent issue, I've tried to show this to @elland last week but it went away) this should be "ready", in the sense explained in the PR description above.

stephen-smith (approved):

elland:
@stephen-smith Matthias won't be back for a while, feel free to merge this.

supersven:
> @stephen-smith Matthias won't be back for a while, feel free to merge this.

Hey @elland :wave: ,
I got the same information from @fisx : We can merge it when we're fine with it or adjust it if we like.

PR #2747, opened 2022-10-04
Bump servant-swagger-ui package

fisx:
## Checklist


pcapriotti (changes requested):
I think we're still supposed to use the `stack.yaml` file to generate `cabal.project.freeze`, and not change it by hand.

fisx:
thanks @pcapriotti for paying attention!  :)

cabal.project
```diff
@@ -158,12 +158,6 @@ source-repository-package
     location: https://gitlab.com/twittner/wai-routing
     tag: 7e996a93fec5901767f845a50316b3c18e51a61d
 
-source-repository-package
-    type: git
-    location: https://github.com/wireapp/tasty.git
-    tag: 394943c7672e5ad269e5587528b7678caf3b0720
-    subdir: hunit
-
```
pcapriotti, line 166: Why has this disappeared now?

fisx: just ran the script without thinking about it :)

i shall investigate!

fisx: i think this has been added to cabal at some point in the past, without having been added to stack.yaml.  is that plausible?  fix coming up!

(what's the reason we keep stack.yaml around, btw?  is there a ticket to remove it?  does anybody know?)

elland: I thought we had completely removed stack by now...

pcapriotti (approved):
LGTM.

PR #2753, opened 2022-10-06
[FS-926] Create new conversation type for global team conversation

elland:
This should cover [FS-927](https://wearezeta.atlassian.net/browse/FS-927), as the endpoint has been added.

## Checklist

### DIFF
--- a/libs/wire-api/src/Wire/API/Conversation.hs
+++ b/libs/wire-api/src/Wire/API/Conversation.hs
@@ -260,7 +260,8 @@ instance ToSchema Conversation where
       (description ?~ "A conversation object as returned from the server")
       $ Conversation
         <$> cnvQualifiedId .= field "qualified_id" schema
-        <* (qUnqualified . cnvQualifiedId) .= optional (field "id" (deprecatedSchema "qualified_id" schema))
+        <* (qUnqualified . cnvQualifiedId)
+          .= optional (field "id" (deprecatedSchema "qualified_id" schema))
         <*> cnvMetadata .= conversationMetadataObjectSchema
         <*> cnvMembers .= field "members" schema
         <*> cnvProtocol .= protocolSchema
--- a/libs/wire-api/src/Wire/API/Conversation.hs
+++ b/libs/wire-api/src/Wire/API/Conversation.hs
@@ -25,6 +25,7 @@ module Wire.API.Conversation
     ConversationMetadata (..),
     defConversationMetadata,
     Conversation (..),
+    conversationMetadataObjectSchema,
     cnvType,
     cnvCreator,
     cnvAccess,
@@ -100,7 +101,7 @@ import Data.Aeson (FromJSON (..), ToJSON (..))
 import qualified Data.Aeson as A
 import qualified Data.ByteString.Lazy as LBS
 import Data.Id
-import Data.List.Extra (disjointOrd)
+import Data.List.Extra (disjointOrd, enumerate)
 import Data.List.NonEmpty (NonEmpty)
 import Data.List1
 import Data.Misc
@@ -259,8 +260,7 @@ instance ToSchema Conversation where
       (description ?~ "A conversation object as returned from the server")
       $ Conversation
         <$> cnvQualifiedId .= field "qualified_id" schema
-        <* (qUnqualified . cnvQualifiedId)
-          .= optional (field "id" (deprecatedSchema "qualified_id" schema))
+        <* (qUnqualified . cnvQualifiedId) .= optional (field "id" (deprecatedSchema "qualified_id" schema))
         <*> cnvMetadata .= conversationMetadataObjectSchema
         <*> cnvMembers .= field "members" schema
         <*> cnvProtocol .= protocolSchema
@@ -436,6 +436,10 @@ data Access
     LinkAccess
   | -- | User can join knowing [changeable/revokable] code
     CodeAccess
+  | -- | In MLS the user can join the global team conversation with their
+    -- | clients via an external commit, thereby inviting their own clients to
+    -- | join.
+    SelfInviteAccess
   deriving stock (Eq, Ord, Bounded, Enum, Show, Generic)
   deriving (Arbitrary) via (GenericUniform Access)
   deriving (ToJSON, FromJSON, S.ToSchema) via Schema Access
@@ -448,7 +452,8 @@ instance ToSchema Access where
           [ element "private" PrivateAccess,
             element "invite" InviteAccess,
             element "link" LinkAccess,
-            element "code" CodeAccess
+            element "code" CodeAccess,
+            element "self_invite" SelfInviteAccess
           ]
 
 typeAccess :: Doc.DataType
@@ -496,6 +501,7 @@ defRole = activatedAccessRole
 
 maybeRole :: ConvType -> Maybe (Set AccessRoleV2) -> Set AccessRoleV2
 maybeRole SelfConv _ = privateAccessRole
+maybeRole GlobalTeamConv _ = teamAccessRole
 maybeRole ConnectConv _ = privateAccessRole
 maybeRole One2OneConv _ = privateAccessRole
 maybeRole RegularConv Nothing = defRole
@@ -578,7 +584,8 @@ data ConvType
   | SelfConv
   | One2OneConv
   | ConnectConv
-  deriving stock (Eq, Show, Generic)
+  | GlobalTeamConv
+  deriving stock (Eq, Show, Generic, Enum, Bounded)
   deriving (Arbitrary) via (GenericUniform ConvType)
   deriving (FromJSON, ToJSON, S.ToSchema) via Schema ConvType
 
@@ -589,11 +596,12 @@ instance ToSchema ConvType where
         [ element 0 RegularConv,
           element 1 SelfConv,
           element 2 One2OneConv,
-          element 3 ConnectConv
+          element 3 ConnectConv,
+          element 4 GlobalTeamConv
         ]
 
 typeConversationType :: Doc.DataType
-typeConversationType = Doc.int32 $ Doc.enum [0, 1, 2, 3]
+typeConversationType = Doc.int32 $ Doc.enum $ fromIntegral . fromEnum <$> enumerate @ConvType
 
 -- | Define whether receipts should be sent in the given conversation
 --   This datatype is defined as an int32 but the Backend does not
--- a/libs/wire-api/src/Wire/API/MLS/GlobalTeamConversation.hs
+++ b/libs/wire-api/src/Wire/API/MLS/GlobalTeamConversation.hs
@@ -24,7 +24,7 @@ import Data.Qualified
 import Data.Schema
 import qualified Data.Swagger as S
 import Imports
-import Wire.API.Conversation
+import Wire.API.Conversation hiding (Conversation)
 import Wire.API.Conversation.Protocol
 import Wire.Arbitrary (Arbitrary (..), GenericUniform (..))
 
--- a/libs/wire-api/src/Wire/API/MLS/GlobalTeamConversation.hs
+++ b/libs/wire-api/src/Wire/API/MLS/GlobalTeamConversation.hs
@@ -1,0 +1,54 @@
+-- This file is part of the Wire Server implementation.
+--
+-- Copyright (C) 2022 Wire Swiss GmbH <opensource@wire.com>
+--
+-- This program is free software: you can redistribute it and/or modify it under
+-- the terms of the GNU Affero General Public License as published by the Free
+-- Software Foundation, either version 3 of the License, or (at your option) any
+-- later version.
+--
+-- This program is distributed in the hope that it will be useful, but WITHOUT
+-- ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+-- FOR A PARTICULAR PURPOSE. See the GNU Affero General Public License for more
+-- details.
+--
+-- You should have received a copy of the GNU Affero General Public License along
+-- with this program. If not, see <https://www.gnu.org/licenses/>.
+
+module Wire.API.MLS.GlobalTeamConversation where
+
+import Control.Lens ((?~))
+import Data.Aeson (FromJSON, ToJSON)
+import Data.Id
+import Data.Qualified
+import Data.Schema
+import qualified Data.Swagger as S
+import Imports
+import Wire.API.Conversation
+import Wire.API.Conversation.Protocol
+import Wire.Arbitrary (Arbitrary (..), GenericUniform (..))
+
+-- | Public-facing global team conversation.
+-- Membership is implicit. Every member of a team is part of it.
+-- Protocol is also implicit: it's always MLS.
+data GlobalTeamConversation = GlobalTeamConversation
+  { gtcId :: Qualified ConvId,
+    gtcMetadata :: ConversationMetadata,
+    gtcMlsMetadata :: ConversationMLSData
+  }
+  deriving stock (Eq, Show, Generic)
+  deriving (Arbitrary) via (GenericUniform GlobalTeamConversation)
+  deriving (FromJSON, ToJSON, S.ToSchema) via Schema GlobalTeamConversation
+
+instance ToSchema GlobalTeamConversation where
+  schema =
+    objectWithDocModifier
+      "GlobalTeamConversation"
+      (description ?~ "The global team conversation object as returned from the server")
+      $ GlobalTeamConversation
+        <$> gtcId
+          .= field "id" schema
+        <*> gtcMetadata
+          .= conversationMetadataObjectSchema
+        <*> gtcMlsMetadata
+          .= mlsDataSchema
--- a/libs/wire-api/src/Wire/API/MLS/GlobalTeamConversation.hs
+++ b/libs/wire-api/src/Wire/API/MLS/GlobalTeamConversation.hs
@@ -33,8 +33,11 @@ import Wire.Arbitrary (Arbitrary (..), GenericUniform (..))
 -- Protocol is also implicit: it's always MLS.
 data GlobalTeamConversation = GlobalTeamConversation
   { gtcId :: Qualified ConvId,
-    gtcMetadata :: GlobalTeamConversationMetadata,
-    gtcMlsMetadata :: ConversationMLSData
+    gtcMlsMetadata :: ConversationMLSData,
+    gtcCreator :: Maybe UserId,
+    gtcAccess :: [Access],
+    gtcName :: Text,
+    gtcTeam :: TeamId
   }
   deriving stock (Eq, Show, Generic)
   deriving (Arbitrary) via (GenericUniform GlobalTeamConversation)
@@ -47,28 +50,14 @@ instance ToSchema GlobalTeamConversation where
       (description ?~ "The global team conversation object as returned from the server")
       $ GlobalTeamConversation
         <$> gtcId .= field "qualified_id" schema
-        <*> gtcMetadata .= gtcMetadataSchema
         <*> gtcMlsMetadata .= mlsDataSchema
-
-data GlobalTeamConversationMetadata = GlobalTeamConversationMetadata
-  { gtcmCreator :: Maybe UserId,
-    gtcmAccess :: [Access],
-    gtcmName :: Text,
-    gtcmTeam :: TeamId
-  }
-  deriving stock (Eq, Show, Generic)
-  deriving (Arbitrary) via (GenericUniform GlobalTeamConversationMetadata)
-
-gtcMetadataSchema :: ObjectSchema SwaggerDoc GlobalTeamConversationMetadata
-gtcMetadataSchema =
-  GlobalTeamConversationMetadata
-    <$> gtcmCreator
-      .= maybe_
-        ( optFieldWithDocModifier
-            "creator"
-            (description ?~ "The creator's user ID")
-            schema
-        )
-    <*> gtcmAccess .= field "access" (array schema)
-    <*> gtcmName .= field "name" schema
-    <*> gtcmTeam .= field "team" schema
+        <*> gtcCreator
+          .= maybe_
+            ( optFieldWithDocModifier
+                "creator"
+                (description ?~ "The creator's user ID")
+                schema
+            )
+        <*> gtcAccess .= field "access" (array schema)
+        <*> gtcName .= field "name" schema
+        <*> gtcTeam .= field "team" schema
--- a/libs/wire-api/src/Wire/API/MLS/GlobalTeamConversation.hs
+++ b/libs/wire-api/src/Wire/API/MLS/GlobalTeamConversation.hs
@@ -33,7 +33,7 @@ import Wire.Arbitrary (Arbitrary (..), GenericUniform (..))
 -- Protocol is also implicit: it's always MLS.
 data GlobalTeamConversation = GlobalTeamConversation
   { gtcId :: Qualified ConvId,
-    gtcMetadata :: ConversationMetadata,
+    gtcMetadata :: GlobalTeamConversationMetadata,
     gtcMlsMetadata :: ConversationMLSData
   }
   deriving stock (Eq, Show, Generic)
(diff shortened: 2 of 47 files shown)


services/galley/src/Galley/API/Teams.hs
```diff
@@ -241,6 +243,7 @@ createNonBindingTeamH zusr zcon (Public.NonBindingNewTeam body) = do
   P.debug $
     Log.field "targets" (toByteString . show $ toByteString <$> zothers)
       . Log.field "action" (Log.val "Teams.createNonBindingTeam")
+  -- TODO: elland: create global team conv from here as well?
```
elland, line 246:
```suggestion
```

services/galley/src/Galley/Cassandra/Access.hs
```diff
@@ -31,6 +31,7 @@ defAccess SelfConv (Just (Set [])) = [PrivateAccess]
 defAccess ConnectConv (Just (Set [])) = [PrivateAccess]
 defAccess One2OneConv (Just (Set [])) = [PrivateAccess]
 defAccess RegularConv (Just (Set [])) = defRegularConvAccess
+defAccess GlobalTeamConv _ = []
```
elland, line 34: Double check this with team.

services/galley/src/Galley/API/Internal.hs
```diff
@@ -685,28 +685,31 @@ rmUser lusr conn = do
       let qUser = qUntagged lusr
       cc <- getConversations ids
       now <- input
+      let deleteIfNeeded c =
```
elland, line 688: Double check this with team.

libs/wire-api/src/Wire/API/Conversation.hs
```diff
@@ -491,6 +512,7 @@ defRole = activatedAccessRole
 
 maybeRole :: ConvType -> Maybe (Set AccessRoleV2) -> Set AccessRoleV2
 maybeRole SelfConv _ = privateAccessRole
+maybeRole GlobalTeamConv _ = teamAccessRole -- TODO(elland): verify this with team
```
elland, line 515: @pcapriotti can you confirm this is the right access role?

services/galley/src/Galley/Cassandra/Conversation.hs
```diff
@@ -191,6 +191,32 @@ getConversation conv = do
...
+createGlobalTeamConversation tid uid = do
+  createConversation (globalTeamConv <$> tid) nc
+  where
+    nc =
+      NewConversation
+        { ncMetadata =
+            ConversationMetadata
+              { cnvmType = GlobalTeamConv,
+                cnvmCreator = uid,
+                cnvmAccess = [LinkAccess],
```
mdimjasevic, line 209: I think this should be `cnvmAccess = [InviteAccess]`, as users should be able to join only by someone else adding them (namely, the team admin). `LinkAccess` means it is sufficient to know the conversation ID.

services/galley/src/Galley/Cassandra/Conversation.hs
```diff
@@ -191,6 +191,32 @@ getConversation conv = do
...
+                cnvmAccessRoles = mempty,
```
mdimjasevic, line 210:
```suggestion
                cnvmAccessRoles = teamAccessRole,
```
The conversation should be accessible only to team members.

services/galley/src/Galley/Cassandra/Conversation.hs
```diff
@@ -191,6 +191,32 @@ getConversation conv = do
...
+                cnvmName = Nothing,
```
mdimjasevic, line 211:
```suggestion
                cnvmName = Just "Global team conversation",
```

services/galley/src/Galley/Cassandra/Access.hs
mdimjasevic, line 34:
```suggestion
defAccess GlobalTeamConv s = maybe [InviteAccess] Set.toList s
```
This seems to be in line with the last catch-all clause for a set with at least one element. If the input set is `Nothing`, team members should be able to join via an invitation.

services/galley/test/integration/API/MLS.hs
```diff
@@ -2011,3 +2054,41 @@ testRemoteUserPostsCommitBundle = do
...
+
+getGlobalConvUserNotMemberOfTeam :: TestM ()
+getGlobalConvUserNotMemberOfTeam = do
+  uid <- randomUser
+  owner <- randomUser
+  tid <- createBindingTeamInternal "sample-team" owner
+  team <- getTeam owner tid
+  assertQueue "create team" tActivate
+  liftIO $ assertEqual "owner" owner (team ^. teamCreator)
+  assertQueueEmpty
```
mdimjasevic, line 2072: I don't understand what queues we're dealing with here. Is that necessary?

services/galley/test/integration/API/MLS.hs
```diff
@@ -2011,3 +2054,41 @@ testRemoteUserPostsCommitBundle = do
...
+
+-- Create global team conversation
+getGlobalConvPresent :: TestM ()
+getGlobalConvPresent = do
+  owner <- randomUser
+  tid <- createBindingTeamInternal "sample-team" owner
+  team <- getTeam owner tid
+  assertQueue "create team" tActivate
+  liftIO $ assertEqual "owner" owner (team ^. teamCreator)
+  assertQueueEmpty
```
mdimjasevic, line 2083: ditto

mdimjasevic (commented):
Looks good! I'd ask you to change a couple of things as I suggested inlined. Once you address that I think this is good to merge.

Also, I didn't see a point where you need a creator client ID. I'm puzzled. This should have shown up somewhere.

services/galley/src/Galley/Cassandra/Conversation.hs
elland: Yeah, I thought you would join by `convId`, by querying the `GET` endpoint for global team convs. This is sadly not specified in the spec either.

libs/wire-api/src/Wire/API/Conversation.hs
```diff
@@ -175,37 +176,68 @@ accessRolesSchemaOpt = toOutput .= accessRolesSchemaTuple `withParser` validate
...
+    <*> cnvmReceiptMode
+    .= optField "receipt_mode" (maybeWithDefault A.Null schema)
 
 instance ToSchema ConversationMetadata where
   schema = object "ConversationMetadata" conversationMetadataObjectSchema
 
+-- | Public-facing global team conversation.
+-- Membership is implicit. Every member of a team is part of it.
+-- Protocol is also implicit: it's always MLS.
+data GlobalTeamConversation = GlobalTeamConversation
```
mdimjasevic, line 222: Now that I think about it, if we're to stay consistent with how we've implemented MLS so far, this should go somewhere under `Wire.API.MLS`, perhaps in a new module.

services/galley/src/Galley/API/Teams.hs
```diff
@@ -250,18 +253,28 @@ createNonBindingTeamH zusr zcon (Public.NonBindingNewTeam body) = do
       (body ^. newTeamIconKey)
       NonBinding
   finishCreateTeam team owner others (Just zcon)
-  pure (team ^. teamId)
+  let tid = team ^. teamId
+  void $ E.createGlobalTeamConversation (qualifyAs lusr tid) zusr
```
mdimjasevic, line 257: After creating the global team conversation, you have to add the creator client `c` to the group with something like this:

```haskell
  E.addMLSClients lcnv (qUntagged lusr) (Set.singleton (c, nullKeyPackageRef))
```
The same goes for the binding team.

services/galley/src/Galley/Cassandra/Conversation.hs
mdimjasevic: Then the list is `[InviteAccess, LinkAccess]`, right?

services/galley/src/Galley/Cassandra/Access.hs
elland: But you shouldn't be invited to the global team conv, right? It should be done automatically when you join a team?

mdimjasevic (dismissed):
I can't see any changes to the Message module in Galley, hence it can't be this PR is ready. There in commit processing you'd need to add the creator client in the first epoch. Paolo has been refactoring that module and introducing a change like this, which I need too in the self-conversation PR.

libs/wire-api/src/Wire/API/MLS/GlobalTeamConversation.hs
```diff
@@ -0,0 +1,54 @@
...
+  deriving (FromJSON, ToJSON, S.ToSchema) via Schema GlobalTeamConversation
+
+instance ToSchema GlobalTeamConversation where
+  schema =
+    objectWithDocModifier
+      "GlobalTeamConversation"
+      (description ?~ "The global team conversation object as returned from the server")
+      $ GlobalTeamConversation
+        <$> gtcId
+          .= field "id" schema
```
pcapriotti, line 50: Better to use the same label and format for the ID as for normal conversations.

services/galley/src/Galley/API/Federation.hs
```diff
@@ -279,7 +279,7 @@ onConversationUpdated requestingDomain cu = do
       SConversationMessageTimerUpdateTag -> pure (Just sca, [])
       SConversationReceiptModeUpdateTag -> pure (Just sca, [])
       SConversationAccessDataTag -> pure (Just sca, [])
-
+      SConversationSelfInviteTag -> pure (Nothing, []) -- TODO(elland): Should not happen. Should we throw?
```
elland, line 282: For now it might make more sense to throw a 403 instead of an empty list.

services/galley/src/Galley/API/Internal.hs
```diff
@@ -691,28 +691,30 @@ rmUser lusr conn = do
...
+             in pure $
+                  Intra.newPushLocal ListComplete (tUnqualified lusr) (Intra.ConvEvent e) (Intra.recipient <$> Data.convLocalMembers c)
+                    <&> set Intra.pushConn conn
+                      . set Intra.pushRoute Intra.RouteDirect
+      pp <- for cc $ \c -> case Data.convType c of
+        SelfConv -> pure Nothing
+        One2OneConv -> deleteMembers (Data.convId c) (UserList [tUnqualified lusr] []) $> Nothing
+        ConnectConv -> deleteMembers (Data.convId c) (UserList [tUnqualified lusr] []) $> Nothing
+        RegularConv -> deleteIfNeeded c >> fireEvent c
+        GlobalTeamConv -> deleteIfNeeded c >> pure Nothing
```
elland, line 717: Actually we need a different logic here. For the global team conv we need to do the MLS client removals regardless of membership, since that's implicit.

services/galley/src/Galley/API/Query.hs
```diff
@@ -147,6 +154,47 @@ getUnqualifiedConversation lusr cnv = do
...
+  E.getGlobalTeamConversation tid >>= \case
+    Nothing -> do
+      conv <- E.createGlobalTeamConversation (qualifyAs lusr tid) uid
+      mlsData <- case Conv.convProtocol conv of
+        ProtocolMLS mls -> pure mls
+        ProtocolProteus -> throw (InternalErrorWithDescription "Wrong protocol, expected MLS, got Proteus.")
+      -- FUTUREWORK: remove this. we are planning to remove the need for a nullKeyPackageRef
+      let convId = Conv.convId conv
+          lconv = qualifyAs lusr convId
+      E.addMLSClients lconv (qUntagged lusr) (Set.singleton (cid, nullKeyPackageRef))
```
elland, line 182: Don't actually create the MLS clients here. Instead add logic in the commit handling code to deal with the global team conversation special case. (Wait for refactoring to be merged, #2844 )

services/galley/src/Galley/API/Teams.hs
```diff
@@ -254,15 +258,24 @@ createNonBindingTeamH zusr zcon (Public.NonBindingNewTeam body) = do
       (body ^. newTeamIconKey)
       NonBinding
   finishCreateTeam team owner others (Just zcon)
-  pure (team ^. teamId)
+  let tid = team ^. teamId
```
elland, line 261: Forgot to clear this after debug. Weird that hlint didn't complain.

services/galley/src/Galley/Cassandra/Conversation.hs
```diff
@@ -191,6 +191,61 @@ getConversation conv = do
       <*> UnliftIO.wait cdata
   runMaybeT $ conversationGC =<< maybe mzero pure mbConv
 
+getGlobalTeamConversation :: TeamId -> Client (Maybe Conversation)
+getGlobalTeamConversation tid =
+  (\c -> c {convLocalMembers = [], convRemoteMembers = []})
+    <$$> getConversation (globalTeamConv tid)
```
elland, line 197: Rewrite this to not actually query the member list.

services/galley/src/Galley/Cassandra/Conversation.hs
```diff
@@ -191,6 +191,61 @@ getConversation conv = do
...
+            cnvmAccessRoles = mempty,
+            cnvmName = Just "Global team conversation",
+            cnvmTeam = Just (tUnqualified tid),
+            cnvmMessageTimer = Nothing,
+            cnvmReceiptMode = Nothing
+          }
+      gid = convToGroupId lconv
+      cs = MLS_128_DHKEMX25519_AES128GCM_SHA256_Ed25519
+      proto =
+        ProtocolMLS
```
elland, line 219: If we use GlobalTeamConversation directly we can simplify this.

services/galley/test/integration/API.hs
```diff
@@ -3837,7 +3837,8 @@ testOne2OneConversationRequest shouldBeLocal actor desired = do
     let req = UpsertOne2OneConversationRequest alice bob actor desired Nothing
     res <-
       iUpsertOne2OneConversation req
-        <!! statusCode === const 200
+        <!! statusCode
```
elland, line 3840: Missed reverting this line from ormolu mess.

services/galley/test/integration/API/MLS.hs
```diff
@@ -185,6 +192,12 @@ tests s =
           test s "add user with a commit bundle to a remote conversation" testAddUserToRemoteConvWithBundle,
           test s "remote user posts commit bundle" testRemoteUserPostsCommitBundle
         ],
+      testGroup
+        "GlobalTeamConv"
+        [ test s "Non-existing team returns 403" testGetGlobalTeamConvNonExistant,
+          test s "Non member of team returns 403" testGetGlobalTeamConvNonMember,
+          test s "Global team conversation is created on get if not present" (testGetGlobalTeamConv s)
```
elland, line 199: Add test for self-invite, MLS stuff.

pcapriotti (changes requested):
There needs to be some special logic either when listing conversations (to exclude the global team and self conversations) or when returning them (to convert them to the correct JSON format).

services/galley/src/Galley/API/Federation.hs
elland: Actually this is federation, we don't throw errors.

services/galley/src/Galley/API/Action.hs
```diff
@@ -282,7 +283,7 @@ ensureAllowed tag loc action conv origUser = do
         (convType conv == GlobalTeamConv)
         $ throwS @'InvalidOperation
     SConversationLeaveTag ->
-      when (convType conv == GlobalTeamConv) $
+      when (convType conv == GlobalTeamConv) $ do
```
pcapriotti, line 286: Why the extra `do`?

services/galley/src/Galley/API/Action.hs
```diff
@@ -634,7 +666,26 @@ updateLocalConversationUnchecked lconv qusr con action = do
       conv = tUnqualified lconv
 
   -- retrieve member
-  self <- noteS @'ConvNotFound $ getConvMember lconv conv qusr
+  self <-
+    if (cnvmType . convMetadata . tUnqualified $ lconv) == GlobalTeamConv
+      then
+        -- TODO(elland): address this problem
+        pure . Left $
```
pcapriotti, line 673: (*Referring to this whole function*)

This is not so terrible, but the GTC querying bits could be moved to a different function to cleanly separate the special logic for GTCs from the rest of the logic.

Of course, the "clean" haskelly solution would be to make `Conversation` a sum type encompassing the two kinds of conversations (i.e. regular or GTC). Given the unfriendliness of record types in Haskell, I understand you might have some reservations about going this route, but maybe it's worth considering.

services/galley/src/Galley/API/MLS/Message.hs
```diff
@@ -499,6 +507,28 @@ postMLSMessageToLocalConv qusr senderClient con smsg lcnv = case rmValue smsg of
     propagateMessage qusr lconv cm con (rmRaw smsg)
 
     pure events
+  where
+    gtcToConv :: GlobalTeamConversation -> Data.Conversation
```
pcapriotti: The `undefined` here is not good, just a horrible runtime bug waiting to happen. Also, this function could be extracted and reused, since it (or a version of it) is defined at least twice. 

If you decide not to go with the sum type solution for the `Conversation` type, then I think the only way would be to make the creator field optional, and adapt the code everywhere else.

services/galley/src/Galley/API/Query.hs
```diff
@@ -171,31 +168,16 @@ getGlobalTeamConversation ::
...
-      mlsData <- case Conv.convProtocol conv of
-        ProtocolMLS mls -> pure mls
-        ProtocolProteus -> throw (InternalErrorWithDescription "Wrong protocol, expected MLS, got Proteus.")
-      pure $
-        GlobalTeamConversation
-          (qUntagged . qualifyAs lusr $ Conv.convId conv)
-          (Conv.convMetadata conv)
-          mlsData
+      gtc <- E.createGlobalTeamConversation ltid uid
+      E.addMLSClients (localGtcId gtc) (qUntagged lusr) (Set.fromList [(cid, nullKeyPackageRef)])
```
pcapriotti: You're still setting the client on on GET, which is prone to race conditions. The `nullKeyPackageRef` hack should not be used anymore.

services/galley/src/Galley/Cassandra/Conversation.hs
```diff
@@ -249,60 +251,111 @@ getConversation conv = do
       <*> UnliftIO.wait cdata
   runMaybeT $ conversationGC =<< maybe mzero pure mbConv
 
-getGlobalTeamConversation :: TeamId -> Client (Maybe Conversation)
-getGlobalTeamConversation tid =
-  (\c -> c {convLocalMembers = [], convRemoteMembers = []})
-    <$$> getConversation (globalTeamConv tid)
+getGlobalTeamConversation ::
```
pcapriotti, line 255: Why the duplication? Can't `getGlobalTeamConversation` simply call `getGlobalTeamConversationById`?

pcapriotti (changes requested):
There are a few issues, see comments below.

I can't find the code that produces the correct rendering of GTCs when fetched through the normal `POST /conversations/list` API. There is also no test for listing.

And one more thing that seems to be missing is creating the GTC entry on the `list-ids` endpoint, so that clients don't need to rely on the mutating semantics of the GET endpoint, just like what we discussed for the self conversation.

services/galley/src/Galley/API/Action.hs
elland: Probably a remnant!

services/galley/src/Galley/API/Query.hs
elland: I think this was a rebase mix-up. Good catch.

elland: Oh, more like I missed a spot.

services/galley/src/Galley/API/Action.hs
elland: > make Conversation a sum type 

I'm all for it but this PR is large enough as it is. Should we create a follow up?

services/galley/src/Galley/API/MLS/Message.hs
```diff
@@ -475,7 +476,15 @@ postMLSMessageToLocalConv ::
...
-    conv <- getLocalConvForUser qusr lcnv
+    gtc <- getGlobalTeamConversationById lcnv
+    conv <- case gtc of
+      Just conv -> do
+        when (isNothing (gtcCreator conv)) $ do
+          setGlobalTeamConversationCreator conv (qUnqualified qusr)
+        localMembers <- getLocalMembers (qUnqualified . gtcId $ conv)
+        pure $ gtcToConv conv localMembers
+      Nothing ->
+        getLocalConvForUser qusr lcnv
```
pcapriotti, line 487: Wouldn't it be better to get the conversation first, then check if it's the GTC looking at its type? That way the normal case of conversations doesn't have to go through yet another db hit.

services/galley/src/Galley/API/MLS/Message.hs
```diff
@@ -475,7 +476,15 @@ postMLSMessageToLocalConv ::
   Sem r [LocalConversationUpdate]
 postMLSMessageToLocalConv qusr senderClient con smsg lcnv = case rmValue smsg of
   SomeMessage tag msg -> do
-    conv <- getLocalConvForUser qusr lcnv
+    gtc <- getGlobalTeamConversationById lcnv
+    conv <- case gtc of
+      Just conv -> do
+        when (isNothing (gtcCreator conv)) $ do
+          setGlobalTeamConversationCreator conv (qUnqualified qusr)
+        localMembers <- getLocalMembers (qUnqualified . gtcId $ conv)
```
pcapriotti, line 484: Won't this return the empty list?

services/galley/src/Galley/API/MLS/Util.hs
```diff
@@ -50,10 +77,29 @@ getLocalConvForUser ::
   Local ConvId ->
   Sem r Data.Conversation
 getLocalConvForUser qusr lcnv = do
-  conv <- getConversation (tUnqualified lcnv) >>= noteS @'ConvNotFound
+  gtc <- getGlobalTeamConversationById lcnv
```
pcapriotti, line 80: This is a bit strange. Can't we simply accept that a GTC does not have a creator? All this fudging around with the creator should not be necessary.

pcapriotti (approved):
Looks generally good, but I would consider changing the way the creator of a GTC is handled. Since a GTC doesn't really have a creator, I would just try to make the code reflect that, instead of forcing one. That means changing the types around a bit, but it shouldn't be a huge effort.

I'll approve for now and let you decide if it's worth doing that or not.

services/galley/src/Galley/API/MLS/Message.hs
elland: Actually this code is now outdated, since I've changed how getting members work further down the chain, we can basically revert this.

services/galley/src/Galley/API/MLS/Message.hs
elland: It now returns the same results as they member list already in the conversation. Dead code.

services/galley/src/Galley/API/MLS/Util.hs
elland: This has large ramifications… I think it makes sense to clean it up when we do the conversation sum type.

services/galley/test/integration/API/Util.hs
```diff
@@ -275,9 +278,17 @@ createBindingTeamInternalNoActivate name owner = do
...
+      ( g
+          . paths ["/i/teams", toByteString' tid]
+          . zUser owner
+          . zConn "conn"
+          . zType "access"
+          . json nt
+      )
+      <!! do
+        const 201 === statusCode
+        const True === isJust . getHeader "Location"
```
jschaul, line 291: why all this reformatting as part of this PR? Makes it much harder to see actual changes that happened.

services/galley/src/Galley/API/MLS/Util.hs
jschaul, line 80: I don't like the sideeffects introduced here into each lookup of a conversation. This looks badly designed.

Elsewhere, it seems global team conversation id is the same as the team Id (even though this is not documented anywhere; and it means there can only ever be one single global team conversation). Can't this mean you can tell earlier, if you have a user object available, you'd know their team, and already know which kind of conv you're dealing with, without an extra lookup?

And why does a GET of any conversation ID lead to an insert of a random user as creator of this conversation? That makes no sense semantically.

elland: We're taking this out in a follow-up  👍

services/galley/test/integration/API/Util.hs
elland: We had some issues with ormolu, I might have missed some places.

PR #2758, opened 2022-10-07
Fix bug: clients without prekeys are not deleted competely

stefanwire:
https://wearezeta.atlassian.net/browse/FS-966

This PR changes the behaviour when clients without prekeys are detected during prekey claiming.

Without this PR: a client without any prekeys is deleted from other services synchronously, but _not_ from brig.
With this PR: a client without any prekeys is deleted
1) synchronously from brig: This leads to the client not be listed anymore. Without this change the client's prekeys are attempted to be fetched over and over again also triggering the deletion attempts over and over again.
2) asynchronously from all other services: This should improve response times when claming prekeys. The removal of clients from the galley service can be particularly slow for MLS clients. with this change it won't block prekey claiming

## Checklist

### DIFF
--- a/changelog.d/3-bug-fixes/pr-2758
+++ b/changelog.d/3-bug-fixes/pr-2758
@@ -1,0 +1,1 @@
+Clients without any prekeys are not deleted completely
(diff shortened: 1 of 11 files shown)


pcapriotti (approved):
Looks good, but unfortunately a lot of polysemy-related changes are mixed in, making it difficult to see what exactly has changed.

PR #2764, opened 2022-10-12
Add client ID to tokens

pcapriotti:
## Checklist


services/brig/test/integration/API/User/Auth.hs
```diff
@@ -968,6 +974,157 @@ getAndTestDBSupersededCookieAndItsValidSuccessor config b n = do
...
+        <!! const 201 === statusCode
+    r <-
+      post
+        ( unversioned
+            . brig
+            . path "/access"
+            . queryItem "client_id" (toByteString' (clientId cl2))
+            . cookie c2
+        )
+        <!! const 200 === statusCode
```
mdimjasevic, line 1118: I suppose this is fine, but would it make more sense to return an error code in this case, i.e., when the provided client ID is different from what was initially provided?

services/nginz/third_party/nginx-zauth-module/README.md
```diff
@@ -31,15 +31,16 @@ the vector of keys in declaration order.
         zauth_key "secret3";
     }
 
-### $zauth_user, $zauth_connection
+### $zauth_user, $zauth_client, $zauth_connection
 
-These variables are replaced with the actual user/connection IDs of an
+These variables are replaced with the actual user/client/connection IDs of an
 access-token and set in upstream requests as headers "Z-User"/"Z-Connection".
```
mdimjasevic, line 37:
```suggestion
access-token and set in upstream requests as headers "Z-User"/"Z-Client"/"Z-Connection".
```

mdimjasevic (approved):
A nice improvement!

One part I wasn't sure of is your flattening of one function that you sprinkled with `MaybeT`s, but I believe you got it right.

services/brig/test/integration/API/User/Auth.hs
pcapriotti: I guess so, I was hesitant about it because of the awkward monad stack that is needed to get that. This is because `MonadLogger` does not have an instance for `ExceptT`. But it's not the end of the world.

services/nginz/third_party/nginx-zauth-module/README.md
pcapriotti: Thanks, fixed

PR #2765, opened 2022-10-13
[FS-919] Support Basic Processing of External Commits

mdimjasevic:
The PR introduces basic support for external commits in MLS. Note that this does not resubmit external pending proposals submitted by the backend; this is to be done in https://wearezeta.atlassian.net/browse/FS-920.

Tracked by https://wearezeta.atlassian.net/browse/FS-919.

## Checklist

### DIFF
--- a/libs/wire-api/src/Wire/API/Error/Galley.hs
+++ b/libs/wire-api/src/Wire/API/Error/Galley.hs
@@ -83,9 +83,6 @@ data GalleyError
   | MLSClientSenderUserMismatch
   | MLSWelcomeMismatch
   | MLSMissingGroupInfo
-  | MLSMissingExternalInit
-  | MLSRemovalUserMismatch
-  | MLSExternalCommitRemoveProposal
   | --
     NoBindingTeamMembers
   | NoBindingTeam
@@ -208,12 +205,6 @@ type instance MapError 'MLSClientSenderUserMismatch = 'StaticError 400 "mls-clie
 type instance MapError 'MLSWelcomeMismatch = 'StaticError 400 "mls-welcome-mismatch" "The list of targets of a welcome message does not match the list of new clients in a group"
 
 type instance MapError 'MLSMissingGroupInfo = 'StaticError 404 "mls-missing-group-info" "The conversation has no group information"
-
-type instance MapError 'MLSMissingExternalInit = 'StaticError 400 "mls-missing-external-init" "The external commit is missing an external init proposal"
-
-type instance MapError 'MLSRemovalUserMismatch = 'StaticError 403 "mls-removal-user-mismatch" "The external commit attempts to remove a client from a user other than themselves"
-
-type instance MapError 'MLSExternalCommitRemoveProposal = 'StaticError 403 "mls-external-commit-proposals" "The external commit has to have at most one remove proposal"
 
 type instance MapError 'NoBindingTeamMembers = 'StaticError 403 "non-binding-team-members" "Both users must be members of the same binding team"
 
--- a/libs/wire-api/src/Wire/API/Error/Galley.hs
+++ b/libs/wire-api/src/Wire/API/Error/Galley.hs
@@ -83,6 +83,9 @@ data GalleyError
   | MLSClientSenderUserMismatch
   | MLSWelcomeMismatch
   | MLSMissingGroupInfo
+  | MLSMissingExternalInit
+  | MLSRemovalUserMismatch
+  | MLSExternalCommitRemoveProposal
   | --
     NoBindingTeamMembers
   | NoBindingTeam
@@ -205,6 +208,12 @@ type instance MapError 'MLSClientSenderUserMismatch = 'StaticError 400 "mls-clie
 type instance MapError 'MLSWelcomeMismatch = 'StaticError 400 "mls-welcome-mismatch" "The list of targets of a welcome message does not match the list of new clients in a group"
 
 type instance MapError 'MLSMissingGroupInfo = 'StaticError 404 "mls-missing-group-info" "The conversation has no group information"
+
+type instance MapError 'MLSMissingExternalInit = 'StaticError 400 "mls-missing-external-init" "The external commit is missing an external init proposal"
+
+type instance MapError 'MLSRemovalUserMismatch = 'StaticError 403 "mls-removal-user-mismatch" "The external commit attempts to remove a client from a user other than themselves"
+
+type instance MapError 'MLSExternalCommitRemoveProposal = 'StaticError 403 "mls-external-commit-proposals" "The external commit has to have at most one remove proposal"
 
 type instance MapError 'NoBindingTeamMembers = 'StaticError 403 "non-binding-team-members" "Both users must be members of the same binding team"
 
--- a/services/galley/src/Galley/API/MLS/Message.hs
+++ b/services/galley/src/Galley/API/MLS/Message.hs
@@ -529,7 +529,7 @@ data ProposalAction = ProposalAction
     paRemove :: ClientMap,
     -- The backend does not process external init proposals, but still it needs
     -- to know if a commit has one when processing external commits
-    paExternalInit :: Maybe ()
+    paExternalInit :: Any
   }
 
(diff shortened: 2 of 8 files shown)


libs/wire-api/src/Wire/API/Error/Galley.hs
```diff
@@ -206,6 +209,12 @@ type instance MapError 'MLSWelcomeMismatch = 'StaticError 400 "mls-welcome-misma
 
 type instance MapError 'MLSMissingGroupInfo = 'StaticError 404 "mls-missing-group-info" "The conversation has no group information"
 
+type instance MapError 'MLSMissingExternalInit = 'StaticError 400 "mls-missing-external-init" "The external commit is missing an external init proposal"
+
+type instance MapError 'MLSRemovalUserMismatch = 'StaticError 403 "mls-removal-user-mismatch" "The external commit attempts to remove a client from a user other than themselves"
+
+type instance MapError 'MLSExternalCommitRemoveProposal = 'StaticError 403 "mls-external-commit-proposals" "The external commit has to have at most one remove proposal"
```
elland, line 216: nitpick: "must have" instead of "has to have"

services/galley/src/Galley/API/MLS/Message.hs
```diff
@@ -660,25 +700,57 @@ processCommitWithAction qusr senderClient con lconv cm epoch groupId action send
...
+          pure $ (pure (), action) -- no key package ref update necessary
         else case (sender, upLeaf <$> cPath commit) of
           (MemberSender senderRef, Just updatedKeyPackage) -> do
             updatedRef <- kpRef' updatedKeyPackage & note (mlsProtocolError "Could not compute key package ref")
             -- postpone key package ref update until other checks/processing passed
             case senderClient of
-              Just cli -> pure $ updateKeyPackageMapping lconv qusr cli (Just senderRef) updatedRef
-              Nothing -> pure $ pure ()
-          (_, Nothing) -> pure $ pure () -- ignore commits without update path
+              Just cli -> pure (updateKeyPackageMapping lconv qusr cli (Just senderRef) updatedRef, action)
```
elland, line 709: Would moving `pure` outside of the case match work?

mdimjasevic: I don't think it would. At first it looked weird to me as well. All the branches of the bigger pattern-match have to have the same type. As far as my understanding goes, this block is written this way so it can be executed later, but then I suppose the same could have been achieved if at the top we had a `let` binding instead of monadic binding and using one level of monads less.

libs/wire-api/src/Wire/API/Error/Galley.hs
mdimjasevic: I'll incorporate this only if there will be another reason to make another commit so as to avoid running CI again just for this nit-pick.

services/galley/src/Galley/API/MLS/Message.hs
elland: I'm fine either way, more of a nitpick :D

services/galley/src/Galley/API/MLS/Message.hs
```diff
@@ -660,25 +700,57 @@ processCommitWithAction qusr senderClient con lconv cm epoch groupId action send
...
+            remRef <-
+              if Map.null (paRemove action)
+                then pure Nothing
+                else do
+                  (remCid, r) <- derefUser (paRemove action) qusr
+                  unless (cidQualifiedUser cid == cidQualifiedUser remCid) $
+                    throwS @'MLSRemovalUserMismatch
+                  pure (Just r)
+
+            updateKeyPackageMapping lconv qusr (ciClient cid) remRef newRef
```
pcapriotti, line 736: Shouldn't you validate the key package first? See for example `applyProposal`.

services/galley/test/integration/API/MLS/Util.hs
```diff
@@ -257,7 +258,9 @@ mlscli :: HasCallStack => ClientIdentity -> [String] -> Maybe ByteString -> MLST
 mlscli qcid args mbstdin = do
   bd <- State.gets mlsBaseDir
   let cdir = bd </> cid2Str qcid
-  liftIO $ spawn (proc "mls-test-cli" (["--store", cdir </> "store"] <> args)) mbstdin
+  liftIO $ do
+    -- Imports.setEnv "RUST_BACKTRACE" "full"
+    spawn (proc "mls-test-cli" (["--store", cdir </> "store"] <> args)) mbstdin
```
pcapriotti, line 263: Dead code?

libs/wire-api/src/Wire/API/Error/Galley.hs
mdimjasevic: I'll drop these errors in favor of the generic `MLSProtocolError`.

libs/wire-api/src/Wire/API/Error/Galley.hs
```diff
@@ -206,6 +209,12 @@ type instance MapError 'MLSWelcomeMismatch = 'StaticError 400 "mls-welcome-misma
...
+
```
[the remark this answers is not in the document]
mdimjasevic: Different error codes was a mistake. In commit 0224a420d I replaced the specific errors with `MLSProtocolError`.

services/galley/src/Galley/API/MLS/Message.hs
```diff
@@ -526,24 +550,31 @@ type HasProposalEffects r =
 
 data ProposalAction = ProposalAction
   { paAdd :: ClientMap,
-    paRemove :: ClientMap
+    paRemove :: ClientMap,
+    -- The backend does not process external init proposals, but still it needs
+    -- to know if a commit has one when processing external commits
+    paExternalInit :: Maybe ()
```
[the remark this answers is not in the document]
mdimjasevic: Sure, done in commit 20162df16. But aren't these two types isomorphic as well, i.e., don't the same arguments you had yesterday in the call apply to `Any` as well?

services/galley/test/integration/API/MLS/Util.hs
mdimjasevic: Thanks for spotting this! Removed in commit 9e1087534.

services/galley/src/Galley/API/MLS/Message.hs
mdimjasevic: Your observation has seemed valid and with that in mind, I looked at other cases when a key package is updated or mapped and they some of them seemed equally wrong because they also just call `updateKeyPackageMapping`. For example, in the case of an existing member updating their key package via the update path we blindly accept the new key package without checking if it is for the same protocol, same cipher suite, etc., but this seems wrong, doesn't it?

I realized `updateKeyPackageMapping` does validation (the ciphersuite, identity matching, signature scheme, etc.) if the old key package is non-existent (i.e., we pass in `Nothing` to `updateKeyPackageMapping`), by calling an internal endpoint in Brig, for which there's a handler `upsertKeyPackage`.

So before making a call to `updateKeyPackageMapping` in the case of a `NewMemberSender`, I just added a call to `addKeyPackageRef`, which calls `upsertKeyPackage` in Brig and validates it before inserting into the database.

Does that seem OK?

services/galley/src/Galley/API/MLS/Message.hs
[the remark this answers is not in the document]
pcapriotti: What arguments? My only argument here is that `Maybe ()` is a circuitous way to express a boolean, while `Any` is more direct. Using `Bool` there is also fine. It's a minor thing, anyway.

services/galley/src/Galley/API/MLS/Message.hs
pcapriotti: As long as validation is happening.

pcapriotti (approved):

services/galley/src/Galley/API/MLS/Message.hs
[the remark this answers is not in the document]
mdimjasevic: Sure, I see your point.

PR #2767, opened 2022-10-15
Polysemy: intersperse combinator

isovector:
This PR implements a new polysemy combinator `intersperse` which allows for injecting logic between every handled action in the effect stack. It also adds a test-suite to `polysemy-wire-zoo`, with a new test showing how to use `intersperse` to poll from an `async` thread. 

The short-term use case for this is to use it to test brig's STOMP client, where we can `intersperse` the listener callback, which otherwise gets in the way of the removing `IO` effort.

## Checklist

### DIFF
--- a/libs/polysemy-wire-zoo/test/Test/IntersperseSpec.hs
+++ b/libs/polysemy-wire-zoo/test/Test/IntersperseSpec.hs
@@ -6,12 +6,12 @@ module Test.IntersperseSpec where
 import qualified Data.Set as S
 import Imports hiding (intersperse)
 import Polysemy
+import Polysemy.Output (output, runOutputList)
+import Polysemy.State (evalState, get, modify)
 import Polysemy.Testing
 import Polysemy.Trace
 import Test.Hspec
 import UnliftIO (async)
-import Polysemy.Output (runOutputList, output)
-import Polysemy.State (evalState, get, modify)
 
 spec :: Spec
 spec = do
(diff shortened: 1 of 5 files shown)


libs/polysemy-wire-zoo/test/Test/IntersperseSpec.hs
```diff
@@ -0,0 +1,77 @@
...
+    let result =
+          fst $
+            run $
+              runTraceList $
+                outputToTrace show $
+                  evalState @Int 0 $
+                    intersperse ((output =<< get) <* modify (+ 1)) $ do
+                      -- 0
+                      trace "start"
+                      pure ()
```
elland, line 43: I think `hlint` will complain about this if we don't add an ignore pragma. 👍

libs/polysemy-wire-zoo/test/Test/IntersperseSpec.hs
```diff
@@ -0,0 +1,77 @@
+{-# LANGUAGE NumDecimals #-}
+{-# OPTIONS_GHC -fplugin=Polysemy.Plugin #-}
+
+module Test.IntersperseSpec where
+
+import qualified Data.Set as S
+import Imports hiding (intersperse)
+import Polysemy
+import Polysemy.Output (output, runOutputList)
```
elland, line 9: Redundant import? 

```suggestion
import Polysemy.Output (output)
```

libs/polysemy-wire-zoo/test/Test/IntersperseSpec.hs
isovector: Thanks!

elland (approved):

libs/polysemy-wire-zoo/test/Test/IntersperseSpec.hs
```diff
@@ -27,6 +29,25 @@ spec = do
...
+                    -- 0
+                    trace "start"
+                    pure ()
+                    -- 1
+                    trace "middle"
+                    -- 2
+                    _ <- get
+                    -- 3
+                    trace "end"
+    result `shouldBe` [ "0", "start", "1", "middle", "2", "3", "end" ]
```
mdimjasevic: @isovector , I would expect the result to be `["0", "start", "1", "2", "middle", "3", "4", "end"]`, i.e., to another `modify (+1)` to happen between `trace "start"` and `pure ()`. Why is that a wrong expectation? Is it because `pure ()` is not a binding operation?

pcapriotti: By the monad laws, `m1 >>= \a -> pure () >> m2 a = m1 >>= m2`, so you shouldn't expect the existence of a `pure ()` in the middle of a do block to be detectable.

mdimjasevic: Right! Thanks for that.

isovector: Furthermore, the combinator works by rewriting calls to `send`, which the standard `m`-quantified `Monad m => ...` operations do not use.

PR #2773, opened 2022-10-19
FS-1059 Change commit-bundle body type to protobuf format

smatting:
This PR changes the API of the `POST /mls/commit-bundles`:

- the request body must be of type `application/vnd.wire.commit-bundle` which is protobuf serialized object define in https://github.com/wireapp/generic-message-proto/commit/f19b5a1b4ed77b661fccf737d8a06a4e06cd22f1

This PR also adds missing tests for the endpoint and removes a superfluous test

Tracked in https://wearezeta.atlassian.net/browse/FS-1059

## Checklist


libs/wire-api/src/Wire/API/ConverProtoLens.hs
```diff
@@ -0,0 +1,31 @@
...
+-- with this program. If not, see <https://www.gnu.org/licenses/>.
+
+module Wire.API.ConverProtoLens where
+
+import Data.Bifunctor (Bifunctor (first))
+import Imports
+
+-- | This typeclass exists to provide overloaded function names for convertion
+-- between data types generated by proto-lens and data types used in wire
+class ConvertProtoLens a b where
```
pcapriotti, line 25: Functional dependencies here would improve type inference. I'd try something like `ConvertProtoLens a b | a -> b, b -> a`.

libs/wire-api/src/Wire/API/ConverProtoLens.hs
```diff
@@ -0,0 +1,31 @@
...
+  fromProtolens :: a -> Either Text b
+  toProtolens :: b -> a
+
+-- | Add labels to error messages
+label :: Text -> Either Text a -> Either Text a
+label lbl = first ((lbl <> ": ") <>)
```
pcapriotti, line 31: Could we call this something more unique, like say `protoLabel` or similar, so the module can imported more easily and without qualification?

libs/wire-api/src/Wire/API/MLS/CommitBundle.hs
```diff
@@ -31,14 +41,40 @@ data CommitBundle = CommitBundle
...
+        "welcome"
+        ( let bs = view Proto.Mls.welcome protoBundle
+           in if BS.length bs == 0
+                then pure Nothing
+                else Just <$> decodeMLS' bs
+        )
+      <*> CP.label "group_info_bundle" (fromProtolens (view Proto.Mls.groupInfoBundle protoBundle))
+  toProtolens bundle =
+    let commitData = rmRaw (cbCommitMsg bundle)
+        welcomeData = maybe mempty rmRaw (cbWelcome bundle)
```
pcapriotti, line 58: `maybe mempty = foldMap`. But this is fine, of course.

libs/wire-api/src/Wire/API/MLS/Servant.hs
```diff
@@ -33,5 +36,19 @@ instance Accept MLS where
 instance {-# OVERLAPPABLE #-} ParseMLS a => MimeUnrender MLS a where
   mimeUnrender _ = mimeUnrenderMLSWith parseMLS
 
+instance MimeRender MLS OpaquePublicGroupState where
+  mimeRender _ = LBS.fromStrict . unOpaquePublicGroupState
+
 mimeUnrenderMLSWith :: Get a -> LByteString -> Either String a
 mimeUnrenderMLSWith p = first T.unpack . decodeMLSWith p
+
+data CommitBundleMimeType
```
pcapriotti, line 45: You could use `Proto` from `Wire.API.ServantProto` here, instead.

libs/wire-api/test/unit/Test/Wire/API/MLS.hs
```diff
@@ -62,7 +62,7 @@ tests =
       testCase "key package ref" testKeyPackageRef,
       testCase "validate message signature" testVerifyMLSPlainTextWithKey,
       testCase "create signed remove proposal" testRemoveProposalMessageSignature,
-      testCase "parse GroupInfoBundle" testParseGroupInfoBundle
+      testCase "parse GroupInfoBundle" testParseGroupInfoBundle -- TODO: remove this also
```
pcapriotti, line 65: Leftover TODO?

pcapriotti (approved):
Looks good. Reusing the existing code in `Wire.API.ServantProto` might be nice, but this is fine too. Regardless, I would maybe change the mime type to `application/x-protobuf` for consistency with the other endpoints.

libs/wire-api/src/Wire/API/MLS/CommitBundle.hs
elland: We could add that to hlint, what do you think?

libs/wire-api/test/unit/Test/Wire/API/MLS.hs
elland: Since @smatting is not here to clarify, I'm unsure whether to make this `futurework` or disable this test. It is currently green for me anyway…

PR #2774, opened 2022-10-19
Added a default .envrc.local with some info

elland:
Please contribute more information, other possible variables we might want here and why.

### DIFF
--- a/.envrc.local.default
+++ b/.envrc.local.default
@@ -1,4 +1,0 @@
-# In order to run integration tests
-# we need ngin(xz) locally built using nix.
-# Copy/move this file to .envrc.local for that to work
-export INTEGRATION_USE_NGINZ=1
--- a/.envrc.local.default
+++ b/.envrc.local.default
@@ -1,0 +1,5 @@
+# In order to run integration tests
+# we need ngin(xz) locally built using nix.
+# Copy/move this file to .envrc.local for that to work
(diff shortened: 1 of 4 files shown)


jschaul (approved):
Perhaps mention this file in developer docs?

.envrc.local.default
```diff
@@ -0,0 +1,5 @@
+# In order to run integration tests
+# we need ngin(xz) locally built using nix.
+# Copy/move this file to .envrc.local for that to work
+export COMPILE_NGINX_USING_NIX=1
+export INTEGRATION_USE_NGINZ=1
```
smatting, line 5: Instead of this, let's just change the default behaviour when `COMPILE_NGINX_USING_NIX` and `INTEGRATION_USE_NGINZ` are not defined.

smatting (changes requested):

.envrc.local.default
smatting: I checked: That's the only two variables I have set in my .envrc.local file

elland: Sounds like a good idea, but I'm not that versed in bash, I wouldn't want to break anything fiddling with that.

smatting (approved):

.envrc.local.default
smatting: I also had to figure it out (and will forget immediately again :)

PR #2776, opened 2022-10-19
nix: Use pkgs.buildEnv for development environment

akshaymankar:
Using haskellPacakges.shellFor directly forced us to use nix-shell which exports too many environment variables. These environment variables cause a lot of problems, specially for people not using NixOS.

This change reads buildInputs and nativeBuildInputs for the derivation produced by haskellPacakges.shellFor and adds it to paths of pkgs.buildEnv. To allow cabal to find C dependencies, we also have to export PKG_CONFIG_PATH.

## Checklist

### DIFF
--- a/.envrc
+++ b/.envrc
@@ -19,6 +19,7 @@ fi
 
 PATH_add "./.env/bin"
 path_add "PKG_CONFIG_PATH" "./.env/lib/pkgconfig"
+path_add "LIBRARY_PATH" "./.env/lib"
 
 # source .profile from `$env`. This sets NIX_PATH to pkgs defined in
 # ./nix/default.nix. Tis is useful for nix tooling that runs inside this direnv,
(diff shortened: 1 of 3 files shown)


smatting (approved):

elland (approved):
Works for me!

akshaymankar:
@pcapriotti @battermann @jschaul can y'all also please test this to see if this also gets rid of your locale problem and other things that we worked around yesterday?

jschaul (approved):
Much better, my warnings went away.

akshaymankar:
@pcapriotti @battermann in case this works for both of you, please merge the PR. If not, let's look at it on Monday.

battermann:
> @pcapriotti @battermann in case this works for both of you, please merge the PR. If not, let's look at it on Monday.

Warnings went away, but the problem with signed commits remains. However, the latter does not speak against merging.

PR #2778, opened 2022-10-20
Re-lint the codebase

elland:
Now that we've bumped dev tools versions, hlint is catching more cases.


mdimjasevic (approved):

mdimjasevic (approved):
Looks good!

PR #2780, opened 2022-10-20
Use ormolu 0.5.0.1 in dev environment

smatting:
Motivation: Consistency with HLS, which uses the same version. That way if you use HLS for formatting you get the same result.

Ormolu now has formatting depending on the fixity of of operators, which it tries to parse out of files. Sometimes this approach fails that's why you can introduce overrides for fixities in .ormolu files which need to be next to .cabal files.

Without these overrides our reformatting is really ugly
```
infixr 10 .=
infix 4 ===
infix 4 =/=
infixr 3 !!!
infixr 3 <!!
```

These definitions are the same as in `Bilge.Assert`
```
infix 4 ===
infix 4 =/=
infixr 3 !!!
infixr 3 <!!
```
somehow ormolu was not able to read these fixities out of the module.

I don't know why we need this (it's not the same fixity as aeson defines it):
```
infixr 10 .=
```
I would consider it a workaround. 

## Checklist


smatting:
We don't like these changes, look how it breaks the lines with the (.=) operator (both schema-profunctor and aeson)

akshaymankar:
> 

I am disappointed that there is a `.ormolu` file despite saying 
> Implementing one “true” formatting style which admits no configuration.

But, let's try it out anyway, hopefully people won't have diverging opinions on this :crossed_fingers:

libs/hscim/src/Web/Scim/Class/Group.hs
```diff
@@ -87,25 +87,25 @@ data GroupSite tag route = GroupSite
     gsGetGroup ::
       route
         :- Capture "id" (GroupId tag)
-        :> Get '[SCIM] (StoredGroup tag),
+          :> Get '[SCIM] (StoredGroup tag),
```
akshaymankar, line 90: Can this also be fixed with the `.ormolu` file?

libs/wire-api/src/Wire/API/Asset.hs
```diff
@@ -200,9 +200,10 @@ newtype AssetToken = AssetToken {assetTokenAscii :: AsciiBase64Url}
 
 instance ToSchema AssetToken where
   schema =
-    AssetToken <$> assetTokenAscii
-      .= schema
-        & doc' . S.schema . S.example ?~ toJSON ("aGVsbG8" :: Text)
+    AssetToken
+      <$> assetTokenAscii
+        .= schema
+      & doc' . S.schema . S.example ?~ toJSON ("aGVsbG8" :: Text)
```
akshaymankar, line 206: Is this right? I am actually not sure if `doc'` is applied to `schema` or `(AssetToken <$> assetTokenAcii .= schema)`. From the original code it looks like it is applied to `schema`, now it looks like that later.

services/brig/src/Brig/Provider/API.hs
```diff
@@ -137,12 +137,12 @@ routesPublic = do
   get "/provider/activate" (continue activateAccountKeyH) $
     accept "application" "json"
       .&> query "key"
-      .&. query "code"
+        .&. query "code"
```
akshaymankar, line 140: This seems wrong, maybe `.&>` and  `.&.` should be at the same fixity?

libs/wire-api/src/Wire/API/Asset.hs
stephen-smith: `&` is `infixl 1` so it is applying to `AssetToken <$> assetTokenAscii .= schema`

stephen-smith:
> infixr 10 .=

> it's not the same fixity as aeson

Data.Schema is our definition of this.  That file _doesn't have_ a fixity declaration, which might be why Ormolu can't find it.  We should probably add one, either at 4 (to match lens) or 8 (to match aeson).  The "default" is `infixl 9` (per the Haskell Report) but I don't know if that's how Ormolu is treating it.

services/brig/src/Brig/Provider/API.hs
stephen-smith: Currently both are `infixr 3`, though it could be possible that Ormolu isn't picking that up?  Prehaps we need to add them to the .ormolu file?

I don't understand why Ormolu wants to indent these lines.  I can't come up with a logic that is consistent with how it handlles (`:` and `++`) or (`.` and `$`) in other places.

That said, I really don't think "ugly" formatting should be a blocker _for this task_.  If we need to change formatter, or add ORMOLU_DISABLE sections, or quiet hlint, or other we can take that up as an effort separate from aligning CI Ormolu with dev environment HLS/Ormolu.

stephen-smith (approved):
Changes are "fine".

New newer Ormolu adds more line breaks and indentation than I really like.

flokli:
This accidentially reverted the submodule update in `services/nginz/third_party/nginx-module-vts`.

PR #2787, opened 2022-10-21
Remove unused Makefiles

smatting:
## Checklist


elland (approved):
<3

fisx (approved):
in the unlikely event that we agree later that this was a mistake, it should be easy to revert, so :shipit: !

PR #2791, opened 2022-10-24
Use locally built schema binaries for db migrations 

akshaymankar:
Also make sure DB is up to date every time tests are run and not just when DBs
are spun up.

This allows us to not rely on the "latest" tag of these docker images and also make spinning up local ephemeral services faster.


Makefile
```diff
@@ -139,17 +139,41 @@ regen-local-nix-derivations:
...
+	./dist/spar-schema --keyspace spar_test --replication-factor 1 --reset
+else
 	$(EXE_SCHEMA) --keyspace $(package)_test --replication-factor 1 --reset
+endif
 
-# migrate db using cabal
-# For using stack see the Makefile of the package, e.g. services/brig/Makefile
-# Usage: make db-migrate-package package=galley
-.PHONY: db-migrate-package
-db-migrate-package: c
```
fisx, line 151: can you keep a goal that juts emits a deprecation warning tell you what to call instead?  (for both reset and migrate.)

fisx (approved):
LGTM, just one small change request.

PR #2795, opened 2022-10-24
Fix content type used when testing MLS commit bundles

elland:
Fixes the failing tests in CI


fisx (approved):

akshaymankar:
The overall-ci will get green soon, I deployed a change to the pipeline and it fixed it. But it needs to wait for integration tests to also go green before it will be marked as green.

PR #2796, opened 2022-10-24
WIP

jschaul:
⚠️ Do not mess with this PR ⚠️ 

## Checklist


nix/nginz.nix
```diff
@@ -0,0 +1,105 @@
...
+      inputs = [ pkg-config zauth ];
+    };
+  };
+
+  nginz = nginx.override {
+    # openssl = (openssl.overrideAttrs (oldAttrs: {
+    #   patches = oldAttrs.patches or [] ++ [
+    #     ./patch.patch
+    #   ];
+    # }));
```
flokli, line 36:
```suggestion
    openssl = openssl.override { openssl = openssl_1_1; };
```

(as per https://xeiaso.net/blog/nixos-nginx-openssl-1.x), it looks like we should use this for the time being.

flokli:
What's the status of this? Still a draft, or ready to review?

nix/nginz.nix
```diff
@@ -0,0 +1,104 @@
...
+      gnugrep
+      which
+      coreutils
+      nginxFakeNss
+    ];
+    config = {
+      Entrypoint = [ "${dumb-init}/bin/dumb-init" "--" "${reload-script}/bin/nginz_reload.sh" "-g" "daemon off;" "-c" "/etc/wire/nginz/conf/nginx.conf" ];
+      Env = [ "SSL_CERT_FILE=/etc/ssl/certs/ca-bundle.crt" ];
+      ExposedPorts = {
+        "80/tcp" = { };
```
flokli, line 99: Isn't this 8080,8081 and 8082? At least that's the ports I see advertised in k8s…

nix/nginz.nix
```diff
@@ -0,0 +1,104 @@
...
+      };
+    };
+  };
+in
+nginzImage
```
flokli, line 104: This hides away the individual binaries, but it's a problem we also have with the haskell binaries, so happy to tackle this in a followup PR.

nix/nginz.nix
```diff
@@ -0,0 +1,104 @@
...
+, cacert
+, bashInteractive
+, openssl
+, openssl_1_1
+, lib
+, makeWrapper
+, writers
+}:
+let
+  zauth = callPackage ./pkgs/zauth { };
```
flokli, line 24: We should have zauth already in `pkgs`, because we define it in `overlay.nix`.

Try passing zauth as a function argument in the head of the file.

nix/nginz.nix
```diff
@@ -0,0 +1,104 @@
...
+  nginz = nginx.override {
+    openssl = openssl_1_1;
+    modules = [
+      nginzModules.vts
+      nginzModules.moreheaders
+      nginzModules.zauth
+    ];
+  };
+
+  reload-script = stdenv.mkDerivation {
```
flokli, line 42: This name makes it look like it's an "additional side script" that runs and reloads nginz whenever the configuration changes.

But this is a wrapper we run, instead of running nginz directly.

I'd probably rename this `nginzWithReloader` or `nginzWrapped`, and have it write a `$out/bin/nginz` because we execute this as if it was `nginz`.

The fact that this is a shell script executing the real `nginz` is an implementation detail.

nix/nginz.nix
```diff
@@ -0,0 +1,104 @@
...
+  };
+
+  nginz = nginx.override {
+    openssl = openssl_1_1;
+    modules = [
+      nginzModules.vts
+      nginzModules.moreheaders
+      nginzModules.zauth
+    ];
+  };
```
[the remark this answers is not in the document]
jschaul: 👍🏿

nix/nginz.nix
jschaul: > This name makes it look like it's an "additional side script" that runs and reloads nginz whenever the configuration changes.

That's exactly what that script does 😄  so maybe it does have a good name.

I can rename the variables here. I don't like the suggestion of `$out/bin/nginz` because that name has a meaning already inside the rest of wire-server, and it's a different name - locally we actually want to run nginz the binary, not the reload script. So we shouldn't introduce there name clashes that make it harder to talk about things.

jschaul:
With https://github.com/zinfra/cailleach/pull/1437 applied, CI is actually green on this PR. (I undid that change to not block other PRs)

flokli:
With https://github.com/wireapp/wire-server/pull/2828 merged, can we rebase this (and remove the `openssl_1_1` overrides)?

jschaul:
> With #2828 merged, can we rebase this (and remove the `openssl_1_1` overrides)?

Done.

PR #2800, opened 2022-10-26
Disable swagger-ui in prod.

fisx:
this hides https://prod-nginz-https.wire.com/api/swagger-ui/ from prod backends to minimize the attack surface.

it should also hide https://prod-nginz-https.wire.com/swagger-ui/ from prod, which delivers js code needed until we can close https://wearezeta.atlassian.net/browse/BE-529.  i haven't figured out how to do that yet.  but it's somehow related to [this](https://github.com/wireapp/wire-server/blob/807c122a5d99f7b68696c779ebf3468a2c92f400/charts/nginz/templates/swagger-ui-configmap.yaml) and [this](https://github.com/wireapp/wire-server/tree/807c122a5d99f7b68696c779ebf3468a2c92f400/charts/nginz/static).

## Checklist


flokli:
wire-server helm charts shouldn't know what prod and staging is. We should make the fact if we want to expose/run swagger-ui configurable via helm values, and set them then accordingly in environment-specific config (cailleach)

charts/nginz/templates/conf/_nginx.conf.tpl
```diff
@@ -343,10 +342,8 @@ http {
         }
         more_set_headers 'Access-Control-Allow-Origin: $http_origin';
     }
-    {{ end }}
 
     # Swagger UI
-
     location /swagger-ui {
```
jschaul, line 347: Now this block is under `{{ if not (eq $.Values.nginx_conf.env "prod")  }}` as well.

jschaul (approved):

PR #2804, opened 2022-10-31
Fixups

fisx:
## Checklist


libs/wire-api/src/Wire/API/Routes/Public/Galley.hs
```diff
@@ -951,6 +953,7 @@ type TeamAPI =
     "create-non-binding-team"
     ( Summary "Create a new non binding team"
         -- FUTUREWORK: deprecated in https://github.com/wireapp/wire-server/pull/2607
+         -- TODO: take it out of some future version *now*
```
akshaymankar, line 956:
```suggestion
        -- TODO: take it out of some future version *now*
```

PR #2805, opened 2022-10-31
Documentation: add nix cache instructions

isovector:
## Checklist

### DIFF
--- a/docs/src/developer/developer/building.md
+++ b/docs/src/developer/developer/building.md
@@ -1,6 +1,6 @@
 # How to build wire-server
 
-As a prerequisiste install the [nix package manager](https://nixos.org/) and [direnv](https://direnv.net/).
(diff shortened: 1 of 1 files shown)


stephen-smith (approved):
There's already docs for this in docs/src/how-to/install/dependencies.rst (listed as "Optional") and docs/src/developer/developer/dependencies.md (in the configuration), but I don't see much disadvantage to mentioning it here, as well.

isovector:
Is that document linked from the main README? Maybe I missed it?

stephen-smith:
The dependencies.rst isn't linked from anywhere I can see.  The dependencies.md is linked from how-to.md in the same directory, but how-to.md isn't linked from anywhere.  So, I do not see a link chain from README.md to either of those locations.

sysvinit:
We have some internal documentation for Nix and Cachix setup in the README for `cailleach` (which I don't think @isovector is able to see). I think this might have come up once or twice in the past, but as the instructions are relevant both for `wire-server` and `cailleach`, it might be worth moving these docs into the public repo so they're visible to external contributors.

PR #2806, opened 2022-10-31
Support local vim configuration

isovector:
Vim users (especially those using coc.nvim) like keeping local project configuration under `/.vim`. This PR gitignores `/.vim`.

## Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -94,6 +94,7 @@ b.yaml
 telepresence.log
 
 /.ghci
+/.vim
 
 # local config
 .envrc.local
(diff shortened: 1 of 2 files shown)


stephen-smith:
I'm not sure this is appropriate for this particular developer flow to apply this change to every project they interact with.  (I use nvim and Coc and I don't have a .vim directory in each project.)

If a developer needs a common ignore list this they can set core.excludesfile in their global git config, and then add `/.vim/` to it.  I do that myself for backup files created by vim as well as some highlighting files that an old plugin used to generate.

I'm open to be convinced otherwise though, and I won't object if someone else wants to approve.

isovector:
I forgot about global gitignores, will do there instead. Thanks for the reminder!

PR #2810, opened 2022-11-01
Upgrade to GHC 9.2.4

smatting:
This PR upgrades all of `wire-server` to GHC 9.2.4. Some packages needed to be bumped and forked. Please also review the changes in the forks when reviewing this PR.
The commits are organized such that they usually modify only a single package.

## Checklist


docs/src/developer/developer/upgrading.md
```diff
@@ -0,0 +1,32 @@
...
+```
+
+- When a dependency doesn't build anymore because of unmet version constraints see if you can use a never version (`haskell-pins.nix`, or `manual-overrides.nix` if multiple versions of the package are included in the `nixpkgs` pin). Also check out the git repo or issue tracker of the dependency to find forks that work. You can also try removing versions constraints of packages: see `doJailbreak` in `manual-overrides.nix`.
+
+- If test-suites of a dependency don't compile, or cannot be run because they require IO interaction you can disable them: see `dontCheck` in `manual-overrides.nix`
+
+- To force a rebuild of the nix shell delete the `.env` directory and run `direnv reload`
+
+- If you need to fix code in a dependency consider forking it and creating a PR to the upstream after successful adapation. Clone the repo and then symlink it inside `libs/` and run `generate-local-nix-packages.sh` to temporarily include its dependencies in the development shell. Make sure to include the package in `shellFor`. If you've got a working shell you can check the output of `ghc-pkg dump` to see the nix-provided ghc packages used by cabal.
```
elland, line 22:
```suggestion
- If you need to fix code in a dependency consider forking it and creating a PR to the upstream after successful integration. Clone the repo and then symlink it inside `libs/` then run `generate-local-nix-packages.sh` to temporarily include its dependencies in the development shell. Make sure to include the package in `shellFor`. If you've got a working shell you can check the output of `ghc-pkg dump` to see the list of nix-provided ghc packages used by cabal.
```

docs/src/developer/developer/upgrading.md
```diff
@@ -0,0 +1,32 @@
...
+
+- Consider using `ghcid --command "cabal repl <package>"` when fixing packages
```
elland, line 24: Would be nice to give some hints as to why you'd want to do that.

docs/src/developer/developer/upgrading.md
```diff
@@ -0,0 +1,32 @@
...
+
+- Delete the `dist-newstyle` directory everytime you upgrade the nix environment, without any exception. That is because cabal does not pick up changes to the `ghc-pkg` set (defined the development shell). Cabal will complain about missing dependencies in these cases.
+
+- When trying to build any packages with cabal (e.g. for fixing code in a depencency or fixing local packages), make sure you've got the right package set in the `shellFor` argument and the right transitive dependencies in `cabal.project`. It takes a couple of tries to get both: a nix provided environment that work and cabal not complaining about missing dependencies when building inside the environment.
```
elland, line 28:
```suggestion
- When trying to build any packages with cabal (e.g. for fixing code in a depencency or fixing local packages), make sure you've got the right package set in the `shellFor` argument and the right transitive dependencies in `cabal.project`. It takes a couple of tries to get both: a nix provided environment that works, and cabal not complaining about missing dependencies when building inside the environment.
```

docs/src/developer/developer/upgrading.md
```diff
@@ -0,0 +1,32 @@
...
+
+- It might happen that a package's test suite dependencies are not available in the nix environment. When you try to build with cabal it might try to build these external dependencies (which you want to avoid). What might work in these cases temporarily updating the `default.nix` file (generated by `generate-local-nix-packages.sh`) to add the test suits dependencies to the library section.
```
elland, line 30:
```suggestion
- It might happen that a package's test suite dependencies are not available in the nix environment. When you try to build with cabal it might try to build these external dependencies (which you want to avoid). What might work in these cases is to temporarily update the `default.nix` file (generated by `generate-local-nix-packages.sh`) to add the test suits dependencies to the library section.
```

docs/src/developer/developer/upgrading.md
```diff
@@ -0,0 +1,32 @@
...
+
+- If cabal is complaining about missing a missing `libsodium` add `sodium-crypt-sign` to the `shellFor` environment
```
elland, line 32:
```suggestion
- If cabal is complaining about missing `libsodium`: add `sodium-crypt-sign` to the `shellFor` environment.
```

tools/db/assets/src/Assets/Lib.hs
```diff
@@ -188,5 +188,4 @@ instance Semigroup Result where
 
 instance Monoid Result where
   mempty = Result 0 0 []
-  mappend (Result n1 v1 i1) (Result n2 v2 i2) =
-    Result (n1 + n2) (v1 + v2) (i1 <> i2)
+  mappend = (<>)
```
elland, line 191: 👏

elland (commented):
That was way less daunting that I expected 👍. Someone else should review the Nix side of things.

tools/db/assets/src/Assets/Lib.hs
smatting: btw, this was not even my choice, `mappend` must be implemented from Semigroups `(<>)` in the future

docs/src/developer/developer/upgrading.md
smatting: You mean you might need to fix code in a dependency? There is only one reason: the dependency is not compiling.

libs/wire-api/src/Wire/API/Routes/MultiVerb.hs
```diff
@@ -19,7 +19,7 @@
 -- with this program. If not, see <https://www.gnu.org/licenses/>.
 
 module Wire.API.Routes.MultiVerb
-  ( -- * MultiVerb types
+  ( -- Type MultiVerb types
```
pcapriotti, line 22: Search/replace accident here

services/galley/src/Galley/API/Push.hs
```diff
@@ -83,10 +82,6 @@ newBotPush b e = NormalMessagePush {userPushes = mempty, botPushes = pure (b, e)
 
 type BotMap = Map UserId BotMember
 
-type family LocalMemberMap (t :: MessageType) = (m :: *) | m -> t where
-  LocalMemberMap 'NormalMessage = Map UserId LocalMember
-  LocalMemberMap 'Broadcast = ()
-
```
pcapriotti, line 89: Ah, nice, I forgot to remove this when I refactored the push stuff.

docs/src/developer/developer/upgrading.md
elland: Nevermind, I mistead the command :D

elland (approved):
Works on my machine :)

pcapriotti (approved):
Nice job! Minor comments below.

PR #2820, opened 2022-11-03
Split galley servant definition

pcapriotti:
## Checklist


battermann:
I like this, we should do the same with the route definitions in wire-api.

pcapriotti:
> I like this, we should do the same with the route definitions in wire-api.

Sure, I'll make it part of this PR.

stefanwire (approved):
There's no entry in `changelog.d/` yet.

PR #2830, opened 2022-11-07
Avoid client deletion edge case condition which can lead to inconsist…

jschaul:
…ent data between brig and galley's clients tables.

This is improving the invariant from the comment

```haskell
-- nb. We must ensure that the set of clients known to brig is always
-- a superset of the clients known to galley.
```

if the process crashes in between these lines; or if there's a problem with adding something to SQS.

This is still not ideal, as there remains a race condition whereby clients will still get a prompt on message sending to also encrypt for a client that is already no longer existing in brig, if there is A) a bug in processing this queue event, or if B) a client tries to send a message after the deletion is processed in brig, but before the deletion has happened on galley. So another follow-up PR, revisiting or perhaps entirely undoing the asynchronous client deletion would be necessary (or galley should be merged into brig, but we shouldn't wait for this, as dealing with data inconsistencies is much more time-consuming that making a small code change here)

## Checklist

### DIFF
--- a/changelog.d/3-bug-fixes/client-deletion-ordering
+++ b/changelog.d/3-bug-fixes/client-deletion-ordering
@@ -1,0 +1,1 @@
+Avoid client deletion edge case condition which can lead to inconsistent data between brig and galley's clients tables.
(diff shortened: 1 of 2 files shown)


smatting (approved):
Agree. If enqueuing fails the client should not be deleted.

PR #2831, opened 2022-11-08
Treefmt nixpkgs fmt

flokli:
## Checklist


libs/api-bot/default.nix
```diff
@@ -2,27 +2,89 @@
...
 # dependencies are added or removed.
-{ mkDerivation, aeson, ansi-terminal, api-client, async, attoparsec
-, base, base64-bytestring, bilge, bytestring, bytestring-conversion
-, cereal, containers, cryptobox-haskell, cryptonite, exceptions
-, filepath, gitignoreSource, HaskellNet, HaskellNet-SSL
-, http-client, imports, iso639, lib, memory, metrics-core, mime
-, monad-control, mwc-random, optparse-applicative, resource-pool
-, stm, text, time, tinylog, transformers-base, types-common
-, unordered-containers, uuid, vector
+{ mkDerivation
```
jschaul, line 5: could we not exclude these auto-generated files from the formatting? It's a bit useless to format these, no? +4,365 −1,060 seems unnecessary. We should format the files people write themselves.

.github/workflows/ci.yml
```diff
@@ -4,6 +4,25 @@ on:
...
+     - uses: cachix/install-nix-action@v14.1
+     - uses: cachix/cachix-action@v10
+       with:
+         name: wire-server
+         signingKey: '${{ secrets.CACHIX_SIGNING_KEY }}'
+         authToken: '${{ secrets.CACHIX_AUTH_TOKEN }}'
+     - name: Install treefmt and nixpkgs-fmt (without pulling in all of dev-env)
+       run: nix-env -if nix/default.nix -iA pkgs.treefmt pkgs.nixpkgs-fmt pkgs.shellcheck
+     - name: Run treefmt
+       run: treefmt
```
jschaul, line 24: This github action is unnecessary, as linters already run `make lint-all` here: https://github.com/zinfra/cailleach/blob/ab3a921c7a6cc903c0d1d578040e05a566c84bfa/wire-server-private/ci/pipelines/jobs/linters.dhall#L40-L60

please remove this github actions block. Also, we agreed to not introduce more github actions if not strictly necessary.

libs/api-bot/default.nix
jschaul: Alright, okay. 😃

jschaul (commented):
Looks ok modulo the github actions block.

.github/workflows/ci.yml
flokli: I'd like to keep these 20 lines of yaml.

 - They're not a big maintenance burden, as it simply invokes `treefmt` after installing the dependencies.
 - Very quick feedback (40 seconds)
 - No requirement to log into concourse, so useful feedback for external contributors


It would be a different thing if we'd start to model complicated logic in GH actions, but this simply executes a script to give semi-instant feedback.

It looks like Concourse also does it, because it's invoked in the `lint-all` target from the Makefile, but it's way less accessible.

jschaul: it spams emails and re-does something that is already done, giving twice the amount of linting feedback for the same issue. To me, this is just noise and unnecessary. Maybe someone else has an opinion here also.

flokli: You can change if GH actions should send you emails on failure in [your notification settings](https://github.com/settings/notifications):

image

jschaul (approved):

PR #2835, opened 2022-11-08
Add renovate config (SQPIT-416)

supersven:
Trying to reintroduce it after/according to https://github.com/wireapp/wire-server/pull/1651#issuecomment-918937330

## Checklist


supersven:
It's doesn't work that way. But, the original PR (https://github.com/wireapp/wire-server/pull/2314) is still active / alive.

PR #2839, opened 2022-11-09
[FS-925] Lazily create an MLS Self-conversation

mdimjasevic:
This is a follow-up to #2730.

It changes the introduced endpoint's method from `PUT` to `GET`. Furthermore, it creates the self-conversation on the first call to the GET endpoint, but such that the epoch column is not written to so as to avoid undesired race conditions of another client also creating the conversation and bumping the epoch number (the rest of the columns in the conversation record are identical, so there is no harm in overwriting such columns). The epoch number is actually written to the database only upon the first commit.

## Checklist

 - [x] See if there are additional checks that should be done when inserting the creator client's key package ref


services/galley/src/Galley/API/MLS/Message.hs
```diff
@@ -796,10 +796,29 @@ processInternalCommit qusr senderClient con lconv cm epoch groupId action sender
...
         then do
-          -- this is a newly created conversation, and it should contain exactly one
-          -- client (the creator)
-          case (self, cmAssocs cm) of
-            (Left lm, [(qu, (creatorClient, _))])
+          let cType = cnvmType . convMetadata . tUnqualified $ lconv
+          case (self, cType, cmAssocs cm) of
+            (Left _, SelfConv, []) -> do
+              creatorClient <-
+                note (mlsProtocolError "Missing the sender client") senderClient
```
pcapriotti, line 803: It's probably better to make this a 403. It might be good to include more context here in the error message. Something that explains that the client has to refresh their access token and provide their client ID.

services/galley/src/Galley/Cassandra/Conversation.hs
```diff
@@ -56,6 +56,66 @@ import Wire.API.MLS.CipherSuite
 import Wire.API.MLS.Group
 import Wire.API.MLS.PublicGroupState
 
+createMLSSelfConversation ::
```
pcapriotti, line 59: Does this function really need to take in a whole `NewConversation` object? I don't think we should be handling team conversations and proteus conversations here, since this function is specific to MLS self conversations.

services/galley/src/Galley/Cassandra/Conversation.hs
```diff
@@ -266,7 +326,12 @@ toProtocol ::
 toProtocol Nothing _ _ _ = Just ProtocolProteus
 toProtocol (Just ProtocolProteusTag) _ _ _ = Just ProtocolProteus
 toProtocol (Just ProtocolMLSTag) mgid mepoch mcs =
-  ProtocolMLS <$> (ConversationMLSData <$> mgid <*> mepoch <*> mcs)
+  ProtocolMLS <$> (ConversationMLSData <$> mgid <*> mapEpoch mepoch <*> mcs)
+  where
+    -- If there is no epoch in the database, assume the epoch is 0 and this is a
+    -- self-conversation.
+    mapEpoch Nothing = Just (Epoch 0)
+    mapEpoch x = x
```
pcapriotti, line 334: I would just inline `mapEpoch`:

```.haskell
mapEpoch = (<> Just (Epoch 0))
```

services/galley/src/Galley/Cassandra/Conversation.hs
```diff
@@ -266,7 +326,12 @@ toProtocol ::
 toProtocol Nothing _ _ _ = Just ProtocolProteus
 toProtocol (Just ProtocolProteusTag) _ _ _ = Just ProtocolProteus
 toProtocol (Just ProtocolMLSTag) mgid mepoch mcs =
-  ProtocolMLS <$> (ConversationMLSData <$> mgid <*> mepoch <*> mcs)
+  ProtocolMLS <$> (ConversationMLSData <$> mgid <*> mapEpoch mepoch <*> mcs)
+  where
+    -- If there is no epoch in the database, assume the epoch is 0 and this is a
+    -- self-conversation.
```
pcapriotti, line 332: We don't actually need to assume that it is a self-conversation, do we?

pcapriotti (approved):
Looks good modulo comments below.

services/galley/src/Galley/API/MLS/Message.hs
mdimjasevic: Thanks! Addressed it in commit 341aeddde (to be pushed soon).

services/galley/src/Galley/Cassandra/Conversation.hs
mdimjasevic: Fair points! Addressed in commit e2de44d76.

services/galley/src/Galley/Cassandra/Conversation.hs
mdimjasevic: Done almost the same in commit e2de44d76.

services/galley/src/Galley/Cassandra/Conversation.hs
mdimjasevic: I dropped that part of the comment. This was the first time (from this PR's point of view) that we started with this approach to MLS conversation creation, hence the comment does apply to self-conversations only.

PR #2844, opened 2022-11-15
Simplify and cleanup MLS message handling

pcapriotti:
This is an attempt to simplify and streamline the logic of MLS message handling. There is still a lot of complexity there, but hopefully less than before. This PR also introduces optional Client IDs taken from the tokens into the MLS message handlers. Making those IDs mandatory and changing conversation creation logic in the V3 API will happen in a separate PR.

## Checklist

### DIFF
--- a/services/galley/src/Galley/API/MLS/Message.hs
+++ b/services/galley/src/Galley/API/MLS/Message.hs
@@ -610,6 +610,92 @@ processCommit qusr senderClient con lconv cm epoch sender commit = do
   (groupId, action) <- getCommitData lconv epoch commit
   processCommitWithAction qusr senderClient con lconv cm epoch groupId action sender commit
 
+processExternalCommit ::
+  forall r.
+  Members
+    '[ BrigAccess,
+       ConversationStore,
+       Error MLSProtocolError,
+       ErrorS 'MLSStaleMessage,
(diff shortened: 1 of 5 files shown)


services/galley/src/Galley/API/MLS/Message.hs
```diff
@@ -281,6 +286,7 @@ postMLSCommitBundleToLocalConv qusr conn bundle lcnv = do
                 /= Set.fromList (map (snd . snd) (cmAssocs (paAdd action)))
             )
             $ throwS @'MLSWelcomeMismatch
+        storeGroupInfoBundle lconv (cbGroupInfoBundle bundle)
```
mdimjasevic, line 289: Isn't this too eagerly storing group info? Didn't we say this cannot be moved here as there are checks that have to run before this is stored?

mdimjasevic (approved):
@pcapriotti , it all looks good except that I am not too sure of that group info storing before checks. If you confirm it is fine this way, feel free to merge.

smatting (approved):

services/galley/src/Galley/API/MLS/Message.hs
pcapriotti: Thanks for catching it, I had forgotten. Fixed now.

PR #2849, opened 2022-11-16
nginz deactivate unused upstreams (SQPIT-1174)

supersven:
### Problem

Some upstreams are configured by default in `nginz` that have no backing service. This leads to many unnecessary DNS lookups and opens unused routes into our K8s clusters (which might be a security issue).

### Solution
Upstream configuration in `nginz` goes from two entries (`ignored_upstreams`, `upstreams`) to four (`ignored_upstreams`, `upstreams`, `enabled_extra_upstreams`, `extra_upstreams`):

- `ignored_upstreams` : default upstreams that should be ignored (for backwards compatibility)
- `upstreams` : default upstreams (kept for backwards compatibility)
- `enabled_extra_upstreams` : Extra upstreams that should be enabled (routed / rendered; this one is new)
- `extra_upstreams` : Contains upstreams that are usually not deployed by default (this is new, too)

I see two benefits with this:

- Existing on-prems only need to be touched when they are using an extra upstream. This is likely only `proxy`.
- Developers should see that the place to add upstreams for their services is likely `extra_upstreams`. (As long as we don’t add one to `wire-server`, which doesn’t happen very often.)

**Ticket**: https://wearezeta.atlassian.net/browse/SQPIT-1174

### Configuration changes
The environment configurations must be changed before this change is deployed. Otherwise, the extra services (`ibis`, `galeb`, `calling-test` and `proxy`) won't be reachable.

- [`staging`](https://github.com/zinfra/cailleach/pull/1451)
- [`prod`](https://github.com/zinfra/cailleach/pull/1452)

## Checklist

 - [x] Config changes for `staging` and `prod` have been merged and applied

### DIFF
--- a/changelog.d/2-features/disable-extra-nginz-upstreams-by-default
+++ b/changelog.d/2-features/disable-extra-nginz-upstreams-by-default
@@ -4,7 +4,7 @@ be enabled by adding their name (entry's key) to
 `nginx_conf.enabled_extra_upstreams`. `nginx_conf.ignored_upstreams` is only
 applied to upstreams from `nginx_conf.upstreams`. In the default configuration
 of `nginz` extra upstreams are `ibis`, `galeb`, `calling-test` and `proxy`. If one
-of those is deployed, it's name has be be added to
+of those is deployed, its name has be be added to
 `nginx_conf.enabled_extra_upstreams` (otherwise, it won't be reachable). Unless
 `nginx_conf.upstreams` hasn't been changed manually (overriding it's default),
 this should be the only needed migration step.
--- a/changelog.d/2-features/disable-extra-nginz-upstreams-by-default
+++ b/changelog.d/2-features/disable-extra-nginz-upstreams-by-default
@@ -1,0 +1,10 @@
+The list of upstreams is split into `nginx_conf.upstreams` and
+`nginx_conf.extra_upstreams`. Extra upstreams are disabled by default. They can
+be enabled by adding their name (entry's key) to
+`nginx_conf.enabled_extra_upstreams`. `nginx_conf.ignored_upstreams` is only
+applied to upstreams from `nginx_conf.upstreams`. In the default configuration
+of `nginz` extra upstreams are `ibis`, `galeb`, `calling-test` and `proxy`. If one
+of those is deployed, it's name has be be added to
+`nginx_conf.enabled_extra_upstreams` (otherwise, it won't be reachable). Unless
+`nginx_conf.upstreams` hasn't been changed manually (overriding it's default),
+this should be the only needed migration step.
--- a/changelog.d/2-features/disable-extra-nginz-upstreams-by-default
+++ b/changelog.d/2-features/disable-extra-nginz-upstreams-by-default
@@ -6,5 +6,5 @@ applied to upstreams from `nginx_conf.upstreams`. In the default configuration
 of `nginz` extra upstreams are `ibis`, `galeb`, `calling-test` and `proxy`. If one
 of those is deployed, its name has be be added to
 `nginx_conf.enabled_extra_upstreams` (otherwise, it won't be reachable). Unless
-`nginx_conf.upstreams` hasn't been changed manually (overriding it's default),
+`nginx_conf.upstreams` hasn't been changed manually (overriding its default),
 this should be the only needed migration step.
--- a/changelog.d/2-features/disable-extra-nginz-upstreams-by-default
+++ b/changelog.d/2-features/disable-extra-nginz-upstreams-by-default
@@ -1,4 +1,4 @@
-The list of upstreams is split into `nginx_conf.upstreams` and
+**Nginz helm chart**: The list of upstreams is split into `nginx_conf.upstreams` and
 `nginx_conf.extra_upstreams`. Extra upstreams are disabled by default. They can
 be enabled by adding their name (entry's key) to
 `nginx_conf.enabled_extra_upstreams`. `nginx_conf.ignored_upstreams` is only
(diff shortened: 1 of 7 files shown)


jschaul (commented):
I'm a little confused as to the necessity for this extra complexity. 

I find it hard to grasp these new categories of default, extra, extra_enabled and ignored. Is this complexity necessary?

Couldn't this PR simple be a one-line change from

```yaml
# charts/nginz/values.yaml
   ignored_upstreams: []
```

to 

```yaml
# charts/nginz/values.yaml
   ignored_upstreams: ["ibis", "galeb", "calling-test", "proxy"]
```

As a default for the most common deployments? Wouldn't that result in the same behaviour that you're looking for? (and for our own deployments, since we already have a custom override for `ignored_upstreams` we don't need to do anything). Maybe I misunderstand the use cases, please elaborate if I didn't see this well enough 😸

supersven:
> [...] and for our own deployments, since we already have a custom override for ignored_upstreams we don't need to do anything [...]

We would and we would silently not notice.

For `prod` `ignored_upstreams` is configured as:
```yaml
ignored_upstreams: ["cannon"]
```
(https://github.com/zinfra/cailleach/blob/master/targets/wire/prod/app/helm_vars/wire-server/values.yaml.gotmpl#L287)

So, the default is overridden, but the invalid upstream (`calling-test`) is still there.

That's why @comawill came up with the idea of an allow list.

So, this solution is a mixture of:
- Create an allow list for extra services (`enabled_extra_upstreams`)
- Keep the old way such that we don't have to change anything on most On-Prems (those that don't use `proxy`; I guess there are no On-Prems with `ibis`, `calling-test` and `galley`.). 

We could trade Helm chart simplicity for more complex configurations, of course: If we swap the logic from `ignored_upstreams` to `enabled_upstreams`, i.e. an allow list.

One drawback I see is that it will be a longer list. E.g. to add the `proxy` upstream to your environment:

```
enabled_upstreams:
  - brig
  - cannon
  - cargohold
  - gundeck
  - galley
  - spar
  - proxy
```
Vs.
```
enabled_extra_upstreams:
  - proxy
```

From a user's perspective I like my solution more. Though, I can understand that the logic behind the scenes is a bit frightening.

What do you think, @jschaul ?

changelog.d/2-features/disable-extra-nginz-upstreams-by-default
```diff
@@ -0,0 +1,10 @@
+The list of upstreams is split into `nginx_conf.upstreams` and
+`nginx_conf.extra_upstreams`. Extra upstreams are disabled by default. They can
+be enabled by adding their name (entry's key) to
+`nginx_conf.enabled_extra_upstreams`. `nginx_conf.ignored_upstreams` is only
+applied to upstreams from `nginx_conf.upstreams`. In the default configuration
+of `nginz` extra upstreams are `ibis`, `galeb`, `calling-test` and `proxy`. If one
+of those is deployed, it's name has be be added to
```
jschaul, line 7:
```suggestion
of those is deployed, its name has be be added to
```

changelog.d/2-features/disable-extra-nginz-upstreams-by-default
```diff
@@ -0,0 +1,10 @@
...
+`nginx_conf.enabled_extra_upstreams` (otherwise, it won't be reachable). Unless
+`nginx_conf.upstreams` hasn't been changed manually (overriding it's default),
```
jschaul, line 9:
```suggestion
`nginx_conf.upstreams` hasn't been changed manually (overriding its default),
```

changelog.d/2-features/disable-extra-nginz-upstreams-by-default
```diff
@@ -0,0 +1,10 @@
+The list of upstreams is split into `nginx_conf.upstreams` and
```
jschaul, line 1:
```suggestion
**Nginz helm chart**: The list of upstreams is split into `nginx_conf.upstreams` and
```

jschaul (approved):
I see. It's a valid point, and good backwards compatibility is easier to deal with than getting people to change their config.

jschaul:
Perhaps you could document this new behaviour and the defaults somewhere in the nginz chart/README or (preferred) inside docs/src/ somewhere. Can be part of this PR or in a separate PR.

supersven:
> Perhaps you could document this new behaviour and the defaults somewhere in the nginz chart/README or (preferred) inside docs/src/ somewhere. Can be part of this PR or in a separate PR.

Hey @jschaul :wave: 
I've added some docs here: https://github.com/wireapp/wire-server/pull/2849/files#diff-b3222f37f5bc811758c8c49b8109f457707875d62428ba2cb89251134afa14a6R180

Could you please do me a favor and proof-read it before I merge this branch?

jschaul:
This PR has some weirdness and seems to have been rebased on an old develop maybe or otherwise force-pushed. Could you re-rebase this? I think I can't make code suggestions easily on this branch as it is right now. Documentation looks fine content and grammar wise, there are just a few syntax errors in the restructured text that I can fix after a rebase. Thanks!

supersven:
> This PR has some weirdness and seems to have been rebased on an old develop maybe or otherwise force-pushed. Could you re-rebase this? I think I can't make code suggestions easily on this branch as it is right now. Documentation looks fine content and grammar wise, there are just a few syntax errors in the restructured text that I can fix after a rebase. Thanks!

@jschaul , Indeed, it looks like I've done `git rebase develop` instead of `git rebase origin/develop`. :disappointed: 

This should be fixed now.

jschaul (approved):
I pushed two commits to the docs; looks good otherwise, thanks for documenting this!

supersven:
> I pushed two commits to the docs; looks good otherwise, thanks for documenting this!

@jschaul , Well, thanks a lot for overhauling this! :+1:

PR #2851, opened 2022-11-17
[SQSERVICE-253] Support provisioning role information with SCIM

battermann:
https://wearezeta.atlassian.net/browse/SQSERVICES-253

- `POST` and `PUT`are supported
- `PATCH` is not supported, yet, see subsequent PR
- SCIM will override changes from team management
- an empty `roles` list will always be interpreted as `member` and will override any previously set role

## Checklist

### DIFF
--- a/services/brig/src/Brig/API/User.hs
+++ b/services/brig/src/Brig/API/User.hs
@@ -256,7 +256,7 @@ createUserSpar new = do
     pure account
 
   -- Add to team
-  userTeam <- withExceptT CreateUserSparRegistrationError $ addUserToTeamSSO account tid (SSOIdentity ident Nothing Nothing) (fromMaybe defaultRole $ newUserSparRole new)
+  userTeam <- withExceptT CreateUserSparRegistrationError $ addUserToTeamSSO account tid (SSOIdentity ident Nothing Nothing) (newUserSparRole new)
 
   -- Set up feature flags
   let uid = userId (accountUser account)
--- a/services/brig/src/Brig/API/User.hs
+++ b/services/brig/src/Brig/API/User.hs
@@ -256,7 +256,7 @@ createUserSpar new = do
     pure account
 
   -- Add to team
-  userTeam <- withExceptT CreateUserSparRegistrationError $ addUserToTeamSSO account tid (SSOIdentity ident Nothing Nothing)
+  userTeam <- withExceptT CreateUserSparRegistrationError $ addUserToTeamSSO account tid (SSOIdentity ident Nothing Nothing) (fromMaybe defaultRole $ newUserSparRole new)
 
   -- Set up feature flags
   let uid = userId (accountUser account)
@@ -274,10 +274,10 @@ createUserSpar new = do
         Just handl -> withExceptT CreateUserSparHandleError $ changeHandle uid Nothing handl AllowSCIMUpdates
         Nothing -> throwE $ CreateUserSparHandleError ChangeHandleInvalid
 
-    addUserToTeamSSO :: UserAccount -> TeamId -> UserIdentity -> ExceptT RegisterError (AppT r) CreateUserTeam
-    addUserToTeamSSO account tid ident = do
+    addUserToTeamSSO :: UserAccount -> TeamId -> UserIdentity -> Role -> ExceptT RegisterError (AppT r) CreateUserTeam
+    addUserToTeamSSO account tid ident role = do
       let uid = userId (accountUser account)
-      added <- lift $ liftSem $ GalleyProvider.addTeamMember uid tid (Nothing, defaultRole)
+      added <- lift $ liftSem $ GalleyProvider.addTeamMember uid tid (Nothing, role)
       unless added $
         throwE RegisterErrorTooManyTeamMembers
       lift $ do
--- a/services/brig/src/Brig/API/User.hs
+++ b/services/brig/src/Brig/API/User.hs
@@ -538,13 +538,14 @@ initAccountFeatureConfig uid = do
 createUserInviteViaScim ::
   Members
     '[ BlacklistStore,
-       UserPendingActivationStore p
+       UserPendingActivationStore p,
+       GalleyProvider
      ]
     r =>
   UserId ->
   NewUserScimInvitation ->
   ExceptT Error.Error (AppT r) UserAccount
-createUserInviteViaScim uid (NewUserScimInvitation tid loc name rawEmail) = do
+createUserInviteViaScim uid (NewUserScimInvitation tid loc name rawEmail _) = do
   email <- either (const . throwE . Error.StdError $ errorToWai @'E.InvalidEmail) pure (validateEmail rawEmail)
   let emKey = userEmailKey email
   verifyUniquenessAndCheckBlacklist emKey !>> identityErrorToBrigError
@@ -565,7 +566,6 @@ createUserInviteViaScim uid (NewUserScimInvitation tid loc name rawEmail) = do
         -- the SCIM user.
         True
   lift . wrapClient $ Data.insertAccount account Nothing Nothing activated
-
   pure account
 
 -- | docs/reference/user/registration.md {#RefRestrictRegistration}.
--- a/services/galley/src/Galley/API/Teams.hs
+++ b/services/galley/src/Galley/API/Teams.hs
@@ -50,6 +50,7 @@ module Galley.API.Teams
     uncheckedGetTeamMember,
     uncheckedGetTeamMembersH,
     uncheckedDeleteTeamMember,
+    uncheckedUpdateTeamMember,
     userIsTeamOwner,
     canUserJoinTeam,
     ensureNotTooLargeForLegalHold,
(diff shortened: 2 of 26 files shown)


services/spar/test-integration/Test/Spar/APISpec.hs
```diff
@@ -1305,6 +1307,76 @@ specScimAndSAML = do
...
+    it "update user" $ do
+      (tok, (owner, tid, _idp, (_, _privcreds))) <- ScimT.registerIdPAndScimTokenWithMeta
+      scimUserWithDefaultRole <- ScimT.randomScimUser
+      userId <- ScimT.scimUserId <$> ScimT.createUser tok scimUserWithDefaultRole
+      let testUpdateUserWithRole role = do
+            let scimUserWithRole = scimUserWithDefaultRole {Scim.roles = [cs $ toByteString $ role]}
+            _ <- ScimT.updateUser tok userId scimUserWithRole
+            ScimT.checkTeamMembersRole tid owner userId role
+      mapM_ testUpdateUserWithRole [minBound .. maxBound]
+    it "update user - default to member if no role given" $ do
```
fisx, line 1354: shouldn't this be the previous value instead of the default?  not sure, though.  i think we have an issue here with not being able to distinguish `[]` from `null` or missing.

services/spar/test-integration/Test/Spar/Scim/UserSpec.hs
```diff
@@ -1704,6 +1775,38 @@ testBrigSideIsUpdated = do
...
+testUpdateUserRole = do
+  env <- ask
+  let brig = env ^. teBrig
+  let galley = env ^. teGalley
+  (owner, tid) <- call $ createUserWithTeam brig galley
+  tok <- registerScimToken tid Nothing
+  let testWithInitialRole r = forM_ [minBound .. maxBound] (testUpdateUserRoles brig tid owner tok r)
+  forM_ [minBound .. maxBound] testWithInitialRole
+
+testUpdateUserRoles :: BrigReq -> TeamId -> UserId -> ScimToken -> Role -> Role -> TestSpar ()
```
fisx, line 1788: better names for `testUpdateUserRole`, `testUpdateUserRoles`?

services/spar/test-integration/Test/Spar/Scim/UserSpec.hs
```diff
@@ -1704,6 +1775,38 @@ testBrigSideIsUpdated = do
...
 
+testUpdateUserRole :: TestSpar ()
+testUpdateUserRole = do
+  env <- ask
+  let brig = env ^. teBrig
+  let galley = env ^. teGalley
+  (owner, tid) <- call $ createUserWithTeam brig galley
+  tok <- registerScimToken tid Nothing
+  let testWithInitialRole r = forM_ [minBound .. maxBound] (testUpdateUserRoles brig tid owner tok r)
+  forM_ [minBound .. maxBound] testWithInitialRole
```
fisx, line 1786:
```suggestion
  testUpdateUserRoles brig tid owner tok `mapM` [minBound..] `mapM` [minBound..]
```

not sure that'll compile, but it looks nicer.  :-)

services/spar/src/Spar/Scim/User.hs
```diff
@@ -271,9 +271,28 @@ validateScimUser' errloc midp richInfoLimit user = do
     either err pure $ Brig.mkUserName (Scim.displayName user) veid
   richInfo <- validateRichInfo (Scim.extra user ^. ST.sueRichInfo)
   let active = Scim.active user
-  lang <- maybe (error "Could not parse language. Expected format is ISO 639-1.") pure $ mapM parseLanguage $ Scim.preferredLanguage user
```
fisx, line 274: oops!

services/spar/src/Spar/Scim/User.hs
```diff
@@ -271,9 +271,28 @@ validateScimUser' errloc midp richInfoLimit user = do
...
+            (throwError $ badRequest $ "The role '" <> roleName <> "' is not valid. Valid roles are " <> validRoleNames <> ".")
+            (pure . Just)
+            (fromByteString $ cs roleName)
+        _ -> throwError $ badRequest "A user cannot have more than one role."
+
+    badRequest :: Text -> Scim.ScimError
+    badRequest msg =
+      Scim.badRequest
+        Scim.InvalidValue
+        (Just $ msg <> " (" <> errloc <> ")")
```
fisx, line 295:
```suggestion
        (Just $ msg <> " (" <> errloc <> ")")
        
```

services/spar/src/Spar/Scim/User.hs
```diff
@@ -271,9 +271,28 @@ validateScimUser' errloc midp richInfoLimit user = do
...
+
+    validateRole =
+      Scim.roles <&> \case
+        [] -> pure Nothing
+        [roleName] ->
+          maybe
+            (throwError $ badRequest $ "The role '" <> roleName <> "' is not valid. Valid roles are " <> validRoleNames <> ".")
+            (pure . Just)
+            (fromByteString $ cs roleName)
+        _ -> throwError $ badRequest "A user cannot have more than one role."
```
fisx, line 289:
```suggestion
        (_ : _ : _) -> throwError $ badRequest "A user cannot have more than one role."
```

this way, order of cases doesn't matter, and it's more clear what you're doing.

services/brig/src/Brig/API/User.hs
```diff
@@ -538,13 +538,14 @@ initAccountFeatureConfig uid = do
...
-       UserPendingActivationStore p
+       UserPendingActivationStore p,
+       GalleyProvider
      ]
     r =>
   UserId ->
   NewUserScimInvitation ->
   ExceptT Error.Error (AppT r) UserAccount
-createUserInviteViaScim uid (NewUserScimInvitation tid loc name rawEmail) = do
+createUserInviteViaScim uid (NewUserScimInvitation tid loc name rawEmail _) = do
```
fisx, line 548: given you've tested this well, you probably don't want to write the role to cassandra here, but...  why?  i'm confused.  :)

services/spar/src/Spar/Intra/Galley.hs
```diff
@@ -91,3 +110,20 @@ isEmailValidationEnabledTeam tid = do
...
     )
+
+-- | Update a team member.
+updateTeamMember ::
+  (MonadIO m, HasCallStack, MonadError SparError m, MonadSparToGalley m) =>
+  UserId ->
+  TeamId ->
+  Role ->
+  m ()
+updateTeamMember u tid role = do
```
fisx, line 121: if you're only passing in a role here, you shouldn't construct the entire member from scratch and write that.  you should only update the role.  otherwise you may overwrite current or yet-to-be-introduced attributes of that member with default values.

services/spar/src/Spar/App.hs
```diff
@@ -169,7 +170,7 @@ createSamlUserWithId ::
   Sem r ()
 createSamlUserWithId teamid buid suid = do
   uname <- either (throwSparSem . SparBadUserName . cs) pure $ Intra.mkUserName Nothing (UrefOnly suid)
-  buid' <- BrigAccess.createSAML suid buid teamid uname ManagedByWire Nothing Nothing Nothing
+  buid' <- BrigAccess.createSAML suid buid teamid uname ManagedByWire Nothing Nothing Nothing defaultRole
```
fisx, line 173: there is only one place where we call this function; better take a role as argument here and pass the default in there; that'll make it harder to forget about roles in future refactorings.

fisx (commented):
still need to read brig and galley, otherwise looks good.  a few comments.

services/spar/test-integration/Test/Spar/APISpec.hs
battermann: IMO it is not surprising for the SCIM admin when they change the `roles` field from the previous SCIM provider setting (e.g. `[owner]`) to `[]` or `null` that this leads to the role being reset to the default. What would you expect as a user?

services/brig/src/Brig/API/User.hs
battermann: The role is stored in C* just before this function is called.
In this function we create an entry in the pending activation store where we do not need the role, but since it's part of the type it needs to be included in the deconstructor.

services/spar/src/Spar/Intra/Galley.hs
battermann: AFAICT this endpoint only exists to update a team member's permissions (equivalent to role). Are you concerned about not yet existing future functionality?

services/spar/test-integration/Test/Spar/APISpec.hs
fisx: ok, makes sense.  maybe we can add something to that end somewhere under docs.wire.com?  (separate PR though.)

services/spar/src/Spar/Intra/Galley.hs
fisx: > Are you concerned about not yet existing future functionality?

yes.  that, and i didn't actually check the types.  i was expecting there to be something else, too.

services/galley/src/Galley/API/Teams.hs
```diff
@@ -783,58 +784,40 @@ updateTeamMember ::
...
-  let zusr = tUnqualified lzusr
+uncheckedUpdateTeamMember mlzusr mZcon tid newMember = do
+  let mZusr = tUnqualified <$> mlzusr
   let targetMember = ntmNewTeamMember newMember
   let targetId = targetMember ^. userId
       targetPermissions = targetMember ^. permissions
   P.debug $
     Log.field "targets" (toByteString targetId)
-      . Log.field "action" (Log.val "Teams.updateTeamMember")
+      . Log.field "action" (Log.val "Teams.uncheckedUpdateTeamMember")
```
fisx, line 799: i would leave the old value.  they are both equally wrong, but the old one is less misleading when you expect it to be checked, and it actually is called through `updateTeamMember`.

fisx (approved):

services/galley/src/Galley/API/Teams.hs
battermann, line 799:
```suggestion
      . Log.field "action" (Log.val "Teams.updateTeamMember")
```

PR #2859, opened 2022-11-21
Use MLS member table indexed by group id

smatting:
This PR adds a table `galley.mls_group_member_client` (indexed by group id) which replaces `galley.member_client` (index by conv). Motivation: This table may be used by MLS subconversations. This PR also includes a data migration from  `galley.member_client` to `galley.mls_group_member_client` (tested locally).
## Checklist

### DIFF
--- a/cassandra-schema.cql
+++ b/cassandra-schema.cql
@@ -364,6 +364,7 @@ CREATE TABLE galley_test.group_id_conv_id (
     AND read_repair_chance = 0.0
     AND speculative_retry = '99PERCENTILE';
 
+-- NOTE: this table is unused. It was replaced by mls_group_member_client
 CREATE TABLE galley_test.member_client (
     conv uuid,
     user_domain text,
@@ -419,6 +420,29 @@ CREATE TABLE galley_test.conversation_codes (
     AND caching = {'keys': 'ALL', 'rows_per_partition': 'NONE'}
     AND comment = ''
     AND compaction = {'class': 'org.apache.cassandra.db.compaction.LeveledCompactionStrategy'}
+    AND compression = {'chunk_length_in_kb': '64', 'class': 'org.apache.cassandra.io.compress.LZ4Compressor'}
+    AND crc_check_chance = 1.0
+    AND dclocal_read_repair_chance = 0.1
+    AND default_time_to_live = 0
+    AND gc_grace_seconds = 864000
+    AND max_index_interval = 2048
+    AND memtable_flush_period_in_ms = 0
+    AND min_index_interval = 128
+    AND read_repair_chance = 0.0
+    AND speculative_retry = '99PERCENTILE';
+
+CREATE TABLE galley_test.mls_group_member_client (
+    group_id blob,
+    user_domain text,
+    user uuid,
+    client text,
+    key_package_ref blob,
+    PRIMARY KEY (group_id, user_domain, user, client)
+) WITH CLUSTERING ORDER BY (user_domain ASC, user ASC, client ASC)
+    AND bloom_filter_fp_chance = 0.01
+    AND caching = {'keys': 'ALL', 'rows_per_partition': 'NONE'}
+    AND comment = ''
+    AND compaction = {'class': 'org.apache.cassandra.db.compaction.SizeTieredCompactionStrategy', 'max_threshold': '32', 'min_threshold': '4'}
     AND compression = {'chunk_length_in_kb': '64', 'class': 'org.apache.cassandra.io.compress.LZ4Compressor'}
     AND crc_check_chance = 1.0
     AND dclocal_read_repair_chance = 0.1
--- a/services/galley/schema/src/Main.hs
+++ b/services/galley/schema/src/Main.hs
@@ -79,6 +79,7 @@ import qualified V73_MemberClientTable
 import qualified V74_ExposeInvitationsToTeamAdmin
 import qualified V75_MLSGroupInfo
 import qualified V76_ProposalOrigin
+import qualified V77_MLSGroupMemberClient
 
 main :: IO ()
 main = do
@@ -143,7 +144,8 @@ main = do
       V73_MemberClientTable.migration,
       V74_ExposeInvitationsToTeamAdmin.migration,
       V75_MLSGroupInfo.migration,
-      V76_ProposalOrigin.migration
+      V76_ProposalOrigin.migration,
+      V77_MLSGroupMemberClient.migration
       -- When adding migrations here, don't forget to update
       -- 'schemaVersion' in Galley.Cassandra
       -- (see also docs/developer/cassandra-interaction.md)
--- a/services/galley/src/Galley/API/Action.hs
+++ b/services/galley/src/Galley/API/Action.hs
@@ -57,7 +57,6 @@ import Galley.API.Util
 import Galley.App
 import Galley.Data.Conversation
 import qualified Galley.Data.Conversation as Data
-import qualified Galley.Data.Conversation.Types as Data
 import Galley.Data.Services
 import Galley.Data.Types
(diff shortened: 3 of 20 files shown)


cassandra-schema.cql
```diff
@@ -430,6 +430,26 @@ CREATE TABLE galley_test.conversation_codes (
     AND read_repair_chance = 0.0
     AND speculative_retry = '99PERCENTILE';
 
+CREATE TABLE galley_test.billing_team_member (
```
smatting, line 433: This diff is a bit misleading

services/galley/schema/src/Main.hs
```diff
@@ -143,7 +144,8 @@ main = do
       V73_MemberClientTable.migration,
       V74_ExposeInvitationsToTeamAdmin.migration,
       V75_MLSGroupInfo.migration,
-      V76_ProposalOrigin.migration
+      V76_ProposalOrigin.migration,
+      V77_MLSSubconversation.migration
```
smatting, line 148: This module should be renamed

services/galley/schema/src/Main.hs
```diff
@@ -143,7 +144,8 @@ main = do
...
       -- When adding migrations here, don't forget to update
       -- 'schemaVersion' in Galley.Cassandra
```
smatting, line 150: Should maybe add a FUTUREWORK here to delete old galley.member_clients table

services/galley/schema/src/V77_MLSSubconversation.hs
```diff
@@ -0,0 +1,36 @@
...
+
+import Cassandra.Schema
+import Imports
+import Text.RawString.QQ
+
+migration :: Migration
+migration =
+  Migration 77 "Add a new version member_client that is indexed by group_id" $ do
+    schema'
+      [r| CREATE TABLE member_client_v2 (
```
mdimjasevic, line 28: How about naming the table with something more descriptive, e.g., `member_client_by_groupid`? "v2" can be confused with lots of things, including the client API v2.

smatting: I rename it to `mls_group_member_client`. Motivation for the `v2` suffix was so that it is listed next to the table is deprecating (for better discoverability). Found a different solutoin: Add the note when generating the schema dump

services/galley/src/Galley/API/MLS/Message.hs
```diff
@@ -598,21 +601,19 @@ getCommitData ::
...
-    preview (to convProtocol . _ProtocolMLS) (tUnqualified lconv)
-      & noteS @'ConvNotFound
-
-  let curEpoch = cnvmlsEpoch convMeta
-      groupId = cnvmlsGroupId convMeta
+  Sem r ProposalAction
+getCommitData lconv mlsMeta epoch commit = do
+  let curEpoch = cnvmlsEpoch mlsMeta
+      groupId = cnvmlsGroupId mlsMeta
+      suite = cnvmlsCipherSuite mlsMeta
```
mdimjasevic, line 611: This now allows to drop the `Member (ErrorS 'ConvNotFound) r` constraint from this function: it is not thrown directly here anymore, nor `applyProposalRef` needs to throw it because it does have all the MLS metadata needed.

services/galley/src/Galley/API/MLS/Message.hs
```diff
@@ -669,13 +671,13 @@ processExternalCommit ::
...
   Local Data.Conversation ->
+  ConversationMLSData ->
   ClientMap ->
   Epoch ->
-  GroupId ->
   ProposalAction ->
   Maybe UpdatePath ->
   Sem r ()
-processExternalCommit qusr mSenderClient lconv cm epoch groupId action updatePath = withCommitLock groupId epoch $ do
+processExternalCommit qusr mSenderClient lconv mlsMeta cm epoch action updatePath = withCommitLock (cnvmlsGroupId mlsMeta) epoch $ do
```
mdimjasevic, line 679: The `ConvNotFound` error constraint likely isn't needed here either.

services/galley/src/Galley/API/MLS/Message.hs
```diff
@@ -776,17 +778,17 @@ processCommitWithAction ::
...
+  ConversationMLSData ->
   ClientMap ->
   Epoch ->
-  GroupId ->
   ProposalAction ->
   Sender 'MLSPlainText ->
   Commit ->
   Sem r [LocalConversationUpdate]
-processCommitWithAction qusr senderClient con lconv cm epoch groupId action sender commit =
+processCommitWithAction qusr senderClient con lconv mlsMeta cm epoch action sender commit =
```
mdimjasevic, line 787: Can you check if `ConvNotFound` can be removed from this one too?

services/galley/src/Galley/API/MLS/Removal.hs
```diff
@@ -110,9 +110,10 @@ removeClient ::
   ClientId ->
   Sem r ()
 removeClient lc qusr cid = do
```
mdimjasevic, line 112: Can you see if you can change the signature of the function such that you don't need to do `for_` in the function? You know you do this only for MLS clients, so it feels a bit awkward to have to check if this is an MLS conversation.

I think you could pass in `ConversationMLSData` and as little as needed from `Local Data.Conversation` instead of passing in the whole conversation.

mdimjasevic (approved):
Looks good! There are some questions I have though; see them inlined. In particular, if possible it'd be great to avoid conditional execution of various MLS actions if we're in a function that deals with MLS conversations only.

services/galley/src/Galley/API/Action.hs
```diff
@@ -325,7 +326,8 @@ performAction tag origUser lconv action = do
                       }
                 )
                 origUser
-      traverse_ (removeUser lconv') victims
+      for_ (Data.mlsMetadata (tUnqualified lconv')) $ \meta ->
+        traverse_ (removeUser lconv' (cnvmlsGroupId meta)) victims
```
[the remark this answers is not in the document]
smatting: No, the `removeUser` function is from the MLS modules. It only applies to MLS conversations, that's why it's okay to require the MLS metadata as argument

services/galley/src/Galley/API/MLS/Removal.hs
smatting: Downstream of `removeClient` is `propagateMessage` and `runMessagePush` which require the full conversation object. Maybe we could refactor them as well, but I woudn't want to mix it into this PR

services/galley/src/Galley/API/MLS/Message.hs
smatting: I could remove the `ConvNotFound` constraint from `checkEpoch`, but `checkGroup` still throws it. I refactored both functions to not used the converation object anymore but rather the `ConversationMLSData` in
https://github.com/wireapp/wire-server/pull/2859/commits/8e9278987a2a127391a9dbf952cdeaf9a1d04d48

services/galley/src/Galley/API/MLS/Message.hs
smatting: Unfortunately because of `checkGroup` it can't be removed

services/galley/src/Galley/API/MLS/Message.hs
smatting: Unfortunately it can't because of downstream `checkGroup`.

services/galley/src/Galley/API/MLS/Removal.hs
smatting: Fair point about inlining the `for_` part into `removeUser` itself. Did this is acf5d67bb0fc9e61753e927bb8f7751377a33c26

PR #2862, opened 2022-11-23
Improve global team conversation handling and self conversation creation error.

elland:
Removed global team conversation creator, improved tests, improve adding users.
Improves error thrown when creating the self conversation, 500 bad.

### DIFF
--- a/services/galley/src/Galley/API/MLS/Message.hs
+++ b/services/galley/src/Galley/API/MLS/Message.hs
@@ -793,20 +793,23 @@ processCommitWithAction qusr senderClient con lconv mlsMeta cm epoch action send
 processInternalCommit ::
   forall r.
   ( HasProposalEffects r,
-    Member (Error FederationError) r,
-    Member (Error InternalError) r,
-    Member (ErrorS 'ConvNotFound) r,
-    Member (ErrorS 'MLSClientSenderUserMismatch) r,
-    Member (ErrorS 'MLSCommitMissingReferences) r,
-    Member (ErrorS 'MLSMissingSenderClient) r,
-    Member (ErrorS 'MLSProposalNotFound) r,
-    Member (ErrorS 'MLSSelfRemovalNotAllowed) r,
-    Member (ErrorS 'MLSStaleMessage) r,
-    Member (ErrorS 'MissingLegalholdConsent) r,
-    Member (Input (Local ())) r,
-    Member ProposalStore r,
-    Member BrigAccess r,
-    Member Resource r
+    Members
+      [ Error FederationError,
+        Error InternalError,
+        ErrorS 'ConvNotFound,
+        ErrorS 'MLSClientSenderUserMismatch,
+        ErrorS 'MLSCommitMissingReferences,
+        ErrorS 'MLSMissingSenderClient,
+        ErrorS 'MLSProposalNotFound,
+        ErrorS 'MLSSelfRemovalNotAllowed,
+        ErrorS 'MLSStaleMessage,
+        ErrorS 'MissingLegalholdConsent,
+        Input (Local ()),
+        ProposalStore,
+        BrigAccess,
+        Resource
+      ]
+      r
   ) =>
   Qualified UserId ->
   Maybe ClientId ->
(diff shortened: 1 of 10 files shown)


services/galley/src/Galley/API/Query.hs
```diff
@@ -313,12 +318,24 @@ getConversationRoles lusr cnv = do
...
   Local UserId ->
   Maybe ConvId ->
   Maybe (Range 1 1000 Int32) ->
   Sem r (Public.ConversationList ConvId)
 conversationIdsPageFromUnqualified lusr start msize = do
+  void $
+    E.getUserTeams (tUnqualified lusr) >>= \tids ->
+      runError @InternalError $
+        runError @(Tagged 'NotATeamMember ())
+          (for_ tids $ \tid -> getGlobalTeamConversation lusr tid)
```
jschaul, line 337: Why leave a side effect in listing conversations; which, as a sideeffect of getting global team conversations, creates one?

Why not create the global team conversation during the creation of a *team*?

And if you'd like to enable this functionality for existing teams, you could add a one-off data migration as done in https://github.com/wireapp/wire-server/tree/develop/services/galley/migrate-data/src - read the list of teams, insert one record into conversations, and from then on all teams will have this.

pcapriotti: I think it would also be possible to use the approach you suggest, but after some discussion with other backend people, I'm not sure it's really worth it.

Ultimately, the "side effect" of creating the global team conversation is not really a side effect, since it's not observable from the outside. The small extra lookup needed when listing conversations doesn't seem like a huge problem, considering that listing conversations is already a slow operation.

On the other hand, preemptively creating all GTCs (and for consistency, all MLS self conversations) forces the creation of all these extra conversations for inactive teams and users. I admit that this is probably not a big concern.

Therefore, we decided to stick with the current approach for now. I'm happy to discuss this further if you still think it's a mistake.

services/galley/src/Galley/API/Query.hs
```diff
@@ -314,12 +319,24 @@ getConversationRoles lusr cnv = do
...
+    ]
+    r =>
   Local UserId ->
   Maybe ConvId ->
   Maybe (Range 1 1000 Int32) ->
   Sem r (Public.ConversationList ConvId)
 conversationIdsPageFromUnqualified lusr start msize = do
+  void $
+    E.getUserTeams (tUnqualified lusr) >>= \tids ->
+      runError @InternalError $
```
smatting, line 336: Can the `InternalError` member from `getGlobalTeamConversation` not be dropped? Looks like it's just doing DB access

services/galley/src/Galley/API/MLS/Util.hs
```diff
@@ -123,3 +90,30 @@ getPendingBackendRemoveProposals gid epoch = do
...
+  let mlsData = gtcMlsMetadata gtc
+   in Conversation
+        { convId = qUnqualified $ gtcId gtc,
+          convLocalMembers = lm,
+          convRemoteMembers = mempty,
+          convDeleted = False,
+          convMetadata =
+            ConversationMetadata
+              { cnvmType = GlobalTeamConv,
+                -- FUTUREWORK: Make this a qualified user ID.
```
smatting, line 109: That FUTUREWORK is proably better placed at the definition of `ConversationMetadata`

services/galley/src/Galley/API/MLS/Message.hs
```diff
@@ -843,11 +863,9 @@ processInternalCommit qusr senderClient con lconv mlsMeta cm epoch action sender
...
                 (Set.singleton (creatorClient, creatorRef))
             (Left _, SelfConv, _) ->
-              throw . InternalErrorWithDescription $
-                "Unexpected creator client set in a self-conversation"
-            -- this is a newly created conversation, and it should contain exactly one
-            -- client (the creator)
-
+              -- this is a newly created conversation, and it should contain exactly one
+              -- client (the creator)
+              throwS @'MLSUnexpectedSenderClient
```
smatting, line 868: This updates error retporting of the self conversation, not the GTC. Can you please update PR description / title?

smatting (approved):
LGTM safe for some minor things

services/galley/src/Galley/API/MLS/Util.hs
elland: It is already there. Copy paste remnants :)

PR #2865, opened 2022-11-24
hlint: fix inplace option, less verbosity on ci

smatting:
This PR fixes some issues of the `./tools/hlint.sh` script

- adds the missing `refactor` binary to nix. Without the `-m inplace` option doesn't work
- adds missing implementation for the `-f all -m inplace` combination
- Removes verbosity for `-f all -m check`. This prevents a wall of text on our CI and makes it easier to read the suggestions of hlint

### DIFF
--- a/nix/wire-server.nix
+++ b/nix/wire-server.nix
@@ -300,6 +300,7 @@ let
     pkgs.helm
     pkgs.helmfile
     pkgs.hlint
+    ( hlib.justStaticExecutables pkgs.haskellPackages.apply-refact )
     pkgs.jq
     pkgs.kubectl
     pkgs.nixpkgs-fmt
--- a/tools/hlint.sh
+++ b/tools/hlint.sh
@@ -4,14 +4,13 @@
 usage() { echo "Usage: $0 -f [all, changeset] -m [check, inplace]" 1>&2; exit 1; }
 
 files=''
-check=true
 
 while getopts ':f:m:k' opt
  do
      case $opt in
          f) f=${OPTARG}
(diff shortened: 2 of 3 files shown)


nix/wire-server.nix
```diff
@@ -300,6 +300,7 @@ let
     pkgs.helm
     pkgs.helmfile
     pkgs.hlint
+    ( hlib.justStaticExecutables pkgs.haskellPackages.apply-refact )
```
elland, line 303: 🫶

tools/hlint.sh
```diff
@@ -22,9 +21,9 @@ while getopts ':f:m:k' opt
             ;;
          m) m=${OPTARG}
             if [ "$m" = "inplace" ]; then
-              check=false
+            :
```
elland, line 24: What does this do? Is this just a no-op?

tools/hlint.sh
```diff
@@ -43,15 +42,15 @@ if [ "${k}" ]; then
...
+if [ "$f" = "all" ] && [ "$m" = "check" ]; then
+  hlint -g
 else
   count=$(echo "$files" | grep -c -v -e '^[[:space:]]*$')
   echo "Analysing $count file(s)…"
   for f in $files
   do
     echo "$f"
-    if [ $check = true ]; then
+    if [ "$m" = "check" ]; then
```
elland, line 53: 👍

tools/hlint.sh
smatting: Yeah, that's the weird syntax for no-op in bash

PR #2866, opened 2022-11-24
Speed up "make formatf"

smatting:
This PR speeds up `make formatf` (not used on CI) by only restricting to files that
- are changed in the branch compared to `origin/develop`
- are changed compared to HEAD

This PR adds `make formatf-all` target that behaves like the the `make formatf` without this PR.


elland (approved):
Looks fine by me but I'm not a bash expert.

PR #2872, opened 2022-11-25
[FS-1249] Do Not List MLS Self-conversation in client API v1 and v2

mdimjasevic:
Currently the call to `POST /conversation/list-ids` will list the MLS self-conversation in client API v1 and v2 if a v3 client created the conversation. Filter out the conversation in versions below v3.

Tracked by https://wearezeta.atlassian.net/browse/FS-1249.

## Checklist

### DIFF
--- a/services/galley/src/Galley/API/MLS/Types.hs
+++ b/services/galley/src/Galley/API/MLS/Types.hs
@@ -19,6 +19,7 @@ module Galley.API.MLS.Types
   ( ClientMap,
     mkClientMap,
     cmAssocs,
+    ListGlobalSelfConvs (..),
   )
 where
 
@@ -41,3 +42,9 @@ mkClientMap = foldr addEntry mempty
 
 cmAssocs :: ClientMap -> [(Qualified UserId, (ClientId, KeyPackageRef))]
 cmAssocs cm = Map.assocs cm >>= traverse toList
+
+-- | Inform a handler for 'POST /conversations/list-ids' if the MLS global team
+-- conversation and the MLS self-conversation should be included in the
+-- response.
+data ListGlobalSelfConvs = ListGlobalSelf | DoNotListGlobalSelf
+  deriving (Eq)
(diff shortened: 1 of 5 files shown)


services/galley/src/Galley/API/MLS/Types.hs
```diff
@@ -41,3 +42,9 @@ mkClientMap = foldr addEntry mempty
 
 cmAssocs :: ClientMap -> [(Qualified UserId, (ClientId, KeyPackageRef))]
 cmAssocs cm = Map.assocs cm >>= traverse toList
+
+-- | Inform a handler for 'POST /conversations/list-ids' if the MLS global team
+-- conversation and the MLS self-conversation should be included in the
+-- response.
+data ListGlobalSelfConvs = ListGlobalSelf | DoNotListGlobalSelf
```
elland, line 49: I think it makes sense to generalise this to the implicit MLS conversations and re-use it for the global team conv as well.

services/galley/src/Galley/API/Query.hs
```diff
@@ -398,6 +406,22 @@ conversationIdsPageFromV2 lusr Public.GetMultiTablePageRequest {..} = do
           mtpPagingState = Public.ConversationPagingState table (LBS.toStrict . C.unPagingState <$> pwsState)
         }
 
+    -- MLS self-conversation of this user
+    selfConvId = mlsSelfConvId (tUnqualified lusr)
+    isNotSelfConv = (/= selfConvId) . qUnqualified
```
elland, line 411: How about the global team conversation?

elland (commented):
Should I take over to add the GTC filter?

services/galley/src/Galley/API/Query.hs
mdimjasevic: I'd suggest to do that in a separate PR. I suppose it won't be hard to add it on top of this (assuming it gets merged), but you would know best.

smatting (approved):

PR #2875, opened 2022-11-28
Filter out global team conversations for API v1 and v2.

elland:
Builds upon #2872 

## Checklist

### DIFF
--- a/services/galley/src/Galley/API/Query.hs
+++ b/services/galley/src/Galley/API/Query.hs
@@ -360,10 +360,11 @@ conversationIdsPageFromV2 ::
   ( p ~ CassandraPaging,
     Members
       '[ ConversationStore,
+         TeamStore,
          Error InternalError,
-         Input Env,
          ListItems p ConvId,
          ListItems p (Remote ConvId),
+         Input Env,
          P.TinyLog
        ]
       r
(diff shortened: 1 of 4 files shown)


services/galley/src/Galley/API/Query.hs
```diff
@@ -415,11 +422,11 @@ conversationIdsPageFromV2 listGlobalSelf lusr Public.GetMultiTablePageRequest {.
     -- self-conversation.
     --
     -- FUTUREWORK: This is yet to be implemented for global team conversations.
```
smatting, line 424: this FUTUREWORK can be removed

services/galley/test/integration/API/MLS.hs
```diff
@@ -2210,6 +2211,42 @@ testGetGlobalTeamConv setup = do
...
+    let (Just gtc) = Aeson.decode rs :: Maybe GlobalTeamConversation
+        gid = cnvmlsGroupId $ gtcMlsMetadata gtc
+
+    void $ uploadNewKeyPackage alice1
+
+    -- create mls group
+    createGroup alice1 gid
+    void $ createAddCommit alice1 [] >>= sendAndConsumeCommitBundle
+
+  -- Now we should have the user as part of that conversation also in the backend
```
smatting, line 2245: this comment should read differently, I think

services/galley/src/Galley/API/Query.hs
```diff
@@ -358,34 +366,43 @@ conversationIdsPageFromV2 ::
   Sem r Public.ConvIdsPage
 conversationIdsPageFromV2 listGlobalSelf lusr Public.GetMultiTablePageRequest {..} = do
   let localDomain = tDomain lusr
+  gtcIds <-
+    E.getUserTeams (tUnqualified lusr) >>= mapM (pure . Data.globalTeamConv)
+
+  -- implicitly add user to the global team conv, so it gets listed.
+  for_ gtcIds $ \tid -> E.createMembers tid (UserList [tUnqualified lusr] mempty)
```
pcapriotti, line 389: I thought the idea was that the GTC never has any explicit members.

services/galley/test/integration/API/MLS.hs
elland: Other way around, it should list for this user, since we're calling the right version of the API, no?

pcapriotti (changes requested):
This doesn't look right to me. Since the normal listing function is not seeing the GTC, we can simply add it manually at the beginning (or the end) of the list we return, without the need for any special tricks.

PR #2877, opened 2022-11-28
Revert global team conversations PRs #2862 and #2753

smatting:
Reverts wireapp/wire-server#2862 and #2753


nix/pkgs/mls-test-cli/default.nix
```diff
@@ -15,8 +15,8 @@ rustPlatform.buildRustPackage rec {
   src = fetchFromGitHub {
     owner = "wireapp";
     repo = "mls-test-cli";
-    sha256 = "sha256-FjgAcYdUr/ZWdQxbck2UEG6NEEQLuz0S4a55hrAxUs4=";
-    rev = "82fc148964ef5baa92a90d086fdc61adaa2b5dbf";
+    sha256 = "sha256-/XQ/9oQTPkRqgMzDGRm+Oh9jgkdeDM1vRJ6/wEf2+bY=";
```
mdimjasevic, line 18: @smatting , it seems wrong to revert the version of `mls-test-cli` to use.

services/galley/test/unit/Test/Galley/Intra/User.hs
```diff
@@ -20,6 +20,7 @@
 
 module Test.Galley.Intra.User where
 
+-- import Debug.Trace (traceShow)
```
mdimjasevic, line 23: No need to bring this back.

mdimjasevic (approved):
Generally it looks good, but I think we can and should keep the latest `mls-test-cli` tool version used.

elland:
I'm wondering if it wouldn't be simpler/easier to just disable the endpoints for now? 🤔

elland:
We've decided to just comment out the endpoints for the release instead.

PR #2878, opened 2022-11-29
add initial delay seconds to brig liveness and readiness probes

amitsagtani97:
Ticket - https://wearezeta.atlassian.net/browse/SQPIT-491

Brig pods continuously restarts in case of liveness and readiness probes failing on the startup in the low compute resource environment.

To reproduce - 
- reduce the memory and cpu resource limits in charts/brig/values.yaml and deploy the chart.
- Components of brig will take time to start which is greater than the wait time for before running first liveness probe, hence the pod is marked as failed and restarted again and again.

Added a initial delay before running first liveness and readiness probe, which allows all the brig containers to start on a low resource environment before running the probes.

## Checklist

### DIFF
--- a/charts/brig/templates/deployment.yaml
+++ b/charts/brig/templates/deployment.yaml
@@ -138,11 +138,13 @@ spec:
               scheme: HTTP
               path: /i/status
               port: {{ .Values.service.internalPort }}
+            initialDelaySeconds: 30
           readinessProbe:
             httpGet:
               scheme: HTTP
               path: /i/status
               port: {{ .Values.service.internalPort }}
+            initialDelaySeconds: 30
           resources:
 {{ toYaml .Values.resources | indent 12 }}
         {{- if .Values.config.geoip.enabled }}
--- a/charts/brig/templates/deployment.yaml
+++ b/charts/brig/templates/deployment.yaml
@@ -133,18 +133,23 @@ spec:
           {{- end }} 
           ports:
             - containerPort: {{ .Values.service.internalPort }}
+          startupProbe:
+            httpGet:
+              scheme: HTTP
+              path: /i/status
+              port: {{ .Values.service.internalPort }}
+            failureThreshold: 6
+            periodSeconds: 5
           livenessProbe:
             httpGet:
               scheme: HTTP
               path: /i/status
               port: {{ .Values.service.internalPort }}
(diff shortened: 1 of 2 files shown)


charts/brig/templates/deployment.yaml
```diff
@@ -138,11 +138,13 @@ spec:
               scheme: HTTP
               path: /i/status
               port: {{ .Values.service.internalPort }}
+            initialDelaySeconds: 30
           readinessProbe:
             httpGet:
               scheme: HTTP
               path: /i/status
               port: {{ .Values.service.internalPort }}
+            initialDelaySeconds: 30
```
jschaul, line 147: The readinessProbe should not get the same initialDelaySeconds as it will ensure brig doesn't get traffic when not ready yet - this is correct behaviour, and the pod should not get traffic while still starting up, so please remove this line above.

charts/brig/templates/deployment.yaml
```diff
@@ -138,11 +138,13 @@ spec:
...
```
jschaul, line 141: I think your problem would be better solved with a startup probe, see also https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/

```
startupProbe:
  httpGet:
   ...
   failureThreshold: 6
   periodSeconds: 5
  ```
 
 That would wait for 30 seconds before moving over to the liveness probe which restarts brig.
 
 Overall, if /i/status fails, however, then this begs the question if the installation works correctly otherwise? If brig doesn't have enough resources likely users will not have adequate latencies either.
 
 Also, could it be that something else regarding networking is not working as it should, rather than this just being a resource problem?

jschaul (changes requested):

charts/brig/templates/deployment.yaml
amitsagtani97: Thank you, yes the startupProbe make more sense here, added it.

Yeah, there can be issues due to high network latencies as well, there is a argument "timeoutSeconds", which has default value of 1 second. Maybe, we can increase that as well for these probes.

jschaul (approved):

jschaul:
@amitsagtani97 PR looks good now; could you add one line to a changelog file (maybe under internal) then this PR is good to be (squash-) merged.

PR #2879, opened 2022-11-29
Commented out GTC for release.

elland:
We can revert this after.


smatting (approved):

pcapriotti (approved):
This is probably fine. It would also be ok to simply delete the code, and revert this PR afterwards.

PR #2887, opened 2022-11-30
[FS-1267] Patch issue with initial commit bundle throwing a global team conversation error

elland:
This fixes an edge case where a regular conversation would be mistaken for a GTC, adds a test case against regressions.

## Checklist


smatting (approved):

pcapriotti (commented):
I still think we should not be fetching the global team conversation unconditionally here, rather check if the ID matches the one of the global team conversation, and only fetch it then. Otherwise we're doing two almost identical queries for each conversation. Am I missing something?

PR #2888, opened 2022-11-30
removed obsolete changelogs

battermann:


mdimjasevic:
Do you want to rewrite the MLS changelog for the GET endpoint as part of this PR too? If so, I can do that quickly, just let me know.

PR #2892, opened 2022-12-01
[FS-1188] Enable the typing indicator in federated environments.

elland:
## Checklist


libs/wire-api-federation/src/Wire/API/Federation/API/Galley.hs
```diff
@@ -72,6 +73,16 @@ type GalleyApi =
     :<|> FedEndpoint "send-mls-commit-bundle" MessageSendRequest MLSMessageResponse
     :<|> FedEndpoint "query-group-info" GetGroupInfoRequest GetGroupInfoResponse
     :<|> FedEndpoint "on-client-removed" ClientRemovedRequest EmptyResponse
+    :<|> FedEndpoint "on-typing-indicator-updated" TypingDataUpdateRequest EmptyResponse
+
+data TypingDataUpdateRequest = TypingDataUpdateRequest
+  { tdurTypingData :: TypingData,
+    tdurConnection :: ConnId,
```
pcapriotti, line 80: You should not pass a connection ID as part of a federated request. Rather the function which is taking care of sending events should accept an optional connection ID, because there won't be one when the request originates from a remote backend.

pcapriotti (changes requested):
Left one comment for now. Will review the rest when this is fixed.

libs/wire-api-federation/src/Wire/API/Federation/API/Galley.hs
elland: I wasn't sure if this was needed over federation, thanks for clarifying. Removed / made optional as applicable.

libs/wire-api/src/Wire/API/Routes/Public/Galley/Conversation.hs
```diff
@@ -615,6 +614,7 @@ type ConversationAPI =
     :<|> Named
            "member-typing-unqualified"
            ( Summary "Sending typing notifications"
+               :> Until 'V2
```
pcapriotti, line 617: Should be `Until 'V3`.

services/galley/test/integration/API.hs
```diff
@@ -3665,25 +3670,115 @@ putReceiptModeWithRemotesOk = do
...
+  connectUsers alice (singleton bob)
+
+  conv <- decodeConvId <$> postO2OConv alice bob Nothing
+  lcnv <- qualifyLocal conv
+
+  WS.bracketR2 c alice bob $ \(wsAlice, wsBob) -> do
+    post
+      ( unversioned
+          . g
+          . paths ["conversations", toByteString' conv, "typing"]
```
pcapriotti, line 3693: You should add a `/v2` prefix.

services/galley/test/integration/API/MLS.hs
```diff
@@ -215,6 +217,11 @@ tests s =
           test s "listing conversations without MLS configured" testSelfConversationMLSNotConfigured,
           test s "attempt to add another user to a conversation fails" testSelfConversationOtherUser,
           test s "attempt to leave fails" testSelfConversationLeave
+        ],
+      testGroup
+        "Typing indicators"
+        [ test s "POST /federation/on-typing-indicator-updated : Update typing indicator by remote user" updateTypingIndicatorFromRemoteUser,
+          test s "POST /federation/on-typing-indicator-updated : Update typing indicator to remote user" updateTypingIndicatorToRemoteUser
```
pcapriotti, line 224: These should be moved to `API`, I think. You don't need to use MLS in these tests. Also, there is a missing case: remote conversation with a local user sending a typing notification.

services/galley/test/integration/API/Util.hs
```diff
@@ -3083,3 +3093,35 @@ createAndConnectUsers domains = do
...
+
+removeVersionPrefix :: ByteString -> Maybe ByteString
+removeVersionPrefix bs = do
+  let (x, s) = B8.splitAt 1 bs
+  guard (x == B8.pack "v")
+  (_, s') <- B8.readInteger s
+  pure (B8.tail s')
+
+-- | Note: Apply this function last when composing (Request -> Request) functions
+unversioned :: Request -> Request
```
pcapriotti, line 3119: You can use `view tsUnversionedGalley` to get galley without any version prefix. You shouldn't need to do any explicit path manipulation besides adding a prefix.

pcapriotti (commented):
Looks generally good. Some comments below.

libs/wire-api/src/Wire/API/Routes/Public/Galley/Conversation.hs
```diff
@@ -624,6 +624,18 @@ type ConversationAPI =
                :> ReqBody '[JSON] TypingData
                :> MultiVerb 'POST '[JSON] '[RespondEmpty 200 "Notification sent"] ()
            )
+    :<|> Named
+           "member-typing-qualified"
+           ( Summary "Sending typing notifications with domain"
```
mdimjasevic, line 629: I would leave out the "with domain" part. This is the default (and starting with V3 the only) endpoint, so no need to say that it takes a qualified ID; we don't do that in other endpoint descriptions.

services/galley/src/Galley/API/Util.hs
```diff
@@ -881,3 +881,28 @@ instance
     if err' == demote @e
       then throwS @e
       else rethrowErrors @effs @r err'
+
+isTyping ::
```
mdimjasevic, line 885: A comment separator (just like before the remote error section) would be nice to have.

services/galley/test/integration/API/MLS.hs
```diff
@@ -45,10 +45,14 @@ import Data.Singletons
...
 import qualified Data.Text as T
 import Data.Time
-import Federator.MockServer hiding (withTempMockFederator)
-import qualified Galley.Options as Opts
 -- import Galley.Data.Conversation
 -- import Galley.Options
+-- import Wire.API.MLS.CipherSuite
+-- import Wire.API.MLS.GlobalTeamConversation
+-- import Wire.API.MLS.Group
+-- import Wire.API.Team (teamCreator)
```
mdimjasevic, line 53: Left overs. Please remove.

services/galley/test/integration/API.hs
```diff
@@ -3665,25 +3673,113 @@ putReceiptModeWithRemotesOk = do
...
+  aliceL <- qualifyLocal alice
+  bobL <- qualifyLocal bob
+
   connectUsers alice (singleton bob)
+
   conv <- decodeConvId <$> postO2OConv alice bob Nothing
+  lcnv <- qualifyLocal conv
+
+  WS.bracketR2 c alice bob $ \(wsAlice, wsBob) -> do
+    -- alice to bob
```
mdimjasevic, line 3739: Shouldn't the comment on this line be swapped with the one on line 3754?

mdimjasevic (approved):
Generally looks good. The only part I'm not quite sure of is the unqualified user ID in the payload sent over to the remote backend. What does it correspond to? Is it the user ID of the user making the request or is it the user ID of the recipient of the notification?

A minor comment or two is inlined.

services/galley/test/integration/API.hs
elland: Good catch!

pcapriotti (approved):
LGTM.

PR #2898, opened 2022-12-02
[SQSERVICES-1773] Allow pagination for team search endpoints - `has_more`

battermann:
This is an extension of https://github.com/wireapp/wire-server/pull/2895 and adds the `has_more` field to the search result.

https://wearezeta.atlassian.net/browse/SQSERVICES-1773

## Checklist

 - [x] No new changelog entry needed as this already exists in the base PR

### DIFF
--- a/services/brig/src/Brig/User/Search/TeamUserSearch.hs
+++ b/services/brig/src/Brig/User/Search/TeamUserSearch.hs
@@ -50,12 +50,12 @@ teamUserSearch ::
   Range 1 500 Int32 ->
   Maybe PagingState ->
   m (SearchResult TeamContact)
-teamUserSearch tid mbSearchText mRoleFilter mSortBy mSortOrder (fromRange -> s) mPagingState = liftIndexIO $ do
+teamUserSearch tid mbSearchText mRoleFilter mSortBy mSortOrder (fromRange -> size) mPagingState = liftIndexIO $ do
   let (IndexQuery q f sortSpecs) = teamUserSearchQuery tid mbSearchText mRoleFilter mSortBy mSortOrder
   idx <- asks idxName
   let search =
         (ES.mkSearch (Just q) (Just f))
-          { ES.size = ES.Size (fromIntegral s),
+          { ES.size = ES.Size (fromIntegral size + 1),
             ES.sortBody = Just (fmap ES.DefaultSortSpec sortSpecs),
             ES.searchAfterKey = toSearchAfterKey =<< mPagingState
           }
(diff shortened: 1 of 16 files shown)


services/brig/src/Brig/User/Search/TeamUserSearch.hs
```diff
@@ -50,12 +50,12 @@ teamUserSearch ::
...
   Maybe PagingState ->
   m (SearchResult TeamContact)
-teamUserSearch tid mbSearchText mRoleFilter mSortBy mSortOrder (fromRange -> s) mPagingState = liftIndexIO $ do
+teamUserSearch tid mbSearchText mRoleFilter mSortBy mSortOrder (fromRange -> size) mPagingState = liftIndexIO $ do
   let (IndexQuery q f sortSpecs) = teamUserSearchQuery tid mbSearchText mRoleFilter mSortBy mSortOrder
   idx <- asks idxName
   let search =
         (ES.mkSearch (Just q) (Just f))
-          { ES.size = ES.Size (fromIntegral s),
+          { ES.size = ES.Size (fromIntegral size + 1),
```
mdimjasevic, line 58: Why is it + 1?

mdimjasevic (approved):
Please add a changelog. You can extend the one from the previous PR by listing both PR numbers, which should look nice in the release notes.

The part I didn't get is +1; a comment is inlined.

Furthermore, is there a way for the API user to iterate over result pages?

battermann:
> Please add a changelog. You can extend the one from the previous PR by listing both PR numbers, which should look nice in the release notes.
> 
> The part I didn't get is +1; a comment is inlined.

Good points, I'll add a comment about the +1 and update the changelog.
 
> Furthermore, is there a way for the API user to iterate over result pages?

Not sure what you mean exactly. This PR adds the `has_more` field. If it is true, it means that there is a next page. They can use the `paging_state` and include it in a request, and this will give them the next page.

mdimjasevic:
> 
> > Furthermore, is there a way for the API user to iterate over result pages?
> 
> Not sure what you mean exactly. This PR adds the `has_more` field. If it is true, it means that there is a next page. They can use the `paging_state` and include it in a request, and this will give them the next page.

Ok, that's what I was asking about. Thanks for clarifying.

Given the experience with the Cassandra paging endpoints, I expect client devs to be asking how to use this Es paging system. Perhaps it'd be good to include a thorough description in Swagger analogous to the way the Cassandra paging endpoints have it, by extending the endpoint description.

battermann:
> Given the experience with the Cassandra paging endpoints, I expect client devs to be asking how to use this Es paging system. Perhaps it'd be good to include a thorough description in Swagger analogous to the way the Cassandra paging endpoints have it, by extending the endpoint description.

I updated the docs, and tried to be more clear. From the client perspective, ES's and C*'s paging mechanisms works exactly the same. In other words, the API is the same (only the implementation differs). So I assume they are already fairly familiar with it. However, it can't hurt to be more clear.

PR #2902, opened 2022-12-05
Cleanup: Remove cassandra queries to the user_keys_hashed table

jschaul:
As they are never read anymore since 'onboarding' / auto-connect was removed in https://github.com/wireapp/wire-server/pull/1005. See [here](https://github.com/zinfra/brig/blob/bf98d664b813548c3929c3c9944b8fa76011e53f/brig/src/Brig/API/Connection.hs#L340) for old code with past usage.

## Checklist


tools/db/move-team/src/Work.hs
```diff
@@ -157,8 +157,6 @@ runFullScans env@Env {..} users = do
       .| mapC (filter (haveId . view _2))
 
   -- FUTUREWORK: no need to read this table, it can be populated from `brig.user`
```
battermann, line 159: maybe remove this comment as well?

battermann (approved):

tools/db/move-team/src/Work.hs
jschaul, line 159:
```suggestion
```

jschaul: 👍🏿

mdimjasevic:
The PR didn't run through the PR pipeline in CI. Is that a problem for this PR?

mdimjasevic (commented):
Is this to be reviewed only after July 2023?

jschaul:
> The PR didn't run through the PR pipeline in CI. Is that a problem for this PR?

it ran in the past but had a flaky integration test. I rebased and pushed... and CI doesn't seem to pick it up anymore.

jschaul:
CI ran after another commit.

PR #2906, opened 2022-12-07
Removed Global Team Conversation.

elland:
With this we can have a clean slate on `develop` to start the long lived `mls` branch.

### DIFF
--- a/services/galley/src/Galley/API/Util.hs
+++ b/services/galley/src/Galley/API/Util.hs
@@ -192,7 +192,7 @@ ensureActionAllowed action self = case isActionAllowed (fromSing action) (convMe
 ensureGroupConversation :: Member (ErrorS 'InvalidOperation) r => Data.Conversation -> Sem r ()
 ensureGroupConversation conv = do
   let ty = Data.convType conv
-  unless (ty `elem` [RegularConv]) $ throwS @'InvalidOperation
+  unless (ty == RegularConv) $ throwS @'InvalidOperation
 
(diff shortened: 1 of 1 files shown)


services/galley/src/Galley/Cassandra/Instances.hs
```diff
@@ -56,14 +56,12 @@ instance Cql ConvType where
...
   toCql One2OneConv = CqlInt 2
   toCql ConnectConv = CqlInt 3
-  toCql GlobalTeamConv = CqlInt 4
 
   fromCql (CqlInt i) = case i of
     0 -> pure RegularConv
     1 -> pure SelfConv
     2 -> pure One2OneConv
     3 -> pure ConnectConv
-    4 -> pure GlobalTeamConv
```
smatting, line 66: I think we need to keep this. There are currently already values in the DB with "4". If we don't keep this then processes will crash upon reading

smatting (changes requested):

mdimjasevic:
Could this PR have started off by introducing revert commits of earlier GTC PRs and then making changes on top? That would have made reviewing much easier.

smatting:
Closing this in favor of https://github.com/wireapp/wire-server/pull/2916
which is just reverts with this outcome
```
> 46 files changed, 265 insertions(+), 1033 deletions(-)
```
Compare to the changes from this branch
```
19 files changed, 47 insertions(+), 592 deletions(-)
```

PR #2919, opened 2022-12-12
Removed typing data wrapper type.

elland:
Should not impact clients.


pcapriotti (approved):
Looks good, but make sure integration tests run in CI before merging.

PR #2937, opened 2022-12-19
FS-902: Fix empty client set & set subconv field in message propagation

smatting:
This PR:

- Fixes a bug: on the first commit the creator's client is not added to member clients of the subconversation
- MLS message are propagated with the `subconv` field populated

## Checklist

### DIFF
--- a/services/galley/src/Galley/API/MLS/Message.hs
+++ b/services/galley/src/Galley/API/MLS/Message.hs
@@ -111,7 +111,8 @@ type MLSMessageStaticErrors =
      ErrorS 'MLSSelfRemovalNotAllowed,
      ErrorS 'MLSClientSenderUserMismatch,
      ErrorS 'MLSGroupConversationMismatch,
-     ErrorS 'MLSMissingSenderClient
+     ErrorS 'MLSMissingSenderClient,
+     ErrorS 'MLSSubConvClientNotInParent
    ]
 
 type MLSBundleStaticErrors =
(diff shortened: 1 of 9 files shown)


elland:
LGTM 🥳

services/galley/src/Galley/API/MLS/Message.hs
```diff
@@ -893,9 +901,24 @@ processInternalCommit qusr senderClient con lConvOrSub epoch action senderRef co
...
+                  (pure senderRef)
+                  ( note (mlsProtocolError "Could not compute key package ref")
+                      . kpRef'
+                      . upLeaf
+                  )
+                  $ cPath commit
+              addMLSClients
+                (cnvmlsGroupId mlsMeta)
+                qusr
+                (Set.singleton (creatorClient, creatorRef))
```
mdimjasevic, line 919: We've had this pattern of validation and addition before, so perhaps it'd be good to extract it into a function, maybe in a separate PR.

pcapriotti (approved):
Looks good, but since you introduced a new error for this, we should be using it also when validating an external commit (line 748).

mdimjasevic (approved):

smatting:
> Looks good, but since you introduced a new error for this, we should be using it also when validating an external commit (line 748).

I'll add it in separate PR

services/galley/src/Galley/API/MLS/Message.hs
smatting: The validation part I could factor out I think: https://github.com/wireapp/wire-server/pull/2945

PR #2940, opened 2022-12-19
Track federated calls

isovector:
As part of gracefully handling offline backends, we want to make sure we have all of our bases covered when it comes to tracking what changes need to happen on the federation side. But it's unclear which endpoints call which federated endpoints. 

This PR tracks federated calls in the type system, ensuring all code paths (including endpoints) document which federated calls they make. For the time being, these constraints get eliminated by orphan instances at the very edge of the APIs, but #2950 follows up on this change, adding a servant combinator for better dissolution of the new `CallsFed` constraints, ensuring they get documented in swagger. 

The relevant change is in `Wire.API.Federation.API`, adding `CallsFed`. Everything else was done automatically in a tireless cycle of asking HLS to add a necessary constraint until the project compiled again.

## Checklist

### DIFF
--- a/libs/wire-api-federation/src/Wire/API/Federation/API.hs
+++ b/libs/wire-api-federation/src/Wire/API/Federation/API.hs
@@ -53,6 +53,8 @@ type HasFedEndpoint comp api name = ('Just api ~ LookupEndpoint (FedApi comp) na
 
 class CallsFed (comp :: Component) (name :: Symbol)
 
+instance CallsFed comp name
+
 -- | Return a client for a named endpoint.
 fedClient ::
   forall (comp :: Component) (name :: Symbol) m api.
--- a/libs/wire-api-federation/src/Wire/API/Federation/API.hs
+++ b/libs/wire-api-federation/src/Wire/API/Federation/API.hs
@@ -20,6 +20,7 @@ module Wire.API.Federation.API
     HasFedEndpoint,
     fedClient,
     fedClientIn,
+    CallsFed,
 
     -- * Re-exports
     Component (..),
@@ -48,12 +49,14 @@ type instance FedApi 'Brig = BrigApi
 
 type instance FedApi 'Cargohold = CargoholdApi
 
-type HasFedEndpoint comp api name = ('Just api ~ LookupEndpoint (FedApi comp) name)
+type HasFedEndpoint comp api name = ('Just api ~ LookupEndpoint (FedApi comp) name, CallsFed comp name)
+
+class CallsFed (comp :: Component) (name :: Symbol)
 
 -- | Return a client for a named endpoint.
 fedClient ::
   forall (comp :: Component) (name :: Symbol) m api.
-  (HasFedEndpoint comp api name, HasClient m api, m ~ FederatorClient comp) =>
+  (CallsFed comp name, HasFedEndpoint comp api name, HasClient m api, m ~ FederatorClient comp) =>
   Client m api
 fedClient = clientIn (Proxy @api) (Proxy @m)
 
--- a/libs/wire-api-federation/src/Wire/API/Federation/API.hs
+++ b/libs/wire-api-federation/src/Wire/API/Federation/API.hs
@@ -53,8 +53,6 @@ type HasFedEndpoint comp api name = ('Just api ~ LookupEndpoint (FedApi comp) na
 
 class CallsFed (comp :: Component) (name :: Symbol)
 
-instance CallsFed comp name
-
 -- | Return a client for a named endpoint.
 fedClient ::
   forall (comp :: Component) (name :: Symbol) m api.
--- a/services/brig/src/Brig/API/Internal.hs
+++ b/services/brig/src/Brig/API/Internal.hs
@@ -87,6 +87,7 @@ import UnliftIO.Async
 import Wire.API.Connection
 import Wire.API.Error
 import qualified Wire.API.Error.Brig as E
+import Wire.API.Federation.API
 import Wire.API.MLS.Credential
 import Wire.API.MLS.KeyPackage
 import Wire.API.MLS.Serialisation
@@ -100,18 +101,19 @@ import Wire.API.User.Activation
 import Wire.API.User.Client
 import Wire.API.User.Password
 import Wire.API.User.RichInfo
-import Wire.API.Federation.API
 
 ---------------------------------------------------------------------------
 -- Sitemap (servant)
 
 servantSitemap ::
-  (Members
-    '[ BlacklistStore,
-       GalleyProvider,
-       UserPendingActivationStore p
-     ]
-    r, CallsFed 'Brig "on-user-deleted-connections") =>
+  ( Members
+      '[ BlacklistStore,
+         GalleyProvider,
+         UserPendingActivationStore p
+       ]
+      r,
+    CallsFed 'Brig "on-user-deleted-connections"
+  ) =>
   ServerT BrigIRoutes.API (Handler r)
 servantSitemap =
   ejpdAPI
@@ -151,12 +153,14 @@ mlsAPI =
     :<|> Named @"put-key-package-add" upsertKeyPackage
 
 accountAPI ::
-  (Members
-    '[ BlacklistStore,
-       GalleyProvider,
-       UserPendingActivationStore p
-     ]
-    r, CallsFed 'Brig "on-user-deleted-connections") =>
+  ( Members
+      '[ BlacklistStore,
+         GalleyProvider,
+         UserPendingActivationStore p
+       ]
+      r,
+    CallsFed 'Brig "on-user-deleted-connections"
+  ) =>
   ServerT BrigIRoutes.AccountAPI (Handler r)
 accountAPI =
   Named @"createUserNoVerify" createUserNoVerify
@@ -285,15 +289,17 @@ swaggerDocsAPI = swaggerSchemaUIServer BrigIRoutes.swaggerDoc
 -- Sitemap (wai-route)
 
 sitemap ::
-  (Members
-    '[ CodeStore,
-       PasswordResetStore,
-       BlacklistStore,
-       BlacklistPhonePrefixStore,
-       GalleyProvider,
-       UserPendingActivationStore p
-     ]
-    r, CallsFed 'Brig "on-user-deleted-connections") =>
+  ( Members
+      '[ CodeStore,
+         PasswordResetStore,
+         BlacklistStore,
+         BlacklistPhonePrefixStore,
+         GalleyProvider,
+         UserPendingActivationStore p
+       ]
+      r,
+    CallsFed 'Brig "on-user-deleted-connections"
+  ) =>
   Routes a (Handler r) ()
 sitemap = do
   get "/i/status" (continue $ const $ pure empty) true
@@ -457,10 +463,12 @@ sitemap = do
 
 -- | Add a client without authentication checks
 addClientInternalH ::
-  (Members
-    '[ GalleyProvider
-     ]
-    r, CallsFed 'Brig "on-user-deleted-connections") =>
+  ( Members
+      '[ GalleyProvider
+       ]
+      r,
+    CallsFed 'Brig "on-user-deleted-connections"
+  ) =>
   UserId ::: Maybe Bool ::: JsonRequest NewClient ::: Maybe ConnId ::: JSON ->
   (Handler r) Response
 addClientInternalH (usr ::: mSkipReAuth ::: req ::: connId ::: _) = do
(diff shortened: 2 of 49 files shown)


isovector:
@julialongtin @fisx

libs/wire-api-federation/src/Wire/API/Federation/API.hs
```diff
@@ -48,12 +49,14 @@ type instance FedApi 'Brig = BrigApi
 
 type instance FedApi 'Cargohold = CargoholdApi
 
-type HasFedEndpoint comp api name = ('Just api ~ LookupEndpoint (FedApi comp) name)
+type HasFedEndpoint comp api name = ('Just api ~ LookupEndpoint (FedApi comp) name, CallsFed comp name)
+
+class CallsFed (comp :: Component) (name :: Symbol)
```
mdimjasevic, line 54: I'm yet to look at this thoroughly, but isn't `CallsFed comp name` the same as

```haskell
type CallsFed comp name = forall e. HasFedEndpoint comp e name
```

that is, so long as there exists a federation API endpoint `e` by name `name` in component `comp` (Galley, Brig or Cargohold), `CallsFed comp name` holds?

pcapriotti: No, `forall` is the universal quantifier ∀. So `forall e. X e` means that the constraint `X` holds for all possible values of `e`. What you are probably thinking about here is `exists` aka `∃`, which haskell doesn't have for whatever reason. You can emulate `exists e. X e` using "impredicative encoding"/CPS/Church encoding as follows:
```haskell
data ExistsX = forall e. ExistsX (X e)
```
This means that `ExistsX` is equipped with a constructor of type `ExistsX :: forall e. X e → ExistsX`, which by the adjunction relating `exists` and `forall`, is equivalent to `(exists e. X e) → ExistsX`.

In any case, I don't think this is very realistic. For one, if I'm not mistaken, existentially quantified constraints are simply not supported. Even if they were, I doubt you would get any sort of reasonable type inference out of them, so you'd still have to specify them manually. Haskell is just not the right language for encoding logic in types.

supersven:
Hey @isovector :wave: 

Please fix this linter issue:
```
+ make lint-all

./tools/ormolu.sh -c

ormolu mode: check

language extensions: AllowAmbiguousTypes BangPatterns ConstraintKinds DataKinds DefaultSignatures DerivingStrategies DerivingVia DeriveFunctor DeriveGeneric DeriveLift DeriveTraversable EmptyCase FlexibleContexts FlexibleInstances FunctionalDependencies GADTs InstanceSigs KindSignatures LambdaCase MultiParamTypeClasses MultiWayIf NamedFieldPuns NoImplicitPrelude OverloadedStrings PackageImports PatternSynonyms PolyKinds QuasiQuotes RankNTypes ScopedTypeVariables StandaloneDeriving TupleSections TypeApplications TypeFamilies TypeFamilyDependencies TypeOperators UndecidableInstances ViewPatterns

Checking 1382 file(s)…

libs/api-client/src/Network/Wire/Client/Session.hs...  ok

libs/brig-types/src/Brig/Types/User.hs...  ok

libs/gundeck-types/src/Gundeck/Types/Push/V2.hs...  ok

libs/types-common-aws/src/Util/Test/SQS.hs...  ok

libs/wire-api/test/unit/Test/Wire/API/User/Search.hs...  ok

services/proxy/src/Proxy/Run.hs...  ok

tools/stern/src/Stern/Types.hs...  ok

./tools/hlint.sh -f all -m check

services/brig/src/Brig/Run.hs:123:10-23: Warning: Redundant bracket

Found:

  (HasCallStack)

Perhaps:

  HasCallStack


1 hint
```

If I got this right, the general approach is implemented and discussed in https://github.com/wireapp/wire-server/pull/2950 which is still a *draft*. I'm wondering if we should merge this PR, before we agreed upon the general approach. My spontaneous answer would be: No.

Would you mind if we don't merge this PR now, but wait until https://github.com/wireapp/wire-server/pull/2950 is ready and merge them at the same time? I'm very sorry, but I'm hesitant to merge a PR that doesn't add functionality (as that's delivered with https://github.com/wireapp/wire-server/pull/2950.) Please, correct me if I'm wrong: I haven't had time to take a deep dive into this.

libs/wire-api-federation/src/Wire/API/Federation/API.hs
isovector: >  so long as there exists a federation API endpoint e by name name in component comp (Galley, Brig or Cargohold), CallsFed comp name holds?

The goal of `CallsFed`  is that it *doesn't have any instances,* ie, it shouldn't be dischargable. This forces all downstream callers to also add the constraint, so that we can guarantee it makes it down to the servant `Server` implementation for the API. I'll add some commentary to that effect.

isovector:
> @supersven  If I got this right, the general approach is implemented and discussed in https://github.com/wireapp/wire-server/pull/2950 which is still a draft. I'm wondering if we should merge this PR, before we agreed upon the general approach.

#2950 is done, just marked as a draft because it depends on this one. We can merge only #2950 if you'd like. My thought was because this PR is so big  it's likely to cause lots of merge conflicts the longer it goes unlanded, but #2950 came out faster than I was expecting. Your call.

supersven:
> https://github.com/wireapp/wire-server/pull/2950 is done, just marked as a draft because it depends on this one. We can merge only https://github.com/wireapp/wire-server/pull/2950 if you'd like. My thought was because this PR is so big it's likely to cause lots of merge conflicts the longer it goes unlanded, but https://github.com/wireapp/wire-server/pull/2950 came out faster than I was expecting. Your call.

Ah, I see. :bulb: I think it would be good to wait until the reviewers of https://github.com/wireapp/wire-server/pull/2950 agree that this is the way to go and the name and structure of `CallsFed` is stable. (Only that. Details may be left aside.)

However, I'll be on vacation during the whole next week. So, please feel free to make other plans with your PR's current reviewer(s).

services/brig/src/Brig/API/Internal.hs
```diff
@@ -105,12 +106,14 @@ import Wire.API.User.RichInfo
...
-       UserPendingActivationStore p
-     ]
-    r =>
+  ( Members
+      '[ BlacklistStore,
+         GalleyProvider,
+         UserPendingActivationStore p
+       ]
+      r,
+    CallsFed 'Brig "on-user-deleted-connections"
```
mdimjasevic, line 115: I'd expect the constraint to be gone by now, i.e., exactly at the point where endpoints are served. Has the constraint made it to here by mistake?

services/brig/src/Brig/API/Internal.hs
```diff
@@ -284,15 +289,17 @@ swaggerDocsAPI = swaggerSchemaUIServer BrigIRoutes.swaggerDoc
...
+  ( Members
+      '[ CodeStore,
+         PasswordResetStore,
+         BlacklistStore,
+         BlacklistPhonePrefixStore,
+         GalleyProvider,
+         UserPendingActivationStore p
+       ]
+      r,
+    CallsFed 'Brig "on-user-deleted-connections"
```
mdimjasevic, line 301: ditto

services/brig/src/Brig/API/Public.hs
```diff
@@ -167,20 +168,31 @@ versionedSwaggerDocsAPI Nothing = versionedSwaggerDocsAPI (Just maxBound)
...
+    CallsFed 'Brig "get-user-by-handle",
+    CallsFed 'Brig "get-users-by-ids",
+    CallsFed 'Brig "search-users",
+    CallsFed 'Brig "claim-key-packages",
+    CallsFed 'Brig "on-user-deleted-connections",
+    CallsFed 'Brig "claim-multi-prekey-bundle",
+    CallsFed 'Brig "send-connection-action",
+    CallsFed 'Brig "claim-prekey",
+    CallsFed 'Brig "claim-prekey-bundle",
+    CallsFed 'Brig "get-user-clients"
```
mdimjasevic, line 194: ditto

services/brig/src/Brig/Team/API.hs
```diff
@@ -97,12 +98,14 @@ servantAPI =
...
-       UserPendingActivationStore p
-     ]
-    r =>
+  ( Members
+      '[ BlacklistStore,
+         GalleyProvider,
+         UserPendingActivationStore p
+       ]
+      r,
+    CallsFed 'Brig "on-user-deleted-connections"
```
mdimjasevic, line 107: Shouldn't the constraint be gone by now?

services/cargohold/src/CargoHold/API/Public.hs
```diff
@@ -35,11 +35,12 @@ import Servant.API
 import Servant.Server hiding (Handler)
 import URI.ByteString
 import Wire.API.Asset
+import Wire.API.Federation.API
 import Wire.API.Routes.AssetBody
 import Wire.API.Routes.Internal.Cargohold
 import Wire.API.Routes.Public.Cargohold
 
-servantSitemap :: ServerT ServantAPI Handler
+servantSitemap :: (CallsFed 'Cargohold "get-asset", CallsFed 'Cargohold "stream-asset") => ServerT ServantAPI Handler
```
mdimjasevic, line 43: ditto

services/galley/src/Galley/API/Federation.hs
```diff
@@ -100,7 +100,18 @@ import Wire.API.ServantProto
...
+federationSitemap ::
+  ( CallsFed 'Galley "on-conversation-updated",
+    CallsFed 'Galley "on-mls-message-sent",
+    CallsFed 'Brig "get-mls-clients",
+    CallsFed 'Galley "on-new-remote-conversation",
+    CallsFed 'Galley "send-mls-message",
+    CallsFed 'Galley "mls-welcome",
+    CallsFed 'Galley "send-mls-commit-bundle",
+    CallsFed 'Galley "on-message-sent",
+    CallsFed 'Brig "get-user-clients"
```
mdimjasevic, line 112: Shouldn't the constraints be gone by now?

services/galley/src/Galley/API/Internal.hs
```diff
@@ -112,6 +113,8 @@ import Wire.API.Team.SearchVisibility
 import Wire.Sem.Paging
 import Wire.Sem.Paging.Cassandra
 
+instance CallsFed comp name
```
mdimjasevic, line 116: Can this instance be moved to somewhere further up, e.g., to `Galley.Run`?

services/galley/src/Galley/API/Public/Servant.hs
```diff
@@ -28,10 +28,29 @@ import Galley.API.Public.Team
...
+    CallsFed 'Galley "on-message-sent",
+    CallsFed 'Galley "on-new-remote-conversation",
+    CallsFed 'Galley "on-typing-indicator-updated",
+    CallsFed 'Galley "send-message",
+    CallsFed 'Brig "get-mls-clients",
+    CallsFed 'Galley "query-group-info",
+    CallsFed 'Galley "mls-welcome",
+    CallsFed 'Galley "update-conversation",
+    CallsFed 'Galley "send-mls-commit-bundle",
+    CallsFed 'Galley "send-mls-message"
```
mdimjasevic, line 51: Why they haven't dissolved by now?

mdimjasevic (approved):
Thanks for explaining what I didn't get at the beginning of reviewing!

There's one thing I thought would happen, but I don't see it in PR: shouldn't the `CallsFed` constraint dissolve when you are combining endpoint handlers into Servant `ServerT` in sitemaps (and similarly for wai-routes)?

Overall, this looks good and I'm approving. If I got the thing above wrong, feel free to merge right away.

isovector:
> There's one thing I thought would happen, but I don't see it in PR: shouldn't the CallsFed constraint dissolve when you are combining endpoint handlers into Servant ServerT in sitemaps (and similarly for wai-routes)?

You're right, nice catch. I'll get that cleaned up today.

isovector:
After 30 minutes of looking into this, it's surprisingly hard to do better here. GHC rejects any attempts to automatically solve all `CallsFed` constraints locally without giving an instance. If we give an instance higher up in Brig, we now require a separate module to contain the orphan instance, since we can't give the same orphan instance in the two files that would need it. Seeing as this all is going to get torn out immediately in #2950 when we introduce the servant combinators, I don't think it's worth putting more energy in here.

Instead I'll add the changelog and land this PR, and we can get working on getting #2950 in sooner.

isovector:
@mdimjasevic Unless you disagree with my earlier comment, please merge this when you get a chance (I don't have the necessary permissions.)

mdimjasevic:
> @mdimjasevic Unless you disagree with my earlier comment, please merge this when you get a chance (I don't have the necessary permissions.)

Merged! You can now proceed with rebasing #2950.

PR #2947, opened 2022-12-20
Upgrade Polysemy to latest to fix OOM in Galley.Run

isovector:
This PR applies a fix in Polysemy to dramatically cut down on  (useless) compile-time inlining. 

Currently marked as draft to see if it does in fact fix CI. If so, I'll release a new version of polysemy directly rather than depend on a pinned commit.

**Update**

@mdimjasevic and @pcapriotti ran a local memory usage comparison. This PR does seem to help substantially: before the patch compiling galley was using 49584 MB of RAM on @pcapriotti's system, compared to 34392 MB after the patch. It's still insane, but it's an improvement of about 44%.

## Checklist


isovector:
@smatting

mdimjasevic:
@isovector , do you have any observations when it comes to improvements in compilation performance that this PR brings?

isovector:
I don't; but this patch lets me build galley on my laptop with 16GB of RAM. Polysemy really doesn't do much that should trip up the optimizer with the exception of trying to inline all of its `Member` proofs, so this patch eliminates its desire to do that.

mdimjasevic:
> I don't; but this patch lets me build galley on my laptop with 16GB of RAM. Polysemy really doesn't do much that should trip up the optimizer with the exception of trying to inline all of its `Member` proofs, so this patch eliminates its desire to do that.

Just to confirm: was this with the O2 optimization level? That's the most challenging way of compiling we've been facing.

mdimjasevic:
Thanks Sandy for the PR! It significantly reduces the hardware requirements for building Wire server.

Now if you could make a Polysemy release with this so we can unpin it in Nix, that would be great!

PR #2950, opened 2022-12-22
MakesFederatedCall servant combinator

isovector:
This PR introduces a new servant combinator `MakesFederatedCall` which allows us to track federated calls across services, as part of the Offline Federation initiative. It also adds this combinator to all endpoints which make federated calls.

The idea here is that calls to federated clients emit a `CallsFed service endpoint` constraint which has no instances and thus can never be discharged under normal circumstances. The `MakesFederatedCall` servant combinator is the only means of eliminating this constraint, which thus forces developers to document their federated calls in the API in order to get things to compile.

In gory detail, we can eliminate a constraint `c` by coercing a `Dict c0` into a `Dict c` so long as `c0` and `c` are identical; see `synthesizeCallsFed` for the implementation. The `HasServer` implementation of `MakesFederatedCall` mangles the underlying handler adding a `Dict CallsFed ->` argument, which the API-definition site must then eliminate via `callsFed` . This is a slightly unfortunate encoding, as we'd prefer a `CallsFed =>` *constraint* instead, but Haskell has some limitations in what we can do with type families.

In addition, the `HasSwagger` implementation injects the federated calls into the `x-wire-makes-federated-calls-to` extension of the swagger `Operation`s. `swagger2` doesn't support extensions, but that work is upstreamed to https://github.com/GetShopTV/swagger2/pull/245, and thus the weird pin here. 
 

## Checklist


libs/wire-api/src/Wire/API/MakesFederatedCall.hs
```diff
@@ -0,0 +1,135 @@
...
+-- This program is distributed in the hope that it will be useful, but WITHOUT
+-- ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+-- FOR A PARTICULAR PURPOSE. See the GNU Affero General Public License for more
+-- details.
+--
+-- You should have received a copy of the GNU Affero General Public License along
+-- with this program. If not, see <https://www.gnu.org/licenses/>.
+{-# LANGUAGE OverloadedLists #-}
+
+module Wire.API.MakesFederatedCall
```
mdimjasevic, line 19: Doesn't this module belong to the `wire-api-federation` library instead?

mdimjasevic:
@isovector , there were compilation errors in integration tests in Brig so I fixed them. The problem was in calling `runFedClient` because a `CallsFed` constraint couldn't be inferred. I addressed that by prefixing it with `unsafeCallsFed @comp @name`. Before that I wanted to address the issue by defining:

```haskell
unsafeRunFedClient ::
  forall (name :: Symbol) comp api.
  ( HasFedEndpoint comp api name,
    Servant.HasClient Servant.ClientM api
  ) =>
  FedClient comp ->
  Domain ->
  Servant.Client Http api
unsafeRunFedClient fc domain = unsafeCallsFed @comp @name (runFedClient @name @comp @api)
```

but for reasons beyond me type inference doesn't do the job, i.e., the `CallsFed` constraint is not inferred. Can you figure out why?

It's a minor thing, but it'd be great if you could figure out this problem with type inference so we can have nicer call sites.

libs/wire-api/src/Wire/API/MakesFederatedCall.hs
isovector: Spiritually, but the package topography doesn't allow it to be so: `wire-api-federation` depends on `wire-api`, but `wire-api` is where all the APIs are defined.

isovector:
@mdimjasevic The problem is that `HasFedEndpoint` _itself_ propagates the `CallsFed` constraint. I've added a `HasUnsafeFedEndpoint` constraint that doesn't do so.

libs/wire-api/src/Wire/API/Routes/Public/Brig.hs
```diff
@@ -1177,7 +1219,7 @@ type SearchAPI =
              (SearchResult TeamContact)
     )
 
-type MLSAPI = LiftNamed (ZLocalUser :> "mls" :> MLSKeyPackageAPI)
+type MLSAPI = LiftNamed ("mls" :> MLSKeyPackageAPI)
```
mdimjasevic, line 1222: Doesn't this removal break the whole API?

And how is it related to this PR?

mdimjasevic (commented):
Everything looks good but that `MLSAPI` type change. How come it didn't cause the change of endpoint handler types? Dropping `ZLocalUser` would mean that these endpoints can be used by unauthenticated parties. Or am I missing something?

libs/wire-api/src/Wire/API/Routes/Public/Brig.hs
isovector: I inlined them into `MLSKeyPackageAPI` directly. The problem is that the `callsFed` combinator requires the `MakesFederatedCall` api modifications to come before any arguments that show up in the `ServerT` (of which `ZLocalUser` is one).

mdimjasevic: Thanks for the explanation! Somehow this went under radar when I was reviewing it.

mdimjasevic (approved):

mdimjasevic:
@isovector , I was about to merge, but I see you introduced a new thing an hour ago. Do you need more time to wrap it up? If not, you can ping anyone to merge this for you or I will merge it next time I check the PR.

isovector:
@mdimjasevic it's good for a final review/merge. Sorry for the bait and switch!

mdimjasevic:
@isovector , I don't see the `SolveCallsFed` class/constraint used anywhere, so maybe it shouldn't have been part of this PR. Anyhow, I'm merging it.

isovector:
@mdimjasevic 

> I don't see the `SolveCallsFed` class/constraint used anywhere

that's the point :) it immediately inlines and lets us replace the old `callsFed (callsFed (callsFed ...)))` --- one for each CallsFed --- with a single `callsFed`.

mdimjasevic:
> @mdimjasevic
> 
> > I don't see the `SolveCallsFed` class/constraint used anywhere
> 
> that's the point :) it immediately inlines and lets us replace the old `callsFed (callsFed (callsFed ...)))` --- one for each CallsFed --- with a single `callsFed`.

Wow! It will take some time for me to wrap my head around this.

PR #2953, opened 2022-12-27
hack/bin/upload-image: Retry despite `set -e`

akshaymankar:
Executing `"$@"` within first argument of `if` prevents `set -e` from immediately failing the whole script.

It could also be written as `while ! "$@"; do ...`, but then getting status of `"$@"` is more complicated as `! "$@"` has status=0 and overwrites the value of `$?`.

## Checklist

 - [x] No changelog

### DIFF
--- a/hack/bin/upload-image.sh
+++ b/hack/bin/upload-image.sh
@@ -35,29 +35,24 @@ function retry {
     local maxAttempts=$1
     local secondsDelay=1
     local attemptCount=1
-    local output=
     shift 1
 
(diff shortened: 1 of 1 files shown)


mdimjasevic (approved):
My Bash skills are rusty, but I'd say the PR is OK.

jschaul:
One could also do

```
set -e

# ... do stuff and if something fails fail the script

set +e

# ... do stuff which can fail but we don't care ....

set -e

# fail script if something fails
```

Which might be more readable (to me). But whatever works, really.

PR #2955, opened 2022-12-28
Use amazonka from upstream head intead of fork

akshaymankar:
https://github.com/brendanhay/amazonka/pull/832 should have solved our problems, so we shouldn't need the fork.

## Checklist

### DIFF
--- a/nix/haskell-pins.nix
+++ b/nix/haskell-pins.nix
@@ -93,9 +93,9 @@ let
     };
     amazonka = {
       src = fetchgit {
-        url = "https://github.com/wireapp/amazonka";
-        rev = "7ced54b0396296307b9871d293cc0ac161e5743d";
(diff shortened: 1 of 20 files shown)


stefanwire (approved):

nix/haskell-pins.nix
```diff
@@ -93,9 +93,9 @@ let
     };
     amazonka = {
       src = fetchgit {
-        url = "https://github.com/wireapp/amazonka";
```
mdimjasevic, line 96: Once the PR is merged, is it safe to delete our fork of Amazonka?

akshaymankar: Maybe let's not delete it, because that would break older revisions of wire-server code. But we can archive it to suggest that we don't use it anymore.

PR #2957, opened 2022-12-29
Add MakesFederatedCall combinators to Galley

isovector:
This PR follows the work done in #2950, also adding the combinators to Galley's endpoints.

## Checklist


isovector:
@mdimjasevic thanks so much. Rebased and ready for review!

libs/wire-api/src/Wire/API/Routes/Public/Galley/Conversation.hs
```diff
@@ -898,6 +965,9 @@ type ConversationAPI =
     :<|> Named
            "update-conversation-access"
            ( Summary "Update access modes for a conversation"
+               :> MakesFederatedCall 'Galley "on-conversation-updated"
```
mdimjasevic, line 968: This applies to any of these endpoints and any federated call that might happen and I'm picking just this one as an example. Suppose one day this endpoint does not make the federated call in Galley for the "on-conversation-updated" endpoint any more. Will the type checker tell us this is a redundant constraint? I suppose it will not.

We're in a similar problem with the `CanThrow` combinator. We have to declare all the errors an endpoint's handler can throw, but should we remove an error from the handler we get no warning the `CanThrow` combinator in the endpoint definition is now redundant.

libs/wire-api/src/Wire/API/Routes/Public/Galley/Feature.hs
```diff
@@ -100,10 +105,14 @@ type FeatureStatusGet f =
     '("get", f)
     (ZUser :> FeatureStatusBaseGet f)
 
-type FeatureStatusPut errs f =
+type family FoldSegments (calls :: [*]) r where
+  FoldSegments '[] r = r
+  FoldSegments (seg ': segs) r = seg :> FoldSegments segs r
```
mdimjasevic, line 110: This is the same as the `ApplyMods` type family, right? It would be great to at least leave a comment on that, or even better extract it out in a module and reuse it wherever needed.

services/galley/src/Galley/API/Internal.hs
```diff
@@ -396,9 +405,13 @@ type ITeamsAPIBase =
 
 type IFeatureStatusGet f = Named '("iget", f) (FeatureStatusBaseGet f)
 
-type IFeatureStatusPut errs f = Named '("iput", f) (FeatureStatusBasePutInternal errs f)
+type family FoldFederatedCalls (calls :: [*]) r where
+  FoldFederatedCalls '[] r = r
+  FoldFederatedCalls (call ': calls) r = call :> FoldFederatedCalls calls r
```
mdimjasevic, line 410: And again, this is the same as the `FoldSegments` and `ApplyMods` type families, isn't it? Seeing it for the third time I'm convinced this ought to be written only once and then just used.

mdimjasevic (commented):
Great, I overall like this!

I hope I'm not getting this wrong, but I saw a redefinition of one type family which is exactly the same as the existing `ApplyMods` one. I thought, OK, not great and I can live with it. But then I saw yet another definition of the same type family, and I thought this is an excuse to ask you to get rid of two redundant definitions and have just one. I left comments to these type families inlined.

Do you agree with the assessment?

mdimjasevic:
I forgot to mention it, but I like how you got rid of the fed constraints in types of Servant sitemaps!

libs/wire-api/src/Wire/API/Routes/Public/Galley/Conversation.hs
isovector: I just tested this, and you're right; we don't get redundant constraint warnings for this. I don't think such a thing is possible, unfortunately; GHC's support for constraint kinds is pretty lacking, and it was pretty hard to get this far.

isovector: There is a possible solution here by way of a GHC plugin, but I think it's ridiculously overkill. Pinging @julialongtin so she knows it's possible if it ever turns out to be a problem in the future, but I don't think it's worth doing now.

mdimjasevic (approved):

PR #2964, opened 2023-01-06
[FS-1334] Reset a Remote Subconversation

mdimjasevic:
As part of https://wearezeta.atlassian.net/browse/FS-1334, this PR adds support for resetting a remote subconversation. The local case was done in PR #2956.

## Checklist

### DIFF
--- a/changelog.d/2-features/delete-remote-subconversation
+++ b/changelog.d/2-features/delete-remote-subconversation
@@ -1,0 +1,1 @@
+Support deleting a remote subconversation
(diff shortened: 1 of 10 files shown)


smatting:
Closing this in favor of https://github.com/wireapp/wire-server/pull/2956

smatting:
Err nevermind

mdimjasevic:
> Looks good. However, the test coverage is a bit strange. If I'm not mistaken, we're testing the failure case of a non-member attempting to reset only in the remote case.

A fair point. I just added a test for the local case.


> Another question: is a client that is not part of a subconversation allowed to reset it?

My understanding is that a client is allowed to reset a subconversation so long as they're member of the parent conversation. They don't necessarily have to be in the subconversation.

PR #2969, opened 2023-01-10
[FS-1336] Leaving subconversations

pcapriotti:
https://wearezeta.atlassian.net/browse/FS-1336

## Checklist

### DIFF
--- a/services/galley/src/Galley/API/MLS/Message.hs
+++ b/services/galley/src/Galley/API/MLS/Message.hs
@@ -1397,6 +1397,7 @@ executeProposalAction qusr con lconvOrSub action = do
       Sem r (Maybe (Qualified UserId))
     checkRemoval cm qtarget clients = do
       let clientsInConv = Map.keysSet (Map.findWithDefault mempty qtarget cm)
+      -- TODO: skip the next check for subconversations
       when (clients /= clientsInConv) $ do
         -- FUTUREWORK: turn this error into a proper response
         throwS @'MLSClientMismatch
--- a/services/galley/src/Galley/API/MLS/Message.hs
+++ b/services/galley/src/Galley/API/MLS/Message.hs
@@ -1302,15 +1302,21 @@ executeProposalAction qusr con lconvOrSub action = do
   -- Type 2 requires no special processing on the backend, so here we filter
   -- out all removals of that type, so that further checks and processing can
   -- be applied only to type 1 removals.
-  removedUsers <- mapMaybe hush <$$> for (Map.assocs (paRemove action)) $
-    \(qtarget, Map.keysSet -> clients) -> runError @() $ do
-      -- fetch clients from brig
-      clientInfo <- Set.map ciId <$> getClientInfo lconvOrSub qtarget ss
-      -- if the clients being removed don't exist, consider this as a removal of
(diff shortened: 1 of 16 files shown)


services/galley/src/Galley/API/MLS/Message.hs
```diff
@@ -1242,44 +1241,51 @@ executeProposalAction ::
     Member TinyLog r,
     CallsFed 'Galley "on-conversation-updated",
     CallsFed 'Galley "on-mls-message-sent",
-    CallsFed 'Galley "on-new-remote-conversation",
+    CallsFed 'Galley "on-new-remote-conversation"
+  )
+
+executeProposalAction ::
+  forall r.
+  ( HasProposalActionEffects r,
```
elland, line 1249: 👍

mdimjasevic: Why are `CallsFed 'Galley` constraints part of the `HasProposalActionEffects` type alias, but the `CallsFed 'Brig` constraint is not?

services/galley/src/Galley/API/MLS/Message.hs
```diff
@@ -1344,43 +1357,63 @@ executeProposalAction loc qusr con (Conv mlsConv) action = do
...
+      ( handleNoChanges
+          . handleMLSProposalFailures @ProposalErrors
+          . fmap pure
+          . updateLocalConversationUnchecked @'ConversationRemoveMembersTag lconv qusr con
+      )
+      . nonEmpty
+      . filter (flip Set.member (existingMembers lconv))
+      . toList
+      $ users
+  SubConv _ _ -> pure []
```
mdimjasevic, line 1417: Removing is supported, isn't it? I can't find that in the MLS implementation in wire-server page, but I vaguely recall someone giving a use case for this operation.

services/galley/src/Galley/API/MLS/Message.hs
pcapriotti: Because it's not used by the other functions that use the constraint synonym. It doesn't really matter much, one way or the other.

services/galley/src/Galley/API/MLS/Message.hs
pcapriotti: Yes, but there's nothing to do here anyway, because we're not tracking user membership in subconversation (only client membership).

smatting (approved):

services/galley/test/integration/API/MLS.hs
```diff
@@ -2722,3 +2728,155 @@ testDeleteRemoteSubConv isAMember = do
...
+    -- charlie1 leaves
+    [charlie1KP] <-
+      map snd . filter (\(cid, _) -> cid == charlie1)
+        <$> getClientsFromGroupState alice1 charlie
+    mlsBracket [alice1, bob2] $ \wss -> do
+      leaveCurrentConv charlie1 qsub
+
+      msgs <-
+        WS.assertMatchN (5 # WS.Second) wss $
+          wsAssertBackendRemoveProposal charlie qcnv charlie1KP
```
smatting, line 2798: we should probably assert that the event is for qsub here

PR #2970, opened 2023-01-10
Optimize memory usage while creating large conversations

akshaymankar:
https://wearezeta.atlassian.net/browse/SQCORE-1373

There were 4 major changes:
1. Avoid duplicate work which was being done to create notifications. This actually doesn't save much memory, but it was nice.
2. Remove `StrictData` pragma from `Galley.Intra.Push.Internal`. This doesn't solve anything directly, but it is required for the next two solutions to work.
3. Refactor logic to chunk pushes: This reduced the memory usage quite a lot. Without this a lot of memory was retained for components of the `Value` type. But there was still a lot of memory consumption for the `ByteString`s that actually get sent.
4. Use chunked encoding for pushes: This ensures that the `ByteString`s in each request are generated only for the purpose of transmitting, because it avoids sending the `Content-Length` header. This makes the bytestring chunks to be generated and sent immmediately and hence allocated and deallocated immediately.

For more details, please read the commit messages.

### Test

Create a conversation with 500 users using `POST /v3/conversations`

<details>

Steps to actually run the test:
1. Modify `services/integration.sh` to not run galley
2. Compile everything and run `INTEGRATION_USE_NGINZ=0 ./services/start-services-only.sh`
3. Compile galley like this: `cabal build exe:galley --enable-profiling`
4. Prepare a JSON to create conversation with 500 users (ensure users are connected to each other).
5. Run galley from `services/galley` like this:
    ```
   ../../dist-newstyle/build/x86_64-linux/ghc-8.10.7/galley-0.83.0/x/galley/noopt/build/galley/galley -c galley.integration.yaml +RTS -l -hT; eventlog2html galley.eventlog
    ```
    Replace `-hT` with any of the heap profiling flags: https://downloads.haskell.org/~ghc/8.10.7/docs/html/users_guide/profiling.html#rts-options-for-heap-profiling 
6. Make the request
   ```
   curl localhost:8085/v3/conversations -X POST -d @create-conv.json -H 'Accept: application/json' -H 'Content-type: application/json' -H "Z-User: $CONV_CREATOR_ID" -H 'Z-Connection: foo' && pkill galley
   ```
7. Open `services/galley/galley.eventlog.html` in a browser.

</details>

### Results:

Before: 

Memory consumption peaks initially to about 280 MB and then slowly drops

After:
Memory rises quickly to about 3.4 MB and fluctuates a little until the end of the request

## Checklist


jschaul (approved):
💪🏿 🏆 🧠 🎆 🥳 😎 ✨ !!

I suppose this is also helpful in other code paths that send messages to a lot of people, right?

akshaymankar:
> I suppose this is also helpful in other code paths that send messages to a lot of people, right?

It follows the same push path, so I think so.

PR #2976, opened 2023-01-12
add read-only securitycontext to brig containers and volume mounts

amitsagtani97:
## Checklist

### DIFF
--- a/charts/brig/templates/tests/brig-integration.yaml
+++ b/charts/brig/templates/tests/brig-integration.yaml
@@ -54,15 +54,21 @@ spec:
     # The other test, "user.auth.cookies.limit", is skipped as it is flaky.
     # This is tracked in https://github.com/zinfra/backend-issues/issues/1150.
     command: [ "brig-integration", "--pattern", "!/turn/ && !/user.auth.cookies.limit/" ]
+    securityContext:
+      readOnlyRootFilesystem: true
     volumeMounts:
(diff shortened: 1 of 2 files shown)


charts/brig/templates/tests/brig-integration.yaml
```diff
@@ -54,22 +54,29 @@ spec:
     # The other test, "user.auth.cookies.limit", is skipped as it is flaky.
     # This is tracked in https://github.com/zinfra/backend-issues/issues/1150.
     command: [ "brig-integration", "--pattern", "!/turn/ && !/user.auth.cookies.limit/" ]
+    securityContext:
```
supersven, line 57: I think you don't have to care about any containers with names ending with `-integration`. These are only meant for integration tests.

jschaul (commented):
Making all containers and volumes read-only isn't correct. Some volumes are being written to by the processes. Some integration tests already fail here (see concourse logs). But in addition, some subtle functionality might stop working. E.g. I think the geoip container might normally write to its volume, if it doesn't, then "a device from IP=1.2.3.4 i.e. Germany/Berlin was added to your account" stops working; there might be more cases like this.

supersven:
I fear there was a misunderstanding regarding `geoip`: 
- As far as I can see, `brig` does not write to it's volume (only reads the geoip database).
- There is a service that updates the database.

PR #2978, opened 2023-01-12
Fix `ToSchema Version` instance.

fisx:
This partially reverts commit 6f911515dd50303590d14ed923d67acf4dde7db4 (6f911515dd50303590d14ed923d67acf4dde7db4).

## Checklist

### DIFF
--- a/libs/wire-api/src/Wire/API/Routes/Version.hs
+++ b/libs/wire-api/src/Wire/API/Routes/Version.hs
@@ -39,6 +39,7 @@ module Wire.API.Routes.Version
   )
 where
 
+import Control.Exception (assert)
 import Control.Lens ((?~))
 import Data.Aeson (FromJSON, ToJSON (..))
 import qualified Data.Aeson as Aeson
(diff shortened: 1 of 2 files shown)


libs/wire-api/src/Wire/API/Routes/Version.hs
```diff
@@ -63,8 +64,20 @@ data Version = V0 | V1 | V2 | V3
...
-      (\v -> element (fromIntegral $ fromEnum v) v) <$> [minBound @Version ..]
+    assert
+      (explicit == constructed)
+      (enum @Integer "Version" . mconcat . fmap (uncurry element) $ explicit)
+    where
+      constructed, explicit :: [(Integer, Version)]
+      constructed =
+        -- (fromEnum doesn't work as soon as we drop support for V0.)
+        (\v -> (read . Imports.tail . show $ v, v)) <$> [minBound @Version ..]
+      explicit =
```
elland, line 75: I don't like how easy it is to forget or make a mistake about this, but I don't have a better idea in mind at the moment. 😰

fisx: what do you mean?  how would you do a mistake here that'd go uncaught by the CI?

elland: I mean forgetting to add / update versions and it not being caught by GHC/ type checking. But hopefully CI is enough to prevent us from messing up 🤞

fisx:
> Can't we just revert and use an explicit list of versions? This is a mess.

my idea was that this way, you get an error in the integration tests if you forget to update `ToSchema`.  and i don't think it's that hard to read.

but i'll let you decide: please approve or reject.

pcapriotti (changes requested):
If you forget to add a new version that has some changes, you'll get an error in the tests anyway as soon as you make use of the new stuff. On that basis, I'm for restoring the simple code.

elland (approved):

pcapriotti (approved):

PR #2980, opened 2023-01-12
Convert docs from reStructuredText to MyST markdown

smatting:
The conversition was perfomed with [rst-to-myst](https://rst-to-myst.readthedocs.io/en/latest/).
A lot of manul preparation work was needed and has been squashed into a single commit.

I've QA'd the result using selenium to obtain screenshots of each of the 115 pages and diffing them with Imagemagick against the screenshots of the reference pages from http://docs.wire.com.
For non-trivial diffs I've inspected manually to find problems and fix.

The result builds with only a few warnings concerning duplicated labels and some minor layout problems, which I'll fix in a followup PR.

The scripts to convert and generate the screenshots are included in the `convert/` directory.


docs/src/conf.py
```diff
@@ -128,6 +128,5 @@
         "security-responses/log4shell": "2021-12-15_log4shell.html",
         "security-responses/cve-2021-44521": "2022-02-21_cve-2021-44521.html",
         "security-responses/2022-05_website_outage": "2022-05-23_website_outage.html",
-        "how-to/single-sign-on/index": "../../understand/single-sign-on/main.html#setting-up-sso-externally",
-        "how-to/scim/index": "../../understand/single-sign-on/main.html#user-provisioning",
+        "how-to/scim/index": "../../understand/single-sign-on/main.html#user-provisioning"
```
jschaul, line 131: These links are for people who may have bookmarked old URLs, they then get a redirect and the old links still work.
I'm not sure how important this is; but it's certainly nice user experience if previous links someone may have saved/bookmarked still work. Perhaps we should add a comment here for the next person stumbling over this

jschaul:
Cool beans! 😃

docs/src/conf.py
smatting: I thought about this, since `how-to-/single-sign-on/index` now points to https://docs.wire.com/how-to/single-sign-on/index.html which is a nice index of all thing sso I thought it's okay enough

PR #2985, opened 2023-01-13
[SQSERVICES-1825] OAuth refresh token generation (6/n)

battermann:
https://wearezeta.atlassian.net/browse/SQSERVICES-1825

## Checklist

 - [ ] Add a new entry in an appropriate subdirectory of `changelog.d`
 - [ ] Read and follow the [PR guidelines](https://docs.wire.com/developer/developer/pr-guidelines.html)

PR #2931, opened 2022-12-16
[SQSERVICES-1780] OAuth token handling (5/n)

battermann:
https://wearezeta.atlassian.net/browse/SQSERVICES-1780

## Checklist
 - [ ] Rebase on develop once #2922 is merged (or merge everything into one branch before merging to develop)
 - [x] Add a new entry in an appropriate subdirectory of `changelog.d`
 - [x] Read and follow the [PR guidelines](https://docs.wire.com/developer/developer/pr-guidelines.html)

PR #2922, opened 2022-12-13
[SQSERVICES-1783] OAuth Feature Flag (4/n)

battermann:
https://wearezeta.atlassian.net/browse/SQSERVICES-1783

## Checklist

 - [ ] Rebase on develop once #2907 is merged (or merge everything into one branch before merging to develop)
 - [x] Add a new entry in an appropriate subdirectory of `changelog.d`
 - [x] Read and follow the [PR guidelines](https://docs.wire.com/developer/developer/pr-guidelines.html)

PR #2907, opened 2022-12-07
[SQSERVICE-1779] OAuth token endpoint (3/n)

battermann:
https://wearezeta.atlassian.net/browse/SQSERVICES-1779

This PR is based on #2901 

## Checklist

 - [ ] Rebase on develop once #2901 is merged (or merge everything into one branch before merging to develop)
 - [x] Add a new entry in an appropriate subdirectory of `changelog.d`
 - [x] Read and follow the [PR guidelines](https://docs.wire.com/developer/developer/pr-guidelines.html)

PR #2989, opened 2023-01-16
OAuth

fisx:
https://wearezeta.atlassian.net/browse/SQSERVICES-1782

replaces #3086, #3046, #3010, #3004, #2985, #2931, #2922, #2907, #2901, #2882

## Checklist


docs/src/how-to/install/oauth.md
```diff
@@ -0,0 +1,72 @@
...
+        "kid":"QVapB_J...",
+        "qi":"sYHbPsy...",
+        "dp":"LFmnVNPW...",
+        "alg":"RS256",
+        "dq":"UXTY7...",
+        "n":"1mnyGVT..."
+      }
+```
+
+Note that the JWK is a sensitive configuration value, so it is recommended to use Helm's support for managing secrets instead of including it in a plaintext values.yaml file.
```
fisx, line 40: maybe reference how that support works, especially how it works in other places in this project?

docs/src/how-to/install/oauth.md
```diff
@@ -0,0 +1,72 @@
...
+
+Please keep in mind that OAuth is currently under development and may not be available for use yet. Once it is ready, you will be able to use the OAuth functionality by setting up the JWK as described above.
```
fisx, line 42:
```suggestion
[2023-01-16] OAuth is currently under development and may not be available for use yet. Once it is ready, you will be able to use the OAuth functionality by setting up the JWK as described above.
```

this way whoever stumbles accross this knows which time window in the changelog to scan for updates on this.

docs/src/how-to/install/oauth.md
```diff
@@ -0,0 +1,72 @@
...
+
+```yaml
+brig:
+  # ...
+  config:
+    # ...
+    optSettings:
+      # ...
+      setOAuthMaxActiveRefreshTokens: 20
+```
```
fisx, line 72: replace with a link to `reference/config-options.md`?  also, shouldn't the jwk setup be part of the reference docs as well, and this file only contains the story in broad strokes?

libs/wire-api/src/Wire/API/Routes/Bearer.hs
```diff
@@ -35,6 +35,10 @@ instance FromHttpApiData a => FromHttpApiData (Bearer a) where
     _ -> Left "Invalid authorization scheme"
   parseUrlPiece = parseHeader . T.encodeUtf8
 
+instance ToHttpApiData a => ToHttpApiData (Bearer a) where
```
fisx, line 38: what is this needed for?  there may be security implications if this is passed in the path or as a query param.  is there a way to avoid that?  (or is it "best practice" to do it this way?

services/brig/brig.integration.yaml
```diff
@@ -193,7 +193,12 @@ optSettings:
   setDpopTokenExpirationTimeSecs: 300 # 5 minutes
   setPublicKeyBundle: test/resources/jwt/ed25519_bundle.pem
   setEnableMLS: true
-
+  setOAuthJwkKeyPair: test/resources/oauth/jwk.json
+  setOAuthAuthCodeExpirationTimeSecs: 3 # 3 secs
+  setOAuthAccessTokenExpirationTimeSecs: 1814400 # 3 weeks
+  setOAuthEnabled: true
+  setOAuthRefreshTokenExpirationTimeSecs: 14515200 # 24 weeks
+  setOAuthMaxActiveRefreshTokens: 10
```
fisx, line 201:
```suggestion
  setOAuthJwkKeyPair: test/resources/oauth/jwk.json
  setOAuthAuthCodeExpirationTimeSecs: 3 # 3 secs
  setOAuthAccessTokenExpirationTimeSecs: 1814400 # 3 weeks
  setOAuthEnabled: true
  setOAuthRefreshTokenExpirationTimeSecs: 14515200 # 24 weeks
  setOAuthMaxActiveRefreshTokens: 10

```

libs/wire-api/src/Wire/API/Routes/Public.hs
```diff
@@ -275,3 +286,74 @@ instance HasServer api ctx => HasServer (OmitDocs :> api) ctx where
 
 instance RoutesToPaths api => RoutesToPaths (OmitDocs :> api) where
   getRoutes = getRoutes @api
+
+--------------------------------------------------------------------------------
+-- Z-OAuth
+
+data ZUserOrOAuth (scope :: OAuthScope)
```
fisx, line 293:
```suggestion
-- FUTUREWORK: it would be nice to have unit tests for this and the instances (esp. `HasServer`, but we cover this with integration tests in brig et al. for now.
data ZUserOrOAuth (scope :: OAuthScope)
```

(right?)

libs/wire-api/src/Wire/API/Routes/Bearer.hs
fisx: also, if you keep it please add a roundtrip-test.

services/brig/test/resources/oauth/jwk.json
```diff
@@ -0,0 +1 @@
...ET_B_LWL_EQQRB2cqfiThkDghN5HWZn18-QskyYgS0vWS_EYquM8LS3s0emQhXkz77ZX4fQrhZGRCAmDeI-iZDt90uRTF6OceKKoC5eTimx9YimQ0Z-qOlKiCpIjplSgHL8QgIZrp6mLdCTVy3vxaegwSNKwGqCCwcewtdBky9nb9pX6sEc6CA3WfyYZMbthMYIYTi0KJ2kaaF2QdR3l7VLTdPBwK-Q","e":"AQAB","use":"sig","kid":"QVapB_JRK66AMCYKN7LHsl4DlkICTNYB6ExJrtmva88","qi":"sYHbPsyiiRIBKbN-chbXYTLsd03jL_kiT2VdyMsP_tNjOz87WcC8Td-lBUIViw_aMq1VOjJEyhB6yE9-Po4YmlAFbPSf1rmypAdjTKRQznkuTGDue6yd4z1t4SgZJOSlpJSGBtkEiYaIlI9-fviPxtIpDh0JfAbE7XgPItu5vT8","dp":"LFmnVNPWVx6H6yHW2dPF8osTb6P5O3HUnftwTQHt4WAVY_Fl4vcWEfj3_ZD8s9VSFhM3apuG7zC1rV-WoZTf6rvhE2--R4creXOFKc4ZOEQo6pU61pqf6VmAoF8chqPI_178q0CSxV_JoERhIZUKizgD3mpFyS1ArjygBREiI5c","alg":"RS256","dq":"UXTY7yryQhql-mfugc_q98CanKuU8G17r5xrbw5wriTwCPOmsFJzZr8UdCorkA_oTw8CpsrwXUBEJJUA-9R3tKhYjfxsGP-nIAsMyfHdkI2SlsztYQ9f8wE39Bs_z4qKZTJyprvdKY76DKvMh7YDm-Wx8_8VA0GZWUN6ldFa2lU","n":"1mnyGVTdwCi4umlGMvT6y7aTmibMQuBvySoT5eZcwKti4_sfnEJhyWXfsE6tasUC8ce7YGzFiq886S9-iZc6hYW0ReIPQYCLZ5hn-fUCxefEZdX03e-uJww7OwX_kNQifgdYwzM__QJsy1_nbgEn7olTzplJfJmUSk9dkfFJ-3o5L3AJ1YMbFPgXae0OLmkxjVKfT_6093a0RLSpUCzcti4OhtNa-OEUMX4rYl0jQyzZyO281IH-MatecA1O_o6JbCI1wH46jWNY8of43cemzFwQaS2SnSyppOmJzGf9fpm5j5dozYrAdTdPYdZEA62CimvdLvJqEt4OjyQuRTAuCw"}
```
fisx, line 1: how did you generate this?  maybe explain somewhere in the tests where this is used?  or store it as yaml and leave a comment here?

services/brig/test/integration/API/OAuth.hs
```diff
@@ -0,0 +1,501 @@
...
...o46dJ5iw0wnuSF0dAiqVG0rSJK-gVBdIbzZYxhSBW4ZF3n4CqtFb6lc1stfZHcnzWHyF6Cofzup6pJumeFe7xXF9-aGU-3UcTSzTnMa21NVP-vT2CXkH8dSfwLI-PuJwlW6tcpBwT2PXrCGyAGqQ3h5cdAmwcgfbla8wqrzj1A08SlkKHvTDixVvnnzpQ\",\"e\":\"AQAB\",\"use\":\"sig\",\"kid\":\"0makAydOdX3vNv4YTToO45ccQUCOoLisvAFVyhiKA4c\",\"qi\":\"phNbA_tiDLQq1omVgM1dHtOe6Dd7J_ZoRdz1Rmc4uaSQyJe-yn88DxXlX10DJkM9uqyzcojOtD5awBUXgYSzmasZvcZ0e2XNi7iXmSwsggTux3lUVVqKWV8HreaSywJ-HqitxjitooWSWOyD9o8yq9RS4r2QdXyuCfthwnEZdpc\",\"dp\":\"q0IJJmjZYolFiYsdq5sq5erWerPGyl0l6gRuiECcqiTVmeQINu81_Wm5gPuNFwHO0JBkt-NBpOprUFHHLvwCwmu3n77ZGfH3VqCq-FT7fMlQ5NCngmvF1bqtmlHJ84X_MCpdY4oDioxcwEl4HDYDrHO17774UItVWxDmXl0rCPs\",\"alg\":\"PS512\",\"dq\":\"NznQQDVsPTofSIPEQeLisIyDoZvsoCk4ael_nPUjaZZ-32L_FNvrLQTZeMl8JVf0yJ4d0ePa8EyTaZb8AqflXT_i1mRw-n-6BP5earMG5_FMGMXfXsKJ04lVEJ94eT-jGTOH--qjJ1fxk_6vNEy73RgrtXmYMGzU1Yhx-duqsrk\",\"n\":\"o9VozUwUc1mQMrAH2fEna_ihmNa3CVRzK7MUgDHEbfY0T71wREpK4f4fOkDysKIqnmMdxRzJhsXTDpxX8_8AlKcimPgR8Qb2z7GwDsnDZOdgAYrZ7l7gj0nX02IX35MBk7a7tWr0nILFLV9SxEu6UFcZo0bL2Rhck81TRqLbomJpIzAq8VCS8uMQeg6hEMarl9tGvSKyFuMdTCV3JE9dSv_NErAWx7uBIgkai3Imjs4ufatvRsi9ZHaUV5V3NtrFbYDulg-GOH1eXZwnO6UrKgcAdB3nS1WKL-vcxqupceAHeFHRjARm6AV07hJyXVOVHxdffv6BFX5GihFPFvpQXQ\"}"
```
fisx, line 501: either store both good and bad jwk in a file, or neither.

docs/src/developer/reference/config-options.md
```diff
@@ -605,11 +605,57 @@ If there is no configuration for a domain, it's defaulted to `no_search`.
...
+
+Optionally, configure the OAuth authorization code, access token, and refresh token expiration time in seconds as follows:
+
+```yaml
+# [brig.yaml]
+optSettings:
+  # ...
+  setOAuthAuthCodeExpirationTimeSecs: 300 # 5 minutes
+  setOAuthAccessTokenExpirationTimeSecs: 300 # 5 minutes
+  setOAuthRefreshTokenExpirationTimeSecs: 14515200 # 24 weeks
```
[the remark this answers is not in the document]
battermann: There is a comment in `Brig.Options`:

```
    -- | The expiration time of an OAuth access token in seconds.
    -- use `setOAuthAccessTokenExpirationTimeSecs` as the getter function which always provides a default value
    setOAuthAccessTokenExpirationTimeSecsInternal :: !(Maybe Word64),
    -- | The expiration time of an OAuth authorization code in seconds.
    -- use `setOAuthAuthCodeExpirationTimeSecs` as the getter function which always provides a default value
    setOAuthAuthCodeExpirationTimeSecsInternal :: !(Maybe Word64),
    -- | En-/Disable OAuth
    -- use `setOAuthEnabled` as the getter function which always provides a default value
    setOAuthEnabledInternal :: !(Maybe Bool),
    -- | The expiration time of an OAuth refresh token in seconds.
    -- use `setOAuthRefreshTokenExpirationTimeSecs` as the getter function which always provides a default value
    setOAuthRefreshTokenExpirationTimeSecsInternal :: !(Maybe Word64),
 ```
 
 I am a bit hesitant to duplicate this. But it's your call.

File: docs/src/how-to/install/oauth.md (hunk 1 above)
Comment (battermann) [reply docs/src/how-to/install/oauth.md:40]: I currently don't know how this works. What about a `FUTURWORK: fill in the details on how this works once we have implemented it for prod`?

File: libs/wire-api/src/Wire/API/Routes/Bearer.hs (hunk 4 above)
Comment (battermann) [reply libs/wire-api/src/Wire/API/Routes/Bearer.hs:38]: It's only needed in test, but yes, we can make the production code safer by moving this to the test code.

File: libs/wire-api/src/Wire/API/Routes/Public.hs (hunk 6 above)
Comment (battermann) [reply libs/wire-api/src/Wire/API/Routes/Public.hs:293]: That's right

File: services/brig/test/integration/API/OAuth.hs [hunk 10]
```diff
@@ -0,0 +1,780 @@
...
+  let newOAuthClient@(NewOAuthClient _ redirectUrl) = newOAuthClientRequestBody "E Corp" "https://example.com"
+  cid <- occClientId <$> registerNewOAuthClient brig newOAuthClient
+  uid <- randomId
+  let scope = OAuthScopes $ Set.fromList [ConversationCreate, ConversationCodeCreate]
+  let state = "foobar"
+  createOAuthCode brig uid (NewOAuthAuthCode cid scope OAuthResponseTypeCode redirectUrl state) !!! do
+    const 302 === statusCode
+    const (Just $ unRedirectUrl redirectUrl ^. pathL) === (fmap getPath . getLocation)
+    const (Just $ ["code", "state"]) === (fmap (fmap fst . getQueryParams) . getLocation)
+    const (Just $ cs state) === (getLocation >=> getQueryParamValue "state")
```
fisx, line 135: also check some syntactic property of code?  like, minimal length, or actual length?

services/brig/test/integration/API/OAuth.hs
```diff
@@ -0,0 +1,780 @@
...
+testCreateOAuthCodeRedirectUrlMismatch :: Brig -> Http ()
+testCreateOAuthCodeRedirectUrlMismatch brig = do
+  let newOAuthClient@(NewOAuthClient _ redirectUrl) = newOAuthClientRequestBody "E Corp" "https://example.com"
+  cid <- occClientId <$> registerNewOAuthClient brig newOAuthClient
+  uid <- randomId
+  let state = "foobar"
+  let differentUrl = mkUrl "https://wire.com"
+  createOAuthCode brig uid (NewOAuthAuthCode cid (OAuthScopes Set.empty) OAuthResponseTypeCode differentUrl state) !!! do
+    const 400 === statusCode
+    const (Just $ "access_denied") === (getLocation >=> getQueryParamValue "error")
```
fisx, line 146:
```suggestion
    const (Just $ "access-denied") === (getLocation >=> getQueryParamValue "error")
```

?  unless this is determined by the oauth rfcs, of course.  but we do dashes everywhere else.

services/brig/test/integration/API/OAuth.hs
```diff
@@ -0,0 +1,780 @@
...
+    expectedAppName @?= ocName oauthClientInfo
+    expectedUrl @?= ocRedirectUrl oauthClientInfo
+
+testCreateOAuthCodeSuccess :: Brig -> Http ()
+testCreateOAuthCodeSuccess brig = do
+  let newOAuthClient@(NewOAuthClient _ redirectUrl) = newOAuthClientRequestBody "E Corp" "https://example.com"
+  cid <- occClientId <$> registerNewOAuthClient brig newOAuthClient
+  uid <- randomId
+  let scope = OAuthScopes $ Set.fromList [ConversationCreate, ConversationCodeCreate]
+  let state = "foobar"
```
fisx, line 130: it's good practice to not have something too common in here, or there is a (theoretical) risk of getting a "foobar" that's actually coming from somewhere else.

```suggestion
  let state = "10b2df8a-9e22-11ed-93eb-47378d658ebd"
```

(or at least hit the keyboard with the fist once, instead of using such a common word.)

(also below.)

services/brig/test/integration/API/OAuth.hs
```diff
@@ -0,0 +1,780 @@
...
+  let newOAuthClient@(NewOAuthClient _ redirectUrl) = newOAuthClientRequestBody "E Corp" "https://example.com"
+  cid <- occClientId <$> registerNewOAuthClient brig newOAuthClient
+  uid <- randomId
+  let state = "foobar"
+  let differentUrl = mkUrl "https://wire.com"
+  createOAuthCode brig uid (NewOAuthAuthCode cid (OAuthScopes Set.empty) OAuthResponseTypeCode differentUrl state) !!! do
+    const 400 === statusCode
+    const (Just $ "access_denied") === (getLocation >=> getQueryParamValue "error")
+    const (Just $ cs state) === (getLocation >=> getQueryParamValue "state")
+    const (Just $ unRedirectUrl redirectUrl ^. pathL) === (fmap getPath . getLocation)
```
fisx, line 148: should this be Nothing as well?  is it bad that we give away information about the client that appears to be under attack here?  (same in a few places below.)

services/brig/test/integration/API/OAuth.hs
battermann: It is indeed determined by the OAuth RFC [here](https://www.rfc-editor.org/rfc/rfc6749#section-4.1.2.1).

services/brig/test/integration/API/OAuth.hs
battermann: No, it's correct according to the specs. We do not reveal any additional information, since the returned redirect URL is the same as was passed in the request by the client. The redirect is, in fact, the mechanism to communicate back to the client. That's why errors are added as query parameters.

services/brig/test/integration/API/OAuth.hs
fisx: did you resolve these comments without changing anything, and without explaining why not?  or did you forget to push?

services/brig/test/integration/API/OAuth.hs
fisx: I don't understand.  `redirectUrl` is part of the oauth client's identity, communicated between oauth client and brig.  the `createOAuthCode` request can be sent by an attacker who knows the client id, but not the redirect url, and we want to avoid leaking the redirect url, given that the sender clearly doesn't know it.

I don't think this is a big deal, just that it's not necessary to reveal that information here, and (at least theoretically) harmful.

battermann: Yes, the redirect URL is part of the client's identity. And it is additionally provided by the client with the authorization request. The redirect URL that is passed has to match the one that was registered with the client. And the URL that has been passed is returned, not the one registered with the client.

battermann: [from the specs](https://www.rfc-editor.org/rfc/rfc6749#section-4.1.2.1)

> If the request fails due to a missing, invalid, or mismatching
   redirection URI, or if the client identifier is missing or invalid,
   the authorization server SHOULD inform the resource owner of the
   error and MUST NOT automatically redirect the user-agent to the
   invalid redirection URI.
   
so you are right, we must not redirect but return an error.

libs/libzauth/libzauth-c/src/zauth.h
```diff
@@ -49,14 +55,15 @@ void        zauth_keystore_delete(ZauthKeystore * store);
...
-ZauthTokenType zauth_token_type(ZauthToken const *);
-long           zauth_token_time(ZauthToken const *);
-uint8_t        zauth_token_version(ZauthToken const *);
-Range          zauth_token_lookup(ZauthToken const *, uint8_t);
-ZauthResult    zauth_token_allowed(ZauthToken const *, ZauthAcl const *, uint8_t const * path, size_t len, uint8_t * result);
-void           zauth_token_delete(ZauthToken *);
+ZauthResult            zauth_token_parse(uint8_t const * str, size_t len, ZauthToken **);
+ZauthResult            zauth_token_verify(ZauthToken const *, ZauthKeystore const *);
+ZauthTokenType         zauth_token_type(ZauthToken const *);
+ZauthTokenVerification zauth_token_verification(ZauthToken const *);
```
fisx, line 61: note to self: this is the only line that has been changed in this diff block (except for indentation).

libs/libzauth/libzauth/src/zauth.rs
```diff
@@ -85,6 +85,15 @@ impl FromStr for TokenType {
     }
 }
 
+// Token Verification ////////////////////////////////////////////////////////
+
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
```
fisx, line 90: `PartialEq`?  what happens if we remove that?  (i'm clueless with rust)

services/nginz/third_party/nginx-zauth-module/zauth_module.c
```diff
@@ -455,11 +542,45 @@ static ngx_int_t zauth_token_typeinfo (ngx_http_request_t * r, ngx_http_variable
...
 static ngx_int_t zauth_token_var (ngx_http_request_t * r, ngx_http_variable_value_t * v, uintptr_t data) {
         ZauthToken const * t = ngx_http_get_module_ctx(r, zauth_module);
         if (t == NULL) {
                 return NGX_ERROR;
         }
+        if (!zauth_is_authorized_and_allowed(r)) {
+                zauth_empty_val(v);
+                return NGX_OK;
+        }
+
```
fisx, line 583: call this in `zauth_handle_zauth_request` instead?

fisx (commented):
notes from pair review session

charts/nginz/values.yaml
```diff
@@ -377,6 +378,21 @@ nginx_conf:
...
       - all
       disable_zauth: true
+    - path: /oauth/clients/([^/]*)$
+      envs:
+      - all
+    - path: i/oauth/clients$
+      envs:
+      - staging
+      disable_zauth: true
+      basic_auth: true
```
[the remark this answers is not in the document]
battermann: I guess it makes sense to have this exposed (and secured via basic auth) in staging for QA, no?

services/nginz/third_party/nginx-zauth-module/zauth_module.c
battermann: Not exactly possible. In `zauth_handle_zauth_request` the token signature is being verified.
In `zauth_is_authorized_and_allowed` however, we check if the signature HAS BEEN verified. We do not want to do the crypto multiple times.

libs/libzauth/libzauth/src/zauth.rs
battermann: It looks like to derive `Eq` we need to derive `PartialEq` as well.

docs/src/how-to/install/oauth.md
flokli: This is specific to an individual deployment, and not even limited to the oauth secrets, but other secrets as well. It might be fine to explicitly point it out like that here, but how secrets are handled really is a case-by-case thing, and whoever deploys wire should form their own informed decision.

services/nginz/third_party/nginx-zauth-module/zauth_module.c
```diff
@@ -59,7 +66,15 @@ static ngx_command_t zauth_commands [] = {
         , NGX_HTTP_LOC_CONF | NGX_CONF_TAKE1
         , ngx_conf_set_flag_slot
         , NGX_HTTP_LOC_CONF_OFFSET
-        , offsetof (ZauthLocationConf, toggle)
+        , offsetof (ZauthLocationConf, zauth)
+        , NULL
+        }
+
+       , { ngx_string ("oauth")
```
flokli, line 73: I'm personally no fan of suddenly having the nginz zauth module now also handle oauth as well, and pushing more of this logic into nginz.

Is there a way for backend services to do this verification?

If there is not, we should at least namespace these options, and make sure this behaviour is properly documented.

battermann: The intention was to have wire-server handle OAuth requests and do all the verifications, and that's how it's implemented.
At the same time, we want the same routes for both OAuth and ZAuth.
Therefore, nginz needs to let OAuth requests through to wire-server. The changes that are made to nginz in this PR are IMO the most minimal invasive to accomplish this.

I agree with the documentation.

I don't know what you mean with the namespaces?

flokli: What I meant to say was i don't know / understand what the new `oauth yes` directive in the zauth nginx module is doing, what it does with these requests if "it let's these requests through" and "wire-server handles oauth requests". Why does zauth need to handle these specifically at all? Can't we just bypass it for these paths?

battermann: The problem is, that it is exactly the same path. No way to decide if this is an OAuth or ZAuth request other than by trying to parse it as ZAuth and see if it succeeds or not. Or do you have a better idea of how to achieve this?

flokli: So wherever `oauth yes` is set, this simply means "if authentication (seemingly) fails, simply let through anyways?

Do we do more zauth-style validation inside each service as well, or what happens if I access one of these `oauth yes` endpoints?

fisx: @flokli i agree with your sentiment, but maybe a little documentation will make this a tolerable design.  something like this:

> zauth paths can be set up to additionally support oauth: for routes configured accordingly, the `Authorization` header is processed for zauth credentials, and if that fails, instead of responding with 403, the `Authorization` header is passed to the service intact for oauth processing.

while i'm writing this, i'm wondering if

- (1) we can do anything about the risk of nginz expecting brig to handle security, and brig expecting nginz to do that.  some ad-hoc script that reads both the swagger of the public api and the nginz yaml, and breaks CI if anything goes through by accident?
- (2) should we even name that oauth, or should we call it "custom-service-auth" or something?
- (3) where should we document this?

battermann: If `oauth yes` is set and authentication fails, nginz lets the request through under the following conditions:
- The request has an Authorization header with a bearer token
- All the other Zauth headers (e.g. Z-User etc.) are emptied
- The token (from the Authorization header) is written to the `Z-OAuth` header

In wire-server, we check if there are proper ZAuth headers (they can only be present if nginz successfully verified the request as a zauth request).

If not, wire-server looks at the `Z-OAuth` header and does the proper OAuth verification including checking the correct sope. If that succeeds, the request handler is called.

battermann (approved):

battermann (approved):

battermann (approved):

battermann (approved):

battermann (dismissed):

libs/wire-api/src/Wire/API/OAuth.hs
```diff
@@ -0,0 +1,597 @@
...
+import Data.Time
+import Imports hiding (exp, head)
+import Servant hiding (Handler, JSON, Tagged, addHeader, respond)
+import Servant.Swagger.Internal.Orphans ()
+import Test.QuickCheck (Arbitrary)
+import URI.ByteString
+import Web.FormUrlEncoded (Form (..), FromForm (..), ToForm (..), parseUnique)
+import Wire.API.Error
+import Wire.Arbitrary (GenericUniform (..))
+import qualified SAML2.WebSSO as configs
```
fisx, line 50: what's this?  doesn't compile for me locally.

fisx: yes, don't know where this came from, but it's gone now: [5b07b11](https://github.com/wireapp/wire-server/pull/2989/commits/5b07b111c96e92fda50f94739dfddad08a7e4a32)

libs/wire-api/src/Wire/API/OAuth.hs
```diff
@@ -0,0 +1,596 @@
...
+
+--------------------------------------------------------------------------------
+-- Types
+
+newtype RedirectUrl = RedirectUrl {unRedirectUrl :: URIRef Absolute}
+  deriving (Eq, Show, Generic)
+  deriving (A.ToJSON, A.FromJSON, S.ToSchema) via (Schema RedirectUrl)
+
+addParams :: [(ByteString, ByteString)] -> RedirectUrl -> RedirectUrl
+addParams ps (RedirectUrl uri) = uri & (queryL . queryPairsL) .~ (ps ++ (uri ^. (queryL . queryPairsL))) & RedirectUrl
```
fisx, line 59: not tested, but something like this should work:

```suggestion
addParams ps (RedirectUrl uri) = uri & (queryL . queryPairsL) %~ (ps <>) & RedirectUrl
```

libs/wire-api/src/Wire/API/OAuth.hs
```diff
@@ -0,0 +1,596 @@
...
+
+instance ToHttpApiData RedirectUrl where
+  toUrlPiece = TE.decodeUtf8With TErr.lenientDecode . toHeader
+  toHeader = serializeURIRef' . unRedirectUrl
+
+instance FromHttpApiData RedirectUrl where
+  parseUrlPiece = parseHeader . TE.encodeUtf8
+  parseHeader = bimap (T.pack . show) RedirectUrl . parseURI strictURIParserOptions
+
+newtype OAuthApplicationName = OAuthApplicationName {unOAuthApplicationName :: Range 1 256 Text}
```
fisx, line 83:
```suggestion
newtype OAuthApplicationName = OAuthApplicationName {unOAuthApplicationName :: Range 6 256 Text}
```

?

libs/wire-api/src/Wire/API/OAuth.hs
```diff
@@ -0,0 +1,596 @@
...
+    enum @Text "OAuthResponseType" $
+      mconcat
+        [ element "code" OAuthResponseTypeCode
+        ]
+
+-- | The OAuth scopes that are supported by the backend.
+-- This type is a bit redundant and unfortunately has to be kept in sync
+-- with the supported scopes defined in the nginx configs.
+-- However, having this typed makes it easier to handle scopes in the backend,
+-- and e.g. provide more meaningful error messages when the scope is invalid.
```
fisx, line 172: why isn't the scope validated and eliminated in nginz?  what can the backend not do if it doesn't know anything about scopes?  i think the answer has something to do with claim sets.  i'll keep reading, but if you have an easy explanation it might help!

libs/wire-api/src/Wire/API/OAuth.hs
```diff
@@ -0,0 +1,596 @@
...
+  builder = \case
+    WriteConversation -> "write:conversations"
+    WriteConversationCode -> "write:conversations_code"
+    ReadSelf -> "read:self"
+    ReadFeatureConfigs -> "read:feature_configs"
+
+instance FromByteString OAuthScope where
+  parser = do
+    s <- parser
+    case s & T.toLower of
```
fisx, line 191:
```suggestion
    case T.toLower s of
```

libs/wire-api/src/Wire/API/OAuth.hs
```diff
@@ -0,0 +1,596 @@
...
+  builder = builder . unOAuthAuthCode
+
+instance FromByteString OAuthAuthCode where
+  parser = OAuthAuthCode <$> parser
+
+instance FromHttpApiData OAuthAuthCode where
+  parseQueryParam = bimap cs OAuthAuthCode . validateBase16 . cs
+
+instance ToHttpApiData OAuthAuthCode where
+  toQueryParam = toText . unOAuthAuthCode
```
fisx, line 258: is this needed?  if not, it would be nice to have it removed so we don't accidentally use it to leak the code.

libs/wire-api/src/Wire/API/Routes/Public/Brig/OAuth.hs
```diff
@@ -0,0 +1,160 @@
...
+  toUnion (CreateOAuthCodeUnsupportedResponseType url) = S (S (Z (I url)))
+  toUnion (CreateOAuthCodeClientNotFound url) = S (S (S (Z (I url))))
+  toUnion CreateOAuthCodeRedirectUrlMissMatch = S (S (S (S (Z (I (dynError @(MapError 'OAuthRedirectUrlMissMatch)))))))
+  fromUnion :: Union (ResponseTypes CreateOAuthAuthCodeResponses) -> CreateOAuthCodeResponse
+  fromUnion (Z (I url)) = CreateOAuthCodeSuccess url
+  fromUnion (S (Z (I url))) = CreateOAuthCodeFeatureDisabled url
+  fromUnion (S (S (Z (I url)))) = CreateOAuthCodeUnsupportedResponseType url
+  fromUnion (S (S (S (Z (I url))))) = CreateOAuthCodeClientNotFound url
+  fromUnion (S (S (S (S (Z (I _)))))) = CreateOAuthCodeRedirectUrlMissMatch
+  fromUnion (S (S (S (S (S x))))) = case x of {}
```
fisx, line 157: what the?!  isn't there a less boilerplaty way of doing this?

docs/src/how-to/install/index.rst
```diff
@@ -0,0 +1,31 @@
+Installing wire-server
```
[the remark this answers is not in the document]
battermann: IDK, this confused me, too.

libs/libzauth/libzauth-c/Makefile
```diff
@@ -36,7 +36,7 @@ install: build-release
 	sed -e "s~<<VERSION>>~$(VERSION)~" \
 		-e "s~<<PREFIX>>~$(PREFIX_INSTALL)~" \
 		src/libzauth.pc > $(PREFIX_INSTALL)/lib/pkgconfig/libzauth.pc
-	cp target/release/libzauth.$(LIB_TYPE) $(PREFIX_INSTALL)/lib/
+	cp target/debug/libzauth.$(LIB_TYPE) $(PREFIX_INSTALL)/lib/
```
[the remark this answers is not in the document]
battermann: Good catch, this must be reverted.

libs/wire-api/src/Wire/API/OAuth.hs
battermann: If backend doesn't know about scopes, it will just accept any text/string here. Having this type makes it much more restricted as the input it gets is parsed and if the scope is not known or formatted invalidly, it can reject the request with a proper error message. The problem with this approach, though, is that adding/changing scopes will need a new backend deployment, but this is the case anyway because the scopes are defined in the nginx.config. However, the danger is that we add a new scope but do not update this type. I left it here, because it was left over from the previous and outdated design. But I am ok with removing this type, too.

libs/wire-api/src/Wire/API/OAuth.hs
battermann: Hehe, I like it with the `&` because it reads better (left to right), but I don't mind. :)

libs/wire-api/src/Wire/API/OAuth.hs
battermann: We need this because the code is returned via a query parameter in the redirect url (in the location header)

libs/wire-api/src/Wire/API/Routes/Public/Brig/OAuth.hs
battermann: :shrug:

docs/src/how-to/install/index.rst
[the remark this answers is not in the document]
fisx: ah, maybe this gives you a hint!  :)

```
$ make dev-run
[...]
WARNING: multiple files found for the document "how-to/install/index": ['how-to/install/index.md', 'how-to/install/index.rst']
```

services/brig/docs/swagger.md
```diff
@@ -1,6 +1,19 @@
...
+The end-points that require zauth are labelled so in the description
+below.  The end-points that support oauth as an alternative to zauth
+have the required oauth scopes listed in the same description.
+
+Futher reading:
+- https://docs.wire.com/how-to/install/oauth.html
+- https://github.com/wireapp/wire-server/blob/develop/libs/wire-api/src/Wire/API/Routes/Public.hs (search for HasSwagger instances)
+- `curl https://staging-nginz-https.zinfra.io/v4/api/swagger.json | jq '.security, .securityDefinitions`
+
+## SSO Endpoints
```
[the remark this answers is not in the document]
battermann: hm, yes, but how is this related to this PR?

services/brig/src/Brig/API/Public.hs
```diff
@@ -203,35 +207,43 @@ internalEndpointsSwaggerDocsAPI service examplePort swagger Nothing =
...
-    Member CodeStore r,
-    Member (Concurrency 'Unsafe) r,
-    Member GalleyProvider r,
-    Member JwtTools r,
-    Member Now r,
-    Member PasswordResetStore r,
-    Member PublicKeyBundle r,
-    Member (UserPendingActivationStore p) r
-  ) =>
+  Members
```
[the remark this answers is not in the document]
battermann: Happened due to a merge. I changed it back.

docs/src/developer/reference/oauth.md
```diff
@@ -0,0 +1,424 @@
...
+       participant C as Outlook Calendar Extension
+       participant A as Authorization Server (wire-server)
+       participant R as Resource Server (wire-server)
+
+       U->>C: Click login
+       C->>A: Authorization code request /authorize
+       A->>U: Redirect to login/authorization prompt
+       U->>A: Authenticate and consent
+       A->>C: Authorization code
+       C->>A: Authentication code + client crednetials
```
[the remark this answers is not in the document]
battermann: The secret is the OAuth client's password.
The authorization code is sent to the authorization server indirectly via the `GET /authorize` request as a query parameter.

services/brig/test/integration/API/OAuth.hs
```diff
@@ -0,0 +1,863 @@
...
+    const Nothing === (getLocation >=> getQueryParamValue "code")
+
+testCreateAccessTokenSuccess :: Opt.Opts -> Brig -> Http ()
+testCreateAccessTokenSuccess opts brig = do
+  now <- liftIO getCurrentTime
+  uid <- userId <$> createUser "alice" brig
+  let redirectUrl = mkUrl "https://example.com"
+  let scopes = OAuthScopes $ Set.singleton ReadSelf
+  (cid, secret, code) <- generateOAuthClientAndAuthCode brig uid scopes redirectUrl
```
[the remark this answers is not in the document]
fisx: no, my beef is with the abbreviation `auth`, which can stand for both `authorization` and `authentication`.  and the fact that `authentication` is used as a synonym for `secret`.

docs/src/developer/reference/oauth.md
```diff
@@ -0,0 +1,428 @@
...
+
+The resource server is the API server the 3rd party app attempts to access on behalf of the user. In our case the resource server is `wire-server`.
+
+#### Authorization server
+
+The authorization server does the authentication of the user and establishes whether the user approves or denies the client's access request. In this case the authorization server is the same server as the resource server which is `wire-server`.
+
+### Supported OAuth flow
+
+`wire-server` currently only supports the [authorization code flow](https://www.rfc-editor.org/rfc/rfc6749#section-4.1) which is optimized for confidential clients such as Outlook Calendar Extension.
```
fisx, line 41: could you somehow link the following sections to the arrows in this diagram?

docs/src/developer/reference/oauth.md
```diff
@@ -0,0 +1,428 @@
...
+}
+```
+
+These credentials have to be stored in a safe place and cannot be recovered if they are lost.
+
+### Get authorization code
+
+When the user wants to use the 3rd party app for the first time, they need to authorize it to access Wire resources on their behalf.
+
+To do so the user is redirected to a login screen for authentication. Once authenticated, the user can either grant or deny the client's access request and the corresponding scope (a list of permissions to give to the 3rd party app).
```
fisx, line 105: is that the wire login screen or the google one?  if it's the wire one, please say so.  also: why do we redirect to the login screen?  odds are the user is already logged in and could be redirected to the oauth page directly.  also, how does the user get to the grant page after login?

docs/src/developer/reference/oauth.md
```diff
@@ -0,0 +1,428 @@
...
+| `response_type` | Required. Value MUST be set to  `code`.                                                                                                                                                                                                                                           |
+| `client_id`     | Required. The client identifier.                                                                                                                                                                                                                                                  |
+| `redirect_url`  | Required. MUST match the URL that was provided during client registration                                                                                                                                                                                                         |
+| `state`         | Required. An opaque value used by the client to maintain state between the request and callback.</br>The authorization server includes this value when redirecting the user-agent back to the client.</br>The parameter SHOULD be used for preventing cross-site request forgery. |
```
fisx, line 126: if you say "should" (and i think "must" would be bad because we need to support whatever other people are doing here, if it's the standard) there should be an integration-test that it works without.  did i miss that test?

battermann: In the standard this is recommended, and we actually require it. If this doesn't work, we can still change it later, IMO.

docs/src/developer/reference/config-options.md
```diff
@@ -618,11 +618,60 @@ If there is no configuration for a domain, it's defaulted to `no_search`.
 
 This options determines whether development versions should be enabled. If set to `False`, all development versions are removed from the `supported` field of the `/api-version` endpoint. Note that they are still listed in the `development` field, and continue to work normally.
 
+### OAuth
```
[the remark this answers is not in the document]
battermann: I added a link to the OAuth docs. Please decide if that is sufficient. I am struggling to find a between documenting this here in detail and having minimal redundancy.

docs/src/developer/reference/config-options.md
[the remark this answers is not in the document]
battermann: See comment above.

docs/src/how-to/install/oauth.md
battermann: The issue I have with linking to `config-options.md` is that this refers to the helm charts and is relevant for side admins and the section in `config-options.md` refers to the local settings for developers in `brig.integration.yaml`. I am lacking of a good strategy to resolve this.

services/brig/test/resources/oauth/jwk.json
battermann: I added a section in `config-option.md` on how to generate this key.

services/brig/test/integration/API/OAuth.hs
battermann: We need the file with the correct key for CI. However, we cannot provide the wrong key (which is only used in the tests) as a file in the CI environment without configuring it via the helm charts, and I don't think we want that. At least this is my understanding.

libs/libzauth/libzauth-c/src/lib.rs
```diff
@@ -253,9 +291,218 @@ impl From<io::Error> for ZauthResult {
...
+                status: OAuthResultStatus::NullArg,
+            };
+        }        
+        let bytes = unsafe { slice::from_raw_parts(token, token_len) };
+        let token = try_unwrap!(str::from_utf8(bytes));
+        let bytes = unsafe { slice::from_raw_parts(scope, scope_len) };
+        let scope = try_unwrap!(str::from_utf8(bytes));
+        let bytes = unsafe { slice::from_raw_parts(method, method_len) };
+        let method = str::from_utf8(bytes).unwrap();
+        let subject = try_unwrap!(verify_oauth_token(&jwk.0, token, scope, method));
```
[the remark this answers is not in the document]
battermann: I don't think that this has to be commented explicitly. It's basically what all the functions in this file are doing, translating inputs from c to rust, calling a function from the libzauth lib, error handling and translating the response back to c.

libs/wire-api/src/Wire/API/OAuth.hs
```diff
@@ -0,0 +1,596 @@
...
+import Test.QuickCheck (Arbitrary)
+import URI.ByteString
+import Web.FormUrlEncoded (Form (..), FromForm (..), ToForm (..), parseUnique)
+import Wire.API.Error
+import Wire.Arbitrary (GenericUniform (..))
+
+--------------------------------------------------------------------------------
+-- Types
+
+newtype RedirectUrl = RedirectUrl {unRedirectUrl :: URIRef Absolute}
```
fisx, line 54: lots of the types in this module are missing roundtrip tests.

docs/src/developer/reference/oauth.md
```diff
@@ -0,0 +1,429 @@
...
+  -d 'code=1395a1a44b72e0b81ec8fe6c791d2d3f22bc1c4df96857a88c3e2914bb687b7b&client_id=b9e65569-aa61-462d-915d-94c8d6ef17a7&grant_type=authorization_code&redirect_uri=https%3A%2F%2Fclient.example.com&client_secret=3f6fbd62835859b2bac411b2a2a2a54699ec56504ee32099748de3a762d41a2d'
+```
+
+Parameters:
+
+| Parameter       | Description                                                                             |
+| --------------- | --------------------------------------------------------------------------------------- |
+| `code`          | Required. The authorization code received from the authorization server.                |
+| `client_id`     | Required. The client identifier.                                                        |
+| `grant_type`    | Required. Value MUST be set to `authorization_code`.                                    |
```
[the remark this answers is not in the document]
battermann: I don't understand the question.

docs/src/developer/reference/oauth.md
```diff
@@ -0,0 +1,429 @@
...
+
+#### Refresh token
+
+refresh tokens are assigned to a user and a 3rd party app (OAuth client)
+
+a user can have more than one active refresh token for the same 3rd party app (e.g. they might use multiple devices, replace devices, or run multiple instances of the app somehow)
+
+the maximum number of active refresh tokens per user and app should be limited (what is a sensitive number, 10?)
+
+once a new token is requested and the limit is exceeded, the oldest refresh token will be deleted/invalidated
```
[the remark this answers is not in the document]
battermann: This depends, if the new token was requested for the same client instance with the same set of permissions, then there is no need to re-authorize.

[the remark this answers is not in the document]
battermann: I'll add this to the docs.

docs/src/developer/reference/oauth.md
```diff
@@ -0,0 +1,429 @@
...
+  ```http
+GET /oauth/client/:id HTTP/1.1
+```
+
+Example response:
+
+```json
+{
+  "client_id": "922fcc9c-07a5-4306-86ce-7b6838442372",
+  "application_name": "Outlook Calendar Extension",
```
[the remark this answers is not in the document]
fisx: never mind, we've established that `client` and `application` are synonyms.  i'm not entirely happy with that, but i don't think it should be fixed now.

resolved, thanks!

docs/src/developer/reference/oauth.md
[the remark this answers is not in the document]
fisx: you just found a counter-example on the phone, so never mind.

services/brig/docs/swagger.md
```diff
@@ -1,4 +1,17 @@
...
+[zauth](), and some support both [zauth]() and [oauth]().
+
+The end-points that require zauth are labelled so in the description
+below.  The end-points that support oauth as an alternative to zauth
+have the required oauth scopes listed in the same description.
+
+Futher reading:
+- https://docs.wire.com/how-to/install/oauth.html
+- https://github.com/wireapp/wire-server/blob/develop/libs/wire-api/src/Wire/API/Routes/Public.hs (search for HasSwagger instances)
+- `curl https://staging-nginz-https.zinfra.io/v4/api/swagger.json | jq '.security, .securityDefinitions`
```
fisx, line 14: this is not up to date any more; we don't have swagger docs about oauth, nor `HasSwagger` instances.  what we can do (at least in this PR) is mark an end-point as "can either do zauth, or oauth, or both".

in the long run, we may want to read the nginz config into a small TH blob that renders the information into swagger from there?  doesn't strike me as particularly hard or hard to maintain.

services/brig/src/Brig/API/OAuth.hs
```diff
@@ -0,0 +1,397 @@
...
+
+--------------------------------------------------------------------------------
+
+revokeOAuthAccountAccess :: UserId -> OAuthClientId -> (Handler r) ()
+revokeOAuthAccountAccess uid cid = do
+  rts <- lift $ wrapClient $ lookupOAuthRefreshTokens uid
+  for_ rts $ \rt -> when (oriClientId rt == cid) $ lift $ wrapClient $ deleteOAuthRefreshToken rt
+
+--------------------------------------------------------------------------------
+-- DB
```
fisx, line 298: Traditionally this would be located in `/services/brig/src/Brig/Data/OAuth.hs`.  I am leaning towards "please move it there for consistency, but I'm not feeling very strongly about it.

services/brig/src/Brig/API/OAuth.hs
```diff
@@ -0,0 +1,397 @@
...
+    q :: PrepQuery R (Identity OAuthAuthorizationCode) (OAuthClientId, UserId, C.Set OAuthScope, RedirectUrl)
+    q = "SELECT client, user, scope, redirect_uri FROM oauth_auth_code WHERE code = ?"
+
+deleteOAuthAuthorizationCode :: (MonadClient m) => OAuthAuthorizationCode -> m ()
+deleteOAuthAuthorizationCode code = retry x5 . write q $ params LocalQuorum (Identity code)
+  where
+    q :: PrepQuery W (Identity OAuthAuthorizationCode) ()
+    q = "DELETE FROM oauth_auth_code WHERE code = ?"
+
+lookupAndDeleteOAuthAuthorizationCode :: (MonadClient m) => OAuthAuthorizationCode -> m (Maybe (OAuthClientId, UserId, OAuthScopes, RedirectUrl))
```
fisx, line 343: would it make sense to store this code in hashed form as well?  we'd need the client id for that and restructure the tables, and it's only a short-lived secret, but i'm wondering...

services/brig/src/Brig/API/OAuth.hs
```diff
@@ -0,0 +1,397 @@
...
+  where
+    qInsertInfo :: PrepQuery W (OAuthRefreshTokenId, OAuthClientId, UserId, C.Set OAuthScope, UTCTime) ()
+    qInsertInfo = fromString $ "INSERT INTO oauth_refresh_token (id, client, user, scope, created_at) VALUES (?, ?, ?, ?, ?) USING TTL " <> show ttl
+
+    qInsertId :: PrepQuery W (UserId, OAuthRefreshTokenId) ()
+    qInsertId = fromString $ "INSERT INTO oauth_user_refresh_token (user, token_id) VALUES (?, ?) USING TTL " <> show ttl
+
+    determineOldestTokensToBeDeleted :: [OAuthRefreshTokenInfo] -> [OAuthRefreshTokenInfo]
+    determineOldestTokensToBeDeleted tokens =
+      take (length sorted - fromIntegral maxActiveTokens + 1) sorted
```
fisx, line 362:
```
length sorted === 13
maxActiveTokens === 10
```

then this would yield the oldes `13 - 10 + 1 === 4` tokens, but it should yield 3, no?

charts/nginz/templates/secret.yaml
```diff
@@ -15,4 +15,5 @@ data:
   {{- with .Values.secrets }}
   zauth.conf: {{ .zAuth.publicKeys | b64enc | quote }}
   basic_auth.txt: {{ .basicAuth | b64enc | quote }}
+  oauth_ed25519_pub.jwk: {{ .oAuth.publicKeys | b64enc | quote }}
```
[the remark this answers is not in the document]
battermann: Same as for `zauth.conf`, I guess it's ok?

services/brig/src/Brig/API/OAuth.hs
battermann: It's not possible because we have to be able to look up the code. If it is hashed with scrypt, that's not possible anymore.

services/brig/src/Brig/API/OAuth.hs
```diff
@@ -0,0 +1,397 @@
...
+
+    hashClientSecret :: MonadIO m => OAuthClientPlainTextSecret -> m Password
+    hashClientSecret = mkSafePassword . PlainTextPassword . toText . unOAuthClientPlainTextSecret
+
+--------------------------------------------------------------------------------
+
+getOAuthClient :: UserId -> OAuthClientId -> (Handler r) (Maybe OAuthClient)
+getOAuthClient _ cid = do
+  unlessM (Opt.setOAuthEnabled <$> view settings) $ throwStd $ errorToWai @'OAuthFeatureDisabled
+  lift $ wrapClient $ lookupOauthClient cid
```
fisx, line 103:
```suggestion
guardOAuthEnabled :: (Handler r) a -> (Handler r) ((Handler r) a)
guardOAuthEnabled action = do
  unlessM (Opt.setOAuthEnabled <$> view settings) $ throwStd $ errorToWai @'OAuthFeatureDisabled
  action

getOAuthClient :: UserId -> OAuthClientId -> (Handler r) (Maybe OAuthClient)
getOAuthClient _ cid = guardOAuthEnabled $ do
  lift $ wrapClient $ lookupOauthClient cid
```

Then this should be called in all handler functions in this module.  If you really have to respond with different errors on different end-points, then make the error an argument to the guard function.

services/brig/src/Brig/API/OAuth.hs
battermann: I agree that this should go to `/.../Data/` *if* we want to split the file. In the meantime, I think it should be ok to keep all of this in one place in a single file.

services/brig/src/Brig/API/OAuth.hs
```diff
@@ -0,0 +1,397 @@
...
+
+createNewOAuthAuthorizationCode :: UserId -> NewOAuthAuthorizationCode -> (Handler r) CreateOAuthCodeResponse
```
fisx, line 105: I'm still a bit unhappy with the naming.  `NewOAuthAuthorizationCode` is a thing you need to create an `AuthorizationCode`?  This is [step 2-5](http://localhost:3000/developer/reference/oauth.html#supported-oauth-flow), correct?  And arrow 5 is not really going from wire-server to oauth client, but from wire-server to browser and then to oauth app by redirect, no?  Maybe add those steps from the docs here to the haddocks as well?

fisx, line 105:
```suggestion
--------------------------------------------------------------------------------

createNewOAuthAuthorizationCode :: UserId -> NewOAuthAuthorizationCode -> (Handler r) CreateOAuthCodeResponse
```

(forgot one...)

services/brig/src/Brig/API/OAuth.hs
```diff
@@ -0,0 +1,397 @@
...
+    q :: PrepQuery R (Identity OAuthClientId) (Identity Password)
+    q = "SELECT secret FROM oauth_client WHERE id = ?"
+
+insertOAuthAuthorizationCode :: (MonadClient m) => Word64 -> OAuthAuthorizationCode -> OAuthClientId -> UserId -> OAuthScopes -> RedirectUrl -> m ()
+insertOAuthAuthorizationCode ttl code cid uid scope uri = do
+  let cqlScope = C.Set (Set.toList (unOAuthScopes scope))
+  retry x5 . write q $ params LocalQuorum (code, cid, uid, cqlScope, uri)
+  where
+    q :: PrepQuery W (OAuthAuthorizationCode, OAuthClientId, UserId, C.Set OAuthScope, RedirectUrl) ()
+    q = fromString $ "INSERT INTO oauth_auth_code (code, client, user, scope, redirect_uri) VALUES (?, ?, ?, ?, ?) USING TTL " <> show ttl
```
fisx, line 327: as below: should this be `scrypt`-ed?

libs/wire-api/src/Wire/API/Routes/Public/Brig.hs
```diff
@@ -259,6 +261,7 @@ type SelfAPI =
   Named
     "get-self"
     ( Summary "Get your own profile"
+        :> Description "OAuth scope: `read:self`"
```
fisx, line 264: I would have introduced a type `DescriptionOAuthScopes`.  Easier to refactor once we have something more correct-by-construction.

services/brig/docs/swagger.md
```diff
@@ -1,4 +1,17 @@
-## General
+## Authentication / Authorization
+
+The end-points in this API support differing authorization protocols:
+some are unauthenticated (`/api-version`, `/login`), some require
+[zauth](), and some support both [zauth]() and [oauth]().
```
fisx, line 5: oops, and i forgot to provide those links it seems.  sorry...

services/brig/src/Brig/API/OAuth.hs
```diff
@@ -0,0 +1,397 @@
...
+
+rand32Bytes :: MonadIO m => m AsciiBase16
+rand32Bytes = liftIO . fmap encodeBase16 $ randBytes 32
+
+--------------------------------------------------------------------------------
+
+revokeRefreshToken :: Member Jwk r => OAuthRevokeRefreshTokenRequest -> (Handler r) ()
+revokeRefreshToken req = do
+  key <- signingKey
+  info <- verifyTokenAndGetInfo key (ortrRefreshToken req)
```
fisx, line 272: `verifyTokenAndGetInfo` removes the authorization token.  in what context does the oauth client still have the authorization token, it's not deleted on wire-server yet, and the oauth client has a refresh token that it wants to revoke?

services/brig/src/Brig/API/OAuth.hs
```diff
@@ -0,0 +1,397 @@
...
+  let uid = oriUserId info
+      cid = oriClientId info
+  void $ getOAuthClient uid cid >>= maybe (throwStd $ errorToWai @'OAuthClientNotFound) pure
```
fisx, line 275: `uid` is provided by the client.  where do you validate that it's not something the client has made up?  does this end-point need to be oauth-authenticated?

services/brig/src/Brig/API/OAuth.hs
fisx: alternatively, we could reject creating the new tokens.  i would be less surprised by an error here if i were an oauth client.

services/brig/src/Brig/API/OAuth.hs
```diff
@@ -0,0 +1,397 @@
...
+    Right reqRT -> createAccessTokenWithRefreshToken reqRT
+
+createAccessTokenWithRefreshToken :: (Member Now r, Member Jwk r) => OAuthRefreshAccessTokenRequest -> (Handler r) OAuthAccessTokenResponse
+createAccessTokenWithRefreshToken req = do
+  unless (oartGrantType req == OAuthGrantTypeRefreshToken) $ throwStd $ errorToWai @'OAuthInvalidGrantType
+  key <- signingKey
+  info <- verifyTokenAndGetInfo key (oartRefreshToken req)
+  let uid = oriUserId info
+      cid = oriClientId info
+      scope = oriScopes info
```
fisx, line 156: match this with `@` in the function header?

services/brig/src/Brig/API/OAuth.hs
battermann: Let's say:
- max active tokens is 10
- there are currently 13 tokens (however that happened)
- user wants to create a new token
- then we need to delete the 4 oldest tokens so that after adding the new token we have a max of 10

services/brig/src/Brig/API/OAuth.hs
battermann: IDK, it's one line. If we extract this, we even increase LOC. Following DRY not worth it here. But this is only my opinion, can change it if you want.

services/brig/src/Brig/API/OAuth.hs
battermann: It can't be. We need to be able to look this up by the code.

services/brig/src/Brig/API/OAuth.hs
battermann: No, the user ID is part of the refresh token, which validity is verified. Also, the client needs to authenticate by providing their client secret.

services/brig/src/Brig/API/OAuth.hs
battermann: Followed an inconsistent convention here with the naming. Will change to `CrearteXRequest`.

services/brig/test/integration/API/OAuth.hs
```diff
@@ -0,0 +1,865 @@
...
+  get (nginz . paths ["self"] . authHeader oauthToken) !!! const 200 === statusCode
+
+testAccessResourceInsufficientScope :: Brig -> Nginz -> Http ()
+testAccessResourceInsufficientScope brig nginz = do
+  uid <- userId <$> createUser "alice" brig
+  let redirectUrl = mkUrl "https://example.com"
+  let scopes = OAuthScopes $ Set.fromList [WriteConversations]
+  (cid, secret, code) <- generateOAuthClientAndAuthorizationCode brig uid scopes redirectUrl
+  let accessTokenRequest = OAuthAccessTokenRequest OAuthGrantTypeAuthorizationCode cid secret code redirectUrl
+  accessToken <- createOAuthAccessToken brig accessTokenRequest
```
fisx, line 359: this could also be another helper function taking `scopes` as an argument, making the tests yet a bit more concise.

services/brig/test/integration/API/OAuth.hs
```diff
@@ -0,0 +1,865 @@
...
+  badRefreshToken <- liftIO $ do
+    claims <- verifyRefreshToken key (unOAuthToken $ oatRefreshToken accessToken)
+    OAuthToken <$> signRefreshToken badKey claims
+  let refreshAccessTokenRequest = OAuthRefreshAccessTokenRequest OAuthGrantTypeRefreshToken cid secret badRefreshToken
+  refreshOAuthAccessToken' brig refreshAccessTokenRequest !!! do
+    const 403 === statusCode
+    const "Forbidden" === statusMessage
+
+testRefreshTokenNoTokenId :: Opts -> Brig -> Http ()
+testRefreshTokenNoTokenId opts brig = do
```
fisx, line 495:
```suggestion
testRefreshTokenNoTokens opts brig = do
```

(it's not about the token id, but about access and refresh token missing, no?)

services/brig/test/integration/API/OAuth.hs
```diff
@@ -0,0 +1,865 @@
...
+  let redirectUrl = mkUrl "https://example.com"
+  let scopes = OAuthScopes $ Set.fromList [ReadSelf]
+  (cid, secret, _) <- generateOAuthClientAndAuthorizationCode brig uid scopes redirectUrl
+  key <- liftIO $ readJwk (fromMaybe "path to jwk not set" (Opt.setOAuthJwkKeyPair $ Opt.optSettings opts)) <&> fromMaybe (error "invalid key")
+  badRefreshToken <-
+    liftIO $
+      OAuthToken <$> do
+        rid :: OAuthRefreshTokenId <- randomId
+        sub <- maybe (error "creating sub claim failed") pure $ idToText rid ^? stringOrUri
+        let claims = emptyClaimsSet & claimSub ?~ sub
```
fisx, line 519: shouldn't this be a legit claim set with just one claim changed, rather than the empty claim set with one claim added?

services/brig/test/integration/API/OAuth.hs
```diff
@@ -0,0 +1,865 @@
...
+    liftIO $ assertEqual "apps" 0 (length apps)
+  void $ createOAuthApplicationWithAccountAccess brig uid
+  void $ createOAuthApplicationWithAccountAccess brig uid
+  do
+    apps <- listOAuthApplications brig uid
+    liftIO $ assertEqual "apps" 2 (length apps)
+  void $ createOAuthApplicationWithAccountAccess brig uid
+  do
+    apps <- listOAuthApplications brig uid
+    liftIO $ assertEqual "apps" 3 (length apps)
```
fisx, line 611: create an app with a different uid and check what first and second user see.

services/brig/test/integration/API/OAuth.hs
```diff
@@ -0,0 +1,865 @@
...
+      do
+        apps <- listOAuthApplications brig uid
+        liftIO $ assertEqual "apps" 1 (length apps)
+      revokeOAuthApplicationAccess brig uid cid3
+      do
+        apps <- listOAuthApplications brig uid
+        liftIO $ assertEqual "apps" 0 (length apps)
+    _ -> liftIO $ assertFailure "unexpected number of apps"
+
+testWriteConversationsSuccessInternal :: Brig -> Nginz -> Http ()
```
fisx, line 638: up to here, the order was the same in the test groups and the test definitions.  that was nice.  :)

fisx: also the 3 `...Internal` functions are not used anywhere, plus they are exact copies of the `Nginz` functions.  remove, i think?

services/brig/test/integration/API/OAuth.hs
```diff
@@ -0,0 +1,865 @@
...
+        "accessing a resource"
+        [ test m "success" $ testAccessResourceSuccessNginz b n,
+          test m "failure - insufficient scope" $ testAccessResourceInsufficientScope b n,
+          test m "failure - expired token" $ testAccessResourceExpiredToken b n,
+          test m "failure - nonsense token" $ testAccessResourceNonsenseToken n,
+          test m "failure - no token" $ testAccessResourceNoToken n,
+          test m "failure - invalid signature" $ testAccessResourceInvalidSignature o b n
+        ],
+      testGroup
+        "accessing resources (only testing happy path to ensure scopes are valid)"
```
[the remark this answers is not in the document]
battermann: I am undecided here.

services/brig/test/integration/API/OAuth.hs
```diff
@@ -0,0 +1,865 @@
...
+    cid <- randomId
+    uid <- randomId
+    state <- UUID.toText <$> liftIO nextRandom
+    let redirectUrl = mkUrl "https://example.com"
+    createOAuthCode brig uid (NewOAuthAuthorizationCode cid mempty OAuthResponseTypeCode redirectUrl state) !!! do
+      const 403 === statusCode
+      const (Just $ "access_denied") === (getLocation >=> getQueryParamValue "error")
+      const (Just $ cs state) === (getLocation >=> getQueryParamValue "state")
+      const (Just $ unRedirectUrl redirectUrl ^. pathL) === (fmap getPath . getLocation)
+      const Nothing === (getLocation >=> getQueryParamValue "code")
```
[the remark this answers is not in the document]
battermann: The error is propagated back via the redirect URL as a query parameter e.g. `?error=acces_denied`. I don't know if the client relies on the other values as well. In any case this is following the RFC.

services/brig/test/integration/API/OAuth.hs
battermann: We need the url again further down, and sometimes we need uid, too. I found this to be a good balance,

services/brig/test/integration/API/OAuth.hs
battermann: No, it's actually about a refresh token that does not contain an ID.

changelog.d/0-release-notes/pr-2989
```diff
@@ -0,0 +1 @@
+Before the deployment of wire-server the private and public keys for OAuth have to be provided for `brig` and `nginz` (see `docs/src/developer/reference/oauth.md` for more information)
```
fisx, line 1:
```suggestion
If you are using OAuth (`optSettings.setOAuthEnabled: true` in brig config): before the deployment of wire-server, the private and public keys for OAuth have to be provided for `brig` and `nginz` (see `docs/src/developer/reference/oauth.md` for more information)
```

services/galley/galley.cabal
```diff
@@ -229,6 +229,7 @@ library
     , http-types                 >=0.8
     , imports
     , insert-ordered-containers
+    , jose
```
fisx, line 232: this is not needed any more, right?

fisx: commit coming up!

charts/nginz/templates/conf/_nginx.conf.tpl
```diff
@@ -193,6 +193,7 @@ http {
 
     zauth_keystore {{ .Values.nginx_conf.zauth_keystore }};
     zauth_acl      {{ .Values.nginx_conf.zauth_acl }};
+    oauth_key      {{ .Values.nginx_conf.oauth_key }};
```
flokli, line 196: Is this a pubkey or a privkey? I guess it's a pubkey, but this should probably be named explicitly…

libs/libzauth/libzauth-c/src/zauth.h
```diff
@@ -39,24 +39,52 @@ typedef enum {
...
-ZauthResult    zauth_token_parse(uint8_t const * str, size_t len, ZauthToken **);
-ZauthResult    zauth_token_verify(ZauthToken const *, ZauthKeystore const *);
-ZauthTokenType zauth_token_type(ZauthToken const *);
-long           zauth_token_time(ZauthToken const *);
-uint8_t        zauth_token_version(ZauthToken const *);
-Range          zauth_token_lookup(ZauthToken const *, uint8_t);
-ZauthResult    zauth_token_allowed(ZauthToken const *, ZauthAcl const *, uint8_t const * path, size_t len, uint8_t * result);
-void           zauth_token_delete(ZauthToken *);
+OAuthResultStatus oauth_key_open(uint8_t const * fname, size_t len, OAuthJwk **);
+void              oauth_key_delete(OAuthJwk * store);
```
flokli, line 75: Same her, pubkey or private key?

libs/libzauth/libzauth/Cargo.toml
```diff
@@ -13,6 +13,15 @@ rustc-serialize = ">= 0.3"
...
+jwt-simple      = "0.11"
+serde_json      = "1.0"
+serde           = "1.0"
+thiserror       = "1.0"
+base64          = "0.21"
 
 [dev-dependencies]
 clap = ">= 2.0"
+
+[patch.crates-io.jwt-simple]
```
flokli, line 25:
```
[dependencies.jwt-simple]
git = "https://github.com/wireapp/rust-jwt-simple"
rev = "commit_id"
```

Also, can you add a comment stating why this was forked (what's missing from upstream), and did we send it along already?

services/nginz/integration-test/resources/oauth/ed25519_public.jwk
```diff
@@ -0,0 +1 @@
+{"kty":"OKP","crv":"Ed25519","x":"mhP-NgFw3ifIXGZqJVB0kemt9L3BtD5P8q4Gah4Iklc"}
```
flokli, line 1: There's a `services/brig/test/resources/oauth/ed25519.jwk`, and a `services/nginz/integration-test/resources/oauth/ed25519_public.jwk`. Maybe it makes sense to move these together and use it where needed?

charts/nginz/templates/conf/_nginx.conf.tpl
battermann: It's a public key. Was following the examples above: `zauth_keystore` is the path to the public zauth key.

flokli: Yes, zauth has the same naming problem, and we probably can't change that. I'd still prefer if try to keep the new stuff less confusingly named than staying consistent with the bad existing naming.

services/nginz/integration-test/conf/nginz/nginx.conf
```diff
@@ -118,6 +118,7 @@ http {
 
     zauth_keystore resources/zauth/pubkeys.txt;
     zauth_acl conf/nginz/zauth_acl.txt;
+    oauth_key resources/oauth/ed25519_public.jwk;
```
flokli, line 121:
```suggestion
    # needs to be kept in sync with services/brig/test/resources/oauth/ed25519.jwk
    oauth_key resources/oauth/ed25519_public.jwk;
```

services/brig/brig.integration.yaml
```diff
@@ -199,6 +199,12 @@ optSettings:
   # To only allow specific phone number prefixes to register uncomment and update the settings below
   # setAllowlistPhonePrefixes:
   #   - "+1555555"
+  setOAuthJwkKeyPair: test/resources/oauth/ed25519.jwk
```
flokli, line 202:
```suggestion
  # needs to be kept in sync with services/nginz/integration-test/resources/oauth/ed25519_public.jwk
  setOAuthJwkKeyPair: test/resources/oauth/ed25519.jwk
```

services/nginz/integration-test/resources/oauth/ed25519_public.jwk
flokli: Addressed in https://github.com/wireapp/wire-server/pull/2989#pullrequestreview-1338926862.

services/nginz/integration-test/conf/nginz/nginx.conf
battermann: it has conflicts now, I'll add it manually.

libs/wire-api/src/Wire/API/Routes/Public/Galley/Conversation.hs
```diff
@@ -342,14 +344,15 @@ type ConversationAPI =
                :> Description "This returns 201 when a new conversation is created, and 200 when the conversation already existed"
                :> ZLocalUser
                :> ZOptClient
-               :> ZConn
+               :> ZOptConn
```
mdimjasevic, line 347: I'm sure this was justified, but can you clarify why is the connection ID now optional? Only the local user can call this endpoint, hence I believe we always have the connection ID.

CC: @fisx , @battermann

battermann: When this endpoint is called by an OAuth application, we do not have a connection ID.

PR #2991, opened 2023-01-16
[REFACTORING] ID tag expand abbreviation to full name

battermann:
## Checklist

### DIFF
--- a/libs/types-common/src/Data/Id.hs
+++ b/libs/types-common/src/Data/Id.hs
@@ -187,7 +187,7 @@ instance Read (Id a) where
 instance FromByteString (Id a) where
   parser = do
     match <-
-      -- we only want the matching part of the ByteString, so the parser doesn'Team
+      -- we only want the matching part of the ByteString, so the parser doesn't
       -- consume additional input.
       -- This allows the parser to be composed.
       matching $ do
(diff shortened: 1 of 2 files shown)


libs/types-common/src/Data/Id.hs
```diff
@@ -87,54 +87,62 @@ import Servant (FromHttpApiData (..), ToHttpApiData (..))
...
-idTagName A = "Asset"
-idTagName C = "Conv"
-idTagName I = "Invitation"
-idTagName U = "User"
-idTagName P = "Provider"
-idTagName S = "Service"
-idTagName T = "Team"
-idTagName STo = "ScimToken"
+idTagName Asset = "Asset"
+idTagName Conversation = "Conv"
```
elland, line 102: Kind of annoying having this to basically replace a `Show` instance, now after the refactor, but don't want to pile up more stuff for you to do 👍

libs/types-common/src/Data/Id.hs
```diff
@@ -179,7 +187,7 @@ instance Read (Id a) where
 instance FromByteString (Id a) where
   parser = do
     match <-
-      -- we only want the matching part of the ByteString, so the parser doesn't
+      -- we only want the matching part of the ByteString, so the parser doesn'Team
```
elland, line 190: Regex is hard <3

elland (changes requested):
Some docs got mangled b/c of regex replaces.

libs/types-common/src/Data/Id.hs
akshaymankar, line 190: collateral regex damange :sweat_smile:

akshaymankar (commented):
Looks good apart from the comments being attacked by the regex

libs/types-common/src/Data/Id.hs
battermann: oops

elland:
Love seeing this kind of PR <3

PR #2994, opened 2023-01-17
Refactor Writetime

smatting:
This PR refactors multiple definitions of the `Writetime a` type to a single one, which wraps a UTCTime value.
This means no loss of precision: The function `writetimeToInt64` can recover the original integer that cassandra uses to represent writetimes.

## Checklist

### DIFF
--- a/libs/cassandra-util/src/Cassandra/Util.hs
+++ b/libs/cassandra-util/src/Cassandra/Util.hs
@@ -14,27 +14,27 @@
 --
 -- You should have received a copy of the GNU Affero General Public License along
 -- with this program. If not, see <https://www.gnu.org/licenses/>.
+{-# LANGUAGE NumericUnderscores #-}
 
 module Cassandra.Util
-  ( writeTimeToUTC,
-    defInitCassandra,
-    Writetime,
+  ( defInitCassandra,
+    Writetime (..),
+    writetimeToInt64,
   )
 where
 
-import Cassandra (ClientState, Keyspace (Keyspace), init)
+import Cassandra (ClientState, init)
+import Cassandra.CQL
 import Cassandra.Settings (defSettings, setContacts, setKeyspace, setLogger, setPortNumber)
+import Data.Aeson
+import Data.Fixed
 import Data.Text (unpack)
-import Data.Time (UTCTime)
-import Data.Time.Clock.POSIX (posixSecondsToUTCTime)
+import Data.Time (UTCTime, nominalDiffTimeToSeconds)
+import Data.Time.Clock (secondsToNominalDiffTime)
+import Data.Time.Clock.POSIX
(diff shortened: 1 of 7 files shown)


services/brig/test/integration/API/Internal.hs
```diff
@@ -370,3 +374,20 @@ getFeatureConfig galley uid = do
...
+testWritetimeRepresentation :: forall m. TestConstraints m => Opt.Opts -> Manager -> Cass.ClientState -> Brig -> Endpoint -> Galley -> m ()
+testWritetimeRepresentation _ _mgr db brig _brigep _galley = do
+  quid <- userQualifiedId <$> randomUser brig
+  let uid = qUnqualified quid
+
+  ref <- fromJust <$> (runIdentity <$$> Cass.runClient db (C.query1 q1 (C.params C.LocalQuorum (Identity uid))))
+
+  wt <- fromJust <$> (runIdentity <$$> Cass.runClient db (C.query1 q2 (C.params C.LocalQuorum (Identity uid))))
+
+  liftIO $ assertEqual "ts representaiton do not match" ref (writetimeToInt64 wt)
```
mdimjasevic, line 387: What's "ts"? There's a typo too in "representaiton".

libs/cassandra-util/src/Cassandra/Util.hs
```diff
@@ -44,3 +44,37 @@ defInitCassandra ks h p lg =
...
     $ defSettings
+
+-- | Read cassandra's writetimes https://docs.datastax.com/en/dse/5.1/cql/cql/cql_using/useWritetime.html
+-- as UTCTime values without any loss of precision
+newtype Writetime a = Writetime {writetimeToUTC :: UTCTime}
+
+instance Cql (Writetime a) where
+  ctype = Tagged BigIntColumn
+  toCql = CqlBigInt . writetimeToInt64
+  fromCql (CqlBigInt n) =
```
mdimjasevic, line 55: Should we have a roundtrip test or two for these two?

mdimjasevic (approved):
Looks good! Just a few minor comments inlined.

libs/cassandra-util/src/Cassandra/Util.hs
smatting: The integration test already shows that `toCql . fromCql` is the identity. The other direction (`toCql . fromCql`) we don't need, because we never convert any arbitrary `UTCTime` values to Writetime values, because `Writetime` values always originate in the database.

services/brig/test/integration/API/Internal.hs
smatting: Fixed it in https://github.com/wireapp/wire-server/pull/2994/commits/207971cb567c7a40f9b35d4f027b2d6154d3ab7a

PR #2997, opened 2023-01-17
extend on-new-remote-conversation by the subconv ID

stefanwire:
When a subconversation is created, all backends involved in the main conversation need to be notified, so that they create a group ID → subconversation mapping. Otherwise, commits coming from clients on those backends cannot be routed.

Similarly, when a user from a new backend B joins a conversation containing non-empty subconversations, it needs to be informed of those subconversations, so that (external) commits for those subconversations coming from clients in B can be routed.

https://wearezeta.atlassian.net/browse/FS-1451

## TODO

 - [x] Split federation endpoint into `on-new-remote-conversation` and `on-new-remote-subconversation`
 - [x] Call `on-new-remote-subconversation` when a new subconversation is created
 - [x] Call `on-new-remote-subconversation` for all existing subconversations when a new backend gets involved
 - [x] Call `on-new-remote-subconversation` when a subconversation is reset
 - [x] Integration tests
 - [x] Subconversation End-to-end tests
 - [ ] (optional) Split `executeProposalAction` into one function for conversations and one for subconversations

## Checklist

### DIFF
--- a/libs/wire-api-federation/src/Wire/API/Federation/API/Galley.hs
+++ b/libs/wire-api-federation/src/Wire/API/Federation/API/Galley.hs
@@ -220,6 +220,8 @@ ccRemoteOrigUserId cc = qualifyAs (ccCnvId cc) (ccOrigUserId cc)
 data NewRemoteConversation = NewRemoteConversation
   { -- | The conversation ID, local to the backend invoking the RPC.
     nrcConvId :: ConvId,
+    -- | The subconversation, if set.
+    nrcSubConvId :: Maybe SubConvId,
     -- | The conversation protocol.
     nrcProtocol :: Protocol
   }
--- a/libs/wire-api-federation/src/Wire/API/Federation/API/Galley.hs
+++ b/libs/wire-api-federation/src/Wire/API/Federation/API/Galley.hs
@@ -56,6 +56,7 @@ type GalleyApi =
     -- This endpoint is called the first time a user from this backend is
     -- added to a remote conversation.
     :<|> FedEndpoint "on-new-remote-conversation" NewRemoteConversation EmptyResponse
+    :<|> FedEndpoint "on-new-remote-subconversation" NewRemoteSubConversation EmptyResponse
     :<|> FedEndpoint "get-conversations" GetConversationsRequest GetConversationsResponse
     -- used by the backend that owns a conversation to inform this backend of
     -- changes to the conversation
@@ -63,7 +64,8 @@ type GalleyApi =
     :<|> FedEndpointWithMods
            '[ MakesFederatedCall 'Galley "on-conversation-updated",
               MakesFederatedCall 'Galley "on-mls-message-sent",
-              MakesFederatedCall 'Galley "on-new-remote-conversation"
+              MakesFederatedCall 'Galley "on-new-remote-conversation",
+              MakesFederatedCall 'Galley "on-new-remote-subconversation"
             ]
            "leave-conversation"
            LeaveConversationRequest
@@ -83,7 +85,8 @@ type GalleyApi =
     :<|> FedEndpointWithMods
            '[ MakesFederatedCall 'Galley "on-mls-message-sent",
               MakesFederatedCall 'Galley "on-conversation-updated",
-              MakesFederatedCall 'Galley "on-new-remote-conversation"
+              MakesFederatedCall 'Galley "on-new-remote-conversation",
+              MakesFederatedCall 'Galley "on-new-remote-subconversation"
             ]
            "on-user-deleted-conversations"
            UserDeletedConversationsNotification
@@ -91,7 +94,8 @@ type GalleyApi =
     :<|> FedEndpointWithMods
            '[ MakesFederatedCall 'Galley "on-conversation-updated",
               MakesFederatedCall 'Galley "on-mls-message-sent",
-              MakesFederatedCall 'Galley "on-new-remote-conversation"
+              MakesFederatedCall 'Galley "on-new-remote-conversation",
+              MakesFederatedCall 'Galley "on-new-remote-subconversation"
             ]
            "update-conversation"
            ConversationUpdateRequest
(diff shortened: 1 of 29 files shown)


services/galley/src/Galley/API/Federation.hs
```diff
@@ -203,14 +203,27 @@ onConversationCreated domain rc = do
...
   Domain ->
   F.NewRemoteConversation ->
   Sem r EmptyResponse
 onNewRemoteConversation domain nrc = do
   -- update group_id -> conv_id mapping
   for_ (preview (to F.nrcProtocol . _ProtocolMLS) nrc) $ \mls ->
-    E.setGroupIdForConversation (cnvmlsGroupId mls) (Qualified (F.nrcConvId nrc) domain)
+    case F.nrcSubConvId nrc of
+      Just subConvId ->
+        setGroupIdForSubConversation
```
elland, line 219: Probably out of scope but we should maybe be more consistent with imports. Pretty much every Effect interpreter everywhere is imported qualified as `E`, with a few exceptions here and there.

libs/wire-api-federation/src/Wire/API/Federation/API/Galley.hs
```diff
@@ -220,6 +220,8 @@ ccRemoteOrigUserId cc = qualifyAs (ccCnvId cc) (ccOrigUserId cc)
 data NewRemoteConversation = NewRemoteConversation
   { -- | The conversation ID, local to the backend invoking the RPC.
     nrcConvId :: ConvId,
+    -- | The subconversation, if set.
+    nrcSubConvId :: Maybe SubConvId,
```
elland, line 224: Not a fan of encoding things with such ambiguity. It's a bit unintuitive that this maps to a sub-conversation if this value is present. Could we change this into a sum type instead?

pcapriotti: It is a sum type (`Maybe`). I'm not sure where the ambiguity is. We are using the same representation for events.

mdimjasevic: The way we handled a need for a change like this is to use the `ConvOrSubConvId` type instead of `ConvId`, and that type can carry either of the conv ID or the full subconv ID (maybe modulo the domain). This would be a change that affects the federation API beyond MLS, so I'd recommend a separate PR to `develop` just to change these types/API and then rebase the PR to `mls` on top of it.

pcapriotti: I'm still not sure what we are looking to achieve. What is the problem with the current representation?

mdimjasevic: There's nothing wrong. It will get the job done. It's just not consistent with how it was done before, but maybe we shouldn't worry about it.

Another note: the way I did such a change before is to split the request type into two: one for Proteus (which doesn't feature subconversations) and one for MLS. It could be done here too.

services/galley/src/Galley/API/Federation.hs
stefanwire: Fixed it.

libs/wire-api-federation/src/Wire/API/Federation/API/Galley.hs
stefanwire: It would be the [first commit](https://github.com/wireapp/wire-server/pull/2997/commits/3e6e87138fd9063653360cf16a83856e666146d8) in `develop` and [the second](https://github.com/wireapp/wire-server/pull/2997/commits/ba0b055bafc45a46df4a52e226e51659220c85f6) in `mls`. This would require `SubConversationStore` in `develop` which doesn't exist there yet.

mdimjasevic: > It would be the [first commit](https://github.com/wireapp/wire-server/pull/2997/commits/3e6e87138fd9063653360cf16a83856e666146d8) in `develop` and [the second](https://github.com/wireapp/wire-server/pull/2997/commits/ba0b055bafc45a46df4a52e226e51659220c85f6) in `mls`. This would require `SubConversationStore` in `develop` which doesn't exist there yet.

Unfortunately, the links lead to nowhere now (I suppose due to a rebase). The split of the message sending types that I was referring to was in PR #2925.

services/galley/src/Galley/API/MLS/SubConversation.hs
```diff
@@ -287,6 +293,17 @@ deleteLocalSubConversation qusr lcnvId scnvId dsc = do
 
     -- the following overwrites any prior information about the subconversation
     Eff.createSubConversation cnvId scnvId cs (Epoch 0) newGid Nothing
+    pure (scMLSData sconv)
+
+  -- notify all backends that the subconversation has a new ID
+  let remotes = Set.fromList (map (void . rmId) (convRemoteMembers cnv))
```
smatting, line 299: nit-pick: in other places we've used `bucketRemote` instead of `Set.fromList`

services/galley/src/Galley/API/MLS/Message.hs
```diff
@@ -1343,6 +1369,25 @@ executeProposalAction qusr con lconvOrSub action = do
   for_ (Map.assocs (paRemove action)) $ \(qtarget, clients) -> do
     removeMLSClients (cnvmlsGroupId mlsMeta) qtarget (Map.keysSet clients)
 
+  -- TODO: only do this for new subconversations (i.e. Epoch 1)
+  -- call `on-new-remote-conversation` on all the remote backends involved in
+  -- the main conversation
```
smatting, line 1374: still left to do?

services/galley/test/integration/API/MLS.hs
```diff
@@ -226,6 +226,7 @@ tests s =
             [ test s "get subconversation of remote conversation - member" (testGetRemoteSubConv True),
               test s "get subconversation of remote conversation - not member" (testGetRemoteSubConv False),
               test s "join remote subconversation" testJoinRemoteSubConv,
+              test s "backends and notified about subconvs when a user joins" testRemoteSubConvNotificationWhenUserJoins,
```
smatting, line 229:
```suggestion
              test s "backends are notified about subconvs when a user joins" testRemoteSubConvNotificationWhenUserJoins,
```

smatting (approved):

PR #3003, opened 2023-01-18
internal endpoints swagger (brig, cannon)

supersven:
Swagger docs for internal endpoints. This PR contains swagger docs for `brig` and `cannon`. The other Haskell services will follow. (I thought it might be better to reach out for feedback early instead of waiting until this PR gets really big...)

I decided to use `schema-profunctor` where feasible.

[Ticket](https://wearezeta.atlassian.net/browse/SQPIT-1184)

### Screenshot
image

## Checklist


libs/types-common/src/Data/Json/Util.hs
```diff
@@ -67,14 +68,14 @@ import qualified Data.Swagger as S
 import qualified Data.Text as Text
 import qualified Data.Text.Encoding as Text
 import qualified Data.Text.Encoding.Error as Text
+-- for UTCTime
```
fisx, line 71: explicit import, then?

fisx: oh, i just realized this has moved a little.  is it in the right place now or was it before?  either way it's maybe dangerous to have comments in import lists rendered by ormolu?

supersven: I'll remove the comment. When we start to add such things, where does it end... Also, HLS can show this information.

fisx (commented):
can you add the relevant links [here](https://docs.wire.com/understand/api-client-perspective/index.html) please?

i'll read the rest of this PR soon!

docs/src/understand/api-client-perspective/swagger.md
```diff
@@ -12,9 +19,33 @@ looking for, double-check the old.
 
 ## New docs
 
-These docs show swagger 2.0:
-
-[new staging swagger page](https://staging-nginz-https.zinfra.io/api/swagger-ui/)
+The Swagger documentation for public endpoints depends on the API version.
+Though, the version prefix in URL paths works for internal (private) endpoints,
+too, they are not versioned.
```
fisx, line 24: Isn't the internal API just as versioned, and we merely don't make use of it yet, ie. all versions happen to look exactly alike?

docs/src/understand/api-client-perspective/swagger.md
```diff
@@ -12,9 +19,33 @@ looking for, double-check the old.
...
+    endpoints](https://staging-nginz-https.zinfra.io/v1/api/swagger-ui/)
+- Version `v2`:
+    - [new staging swagger page - **public**
+    endpoints](https://staging-nginz-https.zinfra.io/v2/api/swagger-ui/)
+- Version `v3` (current *development* version):
+    - [new staging swagger page - **public**
+    endpoints](https://staging-nginz-https.zinfra.io/v3/api/swagger-ui/)
+- Unversioned
+    - [new staging swagger page - **internal** (private)
+    endpoints](https://staging-nginz-https.zinfra.io/api-internal/swagger-ui/)
```
fisx, line 43: I guess it's silly to generate this list, but maybe we can write it up so that the reader can construct the correct URL themselves?  (put an `N` in there and ask the user to replace it, and find the latest version number under `/api-versions`?)

That's slightly more awkward, but also a lot more future-proof then this.

Also, I think internal API is just as versions as external (or should be); see above.

fisx: wait -- does skippping the version path segment give you V0 or the most recent version?  V0 would be consistent with all the other paths, most recent would be most useful, "error - please pick a version number" would be most robust and least misleading.  i'm in favor of either useful or robust.

supersven: Paolo (not using GitHub mention by intention) explained this detail in https://docs.wire.com/developer/developer/api-versioning.html?highlight=versioning#making-requests-to-a-particular-version :

> To support clients that have not yet implemented versioning, backends that support version 0 will also accept unversioned requests, which will automatically be rewritten to /v0/ requests.

I agree with you proposal regarding explaining the URL pattern vs. a prepared list. I'll adjust the docs accordingly.

docs/src/understand/api-client-perspective/swagger.md
supersven: :thinking: 

I'll figure that out.

docs/src/understand/api-client-perspective/swagger.md
fisx: yes, i'm aware of how "no version segment" works for the rest of the api, i just thought since clients won't have to call swagger code in the wild we could get away with giving /api/swagger-ui and /api/swagger-docs a different semantics.

anyway, the important thing is that the docs explain it correctly and for all future versions.

libs/types-common/src/Data/Json/Util.hs
```diff
@@ -157,6 +157,18 @@ class ToJSONObject a where
...
+-- Aeson Object
+
+instance S.ToParamSchema A.Object where
+  toParamSchema _ =
+    mempty & S.type_ ?~ S.SwaggerString
+
+instance ToSchema A.Object where
+  schema =
+    named "Object" $
+      id .= jsonObject
```
fisx, line 170:
```suggestion
objectToParamSchema :: ...
objectToParamSchema _ = mempty & S.type_ ?~ S.SwaggerString

objectSchema :: ...
objectSchema = named "Object" $ id .= jsonObject      
```

i think this could work for `Schema`, not sure about `ParamSchema`.  if it works, it'll reduce the need for orphans, and the places where we use this are pretty few.

libs/wire-api/src/Wire/API/SwaggerServant.hs
```diff
@@ -0,0 +1,97 @@
...
+      prependTitle' :: Text -> S.PathItem -> S.PathItem
+      prependTitle' tagName it =
+        it
+          & (S.get . traverse . S.tags) <>~ tag tagName
+          & (S.put . traverse . S.tags) <>~ tag tagName
+          & (S.post . traverse . S.tags) <>~ tag tagName
+          & (S.delete . traverse . S.tags) <>~ tag tagName
+          & (S.options . traverse . S.tags) <>~ tag tagName
+          & (S.patch . traverse . S.tags) <>~ tag tagName
+          & (S.head_ . traverse . S.tags) <>~ tag tagName
```
fisx, line 57: would [allOperations](https://hoogle.zinfra.io/file/nix/store/568lw38vf5mdj3f8ivgbp9sfrbmczxrk-swagger2-2.8.6-doc/share/doc/swagger2-2.8.6/html/Data-Swagger-Operation.html#v:allOperations) help here?

libs/wire-api/src/Wire/API/SwaggerServant.hs
```diff
@@ -0,0 +1,97 @@
...
+  clientWithRoute proxyM _ = clientWithRoute proxyM (Proxy @api)
+  hoistClientMonad proxyM _ = hoistClientMonad proxyM (Proxy @api)
+
+-- | A type-level tag that lets us omit any branch from Swagger docs.
+--
+-- Those are likely to be:
+--
+--   * Endpoints for which we can't generate Swagger docs.
+--   * The endpoint that serves Swagger docs.
+--   * Internal endpoints.
```
fisx, line 83:
```suggestion
-- FUTUREWORK(fisx): this is currently only used for the spar internal api
-- and spar scim, and we should probably eliminate those uses and this combinator.
-- it's only justification is laziness.
```

fisx (approved):
neat!  i have lots of ideas, but none of them are blockers (i think?)

libs/wire-api/src/Wire/API/SwaggerServant.hs
supersven: Helped :+1:

libs/wire-api/src/Wire/API/Routes/Public/Spar.hs
```diff
@@ -35,6 +35,7 @@ import Web.Scim.Class.User as Scim.User
 import Wire.API.Error
 import Wire.API.Error.Brig
 import Wire.API.Routes.Public
+import Wire.API.SwaggerServant
```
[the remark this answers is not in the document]
supersven: Well, it's used! :smile_cat: 

I moved `OmitDocs` into `Wire.API.SwaggerServant`.

libs/types-common/src/Data/Json/Util.hs
supersven: Well, the issue isn't the Swagger of `Object`, it's that `List1 Object` has to fulfill this constraint: `instance ToSchema a => ToSchema (List1 a)`

I don't know a good way out: One the one hand we'll likely want to show the Swagger of the payload of `List1`. On the other hand, this orphan is a bit unfortunate... :roll_eyes: 

https://github.com/wireapp/wire-server/pull/3003/files#diff-04dac01f8df7b94a25e2d4350eb8e448d47ea92bfb5fe15fa0ba7db8720334aeR67

docs/src/understand/api-client-perspective/swagger.md
```diff
@@ -1,15 +1,81 @@
...
+code), and they are complete up to bots/services and event notification
 payloads (as of 2023-01-16).
 
-Please check the new docs first, and if you can't find what you're
-looking for, double-check the old.
+There are several ways to interpret this kind of documentation:
+
+- Read it as a reference
+- Generate client code from it
+- Interactively explore the API by making requests
```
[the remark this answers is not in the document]
supersven: Not correct :wink: 

image

The *Authorize* button works due to `HasSwagger` instance of `ZUser`: https://hoogle.zinfra.io/file/nix/store/2gh8v2np9adpkrnbpw5ssz8nc1jld0l9-wire-api-0.1.0-doc/share/doc/wire-api-0.1.0/html/src/Wire.API.Routes.Public.html#ZUser

docs/src/understand/api-client-perspective/swagger.md
```diff
@@ -1,15 +1,81 @@
...
+    endpoints](https://staging-nginz-https.zinfra.io/v2/api/swagger-ui/)
+- Version `v3`:
+    - [new staging swagger page - **public**
+    endpoints](https://staging-nginz-https.zinfra.io/v3/api/swagger-ui/)
+- Unversioned
+    - [new staging swagger page - **public**
+    endpoints](https://staging-nginz-https.zinfra.io/api/swagger-ui/)
+
+The first part of the URL's path is the version. No specified version means
+Swagger docs of the *latest* API version. This differs from other API endpoints
```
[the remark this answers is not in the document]
supersven: I've extracted this into a new ticket. https://wearezeta.atlassian.net/browse/SQPIT-1617

(I'm a bit eager to get this one finally merged ... :smile_cat: )

libs/types-common/src/Data/Json/Util.hs
fisx: ah, i see.  then orphans it is i guess!  :+1:

fisx (approved):

PR #2935, opened 2022-12-19
[DO NOT MERGE] add roman chart

amitsagtani97:
## Checklist

- Added helm chart for Roman service.
- Added postgres 13 chart as optional external dependency

Example values file for both charts can be found at - https://github.com/wireapp/wire-server-deploy/pull/597

 - [x] Add a new entry in an appropriate subdirectory of `changelog.d`
 - [x] Read and follow the [PR guidelines](https://docs.wire.com/developer/developer/pr-guidelines.html)

PR #3012, opened 2023-01-24
bitnami/postgresql wrapper to use in different bots and related services

amitsagtani97:
Ticket - https://wearezeta.atlassian.net/browse/SQPIT-574

This the wrapper chart for bitnami/postgresql that can be used in different bots and related services.
An example use case - https://github.com/wireapp/wire-server/pull/2935 

## Checklist


jschaul (changes requested):
This should not be called `-external`. Since postgres from this chart runs inside kubernetes. 

(The `cassandra-external` chart forwards traffic to cassandra deployed *outside* kubernetes on bare VMs. This is different.)

I'm not sure why we need a hand-rolled postgres helm chart here. There are many postgres charts maintained by the community, for example from bitnami, here: https://github.com/bitnami/charts/tree/main/bitnami/postgresql

Let's use their existing chart instead of needing to maintain a helm chart on software we do not develop.

If you need to re-host this helm chart as part of wire-server (why though?), you could require a third-party helm chart using a requirements.yaml file. But I don't see the benefit really, just use the third-party postgres chart wherever you wish to make use of it.

Your PR does not link to any issue / jira ticket so unfortunately I cannot know more context.

amitsagtani97:
> This should not be called `-external`. Since postgres from this chart runs inside kubernetes.
> 
> (The `cassandra-external` chart forwards traffic to cassandra deployed _outside_ kubernetes on bare VMs. This is different.)
> 
> I'm not sure why we need a hand-rolled postgres helm chart here. There are many postgres charts maintained by the community, for example from bitnami, here: https://github.com/bitnami/charts/tree/main/bitnami/postgresql
> 
> Let's use their existing chart instead of needing to maintain a helm chart on software we do not develop.
> 
> If you need to re-host this helm chart as part of wire-server (why though?), you could require a third-party helm chart using a requirements.yaml file. But I don't see the benefit really, just use the third-party postgres chart wherever you wish to make use of it.
> 
> Your PR does not link to any issue / jira ticket so unfortunately I don't see the added value from this PR, if there might be a usecase I can't easily imagine, please provide more context.

I was little worried if the latest postgresql versions will work with the existing bots, and in the Bitnami chart I tried to replace the image and tags to use older postgresql version, but that didn't worked.
But, the latest versions of Postgresql(at least 14) works with bots, so removing this custom chart and replacing with Bitnami one as you suggested.

charts/postgresql/values.yaml
```diff
@@ -0,0 +1,14 @@
...
+  replicationPassword: externalPostgresql
+  postgresPassword: externalPostgresql
+  username: externalPostgresql
+  password: externalPostgresql
+  database: externalPostgresql
+readReplicas:
+  name: read
+  replicaCount: 2
+replication:
+  numSynchronousReplicas: 2
```
jschaul, line 14: why do you need more than one replica in your use case (bots)? Would it not be sufficient to have one single replica (and no read replicas)?

Perhaps the first line `architecture: replication` could be changed to `architecture: standalone` (already the default). If not, please motivate why you need higher availability here at this level, and what this chart is used for (and also where it is used).

charts/postgresql/values.yaml
```diff
@@ -0,0 +1,14 @@
+architecture: replication
+global:
+  storageClass: local-storage
+auth:
+  replicationPassword: externalPostgresql
+  postgresPassword: externalPostgresql
+  username: externalPostgresql
+  password: externalPostgresql
+  database: externalPostgresql
```
jschaul, line 9: Ugh. Can you remove the whole auth block? There should never be any default passwords in plaintext. The user of a helm chart should specify their own passwords. Same for database name and username. Please remove the whole auth block here.

charts/postgresql/values.yaml
```diff
@@ -0,0 +1,14 @@
+architecture: replication
+global:
```
jschaul, line 2: All of the values inside `values.yaml` currently have no effect. If you wish to override the defaults provided inside the upstream values.yaml, here: https://github.com/bitnami/charts/blob/main/bitnami/postgresql/values.yaml, then you need to indent them under a top-level `postgresql:` section.

jschaul (changes requested):

charts/postgresql/values.yaml
amitsagtani97: Fixed 👍

charts/postgresql/values.yaml
amitsagtani97: fixed

PR #3026, opened 2023-01-26
charts/galley/configmap: remove duplicate feature flag searchVisibilityInbound

lucendio:
Flag configuration already exists in [LOCs L90-L93](https://github.com/wireapp/wire-server/blob/develop/charts/galley/templates/configmap.yaml#L90-L93). Accidentally introduced in https://github.com/wireapp/wire-server/pull/2494.


jschaul (approved):
Thanks for spotting this!

PR #3035, opened 2023-01-30
fix conversation list parser in stern

battermann:
## Checklist

### DIFF
--- a/tools/stern/src/Stern/API.hs
+++ b/tools/stern/src/Stern/API.hs
@@ -402,15 +402,15 @@ getUserData uid = do
   convs <- Intra.getUserConversations uid
   clts <- Intra.getUserClients uid
   notfs <- Intra.getUserNotifications uid
-  consent <- (Intra.getUserConsentValue uid <&> Just) `catchE` \_ -> pure Nothing
-  consentLog <- (Intra.getUserConsentLog uid <&> Just) `catchE` \_ -> pure Nothing
+  consent <- (Intra.getUserConsentValue uid <&> toJSON @ConsentValue) `catchE` (pure . String . cs . show)
+  consentLog <- (Intra.getUserConsentLog uid <&> toJSON @ConsentLog) `catchE` (pure . String . cs . show)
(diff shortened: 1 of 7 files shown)


tools/stern/src/Stern/API.hs
```diff
@@ -402,15 +402,15 @@ getUserData uid = do
   convs <- Intra.getUserConversations uid
   clts <- Intra.getUserClients uid
   notfs <- Intra.getUserNotifications uid
-  consent <- Intra.getUserConsentValue uid
-  consentLog <- Intra.getUserConsentLog uid
+  consent <- (Intra.getUserConsentValue uid <&> Just) `catchE` \_ -> pure Nothing
```
fisx, line 405: do these catch internal errors as well?  do we want that?

tools/stern/src/Stern/Intra.hs
```diff
@@ -744,12 +746,12 @@ getUserConversations uid = do
...
             ( method GET
                 . header "Z-User" (toByteString' uid)
-                . path "/conversations"
+                . path "/v2/conversations"
                 . queryItem "size" (toByteString' batchSize)
                 . maybe id (queryItem "start" . toByteString') start
                 . expect2xx
             )
-      parseResponse (mkError status502 "bad-upstream") r
+      unVersioned @'V2 <$> parseResponse (mkError status502 "bad-upstream") r
```
fisx, line 754: this wouldn't have happened if we used servant-client for calling internal api end-points...  :)

fisx (approved):

tools/stern/src/Stern/API.hs
battermann: I don't know, I just noticed that these fail and make the whole request fail, which is unnecessary, IMO.

battermann: made a change, see comment below

battermann:
Internally, stern makes multiple requests, some of which can fail and in this case the error will be shown in the response without making the overall request fail.
image

PR #3036, opened 2023-01-30
Make stern assert backend api version

fisx:
## Checklist

### DIFF
--- a/tools/stern/src/Stern/Intra.hs
+++ b/tools/stern/src/Stern/Intra.hs
@@ -66,6 +66,7 @@ import Bilge hiding (head, options, requestId)
 import Bilge.RPC
 import Brig.Types.Intra
 import Control.Error
+import Control.Exception (ErrorCall (ErrorCall))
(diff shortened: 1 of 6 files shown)


fisx:
(now it would be nice to use `servant-client` instead of bilge for even more guarantees.)

tools/stern/src/Stern/Intra.hs
```diff
@@ -114,6 +118,28 @@ import Wire.API.User.Search
 
 -------------------------------------------------------------------------------
 
+backendApiVersion :: Version
+backendApiVersion = V2
+
+-- | Make sure the backend supports `backendApiVersion`.  Crash if it doesn't.  (Call this so
+-- problems are caught by the integration tests.)
```
fisx, line 125:
```suggestion
-- | Make sure the backend supports `backendApiVersion`.  Crash if it doesn't.  (This is called
-- in `Stern.API` so problems make `./services/integration.sh` crash.)
```

PR #3037, opened 2023-01-30
charts: Mark all test resources to be only created while running tests

akshaymankar:
## Checklist

### DIFF
--- a/changelog.d/5-internal/helm-test
+++ b/changelog.d/5-internal/helm-test
@@ -1,0 +1,1 @@
+charts: Mark all test resources to be only created while running tests
--- a/charts/brig/templates/tests/brig-integration.yaml
+++ b/charts/brig/templates/tests/brig-integration.yaml
@@ -2,6 +2,8 @@ apiVersion: v1
 kind: Service
 metadata:
   name: "brig-integration"
+  annotations:
+    "helm.sh/hook": test
   labels:
     app: brig-integration
     chart: {{ .Chart.Name }}-{{ .Chart.Version | replace "+" "_" }}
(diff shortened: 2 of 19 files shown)


jschaul (approved):
As long as CI is happy, this looks good.

I'm not sure configmaps need this annotation, but it probably also doesn't hurt.

akshaymankar:
> I'm not sure configmaps need this annotation, but it probably also doesn't hurt.

I thought why create these resources at all in the non-testing scenarios.

smatting (approved):

akshaymankar:
`helm test --logs` assumes that every test resource is a pod :facepalm:

jschaul:
> `helm test --logs` assumes that every test resource is a pod facepalm [helm/helm#11236](https://github.com/helm/helm/issues/11236)

I'm reworking how helm test works in https://github.com/wireapp/wire-server/pull/3040

then this issue goes away. So review and merge that PR first, then this one might go green?

akshaymankar:
Ah thanks! I was very confused to see the test fail in exactly the same way. I will change the pipeline I guess.

jschaul:
> Ah thanks! I was very confused to see the test fail in exactly the same way. I will change the pipeline I guess.

Or bump the version of helm in use inside cailleach with the same patch, that would also work I suppose.

akshaymankar:
> 

I vaguely remember wanting to get rid of cailleach-image as much as possible for running integration tests and since we don't rely on any sops machinery anymore we shouldn't need it. I have changed the pipeline already. Hopefully things go green now :crossed_fingers:

smatting (approved):

PR #3039, opened 2023-01-30
Streamline team feature ExposeInvitationURLsToTeamAdminConfig implementation

battermann:
Changed the code a little that was introduced by https://github.com/wireapp/wire-server/pull/2684 to streamline the feature flag implementation.

## Checklist

### DIFF
--- a/services/galley/src/Galley/API/Teams/Features.hs
+++ b/services/galley/src/Galley/API/Teams/Features.hs
@@ -675,8 +675,6 @@ instance GetFeatureConfig db ValidateSAMLEmailsConfig where
 instance SetFeatureConfig db ValidateSAMLEmailsConfig where
   setConfigForTeam tid wsnl = persistAndPushEvent @db tid wsnl
 
--- FUTUREWORK: we may also want to get a default from the server config file here, like for
--- sso, and team search visibility.
 instance GetFeatureConfig db DigitalSignaturesConfig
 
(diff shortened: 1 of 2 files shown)


services/galley/src/Galley/API/Teams/Features.hs
```diff
@@ -670,10 +675,9 @@ instance GetFeatureConfig db ValidateSAMLEmailsConfig where
 instance SetFeatureConfig db ValidateSAMLEmailsConfig where
   setConfigForTeam tid wsnl = persistAndPushEvent @db tid wsnl
 
-instance GetFeatureConfig db DigitalSignaturesConfig where
-  -- FUTUREWORK: we may also want to get a default from the server config file here, like for
-  -- sso, and team search visibility.
-  getConfigForServer = pure defFeatureStatus
+-- FUTUREWORK: we may also want to get a default from the server config file here, like for
+-- sso, and team search visibility.
```
fisx, line 679:
```suggestion
```

(it's true that that may be the case, but it seems arbitrary that we point this out, and not a million other similar things.)

fisx (approved):
makes sense, thanks!

services/galley/src/Galley/API/Teams/Features.hs
battermann: Yes, 100%. I just moved the comment, I didn't add this. However, removing this is fine.

PR #3040, opened 2023-01-30
Run helm tests in parallel

jschaul:
## Checklist


jschaul:
Hm, the changes to nix are not being picked up by the PR job it seems; do I need to add the dependency to develop first? Or where is that container built that gets used?

hack/bin/integration-test.sh
```diff
@@ -1,9 +1,79 @@
...
+}
+
+# Run tests in parallel using GNU parallel (see https://www.gnu.org/software/parallel/)
+# The below commands are a little convoluted, but we wish to:
+# - run integration tests. If they fail, keep track of this, but still go and get logs, so we see what failed
+# - run all tests. Perhaps multiple flaky tests in multiple services exist, if so, we wish to see all problems
+touch ~/.parallel/will-cite
+printf '%s\n' "${tests[@]}" | parallel echo "Running helm tests for {}..."
+printf '%s\n' "${tests[@]}" | parallel -P 6 \
+    helm test -n "${NAMESPACE}" "${NAMESPACE}-${CHART}" --timeout 900s --filter name="${NAMESPACE}-${CHART}-{}-integration" '> logs-{};' \
```
akshaymankar, line 35: This will most likely not work with changes in #3037 as the filter will only match the pods and not actually install other components.

jschaul: Let's try: https://github.com/wireapp/wire-server/pull/3042

jschaul:
> The relevant log section seems to be missing the line of the failing test. In this case this line is missing
> 
> ```
> POST /register - can add team members above fanout limit when whitelisting is enabled:
> ```
> 
> looks like it got overwritten by
> 
> ```
> galley-integration: Thread killed by timeout manager
> ```
> 
> https://concourse.ops.zinfra.io/teams/main/pipelines/wire-server-pr/jobs/integration-tests/builds/1461#L639a4d00:4361

Good catch. It happens when there is a line like

```
some test name:                      {"level": "Info", ....
```

where the test name is filtered out as it's on the same line as some noisy log lines. I added some sed magic to deal with these cases in the last commit.

hack/bin/integration-logs-relevant-bits.sh
```diff
@@ -0,0 +1,52 @@
...
+    # add newline between interleaved test name and log output lines
+    sed 's/\x08\+/\n/g' |
+        sed '/^Progress [0-9]\+/d' |
+        sed '/^\s\+$/d' |
+        sed 's/:\s\+{/:\n{/g'
+}
+
+grepper() {
+    # print 10 lines before/after for context
+    rg "$problem_markers" --color=always -A 10 -B 10
```
akshaymankar, line 49: I think this hopes that the red colour is reset during the next 10 lines and no other colour is set at that point, which might not be true and mess up rest of the output. I think we should explicitly print a colour reset code after this `rg` to ensure that doesn't happen.

hack/bin/integration-test.sh
```diff
@@ -1,9 +1,81 @@
...
+            echo "$t-integration passed ✅."
+        fi
+    done
+}
+
+# Run tests in parallel using GNU parallel (see https://www.gnu.org/software/parallel/)
+# The below commands are a little convoluted, but we wish to:
+# - run integration tests. If they fail, keep track of this, but still go and get logs, so we see what failed
+# - run all tests. Perhaps multiple flaky tests in multiple services exist, if so, we wish to see all problems
+mkdir -p ~/.parallel && touch ~/.parallel/will-cite
```
akshaymankar, line 34: Will this directory need to be cleaned up to run the test next time?

hack/bin/integration-logs-relevant-bits.sh
jschaul: this is kind of the end of the ouput for the integration tests. I've not observde any problems with colour so far. But happy to add something. Would a `echo -e "\033[0m"` line after the `rg` line do the trick?

hack/bin/integration-test.sh
jschaul: this is just a marker file to avoid gnu parallel complaining about wanting to be cited in academic papers. No need for cleanup if this runs locally.

(the logs-brig and stat-brig files so far remain on disk if you run the integration test locally)

jschaul: (update: the script will now clean up these local temporary files by default)

akshaymankar:
Maybe I should've pointed it out in https://github.com/wireapp/wire-server/pull/3049, but now we don't need the forked helm, maybe we can also get rid of that?

akshaymankar (approved):

hack/bin/integration-logs-relevant-bits.sh
```diff
@@ -0,0 +1,54 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+USAGE="
+Filter out garbage from logs (but keep colors and highlight problems).
+
+(Adapted from logfilter.sh by Stefan Matting)
```
smatting, line 7: No need to attribute me here ;)

smatting (approved):

migueleliasweb:
> I already thought of using the fork to get another patch in, this one: [helm/helm#11766](https://github.com/helm/helm/pull/11766)
> 
> But perhaps we can also wait until that is merged (if it ever happens).

I really hope it does eventually get merged! :pray:

PR #3043, opened 2023-02-01
Make account registration whitelists local

fisx:
https://wearezeta.atlassian.net/browse/SQPIT-405 (a related wire infrastructure PR is linked in the ticket)

This is changing a feature wire has been using on our staging environment, and (probably?) not anywhere else.  See the changelog if you think you may be affected.

Since the service is both outdated and almost unused, this PR moves the data from that service into the local server config yaml.

Migration should be painless, since the new settings are in a different place than the old ones.  Just make sure the new fields are added to the config before the upgrade, and then you can remove the old ones at any time after. 

## Checklist


jschaul (commented):
While you're at it, could you rename `Whitelist` to `Allowlist` in the types everywhere consistently? I see the config is called `Allowlist`; but the code still uses `Whitelist*`.

fisx:
> While you're at it, could you rename `Whitelist` to `Allowlist` in the types everywhere consistently? I see the config is called `Allowlist`; but the code still uses `Whitelist*`.

oops, thanks!  b1d530c87955779a5ef5559125f99c3bfb0e6bea

services/brig/src/Brig/API/Handler.hs
```diff
@@ -31,13 +31,13 @@ module Brig.API.Handler
...
+import Bilge (RequestId (..))
 import Brig.API.Error
 import qualified Brig.AWS as AWS
 import Brig.App
 import Brig.CanonicalInterpreter (BrigCanonicalEffects, runBrigToIO)
 import Brig.Email (Email)
-import Brig.Options (setWhitelist)
+import Brig.Options (setAllowlistEmailDomains, setAllowlistPhonePrefixes)
 import Brig.Phone (Phone, PhoneException (..))
 import qualified Brig.Whitelist as Whitelist
```
elland, line 42: Maybe worth going all in and changing terms everywhere? :)

fisx: oh man!  :)

cd42e8d944eddcf75dbe49683e4ff67b7a61334c

also 8c1626bdbb4bd4a47ab1e2121c04b0a69bf25301

what else did i miss?

fisx:
@jschaul or @elland can i get another chance?  :)

if you have no objections and feel confident, please merge right away.  our infra is ready (see ticket).

jschaul (approved):
🥳

PR #3047, opened 2023-02-01
websocket ping-pong

jschaul:
* introduces an integration test / regression test to check that control-level pings with a payload result in a control-level pong with the same payload as specified in the [RFC](https://www.rfc-editor.org/rfc/rfc6455#section-5.5.2)

This is about debugging https://wearezeta.atlassian.net/browse/FS-1489

(related ping-pong prior work: https://github.com/wireapp/wire-server/pull/561 and prior discussion: https://github.com/wireapp/wire-server/issues/560)

### DIFF
--- a/services/gundeck/test/integration/API.hs
+++ b/services/gundeck/test/integration/API.hs
@@ -117,6 +117,7 @@ tests s =
       testGroup
         "Websocket pingpong"
         [ test s "pings produce pongs" testPingPong,
+          test s "pings with payload produce pongs with the same payload" testPingPongWithData,
           test s "non-pings are ignored" testNoPingNoPong
         ],
       testGroup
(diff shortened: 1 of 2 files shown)


services/gundeck/test/integration/API.hs
```diff
@@ -1007,6 +1023,22 @@ connectUsersAndDevicesWithSendingClients ca uidsAndConnIds = do
...
+  TestM [(UserId, [(TChan WS.Message, TChan ByteString)])]
+connectUsersAndDevicesWithSendingClientsRaw ca uidsAndConnIds = do
+  chs <- forM uidsAndConnIds $ \(uid, conns) ->
+    (uid,) <$> do
+      forM conns $ \conn -> do
+        chread <- liftIO $ atomically newTChan
+        chwrite <- liftIO $ atomically newTChan
+        _ <- wsRun ca uid conn (wsReaderWriterPing chread chwrite)
+        pure (chread, chwrite)
+  (\(uid, conns) -> wsAssertPresences uid (length conns)) `mapM_` uidsAndConnIds
```
supersven, line 1039: I'm wondering if this wouldn't become a bit simpler, if you would move `wsAssertPresences uid (length conns)` into the scope of the `forM` expression above :thinking:

services/gundeck/test/integration/API.hs
```diff
@@ -1047,10 +1085,13 @@ retryWhileN n f m =
     (const (pure . f))
     (const m)
 
-waitForMessage :: TChan ByteString -> IO (Maybe ByteString)
+waitForMessageRaw :: TChan WS.Message -> IO (Maybe WS.Message)
+waitForMessageRaw = System.Timeout.timeout 1000000 . liftIO . atomically . readTChan
```
supersven, line 1089: `1000000` means *1 Second*, right? Wouldn't this test become more stable if we would allow to wait for e.g. *3 Seconds*?

supersven (approved):
LGTM :+1: 

Only two remarks that aren't mandatory to solve... (AKA nit pickings)

services/gundeck/test/integration/API.hs
jschaul: sure. I've not seen these particular tests flake, but they could of course, and increasing timeout a little should be fine.

services/gundeck/test/integration/API.hs
jschaul: I refactored this a little in https://github.com/wireapp/wire-server/pull/3047/commits/61791b5ef670a129fdc89b5c031685fe1afc2674

PR #3051, opened 2023-02-01
Use openssl instead of tls in federator http2 client

akshaymankar:
## Checklist

### DIFF
--- a/services/federator/src/Federator/Monitor/Internal.hs
+++ b/services/federator/src/Federator/Monitor/Internal.hs
@@ -359,6 +359,22 @@ mkSSLContext ::
   RunSettings ->
   Sem r SSLContext
 mkSSLContext settings = do
+  ctx <- mkSSLContextWithoutCert settings
+
+  Polysemy.fromExceptionVia @SomeException (InvalidClientCertificate . displayException) $
+    SSL.contextSetCertificateFile ctx (clientCertificate settings)
+
+  Polysemy.fromExceptionVia @SomeException (InvalidClientPrivateKey . displayException) $
+    SSL.contextSetPrivateKeyFile ctx (clientPrivateKey settings)
+
+  privateKeyCheck <- Polysemy.fromExceptionVia @SSL.SomeSSLException SSLException $ SSL.contextCheckPrivateKey ctx
+  unless privateKeyCheck $ do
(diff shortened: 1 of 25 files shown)


services/federator/src/Federator/Monitor/Internal.hs
```diff
@@ -323,57 +320,84 @@ watchedDirs resolve path = do
...
   deriving (Show)
 
 instance Exception FederationSetupError
 
 showFederationSetupError :: FederationSetupError -> Text
-showFederationSetupError (InvalidCAStore path) = "invalid CA store: " <> Text.pack path
+showFederationSetupError (InvalidCAStore path msg) = "invalid CA store: " <> Text.pack path <> ", error: " <> Text.pack msg
 showFederationSetupError (InvalidClientCertificate msg) = Text.pack msg
+showFederationSetupError (InvalidClientPrivateKey msg) = Text.pack msg
+showFederationSetupError (CertificateAndPrivateKeyDoNotMatch cert key) = Text.pack $ "Certificate and private key do not match, certificate: " <> cert <> ", private key: " <> key
```
pcapriotti, line 336: Isn't it a bit dangerous to print the private key here?

services/federator/src/Federator/Monitor/Internal.hs
```diff
@@ -323,57 +320,84 @@ watchedDirs resolve path = do
...
+    embed (SSL.contextSetDefaultVerifyPaths ctx)
+
+  pure ctx
+
+-- Context and possible future work see
+-- https://wearezeta.atlassian.net/browse/FS-33
+-- https://wearezeta.atlassian.net/browse/FS-444
+-- https://wearezeta.atlassian.net/browse/FS-443
+--
+-- The current list is compliant to TR-02102-2
```
pcapriotti, line 394:
```suggestion
-- The current list is compliant with TR-02102-2
```

pcapriotti (approved):
Nice!

services/federator/src/Federator/Monitor/Internal.hs
akshaymankar: Its just the path the file.

pcapriotti (approved):

PR #3056, opened 2023-02-02
Lower the log level of federator inotify

pcapriotti:
Lower the log level of federator inotify. Now it doesn't spam the console output anymore when running integration tests.

## Checklist

### DIFF
--- a/changelog.d/5-internal/federator-log
+++ b/changelog.d/5-internal/federator-log
@@ -1,0 +1,1 @@
+Lower the log level of federator inotify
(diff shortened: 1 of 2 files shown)


mdimjasevic (commented):
Was the intention to reduce the number of messages? In other words, what do higher and lower map to when it comes to log levels?

jschaul (approved):

jschaul:
@mdimjasevic the log levels are trace < debug < info < warn < error. Warn is higher than debug. The default for federator is specified in [the helm chart defaults, currently Info](https://github.com/wireapp/wire-server/blob/develop/charts/federator/values.yaml#L26) for CI and deployments; and it's specified as Debug for local integration test runs [here](https://github.com/wireapp/wire-server/blob/develop/services/federator/federator.integration.yaml#L17). Feel free to also set the log level to Info locally if it's too noisy otherwise.

PR #3057, opened 2023-02-02
MLS integration testing: make client group state transparent in state

smatting:
This PR changes the way `MLSTest` monad manages client's group state: The client's group state is now transparently availble in the monad's state instead of in the file system.

### DIFF
--- a/services/galley/test/integration/API/MLS/Util.hs
+++ b/services/galley/test/integration/API/MLS/Util.hs
@@ -47,11 +47,12 @@ import qualified Data.Set as Set
 import qualified Data.Text as T
 import qualified Data.Text.Encoding as T
 import Data.Time
+import qualified Data.UUID as UUID
+import qualified Data.UUID.V4 as UUIDV4
 import Galley.Keys
 import Galley.Options
 import qualified Galley.Options as Opts
 import Imports hiding (getSymbolicLinkTarget)
-import System.Directory (getSymbolicLinkTarget)
 import System.FilePath
 import System.IO.Temp
 import System.Posix hiding (createDirectory)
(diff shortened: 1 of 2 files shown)


services/galley/test/integration/API/MLS/Util.hs
```diff
@@ -316,12 +319,54 @@ takeLastPrekeyNG = do
...
+        pure (argSubst "<group-in>" fn)
+      else pure mempty
+
+  out <-
+    liftIO $
+      spawn
+        ( proc
+            "mls-test-cli"
+            ( ["--store", cdir </> "store"]
+                <> map (appEndo (substIn <> substOut)) args
```
pcapriotti, line 356: It seems to me this is the only place where the monoid instance of `Endo` is used. So why not simply replace with a function and use `.` here?

pcapriotti (approved):
Looks great, thanks for this!
Just for fun: a more haskelly way to do the argument substitution would be to have an `Arg` sum type, as `GroupIn | GroupOut | Arg String`, implement `IsString` for it, then realise the substitution as a function `Arg → String`. This makes the whole thing less stringly typed, and the `IsString` instance should make it equally nice to specify arguments. Probably not worth doing it, though.

Small comment below.

services/galley/test/integration/API/MLS/Util.hs
smatting: I think I got a little bit to excited when I realized that I have an endomorphism, it's really not needed. I'm going to replace it with plain functions and use the IsString strategy as well

smatting:
I gave the less stringly typed versoion a try: There are over 20 instances where `IsString` wouldn't help because we don't have literals but computed values for arguments. I think it's a bit too clunky to wrap all of them with `fromString` (or event worse `fromString . T.unpack`, so I kept the strings.

smatting:
Ah, `Endo` was really unnecessary, because I didn't even need the identity.

smatting:
.. (String -> String) has the wrong kind of instance here: I'd rather the argument unmodified instead of replacing it with an empty string. Fixed directly on the feature branch https://github.com/wireapp/wire-server/commit/9a2f4a0f2b88affccb7d06b1eaecdaa921d7ed11

PR #3062, opened 2023-02-07
Local setup for end2end tests

pcapriotti:
This introduces a replacement for `integration.sh`, written in python, that is able to run two (or more) backends at the same time. It works by creating configuration files on the fly for a set of services, modifying them so that they work together on a new set of ports.

## TODO

 - [x] Set up local SRV records
 - [x] Add spawn timeout
 - [x] Add a second cannon instance
 - [x] Spawn only one backend when `INTEGRATION_FEDERATION_TESTS` is not set
 - [x] `db-reset` should reset the second test keyspace as well
 - [x] Remove legacy integration.sh script
 - [x] The two backends should use different elasticsearch indices

## Checklist


hack/bin/cabal-run-integration.sh
```diff
@@ -46,7 +46,7 @@ run_integration_tests() {
   service_dir="$TOP_LEVEL/services/$package"
 
   cd "$service_dir"
-  "$TOP_LEVEL/services/integration.sh" \
+  "/usr/bin/python3" "$TOP_LEVEL/services/run-services" \
```
stefanwire, line 49: Shouldn't this be the Nix's python?

services/brig/test/integration/Federation/End2end.hs
```diff
@@ -183,14 +183,14 @@ testGetUsersById brig1 brig2 = do
 
 testClaimPrekeySuccess :: Brig -> Brig -> Http ()
 testClaimPrekeySuccess brig1 brig2 = do
-  self <- randomUser brig1
-  user <- randomUser brig2
+  self <- randomUser brig2
+  user <- randomUser brig1
```
stefanwire, line 187: Out of curiosity, would it still work if the two brigs are switched back?

services/federator/src/Federator/Error.hs
```diff
@@ -22,6 +22,7 @@ module Federator.Error
 where
 
 import qualified Data.Aeson as A
+import Debug.Trace
```
stefanwire, line 25: Is this a leftover?

services/start-services-only.sh
```diff
@@ -7,5 +7,5 @@ set -eo pipefail
 
 SERVICES_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 
-# call integration.sh, show a message, then sleep (instead of executing a test executable)
-"$SERVICES_DIR/integration.sh" bash -c 'printf "(This will hang, Control+C to close.)\nNow you can manually curl them or start an integration test executable manually with e.g. \n(first cd to a service dir for correct working directory)\n  cd services/brig && ../../dist/brig-integration -s brig.integration.yaml -i ../integration.yaml\n" && sleep 1000000'
+# call run-services, show a message, then sleep (instead of executing a test executable)
+"/usr/bin/python3" "$SERVICES_DIR/run-services" bash -c 'printf "(This will hang, Control+C to close.)\nNow you can manually curl them or start an integration test executable manually with e.g. \n(first cd to a service dir for correct working directory)\n  cd services/brig && ../../dist/brig-integration -s brig.integration.yaml -i ../integration.yaml\n" && sleep 1000000'
```
stefanwire, line 11: Depending on whether Nix's python should be used or not, this needs to be changed as well.

hack/bin/cabal-run-integration.sh
pcapriotti: Yes, thanks! I forgot we still have to figure out how to make this work in nix.

services/brig/test/integration/Federation/End2end.hs
pcapriotti: It should, this is just a leftover. Thanks for catching.

services/federator/src/Federator/Error.hs
pcapriotti: Yup, thanks.

nix/overlay.nix
```diff
@@ -87,7 +87,7 @@ self: super: {
     inherit (super) stdenv fetchurl;
   };
 
-  helm = super.callPackage ./pkgs/helm {};
+  helm = super.callPackage ./pkgs/helm { };
```
stefanwire, line 90: This is not a semantic change, or is it?

stefanwire (approved):
Just a nit left.

nix/overlay.nix
pcapriotti: It's not, but `treefmt` put it there.

PR #3065, opened 2023-02-08
Break down Galley API tests into subfiles.

elland:
Currently we have a 4kloc file handling all sort of tests, here I've tried to break them down a bit by what at a glance seemed like a good way to arrange them. Feedback more than welcome in ways we can improve and simplify.


services/galley/src/Galley/API/MLS/SubConversation.hs
```diff
@@ -412,7 +412,7 @@ leaveLocalSubConversation cid lcnv sub = do
   kp <-
     note (mlsProtocolError "Client is not a member of the subconversation") $
       cmLookupRef cid (scMembers subConv)
-  removeClientsWithClientMap
+  createAndSendRemoveProposals
```
elland, line 415: Left over from #3063 in order for tests to run and galley to build.

pcapriotti (approved):
LGTM, assuming CI is happy.

elland:
@pcapriotti CI is never happy nowadays. Re-running it yielded the green checks, but it did not update the state in the PR here. I'll merge.

PR #3068, opened 2023-02-09
Tweak docker-ephemeral/run.sh

fisx:
Needed to put docker into a consistent state somehow more reliably after shutdown.  This works as you'd expect for a single `^C`.  If you press ^C twice, you're on your own!

### DIFF
--- a/deploy/dockerephemeral/run.sh
+++ b/deploy/dockerephemeral/run.sh
@@ -1,7 +1,10 @@
 #!/usr/bin/env bash
 
+set -x
+
 # run.sh should work no matter what is the current directory
 SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 DOCKER_FILE="$SCRIPT_DIR/docker-compose.yaml"
 
-docker-compose --file "$DOCKER_FILE" up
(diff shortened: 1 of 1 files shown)


fisx:
please merge if you see & like this!

akshaymankar (approved):

deploy/dockerephemeral/run.sh
```diff
@@ -1,7 +1,10 @@
 #!/usr/bin/env bash
 
+set -x
+
 # run.sh should work no matter what is the current directory
 SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 DOCKER_FILE="$SCRIPT_DIR/docker-compose.yaml"
 
-docker-compose --file "$DOCKER_FILE" up
+docker compose --file "$DOCKER_FILE" up
```
elland, line 9: Can we change this to `docker-compose`?

fisx: why?!  this breaks things for me.  (debian, i'm assuming i'll have to spend some time installing the apt sources from docker now.  but what's the point of changing this?)

elland: 🤔 this broke the script for me on Ubuntu hehe

PR #3070, opened 2023-02-10
Use valid versions in URL path

supersven:
The pattern is `/v<number>/<path>`, not `/<number>/<path>`.

## Checklist

### DIFF
--- a/tools/stern/src/Stern/Intra.hs
+++ b/tools/stern/src/Stern/Intra.hs
@@ -133,10 +133,10 @@ assertBackendApiVersion = recoverAll (constantDelay 1000000 <> limitRetries 5) $
     throwIO . ErrorCall $ "newest supported backend api version must be " <> show backendApiVersion
 
 path :: ByteString -> Request -> Request
-path = Bilge.path . ((toByteString' backendApiVersion <> "/") <>)
+path = Bilge.path . (("v" <> toByteString' backendApiVersion <> "/") <>)
 
 paths :: [ByteString] -> Request -> Request
-paths = Bilge.paths . (toByteString' backendApiVersion :)
+paths = Bilge.paths . ("v" <> toByteString' backendApiVersion :)
 
 -------------------------------------------------------------------------------
 


mdimjasevic (commented):
This would fix it in one place and in a too low-level way if you're asking me.

A slightly better option would be to have a tiny helper like this in a library module (so it can be used by other services too):

```haskell
toPathComponent :: Version -> ByteString
toPathComponent = pack . fmap toLower . show
```

and then use this whenever we want to use a `Version` to construct a path.

pcapriotti:
> This would fix it in one place and in a too low-level way if you're asking me.
> 
> A slightly better option would be to have a tiny helper like this in a library module (so it can be used by other services too):
> 
> ```haskell
> toPathComponent :: Version -> ByteString
> toPathComponent = pack . fmap toLower . show
> ```
> 
> and then use this whenever we want to use a `Version` to construct a path.

Makes sense to have a function do this, but not the way it's implemented. We should never rely on the Show instance. Instead, implement it as `"v" <> toHeader v` or similar.

tools/stern/src/Stern/Intra.hs
```diff
@@ -133,10 +133,10 @@ assertBackendApiVersion = recoverAll (constantDelay 1000000 <> limitRetries 5) $
     throwIO . ErrorCall $ "newest supported backend api version must be " <> show backendApiVersion
 
 path :: ByteString -> Request -> Request
-path = Bilge.path . ((toByteString' backendApiVersion <> "/") <>)
+path = Bilge.path . (("v" <> toByteString' backendApiVersion <> "/") <>)
```
fisx, line 136: i think this must have changed under my nose?  or, more likely, i'm misremembering things.  oh well!

fisx (approved):

fisx:
> Makes sense to have a function do this, but not the way it's implemented. We should never rely on the Show instance. Instead, implement it as `"v" <> toHeader v` or similar.

am i correct in feeling that by now we're stuck with the invariant that versions always have the shape `v` + a non-negative integer?  maybe we should write that down, or even have a test for it?  this may also lead to a sane implementation of `latest` as a path segment one day.  :)

fisx:
another idea: if you had `data VersionNumber = VersionNumber Version`, you could use that in the `VersionInfo` type instead of `Version`.   serialization would strip the `v`, the specs would be happy, and `toByteString' V1` would be `"v1"`.

pcapriotti:
> > 
> 
> Yes, I agree, a unit test would be great to have to assert that the output for V0 is actually "v0". It might seem trivial and like there's no need to test, but the fact that we have this PR and that we're discussing it justifies the need for it in my view.
> 
> @pcapriotti , if you could explain why one should not ever rely on the `Show` instance, I'd appreciate it!

`Show` is meant as a debugging tool. As a general rule, it should return a valid haskell expression that evaluates back to the original value (although we violate this rule quite often). In this case, the `Show` instance is automatically generated, meaning that it relies on the source-level name of a version constant, hence the need for the jiggery-pokery with case. I don't think the internal names used in the source should have any effect. But I guess people who like `Generics` and such things might disagree.

mdimjasevic (approved):

supersven:
I've tested this manually on the `sven-test` environment. It works.

fisx:
> I've tested this manually on the `sven-test` environment. It works.

(screenshot would have been even better :).)

PR #3071, opened 2023-02-10
fix: explicitly disallow negative timestamps in libzauth tokens

comawill:
`timestamp` in `Token` is a signed integer, but is required to be an unsigned integer for `Duration`.

Since `is_expired` will always only be used on tokens with non negative values, we can safely reject all attempts to verify a token with negative `timestamp` values.

## Checklist

### DIFF
--- a/changelog.d/5-internal/libzauth-timestamps
+++ b/changelog.d/5-internal/libzauth-timestamps
@@ -1,0 +1,1 @@
+Explicitly disallow negative timestamps in libzauth tokens when validating if a token is stil valid
--- a/libs/libzauth/libzauth/src/zauth.rs
+++ b/libs/libzauth/libzauth/src/zauth.rs
@@ -187,6 +187,10 @@ impl<'r> Token<'r> {
     }
 
     fn is_expired(&self) -> bool {
+        // negative timestamps are always considered as expired
+        if self.timestamp < 0 {
+            return false;
+        }
         let expiration = UNIX_EPOCH + Duration::from_secs(self.timestamp as u64);
         expiration < SystemTime::now()
     }


OtaK (changes requested):
This doesn’t fix the base issue; 

Timestamps are 64-bit values, unsigned. If you try to fit a timestamp further than 2038 (year 2038 problem, when 32-bit timestamps will overflow) you’ll instead have UB (undefined behavior) because you’re trying to fit an unsigned 64-bit value in a signed 64-bit value (which is 32-bit in both directions). Depending on the platform you’ll either have a crash, soft-wrapping or anything in between. In the case of crash, well, stuff is not working. In the case of soft wrapping, any token that expires after 2038 will loop back on negative timestamps and will thus be automatically rejected by your fix.

A more correct fix would be my initial recommendation and completely disallow negative timestamps (a timestamp cannot be negative anyway) + not even parse timestamps for bots as they might contain -1.

comawill:
@OtaK I might ask stupid questions, but I have problems following your arguments for this specific case:
```rust
 fn is_expired(&self) -> bool {
        // negative timestamps are always considered as expired
        if self.timestamp < 0 {
            return false;
        }
        let expiration = UNIX_EPOCH + Duration::from_secs(self.timestamp as u64);
        expiration < SystemTime::now()
    }
```

My braindump:
* `self.timestamp` is a i64 (signed 64 bit integer)
* we make sure that `self.timestamp > 0` -> MSB is 0 -> only 63 bit representation
* we cast `self.timestamp` to u64 (unsigned 64 bit integer)
* a cast of a unsigned 63bit integer into a unsigned 64bit integer is safe
* `Duration::from_secs` expects a `u64`
* rustlang's `SystemTime` is intended to be Y2038 safe even on 32bit systems (e.g. https://doc.rust-lang.org/src/std/sys/unix/time.rs.html#329)


OT:
* timestamps can be negative to represent times before epoch 0, but don't make sense to represent dates in the future

Therefore PR intends to make it 100% sure, that `is_expired` never accepts negative timestamp values (even though they should never appear in the first place).

All changes don't break current compatibility.

PR #3074, opened 2023-02-10
[FS-1541] After the last leaver the subconversation should be reset

mdimjasevic:
According to https://wearezeta.atlassian.net/wiki/spaces/CORE/pages/641171479/Conferencing+in+MLS+Subconversations#Hangup, a subconversation's epoch should be reset to zero once its last member leaves. The test case in this PR shows the code doesn't currently do that. It additionally asserts there should be no members after the last member leaves, which is a reasonable expectation.

Tracked by https://wearezeta.atlassian.net/browse/FS-1541.

## Checklist


elland (approved):
Easy peasy.

PR #3075, opened 2023-02-10
Introduce VersionNumber newtype.

fisx:
See `/libs/wire-api/test/unit/Test/Wire/API/Routes/Version.hs` for explanation.

## Checklist


.hlint.yaml
```diff
@@ -25,3 +25,27 @@
...
+# Found:
+#   tail . show . fromVersionNumber
+#     <$> allVersionNumbers @=? cs . toByteString'
+#     <$> allVersionNumbers
+# Perhaps:
+#   ((tail . show . fromVersionNumber)
+#      . allVersionNumbers @=? cs . toByteString'
+#      <$> allVersionNumbers)
+# ```
+- ignore: { name: Functor law }
```
elland, line 51: Maybe we don't want to ignore `Functor law` everywhere, but rather where hlint gets tripped up, which seems to be when using the Bilge Assert combinators? 🤔

libs/wire-api/src/Wire/API/Routes/Version.hs
```diff
@@ -57,44 +57,63 @@ import Servant
...
+versionInt V2 = 2
+versionInt V3 = 3
 
 instance ToSchema Version where
-  schema =
-    enum @Integer "Version" . mconcat $
-      [ element 0 V0,
-        element 1 V1,
-        element 2 V2,
-        element 3 V3
```
elland, line 72: IIRC we wanted it explicit because we'll soon start dropping support for older versions in code. But I'm no longer sure.

changelog.d/5-internal/play-with-version-types
```diff
@@ -0,0 +1 @@
+Introduce VersionNumber newtype (see `/libs/wire-api/test/unit/Test/Wire/API/Routes/Version.hs` for explanation)
```
elland, line 1: I expected the mentioned file to include some comment explaining the need and usage of this wrapper.

libs/wire-api/src/Wire/API/Routes/Version.hs
fisx: this is still the case, we're just doing the explicit enumeration in `versionString`, `versionInt` above now.

.hlint.yaml
fisx: I'd prefer to turn off broken laws.  It's not like either side of the law is wrong, it's merely an aesthetic preference.  Anyone case to weigh in either way?

changelog.d/5-internal/play-with-version-types
fisx: fair!  :)

a8d8ed7f1b48230e9efb7fb94fb29ee1d09e87d0 ?

libs/wire-api/src/Wire/API/Routes/Version.hs
```diff
@@ -57,44 +57,63 @@ import Servant
...
 
 -- | Version of the public API.
 data Version = V0 | V1 | V2 | V3
-  deriving stock (Eq, Ord, Bounded, Enum, Show)
+  deriving stock (Eq, Ord, Bounded, Enum, Show, Generic)
   deriving (FromJSON, ToJSON) via (Schema Version)
+  deriving (Arbitrary) via (GenericUniform Version)
+
+versionString :: IsString a => Version -> a
+versionString V0 = "v0"
```
supersven, line 71: I think this could be implemented with [`fromEnum`](https://hoogle.zinfra.io/file/nix/store/wnv3042jbcai41bm9djjbkgfxlwx0nay-ghc-9.2.4-doc/share/doc/ghc/html/libraries/ghc-9.2.4/GHC-Prelude.html#v:fromEnum).

The benefit would be that you won't have to touch this function when a new version is introduced. The drawback is that `Version` constructors have to be sorted.

libs/wire-api/src/Wire/API/Routes/Version.hs
```diff
@@ -57,44 +57,63 @@ import Servant
...
+versionString V1 = "v1"
+versionString V2 = "v2"
+versionString V3 = "v3"
+
+versionInt :: Integral i => Version -> i
```
supersven, line 76: Same here: Could be implemented with `fromEnum`.

libs/wire-api/src/Wire/API/Routes/Version/Wai.hs
```diff
@@ -17,34 +17,45 @@
...
+  where
+    err :: Text -> IO ResponseReceived
+    err v =
+      k . errorRs' . mkError HTTP.status404 "unsupported-version" $
+        "Version " <> cs v <> " is not supported"
 
-parseVersion :: Request -> Maybe (Request, Integer)
+data ParseVersionError = NoVersion | BadVersion Text
+
+parseVersion :: Request -> Either ParseVersionError (Request, Version)
```
supersven, line 51: I'm wondering if this function cannot be generalized regarding its Monad: Shouldn't a constraint `MonadError m => ...` be sufficient?

(This is nitpicking as far as nitpicking can get ... :wink: )

libs/wire-api/test/unit/Test/Wire/API/Routes/Version.hs
```diff
@@ -1,21 +1,44 @@
...
-tests :: T.TestTree
+tests :: TestTree
 tests =
-  T.testGroup "Version" $
-    [ T.testGroup
-        "toPathComponent"
-        [testCase "serialise different versions" testToPathComponent]
+  testGroup
+    "Version always has the shape V<int>; serializations of Version and VersionNumber are `v<int>`, `<int>`, resp.  <int> is non-negative."
+    [ testCase "Version, show, 'v' prefix" $ do
```
supersven, line 17:
```suggestion
    [ testCase "Version, show, 'v' prefix" $ 
```

Is the `do` needed? (Please be careful with applying my suggestion: I haven't tested it. :warning: )

libs/wire-api/src/Wire/API/Routes/Version.hs
```diff
@@ -57,44 +57,68 @@ import Servant
...
+-- serialization rules.
+newtype VersionNumber = VersionNumber {fromVersionNumber :: Version}
+  deriving stock (Eq, Ord, Show, Generic)
+  deriving newtype (Bounded, Enum)
+  deriving (FromJSON, ToJSON) via (Schema VersionNumber)
+  deriving (Arbitrary) via (GenericUniform Version)
+
+instance ToSchema VersionNumber where
+  schema =
+    enum @Integer "Version" . mconcat $ (\v -> element (versionInt v) (VersionNumber v)) <$> [minBound ..]
```
supersven, line 109:
```suggestion
    enum @Integer "VersionNumber" . mconcat $ (\v -> element (versionInt v) (VersionNumber v)) <$> [minBound ..]
```

Have you chosen the description string intentionally or did this happen by copy & paste? (If you had a reason to use `"Version"` here, please ignore this comment.)

.hlint.yaml
elland: Maybe it's just me who sometimes write "dumb" functor code and waits for hlint to clean it for me. 🤔

elland (approved):
LGTM, just unsure about the hlint changes, although not a blocker.

libs/wire-api/src/Wire/API/Routes/Version.hs
fisx: i used to agree, but been there!  :)

- https://github.com/wireapp/wire-server/pull/2978#pullrequestreview-1247230615
- d2e52a1bf0f94c5d5dbbc80d1443100a54e7c1af

libs/wire-api/test/unit/Test/Wire/API/Routes/Version.hs
fisx: it's not there because it needs to be, but because i want it to :)

libs/wire-api/src/Wire/API/Routes/Version/Wai.hs
fisx: yes, that would be sufficient, but i intentionally decided against it.  i like both the abstract interpretation and the precise type that offers exactly what's needed, and not more.

always happy to nit-pick, though! :)

.hlint.yaml
supersven: When I comes to dumb mistakes, I'm a good candidate, too :wink:

libs/wire-api/src/Wire/API/Routes/Version.hs
supersven: This simple version is fine with me, too :smile_cat:

pcapriotti (commented):
I would organise this differently:
- I wouldn't use the `Enum` instance of `Version` to define the schemas, as that is bound to cause confusion. Instead, we can have a list `allVersions :: [Version]` somewhere and use that. In fact, we should probably remove the `Enum` instance.
- `versionString` and `versionInt` can then be defined in terms of the schema instance, as they were (implicitly) before.

fisx:
> I would organise this differently:
> 
>     * I wouldn't use the `Enum` instance of `Version` to define the schemas, as that is bound to cause confusion. Instead, we can have a list `allVersions :: [Version]` somewhere and use that. In fact, we should probably remove the `Enum` instance.
> 
>     * `versionString` and `versionInt` can then be defined in terms of the schema instance, as they were (implicitly) before.

I don't disagree, but I don't understand either: what's the confusion that is caused by the `Enum` instance?

I think `versionString` and `versionInt` are internal and only needed as a place where we can write down explicitly which `Int` belongs to which `Version` in order to support removing old versions.

pcapriotti:
> I don't disagree, but I don't understand either: what's the confusion that is caused by the `Enum` instance?

An `Enum` instance is a bijection with an initial segment of the natural numbers. The instance you get automatically simply assigns indices in order to the various constructors. There is of course another bijection between this type and a set of natural numbers, which happens to be the same right now but will diverge as soon as we remove a version. This is the confusion I'm alluding to: using `fromEnum` or such to convert between versions and integers.

fisx:
> > I don't disagree, but I don't understand either: what's the confusion that is caused by the `Enum` instance?
> 
> An `Enum` instance is a bijection with an initial segment of the natural numbers. The instance you get automatically simply assigns indices in order to the various constructors. There is of course another bijection between this type and a set of natural numbers, which happens to be the same right now but will diverge as soon as we remove a version. This is the confusion I'm alluding to: using `fromEnum` or such to convert between versions and integers.

Ok, fair point, but i would still argue that `Enum` is intended for enumeration, and the bijection is an implementation detail.  In the instance of `Version`, it points to internal numbers that do not correspond to the actual version number except by coincidence.

I don't know how to make this code as straight-forward if we drop the instance, so I'd rather we don't, but if you have an idea please send me code!  (The `ToSchema` instance doesn't yield both version strings and versions, but not that I'm haggling with you, I think maybe we can implement both in terms of a function `versionNumberSchema`?)

fisx:
```
federator>   Federator.Remote
federator>     can get response with valid certificate
federator>       when hostname=localhost and certificate-for=localhost:                                                       FAIL (0.03s)
federator>         Error message: Unexpected remote error: RemoteError (SrvTarget {srvTargetDomain = "localhost", srvTargetPort = 34749}) (FederatorClientHTTP2Exception (BadThingHappen ProtocolError "error:05800074:x509 certificate routines::key values mismatch"))
federator>         CallStack (from HasCallStack):
federator>           assertFailure, called at test/unit/Test/Federator/Remote.hs:76:15 in main:Test.Federator.Remote
federator>         Use -p '/when hostname=localhost and certificate-for=localhost/' to rerun this test only.
```

Happens reproducibly on CI, but I can't reproduce it locally.

fisx:
retried to reproduce locally (only the one test) 15878 times.  got the sibling test to fail locally after 57 re-runs:

```
  Federator.Remote
    can get response with valid certificate
      when hostname=localhost and certificate-for=localhost:                                                       OK (0.10s)
      when hostname=localhost. and certificate-for=localhost:                                                      FAIL (0.03s)
        Error message: Unexpected remote error: RemoteError (SrvTarget {srvTargetDomain = "localhost.", srvTargetPort = 45027}) (FederatorClientHTTP2Exception (BadThingHappen ProtocolError "error:05800074:x509 certificate routines::key values mismatch"))
        
          assertFailure, called at test/unit/Test/Federator/Remote.hs:76:15 in main:Test.Federator.Remote
        Use -p '$0=="Tests.Federator.Remote.can get response with valid certificate.when hostname=localhost. and certificate-for=localhost"' to rerun this test only.
      when hostname=localhost. and certificate-for=localhost.:                                                     OK (0.02s)
```

fisx:
```
Failures:


  test-integration/Test/Spar/Scim/AuthSpec.hs:132:38: 

  1) WireIdPAPIV1.Spar.Scim.Auth, POST /auth-tokens, works with verification code

       uncaught exception: ErrorCall

       Error executing request: responseJsonUnsafeWithMsg: Value Error in $: parsing ASCII failed, expected String, but encountered Null

         error, called at src/Bilge/Response.hs:145:7 in bilge-0.22.0-L77WXGV80XFJMqJLnvaA1h:Bilge.Response

         responseJsonUnsafeWithMsg, called at src/Bilge/Response.hs:134:22 in bilge-0.22.0-L77WXGV80XFJMqJLnvaA1h:Bilge.Response

         responseJsonUnsafe, called at test-integration/Test/Spar/Scim/AuthSpec.hs:163:10 in main:Test.Spar.Scim.AuthSpec
```

PR #3079, opened 2023-02-13
fix(docs): Change year in copyright footer automatically to current year

comawill:
Sphinx operates on environment variable SOURCE_DATE_EPOCH to provide reproducible builds.
This PR intends to change the Makefile to set SOURCE_DATE_EPOCH to the latest git revision

Additionally sphinx will only do this correctly, when there are no spaces between the numbers and dash. [\[src\]](https://github.com/sphinx-doc/sphinx/blob/6ffe7be080943130364bac7633bfc1f4f579ef90/sphinx/config.py#L28)

## Checklist

### DIFF
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -6,10 +6,11 @@ MKFILE_DIR = $(abspath $(dir $(abspath $(lastword $(MAKEFILE_LIST)))))
 
 # You can set these variables (with a ?=) from the command line, and also
 # from the environment.
-SPHINXOPTS    ?= -q
-SPHINXBUILD   ?= sphinx-build
-SOURCEDIR     = $(MKFILE_DIR)/src
-BUILDDIR      = $(MKFILE_DIR)/build
+SPHINXOPTS       ?= -q
+SPHINXBUILD      ?= sphinx-build
+SOURCEDIR         = $(MKFILE_DIR)/src
+BUILDDIR          = $(MKFILE_DIR)/build
+SOURCE_DATE_EPOCH = $(shell git log -1 --format=%ct)
 
 .PHONY: Makefile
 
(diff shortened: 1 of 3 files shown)


smatting (approved):

comawill:
Do not merge: worked perfectly locally, but has no `git` in github action

smatting:
@comawill Try adding `git` here https://github.com/wireapp/wire-server/blob/develop/nix/default.nix?plain=1#L28
I think @flokli mentioned before that adding git to nix envs could causes problems. But in this case maybe its fine?

flokli:
This would mean the docs and all derivations that depend on them would not become binary reproducible anymore. Can we just drop the year numbers from the footer entirely?

smatting:
@comawill I guess "date" inside the build returns unix epoch 0. I think the problem is that within a nix build you don't have access to the outside world's time. nix tries hard to make all builds deterministic on the input, so it forbids external inputs such as the time.
@flokli do you have an idea how to do this? pass the date from outside to script somehow?

flokli:
As i wrote above, ideally we don't add a year number there at all. If we really want to, not make it automatic based on the current time, but manually, once per year.

comawill:
@smatting using `date +%s` actually seems to work, but I agree with @flokli that this compromises reproducible builds.

My attempt was to depend on the git history here, but within `nix-build` we don't have the history available anymore.

Using the current year or a manual set one does not make a difference as the problems are caused by sphinx here when `SOURCE_DATE_EPOCH` is set. My proposed workaround is now to ignore `SOURCE_DATE_EPOCH` within sphinx.

flokli (approved):
We probably want to squash this before merge.

flokli:
Already happened in https://github.com/wireapp/wire-server/pull/3110.

PR #3081, opened 2023-02-13
[FS-1493]: Report federated message sending errors to clients

elland:
Previously federated errors were logged and suppressed. Now we want clients to know where things go wrong, allowing users to deal with these issues. For this PR we add a `failed_to_send` field to MLS messaging responses listing unreachable remote users.


services/galley/test/integration/API/MLS.hs
```diff
@@ -1102,6 +1107,28 @@ testAppMessage2 = do
         WS.assertMatchN_ (5 # WS.Second) wss $
           wsAssertMLSMessage conversation bob (mpMessage message)
 
+testAppMessageUnreachable :: TestM ()
+testAppMessageUnreachable = do
```
mdimjasevic, line 1111: This test case is OK, though I'd also appreciate if you could add a mixed outcome test case, namely one where the message can be sent to some, but not all remote backends.

services/galley/test/integration/API/MLS.hs
```diff
@@ -126,7 +129,9 @@ tests s =
             "Local Sender/Local Conversation"
             [ test s "send application message" testAppMessage,
               test s "send remote application message" testRemoteAppMessage,
-              test s "another participant sends an application message" testAppMessage2
+              test s "another participant sends an application message" testAppMessage2,
+              -- FUTUREWORK(elland): add happy path test as well
```
mdimjasevic, line 133: Why this futurework note?

mdimjasevic (commented):
Overall this looks good, but I'll do another pass after your follow-up.

As mentioned in the private chat, it'd be great to see either the `MessageSendResponse` or `MLSMessageSendingStatus` type renamed (although I'm in favor of renaming the first one) to make it clear that they're for different protocols.

services/galley/test/integration/API/MLS.hs
```diff
@@ -1197,9 +1225,9 @@ testRemoteToLocal = do
             }
 
     WS.bracketR cannon (qUnqualified alice) $ \ws -> do
-      resp <- runFedClient @"send-mls-message" fedGalleyClient bobDomain msr
+      MLSMessageResponseUpdates updates _ <- runFedClient @"send-mls-message" fedGalleyClient bobDomain msr
```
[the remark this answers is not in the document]
elland: I can't because it will always be present, since we do not have reachable remote backends.

services/galley/test/integration/API/MLS.hs
```diff
@@ -1164,12 +1191,14 @@ testRemoteToLocal = do
 
   let bobDomain = Domain "faraway.example.com"
   -- create users
-  [alice, bob] <- createAndConnectUsers [Nothing, Just (domainText bobDomain)]
+  alice <- randomQualifiedUser
+  bob <- randomQualifiedId bobDomain
+
+  connectWithRemoteUser (qUnqualified alice) bob
```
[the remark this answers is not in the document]
elland: Not sure anymore. Reverting it was fine.

cassandra-schema.cql
```diff
@@ -1511,7 +1511,7 @@ CREATE TABLE brig_test.oauth_auth_code (
     redirect_uri blob,
     scope set<text>,
     user uuid
-) WITH bloom_filter_fp_chance = 0.1
+) WITH bloom_filter_fp_chance = 0.01
```
mdimjasevic, line 1514: Do you know where this comes from? Have we changed anything in how we generate this file?

services/galley/src/Galley/API/MLS/Propagate.hs
```diff
@@ -107,15 +110,21 @@ propagateMessage qusr lconv cm con raw = do
             (\(c, _) -> (remoteUserId, c))
             (toList (Map.findWithDefault mempty remoteUserQId cm))
 
+    remotesToQIds = fmap (tUntagged . rmId)
+
     handleError ::
       Member TinyLog r =>
-      Either (Remote [a], FederationError) (Remote RemoteMLSMessageResponse) ->
-      Sem r ()
+      -- [RemoteMember] ->
```
mdimjasevic, line 117: Would be good to remove the dead code.

PR #3084, opened 2023-02-15
Bump nixpkgs

akshaymankar:
## Checklist


fisx (approved):
I would really prefer an internal changelog entry saying 'bump nix/sources.json'.  It's a pretty substantial change in terms of code; the benefit would be that this information wouldn't be lossed to anybody debugging issues with it; and I don't see the harm.

Also, if you sort large text passages by line in a separate commit that doesn't have any actual changes, that makes it easier to review.

Otherwise LGTM!

PR #3086, opened 2023-02-15
[SQSERVICES-1913] Move OAuth authentication to nginz

battermann:


libs/libzauth/libzauth-c/Makefile
```diff
@@ -27,7 +27,7 @@ build:
 	cargo build
 
 build-release:
-	cargo build --release
+	cargo build
```
beltram, line 30: revert to release mode

libs/libzauth/libzauth-c/src/lib.rs
```diff
@@ -284,8 +312,191 @@ impl From<TokenVerification> for ZauthTokenVerification {
...
+    NullArg = 2,
+    IoError = 3,
+    Utf8Error = 4,
+    Panic = 99,
+}
+
+pub struct OAuthJwk(String);
+
+#[repr(C)]
+#[derive(Clone, Copy, Debug)]
```
beltram, line 338: You probably don't want to derive Copy here. It's basically an "implicit" clone. And since your struct holds pointer, it could silently copy those pointers and somehow result in double free errors

libs/libzauth/libzauth-c/src/lib.rs
```diff
@@ -284,8 +312,191 @@ impl From<TokenVerification> for ZauthTokenVerification {
...
+            OauthError::JwtSimpleError(_) => Self::Panic,
+            OauthError::Base64DecodeError(_) => Self::Panic,
+            OauthError::InvalidJwk => Self::Panic,
+            OauthError::InvalidJwtNoSubject => Self::Panic,
+            OauthError::InvalidScope => Self::InsufficientScope,
+        }
+    }
+}
+
+impl From<io::Error> for OAuthResultStatus {
```
beltram, line 367: You probably could have handled all these manual `From` implementations with the `thiserror` crate.
But it could be tricky to handle with the discriminant so leave it alone if it's too complex

```rust
#[repr(C)]
#[derive(Error, Clone, Copy, Debug)]
pub enum OAuthResultStatus {
    Ok,
    IoError(#[from] io::Error),
}
```

libs/libzauth/libzauth-c/src/lib.rs
```diff
@@ -284,8 +312,191 @@ impl From<TokenVerification> for ZauthTokenVerification {
...
+    fn from(_: io::Error) -> OAuthResultStatus {
+        OAuthResultStatus::IoError
+    }
+}
+
+impl From<str::Utf8Error> for OAuthResultStatus {
+    fn from(_: str::Utf8Error) -> OAuthResultStatus {
+        OAuthResultStatus::Utf8Error
+    }
+}
```
beltram, line 377:
```suggestion
impl From<str::Utf8Error> for OAuthResultStatus {
    fn from(_: str::Utf8Error) -> Self {
        Self::Utf8Error
    }
}
```

libs/libzauth/libzauth-c/src/lib.rs
```diff
@@ -284,8 +312,191 @@ impl From<TokenVerification> for ZauthTokenVerification {
...
+    method_len: size_t,
+) -> OAuthResult {
+    match panic::catch_unwind(|| {
+        if token.is_null() {
+            return OAuthResult {
+                uid: ptr::null(),
+                status: OAuthResultStatus::NullArg,
+            };
+        }
+        if scope.is_null() {
```
beltram, line 461: You don't check 'method' ?

beltram (approved):

libs/libzauth/libzauth-c/src/lib.rs
battermann: Missed that, thanks!

libs/libzauth/libzauth-c/src/lib.rs
```diff
@@ -284,8 +312,197 @@ impl From<TokenVerification> for ZauthTokenVerification {
...
+        let scope = try_unwrap!(str::from_utf8(bytes));
+        let bytes = unsafe { slice::from_raw_parts(method, method_len) };
+        let method = str::from_utf8(bytes).unwrap();
+        let subject = try_unwrap!(verify_oauth_token(&jwk.0, token, scope, method));
+        let c_str = try_unwrap!(CString::new(subject));
+        OAuthResult {
+            uid: c_str.into_raw(),
+            status: OAuthResultStatus::Ok,
+        }
+    }) {
```
pcapriotti, line 485: Can't you use `catch_unwind_with` here too?

services/nginz/third_party/nginx-zauth-module/zauth_module.c
```diff
@@ -276,65 +315,39 @@ static ngx_int_t zauth_and_oauth_handle_request (ngx_http_request_t * r) {
...
+                        ngx_pool_cleanup_t * finaliser = ngx_pool_cleanup_add(r->pool, 0);
+                        if (finaliser == NULL) {
+                                return NGX_ERROR;
+                        }
+                        finaliser->handler = delete_oauth_uid;
+                        finaliser->data = res.uid;
+                        ngx_http_set_ctx(r, res.uid, zauth_module);
+                        return NGX_OK;
+                } else if (res.status == OAUTH_INSUFFICIENT_SCOPE) {
+                        ngx_log_error(NGX_LOG_ERR, r->connection->log, 0, "OAuth insufficient cope");
```
pcapriotti, line 339:
```suggestion
                        ngx_log_error(NGX_LOG_ERR, r->connection->log, 0, "OAuth insufficient scope");
```

services/nginz/third_party/nginx-zauth-module/zauth_module.c
```diff
@@ -597,6 +600,22 @@ static ngx_int_t zauth_token_var (ngx_http_request_t * r, ngx_http_variable_valu
         return zauth_set_var(r->pool, v, zauth_token_lookup(t, data));
 }
 
+static ngx_int_t zauth_token_var_user (ngx_http_request_t * r, ngx_http_variable_value_t * v, uintptr_t _) {
+        ZauthToken const * t = ngx_http_get_module_ctx(r, zauth_module);
+        if (t != NULL && zauth_is_authorized_and_allowed(r)) {
```
pcapriotti, line 605: This doesn't look right. First `t` is cast to `ZauthToken` and assumed to be a valid pointer to that struct by `zauth_is_authorized_and_allowed`. Later, it is fetched again and cast to a string pointer.

You might want to store a tagged union in the context, so that this function can discriminate if it's the zauth or oauth case.

pcapriotti (commented):
Looks generally good, as far as I can tell. I did find a potential problem though. See below.

libs/libzauth/libzauth-c/src/lib.rs
battermann: Yes!

services/nginz/third_party/nginx-zauth-module/zauth_module.c
```diff
@@ -369,8 +376,79 @@ static ngx_int_t zauth_handle_zauth_request (ngx_http_request_t * r, const Zauth
...
+        if (r->headers_in.authorization == NULL) {
+                return NGX_HTTP_UNAUTHORIZED;
+        }
+
+        ngx_str_t hdr = r->headers_in.authorization->value;
+
+        if (strncmp((char const *) hdr.data, "Bearer ", 7) == 0) {
+                OAuthResult res = oauth_verify_token(key, &hdr.data[7], hdr.len - 7, scope.data, scope.len, r->method_name.data, r->method_name.len);
+                if (res.status == OAUTH_OK) {
+                        ZauthContext * ctx = alloc_oauth_context(r, res.uid);
```
pcapriotti, line 389: For OOM-safety, you can add
```.c
if (ctx == NULL) return NGX_ERROR;
```
or similar.

services/nginz/third_party/nginx-zauth-module/zauth_module.c
```diff
@@ -400,32 +478,18 @@ static ngx_int_t zauth_parse_request (ngx_http_request_t * r) {
         }
 
         if (res == ZAUTH_OK && tkn != NULL) {
-                ngx_pool_cleanup_t * finaliser = ngx_pool_cleanup_add(r->pool, 0);
-                if (finaliser == NULL) {
-                        return NGX_ERROR;
+                ZauthContext * ctx = alloc_zauth_context(r, tkn);
+                ngx_int_t e = setup_zauth_context(r, ctx);
```
pcapriotti, line 482: Here too.

pcapriotti (approved):
I looked at the C code only this time. Looks good to me modulo the comments below.

PR #3089, opened 2023-02-15
Coturn container image update errata

sysvinit:
This is a follow-up to #3078. The pre-stop hook script was not updated when some metrics were relabeled in the update from 4.5.2 to 4.6.0, which would mean that the script would inadvertently exit while there were still active TURN sessions in the coturn process. This change updates the coturn image to a version with a fixed pre-stop hook script.

## Checklist

### DIFF
--- a/changelog.d/2-features/coturn-image-update
+++ b/changelog.d/2-features/coturn-image-update
@@ -1,5 +1,5 @@
 The coturn container image included in the coturn Helm chart was updated to
-version `4.6.0-wireapp.3`.
+version `4.6.0-wireapp.4`.
 
 With this version of coturn, the Prometheus metrics endpoint has been
 updated, and the `turn_active_allocations` metric label has been *renamed* to
--- a/charts/coturn/Chart.yaml
+++ b/charts/coturn/Chart.yaml
@@ -11,4 +11,4 @@ version: 0.0.42
 # This is the version number of the application being deployed. This version number should be
 # incremented each time you make changes to the application. Versions are not expected to
 # follow Semantic Versioning. They should reflect the version the application is using.
-appVersion: 4.6.0-wireapp.3
+appVersion: 4.6.0-wireapp.4


jschaul (approved):
*shrug*. Fine for me.

The other containers we build make use of [dumb-init](https://github.com/Yelp/dumb-init) , see e.g. https://github.com/wireapp/wire-server/blob/6a6bcd71a00b5fac25597fdd8cf54fd92766df41/nix/wire-server.nix#L247 to do sane signal handling inside a container.

These version numbers like `4.6.0-wireapp.4` - how do I know which code they correspond to? There's no git tag or similar I can see inside wireapp/coturn, so I have no idea how this image was built and what code it relates to. How do you look this up? Could we tag git commits with this version number so people could know what this number refers to?

sysvinit:
> The other containers we build make use of [dumb-init](https://github.com/Yelp/dumb-init)

Our coturn container image is based on the upstream container image, which itself is based on Debian and runs coturn as PID 1. See e.g. [this Dockerfile for the base image](https://github.com/wireapp/coturn/blob/wireapp/4.6.0/docker/coturn/debian/Dockerfile), [this Dockerfile which adds the graceful shutdown hook](https://github.com/wireapp/coturn/blob/wireapp/4.6.0/docker/coturn/wireapp/Dockerfile), and the `coturn` pipeline in Concourse.

> These version numbers like 4.6.0-wireapp.4 - how do I know which code they correspond to? There's no git tag or similar I can see inside wireapp/coturn, so I have no idea how this image was built and what code it relates to. How do you look this up? Could we tag git commits with this version number so people could know what this number refers to?

The tags in wireapp/coturn are prefixed with the string `wireapp/`, to distinguish them from upstream tags. So, the code for the image 4.6.0-wireapp.4 can be found at the tag `wireapp/4.6.0-wireapp.4`. These version numbers are incremented automatically in the Concourse pipeline too.

PR #3092, opened 2023-02-15
Deflake invitation too many pending

jschaul:
deflake `post /teams/:tid/invitations - 403 too many pending'`

```   
    FAIL
            Exception: Assertions failed:
             1: 403 =/= 201
             2: Just too-many-team-invitations =/= Nothing
```

### DIFF
--- a/services/brig/test/integration/API/Team.hs
+++ b/services/brig/test/integration/API/Team.hs
@@ -96,7 +96,7 @@ tests conf m n b c g aws = do
             test m "post /teams/:tid/invitations - email lookup nginz" $ testInvitationEmailLookupNginz b n,
             test m "post /teams/:tid/invitations - email lookup register" $ testInvitationEmailLookupRegister b,
             test m "post /teams/:tid/invitations - 403 no permission" $ testInvitationNoPermission b,
-            test m "post /teams/:tid/invitations - 403 too many pending" $ testInvitationTooManyPending b tl,
+            test m "post /teams/:tid/invitations - 403 too many pending" $ testInvitationTooManyPending conf b tl,
             test m "post /teams/:tid/invitations - roles" $ testInvitationRoles b g,
             test m "post /register - 201 accepted" $ testInvitationEmailAccepted b g,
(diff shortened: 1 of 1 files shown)


services/brig/test/integration/API/Team.hs
```diff
@@ -354,14 +354,17 @@ headInvitationByEmail service email expectedCode =
...
-  pooledForConcurrentlyN_ 16 emails $ postInvitation brig tid inviter . stdInvitationRequest
   email <- randomEmail
-  -- TODO: If this test takes longer to run than `team-invitation-timeout`, then some of the
-  --       invitations have likely expired already and this test will actually _fail_
+  -- If this test takes longer to run than `team-invitation-timeout`, then some of the
+  -- invitations have likely expired already and this test will actually _fail_
+  -- therefore we increase the timeout from default 10 to 300 seconds
+  let longerTimeout = opts {Opt.optSettings = (Opt.optSettings opts) {Opt.setTeamInvitationTimeout = 300}}
+  withSettingsOverrides longerTimeout $ do
+    forM_ emails $ postInvitation brig tid inviter . stdInvitationRequest
```
mdimjasevic, line 367: I don't know what's the `setMaxTeamSize` value in options passed to the test, but if it's big, could it be that doing this sequentially takes a bit too long? I was planning to make a very little team size limit, along these lines:

```haskell
  let smallTeam =
        opts
          { Opt.optSettings =
              (Opt.optSettings opts)
                { Opt.setTeamInvitationTimeout = 300,
                  Opt.setMaxTeamSize = 5
                }
          }
  withSettingsOverrides smallTeam $ do
```

mdimjasevic (approved):

services/brig/test/integration/API/Team.hs
jschaul: I was thinking since tests already run in parallel, we may not gain that much by spawning even more threads, as these are just simple http requests that should be at milisecond response time level in theory (when there isn't anything else ongoing and enough CPU/memory available). Right now the team size is 32 that is inside the hack/helm_vars/... override. Setting maxTeamSize to something smaller is a good idea though, your suggestion makes sense, feel free to adjust this test to this in a separate PR (I see you already merged this one as-is).

PR #3097, opened 2023-02-21
FS-51 Report unavailable clients for Proteus messages

lepsa:
Changing the return types to match the ticket.
Adding tests and fixing some logic errors.

Sending a Proteus message will no longer emit an error when the conversation owning server cannot contact federated servers to retrieve client lists. These unavailable users will be listed in the failed to send field of the response sent back to the client. Federation members that can be contacted will have messages fanned out to them.

## Checklist


services/galley/test/integration/API.hs
```diff
@@ -1148,6 +1149,86 @@ postMessageQualifiedLocalOwningBackendFailedToSendClients = do
...
+
+    let expectedFailedToSend =
+          QualifiedUserClients . Map.fromList $
+            [ ( remoteDomain,
+                Map.fromList
+                  [ (deeId, Set.singleton deeClient)
+                  ]
+              )
+            ]
+        expectedRedundant = expectedFailedToSend
```
mdimjasevic, line 1220: Why is `deeClient` a redundant recipient?

services/galley/src/Galley/API/Message.hs
```diff
@@ -463,7 +474,44 @@ postQualifiedOtrMessage senderType sender mconn lcnv msg =
...
+    toDomUserClient :: Map Domain (Map UserId (Set ClientId)) -> [(Domain, (UserId, Set ClientId))]
+    toDomUserClient m = do
+      (d, m') <- Map.assocs m
+      (d,) <$> Map.assocs m'
+    -- Rebuild the map, concatenating results along the way.
+    fromDomUserClient :: [(Domain, (UserId, Set ClientId))] -> Map Domain (Map UserId (Set ClientId))
+    fromDomUserClient = foldr f mempty
+      where
+        f :: (Domain, (UserId, Set ClientId)) -> Map Domain (Map UserId (Set ClientId)) -> Map Domain (Map UserId (Set ClientId))
+        f (d, (u, c)) m = Map.alter (pure . Map.alter (g c . fromMaybe mempty) u . fromMaybe mempty) d m
```
mdimjasevic, line 513: This doesn't seem so straightforward to understand, but perhaps it was the easiest way to write it.

services/galley/src/Galley/API/Message.hs
```diff
@@ -463,7 +474,44 @@ postQualifiedOtrMessage senderType sender mconn lcnv msg =
...
+          { mssFailedToSend =
+              QualifiedUserClients $
+                foldr
+                  (Map.unionWith (Map.unionWith Set.union))
+                  mempty
+                  [ qualifiedUserClients failedToSend,
+                    qualifiedUserClients failedToSendFetchingClients,
+                    fromDomUserClient redundantFailed
+                  ]
+          }
```
mdimjasevic, line 500: Can you please write a few tests that check various cases that can happen here? I don't think the current test provides a good coverage wrt this part of the code.

mdimjasevic (commented):
Overall, this looks good, but there are some parts that are not that easy for me to grasp. It'd be great if we could go over the PR in a call!

mdimjasevic:
@lepsa , please take a look at the federation chat and the reasoning why this is a breaking API change. I'd be happy to pair with you tomorrow morning (Wednesday) should you need help with splitting an endpoint in `wire-api` into two, one for versions up to and including v2, and a new one for all versions from v3 and up. I also tried to incorporate your structure suggestion in Confluence: https://wearezeta.atlassian.net/wiki/spaces/CORE/pages/738557958/Offline+backend+handling+-+analysis+of+prekey+fetching+Proteus#Suggested-change

services/galley/test/integration/API.hs
lepsa: It is due to other tests expecting redundant clients under similar circumstances. If I add code to filter out users from the redundant field those tests will fail.

services/galley/src/Galley/API/Message.hs
lepsa: It basically is the easiest way. It moves into each of the Map values by the given keys and appends the client sets. It needs to be `alter` however so that new keys can be added, which requires the update function to be `Maybe a -> Maybe a`, hence all the `pure`s and `fromMaybe`s.

I'll update the names though.

services/galley/test/integration/API.hs
mdimjasevic: Ok, but I still don't understand why. It would also be beneficial for your understanding to get to the bottom of it.

services/galley/test/integration/API.hs
```diff
@@ -1229,6 +1248,53 @@ postMessageQualifiedLocalOwningBackendFailedToSendClients2 = do
       WS.assertMatch_ t wsBob (wsAssertOtr' encodedData convId aliceOwningDomain aliceClient bobClient encodedTextForBob)
       WS.assertMatch_ t wsChad (wsAssertOtr' encodedData convId aliceOwningDomain aliceClient chadClient encodedTextForChad)
 
+testBuildFailedToSend :: TestM ()
+testBuildFailedToSend = liftIO $ do
```
mdimjasevic, line 1252: The tests in here are unit tests, aren't they? Can you move them to a unit test module?

mdimjasevic (approved):
Apart from moving unit tests into a unit test module, this looks good!

mdimjasevic:
Please move the unit tests and then merge. No need to wait for another approval from me.

PR #3101, opened 2023-02-23
[SQSERVICES-1922] [OAUTH] Documentation

battermann:


docs/src/developer/reference/oauth.md
```diff
@@ -0,0 +1,402 @@
...
+```
+
+These credentials have to be stored in a safe place and cannot be recovered if they are lost.
+
+### Get authorization code
+
+When the user wants to use the 3rd party app for the first time, they need to authorize it to access Wire resources on their behalf.
+
+To do so the user is redirected to a login screen for authentication. Once authenticated, the user can either grant or deny the client's access request and the corresponding scope (a list of permissions to give to the 3rd party app).
+
```
flokli, line 90: A lot of what comes below is mostly taken from [RFC6749](https://www.ietf.org/rfc/rfc6749.txt).

I'd feel much more comfortable if we'd simply refer to that, instead of duplicating it here, so people don't need to put both side by side to check if there's any differences.

docs/src/developer/reference/oauth.md
```diff
@@ -0,0 +1,402 @@
...
+Location: https://example.com?code=1395a1a44b72e0b81ec8fe6c791d2d3f22bc1c4df96857a88c3e2914bb687b7b&state=foobar
+Vary: Accept-Encoding
+```
+
+### Retrieve access and refresh token
+
+The 3rd party app sends the authorization code together with the client credentials and the parameters shown below using the `application/x-www-form-urlencoded` format with character encoding of UTF-8 to the authorization server to retrieve an access token and a refresh token:
+
+```curl
+curl -s -X POST localhost:8080/oauth/token \
```
flokli, line 130: I had to look twice to understand this is the "external" endpoint of wire-server (the internal uses port 8082)

The RFC uses server.example.com and client.example.com. Maybe we can use the same consistently too? (and use server-internal.example.com for the internal endpoint)

docs/src/developer/reference/oauth.md
```diff
@@ -0,0 +1,402 @@
...
+       R->>R: Validate access token with public key from auth server
+       R->>C: Response
+```
+
+### Registering an OAuth client
+
+A new OAuth client can be register via the internal API of `brig` by providing an application name and a redirect URL:
+
+```curl
+  curl -s -X POST localhost:8082/i/oauth/clients \
```
flokli, line 59: server-internal.example.com

docs/src/developer/reference/oauth.md
```diff
@@ -0,0 +1,402 @@
...
+Example response:
+
+```http
+HTTP/1.1 200 OK
+(empty-response-body)
+```
+
+### Retrieve a list of 3rd party apps with account access
+
+Authenticated endpoint to retrieve a list of all applications that have account access via OAuth.
```
flokli, line 221: Is this part of another RFC, or a wire-specific endpoint?

Is it on the internal or external port, same as registering new applications?

docs/src/developer/reference/oauth.md
```diff
@@ -0,0 +1,402 @@
...
+
+A secure JWK key has been generated
+
+JWK key including the private key has been set in brig secrets for prod and staging
+
+Public JWK key has been set in galley options
+
+OAuth end-to-end test works with new JWK key
+
+### Default configs
```
flokli, line 402: I assume this is still a WIP list of notes?

docs/src/developer/reference/oauth.md
```diff
@@ -0,0 +1,402 @@
...
+| `client_secret` | Required. The client's secret.                                                          |
+
+Example response:
+
+```JSON
+{
+  "accessToken": "eyJhbGciOiJFZERTQSJ9.eyJhdWQiOiJleGFtcGxlLmNvbSIsImV4cCI6MS42NzcyMzYyNDkxMTIzMTk3MWU5LCJpYXQiOjEuNjc3MjM2MjQ2MTEyMzE5NzFlOSwiaXNzIjoiZXhhbXBsZS5jb20iLCJzY29wZSI6InJlYWQ6c2VsZiIsInN1YiI6ImJhOTIxY2ZmLWU1ZWEtNDMxNS1iZTNkLWZiNjA3NTU0M2Y3MCJ9.5ksjS7msi9NSNat-qh7-Y5O-u9TcuYeLWTAsiAyes_oLwfjD_jYtGevUAuiVV6RXgPBO00VEMv-ZS86e7sd5Dg",
+  "expiresIn": 300,
+  "refreshToken": "eyJhbGciOiJFZERTQSJ9.eyJzdWIiOiI4NzI1ZTRkNC01Njc5LTQwZGEtOTI3My03YTBkMmIwYjUwMGYifQ.59IICzGoli5nfwJ1ZwRH_b3T-lRgBrralE1EZZRtadI2eKrta0kaLIZpuMWPC2Icj6-LSEBsyYLXpxOm3cNaDw",
+  "tokenType": "Bearer"
```
flokli, line 152: I'm not sure I like having an example here. OAUTH itself says these tokens are opaque, but this seems to be a JWT - even though there's no mention of the endpoints specified by [RFC8414](https://www.rfc-editor.org/rfc/rfc8414.html).

Also, are you sure about the key names here? Section 4.1.4 in [RFC 6749](https://www.ietf.org/rfc/rfc6749.txt) uses snake case.

docs/src/developer/reference/oauth.md
flokli: Do we really want to use the internal endpoint here? It's not reachable from the outside, and I assume someone wanting to add the calendar integration can't forward the internal endpoint to initiate the flow…

hack/bin/oauth_test.sh
```diff
@@ -64,13 +64,17 @@ AUTH_CODE=$(
...
-  curl -s -X POST localhost:8082/oauth/token \
+ACCESS_TOKEN_RESPONSE=$(
+  curl -s -X POST localhost:8080/oauth/token \
     -H "Content-Type: application/x-www-form-urlencoded" \
-    -d 'code='"$AUTH_CODE"'&client_id='"$CLIENT_ID"'&grant_type=authorization_code&redirect_uri=https://example.com&client_secret='"$CLIENT_SECRET" |
-    jq -r '.accessToken'
+    -d 'code='"$AUTH_CODE"'&client_id='"$CLIENT_ID"'&grant_type=authorization_code&redirect_uri=https://example.com&client_secret='"$CLIENT_SECRET"
 )
 
+echo "$ACCESS_TOKEN_RESPONSE" | jq
```
fisx, line 73:
```suggestion
echo "$ACCESS_TOKEN_RESPONSE" | jq >&2
```

ya, i'm nit-picking again :)

services/nginz/integration-test/conf/nginz/nginx.conf
```diff
@@ -291,7 +291,17 @@ http {
...
+
+    location /oauth/revoke {
+      include common_response_no_zauth.conf;
+      proxy_pass http://brig;
+    }  
+
+    location /oauth/applications {
+      include common_response_with_zauth.conf;
+      proxy_pass http://brig;
+    }
```
fisx, line 304: shouldn't this be in https://github.com/wireapp/wire-server/pull/2989?  why not?  (same for [charts/nginz/values.yaml](https://github.com/wireapp/wire-server/pull/3101/files#diff-40768c765dccb009ab75954e5cdcaf502510b04989fe3e4220eb77c748aa7b36))

docs/src/developer/reference/oauth.md
```diff
@@ -0,0 +1,402 @@
+# OAuth
+
+OAuth 2.0 is used to authorize 3rd party applications to access `wire-server` resources on behalf of a Wire user.
+
+Currently, only 3rd party apps that have been implemented and approved by Wire are supported, e.g. Outlook Calendar Extension. OAuth is not open for public use.
```
fisx, line 5: where can i get an exhaustive list of supported apps?  (brig settings, correct?  probably something for later down in the doc, but still.)

services/nginz/integration-test/conf/nginz/nginx.conf
battermann: yes, it is

docs/src/developer/reference/oauth.md
battermann: This info is not in brig settings as it is dynamic.
Currently, there is only Outlook calendar integration. I'll list it here.

fisx: what does "dynamic" mean?  that we only know this as opaque strings?

fisx:
please don't forget to make clear what needs to be done by whom: wire-server devs, site operator, team admin.  (i'm assuming wire-server devs have done their job already, just making sure :).

(i'm writing this because i just saw that some thing can only be done via internal end-points, so they can't be done by team admins.  this is fine, but should be explicit.)

docs/src/developer/reference/oauth.md
battermann: Dynamic means that an OAuth client can be created via the internal API and not via a static config file.

fisx (approved):
feel free to continue this in the parent PR.

docs/src/developer/reference/oauth.md
flokli: Do you still plan to address this?

battermann: Partly, yes. I discussed with @fisx that it makes sense to have a concise reference for OAuth client developers, so that they do not have to skim through the noise of the complete RFC.
I am going to reference the RFC and additionally provide a documentation of the subset that we implemented.
This is also quite helpful for anyone new to the topic to get an overview and basic understand of what this is about.

docs/src/developer/reference/oauth.md
battermann: yes

docs/src/developer/reference/oauth.md
battermann: I changed it to snake case.

PR #3103, opened 2023-02-24
Fix brig config for local testing

battermann:
Fixes an issue with config parser introduced in #3043


pcapriotti (approved):
Thanks, I was just about to submit the same fix :)

PR #3106, opened 2023-02-27
Add sanitize-pr make rule.

fisx:
quick & dirty internal make rule to make sure concourse won't have any avoidable issues with a PR.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -91,6 +91,15 @@ endif
 ci: c db-migrate
 	./hack/bin/cabal-run-integration.sh $(package)
 
+.PHONY: sanitize-pr
+sanitize-pr:
+	./hack/bin/generate-local-nix-packages.sh
+	make formatf-all
+	make hlint-inplace-all
+	make git-add-cassandra-schema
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -91,6 +91,15 @@ endif
...
 	./hack/bin/cabal-run-integration.sh $(package)
 
+.PHONY: sanitize-pr
+sanitize-pr:
+	./hack/bin/generate-local-nix-packages.sh
+	make formatf-all
+	make hlint-inplace-all
+	make git-add-cassandra-schema
+	@git diff-files --quiet -- || ( echo "There are unstaged changes, please take a look, consider committing them, and try again."; exit 1 )
+	@git diff-index --quiet --cached HEAD -- || ( echo "There are staged changes, please take a look, consider committing them, and try again."; exit 1 )
```
mdimjasevic, line 101: What are these @-signs at the beginning for?

mdimjasevic (approved):

Makefile
fisx: they ask make to not print it before running it.  i don't insist to keep them, but it seemed tidier to me.

PR #3109, opened 2023-02-28
Python integration tests

pcapriotti:
This PR introduces the beginning of a new integration test suite, written in python and based on pytest.

To try this out, first spawn services:
```bash
make c && INTEGRATION_FEDERATION_TESTS=1 ./services/run-services
```
then (in a separate terminal):
```bash
cd integration
pytest
```

At the moment, the test suite contains some API wrappers, including a system for asserting on websocket events, and a few setup utilities, like functions for creating and connecting users, and a port of the MLS testing setup from Haskell.

A refactoring to bring all integration tests together into a single package is long overdue. Having integration tests for services reside in specific service packages has caused enormous amounts of code duplication, and this would only get worse as we go in the direction of making federation tests more end-to-end.

The choice of going with python might be controversial, but I believe there are several strong advantages to using a "looser" language like python, and a system based on convention over configuration, for example:

- writing tests becomes a much faster and seemless experience, thanks to not having to recompile test code over and over,
- stack traces work automatically with no boilerplate,
- not having to "unwrap" Maybes and Eithers, it is much easier to write code that simply crashes with a good error message in case of unexpected values,
- very easy to debug interactively (`import ipdb; ipdb.set_trace()`),
- setup code is more straighforward and obvious, no need to create state monads or anything like that,
- fewer imports, less boilerplate: you can literally create a new empty test module and add a test without any import statements.

But most importantly, python tests obviously don't have a dependency on the types in wire-api. The fact that the integration tests depend on those types is a pretty serious flaw with the current system, as it hides many possible bugs and makes a lot of the checks happening in the tests completely tautological.

It would be possible, of course, to access the API types manually from haskell as well, but I think the inconvenience of doing that, even with things like lens-aeson, would be hard to justify, when compared to the alternative of simply using a language that is basically designed for such things.

There are a few regressions compared to the current system, however. Those can be addressed, and I'm planning to do so if this proposal is accepted, but I thought it would make sense to have people look at what this looks like and give their opinions, before sinking even more time into it.

The missing bits are:
 - a replacement for `withSettingsOverrides`
 - running tests in parallel
 - running tests in CI
 - mock federator, although this would mostly be rendered obsolete by making the tests end-to-end.

Overall, I'm pretty happy with how the test suite turned out. Tests are very concise and to the point, with essentially no boilerplate. I have ported a random selection of tests from brig and galley, plus a number of MLS tests, including federated ones (and made them end-to-end in the process). I hope this gives a reasonable idea of what python tests would look like more generally.

It is not entirely clear to me how we would take this further if this proposal is accepted. The obvious approach would be to follow the same strategy as for servantification: write new tests in the new system, port a test to the new system when you need to change it, and otherwise keep the old test suites for as long as they contain tests.

I think this would work fine, but maybe people have been burned by their experience with servantification and are scared to follow the same route again. But note that, unlike servantifying an endpoint, porting a test to the python suite is quite a painless experience that should in most cases take only a couple of minutes.

## Checklist


integration/tests/helpers/context.py
```diff
@@ -0,0 +1,46 @@
...
+    def request(self, method, url, user=None, conn_id='0', headers=None, **kwargs):
+        if headers is None:
+            headers = {}
+        if user is not None:
+            headers['Z-User'] = obj_id(user)
+            headers['Z-Connection'] = conn_id
+        return Response(method, url, kwargs, 
+                        requests.request(method, url, headers=headers, **kwargs))
+    
+    def __getattr__(self, name):
```
smatting, line 31: Note to reviewers: This functions makes it possible to write `ctx.<any-method>` for `<any-method` that is defined in `api`. That way all api methods don't have to be defined as methods of `Context`

integration/tests/helpers/setup/__init__.py
```diff
@@ -0,0 +1,18 @@
...
+# from .mls import MLS
+
+# import asyncio
+# from contextlib import contextmanager
+# import itertools
+# import json
+# import queue
+# import time
+# import threading
+# import websockets
```
flokli, line 14: :eye:

integration/tests/helpers/setup/__init__.py
```diff
@@ -0,0 +1,18 @@
...
+
+from .ws import *
+from .util import *
+from .mls import *
```
flokli, line 18: Wildcard imports like this are usually discouraged. Can you run this through a linter (flake8/pylint) to get some basic linting?

integration/tests/helpers/setup/mls.py
```diff
@@ -0,0 +1,247 @@
+from ..prekeys import Prekeys
+from .. import api
+from . import util
+from ..conversions import *
```
flokli, line 4: same, relative imports

integration/tests/helpers/setup/mls.py
```diff
@@ -0,0 +1,247 @@
...
+        self.new_members = []
+
+    def cli(self, cid, *args, stdin=None):
+        state = self[cid]
+
+        subst = {}
+
+        group_in_file = None
+        if '<group-in>' in args:
+            group_in_file = tempfile.NamedTemporaryFile(mode='w+', dir=self.directory, delete=False)
```
flokli, line 202: these are not used in a contextmanager, so won't necessarily get cleaned up?

integration/tests/helpers/setup/ws.py
```diff
@@ -0,0 +1,140 @@
...
+import queue
+import json
+import websockets
+import random
+import threading
+import time
+from urllib.parse import urlencode
+
+from .. import api
+from ..conversions import *
```
flokli, line 12: relative imports

integration/tests/helpers/setup/mls.py
pcapriotti: What's wrong with relative imports?

integration/tests/helpers/setup/mls.py
pcapriotti: It will be cleaned up on destruction. That's good enough here. The file is stored in a temporary directory anyway, which is used via a context manager, so it will be deleted when that is cleaned up anyway.

integration/tests/helpers/setup/__init__.py
pcapriotti: Fixed

integration/tests/helpers/setup/mls.py
flokli: Because self.directory itself is cleaned up, and it's created in there? If that's the case, please add a commen.

integration/tests/helpers/setup/mls.py
flokli: https://stackoverflow.com/a/16748366

integration/tests/helpers/context.py
jschaul: How about copying this block of text as a comment above this function for future readers? 🙃

integration/tests/helpers/setup/mls.py
pcapriotti: I'm not sure where it says what is wrong with them. I think they are fine.

integration/tests/brig/user/test_client.py
```diff
@@ -0,0 +1,39 @@
...
+    uid = setup.random_user(ctx)['id']
+
+    with ctx.add_client(uid, [prekeys.pop()], prekeys.lpop(), 
+                        ctype="legalhold", internal=True) as r:
+        assert r.status_code == 201
+        resp = r.json()
+    lh_client_id = resp['id']
+
+    with ctx.delete_client(uid, lh_client_id) as r:
+        assert r.status_code == 404
```
smatting, line 14: we get a 404 here because there is a bug in `delete_client()`: the path is wrong

PR #3111, opened 2023-02-28
Add performance test scripts for large MLS conversations

smatting:
This PR adds:

- `hack/bin/performance.py` - a script that allows to measure performance when creating and sending to large mls conversation
- a python `wire` library in `hack/python` that should help writing more python scripts
- `hack/bin/create_team.py` - a script that allows you to create a team at any backend. This script makes use use of the `wire` library

This PR also deletes various scripts:

- `create_team.sh`
- `create_team_members.sh`
- `create_team_request_code.sh`
- `create_test_team_admins.sh`
- `create_test_team_members.sh`
- `create_test_user.sh`

which should be integrated into the `create_team.py` on a by-need basis


smatting:
If https://github.com/wireapp/wire-server/pull/3109 maybe extract common python modules and use in these scripts

pcapriotti (approved):

PR #3112, opened 2023-02-28
Upgrade rusty-jwt-tools

battermann:
## Checklist


services/brig/test/integration/API/User/Client.hs
```diff
@@ -121,7 +121,7 @@ tests _cl _at opts p db b c g =
       test p "get/head nonce/clients" $ testNewNonce b,
       testGroup
         "post /clients/:cid/access-token"
-        [ test p "success" $ testCreateAccessToken b,
+        [ test p "invalid values" $ testCreateAccessTokenInvalidValues b,
```
stefanwire, line 124: Is there no happy path test possible anymore?

stefanwire (approved):

services/brig/test/integration/API/User/Client.hs
battermann: Unfortunately not. There is no simple way to create a valid DPoP proof.

PR #3120, opened 2023-03-02
Introduce Http2Manager to help with reusing http2 connections

akshaymankar:
## Checklist


akshaymankar:
Code needs a lot of cleanup and integration with servant. But early feedback is very much appreciated.

pcapriotti (commented):
I wonder if there is an easier approach. The `sendRequest` continuation we get from the http client should be all we need in order to access the live connection. So we could store that continuation in the map, and simply call it when we want to make a new request. This would avoid the need for another background thread per connection, I believe.

libs/wire-api-federation/src/Wire/API/Federation/HTTP2.hs
```diff
@@ -0,0 +1,99 @@
...
+
+data HTTP2Conn = HTTP2Conn
+  { backgroundThread :: Async (),
+    -- See comment in 'startPersistentHTTP2Connection' about what this 'Chan'
+    -- contains
+    requestChan :: Chan (HTTP2.Request, HTTP2.Response -> IO (), MVar (Async ()))
+  }
+
+data HTTP2Manager = HTTP2Manager
+  { connections :: MVar (Map (ByteString, Int) HTTP2Conn),
```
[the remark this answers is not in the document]
akshaymankar: Yeah, I just used it without thinking. I will try to find an STM based map.

akshaymankar:
> I wonder if there is an easier approach. The `sendRequest` continuation we get from the http client should be all we need in order to access the live connection. So we could store that continuation in the map, and simply call it when we want to make a new request. This would avoid the need for another background thread per connection, I believe.

Good idea, it has made the logic less convoluted. But I think the need for a background thread per connection is still there as we cannot stop the `run` function and just smuggle the `sendRequest` continuation out of it because the background threads in the http2 client (sender and receiver) are racing with the continuation.

pcapriotti:
> Good idea, it has made the logic less convoluted. But I think the need for a background thread per connection is still there as we cannot stop the `run` function and just smuggle the `sendRequest` continuation out of it because the background threads in the http2 client (sender and receiver) are racing with the continuation.

Or we could run the whole thing in Codensity. Then there is no need to smuggle the continuation out of the bracket.

akshaymankar:
> Or we could run the whole thing in Codensity. Then there is no need to smuggle the continuation out of the bracket.

Hmm, I am not sure at what point we can run `Codensity` in this case. Do you want to pair on this tomorrow morning?

libs/http2-manager/http2-manager.cabal
```diff
@@ -0,0 +1,78 @@
...
+  -- cabal-fmt: expand src
+  exposed-modules:
+    HTTP2.Client.Manager
+    HTTP2.Client.Manager.Internal
+
+  other-modules:    Paths_http2_manager
+  hs-source-dirs:   src
+  ghc-options:
+    -O2 -Wall -Wincomplete-uni-patterns -Wincomplete-record-updates
+    -Wpartial-fields -fwarn-tabs -optP-Wno-nonportable-include-path
```
fisx, line 32: what is `-optP-Wno-nonportable-include-path` for?

libs/wire-api-federation/wire-api-federation.cabal
```diff
@@ -92,6 +92,7 @@ library
     , http-media
     , http-types
     , http2
```
fisx, line 94: do we still want to depend on this?  or is there stuff we can re-export from http2-manager and then drop this (direct) dependency?  is it worth asking these questions?

services/brig/src/Brig/App.hs
```diff
@@ -42,6 +42,7 @@ module Brig.App
     templateBranding,
     requestId,
     httpManager,
+    http2Manager,
```
fisx, line 45: do you want to keep using both?  in other words: is http2-manager strictly better than http-manager, or what are the use cases for `httpManager`?

fisx (commented):
i read all the boring stuff.  still looks good to me!  :-)

libs/http2-manager/test/Test/HTTP2/Client/ManagerSpec.hs
```diff
@@ -0,0 +1,325 @@
...
+    specTemplate (Just localhostCtx)
+
+    it "should error appropriately when the server presents a wrong certificate" $ do
+      ctx <- loadWrongServerSSLContext
+      mgr <- mkTestManager
+      withTestServer (Just ctx) $ \TestServer {..} ->
+        echoTest mgr True serverPort `shouldThrow` (\(_ :: SSL.SomeSSLException) -> True)
+
+specTemplate :: Maybe SSL.SSLContext -> Spec
+specTemplate mCtx = do
```
fisx, line 62: add a test case about shutting down the manager with `putMVar (Right ())`?

libs/http2-manager/test/Test/HTTP2/Client/ManagerSpec.hs
```diff
@@ -0,0 +1,325 @@
...
+withTestServerOnSocket mCtx action (serverPort, listenSock) = do
+  acceptedConns <- newIORef 0
+  liveConns <- newIORef mempty
+  let cleanupServer serverThread = do
+        cancel serverThread
+        mapM_ cancel =<< readIORef liveConns
+  bracket (async $ testServerOnSocket mCtx listenSock acceptedConns liveConns) cleanupServer $ \serverThread ->
+    action TestServer {..}
+
+allocServerConfig :: Either Socket SSL.SSL -> IO Server.Config
```
fisx, line 241: could this be useful for library users, ie. should it be exported from at least the internals?

libs/http2-manager/http2-manager.cabal
akshaymankar: I copy pasted it without thinking much :sweat_smile:

libs/http2-manager/test/resources/localhost-key.pem
```diff
@@ -0,0 +1 @@
+../../../../services/federator/test/resources/unit/localhost-key.pem
```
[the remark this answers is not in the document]
akshaymankar: Already did, nix didn't like the symlink.

libs/wire-api-federation/wire-api-federation.cabal
akshaymankar: IMO, its ok to let people explicitly depend on this. This way we don't have to keep updating this library because the other one changed.

libs/http2-manager/src/HTTP2/Client/Manager/Internal.hs
```diff
@@ -0,0 +1,337 @@
...
+import qualified Network.Socket as NS
+import qualified OpenSSL.Session as SSL
+import qualified System.TimeManager
+import Prelude
+
+data HTTP2Conn = HTTP2Conn
+  { backgroundThread :: Async (),
+    disconnect :: IO (),
+    -- See comment in 'startPersistentHTTP2Connection' about why this returns
+    -- '()'
```
fisx, line 36: would be nice to have that info in the haddocks.

services/brig/src/Brig/App.hs
akshaymankar: We need both unfortunately, we cannot assume everyone outside of our internal components runs http2 compatible servers.

services/brig/src/Brig/App.hs
```diff
@@ -354,6 +359,18 @@ initHttpManager = do
...
+initHTTP2Manager = do
+  ctx <- SSL.context
+  SSL.contextAddOption ctx SSL_OP_NO_SSLv2
+  SSL.contextAddOption ctx SSL_OP_NO_SSLv3
+  SSL.contextAddOption ctx SSL_OP_NO_TLSv1
+  SSL.contextSetCiphers ctx "HIGH"
+  SSL.contextSetVerificationMode ctx $
+    SSL.VerifyPeer True True Nothing
+  SSL.contextSetDefaultVerifyPaths ctx
+  http2ManagerWithSSLCtx ctx
```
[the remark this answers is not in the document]
akshaymankar: I copied what we have for the old manager.

libs/http2-manager/src/HTTP2/Client/Manager/Internal.hs
```diff
@@ -0,0 +1,337 @@
...
+          pure Nothing
+        Just (Right ()) -> do
+          writeTVar (connections mgr) $ Map.delete (tlsEnabled, host, port) conns
+          pure Nothing
+
+-- | Disconnects HTTP2 connection if there exists one. If the background thread
+-- running the connection does not finish within 1 second, it is canceled.
+--
+-- NOTE: Any requests in progress might not finish correctly.
+-- FUTUREWORK: Write a safe version of this function.
```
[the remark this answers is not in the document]
fisx: never mind, we're terminating a *connection* here, that's different from terminating a manager.

what would have to go into a safe version of this function?  you'd have to kill dangling requests eventually, right?  no guarantee they are still live.

libs/wire-api-federation/src/Wire/API/Federation/Client.hs
```diff
@@ -173,19 +176,22 @@ withHTTP2Request mSSLCtx req hostname port k = do
...
-                k
-                  Response
-                    { responseStatusCode = status,
-                      responseHeaders = Seq.fromList headers,
-                      responseHttpVersion = HTTP.http20,
-                      responseBody = result
-                    }
+          sendRequest req $ consumeStreamingResponseWith k
+
+consumeStreamingResponseWith :: (StreamingResponse -> a) -> HTTP2.Response -> a
```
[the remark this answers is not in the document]
akshaymankar: I think being abstract is good, that way this doesn't need any lifts in case there is a wrapper around IO somewhere.

libs/http2-manager/test/Test/HTTP2/Client/ManagerSpec.hs
```diff
@@ -0,0 +1,325 @@
...
+    Client.responseStatus res `shouldBe` Just status200
+    readResponseBody res `shouldReturn` "some body"
+
+spec :: Spec
+spec = do
+  describe "HTTP2.Client.Manager [without TLS]" $ do
+    specTemplate Nothing
+  -- TODO: Perhaps there are some TLS specific test cases we might want to
+  -- write. Like ensure that the hostname validation works (it currently
+  -- does't!).
```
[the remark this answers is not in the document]
akshaymankar: I did write that test, I need to think of more tests, not sure what they are.

libs/http2-manager/src/HTTP2/Client/Manager.hs
```diff
@@ -0,0 +1,11 @@
+module HTTP2.Client.Manager
+  ( HTTP2Manager,
+    TLSEnabled,
```
fisx, line 3:
```suggestion
    TLSEnabled,
    Hostname,
    Port,
```

libs/http2-manager/src/HTTP2/Client/Manager/Internal.hs
```diff
@@ -0,0 +1,337 @@
...
+    sendRequestMVar :: MVar (Either Request ())
+  }
+
+type TLSEnabled = Bool
+
+type HostName = ByteString
+
+type Port = Int
+
+type Target = (TLSEnabled, HostName, Port)
```
fisx, line 46: export this as well and use it more in the function signatures, instead of the 3 arguments?  i think currying is not very common here, and it'd be easier to read, i think.

libs/http2-manager/src/HTTP2/Client/Manager.hs
```diff
@@ -0,0 +1,11 @@
...
+    defaultHTTP2Manager,
+    http2ManagerWithSSLCtx,
+    withHTTP2Request,
+    ConnectionAlreadyClosed (..),
+  )
```
fisx, line 8: also export a way to inspect and change the cache limit here?  (or at least initialize.)

libs/http2-manager/src/HTTP2/Client/Manager/Internal.hs
```diff
@@ -0,0 +1,337 @@
...
+http2ManagerWithSSLCtx :: SSL.SSLContext -> IO HTTP2Manager
+http2ManagerWithSSLCtx sslContext = do
+  connections <- newTVarIO mempty
+  let cacheLimit = 20
+  pure $ HTTP2Manager {..}
+
+-- | Does not check whether connection is actually running. Users should use
+-- 'withHTTP2Request'. This function is good for testing.
+sendRequestWithConnection :: HTTP2Conn -> HTTP2.Request -> (HTTP2.Response -> IO a) -> IO a
+sendRequestWithConnection conn req f = do
```
fisx, line 79:
```suggestion
sendRequestWithConnection conn req k = do
```

libs/http2-manager/src/HTTP2/Client/Manager/Internal.hs
```diff
@@ -0,0 +1,337 @@
...
+  result <- newEmptyMVar
+  threadKilled <- newEmptyMVar
+  putMVar (sendRequestMVar conn) (Left (Request req (putMVar result <=< f) threadKilled))
+  race (takeMVar result) (takeMVar threadKilled) >>= \case
+    Left x -> pure x
+    Right (SomeException e) -> throw e
+
+-- | Make an HTTP2 request, if it is the first time the 'HTTP2Manager' sees this
+-- (server,port) comibination, it creates the connection and keeps it around for
```
fisx, line 88:
```suggestion
-- (tlsenabled,server,port) combination, it creates the connection and keeps it around for
```

libs/http2-manager/src/HTTP2/Client/Manager/Internal.hs
```diff
@@ -0,0 +1,337 @@
...
+-- any subsequent requests. Subsequest requests try to use this connection, in
+-- case the connection is already dead (e.g. the background thread has
+-- finished), a new connection is created.
+--
+-- It is important to consume the response body completely before the
+-- continuation can finish.
+withHTTP2Request :: HTTP2Manager -> TLSEnabled -> ByteString -> Int -> HTTP2.Request -> (HTTP2.Response -> IO a) -> IO a
```
fisx, line 95: use Hostname and Port here.  maybe you should make them newtypes to catch all these.  :-)

libs/http2-manager/src/HTTP2/Client/Manager/Internal.hs
```diff
@@ -0,0 +1,337 @@
...
+    Right (SomeException e) -> throw e
+
+-- | Make an HTTP2 request, if it is the first time the 'HTTP2Manager' sees this
+-- (server,port) comibination, it creates the connection and keeps it around for
+-- any subsequent requests. Subsequest requests try to use this connection, in
+-- case the connection is already dead (e.g. the background thread has
+-- finished), a new connection is created.
+--
+-- It is important to consume the response body completely before the
+-- continuation can finish.
```
fisx, line 94:
```suggestion
-- It is important that the continuation consumes the response body completely
-- before it returns.
```

libs/http2-manager/test/Test/HTTP2/Client/ManagerSpec.hs
```diff
@@ -0,0 +1,325 @@
...
+
+specTemplate :: Maybe SSL.SSLContext -> Spec
+specTemplate mCtx = do
+  it "should be able to make an HTTP2 request" $ do
+    withTestServer mCtx $ \TestServer {..} -> do
+      mgr <- mkTestManager
+
+      echoTest mgr (isJust mCtx) serverPort
+
+      readIORef acceptedConns `shouldReturn` 1
```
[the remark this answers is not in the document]
akshaymankar: I think of my ideal test as three things: Arrange -> Act -> Assert. When I can get it neatly like this, I like to leave some empty space. So, I would prefer if we leave it like this. LMK if you have strong feelings.

libs/http2-manager/src/HTTP2/Client/Manager/Internal.hs
akshaymankar: Newtypes would be annoying, this is going to get stabilized soon, so I think its not necessary. I will find all the places and fix them.

libs/http2-manager/src/HTTP2/Client/Manager/Internal.hs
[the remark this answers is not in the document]
akshaymankar: I think this function is actually graceful now, earlier it was ungraceful, so perhaps I just need to change the name and export it from the main module.

libs/http2-manager/http2-manager.cabal
akshaymankar: Removed.

fisx:
I got a flake on 0199d61d0217daff307e5dcff03b06ef8cdc5da6

```
HTTP2.Client.Manager [without TLS]
  should be able to make an HTTP2 request [✔]
  should fail appropriately when a server is not available [✔]
  should be able to re-use an HTTP2 connection for multiple requests [✔]
  shouldn't try to re-use a disconnected connection [✔]
  should re-use connections even for concurrent requests [✔]
  should re-use the connection even if one of the requests is stuck forever [✔]
  should re-use the connection even if an exception is thrown while handling a response [✔]
  should fail with appropriate error when a dead connection is used [✘]
  should create a new connection when the server restarts [✔]
HTTP2.Client.Manager [with TLS]
  should be able to make an HTTP2 request [✔]
  should fail appropriately when a server is not available [✔]
  should be able to re-use an HTTP2 connection for multiple requests [✔]
  shouldn't try to re-use a disconnected connection [✔]
  should re-use connections even for concurrent requests [✔]
  should re-use the connection even if one of the requests is stuck forever [✔]
  should re-use the connection even if an exception is thrown while handling a response [✔]
  should fail with appropriate error when a dead connection is used [✔]
  should create a new connection when the server restarts [✔]
  should error appropriately when the server presents a wrong certificate [✔]

Failures:

  test/Test/HTTP2/Client/ManagerSpec.hs:162:19: 
  1) HTTP2.Client.Manager [without TLS] should fail with appropriate error when a dead connection is used
       predicate failed on expected exception: SomeException (BadThingHappen Network.Socket.recvBuf: resource vanished (Connection reset by peer))

  To rerun use: --match "/HTTP2.Client.Manager [without TLS]/should fail with appropriate error when a dead connection is used/"

Randomized with seed 804891027
```

fisx:
[this test](https://github.com/wireapp/wire-server/pull/3120/commits/3f47efed12d06fbb719051057ad0d540084ba4c1) also fails.  with this change:

```
@@ -335,7 +335,7 @@ testServer req _ respWriter = do
             flush
             threadDelay =<< randomRIO (0, 100_000)
             delayedBSWriter moreLines bsWriter flush
-          delayedResponse = Server.responseStreaming status200 [] (delayedBSWriter reqBodyLines)
+          delayedResponse = Server.responseStreaming status200 [] (delayedBSWriter $ reverse reqBodyLines)
       respWriter delayedResponse []
     _ -> do
       respWriter (Server.responseNoBody status404 []) []
```

...  it passes, but i don't think this is right?  if it is: could i get an explanation next to the reverse statement?  :-)

akshaymankar:
> Failures:
> 
>   test/Test/HTTP2/Client/ManagerSpec.hs:162:19: 
>   1) HTTP2.Client.Manager [without TLS] should fail with appropriate error when a dead connection is used
>        predicate failed on expected exception: SomeException (BadThingHappen Network.Socket.recvBuf: resource vanished (Connection reset by peer))
> 
>   To rerun use: --match "/HTTP2.Client.Manager [without TLS]/should fail with appropriate error when a dead connection is used/"

I've added this error to the list of expected exceptions. As the comment says it is hard to control which exception is thrown.

fisx:
> > ... it passes, but i don't think this is right? if it is: could i get an explanation next to the reverse statement? :-)
> 
> you found a bug in the test function to read the whole response into a bytestring. It is fixed.
> 
> Could you please suggest a better name for your test? I am a little confused as to what it is trying to test, the test to ensure stuck streams don't block other streams already tests three things which this test seems to test:
> 
>     1. Multiline response
> 
>     2. Responses which come in multiple chunks, (i.e multiple flushes in a streaming response)
> 
>     3. Multiple requests running in parallel with things having different speeds of responses.
> 
> 
> Perhaps you could rephrase the test name so it reads like `it "should ..."` to signify what the http2-manager should be able to do.

i'm not sure, it's supposed to be spotting corner cases we both can't think of right now, quickcheck style.  not sure how well it does that.  how about `"should process many concurrent requests correctly under different timing situations"`?

fisx:
> > Failures:
> > test/Test/HTTP2/Client/ManagerSpec.hs:162:19:
> > 
> > 1. HTTP2.Client.Manager [without TLS] should fail with appropriate error when a dead connection is used
> >    predicate failed on expected exception: SomeException (BadThingHappen Network.Socket.recvBuf: resource vanished (Connection reset by peer))
> > 
> > To rerun use: --match "/HTTP2.Client.Manager [without TLS]/should fail with appropriate error when a dead connection is used/"
> 
> I've added this error to the list of expected exceptions. As the comment says it is hard to control which exception is thrown.

yes, but my point was "can't you change the code so that one of these errors is replaced by the other, so it's easy to tell which one, because there is only one?"

libs/http2-manager/src/HTTP2/Client/Manager/Internal.hs
```diff
@@ -0,0 +1,337 @@
...
+    -- consumer returned something else we would need another empty MVar to
+    -- write the result, this is being dealt with in
+    -- 'sendRequestWithConnection'.
+    responseConsumer :: HTTP2.Response -> IO (),
+    -- | MVar to communicate lack of response due to an exception.
+    exceptionMVar :: MVar SomeException
+  }
+
+-- | Used to close a persistent connection gracefully
+type CloseConnection = ()
```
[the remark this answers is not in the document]
fisx: > the more i think about it, the more i want all type aliases to be changed into newtypes here...

never mind, you already turned me down on this.  your library!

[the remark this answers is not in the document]
akshaymankar: The only user facing type aliases are these:

```

type TLSEnabled = Bool

type HostName = ByteString

type Port = Int

type Target = (TLSEnabled, HostName, Port)
```

I think all of them make using the library more clunky, except for `TLSEnabled`, right now I can use it nicely in a couple of `if` statements, so perhaps its ok?

libs/http2-manager/test/Test/HTTP2/Client/ManagerSpec.hs
```diff
@@ -0,0 +1,325 @@
...
+spec :: Spec
+spec = do
+  describe "HTTP2.Client.Manager [without TLS]" $ do
+    specTemplate Nothing
+  -- TODO: Perhaps there are some TLS specific test cases we might want to
+  -- write. Like ensure that the hostname validation works (it currently
+  -- does't!).
+  describe "HTTP2.Client.Manager [with TLS]" $ do
+    localhostCtx <- runIO $ loadServerSSLContext
+    specTemplate (Just localhostCtx)
```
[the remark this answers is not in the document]
akshaymankar: Why?

libs/http2-manager/src/HTTP2/Client/Manager/Internal.hs
akshaymankar: Added it to the `Request` type.

libs/http2-manager/test/Test/HTTP2/Client/ManagerSpec.hs
```diff
@@ -153,7 +154,11 @@ specTemplate mCtx = do
 
     let brokenRequest = sendRequestWithConnection deadConn (Client.requestBuilder "GET" "/echo" [] "some body") $ \_ -> do
           expectationFailure "Expected no response when request is made to a dead server"
-    brokenRequest `shouldThrow` (\ConnectionAlreadyClosed -> True)
+        -- It is hard to control which one of these will be thrown
```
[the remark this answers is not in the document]
akshaymankar: There also a third kind of error now. I think its OK to have this fail for three different reasons, as it is very difficult in general to tell which one of them is happening. This is also supposed to only happen in weird edge cases, so I am not sure if anyone really wants to deal with these errors.

libs/http2-manager/test/Test/HTTP2/Client/ManagerSpec.hs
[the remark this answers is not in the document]
fisx: "easier to read"?  :)

never mind, i have bad habits.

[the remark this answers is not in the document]
akshaymankar: Yeah, we could fight about this forever, let's just leave it :sweat_smile:

fisx:
i realized that i was too tame with the number of parallel requests: 441664efd4c713144a52889dbdd556c4c3ad33ec

```
Failures:

  test/Test/HTTP2/Client/ManagerSpec.hs:99:3: 
  1) HTTP2.Client.Manager [without TLS] should process many concurrent requests correctly under different timing situations
       uncaught exception: HTTP2Error
       BadThingHappen Network.Socket.sendBuf: resource vanished (Broken pipe)

  To rerun use: --match "/HTTP2.Client.Manager [without TLS]/should process many concurrent requests correctly under different timing situations/"

  test/Test/HTTP2/Client/ManagerSpec.hs:99:3: 
  2) HTTP2.Client.Manager [with TLS] should process many concurrent requests correctly under different timing situations
       uncaught exception: HTTP2Error
       BadThingHappen ProtocolError "error:0A00010F:SSL routines::bad length"

  To rerun use: --match "/HTTP2.Client.Manager [with TLS]/should process many concurrent requests correctly under different timing situations/"
```

akshaymankar:
> (trying to understand, though: should we get an error if max connections has been reached? or should it be blocking? probably not.)

I would say we can leave that for upstream to decide? I think for our current use case this shouldn't really be a problem.

akshaymankar:
Created an upstream issue here: https://github.com/kazu-yamamoto/http2/issues/71

fisx (approved):

PR #3122, opened 2023-03-03
Finalise API v3

pcapriotti:
This finalises API v3, makes it supported, and introduces a new development API v4. All the new MLS endpoints have been removed from v3 and moved to v4.

## Checklist

### DIFF
--- a/libs/wire-api/src/Wire/API/Routes/Version.hs
+++ b/libs/wire-api/src/Wire/API/Routes/Version.hs
@@ -59,7 +59,7 @@ import Wire.API.Routes.Named
 import Wire.API.VersionInfo
 
 -- | Version of the public API.
-data Version = V0 | V1 | V2 | V3
+data Version = V0 | V1 | V2 | V3 | V4
   deriving stock (Eq, Ord, Bounded, Enum, Show)
   deriving (FromJSON, ToJSON) via (Schema Version)
 
@@ -69,7 +69,8 @@ instance ToSchema Version where
       [ element 0 V0,
         element 1 V1,
         element 2 V2,
-        element 3 V3
+        element 3 V3,
+        element 4 V4
       ]
 
 mkVersion :: Integer -> Maybe Version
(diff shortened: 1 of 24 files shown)


libs/wire-api/src/Wire/API/Routes/Version.hs
```diff
@@ -69,7 +69,8 @@ instance ToSchema Version where
       [ element 0 V0,
         element 1 V1,
         element 2 V2,
-        element 3 V3
+        element 3 V3,
+        element 4 V4
```
mdimjasevic, line 73: Recently we've had that discussion about using versions in endpoint paths and how we missed a bug because we were not testing what versions get converted to when made a path component. Do you want to add a test for it in `Test.Wire.API.Routes.Version`?

mdimjasevic (approved):
Great!

I also browsed around in `wire-api` to see if there are any more endpoints that should be moved to V4, but I couldn't find any.

:shipit:

mdimjasevic (changes requested):
Hey, make V4 a development version!

pcapriotti:
> > Hey, make V4 a development version!
> 
> Sorry, I clicked approve, then I realized that `developmentVersions` needs an update in `Wire.API.Routes.Version`. Right?

Ah, right, thanks for catching this.

libs/wire-api/src/Wire/API/Routes/Version.hs
pcapriotti: What test do you mean?

mdimjasevic (approved):

libs/wire-api/src/Wire/API/Routes/Version.hs
mdimjasevic: In `Test.Wire.API.Routes.Version` I'd extend the test like so:

```haskell
testToPathComponent :: Assertion
testToPathComponent = do
  "v0" @=? toPathComponent V0
  "v1" @=? toPathComponent V1
  "v2" @=? toPathComponent V2
  "v3" @=? toPathComponent V3
  "v4" @=? toPathComponent V4
```

fisx (approved):

PR #3123, opened 2023-03-03
[FS-1588] No proposals after deleting a subconversation

mdimjasevic:
This makes sure the database is cleaned up from leftover backend-generated remove proposals once a subconversation is deleted. Due to our test setup, this is not so straightforward to test, and doing a database test is not something we have been doing lately.

The test added by the PR just makes sure a recreated subconversation has no leftover proposals, but that holds because the recreated subconversation has a fresh group ID, hence fetching proposals for that new group ID yields no results anyway.

Tracked by https://wearezeta.atlassian.net/browse/FS-1588.

## Checklist


services/galley/src/Galley/Cassandra/SubConversation.hs
```diff
@@ -115,7 +122,13 @@ interpretSubConversationStoreToCassandra = interpret $ \case
   SetSubConversationEpoch cid sconv epoch -> embedClient $ setEpochForSubConversation cid sconv epoch
   DeleteGroupIdForSubConversation groupId -> embedClient $ deleteGroupId groupId
   ListSubConversations cid -> embedClient $ listSubConversations cid
-  DeleteSubConversation convId subConvId -> embedClient $ deleteSubConversation convId subConvId
+  DeleteSubConversation convId subConvId -> do
+    msub <- embedClient (selectSubConversation convId subConvId)
+    case msub of
+      Nothing -> embed @IO . UnliftIO.throwIO $ errorToWai @'ConvNotFound
```
stefanwire, line 128: Why raising the exception and not making this call idempotent by just accepting that the intended state (delete conversation) has been reached?

mdimjasevic: That's a fair point! I'll make this a no-op in case it is not found. I suppose this can happen when two concurrent requests to delete the same subconversation occur.

mdimjasevic: @stefanwire , done in commit d904e392e.

services/galley/src/Galley/Cassandra/SubConversation.hs
```diff
@@ -115,7 +119,10 @@ interpretSubConversationStoreToCassandra = interpret $ \case
   SetSubConversationEpoch cid sconv epoch -> embedClient $ setEpochForSubConversation cid sconv epoch
   DeleteGroupIdForSubConversation groupId -> embedClient $ deleteGroupId groupId
   ListSubConversations cid -> embedClient $ listSubConversations cid
-  DeleteSubConversation convId subConvId -> embedClient $ deleteSubConversation convId subConvId
+  DeleteSubConversation convId subConvId -> do
+    msub <- embedClient (selectSubConversation convId subConvId)
+    for_ msub $ deleteAllProposals . cnvmlsGroupId . scMLSData
```
stefanwire, line 124: Why explicitly deleting the pending proposals? Essentially, this is buying Cassandra space and paying with users' time. I'd argue that users' time is more valuable. Also, the Cassandra records have a TTL and will be deleted eventually _and_ they will not appear in a recreated subconversation, since the new subconversation's groupID is a different one than the one of the subconversation before. This is also demonstrated by your test which passes even without changes in business logic.

stefanwire: Sorry for not seeing and commenting this when I checked this PR previously!

mdimjasevic: Your reasoning is perfectly fine, but if we're to accept this argument, I'd prefer to hear arguments for the deleting a subconversation operation to be on a critical path, ideally backed by numbers showing this DB clean-up noticeably slows down the response time. We could let the proposals linger until they're garbage-collected, but to me this seems not so great.

I expect this delete DB statement to be deleting one or two records on average, and in rare case a handful of them. When this is put next to so many DB operations we do anyway in this endpoint handler, I'd argue the introduced time overhead is negligible.

stefanwire: Can you raise this in the chat? I would argue that removing the records was not forgotten, but deliberately skipped. As to the critical path, ending a call should be done as quick as possible. What would you do if you hang up and nothing happens for a moment? We need a second opinion/review if you insist on cleaning up the non-user-facing database in exchange for user-facing latency.

mdimjasevic:
Any further comments?

stefanwire (approved):

PR #3131, opened 2023-03-06
stern: make api version check a unit test

smatting:
This PR removes the runtime check introduced https://github.com/wireapp/wire-server/pull/3036 and replaces it with a unit test.

Also the overly restrictive test condition
```
maximum (vinfoSupported vinfo) == backendApiVersion
```

is replaced with
```
backendApiVersion `elem` (vinfoSupported vinfo)
```

### DIFF
--- a/changelog.d/3-bug-fixes/stern-api-check
+++ b/changelog.d/3-bug-fixes/stern-api-check
@@ -1,0 +1,1 @@
+Remove overly restricte api check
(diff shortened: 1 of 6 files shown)


stefanwire (approved):

changelog.d/3-bug-fixes/stern-api-check
```diff
@@ -0,0 +1 @@
+Remove overly restricte api check
```
fisx, line 1: arguably it's overly restrictive to make sure we're running into this early, and not when we *have* to make the version bump.  but that's clearly arguable, yes.  :)

PR #3137, opened 2023-03-08
[SQSERVICES-1931] wire server allow backend to enforce stronger password restrictions when setting a password

battermann:
https://wearezeta.atlassian.net/browse/SQSERVICES-1931

- Changed `PlainTextPassword`  type so that it contains a text range according to the previous constraint with a min of 6 characters
- Introduced a new type `PlainTextPasswordMinLength8` with a text range with a min of 8 characters
- Used the new type with min length 8 in the following places:
  - `NewProviderResponse`
  - `NewUser`
  - `NewRawUser`
  - `PasswordChange`
  - `CompletePasswordReset`
- Updated golden tests
- Updated integration tests
- Added integration tests which verifies that passwords with less than 8 chars are rejected with a 400 status code for the following scenarios:
  - register user
  - set/change password
  - password reset
- Introduced convenience functions `plainTextPasswordUnsafe` and `plainTextPasswordLegacyUnsafe` to use where we need to construct a password from a string (in the prod code, these should be used with caution)
- `toLegacy` exists to convert from the new password type to the legacy type, which should always succeed because the constraints of the new type are stronger

It's currently not tested that login with an existing password with a length of less 8 characters still works. But according to how this is implemented, there is no reason it should not. Unfortunately, this would be very difficult to test, as it is not possible to create test users with shorter passwords anymore.

## Checklist


battermann:
I am open to other naming suggestions, e.g. call the old/legacy password type `PlainTextPasswordLegacy` and the new one just `PlainTextPassword`?

fisx (commented):
running gundeck integration tests locally i get this:

```
    Tokens
      register a push token:                                          FAIL
        Exception: Safe.readNote no parse on "NO_HEADER_VALUE", unable to parse Location header
          readNote, called at test/integration/API.hs:1253:7 in main:API
        Use -p '!/RealAWS/&&/Tokens.register a push token/' to rerun this test only.
      unregister a push token:                                        FAIL
        Exception: Safe.readNote no parse on "NO_HEADER_VALUE", unable to parse Location header
          readNote, called at test/integration/API.hs:1253:7 in main:API
        Use -p '!/RealAWS/&&/unregister a push token/' to rerun this test only.
    Websocket pingpong
      data-level pings produce pongs:                                 OK (1.00s)
      control pings with payload produce pongs with the same payload: OK (1.00s)
      data non-pings are ignored:                                     OK (2.00s)
    Redis migration
      redis migration should work:                                    FAIL
        Exception: Error executing request: Safe.readNote no parse on "NO_HEADER_VALUE", unable to parse Location header
          readNote, called at test/integration/API.hs:1253:7 in main:API
          error, called at src/Bilge/Assert.hs:101:18 in bilge-0.22.0-inplace:Bilge.Assert
          <!!, called at src/Bilge/Assert.hs:109:19 in bilge-0.22.0-inplace:Bilge.Assert
          !!!, called at test/integration/API.hs:966:7 in main:API
        Use -p '!/RealAWS/&&/redis migration should work/' to rerun this test only.
```

it doesn't look related, but it's only happening on this branch, not on develop.

> [...] it is not possible to create test users with shorter passwords anymore.

you could either call cassandra directly and inject the short password, or implement an /i/-endpoint that allows this.  the former used to happen in the spar tests, but somebody removed it, so maybe we should aim for the second option.

let's pair on this on monday?

fisx:
before b2adfe056f3157f71cd0ab7466f090887112eb70:

```
$ make ci package=gundeck TASTY_PATTERN='!/RealAWS/&&/redis migration should work/'
[...]
Gundeck
  API tests
    Redis migration
      redis migration should work: 2023-03-13T10:01:10Z, D, Connecting to 127.0.0.1:9042
FAIL
        Exception: Error executing request: Safe.readNote no parse on "NO_HEADER_VALUE", unable to parse Location header
          readNote, called at test/integration/API.hs:1253:7 in main:API
          error, called at src/Bilge/Assert.hs:101:18 in bilge-0.22.0-inplace:Bilge.Assert
          <!!, called at src/Bilge/Assert.hs:109:19 in bilge-0.22.0-inplace:Bilge.Assert
          !!!, called at test/integration/API.hs:966:7 in main:API

1 out of 1 tests failed (0.07s)
[...]
```

after b2adfe056f3157f71cd0ab7466f090887112eb70:

```
$ make ci package=gundeck TASTY_PATTERN='!/RealAWS/&&/redis migration should work/'
[...]
Gundeck
  API tests
    Redis migration
      redis migration should work: 2023-03-13T10:02:01Z, D, Connecting to 127.0.0.1:9042
OK (0.33s)

All 1 tests passed (0.35s)
```

I'd really like to know how this didn't trigger on CI *or* on your machine @battermann...

fisx:
ah, maybe something to do with `RealAWS`?

fisx (approved):
LGTM now!  :)

please take a look at the changes (i think commit-by-commit is easiest, and you can ignore all the automatic renaming noise), and merge if you agree with everything.

battermann:
> > [...] it is not possible to create test users with shorter passwords anymore.
> 
> you could either call cassandra directly and inject the short password, or implement an /i/-endpoint that allows this. the former used to happen in the spar tests, but somebody removed it, so maybe we should aim for the second option.
> 
> let's pair on this on monday?

I added this test.

PR #3138, opened 2023-03-09
allow rate_limit_unlimited to also have an effect on authenticated re…

jschaul:
…quests

See also https://github.com/zinfra/cailleach/pull/1574

## Checklist

### DIFF
--- a/charts/nginz/templates/conf/_nginx.conf.tpl
+++ b/charts/nginz/templates/conf/_nginx.conf.tpl
@@ -162,8 +162,6 @@ http {
   limit_req_log_level warn;
   limit_conn_log_level warn;
 
-  # Limit by $zauth_user if present and not part of rate limit exemptions
-  limit_req zone=reqs_per_user burst=20;
   limit_conn conns_per_user 25;
 
   #
@@ -255,6 +253,12 @@ http {
         limit_req zone=reqs_per_addr burst=10 nodelay;
         limit_conn conns_per_addr 20;
                 {{- end }}
+              {{- end }}
+            {{- else }}
+              {{- if ($location.unlimited_requests_endpoint) }}
+                 # Note that this endpoint has no rate limit per user for autenticated requests
+              {{- else }}
+                 limit_req zone=reqs_per_user burst=20;
               {{- end }}
             {{- end }}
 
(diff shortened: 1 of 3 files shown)


charts/nginz/templates/conf/_nginx.conf.tpl
```diff
@@ -256,6 +254,12 @@ http {
         limit_conn conns_per_addr 20;
                 {{- end }}
               {{- end }}
+            {{- else }}
+              {{- if ($location.unlimited_requests_endpoint) }}
+                 # Note that this endpoint has no rate limit per user for autenticated requests
+              {{- else }}
+                 limit_req zone=reqs_per_user burst=20;
+              {{- end }}
```
akshaymankar, line 262: Few lines above there is also `limit_req zone=reqs_per_addr burst=10 nodelay;`. Do you think this will also cause issues?

jschaul: That is inside the "disable-zauth" block; so should not cause effects.

Overall, we don't have tests for rate limiting, unfortunately. Perhaps we should write some?

akshaymankar (approved):
Looks good. I am slightly confused by the commit message and the PR title: where is `rate_limit_unlimited` defined? I only see `unlimited_requests_endpoint`

jschaul:
> Looks good. I am slightly confused by the commit message and the PR title: where is `rate_limit_unlimited` defined? I only see `unlimited_requests_endpoint`

Ah thanks, I got the words all tangled up. Changed description and changelog to use the same word as in the actual settings.

jschaul:
With this PR, the config diff for the assets endpoint is this:

```diff
          location ~* ^(/v[0-9]+)?/assets/(.*) {
          # ...
+                      # Note that this endpoint has no rate limit per user for authenticated requests
          # ...
```

whereas for other authenticated endpoints:

```diff
          location ~* ^(/v[0-9]+)?/system/settings$ {
          # ...
+                      limit_req zone=reqs_per_user burst=20;
          # ...
```

and at the top

```diff
-       # Limit by $zauth_user if present and not part of rate limit exemptions
-       limit_req zone=reqs_per_user burst=20;
```

PR #3140, opened 2023-03-09
New ingress controller chart

jschaul:
## Main change:

New 'ingress-nginx-controller' wrapper chart compatible with kubernetes versions [1.23 - 1.26]. The old one 'nginx-ingress-controller' (compatible only up to k8s 1.19) is now DEPRECATED.
We advise to upgrade your version of kubernetes in use to 1.23 or higher (we tested on kubernetes version 1.26), and to make use of the new ingress controller chart. Main features:
- up-to-date nginx version ('1.21.6')
- TLS 1.3 support (including allowing specifying which cipher suites to use)
- security fixes
- no more accidental logging of Wire access tokens under specific circumstances [jira ticket](
https://wearezeta.atlassian.net/browse/SEC-47)

The 'kind: Ingress' resources installed via 'nginx-ingress-services' chart remain compatible with both the old and the new ingress controller, and k8s versions [1.18 - 1.26]. In case you upgrade an existing kubernetes cluster (not recommended), you may need to first uninstall the old controller before installing the new controller chart.

In case you have custom overrides, you need to modify the directory name and top-level configuration key:

```diff
# If you have overrides for the controller chart (such as cipher suites), ensure to rename file and top-level key:
-# nginx-ingress-controller/values.yaml
+# ingress-nginx-controller/values.yaml
-nginx-ingress:
+ingress-nginx:
   controller:
     # ...
```

and double-check if all overrides you use are indeed provided under the same name by the upstream chart. See also the default overrides in [the default values.yaml](https://github.com/wireapp/wire-server/blob/develop/charts/ingress-nginx-controller/values.yaml).

In case you use helmfile change your ingress controller like this:

```diff
# helmfile.yaml
releases:
-  - name: 'nginx-ingress-controller'
+  - name: 'ingress-nginx-controller'
     namespace: 'wire'
-    chart: 'wire/nginx-ingress-controller'
+    chart: 'wire/ingress-nginx-controller'
     version: 'CHANGE_ME'
```

For more information read the documentation under https://docs.wire.com/how-to/install/ingress.html (or go to https://docs.wire.com and search for "ingress-nginx-controller")

## Other internal changes:

- integration tests on CI will use either the old or the new ingress controller; depending on which kubernetes version they run on.
- upgrade `kubectl` to default from the nixpkgs channel (currently `1.26`) by removing the manual version pin on 1.19
- upgrade `helmfile` to default from the nixpkgs channel by removing the manual version pin
- upgrade `helm` to default from the nixpkgs channel by removing the manual version pin
- add `kubelogin-oidc` so the kubectl in this environment can also talk to kubernetes clusters using OIDC

## Checklist


supersven (approved):
Great job :+1:

hack/bin/integration-setup-federation.sh
```diff
@@ -20,13 +20,22 @@ ${DIR}/integration-cleanup.sh
...
 mkdir -p ~/.parallel && touch ~/.parallel/will-cite
 printf '%s\n' "${charts[@]}" | parallel -P "${HELM_PARALLELISM}" "$DIR/update.sh" "$CHARTS_DIR/{}"
 
 # FUTUREWORK: use helm functions instead, see https://wearezeta.atlassian.net/browse/SQPIT-723
 echo "Generating self-signed certificates..."
 
+KUBERNETES_VERSION_MAJOR="$(kubectl version -o json | jq -r .serverVersion.major)"
+KUBERNETES_VERSION_MINOR="$(kubectl version -o json | jq -r .serverVersion.minor)"
+export KUBERNETES_VERSION="$KUBERNETES_VERSION_MAJOR.$KUBERNETES_VERSION_MINOR"
+if (( KUBERNETES_VERSION_MINOR >= 23 )); then
```
comawill, line 33:
```suggestion
if (( KUBERNETES_VERSION_MAJOR > 1 || KUBERNETES_VERSION_MAJOR == 1 && KUBERNETES_VERSION_MINOR >= 23 )); then
```

even though a v2.x seems not to exist yet, this should prevent unexpected breaks when `KUBERNETES_VERSION_MINOR` gets below 23 with a major release

hack/bin/integration-teardown-federation.sh
```diff
@@ -12,6 +12,13 @@ export NAMESPACE_2="$NAMESPACE-fed2"
 export FEDERATION_DOMAIN_1="."
 export FEDERATION_DOMAIN_2="."
 
+KUBERNETES_VERSION_MINOR="$(kubectl version -o json | jq -r .serverVersion.minor)"
+if (( KUBERNETES_VERSION_MINOR >= 23 )); then
```
comawill, line 16:
```suggestion
if (( KUBERNETES_VERSION_MAJOR > 1 || KUBERNETES_VERSION_MAJOR == 1 && KUBERNETES_VERSION_MINOR >= 23 )); then
```

hack/bin/integration-teardown.sh
```diff
@@ -6,6 +6,12 @@ TOP_LEVEL="$DIR/../.."
 NAMESPACE=${NAMESPACE:-test-integration}
 # doesn't matter for destruction but needs to be set
 export FEDERATION_DOMAIN="."
+KUBERNETES_VERSION_MINOR="$(kubectl version -o json | jq -r .serverVersion.minor)"
+if (( KUBERNETES_VERSION_MINOR >= 23 )); then
```
comawill, line 10:
```suggestion
if (( KUBERNETES_VERSION_MAJOR > 1 || KUBERNETES_VERSION_MAJOR == 1 && KUBERNETES_VERSION_MINOR >= 23 )); then
```

comawill (approved):
Awesome 👍 LGTM

charts/ingress-nginx-controller/values.yaml
```diff
@@ -0,0 +1,44 @@
...
+    service:
+      type: NodePort  # or LoadBalancer
+      externalTrafficPolicy: Local
+      nodePorts:
+        # The nginx instance is exposed on ports 31773 (https) and 31772 (http)
+        # on the node on which it runs. You should add a port-forwarding rule
+        # on the node or on the loadbalancer that forwards ports 443 and 80 to
+        # these respective ports.
+        https: 31773
+        http: 31772
```
flokli, line 21: A more opportunistic default here would be to use `type: Deployment` and `type: LoadBalancer`, and document how to configure this if you have to resort to NodePort.

The NodePort approach always requires manual configuration of some external load balancer/firewall to round-robin between node IPs and is error-prone. It's also a bit annoying to have to decide on some global ports that may not be used otherwise.

Most managed K8s clusters have support for LoadBalancers, you can also get this for your own clusters in hcloud etc. It's even possible to do it for pure bare metal, without any "load balancer hardware", by using BGP or some leadership election over who's announcing the "load balancer ip" via ARP (https://metallb.universe.tf/configuration/_advanced_l2_configuration/).

jschaul: I swapped the defaults as you suggested and documented how to get the previous behaviour back.

docs/src/how-to/install/ingress.md
```diff
@@ -0,0 +1,102 @@
...
+      type: NodePort
+```
+
+Then, on each of your kubernetes worker nodes, two ports are exposed: ports
+31773 (https) and 31772 (http)
+
+You should add a port-forwarding rule on the node or on the loadbalancer that
+forwards ports 443 and 80 to these respective ports. Any traffic hitting the http port is simply getting a http 30x redirect to https.
+
+Downsides of this approach: The NodePort approach always requires manual configuration of some external load balancer/firewall to round-robin between node IPs and is error-prone. It's also a bit annoying to have to decide on some global ports that may not be used otherwise.
```
flokli, line 72: We now do set some ports for the NodePort approach (31773 and 31772), even though we don't use them as long as the user doesn't explicitly set `type: NodePort`. We might want to remove them from `values.yaml`, and explicitly let the user decide (and put it in the example in L63), so this paragraph makes sense.

docs/src/how-to/install/ingress.md
```diff
@@ -0,0 +1,102 @@
...
+
+Most managed K8s clusters have support for LoadBalancers, you can also get this for your own clusters in hcloud etc. It's even possible to do it for pure bare metal, without any "load balancer hardware", by using BGP or some leadership election over who's announcing the "load balancer ip" via ARP (https://metallb.universe.tf/configuration/_advanced_l2_configuration/).
```
flokli, line 74: This is okay for a github discussion, but a bit too sloppy for public docs.

What about

> Most managed K8s clusters have support for LoadBalancers.
> Manually set up K8S Clusters can also support this, by using a provider/environment-specific CCM
>  (see [hcloud](https://github.com/hetznercloud/hcloud-cloud-controller-manager) and [digitalocean](https://github.com/digitalocean/digitalocean-cloud-controller-manager) for examples).
> In case you're provisioning on bare metal, without any hardware load balancer support in front,
> you might be using [MetalLB](https://metallb.universe.tf/), which supports BGP or Failover L2 ARP announcements.
> The choice of CCM highly depends on the environment you choose to deploy wire-server in.

docs/src/how-to/install/ingress.md
jschaul: Good point. https://github.com/wireapp/wire-server/pull/3173

PR #3143, opened 2023-03-13
Add `pendingTestCase` command and use it.

fisx:
This should make life slightly more bearable for everybody including concourse, while still allowing to run the pending tests locally by changing the definition to `pendingTestCase = testCase`.

## Checklist


fisx:
@akshaymankar I know you're opposed to this, but I don't think it's a sustainable situation that we have to run the test suite 2-3 times consistently *every time* we want to get a PR green.  With `pendingTestCase`, we can keep track of where we need to fix things, while not being in the way of everybody else.

Please please please say ok!

akshaymankar:
What process do imagine around this? When we mark something as pending, how and when does it get addressed?

fisx:
> What process do you imagine around this? When we mark something as pending, how and when does it get addressed?

same process as now, only we don't disrupt everybody else: somebody picks it up and fixes it as time permits.

i've also added a make rule to list pending tests, and a shell variable to run them anyway.  the important thing is that we don't set this shell variable on concourse.  maybe we can set it in a pipeline where we build release artefacts?

fisx:
> maybe we can set it in a pipeline where we build release artefacts?

if you want to pair on this let me know.

elland (approved):

Makefile
```diff
@@ -270,8 +275,9 @@ ifeq ($(INTEGRATION_FEDERATION_TESTS), 1)
 	$(EXE_SCHEMA) --keyspace $(package)_test2 --replication-factor 1 --reset
 endif
 endif
-	./dist/brig-index reset --elasticsearch-index directory_test --elasticsearch-server http://localhost:9200 > /dev/null
-	./dist/brig-index reset --elasticsearch-index directory_test2 --elasticsearch-server http://localhost:9200 > /dev/null
+	@echo "brig-index reset currently does not work, see https://github.com/wireapp/wire-server/pull/3114"
+#	./dist/brig-index reset --elasticsearch-index directory_test --elasticsearch-server http://localhost:9200 > /dev/null
+#	./dist/brig-index reset --elasticsearch-index directory_test2 --elasticsearch-server http://localhost:9200 > /dev/null
```
jschaul, line 280: please rebase on develop; this change should not be necessary anymore.

fisx:
integration tests pass locally.  i don't know why concourse doesn't run them on this PR.

PR #3144, opened 2023-03-13
Upgrade cachix to 1.3.1

smatting:
This PR updates to cachix 1.3.1 in hopes of solving the issue https://github.com/cachix/cachix/issues/509

I've added a separate nixpkgs pin for cachix to avoid having to upgrade all the things and quickly try out if upgrading solves the issue.
If this PR fixes it we can take some time to upgrade our main `nixpkgs` pin and use cachix from there. If not we can consider reverting this PR.


smatting:
Failing compilation is due to a unit test flake.
The hoogle job didn't fail, so let's try this on more runs of the job by merging this PR.

jschaul (approved):

PR #3148, opened 2023-03-13
Revert "Use openssl instead of tls in federator http2 client"

smatting:
Reverts wireapp/wire-server#3051
Context https://wearezeta.atlassian.net/browse/FS-1609


mdimjasevic:
@smatting , will you make a PR to the Frida release branch too? I can do it if you've got something else lined up.

PR #3150, opened 2023-03-14
[FS-1147] Proteus: Support creating a conversation when remote backends are unavailable

mdimjasevic:
Tracked by https://wearezeta.atlassian.net/browse/FS-1147.

## Checklist


elland (approved):
I'd still re-use the handler but final decision is yours.

mdimjasevic:
> I'd still re-use the handler but final decision is yours.

I did reuse the handler. It's just the return type that is parameterised to fit both the `< V4` and the `>= V4` case.

PR #3151, opened 2023-03-14
retry-integration-tests-once

jschaul:
This might be a bad idea. If we introduce this, it has the advantage of not needing to manually retrigger the integration tests if they fail due to a (rare) flake.

Ideally we would try and improve those tests which are flaky. Right now we don't have a process that works very well for this however (might be too hard; no time; working on other things; not feeling responsible; etc). So we could combine this PR with e.g. a regular "let's deflake existing tests" effort happening e.g. weekly or bi-weekly (just an idea).

## Checklist


flokli:
I'm not a fan. Deflaking indeed is currently not a very ejoyable process, due to the super slow feedback cycle (time from commit until CI feedback).

We should try to speed up this feedback cycle - by speeding up CI, or making it easier to run these locally (in a toy cluster), skipping some of the unnecessary steps before.

This will make it less likely for the flaky tests to get tackled, because we just put them further out of sight.

PR #3153, opened 2023-03-15
fix: paths in stern

battermann:
## Checklist


fisx (approved):
looks harmless, but what's the fix?  what goes wrong if we add leading `/`, or use `/` in strings in `paths`?  isn't it bad that we're so brittle to make distinctions there?  i wonder where this should be explained?  maybe in bilge, in the haddocks of `paths`?

PR #3159, opened 2023-03-16
fix: make nginz OAuth secret optional

battermann:


charts/nginz/templates/secret.yaml
```diff
@@ -15,5 +15,7 @@ data:
   {{- with .Values.secrets }}
   zauth.conf: {{ .zAuth.publicKeys | b64enc | quote }}
   basic_auth.txt: {{ .basicAuth | b64enc | quote }}
+  {{- if .oAuth.publicKeys }}
```
jschaul, line 18: this will still fail if `.oAuth` doesn't exist.

```suggestion
  {{- if .oAuth }}
```

PR #3164, opened 2023-03-19
add section about having more than 500 users

arthurwolf:
## Checklist

### DIFF
--- a/docs/src/how-to/install/infrastructure-configuration.md
+++ b/docs/src/how-to/install/infrastructure-configuration.md
@@ -693,11 +693,11 @@ This can be changed in the Brig config, with this option:
    
 ```{note}
 
-   Note however that some messages need to be fanned out to all members of a team, and due to End to End Encryption (E2EE), this is very computationally expensive for large numbers of recipients.
+If you create a team with more than 2000 members then clients won't receive certain team update events (e.g. new member joining) live via websocket anymore, but most of the app will still function normally.
 
-   Therefore, it is recommended to leave the maximum number of users at 500, and to instead create additional teams.
+Irrespective of team size conversations can have at most 2000 members. This limit cannot be overridden.
 
-   Very large teams could cause issues with dropped messages and with Legal Hold being unavailable.
+Wire's backend currently only supports fanning out a single message to at most 2000 recipients, hence the limitations on conversation size. Increasing this limit is work in progress.
 
 ```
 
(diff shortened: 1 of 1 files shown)


docs/src/how-to/install/infrastructure-configuration.md
```diff
@@ -677,3 +677,30 @@ cargohold:
...
+
+```
+   
+```{note}
+
+   Note however that some messages need to be fanned out to all members of a team, and due to End to End Encryption (E2EE), this is very computationally expensive for large numbers of recipients.
+
+   Therefore, it is recommended to leave the maximum number of users at 500, and to instead create additional teams.
+
+   Very large teams could cause issues with dropped messages and with Legal Hold being unavailable.
```
smatting, line 700: The backend event system (gundeck) is really the reason for these limitations. The Proteus e2ee protocol leads to computationally expensive operations for large groups, but it would still work. The backend is just not designed currently to handle large fanouts. We have it on the roadmap to scale gundeck.

```suggestion
If you create a team with more than 2000 members then clients won't receive certain team update events (e.g. new member joining) live via websocket anymore, but the most of the app will still function normally.

Irrespective of team size the every conversation can have at most 2000 members. This limit cannot be overridden.

Wire's backend currently only supports fanning out a single message to at most 2000 recipients, hence the limitations on conversation size. Increasing this limit is work in progress.
```

docs/src/how-to/install/infrastructure-configuration.md
```diff
@@ -677,3 +677,30 @@ cargohold:
...
+
+If you create a team with more than 2000 members then clients won't receive certain team update events (e.g. new member joining) live via websocket anymore, but most of the app will still function normally.
+
+Irrespective of team size conversations can have at most 2000 members. This limit cannot be overridden.
+
+Wire's backend currently only supports fanning out a single message to at most 2000 recipients, hence the limitations on conversation size. Increasing this limit is work in progress.
+
+```
+
+
```
julialongtin, line 704:
```suggestion
```

docs/src/how-to/install/infrastructure-configuration.md
```diff
@@ -677,3 +677,29 @@ cargohold:
...
+
```
julialongtin, line 705:
```suggestion
```

julialongtin (approved):

PR #3171, opened 2023-03-22
New integration test suite

smatting:
See README.md

This PR introduces a new integration test suite that is meant to test accross multiple services
What is already included:
- making requests to API, e.g. `baseRequest`, `submit`, `bindResponse`
- working with JSON-like values as the main workhorse, e.g. `%.`, `setField`
- making assertions, e.g. `assertBool`, `shouldMatch`
- Starting services with modified configuration `withModifiedServices`
- Opening websockets and selecting events from it, e.g. `awaitMatch`


integration/test/TestLib/App.hs
```diff
@@ -0,0 +1,930 @@
...
+
+data AppFailure = AppFailure String
+
+instance Show AppFailure where
+  show (AppFailure msg) = msg
+
+instance Exception AppFailure where
+  displayException (AppFailure msg) = msg
+
+failApp :: String -> App a
```
fisx, line 86: `Text` instead of `String` has my vote.  not for performance (I think that wouldn't be significant), but because it seems more natural to me.

integration/test/API.hs
```diff
@@ -0,0 +1,144 @@
...
+        name = Nothing,
+        team = False
+      }
+
+createUser :: CreateUser -> App Response
+createUser cu = do
+  email <- maybe randomEmail pure cu.email
+  let password = fromMaybe defPassword cu.password
+      name = fromMaybe email cu.name
+  req <- baseRequest Brig Unversioned "/i/users"
```
fisx, line 51: I think ideally we should do all the scaffolding via the public api, like god wants us to, and only *inspect* the state of the services via internal api where needed.  I've found situations in the past where a test created a corrupt state that was just plausible enough for one test, and when I added more logic to the test it got pretty ugly.

Of course that's a lot more work for eg. this function, where we'd need to model the entire invitation flow.  Maybe we can agree to allow scaffolding via internal api as a quick hack as long as there is a `TODO: use public api` in the haddocks?

integration/test/API.hs
```diff
@@ -0,0 +1,144 @@
...
+  submit "POST" $
+    addJSONObject
+      ( [ "email" .= email,
```
fisx, line 54: Is there a way to splice variables into aeson-qq-generated json, or something as neat?  That's probably a nice to have, though...

integration/test/API.hs
smatting: Yes, agreed. For inviting users I don't know any alternative however.

integration/test/TestLib/App.hs
```diff
@@ -0,0 +1,930 @@
...
+
+-- |
+--
+-- Table of Contents:
+-- - SECTION_APP : App, Env, ServiceMap, Context
+-- - SECTION_ASSERTIONS
+-- - SECTION_JSON
+-- - SECTION_REQUEST : requests and responses
+-- - SECTION_PRINTING : Terminal and printingmodule App where
+-- - SECTION_MODSERVICE : Start services with modified configuration
```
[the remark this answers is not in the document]
smatting: We decided it's not worth yet to split the module

charts/wire-server/requirements.yaml
```diff
@@ -115,3 +115,8 @@ dependencies:
   repository: "file://../sftd"
   tags:
     - sftd
+- name: integration
```
fisx, line 118: probably a dumb question, but why is this not loaded in prod?

integration/test/Test/Brig.hs
```diff
@@ -0,0 +1,20 @@
...
+testSearchContactForExternalUsers :: HasCallStack => App ()
+testSearchContactForExternalUsers = do
+  owner <- randomUser def {Internal.team = True}
+  partner <- randomUser def {Internal.team = True}
+
+  bindResponse (Internal.putTeamMember (partner %. "id") (partner %. "team") API.teamRolePartner) $ \resp ->
+    resp.status `shouldMatchInt` 200
+
+  bindResponse (Public.searchContacts (partner %. "id") (owner %. "name")) $ \resp ->
+    resp.status `shouldMatchInt` 200
```
fisx, line 20:
```suggestion
    resp.status `shouldMatchInt` 403
```

(https://github.com/wireapp/wire-server/pull/3252 has been merged)

integration/README.md
```diff
@@ -0,0 +1,30 @@
...
+2. `make devtest test=TESTNAME` to start a ghcid that re-runs the test after each succesful build
+
+
+Original design guidelines / goals:
+
+- Don't import `wire-api` or any other internal Haskell libraries. This goal of this test suite is test against the public API interface.
+- Make input of functions as generic as possible, i.e. use `ProducesJSON` as much as possible. That way users can provide values without binding or converting.
+- Make it as easy to debug tests as possible
+  * Add the `HasCallStack` constraint to any function
+  * Prefer to make a function monadic and throw helpfull error message rather than keeping it pure (purity has little value in tests)
```
fisx, line 21:
```suggestion
  * Prefer to make a function monadic and throw a helpfull error message rather than keeping it pure (purity has little value in tests)
```

integration/README.md
```diff
@@ -0,0 +1,30 @@
...
+  * e.g. show full request/responses by default upon exceptions (use `bindResponse`)
+  * all notifications received on websocket are shown in case of failure
+  * always show JSON in a prettified way (see `prettyJSON`)
+- unit testing libraries such as `HUnit` or `hspec` are not imported. The equivalent functions they provided that are better integrated, for example avoiding using `liftIO`, and providing better error messages: `shouldMatch` shows formatted JSON in case of errors.
```
fisx, line 25:
```suggestion
- do not import unit testing libraries such as `HUnit` or `hspec`. The equivalent functions they would provide are provided by the integration package and are better integrated, for example avoiding using `liftIO`, and providing better error messages: `shouldMatch` shows formatted JSON in case of errors.  (integration is built on top of tasty, but you don't need to know this to use it.)
```

integration/test/Testlib/App.hs
```diff
@@ -0,0 +1,935 @@
...
+        Just dir ->
+          (Just (dir </> srvName), "./dist" </> srvName)
+
+    (port, socket) <- maybe (failApp "the impossible in withServices happened") pure (Map.lookup srv ports)
+    liftIO $ N.close socket
+    (_, _, _, ph) <- liftIO $ createProcess (proc exe ["-c", tempFile]) {cwd = cwd}
+    pure ph
+
+  let stopInstances = liftIO $ do
+        -- when running from repl this hangs for 30 seconds
```
fisx, line 849: is this still true?  (the before of https://github.com/wireapp/wire-server/pull/3171/commits/022a4439f5bf63a4454b3579365862aa10d4e90e looks weird...)

charts/wire-server/requirements.yaml
smatting: Good question! It actually would! Good catch. Fixed in ddbc19771004a3770c7a4161f18a44abe6bec6b5

integration/test/Testlib/App.hs
smatting: No, fixed the comment in 60451d9

integration/test/Test/Brig.hs
smatting: Applied

integration/test/Testlib/App.hs
```diff
@@ -0,0 +1,935 @@
...
+getLastPrekey = App $ do
+  pks <- asks (.lastPrekeys)
+  lpk <- atomicModifyIORef pks getPK
+  pure $ object ["id" .= lastPrekeyId, "key" .= lpk]
+  where
+    getPK [] = error "Out of prekeys"
+    getPK (k : ks) = (ks, k)
+
+    lastPrekeyId :: Int
+    lastPrekeyId = 65535
```
fisx, line 112: these look a bit out of place here...

integration/test/Testlib/App.hs
```diff
@@ -0,0 +1,935 @@
...
+instance IsOption ConfigFile where
+  defaultValue = ConfigFile "services/integration.yaml"
+  parseValue = Just . ConfigFile
+  optionName = fromString "config"
+  optionHelp = fromString "Configuration file for integration tests. Default: services/integration.yaml"
+
+instance IsOption ServiceMap where
+  defaultValue = error "No Default value"
+  parseValue = const Nothing
+  optionName = fromString "config-loaded-from-file"
```
fisx, line 183: ?

integration/test/Testlib/App.hs
```diff
@@ -0,0 +1,935 @@
...
+    "pQABARn//wKhAFgglc6nCoZR2/qjLp0tr7vRyuXqb7ugdHHDadjX7zSl4uMDoQChAFgglacihnqg/YQJHkuHNFU7QD6Pb3KN4FnubaCF2EVOgRkE9g==",
+    "pQABARn//wKhAFgg5ER8h0/bIADXjBXe/XPKdzekgv6nhJ4hp3vJ3jtTSbUDoQChAFgglacihnqg/YQJHkuHNFU7QD6Pb3KN4FnubaCF2EVOgRkE9g==",
+    "pQABARn//wKhAFggsgV6jq+GuNuvXk+ctHh570cNqEmfPhz34wcYCMCf9xIDoQChAFgglacihnqg/YQJHkuHNFU7QD6Pb3KN4FnubaCF2EVOgRkE9g==",
+    "pQABARn//wKhAFggdQdlPqkBw6+phKhohp3YaWQL710euZDnyMLFwf2cS0oDoQChAFgglacihnqg/YQJHkuHNFU7QD6Pb3KN4FnubaCF2EVOgRkE9g==",
+    "pQABARn//wKhAFggKlsI/snuQMoYcZRw/kN+BobPV5gwYeBClp0Wx9btTGUDoQChAFgglacihnqg/YQJHkuHNFU7QD6Pb3KN4FnubaCF2EVOgRkE9g==",
+    "pQABARn//wKhAFggtruFBClEgdPKvjpHsYLlWMev9L4OmYZwlxbY0NwvzOwDoQChAFgglacihnqg/YQJHkuHNFU7QD6Pb3KN4FnubaCF2EVOgRkE9g==",
+    "pQABARn//wKhAFggRUdh4cuYtFNL46RLnPy65goYInyreStKwsEcY3pPlLkDoQChAFgglacihnqg/YQJHkuHNFU7QD6Pb3KN4FnubaCF2EVOgRkE9g==",
+    "pQABARn//wKhAFggQtT7lLZzH171F4jCbHNwxEAt28FwdQ8Kt2tbxFzPgC0DoQChAFgglacihnqg/YQJHkuHNFU7QD6Pb3KN4FnubaCF2EVOgRkE9g==",
+    "pQABARn//wKhAFggQeUPM119c+6zRsEupA8zshTfrZiLpXx1Ji0UMMumq9IDoQChAFgglacihnqg/YQJHkuHNFU7QD6Pb3KN4FnubaCF2EVOgRkE9g=="
+  ]
```
fisx, line 282: also looks out of place.

integration/test/Testlib/App.hs
```diff
@@ -0,0 +1,935 @@
...
+shouldMatchInt ::
+  (ProducesJSON a, HasCallStack) =>
+  -- | The actual value
+  a ->
+  -- | The expected value
+  Int ->
+  App ()
+a `shouldMatchInt` b = do
+  xa <- prodJSON a
+  shouldMatch xa b
```
fisx, line 349:
```suggestion
-- | Specialized variant of `shouldMatch` to avoid the need for type annotations.
shouldMatchInt ::
  (ProducesJSON a, HasCallStack) =>
  -- | The actual value
  a ->
  -- | The expected value
  Int ->
  App ()
shouldMatchInt = shouldMatch

```

seems to compile, and it makes sense to me...

integration/test/Testlib/App.hs
```diff
@@ -0,0 +1,935 @@
...
+-- compose operations that might fail without giving up nice error messages:
+--
+-- callMe (response.json %. "user" & "foo.bar.baz" %.= 2)
+--
+-- This can fail if
+-- 1. the response is not application/json
+-- 2. has no "user" field
+-- 3. the nested update fails
+class ProducesJSON a where
+  prodJSON :: HasCallStack => a -> App Value
```
fisx, line 425:
```suggestion
class MakeSON a where
  makeJSON :: HasCallStack => a -> App Value
```

or `mk`?  or `gen`?

Would be nice if method and class name would be consistent, and `produce` is a bit long for something so heavily used.

fisx (approved):

charts/wire-server/requirements.yaml
smatting: https://github.com/wireapp/wire-server/commit/ddbc19771004a3770c7a4161f18a44abe6bec6b5 caused 

```
Error from server (NotFound): pods "test-la9l8u3xfu93-wire-server-integration-integration" not found
```
(see [logs](https://concourse.ops.zinfra.io/builds/20254955#L6423bec2:426))

We need to double-check, but I think 
```
  annotations:
    "helm.sh/hook": test
```
annotation in the `Pod` definition wouldn't deploy this pod in prod.

integration/test/Testlib/App.hs
smatting: You mean because they are domain-specific? I think it's okay: They modify state in Env, so it makes sense they live in the module that defines the Env.

integration/test/Testlib/App.hs
smatting: Ah, that instance is a leftover. I remove it

integration/test/Testlib/App.hs
smatting: Same argument as for `getLastPrekey`. Could move into own module to have less visual noise, though

integration/test/Testlib/App.hs
smatting: Nice. Applied

integration/test/Testlib/App.hs
```diff
@@ -0,0 +1,935 @@
...
+      ob <- asObject v
+      pure $ Object $ KM.insert (KM.fromString k) newValue ob
+
+assertFailureWithJSON :: HasCallStack => ProducesJSON a => a -> String -> App b
+assertFailureWithJSON v msg = do
+  msg' <- ((msg <> "\n") <>) <$> prettyJSON v
+  assertFailure msg'
+
+-- | Useful for debugging
+printJSON :: ProducesJSON a => a -> App ()
```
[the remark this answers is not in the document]
smatting: This comes in handy when you want to print json values for debugging, e.g. printing json responses, etc. I think it's useful.

integration/test/Testlib/App.hs
smatting: Renamed it to `MakesValue` and `make`

integration/test/TestLib/App.hs
```diff
@@ -0,0 +1,930 @@
...
+data Env = Env
+  { context :: Context,
+    manager :: HTTP.Manager,
+    prekeys :: IORef [(Int, String)],
+    lastPrekeys :: IORef [String],
+    serviceConfigsDir :: FilePath,
+    servicesCwdBase :: Maybe FilePath
+  }
+
+data Context = Context
```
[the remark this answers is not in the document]
smatting: Done

PR #3185, opened 2023-03-24
use standard pager and more stable flake order

stefanwire:


stefanwire:
> Thanks! What's missing add pydoc to the python interpreter (in `wire-serer.nix`?)

It's a standard library and already installed. 🤔

PR #3198, opened 2023-03-29
fix: do not decrease login retry limit for 2fa code

battermann:
On a failed 2FA attempt, the retry limits for login decreased by the call to `loginFailedWith`. This has not effect if `setLimitFailedLogins` is not set or less strict. However, 2FA code verification has its own retry limits config and mechanism and the two should not be mixed.

## Checklist


mdimjasevic (commented):
Would it be possible to write tests for this?

battermann:
outdated

PR #3203, opened 2023-03-29
try if http2 would work

jschaul:
For experiments. Do not merge.

## Checklist

### DIFF
--- a/charts/nginz/templates/conf/_nginx.conf.tpl
+++ b/charts/nginz/templates/conf/_nginx.conf.tpl
@@ -186,7 +186,7 @@ http {
   #
 
   server {
-    listen {{ .Values.config.http.httpPort }};
+    listen {{ .Values.config.http.httpPort }}{{ if (.Values.config.http2.enabled) }} http2 {{end}};
     listen {{ .Values.config.ws.wsPort }}{{ if (.Values.config.ws.useProxyProtocol) }} proxy_protocol{{ end }};
 
     zauth_keystore {{ .Values.nginx_conf.zauth_keystore }};
--- a/charts/nginz/values.yaml
+++ b/charts/nginz/values.yaml
@@ -22,6 +22,8 @@ config:
   ws:
     wsPort: 8081
     useProxyProtocol: true
+  http2:
+    enabled: true
 terminationGracePeriodSeconds: 30
 nginx_conf:
   user: nginx


charts/nginz/templates/conf/_nginx.conf.tpl
```diff
@@ -186,7 +186,7 @@ http {
   #
 
   server {
-    listen {{ .Values.config.http.httpPort }};
+    listen {{ .Values.config.http.httpPort }}{{ if (.Values.config.http2.enabled) }} http2 {{end}};
```
flokli, line 189: Why should offering to connect via http2 be a problem? I'd guess it's more about clients, and making sure their http2 implementations are not broken.

charts/nginz/templates/deployment.yaml
```diff
@@ -62,14 +62,14 @@ spec:
...
-            scheme: HTTP
+            scheme: {{ if .Values.config.http2.enabled }}HTTP2{{ else }}HTTP{{ end }}
         livenessProbe:
           initialDelaySeconds: 30
           timeoutSeconds: 1
           httpGet:
             path: /status
             port: {{ .Values.config.http.httpPort }}
-            scheme: HTTP
+            scheme: {{ if .Values.config.http2.enabled }}HTTP2{{ else }}HTTP{{ end }}
```
flokli, line 72: http2 is not a scheme, that's either http or https (see https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.20/#httpgetaction-v1-core). It's up to the client to upgrade to http2 if it can.

In k8s, it probably already automatically upgrades if it can (see https://pkg.go.dev/golang.org/x/net/http2):

> Package http2 implements the HTTP/2 protocol.
> This package is low-level and intended to be used directly by very few people. Most users will use it indirectly through the automatic use by the net/http package (from Go 1.6 and later). For use in earlier Go versions see ConfigureServer. (Transport support requires Go 1.6 or later)

charts/nginz/templates/conf/_nginx.conf.tpl
jschaul: the ingress controller is in front of nginz; and I'm not sure if it talks http2 to upstreams or not. Hence this PR to try out what happens.

charts/nginz/templates/deployment.yaml
jschaul: Thanks! I saw that in some GCP thing where they explicitly put http2, but I should have checked the official docs.

PR #3209, opened 2023-03-30
cabal.project: drop with-compiler statement

flokli:
This is not used by CI, and whenever our ghc version string (coming from nixpkgs) changes, we need to manually update this.

Our `ghc` is coming from nixpkgs, which is pulled in by the dev env, there's no need to have another pin/restriction here.

## Checklist


supersven (approved):
LGTM :+1:

battermann (approved):

flokli:
Rebased to now drop ` with-compiler: ghc-9.2.7` rather than `with-compiler: ghc-9.2.4`, not waiting for integration tests.

PR #3213, opened 2023-04-03
[SQSERVICES-1619] Tech Debt Remove Non Binding Teams Code

battermann:
## Checklist

### DIFF
--- a/libs/wire-api/src/Wire/API/Routes/Public/Galley/Team.hs
+++ b/libs/wire-api/src/Wire/API/Routes/Public/Galley/Team.hs
@@ -26,6 +26,7 @@ import Wire.API.Error.Galley
 import Wire.API.Routes.MultiVerb
 import Wire.API.Routes.Named
 import Wire.API.Routes.Public
+import Wire.API.Routes.Version
 import Wire.API.Team
 import Wire.API.Team.Permission
 
@@ -33,7 +34,7 @@ type TeamAPI =
   Named
     "create-non-binding-team"
     ( Summary "Create a new non binding team"
-        -- FUTUREWORK: deprecated in https://github.com/wireapp/wire-server/pull/2607
+        :> Until 'V4
         :> ZUser
         :> ZConn
         :> CanThrow 'NotConnected
--- a/libs/wire-api/src/Wire/API/Team.hs
+++ b/libs/wire-api/src/Wire/API/Team.hs
@@ -107,15 +107,18 @@ newTeam tid uid nme ico tb = Team tid uid nme ico Nothing tb DefaultIcon
 
 instance ToSchema Team where
   schema =
-    object "Team" $
+    objectWithDocModifier "Team" desc $
       Team
         <$> _teamId .= field "id" schema
         <*> _teamCreator .= field "creator" schema
         <*> _teamName .= field "name" schema
         <*> _teamIcon .= field "icon" schema
         <*> _teamIconKey .= maybe_ (optField "icon_key" schema)
-        <*> _teamBinding .= (fromMaybe Binding <$> optField "binding" schema)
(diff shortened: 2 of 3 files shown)


changelog.d/1-api-changes/pr-3213
```diff
@@ -0,0 +1 @@
+Non-binding team endpoints are removed from API version V4
```
fisx, line 1:
```suggestion
Remove deprecated non-binding team endpoints from API version V4
```

libs/wire-api/src/Wire/API/Routes/Public/Galley/Team.hs
```diff
@@ -26,14 +26,15 @@ import Wire.API.Error.Galley
...
+import Wire.API.Routes.Version
 import Wire.API.Team
 import Wire.API.Team.Permission
 
 type TeamAPI =
   Named
     "create-non-binding-team"
     ( Summary "Create a new non binding team"
-        -- FUTUREWORK: deprecated in https://github.com/wireapp/wire-server/pull/2607
+        :> Until 'V4
```
fisx, line 37: maybe still link the PR and/or ticket?  not sure, maybe it's unnecessary noise.

libs/wire-api/src/Wire/API/Team.hs
```diff
@@ -107,15 +107,18 @@ newTeam tid uid nme ico tb = Team tid uid nme ico Nothing tb DefaultIcon
...
         <*> _teamCreator .= field "creator" schema
         <*> _teamName .= field "name" schema
         <*> _teamIcon .= field "icon" schema
         <*> _teamIconKey .= maybe_ (optField "icon_key" schema)
-        <*> _teamBinding .= (fromMaybe Binding <$> optField "binding" schema)
+        <*> _teamBinding .= (fromMaybe Binding <$> optFieldWithDocModifier "binding" bindingDesc schema)
         <*> _teamSplashScreen .= (fromMaybe DefaultIcon <$> optField "splash_screen" schema)
+    where
+      desc = description ?~ "`binding` is deprecated, and should be ignored. The non-binding teams API is not used (and will not be supported from API version V4 onwards), and `binding` should always be `true`."
+      bindingDesc = description ?~ "Deprecated, please ignore."
```
fisx, line 121:
```suggestion
      desc = description ?~ "`binding` is deprecated, and should be ignored. The non-binding teams API is not used (and will not be supported from API version V4 onwards), and `binding` will always be `true`."
      bindingDesc = description ?~ "Deprecated, please ignore."
```

why the `desc`, and not just the `bindingDesc` with all the information near what it's about?

fisx (approved):

fisx:
Hum, I think you also want to deprecate https://github.com/wireapp/wire-server/blob/2400e8bcd54b1edde4f8824daee71d55d140dd31/libs/wire-api/src/Wire/API/Routes/Public/Galley.hs#L955

If clients use it, it should be replaced by "get-team" (path "/team"); if nobody uses it, it should go away.  (Feel free to overrule me, but if you do please document the history in swagger for future me.  :)

changelog.d/1-api-changes/pr-3213
battermann: Since this appears in the changelog, it is already in the past. So I prefer past tense or even better "have been removed" (present perfect progressive). But I will not hold my opinion very strongly. :)

libs/wire-api/src/Wire/API/Routes/Public/Galley/Team.hs
battermann: I prefer less noise.

libs/wire-api/src/Wire/API/Team.hs
battermann: Because it is easier to see. To see the field description, you need to scroll down and click on `Model` in the swagger UI.

PR #3218, opened 2023-04-05
Coturn helm chart: Increase liveness timeout

jschaul:
Increase default of liveness/readiness probe and make it configurable.

Under high load, the default of failureThreshold=3 timeoutSeconds=1 can lead to restarts of the coturn pod due to the http port being temporarily starved of CPU, leading to an unnecessary restart of the coturn pods. This change should make this less frequent and improve call stability.

## Checklist

### DIFF
--- a/changelog.d/3-bug-fixes/coturn-liveness
+++ b/changelog.d/3-bug-fixes/coturn-liveness
@@ -1,0 +1,1 @@
+Coturn helm chart: Increase the default timeout of liveness/readiness probe and make it configurable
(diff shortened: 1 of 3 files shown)


supersven:
@jschaul do you know if a not quickly answering `coturn` may cause any kind of other troubles?

jschaul:
> @jschaul do you know if a not quickly answering `coturn` may cause any kind of other troubles? I mean: What happens on the client's side in this case? thinking

As far as I'm aware, clients attempt to contact multiple TURN servers at once, and go with the server that answers first. So that shouldn't be an issue. 

Also, the metrics/http endpoint is different to the udp/tcp/tls endpoints used for actual TURN traffic. So metrics being slow doesn't need to mean actual turn connections are slow, too.

PR #3220, opened 2023-04-05
coturn helm chart: use a memory-backed mount

jschaul:
Add memory-backed mount /var/lib/coturn to store sqllite DB to improve performance, as described on https://github.com/wireapp/coturn/tree/master/docker/coturn#persistence

## Checklist

### DIFF
--- a/changelog.d/3-bug-fixes/coturn-interal-db
+++ b/changelog.d/3-bug-fixes/coturn-interal-db
@@ -1,0 +1,1 @@
+coturn helm chart: use a memory-backed folder to store sqllite DB to improve performance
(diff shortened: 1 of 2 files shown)


flokli:
We are certain we don't need any of that state?

supersven (approved):
@jschaul Well, I give you a conditional approval: If it doesn't turn out to be beneficial (no pun intended :wink: ), please roll the change back.

supersven:
> We are certain we don't need any of that state?

@flokli , @jschaul , AFAIK the state was previously saved in the container itself. That's not very persistent, too... :wink: However, maybe, @jschaul can try to understand what the database's data is about and if it would be beneficial to keep it in a persistent db?

PR #3222, opened 2023-04-06
E2E Test for generating DPoP access token

battermann:
## Checklist


services/brig/test/integration/API/User/Client.hs
```diff
@@ -121,7 +128,7 @@ tests _cl _at opts p db b c g =
       test p "get/head nonce/clients" $ testNewNonce b,
       testGroup
         "post /clients/:cid/access-token"
-        [ test p "invalid values" $ testCreateAccessTokenInvalidValues b,
```
fisx, line 124: why is this test removed?

fisx (approved):
LGTM!  I have skimmed over the test and it looks plausible.  Since errors in the test are less harmful and more likely to show given prod code hasn't changed much, I think this is enough for an "approve".

services/brig/test/integration/API/User/Client.hs
battermann: It is really the most trivial test and doesn't prove much value.

battermann:
obsolete because of #3255

PR #3228, opened 2023-04-12
avoid overlapping versions

stefanwire:
## Checklist


mdimjasevic (commented):
This seems benign, but now that you made it obvious that there are overlapping endpoints, I wonder how is the endpoint and the handler chosen if I make a request to `POST /v2/conversations`?

mdimjasevic (approved):
I think for endpoints `a` and `b` that are combined in an API with `a :<|> b`, I think `a` will be attempted first (to answer my earlier question).

PR #3235, opened 2023-04-19
Dynamic federation remotes

fisx:
https://wearezeta.atlassian.net/browse/FS-1115

store federation remotes in cassandra in brig, in the same form as the remote search policies in the config file.

the idea is that this will become the source of truth for all services which nodes we federate with, and that it can be dynamically updated (via internal end-points for now, and possibly by events like "node down for too long" in the future).

it's not ideal that we store this in brig, but federator doesn't have a cassandra instance.

**[not on this PR yet]** we're also working on a TVar maintained by all federators and other services that are interested (including brig) that keeps a cache of the cassandra table, and a rabbitMQ broadcast queue that propagates updates to the brig pod siblings that haven't received it from the operator, and to the other services.  the federator config will be simplified to not contain a list of federating nodes any more; federators will also get them from brig.

**migration:** my current idea is to always consider the union of brig's config file and the cassandra table the global truth.  this way the operator can safely remove the list from the config file after the cassandra table has been updated (next deployment or so).  the downside of this is that removing edges that are in the config file will silently fail (we could make it fail loudly, of course).

## Checklist

### DIFF
--- a/services/brig/src/Brig/API/Internal.hs
+++ b/services/brig/src/Brig/API/Internal.hs
@@ -119,6 +119,7 @@ servantSitemap =
     :<|> userAPI
     :<|> authAPI
     :<|> internalOauthAPI
+    :<|> federationRemotesAPI
 
 ejpdAPI ::
   Member GalleyProvider r =>
@@ -170,6 +171,12 @@ authAPI =
     :<|> Named @"sso-login" (callsFed (exposeAnnotations ssoLogin))
     :<|> Named @"login-code" getLoginCode
     :<|> Named @"reauthenticate" reauthenticate
+
+federationRemotesAPI :: ServerT BrigIRoutes.FederationRemotesAPI (Handler r)
+federationRemotesAPI =
+  Named @"get-federator-remotes" undefined
+    :<|> Named @"add-federator-remote" undefined
+    :<|> Named @"delete-federator-remote" undefined
 
 -- | Responds with 'Nothing' if field is NULL in existing user or user does not exist.
 getAccountConferenceCallingConfig :: UserId -> (Handler r) (ApiFt.WithStatusNoLock ApiFt.ConferenceCallingConfig)
--- a/services/brig/src/Brig/API/Internal.hs
+++ b/services/brig/src/Brig/API/Internal.hs
@@ -38,6 +38,7 @@ import qualified Brig.Code as Code
 import Brig.Data.Activation
 import qualified Brig.Data.Client as Data
 import qualified Brig.Data.Connection as Data
+import qualified Brig.Data.Federation as Data
 import qualified Brig.Data.MLS.KeyPackage as Data
 import qualified Brig.Data.User as Data
 import Brig.Effects.BlacklistPhonePrefixStore (BlacklistPhonePrefixStore)
@@ -89,6 +90,7 @@ import Wire.API.Federation.API
 import Wire.API.MLS.Credential
 import Wire.API.MLS.KeyPackage
 import Wire.API.MLS.Serialisation
+import Wire.API.Routes.FederationDomainConfig
 import Wire.API.Routes.Internal.Brig
 import qualified Wire.API.Routes.Internal.Brig as BrigIRoutes
 import Wire.API.Routes.Internal.Brig.Connection
@@ -174,9 +176,9 @@ authAPI =
 
 federationRemotesAPI :: ServerT BrigIRoutes.FederationRemotesAPI (Handler r)
 federationRemotesAPI =
-  Named @"get-federation-remotes" undefined
-    :<|> Named @"add-federation-remotes" undefined
-    :<|> Named @"delete-federation-remotes" undefined
+  Named @"get-federation-remotes" (lift $ FederationDomainConfigs <$> wrapClient Data.getFederationRemotes) -- TODO: get this from TVar!  also merge in config file!
+    :<|> Named @"add-federation-remotes" (lift . wrapClient . Data.addFederationRemote)
+    :<|> Named @"delete-federation-remotes" (lift . wrapClient . Data.deleteFederationRemote)
 
 -- | Responds with 'Nothing' if field is NULL in existing user or user does not exist.
 getAccountConferenceCallingConfig :: UserId -> (Handler r) (ApiFt.WithStatusNoLock ApiFt.ConferenceCallingConfig)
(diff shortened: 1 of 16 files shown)


services/brig/src/Brig/API/Internal.hs
```diff
@@ -38,6 +38,7 @@ import qualified Brig.Code as Code
 import Brig.Data.Activation
 import qualified Brig.Data.Client as Data
 import qualified Brig.Data.Connection as Data
+import qualified Brig.Data.Federation as Data
```
akshaymankar, line 41: Did you forget to commit this file?

services/brig/src/Brig/API/Internal.hs
```diff
@@ -171,6 +174,12 @@ authAPI =
     :<|> Named @"login-code" getLoginCode
     :<|> Named @"reauthenticate" reauthenticate
 
+federationRemotesAPI :: ServerT BrigIRoutes.FederationRemotesAPI (Handler r)
+federationRemotesAPI =
+  Named @"get-federation-remotes" (lift $ FederationDomainConfigs <$> wrapClient Data.getFederationRemotes) -- TODO: get this from TVar!  also merge in config file!
```
akshaymankar, line 179: Don't see the code for this, but perhaps this would require a full-table scan, which Cassandra doesn't like. If we have to list all the domain, perhaps we should store them differently.

akshaymankar:
> **migration:** my current idea is to always consider the union of brig's config file and the cassandra table the global truth. this way the operator can safely remove the list from the config file after the cassandra table has been updated (next deployment or so). the downside of this is that removing edges that are in the config file will silently fail (we could make it fail loudly, of course).

Perhaps we don't need to worry about this as federation is not very seriously used by anyone, so causing a bit a breakage while someone makes an API call to load the list in the DB is acceptable? If so we can get rid of this whole thing.

akshaymankar:
> **[not on this PR yet]** we're also working on a TVar maintained by all federators and other services that are interested (including brig) that keeps a cache of the cassandra table, and a rabbitMQ broadcast queue that propagates updates to the brig pod siblings that haven't received it from the operator, and to the other services. the federator config will be simplified to not contain a list of federating nodes any more; federators will also get them from brig.

I am assuming we're looking at storing this in `TVar` and using `RabbitMQ` to propagate changes in order to reduce latency which goes with checking whether or not a domain is allowed to be federated with. Do you have  ballpark estimate of how much we're saving by doing this? To me this sounds like a lot of upfront optimization, or perhaps I am missing something.

services/brig/src/Brig/API/Internal.hs
fisx: yes.  thanks!

fisx:
> > **[not on this PR yet]** we're also working on a TVar maintained by all federators and other services that are interested (including brig) that keeps a cache of the cassandra table, and a rabbitMQ broadcast queue that propagates updates to the brig pod siblings that haven't received it from the operator, and to the other services. the federator config will be simplified to not contain a list of federating nodes any more; federators will also get them from brig.
> 
> I am assuming we're looking at storing this in `TVar` and using `RabbitMQ` to propagate changes in order to reduce latency which goes with checking whether or not a domain is allowed to be federated with. Do you have ballpark estimate of how much we're saving by doing this? To me this sounds like a lot of upfront optimization, or perhaps I am missing something.

we could make everybody poll once every second, but i have a feeling that using a queue isn't much harder, just a lot faster.

come to think of it, we probably need to make sure that we don't have a (unlikely) race condition where pod A writes to cassandra, then pod B writes to cassandra, then pod B notifies everybody of the update, then pod A does.  hum.  would that be resolved by polling?  i guess so.

akshaymankar:
> we could make everybody poll once every second, but i have a feeling that using a queue isn't much harder, just a lot faster.

I meant what is the point of the `TVar`, everybody who needs to know can just make an internal call to brig? How slow is that and is it _too slow_? As in do you have an SLO in mind?

services/brig/src/Brig/API/Internal.hs
akshaymankar: An alternative is to store the list in 1 row with `id = 1`. I think we already do this somewhere.
Then have a separate table to remember all the configuration for each endpoint. This also has the benefit of not forgetting preferences for a remote even if you stop federating with them.

lepsa:
> 

From my talks with Matthias, using a message queue simplifies propagating the domain update notice to all services that need it. Since the various services can only talk within their node makes coordination hard. Having a system sitting outside of these VMs, like the databases, that can fan out messages would be greatly helpful. The K8S pods wouldn't have to know how many other pods are running, or where they are running, all updates for federation domains can be passed through an MQ broker that handles fanout for us.

services/brig/src/Brig/API/Internal.hs
fisx: https://nblair.github.io/2017/02/16/cassandra-full-table-scan/ says it's fine to pull 10k rows in one query.  so wouldn't c553c0efc7b8ad1282af689abfaf6aa88299e5ff be a good solution?

fisx:
> > 
> 
> From my talks with Matthias, using a message queue simplifies propagating the domain update notice to all services that need it. Since the various services can only talk within their node makes coordination hard. Having a system sitting outside of these VMs, like the databases, that can fan out messages would be greatly helpful. The K8S pods wouldn't have to know how many other pods are running, or where they are running, all updates for federation domains can be passed through an MQ broker that handles fanout for us.

I think @akshaymankar you are suggesting that we do a rest api request to brig and table lookup every time the current code does an MVar lookup in federator or brig.  Yes, that could be done, but wouldn't that add a few (10s? 100s?) milliseconds to every remote request or notification?  I'm not aware if we have specified acceptable lag limits, but i feel it's easier to get this right to begin, doing lots of api requests isn't that much easier.

Instead of using a rabbit-queue we could also make every pod poll the table via rest once a second or so, for caching in a TVar.  This is certainly slower during changes of the connection graph, but that's a rare operation and we need to allow for race conditions there anyway.

I still like the fastest option best.  @akshaymankar do you really think it'll be expensive to implement?  Maybe discuss on the phone monday?

akshaymankar:
> I still like the fastest option best. @akshaymankar do you really think it'll be expensive to implement? Maybe discuss on the phone monday?

As I replied in JIRA, I guess y'all already have a decision, let me know if you still want to have a call.

fisx:
closing in favor of https://github.com/wireapp/wire-server/pull/3260 (which contains this branch in full).

PR #3236, opened 2023-04-19
[FS-1845] Deploy RabbitMQ on k8s for CI

akshaymankar:
The scope changed from the discussions. The work will now be split into two PRs, this one deals with CI infrastructure only, the next one incorporates it into Brig.


services/backend-notification-pusher/src/Wire/BackendNotificationPusher/Options.hs
```diff
@@ -0,0 +1,24 @@
...
+
+import Data.Aeson
+import Data.Domain
+import Imports
+import Util.Options
+
+data Opts = Opts
+  { federatorInternal :: !Endpoint,
+    rabbitMQ :: !RabbitMQOpts,
+    remoteBackends :: [Domain]
```
fisx, line 11: this would need to be handled the same way we do things in #3235, but until we've decided that this is fine i suppose.

fisx:
I'm a bit skeptical about making this a separate service.  There is a lot of overlap between the code in federator and in notification-pusher (the requests that are currently going to federator will be pushed to the queue in some form by federator, and then picked up by notification-pusher, is that the plan?  So the forwarding is currently done by federator, and will be done by notification-pusher in the future?).  And I don't think scalability will become a problem if we do everything in one scalable service rather than several.

I am all for separating concerns, but I would prefer doing that using module boundaries rather than microservices.

services/backend-notification-pusher/src/Wire/BackendNotificationPusher/Options.hs
akshaymankar: Yes, that's the plan. If you finish #3235 before this PR, we get to implement the change here (as I guess some test would break otherwise). If we finish this PR before you, then you get to implement it in yours. So its a race :rofl:

akshaymankar:
> I'm a bit skeptical about making this a separate service. There is a lot of overlap between the code in federator and in notification-pusher (the requests that are currently going to federator will be pushed to the queue in some form by federator, and then picked up by notification-pusher, is that the plan? So the forwarding is currently done by federator, and will be done by notification-pusher in the future?). And I don't think scalability will become a problem if we do everything in one scalable service rather than several.
> 
> I am all for separating concerns, but I would prefer doing that using module boundaries rather than microservices.

The idea is that this service will make calls to federator, which will do its thing as usual. The new service will not duplicate code in federator. The idea of why this service should be separate is simple: This is not an HTTP service, it is a background worker. There was a plan to make federator's internal service into haskell library and make federated calls directly from brig, galley and cargohold which follows the reasoning of not making another HTTP calls instead of a function call, but I am not sure why it was not followed. Even if we do that plan, I would say this new service should be running as a separate process as it not being an http server is significant enough change to warrant different scaling and process management.

cabal.project
```diff
@@ -31,6 +31,7 @@ packages:
   , libs/wire-message-proto-lens/
   , libs/zauth/
   , services/brig/
+  , services/backend-notification-pusher/
```
fisx, line 34: can we still rename this to `worker`, or something similarly general?  [discussion here](https://wearezeta.atlassian.net/wiki/pages/diffpagesbyversion.action?pageId=384928204&selectedPageVersions=71&selectedPageVersions=72)

mdimjasevic (approved):

PR #3251, opened 2023-04-26
play with new integration package

fisx:
WIP

PR #3252, opened 2023-04-26
Introduce SearchContacts permission.

fisx:
https://wearezeta.atlassian.net/browse/FS-1880

integration test is under way [elsewhere](https://github.com/wireapp/wire-server/pull/3251) for technical reasons.

## Checklist

### DIFF
--- a/services/brig/src/Brig/Team/Util.hs
+++ b/services/brig/src/Brig/Team/Util.hs
@@ -19,8 +19,10 @@ module Brig.Team.Util where -- TODO: remove this module and move contents to Bri
 
 import Brig.API.Error
 import Brig.App
+import qualified Brig.Data.User as Data
 import Brig.Effects.GalleyProvider (GalleyProvider)
 import qualified Brig.Effects.GalleyProvider as GalleyProvider
+import Brig.Types.User (HavePendingInvitations (NoPendingInvitations))
 import Control.Error
 import Control.Lens
 import Data.Id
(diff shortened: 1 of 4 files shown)


services/brig/src/Brig/Team/Util.hs
```diff
@@ -30,8 +32,17 @@ import Imports
...
+import Wire.API.User (User (userTeam))
 
-ensurePermissions :: Member GalleyProvider r => UserId -> TeamId -> [Perm] -> ExceptT Error (AppT r) ()
+-- | If the user is in a team, it has to have these permissions.  If not, it is a personal
+-- user with account validation and thus given the permission implicitly.  (Used for
+-- `SearchContactcs`.)
+ensurePermissionsOrPersonalUser :: (Member GalleyProvider r, IsPerm perm) => UserId -> [perm] -> ExceptT Error (AppT r) ()
+ensurePermissionsOrPersonalUser u perms = do
+  mbUser <- lift $ wrapHttp $ Data.lookupUser NoPendingInvitations u
+  maybe (pure ()) (\tid -> ensurePermissions u tid perms) (userTeam =<< mbUser :: Maybe TeamId)
```
smatting, line 43: Nit-pick, maybe something more readable, e.g.

```suggestion
  mbUser <- lift $ wrapHttp $ Data.lookupUser NoPendingInvitations u
  for_ mbUser $ \u ->
     ensurePermissions u (userTeam u) perms
```
(untested)

services/brig/src/Brig/User/API/Search.hs
```diff
@@ -92,6 +93,7 @@ search ::
   Maybe (Range 1 500 Int32) ->
   (Handler r) (Public.SearchResult Public.Contact)
 search searcherId searchTerm maybeDomain maybeMaxResults = do
+  ensurePermissionsOrPersonalUser searcherId [SearchContacts]
```
smatting, line 96: This could be moved into `searchLocally`, which already looks up the team id. Federated searching has its own policies, maybe this check isn't relevant there.

smatting (approved):
integration test for this is in https://github.com/wireapp/wire-server/pull/3251

services/brig/src/Brig/User/API/Search.hs
fisx: i agree that's a possible optimization, but i think we should also disallow remote searches for ephemeral users.  i'll leave a comment in the code, ok?

PR #3253, opened 2023-04-26
integration-setup: Pass --skip-deps to helmfile sync

akshaymankar:
We run `helm dep update` right before running the helmfile command, so this shouldn't be required.
This saves 1.5 mins in CI.

## Checklist

 - [x] ~Add a new entry in an appropriate subdirectory of `changelog.d`! No changelog


jschaul (approved):
let's try, and undo if we find issues with this.

PR #3256, opened 2023-04-27
Grepinclude sphinx directive

jschaul:
Some python to include pieces of code inside docs.wire.com that auto-updates and in somewhat resilient to minor moving lines up and down.

Also document TLS ciphers used for coturn.

coturn-docs

## Checklist

### DIFF
--- a/docs/src/grepinclude.py
+++ b/docs/src/grepinclude.py
@@ -1,0 +1,56 @@
+from docutils import nodes
+from sphinx.util.docutils import SphinxDirective
+from sphinx.directives.code import container_wrapper
+import re
+
+# A bit like 'literalinclude', but using a keyword to highlight some code so as
+# to be a little more resilient to code moving lines with time.
+# Usage:
+# ```{eval-rst}
+# .. grepinclude:: ../charts/coturn/values.yaml ciphers:
+#    :lines-before: 3
+#    :lines-after: 0
+#    :language: yaml
+# ```
+class GrepInclude(SphinxDirective):
+    required_arguments = 2
+    optional_arguments = 0
+    final_argument_whitespace = True
+    option_spec = {
+        'lines-before': int,
+        'lines-after': int,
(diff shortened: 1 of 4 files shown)


arthurwolf (approved):
looks great

docs/src/how-to/install/tls.md
```diff
@@ -40,6 +40,17 @@ The list of TLS ciphers for incoming SFT requests (and metrics) are defined in a
...
+## Coturn (kubernetes)
+
+The list of TLS ciphers for "TLS over TCP" TURN are defined in the [coturn helm chart](https://github.com/wireapp/wire-server/blob/master/charts/coturn/)
+
+```{eval-rst}
+.. grepinclude:: ../charts/coturn/values.yaml ciphers:
+   :lines-before: 3
+   :lines-after: 0
+   :language: yaml
+```
```
smatting, line 52: Doe we need the `eval-rst`?

The equivalent markdown should look something like

```

 ```{grepinclude} chiphers
---
lines-before: 3
lines-after: 0
language: yaml
---
...
```

jschaul: makes sense. Adjusted code to match that.

docs/src/grepinclude.py
```diff
@@ -0,0 +1,66 @@
...
+                None, nodes.paragraph(text="Unable to open file: %s" % file_path)
+            )]
+
+        filtered_lines = []
+        for i, line in enumerate(content):
+            if re.search(keyword, line):
+                start = max(0, i - lines_before)
+                end = min(len(content), i + 1 + lines_after)
+                filtered_lines.extend(content[start:end])
+
```
comawill, line 55: I really like this concept 👍 

What do you think about returning an error, `if len(filtered_lines) == 0`?
So this could prevent silently placing empty blocks, if the search string no longer matches.

PR #3260, opened 2023-05-01
Dynamic federator remotes

fisx:
https://wearezeta.atlassian.net/browse/FS-1115

### internal docs: 

http://localhost:8082/api-internal/swagger-ui/brig/

2023-05-19-113146_1920x1080_scrot

### docs for devs:

2023-06-04-143744_1600x900_scrot

### (slightly outdated) docs for operators:

2023-06-04-143310_1600x900_scrot
2023-06-04-143317_1600x900_scrot
2023-06-04-143320_1600x900_scrot
2023-06-04-143325_1600x900_scrot
2023-06-04-143335_1600x900_scrot

## Checklist

## TODO

- [ ] double-check what our default search policy in (a) the case where we allow all domains and a new one is not listed, and (b) the case where we have an allow entry in the cassandra table, but it has no search policy.  also check if we're getting it right in the docs.  (i think (a) should be no_search and (b) should be no NULL allowed in database field, just pick something, it's not that hard.)  maybe write integration tests?
- [x] log 'F' are sometimes too harsh.  'I' or 'D' would do.  eg., for "couldn't pull data from brig *yet*, retrying".
- [ ] yeah, galley integration tests where hanging for me as well, but the last 3 rounds everything went green ok.
- [ ] https://github.com/wireapp/wire-server/pull/3260#pullrequestreview-1487587205


services/brig/src/Brig/API/Internal.hs
```diff
@@ -171,6 +176,34 @@ authAPI =
     :<|> Named @"login-code" getLoginCode
     :<|> Named @"reauthenticate" reauthenticate
 
+federationRemotesAPI :: ServerT BrigIRoutes.FederationRemotesAPI (Handler r)
+federationRemotesAPI =
+  Named @"get-federation-remotes" getFederationRemotes -- TODO: get this from TVar!
```
fisx, line 181: this TODO is important!

services/brig/test/integration/API/Federation.hs
```diff
@@ -457,3 +459,33 @@ testClaimKeyPackagesMLSDisabled opts brig = do
...
+  resetFederationRemotes opts brig
+
+  res1 <- getFederationRemotes brig
+  liftIO $ assertEqual "should return config values" cfgRemotes res1
+
+  let remote1 = FederationDomainConfig (Domain "good.example.com") NoSearch
+  addFederationRemote brig remote1
+  res2 <- getFederationRemotes brig
+  liftIO $ assertEqual "should return config values and good.example.com" (nub $ sort $ cfgRemotes <> [remote1]) (sort res2)
+
```
fisx, line 477: maybe this?

```suggestion
  let remote1 = FederationDomainConfig (Domain "good.example.com") NoSearch
  addFederationRemote brig remote1
  res2 <- getFederationRemotes brig
  liftIO $ assertEqual "should return config values and good.example.com" (nub $ sort $ cfgRemotes <> [remote1]) (sort res2)

  -- idempotency
  addFederationRemote brig remote1
  res2' <- getFederationRemotes brig
  liftIO $ assertEqual "should return config values and good.example.com" (nub $ sort $ cfgRemotes <> [remote1]) (sort res2')
  
```

(i know i wrote this test, just thought of it late.)

fisx (commented):
read about half, looks good, but of course i have comments.  will keep reading later!

services/federator/src/Federator/Options.hs
```diff
@@ -92,7 +92,9 @@ data Opts = Opts
     -- | Logformat to use
     logFormat :: !(Maybe (Last LogFormat)),
     -- | Runtime settings
-    optSettings :: !RunSettings
+    optSettings :: !RunSettings,
+    -- | Domain update interval (microseconds)
+    domainUpdateInterval :: !Int
```
fisx, line 97: only make this configurable in brig, and serve the parameter together with the remotes in the rest api?

services/federator/src/Federator/Run.hs
```diff
@@ -84,6 +114,17 @@ run opts = do
...
         (_, _) -> conf
 
+    getFedRemotes = namedClient @IAPI.API @"get-federation-remotes"
+
+    updateDomains :: ClientEnv -> (FederationDomainConfigs -> IO ()) -> IO ()
+    updateDomains clientEnv update = forever $ do
+      threadDelay $ domainUpdateInterval opts
+      strat <- runClientM getFedRemotes clientEnv
+      either
+        print
```
fisx, line 124: no print, please use logger everywhere!

fisx:
@pcapriotti says we should use `IORef` instead of `TVar`.  he might be right?

fisx:
we have another problem.  looking at federator: since we're storing the entire `Env` in a `TVar`, i'm not sure what happens to the `RequestID`.  do we somehow share it between requests now?  if not, how does that work?

fisx:
https://github.com/wireapp/wire-server/pull/3260/commits/3b8b3837f03f2f8f11b71b306b3de3d5b0aa7d18 (even though you should probably look at

```
$ git diff `git merge-base origin/develop HEAD` services/federator/src/
```

instead, it cleans up a little mess I've made yesterday).

I think I now have federator in a shape that I like.  It solves the problem of sharing requestIDs between threads (which I introduced myself yesterday) by only keeping an IORef of the AllowedDomains that we actually want to change, not IORef-ing the entire Env.  It looks a little tidier and more understandable.  It uses `IORef`.  It keeps the diff to develop minimal.

Missing: unit tests, probably integration tests; make FederationStrategy syntax sane (eg., require operators to copy the strategy to a new syntax in a different place in the yaml tree, keep the old until transition period is over).

libs/types-common/src/Data/MessageQueue.hs
```diff
@@ -0,0 +1,21 @@
...
+where
+
+import Data.Aeson (FromJSON)
+import Data.Text (Text)
+import GHC.Generics (Generic)
+import Prelude (Show, String)
+
+-- | Options for connecting to the message queue system
+data MessageQueueSettings = MessageQueueSettings
+  { mqHost :: String,
```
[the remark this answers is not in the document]
lepsa: I'm happy to do that.

services/brig/brig.integration.yaml
```diff
@@ -209,3 +209,9 @@ optSettings:
 
 logLevel: Warn
 logNetStrings: false
+mqSettings:
```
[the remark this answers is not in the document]
lepsa: I haven't kept up with that PR, but making this config optional certainly makes sense.

services/brig/src/Brig/API/Internal.hs
lepsa: Shouldn't getting the federation remotes be reading from the DB directly in this case? Reading from a TVar/IORef will be introducing another polling delay for services that depend on this API to get the federation domain state.

services/brig/src/Brig/API/Internal.hs
```diff
@@ -171,6 +176,42 @@ authAPI =
...
+    Data.AddFederationRemoteMaxRemotesReached ->
+      throwError . fedError . FederationUnexpectedError $
+        "Maximum number of remote backends reached.  If you need to create more connections, \
+        \please contact wire.com."
+
+getFederationRemotes :: ExceptT Brig.API.Error.Error (AppT r) FederationDomainConfigs
+getFederationRemotes = lift $ do
+  db <- wrapClient Data.getFederationRemotes
+  cfg <- asks (fromMaybe [] . setFederationDomainConfigs . view settings)
+  -- FUTUREWORK: we should solely rely on `db` in the future; `cfg` is just for an easier,
```
lepsa, line 199: @fisx This note ties into my comment above. I don't think that a tvar/ioref is needed when calling the GET API, as a DB call for the latest info would be ideal for other services to be up to date.

services/cannon/src/Cannon/Run.hs
```diff
@@ -75,6 +79,27 @@ run o = do
...
+          Right strat -> pure strat
+          Left err -> do
+            print $ "Could not retrieve the latest list of federation domains from Brig: " <> show err
+            threadDelay $ o ^. domainUpdateInterval
+            getInitialFedDomains
+  fedStrat <- getInitialFedDomains
+  tEnv <- newTVarIO fedStrat
+  let callback :: FederationDomainConfigs -> IO ()
+      callback = atomically . writeTVar tEnv
+  updateDomainsThread <- Async.async $ updateDomains clientEnv callback
```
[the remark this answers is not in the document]
lepsa: The TVar will need to be outside of the update loop so it can eventually be used by other code that I haven't updated yet. Otherwise it will be unreachable by the rest of the application.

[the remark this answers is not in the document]
lepsa: Callback was a bit more complicated once-upon-a-time, but got simplified to its current form. Inlining it certainly a viable option.

services/brig/src/Brig/API/Internal.hs
fisx: 167083a63d376859cfd67e3b076d88a4ed90f3db  :+1:

services/brig/src/Brig/API/Internal.hs
fisx: no, this is about adding the old remotes considered in the config file to the response for a transition period, until the operators have had a chance to load the cassandra table.

i think this should stay.  if you still disagree let's call.

services/cannon/src/Cannon/Run.hs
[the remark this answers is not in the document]
fisx: hm, might be, but this code is outdated anyway and wouldn't fit the tip of the branch any more.  i'll close this comment.

services/cannon/src/Cannon/Run.hs
```diff
@@ -85,20 +83,9 @@ run o = do
...
-            print $ "Could not retrieve the latest list of federation domains from Brig: " <> show err
-            threadDelay $ o ^. domainUpdateInterval
-            getInitialFedDomains
-  fedStrat <- getInitialFedDomains
-  tEnv <- newTVarIO fedStrat
-  let callback :: FederationDomainConfigs -> IO ()
-      callback = atomically . writeTVar tEnv
-  updateDomainsThread <- Async.async $ updateDomains clientEnv callback
+  fedStrat <- getAllowedDomainsInitial clientEnv
+  ioref <- newIORef fedStrat
```
fisx, line 87:
```
  fedStrat <- newIORef =<< getAllowedDomainsInitial clientEnv
```

(`ioref` is not a very helpful name.)

fisx: also, who's using `ioref`?  adding it to `Env` and watching for updates is future work, correct?  would that also happen in `updateDomainsThread`, right after `getAllowedDomainsLoop`?

services/federator/src/Federator/Env.hs
```diff
@@ -35,18 +33,15 @@ import OpenSSL.Session (SSLContext)
...
+import Wire.API.Routes.FederationDomainConfig (FederationDomainConfigs)
 
 data Env = Env
   { _metrics :: Metrics,
     _applog :: LC.Logger,
     _requestId :: RequestId,
     _dnsResolver :: Resolver,
     _runSettings :: RunSettings,
-    _allowedRemoteDomains :: IORef AllowedDomains,
+    _allowedRemoteDomains :: IORef FederationDomainConfigs,
```
fisx, line 44: :+1:

services/galley/src/Galley/Run.hs
```diff
@@ -186,29 +183,20 @@ collectAuthMetrics m env = do
...
-  where
-    getFedRemotes = namedClient @IAPI.API @"get-federation-remotes"
+  
+  liftIO $ do
+    okRemoteDomains <- getAllowedDomainsInitial clientEnv
+    atomicWriteIORef ioref okRemoteDomains
+    let domainListsEqual old new =
+            Set.fromList (domain <$> fromFederationDomainConfigs old)
+              == Set.fromList (domain <$> fromFederationDomainConfigs new)
+        callback old new = unless (domainListsEqual old new) $ do
```
fisx, line 198: move this "unless" to inside `getAllowedDomainsLoop`?  i can't think of anything this callback might want to do if there are no changes.

fisx (commented):
:+1:, and some more nit-picks.

lepsa:
> @pcapriotti says we should use `IORef` instead of `TVar`. he might be right?

Probably, TVar was the tool I'm most familiar with for solving these type of problems. I've updated code to use IORef instead.

lepsa:
> 

I think this is from an older version of the code. federator's env isn't in an ioref/tvar at the current commit.

fisx:
> > 
> 
> I think this is from an older version of the code. federator's env isn't in an ioref/tvar at the current commit.

yeah i already fixed that.

lepsa (approved):

fisx:
> I've updated this text to mention how the update intervals are supplied by Brig in the federation domain update structure.

you mean "where did that text go?"?

i've moved it to the information for operators, but it's not very, erm, polished yet...

libs/types-common/src/Data/MessageQueue.hs
```diff
@@ -0,0 +1,21 @@
...
+  )
+where
+
+import Data.Aeson (FromJSON)
+import Data.Text (Text)
+import GHC.Generics (Generic)
+import Prelude (Show, String)
+
+-- | Options for connecting to the message queue system
+data MessageQueueSettings = MessageQueueSettings
```
fisx, line 12: come to think of it: this is not used for anything yet, right?  i'd prefer to not add new dead code, we already have a lot of that.  i've removed this; here is the commit for when you need to re-add it in the future: c0df89aafa7064484c571ae406dce1dc36747cbb

battermann (approved):

fisx:
one more problem: on this branch, when `./services/run-services` is running for a while, the system load goes crazy:

```
top - 12:21:43 up 13 days, 14:55, 15 users,  load average: 9.14, 5.47, 4.15
Tasks: 453 total,   1 running, 452 sleeping,   0 stopped,   0 zombie
%Cpu(s): 64.0 us, 16.3 sy,  0.0 ni, 14.8 id,  0.1 wa,  0.0 hi,  4.8 si,  0.0 st 
MiB Mem :  31838.1 total,   7409.8 free,  10775.9 used,  14810.5 buff/cache     
MiB Swap:    976.0 total,    829.2 free,    146.8 used.  21062.2 avail Mem 

    PID USER      PR  NI    VIRT    RES    SHR S  %CPU  %MEM     TIME+ COMMAND                                                       
1218137 systemd+  20   0 2180260 601940  40296 S 239.4   1.8  17:36.07 java                                                          
1253078 mf        20   0 1024.4g  98664  46148 S  96.7   0.3   1:02.59 brig                                                          
1253006 mf        20   0 1024.4g 109544  46196 S  96.0   0.3   1:05.70 brig                                                          
1217483 root      20   0 1599200  19416   2584 S  38.1   0.1   2:30.04 docker-proxy                                                  
1253084 mf        20   0 1024.3g  65808  38516 S  14.2   0.2   0:09.23 galley                                                        
1253012 mf        20   0 1024.3g  66296  38524 S  13.6   0.2   0:09.59 galley                                                        
1253018 mf        20   0 1024.3g  41976  21612 S  13.6   0.1   0:09.40 gundeck                                                       
1253090 mf        20   0 1024.3g  41476  21376 S  13.6   0.1   0:08.96 gundeck                                                       
1253120 mf        20   0 1024.4g  34036  17044 S  13.6   0.1   0:08.63 federator                                                     
1253055 mf        20   0 1024.4g  34036  16972 S  13.2   0.1   0:09.07 federator                                                     
1253024 mf        20   0 1024.2g  30880  14236 S  12.6   0.1   0:08.35 cannon                                                        
1253096 mf        20   0 1024.2g  28492  14216 S  12.6   0.1   0:08.01 cannon                                                        
1253427 root      20   0   28140  26300   5860 S  12.6   0.1   0:00.38 python                                                        
1253027 mf        20   0 1024.3g  30708  14300 S  12.3   0.1   0:08.38 cannon                                                        
1253101 mf        20   0 1024.3g  28892  14220 S  12.3   0.1   0:08.02 cannon                                                        
```

the docker-compose logs suggest the java process at the top is sqs.

develop is not misbehaving in this way.

charts/brig/templates/configmap.yaml
```diff
@@ -248,6 +248,9 @@ data:
       setDeleteThrottleMillis: {{ .setDeleteThrottleMillis }}
       setFederationDomain: {{ .setFederationDomain }}
       {{- if .setFederationDomainConfigs }}
+      # 'setFederationDomainConfigs' is deprecated as of https://github.com/wireapp/wire-server/pull/3260.  See
```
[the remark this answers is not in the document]
battermann: Also set defaults in values: `charts/brig/values.yaml`?

hack/helm_vars/wire-server/values.yaml.gotmpl
```diff
@@ -78,7 +78,9 @@ brig:
       # See helmfile for the real value
       setFederationDomain: integration.example.com
       setFederationDomainConfigs:
-        # See helmfile for the real value
+        # 'setFederationDomainConfigs' is deprecated as of https://github.com/wireapp/wire-server/pull/3260.  See
```
battermann, line 81: I think it makes sense to set values for CI here, too, for the following settings (even if it's only the default settings):

- `setFederationStrategy`
- `setFederationDomainConfigsUpdateFreq`

services/federator/federator.integration.yaml
```diff
@@ -39,4 +39,4 @@ optSettings:
   clientCertificate: "test/resources/integration-leaf.pem"
   clientPrivateKey: "test/resources/integration-leaf-key.pem"
   dnsHost: "127.0.0.1"
-  dnsPort: 9053
+  dnsPort: 9053
```
battermann, line 42: Remove `federationStrategy` from this config.

battermann (commented):
Also remove from `federationStrategy` from

- `charts/federator/templates/configmap.yaml`
- `charts/federator/values.yaml`
- `hack/helm_vars/wire-server/values.yaml.gotmpl`

integration/test/API/BrigInternal.hs
```diff
@@ -43,3 +44,70 @@ createUser domain cu = do
...
+createFedConn :: HasCallStack => FedConn -> App Response
+createFedConn fedConn = do
+  res <- createFedConn' fedConn
+  res.status `shouldMatchRange` (200, 299)
+  pure res
+
+createFedConn' :: HasCallStack => FedConn -> App Response
+createFedConn' fedConn = do
+  owndom <- ownDomain
+  req <- rawBaseRequest owndom Brig Unversioned "/i/federation/remotes"
```
smatting, line 76: No need to bind the domain, you can just pass it directly

```suggestion
  req <- rawBaseRequest ownDomain Brig Unversioned "/i/federation/remotes"
```
Also the domain should not be hardcoded in the API.* Modules like this one, so that the api function can be used in other tests as well, see uses of the `objDomain` function

services/brig/src/Brig/API/Internal.hs
```diff
@@ -171,6 +180,125 @@ authAPI =
...
+      "keeping track of remote domains in the brig config file is deprecated, but as long as we \
+      \do that, removing or updating items listed in the config file is not allowed."
+
+-- | Remove the entry from the database if present (or do nothing if not).  This responds with
+-- 533 if the entry was also present in the config file, but only *after* it has removed the
+-- entry from cassandra.
+deleteFederationRemotes :: Domain -> ExceptT Brig.API.Error.Error (AppT r) ()
+deleteFederationRemotes dom = do
+  lift . wrapClient . Data.deleteFederationRemote $ dom
+  assertNoDomainsFromConfigFiles dom
```
lepsa, line 300: The ordering on this delete then check seems weird, but allows us to default all the way back to config file state for a federation domain.

When we come to the point where we drop support for federation domains in the config files, we can remove most of these checks beyond "we know about this domain", and that any given domain isn't being updated to point at another domain. See `assertDomainIsNotUpdated`

integration/test/Test/Brig.hs
```diff
@@ -18,3 +21,74 @@ testSearchContactForExternalUsers = do
...
   bindResponse (Public.searchContacts partner (owner %. "name")) $ \resp ->
     resp.status `shouldMatchInt` 403
+
+testCrudFederationRemotes :: HasCallStack => App ()
+testCrudFederationRemotes = do
+  let parseFedConns :: HasCallStack => Response -> App [Internal.FedConn]
+      parseFedConns resp = do
+        -- TODO: not idiomatic!  try `getJSON 200 resp %. "remotes" & asList & mapM asObjOrSomething`
+        -- Some ideas: There is asList to assert that a Value is actually a an array of Values. Then you can sort that to have a defined order.
+        fromJust . Aeson.decode . Aeson.encode . fromJust <$> ((`lookupField` "remotes") =<< getJSON 200 resp)
```
lepsa, line 31: Why is there a round trip through Aeson here?

libs/wire-api/src/Wire/API/FederationUpdate.hs
```diff
@@ -0,0 +1,66 @@
...
+
+getFedRemotes :: ClientM FederationDomainConfigs
+getFedRemotes = namedClient @IAPI.API @"get-federation-remotes"
+
+-- Initial function for getting the set of domains from brig, and an update interval
+getAllowedDomainsInitial :: L.Logger -> ClientEnv -> IO FederationDomainConfigs
+getAllowedDomainsInitial logger clientEnv =
+  let -- keep trying every 3s for one minute
+      policy :: R.RetryPolicy
+      policy = R.constantDelay 3_081_003 <> R.limitRetries 20
```
lepsa, line 27: Where did this delay constant come from? retry has threadDelay under the hood which only offers lower bound accuracy on delay times, and an extra 3 microseconds seems like it would be lost in the noise of scheduling.

integration/test/Test/Brig.hs
lepsa: It was doing `ByteString -> Maybe a` by re-encoding the already parsed JSON. Using some other aeson functions allows us to get at the desired parse result more directly.

services/cannon/src/Cannon/Run.hs
```diff
@@ -75,6 +79,27 @@ run o = do
       <*> mkClock
   refreshMetricsThread <- Async.async $ runCannon' e refreshMetrics
   s <- newSettings $ Server (o ^. cannon . host) (o ^. cannon . port) (applog e) m (Just idleTimeout)
+
+  -- Get the federaion domain list from Brig and start the updater loop
+  manager <- newManager defaultManagerSettings
```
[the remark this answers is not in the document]
lepsa: We'll need it soon for sending out events to clients notifying them about the federation domain changes. In Galley it is also going to be used to update the databases to remove conversations and users as needed when a domain is dropped.

integration/test/API/BrigInternal.hs
```diff
@@ -72,6 +73,7 @@ createFedConn fedConn = do
 
 createFedConn' :: HasCallStack => FedConn -> App Response
 createFedConn' fedConn = do
+  ownDomain <- asks domain1
```
fisx: i removed these intentionally.  see new tip of branch.

integration/test/Test/Brig.hs
```diff
@@ -28,7 +28,7 @@ testCrudFederationRemotes = do
       parseFedConns resp = do
         -- TODO: not idiomatic!  try `getJSON 200 resp %. "remotes" & asList & mapM asObjOrSomething`
         -- Some ideas: There is asList to assert that a Value is actually a an array of Values. Then you can sort that to have a defined order.
-        fromJust . Aeson.decode . Aeson.encode . fromJust <$> ((`lookupField` "remotes") =<< getJSON 200 resp)
+        fromJust . parseMaybe parseJSON . fromJust <$> ((`lookupField` "remotes") =<< getJSON 200 resp)
```
fisx, line 31: still not idiomatic, but better.  :)  thanks!

services/galley/src/Galley/Run.hs
```diff
@@ -61,23 +60,24 @@ import qualified Network.Wai.Middleware.Gunzip as GZip
...
+import System.Logger.Extended (mkLogger)
 
 run :: Opts -> IO ()
 run opts = lowerCodensity $ do
-  (app, env) <- mkApp opts
+  (ioref, _) <- lift $ do
+    -- Duplicating the logger from App.createEnv so that we don't have to deal
+    -- with recursive monadic actions to get the logger before we have the initial
+    -- IORef of values from brig. It's just easier this way.
+    l <- mkLogger (opts ^. optLogLevel) (opts ^. optLogNetStrings) (opts ^. optLogFormat)
```
fisx, line 78: is this ok?  i'm worried about output from two loggers getting entangled somehow, even though i couldn't explain how.

can't we create the logger first thing in `run`, and then pass it to both `createEnv` and `updateFedDomains`?

integration/test/Test/Brig.hs
smatting: Or don't parse at all! This is the way

services/brig/brig.integration.yaml
```diff
@@ -190,16 +190,10 @@ optSettings:
   setFederationDomain: example.com
   setFeatureFlags:  # see #RefConfigOptions in `/docs/reference`
   setFederationDomainConfigsUpdateFreq: 1
-  setFederationStrategy: allowList
+  setFederationStrategy: allowAll
```
fisx, line 193: probably need to do this in more places?

lepsa: I don't think so. This will flow down to other services calling to brig in the integration tests.

services/galley/src/Galley/Run.hs
```diff
@@ -181,23 +180,12 @@ collectAuthMetrics m env = do
...
-    let domainListsEqual old new =
-          Set.fromList (domain <$> remotes old)
-            == Set.fromList (domain <$> remotes new)
-        callback old new = unless (domainListsEqual old new) $ do
-          -- TODO: perform the database updates here
-          -- This code will only run when there is a change in the domain lists
-          pure ()
-    getAllowedDomainsLoop logger clientEnv ioref callback
+callback :: FedUpdateCallback
+callback old new = unless (domainListsEqual old new) $ do
```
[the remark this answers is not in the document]
lepsa: Done and done.

services/galley/src/Galley/Run.hs
lepsa: Done and done.

integration/test/Test/Brig.hs
lepsa: I've removed more of the FedConn specific code, and now it is comparing json values and leaning on typeclasses to align values.

docs/src/developer/developer/federation-design-aspects.md
```diff
@@ -0,0 +1,29 @@
...
+
+The state is persisted in cassandra table `brig.federation_remotes`.
+brig provides the contents via an internal CRUD API (see
+{ref}`configure-federation-strategy-in-brig` for the links).  In the
+future, we may decide that brig needs to cache the table itself, but
+for now (`GET` is only used for the internal end-point to share it
+with other services) we hope to get away with the simple solution and
+always read from cassandra directly.
+
+(More details to be added?)
```
supersven, line 29: I don't have an answer for this question, yet. However, this comment should act as a marker such that we don't forget to either care for or delete this question.

docs/src/understand/configure-federation.md
```diff
@@ -395,7 +395,166 @@ cargohold:
...
+     - `strategy` (from config): one of `allowNone`, `allowDynamic`, `allowAll` (see above)
+     - `update_interval` (from config): the suggested update frequency with which calling
+       services should refresh their information.
+
+  - It doesn't serve the local domain, which needs to be configured
+   for every service that needs to know it individually.  This may
+   change in the future.
+
+  - This end-point enjoys a comparably high amount of traffic.  If you
+   have many a large instance (say, >100 pods), *and* you set a very
```
supersven, line 451:
```suggestion
   have many (a large instance with say, >100 pods), *and* you set a very
```

docs/src/developer/developer/federation-design-aspects.md
```diff
@@ -0,0 +1,29 @@
+# Federation Design Aspects
+
+(configuring-remote-connections-dev-perspective)=
+
+## keeping track of federator remotes
```
supersven, line 5:
```suggestion
## Keeping track of federator remotes
```

Start headline with uppercase letter for consistency.

docs/src/understand/configure-federation.md
```diff
@@ -395,7 +395,166 @@ cargohold:
...
+      setFederationDomainConfigsUpdateFreq: 10 # seconds
+```
+
+The default strategy of `allowNone` effectively disables federation
+(and probably isn't what you want if you are reading this).
+`allowAll` federates with any backend that requests contact or that a
+user uses in a search.  `allowDynamic` only federates with known
+remote backends listed in cassandra.
+
+The update frequence determines how often other services will refresh
```
supersven, line 424: I understand that having a single source of truth (`brig` in this case) guards against federation configs in other services going out of sync.

However, I'm wondering if this has to be solved with *polling* by the depending services? Can't `brig` tell the other services when its configuration changed by querying an endpoint (e.g. `POST galley/i/federation-config-changed`)? Initial sync could happen when a pod starts.

I haven't read the code, so I might be wrong here.

My impression is that these settings change very seldom and that wouldn't justify querying an endpoint every 10 Seconds or so. :thinking:

fisx: you're right, that would work; the question is is it worth the trouble?  what is worse: running a few 100 bogus requests inside wire-server every 10 secs, or implementing more internal end-points and clients that use it?

see paragraph below on traffic (line 450).

libs/wire-api/src/Wire/API/FederationUpdate.hs
fisx: that's fine, the 3 musecs aren't important.  it just always seems arbitrary to me to have so many 0s in a constant.  nature doens't work that way :-)

(on a more serious note, i think i learned this to make it less likely for systems to have load peaks; it might apply better to cron jobs on 1990ties machines than here.  meh.)

services/brig/src/Brig/API/Internal.hs
```diff
@@ -180,6 +186,125 @@ authAPI =
...
+  when (dom `elem` (domain <$> cfg)) $ do
+    throwError . fedError . FederationUnexpectedError $
+      "keeping track of remote domains in the brig config file is deprecated, but as long as we \
+      \do that, removing or updating items listed in the config file is not allowed."
+
+-- | Remove the entry from the database if present (or do nothing if not).  This responds with
+-- 533 if the entry was also present in the config file, but only *after* it has removed the
+-- entry from cassandra.
+deleteFederationRemotes :: Domain -> ExceptT Brig.API.Error.Error (AppT r) ()
+deleteFederationRemotes dom = do
```
fisx, line 304:
```suggestion
deleteFederationRemote :: Domain -> ExceptT Brig.API.Error.Error (AppT r) ()
deleteFederationRemote dom = do
```

services/federator/src/Federator/Run.hs
```diff
@@ -59,15 +61,17 @@ import qualified Wire.Network.DNS.Helper as DNS
 -- FUTUREWORK(federation): Add metrics and status endpoints
 run :: Opts -> IO ()
 run opts = do
+  logger <- LogExt.mkLogger (Opt.logLevel opts) (Opt.logNetStrings opts) (Opt.logFormat opts)
   let resolvConf = mkResolvConf (optSettings opts) DNS.defaultResolvConf
-  DNS.withCachingResolver resolvConf $ \res ->
-    bracket (newEnv opts res) closeEnv $ \env -> do
+  DNS.withCachingResolver resolvConf $ \res -> do
+    (ioref, updateAllowedDomainsThread) <- updateFedDomains (brig opts) logger (\_ _ -> pure ())
```
fisx, line 67:
```suggestion
    (ioref, updateFedDomainsThread) <- updateFedDomains (brig opts) logger (\_ _ -> pure ())
```

pls ignore me, can't help it.

lepsa:
> https://wearezeta.atlassian.net/browse/FS-1115
> ### internal docs:
> 
> http://localhost:8082/api-internal/swagger-ui/brig/
> 
> 2023-05-19-113146_1920x1080_scrot
> ### docs for devs:
> 
> 2023-06-04-143744_1600x900_scrot
> ### (slightly outdated) docs for operators:
> 
> 2023-06-04-143310_1600x900_scrot 2023-06-04-143317_1600x900_scrot 2023-06-04-143320_1600x900_scrot 2023-06-04-143325_1600x900_scrot 2023-06-04-143335_1600x900_scrot
> ## Checklist
> 
>     * [x]  Add a new entry in an appropriate subdirectory of `changelog.d`
> 
>     * [x]  Read and follow the [PR guidelines](https://docs.wire.com/developer/developer/pr-guidelines.html)
> 
> 
> ## TODO
> 
>     * [ ]  double-check what our default search policy in (a) the case where we allow all domains and a new one is not listed, and (b) the case where we have an allow entry in the cassandra table, but it has no search policy.  also check if we're getting it right in the docs.  (i think (a) should be no_search and (b) should be no NULL allowed in database field, just pick something, it's not that hard.)  maybe write integration tests?
> 
>     * [x]  log 'F' are sometimes too harsh.  'I' or 'D' would do.  eg., for "couldn't pull data from brig _yet_, retrying".
> 
>     * [ ]  yeah, galley integration tests where hanging for me as well, but the last 3 rounds everything went green ok.
> 
>     * [ ]  [Dynamic federator remotes #3260 (review)](https://github.com/wireapp/wire-server/pull/3260#pullrequestreview-1487587205)

TODO answers:

1. a) AllowAll will allow an unlisted domain to be federated with. See `ensureCanFederateWith`
b) The DB query types don't allow a domain to be set or fetched without an associated search policy. See `getFederationRemotes`, `addFederationRemote`, and `updateFederationRemote`. The underlying cassandra table stores the search policy as an integer, presumably using the Enum instance for mapping. This may prove to be problematic in the future if new values are inserted between existing values, and no warning about this is present in the code. A new value inserted into the existing structure will result in some values being incorrectly mapped back into haskell values from the database. My reading of the `cql` library leads me to believe that nulls in this table will result in an exception of some sort being thrown, as these column types aren't Maybe wrapped.

services/galley/src/Galley/Run.hs
```diff
@@ -79,17 +85,15 @@ run opts = lowerCodensity $ do
     void $ Codensity $ Async.withAsync $ collectAuthMetrics (env ^. monitor) (aws ^. awsEnv)
   void $ Codensity $ Async.withAsync $ runApp env deleteLoop
   void $ Codensity $ Async.withAsync $ runApp env refreshMetrics
+  void $ Codensity $ Async.withAsync $ runApp env undefined
```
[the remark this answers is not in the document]
lepsa: I think that this should be removed. `runApp` is going to call `unApp` on the undefined value and pass it into runReaderT. This will fail when it is waited on.

```
Loaded GHCi configuration from /tmp/cabal-repl.-94973/setcwd.ghci
λ> import Control.Monad.Reader
λ> import Control.Concurrent.Async
λ> wait =<< async (runReaderT undefined ())
*** Exception: Prelude.undefined
  undefined, called at <interactive>:3:28 in interactive:Ghci3
λ> 
```

integration/test/API/BrigInternal.hs
```diff
@@ -47,6 +48,72 @@ createUser domain cu = do
                ]
         )
 
+data FedConn = FedConn
+  { domain :: String,
+    searchStrategy :: String
+  }
+  deriving (Eq, Ord, Show)
+
+instance ToJSON FedConn where
```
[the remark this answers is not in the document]
lepsa: The ToJSON instance is needed for an assertion equality, so inlining it here will come with the cost of extra code elsewhere. The FromJSON instance was redundant.

libs/wire-api/src/Wire/API/FederationUpdate.hs
```diff
@@ -0,0 +1,78 @@
...
+  where
+    domainListsEqual o n =
+      Set.fromList (domain <$> remotes o)
+        == Set.fromList (domain <$> remotes n)
+
+updateFedDomains :: Endpoint -> L.Logger -> FedUpdateCallback -> IO (IORef FederationDomainConfigs, Async ())
+updateFedDomains (Endpoint h p) log' cb = do
+  clientEnv <- newManager defaultManagerSettings <&> \mgr -> ClientEnv mgr baseUrl Nothing defaultMakeClientRequest
+  ioref <- newIORef =<< getAllowedDomainsInitial log' clientEnv
+  updateDomainsThread <- async $ getAllowedDomainsLoop log' clientEnv cb ioref
```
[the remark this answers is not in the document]
lepsa: I've added in a finally in the `async` wrapped IO action. It will restart the loop and log an error message.

services/galley/src/Galley/Run.hs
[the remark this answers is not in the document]
fisx: yes, certainly the `undefined` needs to be refined.  :)

battermann:
This is just a comment, and there is no need to take action. Just to document what I have tested.

`setFederationStrategy` in `brig.integration.yaml` should be set to `allowAll` which matches the old setting in federator before this PR.

If we change `setFederationStrategy` to `allowDynamic` and add `example.com` and `b.example.com` to the allow list, a couple of integration tests fail. The reason is that some tests that expect `400 federation-denied` now get `422 invalid-domain`. Since the exact same thing happens when changing the old setting in federator to `allowDomains` this is expected and absolutely fine.

The errors are not listed in the specific endpoint descriptions in the swagger, but they are all listed as federation errors in the overall API descriptions. So clients should be aware and prepared to handle them all.

All looking good :+1:

charts/brig/values.yaml
```diff
@@ -100,6 +100,8 @@ config:
     # Disable one ore more API versions. Please make sure the configuration value is the same in all these charts:
     # brig, cannon, cargohold, galley, gundeck, proxy, spar.
     # setDisabledAPIVersions: [ v3 ]
+    setFederationStrategy: allowNone
```
battermann, line 103: The old setting for cloud was:

```
    federationStrategy:
      allowedDomains: []
``` 

Should we set it to `allowDynamic` with an empty list here, too, to exactly match the old behavior?

fisx:
ready to merge!  :shipit:

charts/brig/values.yaml
fisx: the two are different in behavior if the database table is non-empty.  so i'd argue `allowNone` is keeping behavior as unchanged as can be.

docs/src/developer/developer/federation-design-aspects.md
```diff
@@ -0,0 +1,29 @@
...
+
+The state is persisted in cassandra table `brig.federation_remotes`.
+brig provides the contents via an internal CRUD API (see
+{ref}`configure-federation-strategy-in-brig` for the links).  In the
+future, we may decide that brig needs to cache the table itself, but
+for now (`GET` is only used for the internal end-point to share it
+with other services) we hope to get away with the simple solution and
+always read from cassandra directly.
+
+(More details to be added?)
```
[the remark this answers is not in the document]
fisx: fair point.  but i'm not going to spend another day beating concourse for this change :-P

PR #3263, opened 2023-05-02
fix: forward original http host to proxy request in nginz as `Z-Host`

battermann:
To create a valid DPoP access token for the client's e2eid CSR we have to pass the correct domain to the crypto library.

We can extract the domain from the request via the HTTP host header.

However, with `nginz` in front of `brig` the host header is automatically redefined as `$proxy_host`. (see <http://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_set_header>)

To pass the original header, a new header `Z-Host` is defined and set to the original HTTP host header value via `$host`.

## Checklist

### DIFF
--- a/services/brig/src/Brig/API/Client.hs
+++ b/services/brig/src/Brig/API/Client.hs
@@ -87,7 +87,6 @@ import Data.Misc (PlainTextPassword6)
 import Data.Qualified
 import qualified Data.Set as Set
 import Data.String.Conversions (cs)
-import qualified Data.Text as Text
 import Imports
 import Network.HTTP.Types.Method (StdMethod)
 import Network.Wai.Utilities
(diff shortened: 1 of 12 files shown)


services/brig/src/Brig/API/Client.hs
```diff
@@ -486,10 +485,10 @@ createAccessToken ::
   StdMethod ->
   Link ->
   Proof ->
-  Maybe Text ->
+  Text ->
   ExceptT CertEnrollmentError (AppT r) (DPoPAccessTokenResponse, CacheControl)
-createAccessToken uid cid method link proof mHostHeader = do
-  domain <- maybe (throwE MisconfiguredRequestUrl) (pure . Domain) $ mHostHeader <&> Text.splitOn ":" >>= headMay
+createAccessToken uid cid method link proof host = do
+  let domain = Domain host
```
elland, line 491: 👌🏼

elland (approved):
LGTM

PR #3265, opened 2023-05-02
Improve integration test suite

smatting:
- remove tasty for a custom runner
- remove dependency on "imports" for own Prelude
- fix minor bug in Makefile "ci" target
- Add the federation domain into all `baseRequest`
- Add some API calls
- Don't show library functions in call stack


battermann (approved):

integration/test/Testlib/Run.hs
```diff
@@ -1,46 +1,121 @@
...
 import Testlib.Options
+import Text.Printf
+import UnliftIO.Async
+
+-- TODO
+--
+-- [x] Limit concurrency to the number of capabilities
+-- [x] Parse configuration before running tests
+-- [x] Add -c option for the configuration directory
+-- [ ] Clean up temporary files
```
pcapriotti, line 22: There's a missing item here.

pcapriotti (approved):
LGTM

integration/test/Testlib/Run.hs
smatting: Thanks, fixed it.

PR #3276, opened 2023-05-09
[FS-1207] Start sending backend notifications through rabbitMQ and consuming them

elland:
## Checklist


changelog.d/0-release-notes/background-worker
```diff
@@ -0,0 +1,38 @@
...
+    remoteDomains: []
+  secrets:
+    rabbitmq:
+      username: <YOUR_USERNAME>
+      password: <YOUR_PASSWORD>
+```
+
+The above are the default values (except for secrets, which do not have
+defaults), if they work they are not required to be configured.
+`background-worker.config.remoteDomains` should contain all the remote domains
```
fisx, line 36: this will change again with #3260, but let's worry about that there.

changelog.d/0-release-notes/background-worker
```diff
@@ -0,0 +1,38 @@
+This release introduces a new component: background-worker. This is currently
+only used to forward notifications to federated backends. Enabling federation in
+the wire-server helm chart automatically installs this component.
+
+When federation is enabled, wire-server will require running RabbitMQ. The helm
+chart in `rabbitmq` can be used to install RabbitMQ. Please refer to the
+documentation at https://docs.wire.com to install RabbitMQ in Kubernetes. These
+new configurations are required:
```
fisx, line 8: <plsignoreimtalkingtomyself>move the yaml snippet to docs.wire.com?  or if it's already there: link to it rather than redundantly copy it?  on the other hand, that would complicate the situation in which we change docs.wire.com in a later release, and somebody upgrades to this release afterwards.</plsignoreimtalkingtomyself>

charts/background-worker/templates/deployment.yaml
```diff
@@ -9,12 +9,11 @@ metadata:
     heritage: {{ .Release.Service }}
 spec:
   replicas: {{ .Values.replicaCount }}
-  # TODO(elland): Review this
   strategy:
-    type: RollingUpdate
-    rollingUpdate:
-      maxUnavailable: 0
-      maxSurge: {{ .Values.replicaCount }}
+    # Ensures only one version of the background worker is running at any given
```
fisx, line 13:
```suggestion
    # Ensures only one instance (or k8s pod) of the background worker is running at any given
```

is that what you mean?

charts/background-worker/values.yaml
```diff
@@ -9,12 +9,11 @@ resources:
     cpu: "100m"
   limits:
     memory: "512Mi"
-# TODO(elland): Create issue for a metrics endpoint
+# FUTUREWORK: Implement metrics
```
fisx, line 12: we can do this in a separate PR, but we should make sure we won't forget and have a ticket:

https://wearezeta.atlassian.net/browse/FS-2020

btw, which ticket is this PR about?

charts/background-worker/templates/deployment.yaml
akshaymankar: No, many K8s pods can run at once. This way of deploying just deletes previous version before deploying the new version. Basically does the opposite of a rolling deployment.

charts/background-worker/values.yaml
akshaymankar: It is this: https://wearezeta.atlassian.net/browse/FS-1940
It was in the title, also added it to description now.

akshaymankar: I added a `parent of` relation, maybe that's all that was needed :crossed_fingers:

libs/wire-api/src/Wire/API/MakesFederatedCall.hs
```diff
@@ -78,6 +79,16 @@ data Component
   | Cargohold
   deriving (Show, Eq, Generic)
   deriving (Arbitrary) via (GenericUniform Component)
+  deriving (ToJSON, FromJSON) via (Schema Component)
+
+instance ToSchema Component where
```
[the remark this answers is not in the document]
akshaymankar: It would be so much easier once we merge these services :smile:

services/brig/test/integration/API/User/Account.hs
```diff
@@ -157,8 +154,6 @@ tests _ at opts p b c ch g aws userJournalWatcher =
       test p "delete/by-code" $ testDeleteUserByCode b,
       test p "delete/anonymous" $ testDeleteAnonUser b,
       test p "delete with profile pic" $ testDeleteWithProfilePic b ch,
-      test p "delete with connected remote users" $ testDeleteWithRemotes opts b,
-      test p "delete with connected remote users and failed remote notifcations" $ testDeleteWithRemotesAndFailedNotifications opts b c,
```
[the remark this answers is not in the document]
akshaymankar: The first test is already being tested in end-to-end tests. It is very hard to test in the integration tests.

The second case is not valid anymore because we cannot fail due to remote backend being down, we can fail due to rabbitmq being down, but that is just a 500. I didn't think we should test 500 scenarios.

services/run-services
```diff
@@ -400,7 +401,7 @@ if __name__ == '__main__':
         'AWS_ACCESS_KEY_ID': "dummykey",
         'AWS_SECRET_ACCESS_KEY': "dummysecret",
         'RABBITMQ_USERNAME': 'guest',
-        'RABBITMQ_PASSWORD': 'alpaca-grapefruit'
+        'RABBITMQ_PASSWORD': 'guest'
```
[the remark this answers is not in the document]
akshaymankar: There was botched merge, perhaps this happened due to that. I will add it back.

services/background-worker/test/Test/Wire/BackendNotificationPusherSpec.hs
```diff
@@ -0,0 +1,140 @@
...
+    readIORef envelope.acks `shouldReturn` 0
+    readIORef envelope.rejections `shouldReturn` [False]
+    fedReqs `shouldBe` []
+
+  it "should retry failed deliveries" $ do
+    isFirstReqRef <- newIORef True
+    let returnSuccessSecondTime _ =
+          atomicModifyIORef isFirstReqRef $ \isFirstReq ->
+            if isFirstReq
+              then (False, ("text/html", "<marquee>down for maintenance</marquee>"))
```
smatting, line 92: ;)

smatting (approved):

services/background-worker/src/Wire/BackendNotificationPusher.hs
```diff
@@ -0,0 +1,90 @@
...
+          Log.msg (Log.val "Failed to parse notification, the notification will be ignored")
+            . Log.field "domain" (domainText targetDomain)
+            . Log.field "error" e
+
+        -- FUTUREWORK: This rejects the message without any requeueing. This is
+        -- dangerous as it could happen that a new type of notification is
+        -- introduced and an old instance of this worker is running, in which case
+        -- the notification will just get dropped. On the other hand not dropping
+        -- this message blocks the whole queue. Perhaps there is a better way to
+        -- deal with this.
```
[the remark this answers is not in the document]
akshaymankar: In case of a reverting a deployment, it would get blocked forever.

changelog.d/0-release-notes/background-worker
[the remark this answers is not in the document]
akshaymankar: I believe https://github.com/wireapp/wire-server/pull/3260 is going to solve this.

services/background-worker/src/Wire/BackendNotificationPusher.hs
```diff
@@ -0,0 +1,90 @@
...
+        ceFederator <- asks federatorInternal
+        ceHttp2Manager <- asks http2Manager
+        let ceOriginDomain = notif.ownDomain
+            ceTargetDomain = targetDomain
+            fcEnv = FederatorClientEnv {..}
+        liftIO $ either throwM pure =<< sendNotification fcEnv notif.targetComponent notif.path notif.body
+        lift $ ack envelope
+
+-- FUTUREWORK: Recosider using 1 channel for many consumers. It shouldn't matter
+-- for a handful of remote domains.
```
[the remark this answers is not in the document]
akshaymankar: The channel itself doesn't block when a domain is unreachable. It only blocks the particular consumer who is supposed to be pushing notifications to the unreachable domain. Other consumers keep going.

I suspect that this will reach some limit of consumers or just be slow when we have hundreds of consumers running. But perhaps we can solve that problem when we get there.

PR #3277, opened 2023-05-09
Add conversation.protocol-update event

smatting:
This PR adds sending a `conversation.update-protocol` event to all conversation participants when the protocol is switched.


smatting:
Integration tests passed

PR #3278, opened 2023-05-09
Handle race conditions in /integration

fisx:
## Checklist

### DIFF
--- a/.hlint.yaml
+++ b/.hlint.yaml
@@ -12,15 +12,11 @@
 - ignore: { name: Eta reduce }
 - ignore: { name: Use section }
 - ignore: { name: Use underscore }
-- ignore: { name: no recovering }
-- ignore: { name: no threadDelay }
 
 # custom rules:
 - hint: { lhs: (() <$), rhs: void }
 - hint: { lhs: return, rhs: pure }
 - hint: { lhs:  maybe mempty, rhs: foldMap }
-- hint: { name: no threadDelay, lhs: threadDelay, rhs: unrace }
-- hint: { name: no recovering, lhs: recovering, rhs: unrace }
 
 # We want the latter function because it handles signals properly.
 - error: { name: Use shutdown, lhs: runSettings, rhs: runSettingsWithShutdown }
--- a/.hlint.yaml
+++ b/.hlint.yaml
@@ -12,11 +12,15 @@
 - ignore: { name: Eta reduce }
 - ignore: { name: Use section }
 - ignore: { name: Use underscore }
+- ignore: { name: no recovering }
+- ignore: { name: no threadDelay }
 
 # custom rules:
 - hint: { lhs: (() <$), rhs: void }
 - hint: { lhs: return, rhs: pure }
 - hint: { lhs:  maybe mempty, rhs: foldMap }
+- hint: { name: no threadDelay, lhs: threadDelay, rhs: unrace }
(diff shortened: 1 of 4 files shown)


.hlint.yaml
```diff
@@ -12,11 +12,15 @@
...
 - ignore: { name: Use underscore }
+- ignore: { name: no recovering }
+- ignore: { name: no threadDelay }
 
 # custom rules:
 - hint: { lhs: (() <$), rhs: void }
 - hint: { lhs: return, rhs: pure }
 - hint: { lhs:  maybe mempty, rhs: foldMap }
+- hint: { name: no threadDelay, lhs: threadDelay, rhs: unrace }
+- hint: { name: no recovering, lhs: recovering, rhs: unrace }
```
fisx, line 23: i have tried to limit those rules to package `integration`, but couldn't find a way.  limiting to modules didn't work *and* is inconvenient, as we have to remember to add every new module from `integration` here as well.

introducing the rules and then ignoring them where they are in the way seems the most robust way to do this, but this didn't work either.

i'm open to ideas and opinions, including "stupid idea, don't do this with hlint, we'll remember ourselves".

fisx, line 23: i have tried to limit those rules to package `integration`, but couldn't find a way.  limiting to modules didn't work *and* is inconvenient, as we have to remember to add every new module from `integration` here as well.

introducing the rules and then ignoring them where they are in the way seems the most robust way to do this, but this didn't work either.

i'm open to ideas and opinions, including "stupid idea, don't do this with hlint, we'll remember ourselves".

integration/test/Test/Demo.hs
```diff
@@ -68,3 +70,14 @@ testMultipleBackends = do
...
+
+testUnrace :: App ()
+testUnrace = do
+  {-
+  -- the following would retry for ~30s and only then fail
+  unrace $ do
+    True `shouldMatch` True
+    True `shouldMatch` False
+  -}
+  () <- undefined (recovering @IO) threadDelay
```
fisx, line 82: this will go away once the hlint questions have been answered.

integration/test/Testlib/App.hs
```diff
@@ -49,3 +50,12 @@ ownDomain = asks (.domain1)
 
 otherDomain :: App String
 otherDomain = asks (.domain2)
+
+-- | Run an action, `recoverAll`ing with exponential backoff (min step 8ms, total timeout
+-- ~30s).  Search this package for examples how to use it.
```
fisx, line 55: 15s max would also be ok for me.  or 10?

elland: Yeah, our setup _should_ be able to resolve things within 15s, if not, then maybe something else is wrong?

integration/test/Testlib/App.hs
```diff
@@ -49,3 +50,12 @@ ownDomain = asks (.domain1)
 
 otherDomain :: App String
 otherDomain = asks (.domain2)
+
+-- | Run an action, `recoverAll`ing with exponential backoff (min step 8ms, total timeout
+-- ~15s).  Search this package for examples how to use it.
+--
+-- Ideally, this will be the only thing you'll ever need from the retry package when writing
+-- integration tests.  If you are unhappy with it,, please consider fixing it so everybody can
```
battermann, line 58:
```suggestion
-- integration tests.  If you are unhappy with it, please consider fixing it so everybody can
```

battermann (approved):

PR #3284, opened 2023-05-11
Update docs.wire.com

fisx:
## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -85,7 +85,7 @@ will, eventually, have built a range of docker images. Make sure to [give Docker
 
 See the `Makefile`s and `Dockerfile`s, as well as [build/ubuntu/README.md](build/ubuntu/README.md) for details.
 
-#### 2. Use nix-provided build environment 
+#### 2. Use nix-provided build environment
 
 This is suitable only for local development and testing. See [build instructions](./docs/src/developer/developer/building.md) in the developer documentation.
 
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -95,4 +95,4 @@ You have two options:
 
 * Option 1. (recommended) Install wire-server on kubernetes using the configuration and instructions provided in [wire-server-deploy](https://github.com/wireapp/wire-server-deploy). This is the best option to run it on a server and recommended if you want to self-host wire-server.
 
-* Option 2. Compile everything in this repo, then you can use the `services/start-services-only.sh`. This option is intended as a way to try out wire-server on your local development machine and not suited for production.
+* Option 2. Compile everything in this repo, then you can use the `services/run-services`. This option is intended as a way to try out wire-server on your local development machine and not suited for production.
```
lepsa, line 98: Given that start-services-only.sh is only calling out to run-services, should that script be removed too?

docs/src/developer/developer/how-to.md
```diff
@@ -11,21 +11,18 @@ Check that this file exists by running `ls $KUBECONFIG`.
 
 ## How to look at the swagger docs / UI locally
 
-Terminal 1:
-* Set up backing services: `./deploy/dockerephemeral/run.sh`
```
lepsa, line 15: Local testings shows that it still needs the backing databases. Maybe this line should be left in, or something to the effect of having DBs accessible.

elland: Yeah that's still needed no?

services/start-services-only.sh
```diff
@@ -1,10 +1,5 @@
-#!/usr/bin/env bash
-
-# Run all haskell services without immediately starting a test executable.
-# Can be useful for manually poking at the API.
-set -eo pipefail
+n#!/usr/bin/env bash
```
lepsa, line 1: Is this n intentional?

fisx: nope!

docs/src/developer/developer/how-to.md
fisx: oops, you're right.

README.md
fisx: no, i'm worried somebody will still use it and stumble over the fact that it's gone.  the way i changed it it doesn't do anything, but it'll point everybody who uses it in the right direction.

fisx:
+15 −21 and two mistakes.  that's not great...  :)  thanks for spotting them!

elland (approved):

PR #3287, opened 2023-05-11
Restore deleted scripts

smatting:
[FS-1148]: https://wearezeta.atlassian.net/browse/FS-1148?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/hack/bin/create_test_team_members.sh
+++ b/hack/bin/create_test_team_members.sh
@@ -1,0 +1,137 @@
+#!/usr/bin/env bash
+
+set -e
+
+ADMIN_UUID="a09e9521-e14e-4285-ad71-47caa97f4a16"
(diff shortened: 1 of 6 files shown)


fisx (approved):

hack/bin/create_test_team_members.sh
```diff
@@ -0,0 +1,137 @@
+#!/usr/bin/env bash
```
fisx, line 1:
```suggestion
#!/usr/bin/env bash
#
# consider using create_team.py (you'll have to evolve it a little further to cover this use case, though)
```

PR #3300, opened 2023-05-16
Remove INTEGRATION_FEDERATION_TESTS variable

smatting:
With this PR all end2end tests will always run locally


akshaymankar (approved):
Maybe we can also delete the `sans-federation` make targets and shell scripts. Feel free to do it separately.

PR #3305, opened 2023-05-19
Register and Update OAuth client via Stern/Backoffice

battermann:
## Checklist

### DIFF
--- a/tools/stern/src/Stern/Intra.hs
+++ b/tools/stern/src/Stern/Intra.hs
@@ -888,7 +888,7 @@ getOAuthClient cid = do
   case statusCode r of
     200 -> parseResponse (mkError status502 "bad-upstream") r
     404 -> throwE (mkError status404 "bad-upstream" "not-found")
-    _ -> throwE (mkError status502 "bad-upstream" "")
+    _ -> throwE (mkError status502 "bad-upstream" (cs $ show r))
 
 updateOAuthClient :: OAuthClientId -> OAuthClientConfig -> Handler OAuthClient
 updateOAuthClient cid conf = do
@@ -898,7 +898,7 @@ updateOAuthClient cid conf = do
       rpc'
         "brig"
         b
-        ( method POST
+        ( method PUT
             . Bilge.paths ["i", "oauth", "clients", toByteString' cid]
             . Bilge.json conf
             . contentJson
--- a/tools/stern/src/Stern/Intra.hs
+++ b/tools/stern/src/Stern/Intra.hs
@@ -59,6 +59,10 @@ module Stern.Intra
     getUserClients,
     getUserCookies,
     getUserNotifications,
+    registerOAuthClient,
+    getOAuthClient,
+    updateOAuthClient,
+    deleteOAuthClient,
   )
 where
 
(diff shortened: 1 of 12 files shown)


tools/stern/src/Stern/Intra.hs
```diff
@@ -853,3 +858,64 @@ getUserNotifications uid = do
...
+    rpc'
+      "brig"
+      b
+      ( method GET
+          . Bilge.paths ["i", "oauth", "clients", toByteString' cid]
+      )
+  case statusCode r of
+    200 -> parseResponse (mkError status502 "bad-upstream") r
+    404 -> throwE (mkError status404 "bad-upstream" "not-found")
+    _ -> throwE (mkError status502 "bad-upstream" "")
```
fisx, line 891:
```suggestion
    _ -> throwE (mkError status502 "bad-upstream" (show r))
```

(not tested; not that important, but might save some time if it breaks.)

tools/stern/test/integration/API.hs
```diff
@@ -93,14 +95,34 @@ tests s =
       test s "GET /i/consent" testGetConsentLog,
       test s "GET /teams/:id" testGetTeamInfo,
       test s "GET i/user/meta-info?id=..." testGetUserMetaInfo,
-      test s "/teams/:tid/search-visibility" testSearchVisibility
+      test s "/teams/:tid/search-visibility" testSearchVisibility,
+      test s "i/oauth/clients" testCrudOAuthClient
```
fisx, line 99: why the `/i`?  stern doesn't have any public-facing api, so the distinction doesn't really work here i think.

tools/stern/test/integration/API.hs
```diff
@@ -93,14 +95,34 @@ tests s =
...
       -- The following endpoints can not be tested because they require ibis:
       -- - `GET /teams/:tid/billing`
       -- - `GET /teams/:tid/invoice/:inr`
       -- - `PUT /teams/:tid/billing`
       -- - `POST /teams/:tid/billing`
     ]
 
+testCrudOAuthClient :: TestM ()
+testCrudOAuthClient = do
```
fisx, line 108: i won't insist, but maybe add new stern tests to `/integration` as well in the future?  it's also good to save time, though...

fisx (approved):

tools/stern/test/integration/API.hs
battermann: I've seen it on other stern endpoints, too, so I adapted it. I guess, it just serves as a marker to make it obvious that this is calling an internal endpoint. Should I remove it?

tools/stern/test/integration/API.hs
battermann: Yes, good point. Now the work is done. Let's keep it for now.

tools/stern/test/integration/API.hs
fisx: when i complained about this i thought this was the only end-point in stern with an `/i`, and that stern is only calling internal end-points.  both assumptions may be wrong.  use your own judgement, if it's consistent with what stern does everywhere, keep it.

fisx: ah, i see i'm a bit late.  :)  either way it's not a big deal.

PR #3306, opened 2023-05-22
MLS: Remove incorrect check

smatting:
This check is incorrect because the committer can be anyone from the
conversation. If it all this check should be applied at the level of the proposal.


smatting:
Closing because already included in #3303

PR #3312, opened 2023-05-24
Simplify cabal.project

fisx:
According to [the docs](https://cabal.readthedocs.io/en/3.4/cabal-project.html#package-configuration-options), we can use `*` instead of enumerating all packages.

## Checklist


fisx:
this may cause more trouble than it avoids, as we need to disable -Werror for all dependencies that compile with warnings.  or fix the world upstream.  i'm giving up for now.

PR #3313, opened 2023-05-24
FS-1959: Setup schemas and ES indices for dynamic backend tests

smatting:
This PR sets up separate cassandra schemas and ES indices to be used by integration tests.
The number of backends can be controlled by `INTEGRATION_DYNAMIC_BACKENDS_POOLSIZE` for the local dev environment and `config.dynamicBackendsPoolsize` in the `integration` helm chart.

The db schemas will be named `brig_test_dyn_X`, `galley_test_dyn_X`, etc; the ES index is named `directory_dyn_X_test`, where `X` ranges from 1 to the poolsize.

### DIFF
--- a/charts/integration/templates/integration-integration.yaml
+++ b/charts/integration/templates/integration-integration.yaml
@@ -34,6 +34,24 @@ spec:
         name: "turn"
 
   restartPolicy: Never
+
+  initContainers:
+  - name: integration-scripts
+    image: "{{ .Values.image.repository }}:{{ .Values.image.tag }}"
+    env:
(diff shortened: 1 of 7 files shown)


charts/integration/templates/integration-integration.yaml
```diff
@@ -34,6 +34,24 @@ spec:
         name: "turn"
 
   restartPolicy: Never
+
+  initContainers:
+  - name: integration-scripts
```
smatting, line 39: TODO: rename this to `integration-setup`

battermann (approved):

PR #3314, opened 2023-05-25
brig: Make RabbitMQ config optional

akshaymankar:
It is only required when federation is enabled.

## Checklist

### DIFF
--- a/services/brig/src/Brig/Federation/Client.hs
+++ b/services/brig/src/Brig/Federation/Client.hs
@@ -152,12 +152,22 @@ notifyUserDeleted ::
 notifyUserDeleted self remotes = do
   let remoteConnections = tUnqualified remotes
   let notif = UserDeletedConnectionsNotification (tUnqualified self) remoteConnections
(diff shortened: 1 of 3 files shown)


services/brig/src/Brig/Federation/Client.hs
```diff
@@ -152,12 +152,22 @@ notifyUserDeleted ::
...
+      remoteDomain = tDomain remotes
+  view rabbitmqChannel >>= \case
+    Just chanVar -> do
+      enqueueNotification (tDomain self) remoteDomain Q.Persistent chanVar $ void $ fedQueueClient @'Brig @"on-user-deleted-connections" notif
+    Nothing ->
+      Log.err $
+        Log.msg ("Federation error while notifying remote backends of a user deletion." :: ByteString)
+          . Log.field "user_id" (show self)
+          . Log.field "domain" (domainText remoteDomain)
+          . Log.field "error" (show FederationNotConfigured)
```
fisx, line 164: i think this could be an idiom like "if channel, then do this, otherwise log error" somewhere in wire-api?  doesn't have to happen in this hotfix PR, of course.

fisx (approved):

PR #3319, opened 2023-05-26
Render `Named` names as "internal route ID" in swagger UI.

fisx:
This should make jumping from swagger docs into the source code by grepping for the ID much easier.

2023-05-30-222505_1600x900_scrot
2023-05-30-222522_1600x900_scrot

## Checklist

### DIFF
--- a/libs/wire-api/src/Wire/API/Routes/Named.hs
+++ b/libs/wire-api/src/Wire/API/Routes/Named.hs
@@ -19,7 +19,7 @@
 
 module Wire.API.Routes.Named where
 
-import Control.Lens ((?~))
+import Control.Lens ((%~))
 import Data.Kind
 import Data.Metrics.Servant
 import Data.Proxy
@@ -48,7 +48,11 @@ instance {-# OVERLAPPING #-} (RenderableSymbol a, RenderableSymbol b) => Rendera
 instance (HasSwagger api, RenderableSymbol name) => HasSwagger (Named name api) where
   toSwagger _ =
     toSwagger (Proxy @api)
-      & (info . description) ?~ ("internal route ID: " <> cs (renderSymbol @name))
+      & allOperations . summary %~ (<> Just dscr)
+      & allOperations . description %~ (Just (dscr <> "\n\n") <>)
+    where
+      dscr :: Text
+      dscr = " [internal route ID: " <> cs (renderSymbol @name) <> "]"
 
 instance HasServer api ctx => HasServer (Named name api) ctx where
   type ServerT (Named name api) m = Named name (ServerT api m)
--- a/libs/wire-api/src/Wire/API/Routes/Named.hs
+++ b/libs/wire-api/src/Wire/API/Routes/Named.hs
@@ -25,7 +25,6 @@ import Data.Metrics.Servant
 import Data.Proxy
 import Data.String.Conversions (cs)
 import Data.Swagger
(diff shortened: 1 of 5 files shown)


libs/wire-api/src/Wire/API/Routes/Named.hs
```diff
@@ -19,21 +19,38 @@
...
+-- | For 'HasSwagger' instance of 'Named'.  'KnownSymbol' isn't enough because we're using
+-- types other than string literals in some places.
+class RenderableSymbol a where
+  renderSymbol :: Text
+
+instance {-# OVERLAPPABLE #-} KnownSymbol a => RenderableSymbol a where
+  renderSymbol = cs $ symbolVal (Proxy @a)
+
+instance {-# OVERLAPPING #-} (RenderableSymbol a, Typeable b) => RenderableSymbol (a, b) where
+  renderSymbol = "(" <> renderSymbol @a <> ", " <> (cs . show . typeRep $ (Proxy @b)) <> ")"
```
fisx, line 48: These are not sufficient to prevent this error from being raised:

```
src/Wire/API/Routes/Public/Galley.hs:54:14: error:
    • No instance for (Wire.API.Routes.Named.RenderableSymbol
                         '("put", Wire.API.Team.Feature.LegalholdConfig))
        arising from a use of ‘toSwagger’
    • In the expression: toSwagger (Proxy @ServantAPI)
      In an equation for ‘swaggerDoc’:
          swaggerDoc = toSwagger (Proxy @ServantAPI)
   |
54 | swaggerDoc = toSwagger (Proxy @ServantAPI)
```

But I don't understand why?  `LegalHold` is certainly `Typeable`, because everything is implicitly; and type-level string literals (`Symbol`) are the first instance of `KnownSymbol` that comes to mind.

fisx: 5adbe2be1b0b1168d3a7cd3de53103210da7abb9  thanks @pcapriotti !

libs/wire-api/src/Wire/API/Routes/Named.hs
```diff
@@ -19,21 +19,40 @@
...
+  renderSymbol = cs $ symbolVal (Proxy @a)
+
+instance {-# OVERLAPPING #-} (RenderableSymbol a, RenderableSymbol b) => RenderableSymbol '(a, b) where
+  renderSymbol = "(" <> (renderSymbol @a) <> ", " <> (renderSymbol @b) <> ")"
+
+instance (HasSwagger api, RenderableSymbol name) => HasSwagger (Named name api) where
+  toSwagger _ =
+    toSwagger (Proxy @api)
+      & allOperations . summary %~ (<> Just dscr)
+      & allOperations . description %~ (Just (dscr <> "\n\n") <>)
```
fisx, line 52: not sure we want both, but i can't decide which one to prefer?

fisx:
i also thought maybe we can make it a rule that the internal IDs are strings that can be fed to grep verbatim.

and maybe document this in docs.wire.com somewhere, and link to those docs from the "internal route ID" label.

libs/wire-api/src/Wire/API/Routes/Named.hs
smatting: Let's keep the one in the description, not the summary

smatting (approved):

PR #3320, opened 2023-05-26
Export Data.String.Conversions.cs from Imports

fisx:
changes:

- b0964bcdd30d9e8637c405a38b191a58f73efedc
- `s/SBS/ByteString/g`
- `s/LBS/LByteString/g`
- `s/ST/Text/g`
- `s/LT/LText/g`
- remove `import Data.String.Conversions.*`  (at least in most places...)
- hide `cs` from `Imports` imports in a few places to avoid name shadowing

cs is technically not always safe, because it can crash when trying to convert a bytestring elf binary into text.  but first of all, we already use cs everywhere, and furthermore, unpack does exactly the same thing, so we wouldn't be any safer without cs.  (i think.)

## Checklist

### DIFF
--- a/changelog.d/4-docs/pr-3320
+++ b/changelog.d/4-docs/pr-3320
@@ -1,0 +1,1 @@
+ Export `Data.String.Conversions.cs` from `Imports`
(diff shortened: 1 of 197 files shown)


battermann (approved):
LGTM

elland (approved):

PR #3322, opened 2023-05-29
[FS-1148] Backport general changes from `mls` to `develop`

mdimjasevic:
There are a few things I introduced while working on #3304 after Paolo provided good feedback. These are not specific to MLS, hence this PR that backports them to `develop`.

Tracked by https://wearezeta.atlassian.net/browse/FS-1148.

## Checklist


libs/wire-api-federation/src/Wire/API/Federation/API/Galley.hs
```diff
@@ -423,7 +424,12 @@ data MLSMessageResponse
   = MLSMessageResponseError GalleyError
   | MLSMessageResponseProtocolError Text
   | MLSMessageResponseProposalFailure Wai.Error
-  | MLSMessageResponseUpdates [ConversationUpdate] (Maybe UnreachableUsers)
+  | -- | The conversation-owning backend could not reach some of the backends that
+    -- have users in the conversation when processing a commit.
+    MLSMessageResponseUnreachableBackends (Set Domain)
+  | -- | If the list of unreachable users is non-empty, it corresponds to users
+    -- that an application message could not be sent to.
+    MLSMessageResponseUpdates [ConversationUpdate] (Maybe UnreachableUsers)
```
fisx, line 432: `Maybe (NonEmpty ..)` instead of just `[..]` is a bit silly, but ok.  :)  that discussion doesn't belong into this PR.

fisx: Also, can you explain in the haddocks how users can be unreachable, if not by their backend being unreachable?  (If this is obvious to most readers let me know; I have very little context on MLS yet.)

libs/wire-api-federation/src/Wire/API/Federation/Error.hs
```diff
@@ -304,6 +318,16 @@ federationUnexpectedError msg =
     "federation-unexpected-wai-error"
     ("Could parse body, but got an unexpected error response: " <> LT.fromStrict msg)
 
+federationUnreachableError :: Set Domain -> Wai.Error
+federationUnreachableError (Set.map domainText -> ds) =
+  Wai.mkError
+    status
+    "federation-unreachable-domains-error"
+    ("The following domains are unreachable: " <> (LT.pack . show . Set.toList) ds)
```
fisx, line 326:
```suggestion
federationUnreachableError ds =
  Wai.mkError
    status
    "federation-unreachable-domains-error"
    ("The following domains are unreachable: " <> (LT.pack . show . fmap domainText . Set.toList) ds)
```

services/galley/test/integration/API/MLS.hs
```diff
@@ -1125,10 +1125,9 @@ testAppMessageSomeReachable = do
     let commitMocks =
           receiveCommitMockByDomain [bob1, charlie1]
             <|> welcomeMock
-    (([event], ftpCommit), _) <-
+    ([event], _) <-
       withTempMockFederator' commitMocks $ do
-        sendAndConsumeCommitFederated commit
-    liftIO $ ftpCommit @?= mempty
+        sendAndConsumeCommit commit
```
fisx, line 1130: have you tried translating this test case into `/integration`?  i'd give it a (time-boxed) try after this PR has been merged, `/integration` is really neat!  :)

services/galley/src/Galley/API/Action.hs
```diff
@@ -745,15 +745,15 @@ notifyConversationAction tag quid notifyOrigDomain con lconv targets action = do
     -- For now these users will not be able to join the conversation until
     -- queueing and retrying is implemented.
     let failedNotifies = lefts notifyEithers
+    for_ failedNotifies $
+      logError
```
fisx, line 749: not sure about the log level here.  is this something we're doing elsewhere?  i think for some clients this may be the right log level, others (more hypothetical ones?) will be hit by a lot of noise.  (Yes, this discussion also doesn't belong here, sorry :)

services/galley/src/Galley/API/MLS/Message.hs
```diff
@@ -532,6 +543,12 @@ postMLSMessageToRemoteConv loc qusr _senderClient con smsg rcnv = do
     MLSMessageResponseProtocolError e ->
       throw (mlsProtocolError e)
     MLSMessageResponseProposalFailure e -> throw (MLSProposalFailure e)
+    MLSMessageResponseUnreachableBackends ds ->
+      throw . InternalErrorWithDescription $
```
fisx, line 547: de-duplicate this particular internal error and give it a name?

fisx (approved):
looks good to me!  found some nit-picks.

services/galley/test/integration/API/MLS.hs
mdimjasevic: I entertained the idea, but many times I need at least three backends that federate, and support for testing that is still in the works.

This change in particular looks like simplifying an existing test, though.

libs/wire-api-federation/src/Wire/API/Federation/API/Galley.hs
mdimjasevic: > `Maybe (NonEmpty ..)` instead of just `[..]` is a bit silly, but ok. :) that discussion doesn't belong into this PR.

They are isomorphic. I did get some mileage out of my choice over the alternative, but nothing big.

mdimjasevic: > Also, can you explain in the haddocks how users can be unreachable, if not by their backend being unreachable? (If this is obvious to most readers let me know; I have very little context on MLS yet.)

This comes by extension: if a remote backend is unreachable, all of the users on that backend are unreachable too.

Are you confused by the fact I introduced two data constructors for seemingly the same thing ? In the first one, a commit is considered (therefore, an MLS protocol message) and it either fails (e.g., because some backends are unreachable) or succeeds (all backends are reachable); no partiality allowed. In the second one, an application message is considered and it is fine if it cannot be delivered to some users.

If you can clarify this, that would be great as that would help me understand what should be added to haddocks.

services/galley/src/Galley/API/Action.hs
mdimjasevic: I don't think we're consistent when it comes to logging, including something like this.

services/galley/src/Galley/API/MLS/Message.hs
mdimjasevic: ... as in to have a new data constructor for `InternalError` that is suited for this specific purpose?

mdimjasevic:
@fisx , I just merged the PR. However, I am still interested in hearing from you regarding my follow-ups to your questions. Once we sort that out, I'll be happy to make another PR to incorporate that too.

PR #3326, opened 2023-05-30
Add supported protocols

pcapriotti:
This adds a `supported_protocols` field to users. Clients can set it to any subset of `["proteus", "mls"]` using `PUT /self/supported-protocols`. There is also a new endpoint `GET /users/:domain/:id/supported-protocols`. The backend does not assign any semantics to this field, but it is intended to be used to coordinate migration to MLS across the clients of a user, as well as between two users participating in a 1-1 conversation.

https://wearezeta.atlassian.net/browse/FS-2025

## Checklist


services/brig/src/Brig/Data/Instances.hs
```diff
@@ -287,3 +287,10 @@ instance Cql SearchVisibilityInbound where
   fromCql (CqlInt 0) = pure SearchableByOwnTeam
   fromCql (CqlInt 1) = pure SearchableByAllTeams
   fromCql n = Left $ "Unexpected SearchVisibilityInbound: " ++ show n
+
+instance Cql (Imports.Set BaseProtocolTag) where
+  ctype = Tagged IntColumn
+
+  toCql = CqlInt . fromIntegral . protocolSetBits
+  fromCql (CqlInt bits) = pure $ protocolSetFromBits (fromIntegral bits)
```
mdimjasevic, line 295: If I'm reading this right, we simply ignore any bits other than the lowest two. E.g., if value 5 is read from the DB, this is fine and in particular we'll say that this means that Proteus is supported (because 5 is 101 in binary, and the lowest bit is set, which corresponds to Proteus).

mdimjasevic (approved):

services/brig/src/Brig/Data/Instances.hs
pcapriotti: That's right.

PR #3328, opened 2023-05-30
[chore] Add default suite to make ci

elland:
This should restore the previous behaviour of calling a bare `make ci`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -97,12 +97,10 @@ endif
 .PHONY: ci
 ci: c db-migrate
 ifeq ("$(package)", "all")
-	 make c
 	./hack/bin/cabal-run-integration.sh all
 	./hack/bin/cabal-run-integration.sh integration
 endif
 ifeq ("$(package)", "integration")
-	make c package=integration
 	./hack/bin/cabal-run-integration.sh integration
 else
 	make c package=$(package)
--- a/Makefile
+++ b/Makefile
@@ -100,11 +100,7 @@ ifeq ("$(package)", "all")
 	./hack/bin/cabal-run-integration.sh all
 	./hack/bin/cabal-run-integration.sh integration
 endif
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -98,30 +97,16 @@ endif
...
-    ifneq ("$(suite)", "new")
-		./hack/bin/cabal-run-integration.sh all
-    endif
-    ifneq ("$(suite)", "old")
-		make c package=integration
-		./hack/bin/cabal-run-integration.sh integration
-    endif
-else
-  ifeq ("$(package)", "integration")
+	 make c
```
smatting, line 100:
```suggestion
```
`ci: c db-migrate` Already implies `make c`.

Makefile
```diff
@@ -98,30 +97,16 @@ endif
...
+	./hack/bin/cabal-run-integration.sh all
+	./hack/bin/cabal-run-integration.sh integration
+endif
+ifeq ("$(package)", "integration")
+	make c package=integration
```
smatting, line 105:
```suggestion
```
Same here. 

I think probably all these `if`s can go away. It's a bit hard to say for sure. Can you please test the different targets?

smatting (changes requested):

Makefile
```diff
@@ -98,30 +97,14 @@ endif
...
-		./hack/bin/cabal-run-integration.sh integration
-      else
-		make c package=integration
-		./hack/bin/cabal-run-integration.sh $(package)
-		./hack/bin/cabal-run-integration.sh integration
-      endif
-    endif
-  endif
+else
+	make c package=$(package)
```
smatting, line 106: This would call `make c` for the package a second time. The first time would be because of the `ci: c` line. I think the whole if can go away

smatting (changes requested):

PR #3333, opened 2023-05-31
galley: Send on-user-deleted-conversations backend notification through RabbitMQ

akshaymankar:
https://wearezeta.atlassian.net/browse/WPB-200

Also includes some fixups for #3276 (all brig related changes are fixups)

## Checklist

### DIFF
--- a/charts/galley/values.yaml
+++ b/charts/galley/values.yaml
@@ -23,6 +23,11 @@ config:
     host: aws-cassandra
     replicaCount: 3
   enableFederation: false # keep enableFederation default in sync with brig and cargohold chart's config.enableFederation as well as wire-server chart's tags.federation
+  # Not used if enableFederation is false
+  rabbitmq:
+    host: rabbitmq
+    port: 5672
+    vHost: /
   settings:
     httpPoolSize: 128
     maxTeamSize: 10000
--- a/services/galley/src/Galley/Env.hs
+++ b/services/galley/src/Galley/Env.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE RecordWildCards #-}
 {-# LANGUAGE TemplateHaskell #-}
 
 -- This file is part of the Wire Server implementation.
@@ -31,6 +32,8 @@ import Galley.Options
 import qualified Galley.Queue as Q
 import HTTP2.Client.Manager (Http2Manager)
 import Imports
+import qualified Network.AMQP as Q
+import qualified Network.AMQP.Extended as Q
 import Network.HTTP.Client
 import Network.HTTP.Client.OpenSSL
 import OpenSSL.EVP.Digest
@@ -59,7 +62,8 @@ data Env = Env
     _deleteQueue :: Q.Queue DeleteItem,
     _extEnv :: ExtEnv,
     _aEnv :: Maybe Aws.Env,
-    _mlsKeys :: SignaturePurpose -> MLSKeys
+    _mlsKeys :: SignaturePurpose -> MLSKeys,
+    _rabbitmqChannel :: Maybe (MVar Q.Channel)
   }
 
 -- | Environment specific to the communication with external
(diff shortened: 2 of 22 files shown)


services/galley/src/Galley/Intra/BackendNotificationQueue.hs
```diff
@@ -0,0 +1,62 @@
+{-# LANGUAGE NumericUnderscores #-}
```
elland, line 1: This is in the cabal file now, right?

services/galley/test/integration/API.hs
```diff
@@ -4416,31 +4416,7 @@ removeUser = do
...
-      F.udcvUser udcnC @?= qUnqualified alexDel
-
-    liftIO $ do
-      dReq <- assertOne $ filter (matchFedRequest dDomain "on-user-deleted-conversations") fedRequests
-      frComponent dReq @?= Galley
-      frRPC dReq @?= "on-user-deleted-conversations"
-      Right udcnD <- pure . eitherDecode . frBody $ dReq
-      sort (fromRange (F.udcvConversations udcnD)) @?= sort [convD1]
-      F.udcvUser udcnD @?= qUnqualified alexDel
+      assertEqual ("expect exactly 4 federated requests in : " <> show fedRequests) 4 (length fedRequests)
```
fisx, line 4419: ok, the federated calls don't happen here any more, so we can't test for these events.  but should/can we test for something else?

related: a half-solution to the `MakesFederatedCall` tolerance for removing calls without removing the constraint could be that we add another kind of constraint `PushesFederatedNotifications`.  then we would get a type error if we forget to add the new constraint, and that would help us remember to remove the old one.  not sure that's worth it, though.

services/galley/src/Galley/Env.hs
```diff
@@ -104,3 +108,15 @@ currentFanoutLimit o = do
   let optFanoutLimit = fromIntegral . fromRange $ fromMaybe defFanoutLimit (o ^. (optSettings . setMaxFanoutSize))
   let maxTeamSize = fromIntegral (o ^. (optSettings . setMaxTeamSize))
   unsafeRange (min maxTeamSize optFanoutLimit)
+
+mkRabbitMqChannel :: Logger -> Opts -> IO (Maybe (MVar Q.Channel))
+mkRabbitMqChannel l (view optRabbitmq -> Just RabbitMqOpts {..}) = do
```
fisx, line 113: should this go to wire-api?  that would also pull a few other changes there, and clean up the arguments (`Opts` is not available in wire-api, but `RabbitMqOpts` should be).

services/galley/src/Galley/Intra/BackendNotificationQueue.hs
```diff
@@ -0,0 +1,62 @@
...
+    Nothing -> pure (Left FederationNotConfigured)
+    Just chanVar -> do
+      let policy = limitRetries 3 <> constantDelay 1_000_000
+          handlers =
+            skipAsyncExceptions
+              <> [logRetries (const $ pure True) logError]
+      Right <$> recovering policy handlers (const $ go ownDomain chanVar)
+  where
+    logError willRetry (SomeException e) status = do
+      Log.err $
```
fisx, line 47: This is something that may well happen every now and then without being an error.  eg., if only one backend-pusher is running and needs to be restarted, or during startup.  Should the log level be `info`, and only `error` once the retry count has been exhausted?

fisx (approved):
LGTM!  just nit-picks.

charts/galley/values.yaml
```diff
@@ -23,6 +23,11 @@ config:
     host: aws-cassandra
     replicaCount: 3
   enableFederation: false # keep enableFederation default in sync with brig and cargohold chart's config.enableFederation as well as wire-server chart's tags.federation
+  # Not used if enableFederation is false
```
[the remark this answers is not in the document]
akshaymankar: It is not an error to set the helm value, it just gets ignored. If we want to make it an error it would require us to remove this from here and put it in some docs. Which would make it less discoverable.

services/galley/test/integration/API.hs
akshaymankar: > ok, the federated calls don't happen here any more, so we can't test for these events. but should/can we test for something else?

This functionality is already tested in federation end-to-end tests. If we _really_ want to do something we'd have to create a rabbitmq vhost (its like a namespace) and pass that to our test and then we can assert that there is something in the queue. But honestly, seems a bit redundant given there are other test. I am making sure that I don't delete tests without making sure that there are other tests.

akshaymankar: > related: a half-solution to the MakesFederatedCall tolerance for removing calls without removing the constraint could be that we add another kind of constraint PushesFederatedNotifications. then we would get a type error if we forget to add the new constraint, and that would help us remember to remove the old one. not sure that's worth it, though.

I thought the point behind adding `MakesFederatedCall` was to ensure that possibility of federation errors shows up in swagger. If we fail to connect to RabbitMQ, it is just a 500 because the backend is not well. Do you think it would help swagger somehow that we add this new annotation? Do the clients need to care that this gets send via RabbitMQ?

services/galley/src/Galley/Env.hs
akshaymankar: I think `wire-api` is the wrong place to put it as this has nothing to do with the API. I was thinking of shoving it in the `exteded` package, but forgot about it. Let's do it in the next PR?

services/galley/src/Galley/Intra/BackendNotificationQueue.hs
akshaymankar: This can only happen when RabbitMQ is down or unreachable. So, this is definitely an error.

PR #3344, opened 2023-06-09
add conversation type to group ID serialisation

stefanwire:
## Checklist


pcapriotti (approved):
Looks good. Maybe one thing I would change is make the internal endpoint take a group ID instead of its parts. Then we wouldn't need to add bytestring conversion functions for `ConvType`.

PR #3351, opened 2023-06-13
k8s 1.26/1.27 compatibility: podsecuritycontext

jschaul:
See https://wearezeta.atlassian.net/browse/WPB-1234

This PR adds a `securityContext` to all (wire-server) containers. This builds upon #3352 and enforces some pod security policies everywhere. (no change is done for third-party helm charts which we just wrap).

Drive-by fixes:

* Fixes a deprecation warning about ingresses and k8s >= 1.27
* makes `helmfile destroy` as used by CI a little faster
* updates some alpine images

## Checklist

### DIFF
--- a/changelog.d/5-internal/security-context
+++ b/changelog.d/5-internal/security-context
@@ -1,1 +1,1 @@
-All wire-server containers now run in a restricted securityContext
+All wire-server containers now run in a restricted securityContext when run on k8s >= 1.24


jschaul:
Integration tests ran successfully on k8s 1.27, too (I ran `make kube-integration` against that cluster after setting the right `DOCKER_TAG`).

flokli (approved):

jschaul:
Some of this securityContext stuff is actually not compatible with old k8s (like 1.19), so this got now wrapped in conditional helm logic and is only active on kubernetes 1.24 or above.

PR #3356, opened 2023-06-15
Change MessagePush interface: Multiple recipients, single request

smatting:
This PR changes the `MessagePush` interface between galley and gundeck.

Without this PR: Multiple `Push` objects are abstracted in a single `MessagePush` object.
With this PR: A `MessagePush` object abstracts a single Push object with multiple recipients.

For Proteus' `sendLocalMessages` where each message payload is distinct for each recipient this PR
has the effect that for each recipient a single `Push` from galley to gundeck is sent. Without this PR
all Pushes are sent in a single request.

This PR is preparation for MLS where the message payload is the same for all recipients in a group.

This PR also contains updates to the python scripts to test performance with large MLS groups.


pcapriotti (approved):
LGTM

PR #3359, opened 2023-06-16
[WPB-1085] Two-event member adding process during conversation creation

mdimjasevic:
*This PR is a carved out part of #3324 that relates to conversation creation only.*

When a Proteus conversation is created and remote members are added to it as part of the request to `POST /conversations`, remote clients now get two events: a `conversation.create` event as before with a subset of other members, and also a `conversation.member-join` event once the list of all remote members is confirmed. Local users get only the `conversation.create` event as before with the full list of members.

For details see https://wearezeta.atlassian.net/wiki/spaces/CORE/pages/751435862/Offline+backend+handling+-+analysis+of+creating+a+conversation#Events-for-joining-clients.

Tracked by https://wearezeta.atlassian.net/browse/WPB-1085.

## Checklist


services/galley/src/Galley/Data/Types.hs
```diff
@@ -99,3 +101,9 @@ data LockAcquired
   = Acquired
   | NotAcquired
   deriving (Show, Eq)
+
+data MemberAddStatus = MemberAddStatus
```
fisx, line 105: _note to self: please add this comment to the docs and make sure it's considered in the logic._

we can't just fail to the requesting client with "the following domains could not be reached", because a remote site may go down after adding some of its members to a conversation, but before adding the rest.  so clients need to be informed about failures of adding individual members.

`MemberAddStatus` is confusing me.  i don't think we're using `added` anywhere in either of these two PRs.  maybe this is about the above, and `added` should be renamed to `notNotified`?

services/galley/src/Galley/API/Util.hs
```diff
@@ -759,25 +754,78 @@ fromConversationCreated loc rc@ConversationCreated {..} =
...
-    fedClient @'Galley @"on-conversation-created" rc
+  Local Data.Conversation ->
+  Sem r DataTypes.MemberAddStatus
+registerRemoteConversationMemberships now lc = do
+  let c = tUnqualified lc
+      allRemoteMembers = nubOrd (Data.convRemoteMembers c)
+      allRemoteMembersQualified = remoteMemberQualify <$> allRemoteMembers
+      allRemoteBuckets :: [Remote [RemoteMember]] = bucketRemote allRemoteMembersQualified
+      rc = toConversationCreated now c
+  failedToNotify <- fmap (foldMap (either (sequenceA . fst) mempty)) $
```
fisx, line 771: this looks weird.

the type of the second argument of this `fmap` is `[Either (Remote [RemoteMember], FederationError) (Remote ())]`.

- shouldn't we keep the buckets with the different remotes, rather than filtering out again below in line 786?
- why does the `FederationError` have to be the same value for all remote members?  shouldn't it be allowed to differ between members?

services/galley/src/Galley/Data/Types.hs
```diff
@@ -99,3 +101,7 @@ data LockAcquired
   = Acquired
   | NotAcquired
   deriving (Show, Eq)
+
+-- | Set of users that could not be notified of their addition to a new conversation, and
+-- should be removed from the conv for consistenty.
```
mdimjasevic, line 106:
```suggestion
-- should be removed from the conv for consistency.
```

services/galley/src/Galley/Data/Types.hs
mdimjasevic: There are two parts to your comment.

I believe we addressed the first one: the federation API does not work at this level of granularity, so there's no incremental update to the conversation-owning backend on the progress of member adding on a remote backend's end.

As to the second part regarding `MemberAddStatus`, I believe you've already addressed that with your changes.

I'm marking this comment as resolved, but revert that if there's more to it.

services/galley/src/Galley/API/Util.hs
mdimjasevic: The comment is referring to old code so I'm not sure my reply will address your concerns, but here's an attempt:

1. Regarding keeping them in buckets, there are a few variations of the same data, but in different types, each suitable for a slightly different need. We do have buckets around too.
2. The same `FederationError` question is, I believe, addressed by the observation I made yesterday during the call: there's no incremental update from a remote backend on the progress of processing member updates, so either all of members are added successfully or there is a failure, e.g., a `FederationError` is returned, in which case none of the remote backend's users are considered to be conversation members.

mdimjasevic:
@fisx and @battermann , I've merged the latest `develop` into the PR and cleaned up the PR a bit (fixed a few typos, made tiny changes to haddock, and removed the unused `EventWithUnreachables` type). I believe it's good to be merged now.

Let me know if you have any further feedback, remarks, or questions.

services/galley/src/Galley/API/Util.hs
```diff
@@ -759,25 +752,91 @@ fromConversationCreated loc rc@ConversationCreated {..} =
...
+      rc = toConversationCreated now c
+
+      allRemoteMembers = nubOrd {- (but why would there be duplicates?) -} (Data.convRemoteMembers c)
+      allRemoteMembersQualified = remoteMemberQualify <$> allRemoteMembers
+      allRemoteBuckets :: [Remote [RemoteMember]] = bucketRemote allRemoteMembersQualified
+
+  failedToNotify :: [Remote RemoteMember] <- fmap (foldMap (either (sequenceA . fst) mempty)) $
+    runFederatedConcurrentlyEither allRemoteMembersQualified $
+      \rrms ->
+        fedClient @'Galley @"on-conversation-created" $
```
battermann, line 774:
```suggestion
        fedClient @'Galley @"on-conversation-created"
```

battermann (approved):
LGTM

PR #3364, opened 2023-06-19
Remove user & client thresholds from MLS migration

stefanwire:
## Checklist


libs/wire-api/src/Wire/API/Team/Feature.hs
```diff
@@ -1077,25 +1067,20 @@ instance ToSchema MlsMigrationConfig where
...
         )
         checkConfig
     where
       checkConfig c = do
         when
-          ( isNothing c.finaliseRegardlessAfter
-              && isNothing c.usersThreshold
-              && isNothing c.clientsThreshold
-          )
+          (isNothing c.finaliseRegardlessAfter)
```
pcapriotti, line 1075: We don't need this check anymore. `finaliseRegardlessAfter` can be empty.

services/galley/schema/src/V85_DropUserClientThresholdFromTeamFeatureMlsMigration.hs
```diff
@@ -0,0 +1,34 @@
...
+  ( migration,
+  )
+where
+
+import Cassandra.Schema
+import Imports
+import Text.RawString.QQ
+
+migration :: Migration
+migration = Migration 85 "Remove mls_migration_users_threshold and mls_migration_clients_threshold from feature config for team feature MLS Migration" $ do
```
pcapriotti, line 28: Why not simply modify V84? The whole migration feature only exists in the mls branch and was never deployed anywhere important.

pcapriotti (approved):

PR #3365, opened 2023-06-19
[WPB-2789] Use EmptyResponse consistently across Galley federation endpoints

elland:
## Checklist


fisx (approved):
i think there may also be some "NoContent" cases that you want to change, but i suppose that's not a breaking change, so doesn't have to happen here & now.

services/galley/test/integration/API/Federation.hs
```diff
@@ -511,8 +511,7 @@ notifyUpdateUnavailable extras action etype edata = do
   WS.bracketR2 c alice charlie $ \(wsA, wsC) -> do
     ((), _fedRequests) <-
       withTempMockFederator' (throw $ MockErrorResponse Http.status500 "Down for maintenance") $
-        void $
-          runFedClient @"on-conversation-updated" fedGalleyClient bdom cu
+          void $ runFedClient @"on-conversation-updated" fedGalleyClient bdom cu
```
fisx, line 514: i don't think ormolu will leave that alone :)

PR #3367, opened 2023-06-19
Test swagger docs.

fisx:
## Checklist

### DIFF
--- a/integration/test/API/Brig.hs
+++ b/integration/test/API/Brig.hs
@@ -216,14 +216,14 @@ putUserSupportedProtocols user ps = do
 getApiVersions :: HasCallStack => App Response
 getApiVersions = do
   req <-
-    localRawBaseRequest Brig Unversioned $
+    rawBaseRequest OwnDomain Brig Unversioned $
       joinHttpPath ["api-version"]
   submit "GET" req
 
 getSwaggerPublicTOC :: HasCallStack => App Response
 getSwaggerPublicTOC = do
   req <-
-    localRawBaseRequest Brig Unversioned $
+    rawBaseRequest OwnDomain Brig Unversioned $
       joinHttpPath ["api", "swagger-ui"]
   submit "GET" req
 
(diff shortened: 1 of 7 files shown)


integration/test/Testlib/HTTP.hs
```diff
@@ -101,6 +102,9 @@ onFailureAddResponse r m = App $ do
 
 data Versioned = Versioned | Unversioned | ExplicitVersion Int
 
+localRawBaseRequest :: HasCallStack => Service -> Versioned -> String -> App HTTP.Request
+localRawBaseRequest = rawBaseRequest OwnDomain
```
pcapriotti, line 106: What's the point of this definition? I'd rather see `rawBaseRequest OwnDomain` than `localRawBaseRequest` and have to look up what that is :shrug:.

integration/test/API/Brig.hs
```diff
@@ -212,3 +212,48 @@ putUserSupportedProtocols user ps = do
...
+
+getSwaggerPublicTOC :: HasCallStack => App Response
+getSwaggerPublicTOC = do
+  req <-
+    localRawBaseRequest Brig Unversioned $
+      joinHttpPath ["api", "swagger-ui"]
+  submit "GET" req
+
+getSwaggerInternalTOC :: HasCallStack => App Response
+getSwaggerInternalTOC = error "FUTUREWORK: this API end-point does not exist."
```
pcapriotti, line 231: Doesn't this simply crash? Maybe turn it into a comment.

pcapriotti (approved):

integration/test/API/Brig.hs
fisx: it's meant to crash.  nobody should call this function before it is doing anything interesting.  also, crashing isn't that bad in a test suite.  just stop calling this function, to resolve it?

integration/test/Testlib/HTTP.hs
fisx: the idea was that somebody who doesn't know about `OwnDomain` can find this function and will understand that it's the thing they were looking for.  fewer questions from stupid developers you'll have to answer in the future.  :)

but no strong opinion, i'll remove it.  :+1:

integration/test/API/Brig.hs
pcapriotti: You're right, sorry, I thought this was a test. My mistake.

PR #3372, opened 2023-06-22
Add combinator for maps with arbitrary keys

pcapriotti:
A new combinator to create schemas for objects with arbitrary keys, but where the key type does not necessarily have the right instances for JSON keys. Instead, this combinator takes conversion functions to and from 'Text'.

## Checklist


fisx (approved):
tried this and it works!  thanks!

libs/schema-profunctor/src/Data/Schema.hs
```diff
@@ -490,6 +496,19 @@ map_ sch = mkSchema d i o
     i = A.parseJSON >=> traverse (schemaIn sch)
     o = fmap A.toJSON . traverse (schemaOut sch)
 
+-- | A schema for a JSON object with arbitrary keys of type 'k', where 'k' can
+-- be converted to and from 'Text.
```
pcapriotti, line 500:
```suggestion
-- be converted to and from 'Text'.
```

fisx (approved):

PR #3375, opened 2023-06-23
multi ingress ingress services

supersven:
Render additional K8s objects for additional backend domains ("multi-ingress").

The best entry to this PR is the `values.yaml` file which shows the new configuration settings: https://github.com/wireapp/wire-server/pull/3375/files#diff-6752e60cbd59fb7f3c26228bae9955905976750f391ce424ac849ce0450432abR45

## Checklist

### DIFF
--- a/charts/nginx-ingress-services/templates/ingress.yaml
+++ b/charts/nginx-ingress-services/templates/ingress.yaml
@@ -9,6 +9,10 @@ metadata:
     {{- if not $ingressFieldNotAnnotation }}
     kubernetes.io/ingress.class: "{{ .Values.config.ingressClass }}"
     {{- end }}
+    {{ if .Values.config.renderCSPInIngress }}
+    {{ if ne .Values.config.ingressClass "nginx"}}
+        {{ fail "In ingress CSP header setting only works with 'nginx' controller." }}
+    {{ end }}
     nginx.ingress.kubernetes.io/configuration-snippet: |
       set $CSP "connect-src 'self' blob: data: https://*.giphy.com https://{{ .Values.config.dns.https }}";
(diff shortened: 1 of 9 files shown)


charts/nginx-ingress-services/templates/ingress.yaml
```diff
@@ -4,11 +4,38 @@
...
 metadata:
-  name: nginx-ingress
+  name: nginx-ingress-{{ .Values.ingressName }}
   annotations:
     {{- if not $ingressFieldNotAnnotation }}
     kubernetes.io/ingress.class: "{{ .Values.config.ingressClass }}"
     {{- end }}
+    {{ if .Values.config.renderCSPInIngress }}
+    {{ if ne .Values.config.ingressClass "nginx"}}
+        {{ fail "In ingress CSP header setting only works with 'nginx' controller." }}
```
jschaul, line 14: If this were changed in CI integration tests, we'd run into a problem, as in 

```
hack/helm_vars/nginx-ingress-controller/values.yaml.gotmpl
6:    ingressClass: "nginx-{{ .Release.Namespace }}"
```

We set the ingress class to something different matching the ingress controller. Can the condition be changed to be "if "nginx" is a substring of ....Values.ingressClass" (however the syntax for that is in gotemplate, so as not to conflict here, were we to test this in integration tests?

jschaul (approved):

charts/nginx-ingress-services/templates/ingress.yaml
supersven: I've added a check if the `ingressClass` contains `"nginx"`. Feels a bit hand-wavy, but I don't have a better idea / better field to check.

PR #3378, opened 2023-06-26
FS-1179/WPB-1107: Clean local state when stopping to federate with another backend

lepsa:
https://wearezeta.atlassian.net/browse/WPB-1107

## Checklist

## Changes
Removing conversations upon a defederation request.
- Adding indexes to remote conversation tables so we can search on the remote domain.
- Updating background worker to listen to a RabbitMQ defederation queue and kickoff defederation cleanup.
- Adding internal APIs to Brig and Galley to facilitate defederation. Galley is called by background worker, and Brig is called by Galley for DB access reasons.
- Updating Brig to drop notification queues for deleted domains.
- Updating background worker to pick up on federation domain list changes when pushing notifications.


lepsa:
There is an issue with the tests after merging in develop and migrating what I had to hspec. The integration binary no longer runs in `make ci` as it cannot handle the command args being passed in.

background-worker-integration: unrecognized option `-s'

However going to the service directory and running `cabal exec background-worker-integration` will run the tests that do exist.

This could be resolved by removing the background-worker-integration section from the cabal file, or allowing it to handle the extra arguments so it can work with the `make ci` tests.

lepsa:
We also need to look into if we are happy to let this service die on exceptions and have kubernetes restart it for us. These tasks should be mostly idempotent and rabbit will track what still needs to be done.

changelog.d/6-federation/fs-1179
```diff
@@ -0,0 +1 @@
+Removing a federation domain will now remove all conversations for that domain.
```
fisx, line 1: maybe add an entry in 5-internal that explains how background-worker had to be rewired to allow for proper cleanup of de-federated domains?  is there other rewiring happening in this PR?

fisx:
> There is an issue with the tests after merging in develop and migrating what I had to hspec. The integration binary no longer runs in `make ci` as it cannot handle the command args being passed in.
> 
> background-worker-integration: unrecognized option `-s'
> 
> However going to the service directory and running `cabal exec background-worker-integration` will run the tests that do exist.
> 
> This could be resolved by removing the background-worker-integration section from the cabal file, or allowing it to handle the extra arguments so it can work with the `make ci` tests.



> There is an issue with the tests after merging in develop and migrating what I had to hspec. The integration binary no longer runs in `make ci` as it cannot handle the command args being passed in.
> 
> background-worker-integration: unrecognized option `-s'
> 
> However going to the service directory and running `cabal exec background-worker-integration` will run the tests that do exist.
> 
> This could be resolved by removing the background-worker-integration section from the cabal file, or allowing it to handle the extra arguments so it can work with the `make ci` tests.

fixed here: c90889c62f9abf846ba93b949d430f3c6ddb89a2

it's still weird that `Spec` is added to two cabal goals: the integration executable and the test suite.  but `make ci package=background-worker` runs fine again.

lepsa:
> > There is an issue with the tests after merging in develop and migrating what I had to hspec. The integration binary no longer runs in `make ci` as it cannot handle the command args being passed in.
> > background-worker-integration: unrecognized option `-s'
> > However going to the service directory and running `cabal exec background-worker-integration` will run the tests that do exist.
> > This could be resolved by removing the background-worker-integration section from the cabal file, or allowing it to handle the extra arguments so it can work with the `make ci` tests.
> 
> > There is an issue with the tests after merging in develop and migrating what I had to hspec. The integration binary no longer runs in `make ci` as it cannot handle the command args being passed in.
> > background-worker-integration: unrecognized option `-s'
> > However going to the service directory and running `cabal exec background-worker-integration` will run the tests that do exist.
> > This could be resolved by removing the background-worker-integration section from the cabal file, or allowing it to handle the extra arguments so it can work with the `make ci` tests.
> 
> fixed here: [c90889c](https://github.com/wireapp/wire-server/commit/c90889c62f9abf846ba93b949d430f3c6ddb89a2)
> 
> it's still weird that `Spec` is added to two cabal goals: the integration executable and the test suite. but `make ci package=background-worker` runs fine again.

Galley and Brig both have a similar setup, with a test suite and integration executable defined in cabal. I didn't check other services.
These test-suites are used by `cabal-run-tests.sh` and that script is called from `make c test=1`. The integration binary is for `make ci`, so in theory you can have different test suites running different things. This doesn't seem to see much use however, as galley only has 33 unit tests vs the 363 integration tests.

services/background-worker/background-worker.cabal
```diff
@@ -148,6 +159,96 @@ executable background-worker
     UndecidableInstances
     ViewPatterns
 
+executable background-worker-integration
```
fisx, line 162: ah, no, wait.  please don't introduce any new legacy integration test suites.  we're using /integration for integration tests from now on; `/services/<package>/test/integration` is only still there because it takes time and effort to migrate.

another thing: previously the unit tests and integration tests should be in separate directories, and do separate things.  like in `/services/brig/test/{unit,integration}`.  only the `integration` part is now legacy, and we use `/integration` with a new approach instead.  unit tests in any package are fine, of course.

please pair when you're using `/integration` for the first time!  it has a readme, but it's a bit dense.

services/brig/src/Brig/API/Internal.hs
```diff
@@ -1,3 +1,5 @@
+{-# LANGUAGE TypeApplications #-}
```
fisx, line 1: this should be in the cabal file as language defaults.  if it's not, please add it there.

services/background-worker/background-worker.cabal
lepsa: I'll do some prep reading before you come back online.

services/brig/src/Brig/API/Internal.hs
lepsa: I've removed that. I've also noticed that brig has a lot of redundant language extensions, and some extensions that are used a lot like GeneralisedNewtypeDeriving. Maybe a conversation should be had about enabling them by default?

services/background-worker/background-worker.cabal
fisx: what happened to this?

services/galley/test/integration/Federation.hs
```diff
@@ -0,0 +1,374 @@
+{-# LANGUAGE RecordWildCards #-}
+{-# OPTIONS_GHC -Wno-unused-matches #-}
+
+module Federation where
```
fisx, line 4: this should also not have gone here.  let's move the background-worker suite first and see how painful it is, though.

lepsa:
> > There is an issue with the tests after merging in develop and migrating what I had to hspec. The integration binary no longer runs in `make ci` as it cannot handle the command args being passed in.
> > background-worker-integration: unrecognized option `-s'
> > However going to the service directory and running `cabal exec background-worker-integration` will run the tests that do exist.
> > This could be resolved by removing the background-worker-integration section from the cabal file, or allowing it to handle the extra arguments so it can work with the `make ci` tests.
> 
> > There is an issue with the tests after merging in develop and migrating what I had to hspec. The integration binary no longer runs in `make ci` as it cannot handle the command args being passed in.
> > background-worker-integration: unrecognized option `-s'
> > However going to the service directory and running `cabal exec background-worker-integration` will run the tests that do exist.
> > This could be resolved by removing the background-worker-integration section from the cabal file, or allowing it to handle the extra arguments so it can work with the `make ci` tests.
> 
> fixed here: [c90889c](https://github.com/wireapp/wire-server/commit/c90889c62f9abf846ba93b949d430f3c6ddb89a2)
> 
> it's still weird that `Spec` is added to two cabal goals: the integration executable and the test suite. but `make ci package=background-worker` runs fine again.

This turned out to be a red-herring based on my lack of understanding about what `make ci` was intended to do, and not knowing about `make c test=1`. These were never meant to be unit tests, I erroneously forced them to run along with the integration tests.

elland (approved):

PR #3379, opened 2023-06-26
Avoid MLS messages in sender client's event stream

stefanwire:
## Checklist

### DIFF
--- a/integration/test/Test/MLS.hs
+++ b/integration/test/Test/MLS.hs
@@ -28,9 +28,16 @@ testSendMessageNoReturnToSender = do
     for_ wss $ \ws -> do
       n <- awaitMatch 3 (\n -> nPayload n %. "type" `isEqual` "conversation.mls-message-add") ws
       nPayload n %. "data" `shouldMatch` T.decodeUtf8 (Base64.encode mp.message)
-    expectFailure (const $ pure ()) $ do
-      n <- awaitMatch 3 (\n -> nPayload n %. "type" `isEqual` "conversation.mls-message-add") wsSender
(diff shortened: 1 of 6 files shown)


pcapriotti (approved):
Looks good. Although maybe it would be good to have a test that the sender client is indeed not receiving these messages. And maybe some comments for our future selves explaining that this behaviour is indeed pretty strange, but it's what the clients want.

integration/test/Test/MLS.hs
```diff
@@ -11,6 +11,27 @@ import MLS.Util
...
+  withWebSockets [alice1, alice2, bob1, bob2] $ \(wsSender : wss) -> do
+    mp <- createApplicationMessage alice1 "hello, bob"
+    void . bindResponse (postMLSMessage mp.sender mp.message) $ \resp -> do
+      resp.status `shouldMatchInt` 201
+    for_ wss $ \ws -> do
+      n <- awaitMatch 3 (\n -> nPayload n %. "type" `isEqual` "conversation.mls-message-add") ws
+      nPayload n %. "data" `shouldMatch` T.decodeUtf8 (Base64.encode mp.message)
+    expectFailure (const $ pure ()) $ do
+      n <- awaitMatch 3 (\n -> nPayload n %. "type" `isEqual` "conversation.mls-message-add") wsSender
+      nPayload n %. "data" `shouldMatch` T.decodeUtf8 (Base64.encode mp.message)
```
pcapriotti, line 33: I don't think we want this assertion here. If we get an unexpected extra message in there, the test will flake. Either make the last check part of the `awaitMatch` predicate, or simply remove it.

stefanwire: Made it part of `awaitMatch`.

PR #3396, opened 2023-07-05
replace all instances of example.com with default.domain

arthurwolf:
manually going over all of them now.

(note: See WPB-2621 )

## Checklist


smatting (changes requested):
Replacing all instances will not work. We use `example.com` `b.example.com` `c*.example.com` as domains for our integration tests. We can't change them easily.

Maybe change only in `docs/`?

flokli (changes requested):
Also NAK on the general direction of this PR.

https://www.rfc-editor.org/rfc/rfc2606.html describes what domain names or TLDs to use for testing/documentation examples, and example.com is among them, while default.domain certainly is not.

arthurwolf:
Re-tooling this PR: it will now be only about modifications to the docs, there will be a separate ticket (and maybe PR later) for deeper work.

arthurwolf:
After internal discussions, also changing the demain to "wire.example"

rohan-wire:
> Also NAK on the general direction of this PR.
> 
> https://www.rfc-editor.org/rfc/rfc2606.html describes what domain names or TLDs to use for testing/documentation examples, and example.com is among them, while default.domain certainly is not.

a.example, b.example, c.example, a.test, b.test, are all appropriate domains for documentation.

I thought we agreed that backends with no local domain name should be configured with `local.invalid`?

PR #3397, opened 2023-07-05
WPB-240: Generate and fan out events about stopping to federate

lepsa:
This adds the initial client notifications for the local server de-federating from a remote server. Remote servers de-federating from each other are not covered in this PR.

## Checklist


services/galley/src/Galley/API/Internal.hs
```diff
@@ -543,8 +553,35 @@ internalDeleteFederationDomainH ::
...
+  -- Once before and once after defederation work.
+  -- https://wearezeta.atlassian.net/wiki/spaces/ENGINEERIN/pages/809238539/Use+case+Stopping+to+federate+with+a+domain
+  void sendNotifications
   deleteFederationDomain domain
+  void sendNotifications
   pure (empty & setStatus status200)
+  where
+    sendNotifications = do
+      maxPage <- inputs $ fromRange . currentFanoutLimit . _options -- This is based on the limits in removeIfLargeFanout
+      page <- embedClient $ paginate selectAllMembers (paramsP LocalQuorum () maxPage)
```
elland, line 566: Feels like this should be an effect, instead of accessed directly, but I'm not sure what we want to do.

services/galley/src/Galley/API/Internal.hs
```diff
@@ -543,8 +553,35 @@ internalDeleteFederationDomainH ::
...
+      sendNotificationPage page
```
elland, line 567: Shouldn't we void here instead of the call sites?

services/galley/src/Galley/API/Internal.hs
```diff
@@ -543,8 +553,35 @@ internalDeleteFederationDomainH ::
...
+    pushEvents results = do
+      let (bots, mems) = localBotsAndUsers results
+          recipients = Intra.recipient <$> mems
+          event = Intra.FederationEvent $ Federation.Event Federation.FederationDelete domain
+      for_ (Intra.newPush ListComplete Nothing event recipients) $ \p -> do
+        -- TODO: Transient or not?
```
elland, line 573: TODO or Futurework?

elland (commented):
Just a few open questions, but also wondering if the tests shouldn't be done in `/integration` instead.

services/galley/src/Galley/API/Internal.hs
lepsa: I've moved the void in front of this call.

services/galley/src/Galley/API/Internal.hs
lepsa: I've made a new effect for this.

elland (approved):

PR #3402, opened 2023-07-05
Re-revert: multi ingress ingress services

supersven:
Brings back the changes reverted with https://github.com/wireapp/wire-server/pull/3399 (which reverted https://github.com/wireapp/wire-server/pull/3375 .)

Add backwards compatibility to existing deployments by ensuring that the existing ingress is replaced when the default settings are applied. This happens by using the same name.
(Helm isn't smart enough to understand that the ingress had a changed name.)

## Checklist

### DIFF
--- a/charts/nginx-ingress-services/templates/issuer.yaml
+++ b/charts/nginx-ingress-services/templates/issuer.yaml
@@ -6,7 +6,7 @@ kind: "{{ .Values.tls.issuer.kind }}"
 {{- fail (cat ".tls.issuer.kind can only be one of Issuer or ClusterIssuer, got: " .tls.issuer.kind )}}
 {{- end }}
 metadata:
-  name: {{ .Values.tls.issuer.name }}
+  name: {{ .Values.tls.issuer.name }}-{{ .Values.color }}
   {{- if eq .Values.tls.issuer.kind "Issuer" }}
   namespace: {{ .Release.Namespace }}
   {{- end }}
--- a/charts/nginx-ingress-services/templates/issuer.yaml
+++ b/charts/nginx-ingress-services/templates/issuer.yaml
@@ -6,7 +6,7 @@ kind: "{{ .Values.tls.issuer.kind }}"
 {{- fail (cat ".tls.issuer.kind can only be one of Issuer or ClusterIssuer, got: " .tls.issuer.kind )}}
 {{- end }}
 metadata:
-  name: {{ .Values.tls.issuer.name }}-{{ .Values.color }}
+  name: {{ .Values.tls.issuer.name }}-{{ .Values.ingressName }}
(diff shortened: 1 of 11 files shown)


charts/nginx-ingress-services/templates/issuer.yaml
```diff
@@ -6,7 +6,7 @@ kind: "{{ .Values.tls.issuer.kind }}"
 {{- fail (cat ".tls.issuer.kind can only be one of Issuer or ClusterIssuer, got: " .tls.issuer.kind )}}
 {{- end }}
 metadata:
-  name: {{ .Values.tls.issuer.name }}
+  name: {{ include "nginx-ingress-services.getIssuerName" . | quote }}
```
akshaymankar, line 9: The issuerName is derived from ingressName, but the name of the secret for storing the account key is not derived from it. Would this work if someone actually tries to use different ACME servers?

supersven: Very good question!

You're referring to `getCustomSolversSecretName`, right?

```go
{{/*
Generate the secret name in a conistent way, since it's referred to in multiple places, while
at the same time being used for distinct scenarios
*/}}
{{- define "nginx-ingress-services.getCustomSolversSecretName" -}}
{{- $nameParts := list (include "nginx-ingress-services.fullname" .) -}}
{{- $nameParts = append $nameParts "cert-manager-custom-solvers" -}}
{{- join "-" $nameParts -}}
{{- end -}}
```

This should work, because its name depends on the name of the Helm release `nginx-ingress-services.fullname`. I.e. there is one Helm release for each domain. E.g. (from `mobtown-test`):

```yaml
...
  - name: 'nginx-ingress-services-red'
    namespace: 'wire'
    chart: 'ingress-branch/nginx-ingress-services'
    version: {{ .Values.wireChartVersion | quote }}
    values:
      - './helm_vars/nginx-ingress-services/values-red.yaml.gotmpl'
    secrets:
      - './helm_vars/nginx-ingress-services/secrets.yaml'

  - name: 'nginx-ingress-services-green'
    namespace: 'wire'
    chart: 'ingress-branch/nginx-ingress-services'
    version: {{ .Values.wireChartVersion | quote }}
    values:
      - './helm_vars/nginx-ingress-services/values-green.yaml.gotmpl'
    secrets:
      - './helm_vars/nginx-ingress-services/secrets.yaml'
...
```

I'm not sure if I understood your question correctly. Please let me know if my answer is non-sense ... :smile_cat:

akshaymankar: I am referring to `spec.acme.privateKeySecretRef` in this file. I think If someone installs this chart twice both instances will use the same secret in K8s and if they are talking to separate acme servers, I am not sure if it will work.

akshaymankar (approved):
:rocket:

PR #3401, opened 2023-07-05
Tweak retry policy for pulling federation remotes from brig.

fisx:
https://wearezeta.atlassian.net/browse/WPB-3007

timing of `./services/run-services date`:

- without fix, with federation enabled: `real 0m7.475s`
- without fix, with federation disabled: `real 0m3.912s`
- with fix, with federation enabled: `real 0m2.934s`
- with fix, with federation disabled: `real 0m1.659s`

## Checklist

 - [x] Add a new entry in an appropriate subdirectory of `changelog.d`
 - [x] Read and follow the [PR guidelines](https://docs.wire.com/developer/developer/pr-guidelines.html)

PR #3404, opened 2023-07-06
Removing domain sync threads where not currently used.

lepsa:
- Remove the domain sync thread from galley and gundeck.
- Remove the brig config value from gundeck.

## Checklist

### DIFF
--- a/changelog.d/0-release-notes/domain-sync-thread-cleanup
+++ b/changelog.d/0-release-notes/domain-sync-thread-cleanup
@@ -1,0 +1,1 @@
+Removed brig configuration value from gundeck.
--- a/libs/wire-api/src/Wire/API/FederationUpdate.hs
+++ b/libs/wire-api/src/Wire/API/FederationUpdate.hs
@@ -32,9 +32,9 @@ syncFedDomainConfigs (Endpoint h p) log' cb = do
 -- | Initial function for getting the set of domains from brig, and an update interval
 initialize :: L.Logger -> ClientEnv -> IO FederationDomainConfigs
 initialize logger clientEnv =
-  let -- keep trying every 3s for one minute
+  let -- An initial value of 100000 is used throughout services. Another commonly used value is 50000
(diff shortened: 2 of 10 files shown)


libs/wire-api/src/Wire/API/FederationUpdate.hs
```diff
@@ -32,9 +32,9 @@ syncFedDomainConfigs (Endpoint h p) log' cb = do
 -- | Initial function for getting the set of domains from brig, and an update interval
 initialize :: L.Logger -> ClientEnv -> IO FederationDomainConfigs
 initialize logger clientEnv =
-  let -- keep trying every 3s for one minute
+  let -- An initial value of 100000 is used throughout services. Another commonly used value is 50000
       policy :: R.RetryPolicy
-      policy = R.constantDelay 3_081_003 <> R.limitRetries 20
+      policy = R.exponentialBackoff 100000 <> R.limitRetries 20
```
akshaymankar, line 37: I think we should always add a cap to the exponential backoff, so we don't get crazy wait times like 1 hour.

lepsa: Luckily Matthias has a PR for exactly that! https://github.com/wireapp/wire-server/pull/3401

changelog.d/0-release-notes/domain-sync-thread-cleanup
```diff
@@ -0,0 +1 @@
+Removed brig configuration value from gundeck.
```
fisx, line 1:
```suggestion
Removed federation remote sync look from gundeck and galley
```

libs/wire-api/src/Wire/API/FederationUpdate.hs
fisx: oh, and resolving the conflict made me notice i forgot to remove the comment.  :-)  this mixup was a success!

fisx (approved):

PR #3408, opened 2023-07-06
Improve integration tests: Remove duplication of run-services script

smatting:
This PR:

- Replaces the `run-services` python script and with a binary that shares code with the ModService module from `/integration`
- reduces startup-time of all integration tests by starting all backends in-parallel instead of in sequential order
- Removes nginz acces.log from stdout (Also fixes /var/log/nginx/error.log not found error)


integration/test/Testlib/RunServices.hs
```diff
@@ -0,0 +1,135 @@
...
+      berAwsUserJournalQueue = "integration-user-events.fifo2",
+      berAwsPrekeyTable = "integration-brig-prekeys2",
+      berAwsS3Bucket = "dummy-bucket2",
+      berAwsQueueName = "integration-gundeck-events2",
+      berBrigInternalEvents = "integration-brig-events-internal2",
+      berEmailSMSSesQueue = "integration-brig-events2",
+      berEmailSMSEmailSender = "backend-integration2@wire.com",
+      berGalleyJournal = "integration-team-events.fifo2",
+      -- TODO
+      berVHost = "/",
```
smatting, line 75: @battermann do you know what vhost backend B (b.example.com) should have?

battermann: Not sure. AFAIK the rabbitmq queues are named after the target backend. So, messages going to `example.com` go into the queue named `example.com`. If multiple backends have the same target backend, this will cause a clash. But if there are only 2 backends (like in the old integration tests `example.com` and `b.example.com`), this doesn't matter. However, if backends `example.com` and `b.example.com` want to send messages to the dynamic backends, this can cause a problem. If we ever want integration tests where the old backends talk to the dynamic backends, we need different vhosts for the old backends. The convention is to name the vhost after the domain. I'd propose for consistency to create vhosts for backend A and B, named `example.com` and `b.example.com`. These can be created via curl:

```
curl -u "$RABBITMQ_USERNAME":"$RABBITMQ_PASSWORD" -X PUT "$ENDPOINT_URL"/example.com
```

pcapriotti (approved):

PR #3416, opened 2023-07-10
K8s integration tests: Disable persistence for `redis-cluster` and `rabbitmq`

smatting:
Motivation: Disable persistence for `redis-cluster` and `rabbitmq` for k8s run integration tests.

Disabling persistence became a feature of `redis-cluster` in chart version `8.3.0`: https://github.com/bitnami/charts/commit/eba6b1bc10f4bcf6bb2748c1106be94a17fcc439
but `8.3.0` is not available in the repo. I found `8.6.6` is available.

## Checklist

### DIFF
--- a/hack/helm_vars/redis-cluster/values.yaml.gotmpl
+++ b/hack/helm_vars/redis-cluster/values.yaml.gotmpl
@@ -1,8 +1,3 @@
-global:
-  storageClass: {{ .Values.storageClass }}
-
 redis-cluster:
   persistence:
-    size: 100Mi
-  volumePermissions:
-    enabled: true
+    enabled: false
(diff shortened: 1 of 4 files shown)


hack/helm_vars/redis-cluster/values.yaml.gotmpl
```diff
@@ -1,8 +1,3 @@
-global:
-  storageClass: {{ .Values.storageClass }}
```
jschaul, line 2: this is used for both rabbitmq and redis. removing this here will just lead to helm templating errors, you need to also remove the usage of this; as well as disable persistence in rabbitmq.

jschaul: nevermind, I'm wrong.

smatting:
@jschaul Closing this PR, because I think it's not possible. Disabling persistence in `redis-cluster` was only introduced inhttps://github.com/bitnami/charts/commit/eba6b1bc10f4bcf6bb2748c1106be94a17fcc439 which already comes with redis 7. Our code (hedis) seems not be compatible with redis 7.

PR #3421, opened 2023-07-12
WPB-2986: Background-worker graceful termination

lepsa:
Adding graceful shutdown handling via POSIX signals and exception handling.

Kubernetes signals and timings come from [Kubernetes best practices: terminating with grace](https://cloud.google.com/blog/products/containers-kubernetes/kubernetes-best-practices-terminating-with-grace)

I've included two scripts for manually testing these changes locally. I don't know how we would coordinate pushing messages to rabbit, running background-worker, sending signals, and checking if the service quit gracefully. If someone does have an idea on this I'd like to include that in this PR.

## Checklist


services/background-worker/src/Wire/BackendNotificationPusher.hs
```diff
@@ -148,12 +175,47 @@ getRemoteDomains = do
...
+  consumersRef <- newIORef mempty
+  let -- cleanup the refs when channels die
+      -- This is so we aren't trying to close consumers
+      -- that don't exist when the service is shutdown.
+      l = logger env
+      clearRefs = do
+        atomicWriteIORef consumersRef mempty
+      cleanup chan = do
+        readIORef consumersRef >>= traverse_ \(consumer, mvar) -> do
+          Log'.info l $ Log.msg (Log.val "Cancelling") . Log.field "Consumer" consumer
```
akshaymankar, line 192:
```suggestion
          Log'.info l $ Log.msg (Log.val "Cancelling consumer") . Log.field "consumer" consumer
```

The consumer tag is just a number, probably not as useful as the domain name, which should be the key of the `Map`, maybe use `Map.traverseWithKey`.

services/background-worker/src/Wire/BackendNotificationPusher.hs
```diff
@@ -61,7 +72,21 @@ pushNotification targetDomain (msg, envelope) = do
           <> [ skipChanThreadKilled,
                logRetries (const $ pure True) logErrr
              ]
-  recovering policy handlers $ const go
+  -- Ensure that the mvars are reset correctly.
+  UnliftIO.bracket_ (takeMVar mvar) (putMVar mvar ()) $ do
```
akshaymankar, line 76: Why reset them? Just use it once as a signal to stop retrying?

akshaymankar (changes requested):
1. I don't understand why we have 1 MVar to tell the thread to stop and the same one to mean that the thread stopped. I might be interpreting this wrong, so please explain. I am not sure why not use two MVars: one to signal to the consumer thread to stop retrying, second as a signal from the thread that it stopped. There can be only 1 thread running per domain at any given time anyway.

2. The `MVar` is just called `mvar` everywhere. Can it please have a better variable name?

libs/wai-utilities/src/Network/Wai/Utilities/Server.hs
```diff
@@ -137,8 +142,8 @@ runSettingsWithShutdown s app (fromMaybe defaultShutdownTime -> secs) = do
     initialization = do
       spawnGCMetricsCollector
     catchSignals closeSocket = do
-      void $ installHandler sigINT (Sig.CatchOnce closeSocket) Nothing
-      void $ installHandler sigTERM (Sig.CatchOnce closeSocket) Nothing
+      void $ installHandler sigINT (Sig.CatchOnce $ cleanup >> closeSocket) Nothing
+      void $ installHandler sigTERM (Sig.CatchOnce $ cleanup >> closeSocket) Nothing
```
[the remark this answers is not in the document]
lepsa: Good catch. Will do.

services/background-worker/src/Wire/BackendNotificationPusher.hs
lepsa: These aren't used once, they are paired with the consuming callback, and will be used each time a message for that callback arrives. These MVars are also shared with the cleanup code so it can wait until a given in-flight message is finished processing before shutting down the AMQP connection.

services/background-worker/src/Wire/BackendNotificationPusher.hs
```diff
@@ -61,7 +72,21 @@ pushNotification targetDomain (msg, envelope) = do
...
+    -- to deliver a single message at a time and to wait for confirmation.
+    UnliftIO.onException
+      (recovering policy handlers (const go))
+      -- Reject for redelivery. This is needed for the following reasons
+      -- 1) We have a strict time limit when kubernetes sends SIGTERM. We need to stay within that.
+      -- 2) Just because we failed to push the notification _now_ doesn't mean we won't be able to
+      --    in a few minutes. Requeuing the message will give us another go at it, and the next pod
+      --    to gain single exclusive consumer will get this message.
+      $ lift
+      $ reject envelope True
```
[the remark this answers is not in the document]
lepsa: Messages are requeued in their original position, if possible. https://www.rabbitmq.com/nack.html

However I think we still want to use the MVars so that we can use the graceful shutdown time the kubernetes gives us to try and finish processing any given message before we gracefully shutdown the amqp connection. Without the cross-thread signalling we might tear down the network connection while a notification is being pushed out, before it can be acknowledged. Not having the MVars will increase re-delivery of messages when the pod is shutdown if anything has been delivered by http, but the connection is closed before amqp can signal that.

services/background-worker/src/Wire/BackendNotificationPusher.hs
lepsa: :+1:

services/background-worker/src/Wire/BackendNotificationPusher.hs
```diff
@@ -38,13 +43,19 @@ instance RabbitMQEnvelope Q.Envelope where
...
-  -- FUTUREWORK: Pull these numbers into config
-  let policy = capDelay 300_000_000 $ fullJitterBackoff 10000
+  -- FUTUREWORK: Pull these numbers into config.
+  -- FUTUREWORK: Is this futurework still needed given the new changes?
+  -- Limit retries to a max of 20 seconds, as this is what Kubernetes is going
+  -- to give us for a SIGTERM shutdown notice by default. If we take longer than
+  -- that, it will SIGKILL the pod and there is nothing we can do to stop that.
+  -- If we fail after this, the notification will be NACKed, and will be redelivered
+  -- by RabbitMQ for another attempt.
+  let policy = limitRetriesByCumulativeDelay (20 * 1_000_000) $ fullJitterBackoff 10000
```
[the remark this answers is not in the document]
lepsa: So we still need a config update. Setting the max delay to the kubernetes grace timeout isn't going to work because that won't factor in processing time. We will need a fudge factor if we are setting those values to be equal.

services/background-worker/src/Wire/BackendNotificationPusher.hs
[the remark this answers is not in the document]
lepsa: I've added more comments to make this clearer. Without the MVars there is a brief window where messages could be delivered to endpoint services but not ACKed to Rabbit. This is a problem that we should try to avoid as much as possible.

services/background-worker/src/Wire/BackendNotificationPusher.hs
lepsa: If we use them once there are several problems.
1) The next time the consumer is called it will block forever as the MVar is already empty/full. Nothing would be clearing it for the next invocation.
2) The cleanup thread would have a similar issue, it would know that the first invocation of the consumer is either running or finished, but it would have no idea about all invocations after that and would blindly close the connection to AMQP, potentially while notifications are being sent and before they can be ACKed correctly, leading to redelivery to services.
3) Memory leaks. If the MVar is used once per consumer invocation, then we will have track them somewhere until the cleanup code is called and it would have to check all of them to be sure that the consumers aren't running. Alternatively we'd need another thread to periodically clean out this MVar collection, but that just adds to the multithreading coordination problem we already have.

lepsa:
> 1. I don't understand why we have 1 MVar to tell the thread to stop and the same one to mean that the thread stopped. I might be interpreting this wrong, so please explain. I am not sure why not use two MVars: one to signal to the consumer thread to stop retrying, second as a signal from the thread that it stopped. There can be only 1 thread running per domain at any given time anyway.
> 
>     2. The `MVar` is just called `mvar` everywhere. Can it please have a better variable name?

The single mvar is primarily used to signal to cleanup code when the thread has finished processing a message from Rabbit so that we don't close AMQP connections and channels before threads have finished with them. That the mvar can block the thread from sending out another notification isn't the goal, but it is a nice benefit of using them.
The primary way in which consumers are stopped from ingesting messages is that we remove them from the Rabbit channel. This will tell `amqp` to deregister that consumer, but since we have threads and laziness to deal with being proactive about checking that resources aren't going to be used seems prudent.

akshaymankar:
Ormolu is not happy:

```diff
integration/test/Testlib/Types.hs...  ok
libs/brig-types/src/Brig/Types/User.hs...  ok
libs/gundeck-types/src/Gundeck/Types/Push/V2.hs...  ok
libs/types-common-aws/src/Util/Test/SQS.hs...  ok
libs/wire-api/test/unit/Test/Wire/API/User/Search.hs...  ok
services/background-worker/exec/Main.hs...  ok
services/background-worker/src/Wire/BackendNotificationPusher.hs
@@ -57,7 +57,7 @@
    --
    -- If we fail to deliver the notification after policy, the notification will be NACKed,
    -- and will be redelivered by RabbitMQ for another attempt, most likely by the same pod.
-   let delayUsablePercentage = 2/3 :: Float
+   let delayUsablePercentage = 2 / 3 :: Float
        policy = limitRetriesByCumulativeDelay (floor $ delayUsablePercentage * fromIntegral env.shutdownGraceTime * 1_000_000) $ fullJitterBackoff 10000
        logErrr willRetry (SomeException e) rs = do
          Log.err $
services/background-worker/src/Wire/BackendNotificationPusher.hs...  *** FAILED
services/background-worker/src/Wire/BackgroundWorker.hs...  ok
services/background-worker/src/Wire/BackgroundWorker/Env.hs
@@ -12,6 +12,7 @@
  import Imports
  import Network.AMQP.Extended
  import qualified Network.RabbitMqAdmin as RabbitMqAdmin
+ import Numeric.Natural
  import OpenSSL.Session (SSLOption (..))
  import qualified OpenSSL.Session as SSL
  import Prometheus
@@ -21,6 +22,5 @@
  import qualified System.Logger.Extended as Log
  import Util.Options
  import Wire.BackgroundWorker.Options
- import Numeric.Natural
  type IsWorking = Bool
services/background-worker/src/Wire/BackgroundWorker/Env.hs...  *** FAILED
services/background-worker/src/Wire/BackgroundWorker/Health.hs...  ok
services/background-worker/src/Wire/BackgroundWorker/Options.hs
@@ -3,9 +3,9 @@
  import Data.Aeson
  import Imports
  import Network.AMQP.Extended
+ import Numeric.Natural
  import System.Logger.Extended
  import Util.Options
- import Numeric.Natural
  data Opts = Opts
    { logLevel :: !Level,
services/background-worker/src/Wire/BackgroundWorker/Options.hs...  *** FAILED
services/proxy/src/Proxy/Run.hs...  ok
tools/stern/test/unit/Main.hs...  ok
ormolu failed on 3 files.
you can fix this by running 'make format' from the git repo root.
make: *** [Makefile:216: formatc] Error 1
```

services/background-worker/src/Wire/BackgroundWorker.hs
```diff
@@ -1,11 +1,17 @@
...
+import qualified Data.Map.Strict as Map
 import qualified Data.Metrics.Servant as Metrics
 import Imports
+import qualified Network.AMQP as Q
 import Network.Wai.Utilities.Server
 import Servant
 import Servant.Server.Generic
-import UnliftIO.Async
+import qualified System.Logger as Log
+import qualified System.Logger as Log'
```
akshaymankar, line 14:
```suggestion
import qualified System.Logger as Log
```

services/background-worker/test/Test/Wire/BackendNotificationPusherSpec.hs
```diff
@@ -90,10 +90,10 @@ spec = do
               { Q.msgBody = Aeson.encode notif,
                 Q.msgContentType = Just "application/json"
               }
-
+      mvar <- newMVar ()
```
akshaymankar, line 93:
```suggestion
      runningFlag <- newMVar ()
```

akshaymankar (approved):
Looks good overall, minor nits.

services/background-worker/src/Wire/BackgroundWorker.hs
lepsa: This will break the cleanup code logging. I need Log' so that I can pass in the logger to the IO code, rather than relying on the classy version.

services/background-worker/test/Test/Wire/BackendNotificationPusherSpec.hs
lepsa: I've renamed all of these.

services/background-worker/src/Wire/Defederation.hs
```diff
@@ -53,75 +50,65 @@ deleteFederationDomainInner' go (msg, envelope) = do
...
-          liftIO $ ack envelope
-        else -- ensure that the message is requeued
-        -- This message was able to be parsed but something
-        -- else in our stack failed and we should try again.
-          liftIO $ reject envelope True
+    callGalley d = do
+      env <- ask
+      -- Jittered exponential backoff with 10ms as starting delay and 300s as max
+      -- delay. When 300s is reached, every retry will happen after 300s.
+      let policy = capDelay 300_000_000 $ fullJitterBackoff 10000
```
akshaymankar, line 61:
```suggestion
      -- Jittered exponential backoff with 10ms as starting delay and 300s as max
      -- delay. When 60s is reached, every retry will happen after 300s.
      let policy = capDelay 60_000_000 $ fullJitterBackoff 10000
```

I think 5 mins is too much to wait for a local component

services/background-worker/src/Wire/Defederation.hs
```diff
@@ -53,75 +50,65 @@ deleteFederationDomainInner' go (msg, envelope) = do
...
-    go resp = do
-      let code = statusCode $ responseStatus resp
-      if code >= 200 && code <= 299
-        then do
-          liftIO $ ack envelope
-        else -- ensure that the message is requeued
-        -- This message was able to be parsed but something
-        -- else in our stack failed and we should try again.
-          liftIO $ reject envelope True
+    callGalley d = do
```
akshaymankar, line 57:
```suggestion
    callGalley domain = do
```

services/background-worker/src/Wire/Defederation.hs
```diff
@@ -53,75 +50,65 @@ deleteFederationDomainInner' go (msg, envelope) = do
...
+            liftIO $ reject envelope True
+    req env dom =
+      defaultRequest
+        { method = methodDelete,
+          secure = False,
+          host = galley env ^. epHost . to encodeUtf8,
+          port = galley env ^. epPort . to fromIntegral,
+          path = "/i/federation/" <> toByteString' dom,
+          requestHeaders = ("Accept", "application/json") : requestHeaders defaultRequest,
+          responseTimeout = defederationTimeout env
```
akshaymankar, line 84: Looks like a long running http request, this is likely going to blow up for a big enough instance. I think we need to reconsider this approach, but perhaps in a separate PR.

services/background-worker/src/Wire/Defederation.hs
```diff
@@ -53,75 +50,64 @@ deleteFederationDomainInner' go (msg, envelope) = do
...
+      env <- ask
+      -- Jittered exponential backoff with 10ms as starting delay and 300s as max
+      -- delay. When 300s is reached, every retry will happen after 300s.
+      let policy = capDelay 300_000_000 $ fullJitterBackoff 10000
+          manager = httpManager env
+      recovering policy httpHandlers $ \_ ->
+        bracket_ (takeMVar runningFlag) (putMVar runningFlag ()) $ do
+          -- Non 2xx responses will throw an exception
+          -- So we are relying on that to be caught by recovering
+          resp <- liftIO $ httpLbs (req env d) manager
```
akshaymankar, line 67: This seems like perfect opportunity to use servant, why are we building requests by hand? Perhaps this is also for another PR.

akshaymankar (changes requested):

services/background-worker/src/Wire/Defederation.hs
lepsa: I think that a separate PR would be best, as the other internal routes in galley should also be moved over to servant.

services/background-worker/src/Wire/Defederation.hs
lepsa: Done

services/background-worker/src/Wire/Defederation.hs
lepsa: Done

akshaymankar (approved):

PR #3427, opened 2023-07-12
Disable persistence for redis-cluster in tests (upgrade helm chart with overwritten image)

smatting:
## Checklist

### DIFF
--- a/charts/redis-cluster/requirements.yaml
+++ b/charts/redis-cluster/requirements.yaml
@@ -1,4 +1,4 @@
 dependencies:
 - name: redis-cluster
-  version: 7.6.4
+  version: 8.6.6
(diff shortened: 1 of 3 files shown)


fisx:
CI doesn't like it, though :)

```
  Error: failed to parse /tmp/helmfile3136517592/test-pgdnaddg6dx9-redis-cluster-values-6b446cbc47: error converting YAML to JSON: yaml: line 7: did not find expected key
```

smatting:
Looks like it's not released yet, though, the latest version 0.15.2 doesn't include it. I don't think our motivation (disabling persistence) justify all that work (we also have custom reconnection logic) and risk involved. We have also the option of using `redis-ephemeral` for this, so I would close upgrading redis-cluster for now.

PR #3431, opened 2023-07-13
Do not send member updates to all

pcapriotti:
This PR changes which users receive the following team events:

- `team.update`
- `team.member-join`
- `team.member-leave`
- `team.member-update`

Without this PR these events are sent to arbitrary subset of 2000 team members. This is due to the fanout limit. For teams smaller than 2000 members all team members receive the events.

With this PR these events are sent only to team admins, so that the `team-settings` app can receive them. These events are sent transient, i.e. they are only sent to open websockets (opened by `team-settings`) and not persisted in the admin's notification stream.
This PR also limits the number of team admins of a team to 2000.
This PR also introduced an index of all team admins (table `galley.team_admin`), which is backfilled in an automated data migration.

This PR does not remove the fanout limit in general.

Tracked by https://wearezeta.atlassian.net/browse/WPB-2565

## Checklist


services/galley/schema/src/V82_CreateTableTeamAdmin.hs
```diff
@@ -0,0 +1,36 @@
...
+
+migration :: Migration
+migration = Migration 82 "Create table `team_admin`" $ do
+  schema'
+    [r|
+        CREATE TABLE team_admin (
+            team   uuid,
+            user   uuid,
+            PRIMARY KEY (team, user)
+        );
```
jschaul, line 35: Please use LeveledCompactionStrategy, it's not cassandra's default, but for almost all kind of data we have at Wire it's the better choice.

```suggestion
        ) with compaction = {'class': 'LeveledCompactionStrategy'};
```

jschaul (approved):
I just reviewed cassandra tables; data migration; cql statement use. Looks good! Minor nit-pick inline.

As for notifying users - behaviour was changed, is there a way to add a test to ensure the team admins who are online see the events? Or is there already a test covering that?

services/galley/schema/src/V82_CreateTableTeamAdmin.hs
smatting: Thanks! Updated schema https://github.com/wireapp/wire-server/pull/3431/commits/05ff0045c429763f797e46ebbe101dded772705f

smatting:
Integration tests passed for https://github.com/wireapp/wire-server/pull/3431/commits/b8032600353ecf0bd6ad7daf7efdf9190f500594

PR #3444, opened 2023-07-17
Wpb 2824 disable redis persistence (upgrade hedis)

jschaul:
Just to see what breaks - if anything - when we pin hedis to latest version. Follow-up from https://github.com/wireapp/wire-server/pull/3427

## Checklist


jschaul:
Does not work; hedis unit tests run forever and never stop. Code does not seem ready.

PR #3446, opened 2023-07-18
In CI, use redis-ephemeral master mode

jschaul:
We wish to avoid needing persistence during CI runs, as we run into limits of volume creation, which causes integration test setup failures.

This PR can be reverted at some point in the future when:

- hedis code is both compatible with current redis version AND redis version 7 and its tests pass, and it is released to hackage / nixpkgs.
- (or temporarily if you make changes to redis and/or redis code in gundeck and wish to test these changes on CI)

Then:

- Step 1. We can use latest hedis code (and merge this to develop, ensuring this works on staging)
- Step 2: Subsequently this PR can be reverted, and the redis-cluster helm chart bumped, and persistence disabled. See also #3444
- ensure to do steps 1 and 2 on *separate* PRs, to avoid requiring a redis helm chart update concurrently with a code update - that would not work well in production and cause downtime.

Context:

- https://wearezeta.atlassian.net/browse/WPB-2824
- https://wearezeta.atlassian.net/browse/WPB-2783
- https://github.com/wireapp/wire-server/pull/3427
- https://github.com/wireapp/wire-server/pull/3444
- https://github.com/wireapp/wire-server/pull/3445

## Checklist

### DIFF
--- a/changelog.d/5-internal/ci-redis-ephemeral
+++ b/changelog.d/5-internal/ci-redis-ephemeral
@@ -1,0 +1,1 @@
+In CI integration tests, use redis-ephemeral in master mode (may be reverted in the future, see PR details)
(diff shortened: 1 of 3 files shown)


jschaul:
unrelated integration test failures...

PR #3447, opened 2023-07-18
Revert "Do not send member updates to all (#3431)"

pcapriotti:
Looks like preventing team events from being sent to regular team users has some consequences for clients, not just the team settings app.

## Checklist

 - [x] Revert changelog changes


services/galley/src/Galley/Cassandra.hs
```diff
@@ -20,4 +20,4 @@ module Galley.Cassandra (schemaVersion) where
 import Imports
 
 schemaVersion :: Int32
-schemaVersion = 83
+schemaVersion = 81
```
jschaul, line 23: I would leave this change in.

services/galley/src/Galley/Cassandra/Queries.hs
```diff
@@ -167,15 +167,6 @@ deleteBillingTeamMember = "delete from billing_team_member where team = ? and us
 listBillingTeamMembers :: PrepQuery R (Identity TeamId) (Identity UserId)
```
jschaul, line 167: I would leave this change in, not revert it.

services/galley/src/Galley/Cassandra/Team.hs
```diff
@@ -77,7 +77,6 @@ interpretTeamStoreToCassandra lh = interpret $ \case
   CreateTeam t uid n i k b -> embedClient $ createTeam t uid n i k b
   DeleteTeamMember tid uid -> embedClient $ removeTeamMember tid uid
```
jschaul, line 78: I would leave this change (the whole file) in, not revert it.

jschaul (commented):
Since the galley data migration has now already run; I think we should keep the logic which writes and deletes to the new table in place. And only revert the business logic about who gets events. Otherwise we get a gap of events not recorded in that table from t0=this-PR-deployed to t1=future-time-where-new-PR-gets-deployed.

pcapriotti:
> Since the galley data migration has now already run; I think we should keep the logic which writes and deletes to the new table in place. And only revert the business logic about who gets events. Otherwise we get a gap of events not recorded in that table from t0=this-PR-deployed to t1=future-time-where-new-PR-gets-deployed.

Good point, I didn't think of that. Should be fixed by the latest commit. It's a bit unfortunate that now we have to fetch admins just to check the admin limit, and not use them for notifications. An alternative would be to leave things as they used to be, and re-run the data migration next time. But I guess this is fine.

services/galley/src/Galley/Cassandra.hs
pcapriotti: Fixed.

services/galley/src/Galley/Cassandra/Team.hs
pcapriotti: Fixed.

jschaul (approved):

PR #3450, opened 2023-07-19
integration: Produce non-zero exitcode on failures

akshaymankar:
https://wearezeta.atlassian.net/browse/WPB-3297

## Checklist

 - [x] ~Add a new entry in an appropriate subdirectory of `changelog.d`~ No changelog as this was broken during this release and fixing this shouldn't be necessary to be put in a separate changelog.


integration/test/Test/Conversation.hs
```diff
@@ -106,20 +108,13 @@ testFederationStatus :: HasCallStack => App ()
...
     $ \resp -> do
       resp.status `shouldMatchInt` 200
       resp.json %. "status" `shouldMatch` "fully-connected"
 
-  bindResponse
-    (API.getFederationStatus uid [unknownDomain])
-    $ \resp -> do
-      resp.status `shouldMatchInt` 400
-      resp.json %. "label" `shouldMatch` "discovery-failure"
-
```
akshaymankar, line 122: Note to reviewer: This test cannot be done in K8s because it relies on the DNS server to refuse to answer queries for a domain. This cannot happen in K8s because the DNS server there assumes that it has access to information of all domains.

smatting (approved):

PR #3456, opened 2023-07-27
fix 2 integration tests in new integration test suite

battermann:
2 integration tests from new test suite are failing locally.

- `testSwagger`
  - Not fixed, but the offending commit was found, see code comment above the test
- `testDynamicBackendsNotFederating`
  - `fedration-denied` status code changed from 400 to 422, it is not obvious why, so I changed it back. If that is wrong, we need to update the test instead

## Checklist


elland:
https://github.com/wireapp/wire-server/pull/3450

pcapriotti:
See https://github.com/wireapp/wire-server/pull/3461 and https://github.com/wireapp/wire-server/pull/3458.

PR #3457, opened 2023-07-27
[chore] Migrate from Haskell2010 to GHC2021.

elland:
Reviewers might want to do it per-commit, as commit one does the actual change, the follow up is reformatting.


services/brig/src/Brig/App.hs
```diff
@@ -491,7 +491,7 @@ instance Applicative (AppT r) where
   (AppT x0) <*> (AppT x1) = AppT $ x0 <*> x1
 
 instance Monad (AppT r) where
-  (AppT x0) >>= f = AppT $ x0 >>= unAppT . f
+  (AppT x0) >>= f = AppT $ x0 >>= (\x -> unAppT $ f x)
```
fisx, line 494: i don't particularly like this change.  why did it become necessary?

services/brig/test/integration/Util.hs
```diff
@@ -592,7 +592,7 @@ putConnectionQualified brig from (Qualified to toDomain) r =
   where
     payload = RequestBodyLBS . encode $ object ["status" .= r]
 
-connectUsers :: Brig -> UserId -> List1 UserId -> (MonadIO m, MonadHttp m) => m ()
+connectUsers :: (MonadIO m, MonadHttp m) => Brig -> UserId -> List1 UserId -> m ()
```
fisx, line 595: why?

services/gundeck/test/unit/ThreadBudget.hs
```diff
@@ -1,4 +1,5 @@
 {-# LANGUAGE DeriveAnyClass #-}
+{-# LANGUAGE NoGeneralizedNewtypeDeriving #-}
```
fisx: why?  and if this is part of ghc2010, should we disable it globally instead?

fisx (approved):
looks good!  got some suggestions, but i'd like to leave it to you to decide what to do with them.

services/gundeck/test/unit/ThreadBudget.hs
elland: I'm surprised they released GHC2021 with that extension on. Removing it everywhere might be wise.

services/brig/src/Brig/App.hs
elland: GHC2021 disables the `DeepSubsumption` extension by default, so the options were to either enable it for this file or use an explicit lambda like this, or add some eldritch looking type signatures. After some digging and some thought, I think we're fine with either the extension or the eta expansion. I reverted this change and enabled the extension for this file.

services/brig/test/integration/Util.hs
elland: My understanding of it: The original version relied on GHC + DeepSubsumption to rewrite this to put the constraint "in the right place". The extension removes the requirement that `forall` must be always in the same place, through eta-expansions and other rewrite rules.

PR #3461, opened 2023-07-31
Fix assertion in integration tests

pcapriotti:
This fixes an integration test failure on develop

## Checklist

 - [x] No CHANGELOG entry

### DIFF
--- a/integration/test/Test/Conversation.hs
+++ b/integration/test/Test/Conversation.hs
@@ -44,7 +44,7 @@ testDynamicBackendsNotFederating = do
         $ bindResponse
           (API.getFederationStatus uidA [domainB, domainC])
         $ \resp -> do
-          resp.status `shouldMatchInt` 400
+          resp.status `shouldMatchInt` 422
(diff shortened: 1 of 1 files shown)


integration/test/Test/Conversation.hs
```diff
@@ -44,7 +44,7 @@ testDynamicBackendsNotFederating = do
         $ bindResponse
           (API.getFederationStatus uidA [domainB, domainC])
         $ \resp -> do
-          resp.status `shouldMatchInt` 400
+          resp.status `shouldMatchInt` 422
```
fisx, line 47: why do we have to change error codes all the time?  is it because moving to servant?

also, why wasn't that caught on concourse?

fisx (approved):

integration/test/Test/Conversation.hs
pcapriotti: I don't know why the code changed. These tests aren't running on CI.

PR #3471, opened 2023-08-01
Expose sso-deep-link api on stern/backoffice

fisx:
https://wearezeta.atlassian.net/browse/WPB-2852

## Checklist


fisx:
@arthurwolf could you just double-check that the docs make sense, and ideally fill in the TBD., or let me know if this is documented anywhere else?  This doesn't need to be perfect, we can always improve on it later.

@julialongtin FYI.

tools/stern/test/integration/API.hs
```diff
@@ -103,6 +105,34 @@ tests s =
       -- - `POST /teams/:tid/billing`
     ]
 
+testRudSsoDomainRedirect :: TestM ()
+testRudSsoDomainRedirect = do
+  testGet 1 Nothing
```
battermann, line 110: The `Int` parameter does not make a lot of sense to me. I'd rather have something more expressive, but this is not really important.

battermann (approved):

tools/stern/test/integration/API.hs
fisx: yeah, bad habit of mine.  it's just to locate the failure from the output easier, there isn't much more meaning behind it.

PR #3473, opened 2023-08-02
Fix: SCIM user lookup after changing IdP issuer ID

fisx:
https://wearezeta.atlassian.net/browse/WPB-3530

this only adds a set of integration test with one failing (and thus deactivated) instance.  fix coming up in #3563, but i'd rather merge this now and keep the other PR smaller.

## Checklist

 - [ ] ~~Add a new entry in an appropriate subdirectory of `changelog.d`~~

### DIFF
--- a/libs/wire-api/src/Wire/API/User/IdentityProvider.hs
+++ b/libs/wire-api/src/Wire/API/User/IdentityProvider.hs
@@ -24,6 +24,7 @@ import Control.Lens (makeLenses, (.~), (?~))
 import Control.Monad.Except
 import Data.Aeson
 import Data.Aeson.TH
+import Data.Aeson.Types (parseMaybe)
 import Data.Attoparsec.ByteString qualified as AP
 import Data.Binary.Builder qualified as BSB
 import Data.ByteString.Conversion qualified as BSC
@@ -168,6 +169,10 @@ instance FromJSON IdPMetadataInfo where
 instance ToJSON IdPMetadataInfo where
   toJSON (IdPMetadataValue _ x) =
     object ["value" .= SAML.encode x]
+
+-- | (Returning 'Nothing' would be an internal error.)
+idPMetadataToInfo :: SAML.IdPMetadata -> Maybe IdPMetadataInfo
+idPMetadataToInfo = parseMaybe parseJSON . toJSON . IdPMetadataValue undefined
 
 -- Swagger instances
 
--- a/libs/wire-api/src/Wire/API/User/IdentityProvider.hs
+++ b/libs/wire-api/src/Wire/API/User/IdentityProvider.hs
@@ -170,9 +170,11 @@ instance ToJSON IdPMetadataInfo where
   toJSON (IdPMetadataValue _ x) =
     object ["value" .= SAML.encode x]
 
--- | (Returning 'Nothing' would be an internal error.)
-idPMetadataToInfo :: SAML.IdPMetadata -> Maybe IdPMetadataInfo
-idPMetadataToInfo = parseMaybe parseJSON . toJSON . IdPMetadataValue undefined
+idPMetadataToInfo :: SAML.IdPMetadata -> IdPMetadataInfo
+idPMetadataToInfo =
+  -- 'undefined' is fine because `instance toJSON IdPMetadataValue` ignores it.  'fromJust' is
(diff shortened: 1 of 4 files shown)


libs/wire-api/src/Wire/API/User/IdentityProvider.hs
```diff
@@ -169,6 +170,10 @@ instance ToJSON IdPMetadataInfo where
   toJSON (IdPMetadataValue _ x) =
     object ["value" .= SAML.encode x]
 
+-- | (Returning 'Nothing' would be an internal error.)
```
elland, line 173: Could we encode this better in the types?

services/spar/test-integration/Test/Spar/APISpec.hs
```diff
@@ -1054,34 +1055,99 @@ specCRUDIdentityProvider = do
...
+                  -- FUTUREWORK: deprecate POST?
+                  if updateNotReplace
+                    then callIdpUpdate' (env ^. teSpar) (Just owner1) (idp1 ^. SAML.idpId) (fromJust $ idPMetadataToInfo idpmeta2)
+                    else callIdpCreateReplace (env ^. teWireIdPAPIVersion) (env ^. teSpar) (Just owner1) idpmeta2 (idp1 ^. SAML.idpId)
+
+          idp1' <- call $ callIdpGet (env ^. teSpar) (Just owner1) (idp1 ^. SAML.idpId)
+          idp2' <- call $ callIdpGet (env ^. teSpar) (Just owner1) (idp2 ^. SAML.idpId)
+          liftIO $ do
+            idp1'
+              `shouldBe` ( idp1
```
elland, line 1119: I think this would be a lot more readable and 'obviously correct' if extracted into named bindings instead. Doing all the lens accessors in place makes it rather to reason about.

libs/wire-api/src/Wire/API/User/IdentityProvider.hs
fisx: c22b9602421ffe7f266982faabb60419f06ca90b

services/spar/test-integration/Test/Spar/APISpec.hs
```diff
@@ -1054,34 +1055,99 @@ specCRUDIdentityProvider = do
...
+          issuer2 <- makeIssuer
+          idp2 <- do
+            let idpmeta2 = idpmeta1 & edIssuer .~ issuer2
+             in call $
+                  -- There are two mechanisms for re-aligning your team when your IdP metadata
+                  -- has changed: POST (create a new one, and mark it as replacing the old one),
+                  -- and PUT (updating the existing IdP's metadata).  The reason for having two
+                  -- ways to do this has been lost in history, but we're testing both here.
+                  --
+                  -- FUTUREWORK: deprecate POST?
```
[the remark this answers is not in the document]
fisx: yes, but it doesn't fit into this PR.  there is also the complication that i don't know off the top of my head if the team-management api is versioned, and if that's in lock step with clients.  (i think it is, because the two apis overlap, but then there are customers doing this with curl...)

services/spar/test-integration/Test/Spar/APISpec.hs
fisx: d2abbaae4c2a829285bdb2fcbe533cd0712d4d4a

not sure that does it for you, but i can't think of anything else that makes sense to me.  maybe pair over it, if you're still unhappy?

libs/wire-api/src/Wire/API/User/IdentityProvider.hs
```diff
@@ -169,6 +170,12 @@ instance ToJSON IdPMetadataInfo where
   toJSON (IdPMetadataValue _ x) =
     object ["value" .= SAML.encode x]
 
+idPMetadataToInfo :: SAML.IdPMetadata -> IdPMetadataInfo
+idPMetadataToInfo =
+  -- 'undefined' is fine because `instance toJSON IdPMetadataValue` ignores it.  'fromJust' is
+  -- ok as long as 'parseJSON . toJSON' always yields a value and not 'Nothing'.
+  fromJust . parseMaybe parseJSON . toJSON . IdPMetadataValue undefined
```
elland, line 177: I see how it works now in the instances, still feels weird :D

elland (approved):

PR #3460, opened 2023-07-28
[WPB-3394] Sending proteus messages when a remote backend is offline

battermann:
https://wearezeta.atlassian.net/browse/WPB-3394

## Checklist

 - [x] Add a new entry in an appropriate subdirectory of `changelog.d`
 - [x] Read and follow the [PR guidelines](https://docs.wire.com/developer/developer/pr-guidelines.html)

PR #3474, opened 2023-08-02
[WPB-3394] Sending proteus messages when a remote backend is offline (update)

battermann:
https://wearezeta.atlassian.net/browse/WPB-3394

This is an update to https://github.com/wireapp/wire-server/pull/3460

Users that are in `failed_to_send` are now also included in `failed_to_confirm_clients`.

No need to update change log.

## Checklist

 - [ ] ~Add a new entry in an appropriate subdirectory of `changelog.d`~

### DIFF
--- a/services/galley/src/Galley/API/Message.hs
+++ b/services/galley/src/Galley/API/Message.hs
@@ -492,17 +492,12 @@ postQualifiedOtrMessage senderType sender mconn lcnv msg =
           predicate (d, (u, _)) = any (\(d', (u', _)) -> d == d' && u == u') failed'
           -- Failed users/clients aren't redundant
           (failed, redundant) = partition predicate redundant'
+          collectedFailedToSend = collectFailedToSend [qualifiedUserClients failedToSend, toDomMap unconfirmedUnknownClients, fromDomUserClient failed]
       pure
         otrResult
-          { mssFailedToSend =
-              QualifiedUserClients $
-                collectFailedToSend
-                  [ qualifiedUserClients failedToSend,
-                    toDomMap unconfirmedUnknownClients,
-                    fromDomUserClient failed
-                  ],
+          { mssFailedToSend = QualifiedUserClients collectedFailedToSend,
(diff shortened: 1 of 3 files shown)


services/galley/src/Galley/API/Message.hs
```diff
@@ -492,17 +492,12 @@ postQualifiedOtrMessage senderType sender mconn lcnv msg =
           predicate (d, (u, _)) = any (\(d', (u', _)) -> d == d' && u == u') failed'
           -- Failed users/clients aren't redundant
           (failed, redundant) = partition predicate redundant'
+          collectedFailedToSend = collectFailedToSend [qualifiedUserClients failedToSend, toDomMap unconfirmedUnknownClients, fromDomUserClient failed]
```
battermann, line 495: `collectFailedToSend` would not be needed if we had a proper semigroup instance for `Map` where `a` has a monoid instance. But let's worry about this later.

fisx:
> No need to update change log.

yes, you should explicitly add the *two* PRs that implement the change listed in the first one.

fisx (approved):
for the record (@mc): i think it's a bad smell that we repeatedly avoid trivial, straight-forward changes on clients and instead make the protocol and the backend implementation more convoluted and harder to maintain.

but too late, i missed the meeting, so here you go :)

mastaab:
@fisx I think you maybe wanted to tag @marcoconti83 in your comment

fisx:
> @fisx I think you maybe wanted to tag @marcoconti83 in your comment

oops, yes, thanks!

marcoconti83:
> for the record (@mc): i think it's a bad smell that we repeatedly avoid trivial, straight-forward changes on clients and instead make the protocol and the backend implementation more convoluted and harder to maintain.
> 
> but too late, i missed the meeting, so here you go :)

I agree in general, and I'm not finger pointig, but it also goes both ways :) Changing the backend payload in a more structured manner would have been more work on the backend. This implementaton is a middle ground also for the backend effort :)

PR #3477, opened 2023-08-04
Fix defederate loop

fisx:
this builds on #3478 and collects more internal errors and makes them error-level log entries rather than stumbling into a crash loop.

https://wearezeta.atlassian.net/browse/WPB-3631

## Checklist


fisx:
replaced by https://github.com/wireapp/wire-server/pull/3481

PR #3478, opened 2023-08-04
Fix defederate loop - the easy part

fisx:
if the main defederation background process fails to remove a remote conversation because it's already gone, it crashes, gives the job back to rabbitMQ, and picks it up again immediately.  this PR avoids this particular situation by gracefully ignoring already-removed conversations.

there are more exceptions that can lead to crash loops, work in progress in https://github.com/wireapp/wire-server/pull/3477 and by @lepsa.

https://wearezeta.atlassian.net/browse/WPB-3631

## Checklist

### DIFF
--- a/changelog.d/3-bug-fixes/pr-3477
+++ b/changelog.d/3-bug-fixes/pr-3477
@@ -1,0 +1,1 @@
+Fix: When defederating, don't crash on already-deleted conversations.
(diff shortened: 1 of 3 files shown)


lepsa (approved):
LGTM.
This does raise the question of if there are more points in the code where I made these sort of assumptions about data aligning between tables when that isn't something that Cassandra offers.

PR #3472, opened 2023-08-02
[WPB-195] Prevent users from being added to an existing conversation on NFCG

battermann:
https://wearezeta.atlassian.net/browse/WPB-195

## Checklist

 - [x] Add a new entry in an appropriate subdirectory of `changelog.d`
 - [x] Read and follow the [PR guidelines](https://docs.wire.com/developer/developer/pr-guidelines.html)

PR #3483, opened 2023-08-07
Refactor NFCG error response

pcapriotti:
- Prevent users from being added to an existing conversation on NFCG
- Turn NFCG response into a Polysemy error
- Make it possible for handlers to fail with an arbitrary JSON error value

This supersedes https://github.com/wireapp/wire-server/pull/3472.

https://wearezeta.atlassian.net/browse/WPB-195

## Checklist


integration/test/SetupHelpers.hs
```diff
@@ -107,3 +108,20 @@ fullSearchWithAll =
         let remoteDomains = List.delete ownDomain $ [env.domain1, env.domain2] <> env.dynamicDomains
         addFullSearchFor remoteDomains val
     }
+
+withFederatingBackendsAllowDynamic :: HasCallStack => Int -> ((String, String, String) -> App a) -> App a
+withFederatingBackendsAllowDynamic n a = do
```
elland, line 113: Nitpick: not sure about calling the variable here `a`, instead of `f` or something a bit more explicit.

elland (approved):
🚢

PR #3485, opened 2023-08-08
[WPB-3633] client receives conversation deleted conversation events after federation delete event

battermann:
https://wearezeta.atlassian.net/browse/WPB-3633

## Checklist

### DIFF
--- a/integration/test/Test/Conversation.hs
+++ b/integration/test/Test/Conversation.hs
@@ -251,8 +251,7 @@ testDefederationOneOnOne = do
       connectAllDomainsAndWaitToSync 1 domains
       [uA, uB] <- createAndConnectUsers [domainA, domainB]
       -- figure out on which backend the 1:1 conversation is created
-      conn <- getConnection uA uB
-      qConvId <- conn.json %. "qualified_conversation"
+      qConvId <- getConnection uA uB >>= \c -> c.json %. "qualified_conversation"
 
       -- check conversation exists and uB is a member from POV of uA
       bindResponse (getConversation uA qConvId) $ \r -> do
@@ -282,14 +281,7 @@ testDefederationOneOnOne = do
 
       when (domainA /= conversationOwningDomain && domainB /= conversationOwningDomain) $ do
         -- this should not happen
-        assertFailure $
-          "conversation was expected to be created on "
-            <> domainA
-            <> " or "
-            <> domainB
-            <> " but was created on '"
-            <> conversationOwningDomain
-            <> "'"
+        error "impossible"
   where
     assertFederationTerminatingUserNoConvDeleteEvent :: Value -> Value -> String -> String -> App ()
     assertFederationTerminatingUserNoConvDeleteEvent user convId ownDomain otherDomain = do
--- a/integration/test/Test/Conversation.hs
+++ b/integration/test/Test/Conversation.hs
@@ -1,8 +1,10 @@
 module Test.Conversation where
 
+import API.Brig (getConnection)
 import API.BrigInternal qualified as Internal
-import API.Galley (defProteus, postConversation, qualifiedUsers)
+import API.Galley (defProteus, getConversation, postConversation, qualifiedUsers)
 import API.GalleyInternal qualified as API
+import API.Gundeck (getNotifications)
 import Control.Applicative
 import Data.Aeson qualified as Aeson
 import GHC.Stack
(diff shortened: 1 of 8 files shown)


services/brig/src/Brig/API/Internal.hs
```diff
@@ -366,13 +366,12 @@ deleteFederationRemote dom = do
...
 
 -- | Remove one-on-one conversations for the given remote domain. This is called from Galley as
--- part of the defederation process, and should not be called duriung the initial domain removal
+-- part of the defederation process, and should not be called during the initial domain removal
 -- call to brig. This is so we can ensure that domains are correctly cleaned up if a service
 -- falls over for whatever reason.
 deleteFederationRemoteGalley :: Domain -> ExceptT Brig.API.Error.Error (AppT r) ()
 deleteFederationRemoteGalley dom = do
   lift . wrapClient . Data.deleteRemoteConnectionsDomain $ dom
-  assertNoDomainsFromConfigFiles dom
```
fisx, line 375: the reason for this line is that if you delete a remote that's also present in the config file, it will fail silently (signal success), as read is always the union of cassandra and yaml, and yaml can't be changed here.  there is a comment about this somewhere in this CRUD family.

why did you remove it?

integration/test/Test/Conversation.hs
```diff
@@ -177,3 +179,136 @@ testCreateConversationNonFullyConnected = do
...
+  let setFederationConfig =
+        setField "optSettings.setFederationStrategy" "allowDynamic"
+          >=> removeField "optSettings.setFederationDomainConfigs"
+          >=> setField "optSettings.setFederationDomainConfigsUpdateFreq" (Aeson.Number 1)
+  startDynamicBackends
+    [ def {dbBrig = setFederationConfig},
+      def {dbBrig = setFederationConfig}
+    ]
+    $ \dynDomains -> do
+      domains@[domainA, domainB] <- pure dynDomains
```
fisx, line 250: still don't understand why we have take this detour every time.  maybe it's worth fixing this in package-wide in a separate PR?

integration/test/Test/Conversation.hs
```diff
@@ -177,3 +179,136 @@ testCreateConversationNonFullyConnected = do
...
+      connectAllDomainsAndWaitToSync 1 domains
+      [uA, uB] <- createAndConnectUsers [domainA, domainB]
+      -- figure out on which backend the 1:1 conversation is created
+      conn <- getConnection uA uB
```
fisx, line 254: `conn` usually refers to a cannon web socket connection of one client, but i don't have a better name either.  maybe this is fine?  (`uconn`?)

integration/test/Test/Conversation.hs
```diff
@@ -177,3 +179,136 @@ testCreateConversationNonFullyConnected = do
...
+      when (domainA /= conversationOwningDomain && domainB /= conversationOwningDomain) $ do
+        -- this should not happen
+        assertFailure $
+          "conversation was expected to be created on "
+            <> domainA
+            <> " or "
+            <> domainB
+            <> " but was created on '"
+            <> conversationOwningDomain
+            <> "'"
```
fisx, line 292: we don't expect this to ever happen, right?  what domain should that be?  so i wouldn't spend so much code on an elaborate error message here, but just case-switch over `conversationOwningDomain` with a catch-all saying `error "impossible"`.

services/galley/src/Galley/API/Internal.hs
```diff
@@ -611,50 +602,17 @@ deleteFederationDomainRemoteUserFromLocalConversations dom = do
...
-        -- to the remote backend.
-        -- We don't need to check the conversation type here, as we can't tell the
-        -- remote federation server to delete the conversation. They will have to do a
-        -- similar processing run for removing the local domain from their federation list.
-        onConversationUpdated dom convUpdate
+  remoteConvs <- foldr insertIntoMap mempty <$> E.getLocalMembersByDomain dom
+  for_ (Map.toList remoteConvs) $ \(cnv, lUsers) -> do
+    -- All errors, either exceptions or Either e, get thrown into IO
+    mapError @NoChanges (const (InternalErrorWithDescription "No Changes: Could not remove a local member from a remote conversation.")) $ do
+      E.deleteMembersInRemoteConversation (toRemoteUnsafe dom cnv) (N.toList lUsers)
```
fisx, line 615: this is certainly more to the point!  :)

fisx (approved):

integration/test/Test/Conversation.hs
battermann: I think it is alright, but sure, if we can improve it, I am fine with that.

PR #3503, opened 2023-08-15
Mitigate remotes list update lag

fisx:
https://wearezeta.atlassian.net/browse/WPB-3797

## Checklist


fisx:
This isn't working yet, and i suspect it's not just the last commit.  Will continue tomorrow!

fisx:
increasing the cache update frequency seems a bad idea.  just makes things more complicated, and laptop fans louder.  i'll try following the spirit of the ticket now.

PR #3518, opened 2023-08-21
WPB-3798 incorrect json field names

lepsa:
### Notable
The JSON schema for `NonConnectedBackends` has changed due to the changes in `CustomEncoded`, it is no longer dropping the "prefix" of `non`. @elland also noticed this issue in another ticket. https://github.com/wireapp/wire-server/pull/3489
```
newtype NonConnectedBackends = NonConnectedBackends
  { nonConnectedBackends :: Set Domain
  }
```

### Other
- Updating records that had prefix processing in their JSON schemas to remove leading prefixes from their fields, and updating code and tests to reflect these new field names.
  - Types from the https://github.com/wireapp/saml2-web-sso library weren't updated as part of this as they are in a different repository and should be handled separately.
- `federator` dropped `mtl` as a direct dependency.

## Open Questions: Answered, see Checklist
How do we deal with the API change around `NonConnectedBackends`? I can see a couple of options going forward.
- Breaking the API.
- We create a new API version for the routes that use it. This new version can have the correct field name of `non_connected_backends`. Older versions would need to keep their current semantics of `connected_backends`
- We notify client developers about the change so they can make the parsing accept both field names, and make the backend also accept both names. However clients should send the old name for compatibility.

## Checklist

   - I've talked with @akshaymankar about how we might deal with the change in the federation API, and he mentioned that we can probably perform a breaking change in this particular instance as it is unlikely to affect anyone. Further investigation is warranted to ensure this, however it is a promising start.


services/galley/test/integration/API/MLS/Util.hs
```diff
@@ -94,10 +94,10 @@ mapRemoteKeyPackageRef ::
   (Request -> Request) ->
   KeyPackageBundle ->
   m ()
-mapRemoteKeyPackageRef brig bundle =
+mapRemoteKeyPackageRef galleyCall bundle =
```
elland, line 97: Is this in fact galleyCall and it was a mistake before or is this a mistake now?

libs/wire-api/src/Wire/API/User/Orphans.hs
```diff
@@ -51,9 +53,17 @@ instance ToSchema CountryCode
 
 -- | The options to use for schema generation. Must match the options used
 -- for 'ToJSON' instances elsewhere.
+--
+-- FUTUREWORK: This should be removed once the saml2-web-sso types are updated to remove their prefixes.
```
elland, line 57: Can we add https://wearezeta.atlassian.net/browse/WPB-3972 to the comment?

lepsa: Yup yup

services/spar/test-integration/Util/Core.hs
```diff
@@ -265,9 +265,9 @@ mkEnv _teTstOpts _teOpts = do
   _teMgr :: Manager <- newManager defaultManagerSettings
```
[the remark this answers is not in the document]
lepsa: Prefixes yes, underscores no in the general case. Spar's TestEnv has `makeLenses` so the underscores are the least intrusive way of keeping that working.

elland (approved):

PR #3523, opened 2023-08-22
Integration tests: delete all rabbitmq queues during dynamic backends setup phase

smatting:
https://wearezeta.atlassian.net/browse/WPB-3974

This PR makes sure that all queues are deleted from rabbitmq before any tests start for any dynamically spawned backend.

This PR also refactors modules in `integration` so that `Testlib.Types` does not depend on any other modules. This should make resolving cyclic module dependencies easier to resolve in the future.


akshaymankar (approved):
:rocket:

PR #3526, opened 2023-08-22
[WPB 3796] Fix background worker for federation policy allowAll

fisx:
https://wearezeta.atlassian.net/browse/WPB-3796

## Checklist


services/background-worker/src/Wire/BackgroundWorker/Env.hs
```diff
@@ -47,9 +47,9 @@ data Env = Env
     galley :: Endpoint,
     brig :: Endpoint,
     defederationTimeout :: ResponseTimeout,
-    remoteDomains :: IORef FederationDomainConfigs,
-    remoteDomainsChan :: Chan FederationDomainConfigs,
-    backendNotificationMetrics :: BackendNotificationMetrics,
+    remoteDomains :: IORef FederationDomainConfigs, -- TODO: remove
+    remoteDomainsChan :: Chan FederationDomainConfigs, -- TODO: remove
+    backendNotificationMetrics :: BackendNotificationMetrics, -- TODO: move into BackendNotificationsConfig?
```
akshaymankar, line 52: This is not a config so it cannot be moved in there. Merging them in some way would create hierarchies which just make things more painful to access, so I would leave it like this.

fisx: (i noticed that, too.  :)

services/background-worker/src/Wire/BackendNotificationPusher.hs
```diff
@@ -132,29 +131,15 @@ startPusher consumersRef chan = do
...
-        -- Wait for a new set of domains. This is a blocking action
-        -- so we will only move past here when we get a new set of domains.
-        -- It is a bit nicer than having another timeout value, as Brig is
-        -- already providing one in the domain update message.
-        chanRemotes <- liftIO $ readChan env.remoteDomainsChan
-        -- Make new consumers for the new domains, clean up old ones from the consumer map.
-        ensureConsumers consumersRef chan $ domain <$> chanRemotes.remotes
+        remoteDomains <- getRemoteDomains
+        ensureConsumers consumers chan remoteDomains
+        threadDelay (round $ 1_000_000 * fromMaybe 60 env.backendNotificationsConfig.remotesRefreshInterval)
```
akshaymankar, line 138: Why is this a `Maybe`, we can put default values in all the places, i.e. helm chart and integration test configs? IMO this is unnecessary complexity.

services/background-worker/src/Wire/BackgroundWorker/Options.hs
```diff
@@ -31,7 +30,9 @@ data BackendNotificationsConfig = BackendNotificationsConfig
     -- | Upper limit on amount of time (in microseconds) to wait before retrying
     -- any notification. This exists to ensure that exponential back-off doesn't
     -- cause wait times to be very big.
-    pushBackoffMaxWait :: Int
+    pushBackoffMaxWait :: Int,
+    -- | Number of seconds between two calls to `getRemoteDomains`.
+    remotesRefreshInterval :: Maybe Double
```
akshaymankar, line 35: Instead of using `Double` can we just take number of milliseconds? We won't need rounding like this. I think we can safely assume people running wire are capable of understanding different units of time than seconds.
This also makes it more in line with the config one line above.

fisx: i get the last point, but i just think doubles are more ergonomic.  anyway i'll align it with the rest.

charts/background-worker/values.yaml
```diff
@@ -26,6 +26,7 @@ config:
   backendNotificationPusher:
     pushBackoffMinWait: 10000 # in microseconds, so 10ms
     pushBackoffMaxWait: 300000000 # microseconds, so 300s
+    remotesRefreshIntervalMs: 30000
```
akshaymankar, line 29: Adding `Ms` here has two problems:
1. Capital `M` means `Mega` not `micro`.
2. The things above do not have this suffix, so it is inconsistent.

libs/wire-api-federation/src/Wire/API/Federation/BackendNotifications.hs
```diff
@@ -89,17 +89,24 @@ routingKey t = "backend-notifications." <> t
 -- they are stored in Rabbit.
 type DefederationDomain = Domain
 
+-- | FUTUREWORK: This really doesn't belong here in a module about pushing backend-to-backend
+-- notifications.  Maybe "Wire.API.Federation.Defederate"?  (Same for `ensureTLQueue`.)
```
akshaymankar, line 93: Why not create the module right now?

changelog.d/0-release-notes/WPB-3796
```diff
@@ -0,0 +1,5 @@
+New config value `background-worker.backendNotificationPusher.remotesRefreshIntervalMs`.  If in doubt, set to 30000 ms.  (This controls the delay between backend-to-backend notification queue update pulls from rabbitMQ to background-worker.)
+
+Remote federation domains in the config file are no longer supported (deprecated since Charts...).  Before deploying this release, make sure you have completed the migration steps from [the CHANGELOG of that release](...).
```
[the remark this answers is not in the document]
fisx: incorrect.  we released this last week or so.

changelog.d/0-release-notes/WPB-3796
```diff
@@ -0,0 +1,5 @@
...
+
+If you have enabled federation and therefore rabbitMQ before this upgrade, you may need to remove all queues called `backend-notifications.delete-federation` manually, or this release won't start.
```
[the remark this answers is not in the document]
fisx: the thing that fails is the line below the one where the queue is created: it chokes on the fact that the queue has an invalid domain name and gets trapped in a loop, i think.  but you are right, my wording was inaccurate.

charts/background-worker/values.yaml
fisx: *sigh*

i really don't like the way we do config files.  very brittle and fragile at the same time.

i'll remove the Ms.

libs/wire-api-federation/src/Wire/API/Federation/BackendNotifications.hs
```diff
@@ -89,17 +89,24 @@ routingKey t = "backend-notifications." <> t
...
+ensureB2BQueue :: Q.Channel -> Text -> IO ()
+ensureB2BQueue chan = ensureTLQueue chan . routingKey
+
 -- | If you ever change this function and modify
 -- queue parameters, know that it will start failing in the
 -- next release! So be prepared to write migrations.
-ensureQueue :: Q.Channel -> Text -> IO ()
-ensureQueue chan queue = do
+ensureTLQueue :: Q.Channel -> Text -> IO ()
+ensureTLQueue chan queue = do
```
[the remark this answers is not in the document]
fisx: "top level".  not great, yes.

do you have a suggestion what to use for a name?

i'll think of something...

services/background-worker/src/Wire/BackendNotificationPusher.hs
```diff
@@ -132,29 +131,15 @@ startPusher consumersRef chan = do
...
-        -- Wait for a new set of domains. This is a blocking action
-        -- so we will only move past here when we get a new set of domains.
-        -- It is a bit nicer than having another timeout value, as Brig is
-        -- already providing one in the domain update message.
-        chanRemotes <- liftIO $ readChan env.remoteDomainsChan
-        -- Make new consumers for the new domains, clean up old ones from the consumer map.
-        ensureConsumers consumersRef chan $ domain <$> chanRemotes.remotes
+        remoteDomains <- getRemoteDomains
+        ensureConsumers consumers chan remoteDomains
+        threadDelay env.backendNotificationsConfig.remotesRefreshInterval
```
fisx, line 138: this is maybe the most interesting change in this PR.  it changes back from calling brig for remotes info to calling rabbitMQ, and leaving it to the other services to create the queue if they want to push to it.

fisx (commented):
I went through all changes again, pushed a few small changes, and I think it's ready to merge now.  Let's see what concourse thinks...

fisx:
fails locally in brig-integration:

```
      Exception: Timeout: No matching notification received.
      Use -p '/POST \/federation\/on-user-deleted-connections : 200/' to rerun this test only.
```

fisx:
> fails locally in brig-integration:
> 
> ```
>       Exception: Timeout: No matching notification received.
>       Use -p '/POST \/federation\/on-user-deleted-connections : 200/' to rerun this test only.
> ```

can't reproduce...

charts/brig/templates/tests/configmap.yaml
```diff
@@ -58,6 +58,8 @@ data:
       cert: /etc/wire/integration-secrets/provider-publiccert.pem
       botHost: https://brig-integration
 
+    originDomain: 612c6a18-4e3e-11ee-a205-97507f092421.example.com
```
elland, line 61: Having this hard-coded like this raises an eyebrow. Why is this not dynamic?

charts/integration/templates/integration-integration.yaml
```diff
@@ -207,3 +207,7 @@ spec:
         secretKeyRef:
           name: brig
           key: rabbitmqPassword
+    - name: TEST_INCLUDE
```
elland, line 210: Should probably remove this before merging? :)

fisx:
@akshaymankar this can be closed, right?  i'll close it!  https://github.com/wireapp/wire-server/pull/3588

PR #3527, opened 2023-08-22
Introduce API v5

pcapriotti:
This PR introduces v5 as a new development version, while keeping v4 intact (and still development).

## Checklist

### DIFF
--- a/libs/wire-api/src/Wire/API/Routes/Version.hs
+++ b/libs/wire-api/src/Wire/API/Routes/Version.hs
@@ -22,8 +22,8 @@
 module Wire.API.Routes.Version
   ( -- * API version endpoint
     VersionAPI,
+    VersionAPITag,
     VersionInfo (..),
-    versionSwagger,
     versionHeader,
     VersionHeader,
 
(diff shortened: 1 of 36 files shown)


libs/wire-api/src/Wire/API/Routes/Version.hs
```diff
@@ -179,7 +183,88 @@ type VersionAPI =
...
 $(genSingletons [''Version])
+
+$( promote
+     [d|
+       isDevelopmentVersion :: Version -> Bool
+       isDevelopmentVersion V0 = False
+       isDevelopmentVersion V1 = False
+       isDevelopmentVersion V2 = False
+       isDevelopmentVersion V3 = False
+       isDevelopmentVersion _ = True
```
stefanwire, line 199: Shouldn't `V4` become `False` in this PR?

stefanwire (approved):

libs/wire-api/src/Wire/API/Routes/Version.hs
mdimjasevic: I don't think so because the plan is to first have V4 and V5 co-exist as development versions, and only in the next step/PR we finalize V4.

pcapriotti: That's right, v4 is still development. That's kind of the point of this PR. But that reminds me that we don't need to promote this function any more. It was just part of a failed attempt at removing some of the code duplication.

mdimjasevic (approved):

PR #3528, opened 2023-08-22
Fix flaky tests: Wait for federator ingress health check

smatting:
- abstract ClientCertificate
- client wip

## Checklist


smatting:
Closing this, because this wasn't the problem. We assumed that the flakes we exprienced were to the ingress to the federator not being ready, but that didn't turn out to be the case

PR #3529, opened 2023-08-23
integration: Fix testAddingUserNonFullyConnectedFederation and testNotificationsForOfflineBackends

akshaymankar:
https://wearezeta.atlassian.net/browse/WPB-3974

## Checklist

 - [x] ~Add a new entry in an appropriate subdirectory of `changelog.d`~ No changelog, insignificant change.

### DIFF
--- a/integration/test/Test/Conversation.hs
+++ b/integration/test/Test/Conversation.hs
@@ -431,7 +431,9 @@ testAddingUserNonFullyConnectedFederation = do
     alice <- randomUser own def
     bob <- randomUser other def
     charlie <- randomUser dynBackend def
-    mapM_ (connectUsers alice) [bob, charlie]
+    -- We use retryT here so the dynamic federated connection changes can take
+    -- some time to be propagated. Remove after fixing https://wearezeta.atlassian.net/browse/WPB-3797
+    mapM_ (retryT . connectUsers alice) [bob, charlie]
 
     let newConv = defProteus {qualifiedUsers = []}
     conv <- postConversation alice newConv >>= getJSON 201
--- a/integration/test/Test/Conversation.hs
+++ b/integration/test/Test/Conversation.hs
@@ -1,4 +1,5 @@
 {-# OPTIONS_GHC -Wno-ambiguous-fields #-}
+{-# OPTIONS_GHC -Wno-incomplete-uni-patterns #-}
 
 module Test.Conversation where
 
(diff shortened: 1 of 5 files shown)


integration/test/Test/Conversation.hs
```diff
@@ -410,19 +411,33 @@ testAddUnreachable = do
...
+    own <- asString OwnDomain
+    other <- asString OtherDomain
+
+    -- Ensure that dynamic backend only federates with own domain, but not other
+    -- domain.
+    --
+    -- FUTUREWORK: deleteAllFedConns at the time of acquiring a backend, so
+    -- tests don't affect each other.
+    deleteAllFedConns dynBackend
+    void $ createFedConn dynBackend (FedConn own "full_search")
```
mdimjasevic, line 429: Can you please explain how this dance of deleting and then immediately making a connection works?

akshaymankar: If you have two tests: one of them connects `dynBackend1` with all backends and second one only connects `dynBackend`1 with `OwnDomain`. if the first test runs first, the second test will always fail. So, the second test must always delete all the connections `dynBackend1` has and then only create a connection with `OwnDomain` to ensure that it is not affected by the first test.

mdimjasevic: I see.

Wasn't the plan to provide more isolation between test executions? This dance within the test is very fragile, it incurs what I hope can be called a needless brain load, and is not a general solution.

Is there something we can do in the (dynamic) backend setup that would make them isolated from each other (even if we go with e.g. randomly assigning identifiers (names) and such, which should be good enough if they are very unlikely to collide)?

akshaymankar: Yes, that's why I added the FUTUREWORK. While claiming an environment we could simply truncate the table which keeps this information around.

integration/test/Test/Federation.hs
```diff
@@ -68,12 +69,13 @@ testNotificationsForOfflineBackends = do
     bindResponse (postConversation delUser (defProteus {qualifiedUsers = [otherUser, downUser1]})) $ \resp ->
       resp.status `shouldMatchInt` 533
 
-    -- Adding users to an up backend conversation should work even when one of
-    -- the participating backends is down
-    otherUser2 <- randomUser OtherDomain def
-    connectUsers delUser otherUser2
-    bindResponse (addMembers delUser upBackendConv [otherUser2]) $ \resp ->
-      resp.status `shouldMatchInt` 200
+    -- Adding users to an up backend conversation should work not work when one
```
mdimjasevic, line 72: "should work not work" - must be a left-over in here, but you tell which is which.

akshaymankar: Good catch, It is "should not work".

integration/test/Test/Federation.hs
```diff
@@ -103,23 +108,17 @@ testNotificationsForOfflineBackends = do
...
                 isNotifForUser delUser
               ]
       void $ awaitNotification downUser1 downClient1 (Just newMsgNotif) 1 isDelUserLeaveDownConvNotif
-      void $ awaitNotification otherUser otherClient noValue 1 isDelUserLeaveDownConvNotif
 
       -- FUTUREWORK: Uncomment after fixing this bug: https://wearezeta.atlassian.net/browse/WPB-3664
-      -- void $ awaitNotification downUser1 downClient1 (Just newMsgNotif) 1 (allPreds [isConvLeaveNotif, isNotifConv upBackendConv, isNotifForUser otherUser])
-      -- void $ awaitNotification downUser1 downClient1 (Just newMsgNotif) 1 (allPreds [isConvLeaveNotif, isNotifConv upBackendConv, isNotifForUser delUser])
+      -- void $ awaitNotification downUser1 downClient1 (Just newMsgNotif) 1 isOtherUser2LeaveUpConvNotif
+      -- void $ awaitNotification otherUser otherClient (Just newMsgNotif) 1 isDelUserLeaveDownConvNotif
```
mdimjasevic, line 121: Are you planning to uncomment or remove these two lines?

mdimjasevic (approved):
Looks good to me, but please do something with the two commented out lines at the bottom of changes in the test module `Federation.hs`.

integration/test/Test/Federation.hs
akshaymankar: Yes, see the comment above them. These don't work due to a bug: https://wearezeta.atlassian.net/browse/WPB-3797

PR #3530, opened 2023-08-23
Do not cache federation remote configs on non-brig services

fisx:
https://wearezeta.atlassian.net/browse/WPB-3797

## Checklist


services/federator/src/Federator/Response.hs
```diff
@@ -168,14 +172,24 @@ runFederator env =
...
 
+getFederationDomainConfigs :: Env -> IO FedUp.FederationDomainConfigs
+getFederationDomainConfigs env = do
+  let mgr = env ^. httpManager
+      Endpoint h p = env ^. service $ Brig
+      baseurl = BaseUrl Http (cs h) (fromIntegral p) ""
+      clientEnv = mkClientEnv mgr baseurl
+  FedUp.getFederationDomainConfigs clientEnv >>= \case
+    Right v -> pure v
+    Left e -> error $ show e
```
fisx, line 191: i guess there is a way in polysemy to throw this through an error effect?

lepsa: @fisx Something like this would work. The exact error value could be different, I extended `ServerError` with this new value.

```
getFederationDomainConfigs ::
  ( Member (Embed IO) r
  , Member (Error ServerError) r
  ) =>
  Env ->
  Sem r FedUp.FederationDomainConfigs
getFederationDomainConfigs env = do
  let mgr = env ^. httpManager
      Endpoint h p = env ^. service $ Brig
      baseurl = BaseUrl Http (cs h) (fromIntegral p) ""
      clientEnv = mkClientEnv mgr baseurl
  liftIO (FedUp.getFederationDomainConfigs clientEnv) >>= \case
    Right v -> pure v
    Left e -> throw . InternalError . pack $ show e
   ```

lepsa: > 

If in doubt, shove it into config. Maybe a total retry time?

fisx:
stale, closed.

PR #3537, opened 2023-08-24
[WPB-3664] Bug fix: Notify remote backends of their users removed from conversation when reachable again

mdimjasevic:
A backend that was unreachable, but is reachable now would not get a notification of their user removed from a remote conversation. This PR fixes that bug.

The old way of integration tests does not support testing message queueing among backends. Because of that, a number of integration tests were migrated to the new test framework.

Tracked by https://wearezeta.atlassian.net/browse/WPB-3664.

## Checklist


pcapriotti (approved):
LGTM

PR #3541, opened 2023-08-24
federator: Regenerate certs used by unit-tests

akshaymankar:
Also unmark tests wrongly marked flaky. 

## Checklist

 - [x] ~Add a new entry in an appropriate subdirectory of `changelog.d`~ No changelog, insignificant change.


akshaymankar:
Weird mystery. When I try to run the tests locally (using nix), it fails. In CI they succeed. In CI nix seems to make `/etc/resolv.conf` accessible for the build, but for my local setup this is not the case. @flokli Do you maybe have any ideas about this?

flokli:
This has been discussed recently - `/etc/resolv.conf` is only one part of NSS, and `localhost.` (with the dot) resolution is only done by `nss-myhostname`, (see this [glibc bug](https://sourceware.org/bugzilla/show_bug.cgi?id=27851)), which is a non-glibc module, so nix-built binaries (such as `wire-server`) won't be able to load it on non-NixOS systems for the reasons described in https://flokli.de/posts/2022-11-18-nsncd/.

The fix would be to either add ns(n)cd to your local machine, or find another way to provide these DNS records, like nip.io.

PR #3546, opened 2023-08-28
WPB 2784 multi ingress guest links

smatting:
https://wearezeta.atlassian.net/browse/WPB-2784

### DIFF
--- a/libs/wire-api/src/Wire/API/Routes/Public/Galley/Conversation.hs
+++ b/libs/wire-api/src/Wire/API/Routes/Public/Galley/Conversation.hs
@@ -693,6 +693,7 @@ type ConversationAPI =
                :> CanThrow 'GuestLinksDisabled
                :> CanThrow 'CreateConversationCodeConflict
                :> ZUser
+               :> ZHostOpt
                :> ZOptConn
                :> "conversations"
                :> Capture' '[Description "Conversation ID"] "cnv" ConvId
@@ -737,6 +738,7 @@ type ConversationAPI =
                :> CanThrow 'ConvAccessDenied
                :> CanThrow 'ConvNotFound
                :> CanThrow 'GuestLinksDisabled
+               :> ZHostOpt
                :> ZLocalUser
                :> "conversations"
                :> Capture' '[Description "Conversation ID"] "cnv" ConvId
--- a/integration/test/API/Galley.hs
+++ b/integration/test/API/Galley.hs
@@ -42,6 +42,13 @@ defProteus =
 defMLS :: CreateConv
 defMLS = defProteus {protocol = "mls"}
 
+allowGuests :: CreateConv -> CreateConv
+allowGuests cc =
+  cc
+    { access = Just ["code"],
+      accessRole = Just ["team_member", "guest"]
+    }
+
 instance MakesValue CreateConv where
   make cc = do
     quids <- for (cc.qualifiedUsers) objQidObject
@@ -229,12 +236,14 @@ postConversationCode ::
   user ->
   conv ->
   Maybe String ->
+  Maybe String ->
   App Response
-postConversationCode user conv mbpassword = do
+postConversationCode user conv mbpassword mbZHost = do
   convId <- objId conv
   req <- baseRequest user Galley Versioned (joinHttpPath ["conversations", convId, "code"])
   submit
     "POST"
     ( req
         & addJSONObject ["password" .= pw | pw <- maybeToList mbpassword]
+        & maybe id zHost mbZHost
     )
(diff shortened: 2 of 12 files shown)


docs/src/developer/reference/config-options.md
```diff
@@ -657,39 +655,71 @@ The default setting is that no API version is disabled.
...
-#### Multi-Ingress setup
+
+## Multi-Ingress setup
 
 In a multi-ingress setup the backend is reachable via several domains, each
 handled by a separate Kubernetes ingress. This is useful to obfuscate the
 relationship of clients to each other, as an attacker on TCP/IP-level could only
 see domains and IPs that do not obviously relate to each other.
+Each of these backend domains represents a virtual backend. N.B. these backend
+domains are *DNS domains* only, not to be confused of the "backend domain" term used for federation (see {ref}`configure-federation`). In single-ingress setups the backend dns domain and federation backend domain is usually be the same, but this is not true for multi-ingress setups.
```
supersven, line 671:
```suggestion
domains are *DNS domains* only, not to be confused of the "backend domain" term used for federation (see {ref}`configure-federation`). In single-ingress setups the backend DNS domain and federation backend domain is usually be the same, but this is not true for multi-ingress setups.
```

docs/src/developer/reference/config-options.md
```diff
@@ -724,45 +754,14 @@ Link to diagram:
...
+### Galley
 
-### Multi-Ingress setup
-
-*cannon* sets [CORS
-headers](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS) for direct API
-accesses by clients. To generate them for multiple domains (usually, *cannon*
-works with only one root domain) these need to be defined with
-`nginx_conf.additional_external_env_domains`.
+For conversation invite links to be correct in a multi-ingress setup `settings.multiIngress` needs to be configured as map from `Z-Host` to the conversation uri prefix.  This setting is a `Z-Host` depended version of `settings.conversationCodeURI`. In fact `settings.multiIngress` and `settings.conversationCodeURI` are mutually exclusive.
```
supersven, line 759:
```suggestion
For conversation invite links to be correct in a multi-ingress setup `settings.multiIngress` needs to be configured as map from `Z-Host` to the conversation URI prefix.  This setting is a `Z-Host` depended version of `settings.conversationCodeURI`. In fact `settings.multiIngress` and `settings.conversationCodeURI` are mutually exclusive.
```

libs/wire-api/src/Wire/API/Routes/Public/Galley/Conversation.hs
```diff
@@ -693,6 +693,7 @@ type ConversationAPI =
                :> CanThrow 'GuestLinksDisabled
                :> CanThrow 'CreateConversationCodeConflict
                :> ZUser
+               :> ZHostOpt
```
supersven, line 696: The versioning is unfortunately something to think about: The deployment is planned to happen mid of September. Will all clients use `v4` for this endpoint then? :thinking: 

For the `cargohold` part I've added the `ZHost` to all versions. Hand-wavy justifying that this is not really a change to the API because it happens between `nginz` and the backend.

services/cargohold/src/CargoHold/Options.hs
```diff
@@ -104,7 +104,7 @@ data AWSOpts = AWSOpts
     --
     -- This logic is: If the @Z-Host@ header is provided and found in this map,
     -- the map's values is taken as s3 download endpoint to redirect to;
-    -- otherwise, `_awsS3DownloadEndpoint` is used. This option is only useful
+    -- otherwise an 302 error is thrown. This option is only useful
```
supersven, line 107:
```suggestion
    -- otherwise an 404 error is thrown. This option is only useful
```

Thanks for adjusting this outdated doc!

However, I think a `NoMatchingAssetEndpoint` is thrown in this case (map defined, but host not found in it.); which translates to a `404`. :thinking:

supersven: The `302` you may haven seen in e.g. tests is likely a redirect to the assets host (e.g. AWS S3, MinIO, ...)

supersven (changes requested):
Looking good :+1: 

However, there are a few bits left to discuss...

docs/src/developer/reference/config-options.md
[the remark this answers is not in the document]
smatting: Yes, but this is not a plain markdown file, it's MyST, which is what sphinx uses by default. All directives that are available in sphinx reStructuredText dialect are also available in MyST.

services/cargohold/src/CargoHold/Options.hs
smatting: Ouch, yes of course. 302 is als not an error.

libs/wire-api/src/Wire/API/Routes/Public/Galley/Conversation.hs
smatting: Right, I don't think we can rely on `v4` being used. I've updated also the v3 api. I think we can handwave this again with the same arguments.

supersven (approved):
LGTM :+1:

PR #3558, opened 2023-09-04
Fix ES migration script.

fisx:
https://wearezeta.atlassian.net/browse/WPB-4425

## Checklist

### DIFF
--- a/services/brig/src/Brig/Index/Migrations.hs
+++ b/services/brig/src/Brig/Index/Migrations.hs
@@ -120,14 +120,14 @@ failIfIndexAbsent targetIndex =
 
 -- | Runs only the migrations which need to run
 runMigration :: MigrationVersion -> MigrationActionT IO ()
-runMigration ver = do
-  vmax <- latestMigrationVersion
-  if ver > vmax
+runMigration expectedVersion = do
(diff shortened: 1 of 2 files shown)


elland (approved):

services/brig/src/Brig/Index/Migrations.hs
```diff
@@ -120,14 +120,14 @@ failIfIndexAbsent targetIndex =
...
-  if ver > vmax
+runMigration expectedVersion = do
+  foundVersion <- latestMigrationVersion
+  if expectedVersion > foundVersion
     then do
       Log.info $
         Log.msg (Log.val "Migration necessary.")
-          . Log.field "expectedVersion" vmax
-          . Log.field "foundVersion" ver
+          . Log.field "expectedVersion" ver
```
supersven, line 129: Shouldn't this be 

```suggestion
          . Log.field "expectedVersion" expectedVersion
```
? :thinking: 

(I cannot see where `ver` comes from.)

elland: Somehow I missed this 😅

fisx: good thing we have sven and the compiler!  :)

PR #3559, opened 2023-09-04
[WPB-4361] upgrade jwt-tools

battermann:
## Checklist

### DIFF
--- a/nix/overlay.nix
+++ b/nix/overlay.nix
@@ -49,15 +49,18 @@ let
       '';
     };
 
+  sources = import ./sources.nix;
+  pkgsCargo = import sources.nixpkgs-cargo {};
 in
 
 self: super: {
+
   cryptobox = self.callPackage ./pkgs/cryptobox { };
   zauth = self.callPackage ./pkgs/zauth { };
   mls-test-cli = self.callPackage ./pkgs/mls-test-cli { };
 
   # Named like this so cabal2nix can find it
-  rusty_jwt_tools_ffi = self.callPackage ./pkgs/rusty_jwt_tools_ffi { };
+  rusty_jwt_tools_ffi = self.callPackage ./pkgs/rusty_jwt_tools_ffi { rustPlatform = pkgsCargo.rustPlatform; };
 
   nginxModules = super.nginxModules // {
     zauth = {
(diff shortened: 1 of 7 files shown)


nix/pkgs/rusty_jwt_tools_ffi/default.nix
```diff
@@ -5,17 +5,28 @@
...
 }:
-
 let
-  version = "0.3.4";
+  sources = import ../../sources.nix;
+in
+with import sources.nixpkgs
+{
+  overlays = [
+    (import (fetchTarball "https://github.com/oxalica/rust-overlay/archive/master.tar.gz"))
```
flokli, line 14: This is not a stable pin, but will change - and we don't really need a rust overlay either.

I'd propose adding a different (known-working) nixpkgs pin to `nix/sources.json` instead, and use `cargo` and `rustc` from there instead.

nix/overlay.nix
```diff
@@ -49,15 +49,18 @@ let
...
 
 self: super: {
+
   cryptobox = self.callPackage ./pkgs/cryptobox { };
   zauth = self.callPackage ./pkgs/zauth { };
   mls-test-cli = self.callPackage ./pkgs/mls-test-cli { };
 
   # Named like this so cabal2nix can find it
-  rusty_jwt_tools_ffi = self.callPackage ./pkgs/rusty_jwt_tools_ffi { };
+  rusty_jwt_tools_ffi = self.callPackage ./pkgs/rusty_jwt_tools_ffi { rustPlatform = pkgsCargo.rustPlatform; };
```
flokli, line 63:
```suggestion
  rusty_jwt_tools_ffi = self.callPackage ./pkgs/rusty_jwt_tools_ffi {
    inherit (pkgsCargo) rustPlatform;
  };
```

nix/overlay.nix
```diff
@@ -49,15 +49,18 @@ let
       '';
     };
 
+  sources = import ./sources.nix;
+  pkgsCargo = import sources.nixpkgs-cargo {};
```
flokli, line 53: I'm not sure if we (still) do some custom Darwin stuff when instantiating the regular nixpkgs, you might want to compare it with the invocation there. But we can also deal with it in a followup, no need to block this PR.

PR #3570, opened 2023-09-08
WPB-4240: Migrate from swagger2 to openapi3

lepsa:
https://wearezeta.atlassian.net/browse/WPB-4240

## Changes
- Migrating from the swagger2 to openapi3 library.
  - This is a fairly straight forward migration, as the libraries are very similar.
  - The biggest change is how federation calls are now tracked. Extensions aren't supported in `openapi3` yet, so I've used tags for the same effect.
- Some of the schema derivations were triggering compiler bugs. These bugs were worked around by manually inlining the code used by the instance derivation and carefully picking constraints.
- Adding deprecated flags to routes and types where the deprecation was already noted in the documentation. This is basically a formalization of the existing deprecation, but now in a machine-readable way.

## Checklist


libs/types-common/src/Data/CommaSeparatedList.hs
```diff
@@ -40,10 +40,10 @@ instance FromByteString (List a) => FromHttpApiData (CommaSeparatedList a) where
...
 instance ToParamSchema (CommaSeparatedList a) where
-  toParamSchema _ = mempty & type_ ?~ SwaggerString
+  toParamSchema _ = mempty & type_ ?~ OpenApiString
 
 -- | TODO: is this obsoleted by the instances in "Data.Range"?
 instance (ToParamSchema a, ToParamSchema (Range n m [a])) => ToParamSchema (Range n m (CommaSeparatedList a)) where
   toParamSchema _ =
     toParamSchema (Proxy @(Range n m [a]))
-      & items ?~ SwaggerItemsPrimitive (Just CollectionCSV) (toParamSchema (Proxy @a))
+      & items ?~ OpenApiItemsArray [Inline $ toParamSchema (Proxy @a)]
```
lepsa, line 49: There is no direct equivalent to `SwaggerItemsPrimitive` in openapi3. OpenApiItemsArray seemed the closest.

nix/haskell-pins.nix
```diff
@@ -180,6 +173,13 @@ let
         sha256 = "sha256-SKEE9ZqhjBxHYUKQaoB4IpN4/Ui3tS4S98FgZqj7WlY=";
       };
     };
+    servant-openapi3 = {
+      src = fetchgit {
+        url = "https://github.com/lepsa/servant-openapi3";
```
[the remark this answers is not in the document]
fisx: can you put this comment into the code next time so it won't get lost to future us?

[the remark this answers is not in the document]
lepsa: Done!

libs/wire-api/src/Wire/API/SwaggerHelper.hs
```diff
@@ -47,3 +57,49 @@ cleanupSwagger =
       (S.properties . traverse . S._Inline %~ sanitise)
         . (S.required %~ nubOrd)
         . (S.enum_ . _Just %~ nub)
+    -- servant-openapi and servant-swagger both insert default responses with codes 404 and 400.
+    -- They have a simple structure that we can match against, and remove from the final structure.
+    stripDefaultErrors :: HttpStatusCode -> Referenced Response -> Responses' -> Responses'
```
elland, line 62: Since we're already patching the library, would it make sense to also patch in settings to disable this default responses?

fisx: i'd prefer to do that sepately.  this change is big enough, and we have had odd defaults until now, too.

elland (approved):

PR #3579, opened 2023-09-13
Save MLS migration statistics in S3 bucket

stefanwire:
## Checklist


charts/mlsstats/templates/cronjob.yaml
```diff
@@ -0,0 +1,59 @@
...
+  concurrencyPolicy: Forbid
+  schedule: {{ .Values.schedule | quote }}
+  jobTemplate:
+    metadata:
+      labels:
+        app: mlsstats
+        release: {{ .Release.Name }}
+      # annotations:
+      #   # An annotation of the configmap checksum ensures changes to the configmap cause a redeployment upon `helm upgrade`
+      #   checksum/secret: {{ include (print .Template.BasePath "/secret.yaml") . | sha256sum }}
```
akshaymankar, line 20: Why is this commented out? This looks useful.

charts/mlsstats/templates/cronjob.yaml
```diff
@@ -0,0 +1,59 @@
...
+    spec:
+      backoffLimit: 0
+      template:
+        spec:
+          restartPolicy: Never
+          # volumes:
+          #   - name: "ldap-scim-bridge-config"
+          #     secret:
+          #       secretName: {{ .Release.Name }}
```
akshaymankar, line 29:
```suggestion
```
dead code.

charts/mlsstats/templates/cronjob.yaml
```diff
@@ -0,0 +1,59 @@
...
+                      , "--cassandra-pagesize", {{ .Values.cassandra.pagesize | quote }}
+                      {{if .Values.s3.host -}}
+                      , "--s3-host", {{ .Values.s3.host | quote -}}
+                      {{- end -}}
+                      {{if .Values.s3.port -}}
+                      , "--s3-port", {{ .Values.s3.port | quote -}}
+                      {{- end -}}
+                      {{if .Values.s3.region -}}
+                      , "--s3-region", {{ .Values.s3.region | quote -}}
+                      {{- end -}}
```
akshaymankar, line 48: Can we just have defaults written down in values.yaml instead of these ifs. I think defaulting to real AWS endpoints would be good enough.

charts/mlsstats/templates/cronjob.yaml
```diff
@@ -0,0 +1,59 @@
+apiVersion: {{ include "batch.apiVersion" . }}
```
akshaymankar, line 1: I don't see S3 creds being set as env vars, how does this work?

charts/mlsstats/templates/secret.yaml
```diff
@@ -0,0 +1,13 @@
...
+  name: {{ .Release.Name }}
+  labels:
+    app: ldap-scim-bridge
+    chart: "{{ .Chart.Name }}-{{ .Chart.Version }}"
+    release: "{{ .Release.Name }}"
+    heritage: "{{ .Release.Service }}"
+type: Opaque
+stringData:
+  config.yaml: |
+{{ toYaml .Values.config | indent 4 }}
```
akshaymankar, line 13: Is the whole config a secret? Usually we use `.Values.secrets` and write each secret here. It is more work but makes things clearer for someone figuring out what they need to provide.
I also don't see the structure of this config specified in the `values.yaml`, so not sure what needs to be provided by the user.

tools/mlsstats/exec/Main.hs
```diff
@@ -0,0 +1,36 @@
...
+import Options.Applicative
+
+main :: IO ()
+main = do
+  opts <- execParser (info (helper <*> optsParser) desc)
+  run opts
+  where
+    desc =
+      header "mlsstats"
+        <> progDesc "MlsStats - Proteus to MLS migration statistics for admins"
```
akshaymankar, line 35:
```suggestion
        <> progDesc "MLS Stats - Proteus to MLS migration statistics for admins"
```

or it should be `mlsstats`, `MlsStats` is neither the name of the binary nor how a user will see it as.

charts/wire-server/requirements.yaml
```diff
@@ -129,3 +129,8 @@ dependencies:
   repository: "file://../integration"
   tags:
     - integration
+- name: mlsstats
+  version: "0.0.42"
+  repository: "file://../mlsstats"
+  tags:
+    - integration
```
akshaymankar, line 136: We shouldn' use the `integration` tag for this because this is not integration tests. 
Let's call the tag `mlsstats`, then edit `charts/wire-server/values.yaml` and disable this by default. Add a comment there saying it is not a good idea because of the reasons we also mentioned in the changelog.

hack/helmfile.yaml
```diff
@@ -135,6 +135,8 @@ releases:
         value: {{ .Values.dynBackendDomain2 }}
       - name: brig.config.optSettings.setFederationDomainConfigs[4].domain
         value: {{ .Values.dynBackendDomain3 }}
+      - name: mlsstats.federationDomain
+        value: {{ .Values.federationDomain1 }}
```
akshaymankar, line 139: We probably don't need to enable this component in integration tests because it will probably just not run. So this config won't be required.

hack/helmfile.yaml
```diff
@@ -159,5 +161,7 @@ releases:
         value: {{ .Values.dynBackendDomain2 }}
       - name: brig.config.optSettings.setFederationDomainConfigs[4].domain
         value: {{ .Values.dynBackendDomain3 }}
+      - name: mlsstats.federationDomain
+        value: {{ .Values.federationDomain2 }}
```
akshaymankar, line 165: same as above.

akshaymankar (commented):
Reviewed the helm things, reviewing haskell next.

charts/mlsstats/templates/cronjob.yaml
```diff
@@ -0,0 +1,59 @@
...
+          #   - name: "ldap-scim-bridge-config"
+          #     secret:
+          #       secretName: {{ .Release.Name }}
+          containers:
+            - name: mlsstats
+              image: "{{ .Values.image.repository }}:{{ .Values.image.tag }}"
+              imagePullPolicy: {{ default "" .Values.imagePullPolicy | quote }}
+              args: [ "mlsstats"
+                      , "--cassandra-host", {{ .Values.cassandra.host | quote }}
+                      , "--cassandra-port", {{ .Values.cassandra.port | quote }}
```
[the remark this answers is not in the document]
stefanwire: Didn't know. 👍

charts/mlsstats/templates/cronjob.yaml
stefanwire: Good point, we need to set them the same way as we set them for prod. Will check and fix.

tools/mlsstats/src/MlsStats/Run.hs
```diff
@@ -0,0 +1,247 @@
...
+  env <-
+    maybe id (\reg e -> (e :: Env) {region = Region' reg}) s3.region
+      . (\e -> e {logger = logger})
+      . configureService service
+      <$> newEnv discover
+  runResourceT $ do
+    uploadStream env (BucketName s3.bucketName) "user-client.csv" (userClient brigTables queryPageSize)
+    uploadStream env (BucketName s3.bucketName) "conv-group-team-protocol.csv" (convGroupTeamProtocol galleyTables queryPageSize)
+    uploadStream env (BucketName s3.bucketName) "domain-user-client-group.csv" (domainUserClientGroup mOwnDomain galleyTables queryPageSize)
+    uploadStream env (BucketName s3.bucketName) "user-conv.csv" (userConv galleyTables queryPageSize)
```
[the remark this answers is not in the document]
akshaymankar: Another problem I see with this is that we seem to assume that we will always upload things to root of a bucket, I think we should allow uploads to some directory inside the bucket too.

akshaymankar (commented):
Also reviewed haskal

hack/helmfile.yaml
stefanwire: I would need the own domain name here to filter (even though that's an optimisation). Where would I get from in prod?

akshaymankar (approved):

PR #3580, opened 2023-09-13
WPB-4559 user receives 2 k federation connection removed on anta

battermann:
https://wearezeta.atlassian.net/browse/WPB-4559

## Checklist

### DIFF
--- a/services/galley/src/Galley/API/BackgroundProcesses.hs
+++ b/services/galley/src/Galley/API/BackgroundProcesses.hs
@@ -1,0 +1,127 @@
+-- This file is part of the Wire Server implementation.
+--
+-- Copyright (C) 2022 Wire Swiss GmbH <opensource@wire.com>
+--
+-- This program is free software: you can redistribute it and/or modify it under
+-- the terms of the GNU Affero General Public License as published by the Free
+-- Software Foundation, either version 3 of the License, or (at your option) any
+-- later version.
+--
+-- This program is distributed in the hope that it will be useful, but WITHOUT
+-- ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+-- FOR A PARTICULAR PURPOSE. See the GNU Affero General Public License for more
+-- details.
+--
+-- You should have received a copy of the GNU Affero General Public License along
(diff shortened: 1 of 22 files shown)


services/galley/src/Galley/API/BackgroundProcesses.hs
```diff
@@ -0,0 +1,127 @@
...
+--
+-- This program is distributed in the hope that it will be useful, but WITHOUT
+-- ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+-- FOR A PARTICULAR PURPOSE. See the GNU Affero General Public License for more
+-- details.
+--
+-- You should have received a copy of the GNU Affero General Public License along
+-- with this program. If not, see <https://www.gnu.org/licenses/>.
+
+module Galley.API.BackgroundProcesses where
```
fisx, line 18:
```suggestion
module Galley.BackgroundProcesses where
```

not part of the rest api, right?

services/galley/src/Galley/API/BackgroundProcesses.hs
```diff
@@ -0,0 +1,127 @@
...
+import Galley.Effects.TeamStore qualified as E
+import Galley.Types.Conversations.Members (RemoteMember (RemoteMember, rmId))
+import Galley.Types.UserList
+import Imports hiding (head)
+import Polysemy
+import Polysemy.Input
+import Wire.API.Conversation hiding (Member)
+
+-- | Remove all conversations and members from the local domain. This must only be run in a background process
+unsafeRemoveRemoteMembersFromLocalConversation ::
```
fisx, line 47: why is it `unsafe`?  because it's not fit for HTTP?  then i'd call it `bg` instead.

fisx (commented):
This looks very reasonable, unfortunately it is about to be deprecated by the more robust https://github.com/wireapp/wire-server/pull/3582.

This PR is probably good for code mining when we have a new architecture and implement this again, Galley.API.BackgroundProcesses in particular looks useful.

battermann:
obsolote because of https://github.com/wireapp/wire-server/pull/3582
let's keep the branch though

PR #3585, opened 2023-09-14
WPB-4629 impossible to add users to a conversation if one of the members is from an offline backend

battermann:
https://wearezeta.atlassian.net/browse/WPB-4629

## Checklist

### DIFF
--- a/services/galley/src/Galley/API/Action.hs
+++ b/services/galley/src/Galley/API/Action.hs
@@ -443,19 +443,25 @@ performConversationJoin qusr lconv (ConversationJoin invited role) = do
       Local UserId ->
       Sem r ()
     checkRemoteBackendsConnected lusr = do
-      let invitedRemoteDomains = Set.fromList . filter (/= tDomain lconv) $ tDomain <$> snd (partitionQualified lusr $ NE.toList invited)
+      let remoteUsers = snd (partitionQualified lusr $ NE.toList invited)
+          invitedRemoteDomains = Set.fromList . filter (/= tDomain lconv) $ tDomain <$> remoteUsers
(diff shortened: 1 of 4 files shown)


services/galley/src/Galley/API/Action.hs
```diff
@@ -442,17 +443,25 @@ performConversationJoin qusr lconv (ConversationJoin invited role) = do
...
+          allInvitedAlreadyInConversation = null $ invitedRemoteDomains \\ existingRemoteDomains
+
+      if not allInvitedAlreadyInConversation
+        then -- Note:
+        --
+        -- In some cases, this federation status check might be redundant (for
+        -- example if there are only local users in the conversation). However,
+        -- it is important that we attempt to connect to the backends of the new
+        -- users here, because that results in the correct error when those
+        -- backends are not reachable.
```
pcapriotti, line 458: This comment isn't true anymore.

services/galley/src/Galley/API/Action.hs
```diff
@@ -442,17 +443,25 @@ performConversationJoin qusr lconv (ConversationJoin invited role) = do
...
+          checkFederationStatus (RemoteDomains (invitedRemoteDomains <> existingRemoteDomains))
+        else do
+          -- even if there are no new remotes, we still need to check they are reachable
+          void . (ensureNoUnreachableBackends =<<) $
+            E.runFederatedConcurrentlyEither remoteUsers $ \_ ->
+              void $ fedClient @'Brig @"api-version" ()
```
pcapriotti, line 464: This is weird, because it will result in 2 calls to api-version in the success case. But I guess this "ping" logic already exists elsewhere. Can't it be fixed simply by having `pure ()` as the body of `runFederatedConcurrentlyEither`?

pcapriotti (approved):
I think this is fine. Comments below.

PR #3588, opened 2023-09-19
Fix federationStrategy 'allowAll'

akshaymankar:
https://wearezeta.atlassian.net/browse/WPB-3796

Supersedes #3526

## Checklist

### DIFF
--- a/integration/test/Test/Brig.hs
+++ b/integration/test/Test/Brig.hs
@@ -1,3 +1,5 @@
+{-# OPTIONS_GHC -Wno-incomplete-uni-patterns #-}
+
 module Test.Brig where
 
 import API.Brig qualified as Public
@@ -5,7 +7,6 @@ import API.BrigInternal qualified as Internal
 import API.Common qualified as API
 import API.GalleyInternal qualified as Internal
 import Control.Concurrent (threadDelay)
-import Data.Aeson qualified as Aeson
 import Data.Aeson.Types hiding ((.=))
 import Data.Set qualified as Set
 import Data.String.Conversions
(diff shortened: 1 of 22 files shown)


integration/test/Test/Brig.hs
```diff
@@ -1,11 +1,12 @@
+{-# OPTIONS_GHC -Wno-incomplete-uni-patterns #-}
```
fisx, line 1: i'm also ok with doing this in integration.cabal...

services/brig/test/integration/Federation/End2end.hs
```diff
@@ -114,22 +112,6 @@ spec _brigOpts mg brig galley cargohold cannon _federator brigTwo galleyTwo carg
...
-  resultViaBrigTwo <- getUserInfoFromHandle brigTwo domain handle
-
-  -- query the local-namespace brig for a user sitting on the other backend
-  -- (which will exercise the network traffic via two federators to the remote brig)
-  resultViaBrigOne <- getUserInfoFromHandle brig domain handle
-
-  liftIO $ assertEqual "remote handle lookup via federator should work in the happy case" (profileQualifiedId resultViaBrigOne) (userQualifiedId userBrigTwo)
-  liftIO $ assertEqual "querying brig1 or brig2 about the same user should give same result" resultViaBrigTwo resultViaBrigOne
-
 testSearchUsers :: Brig -> Brig -> Http ()
```
fisx, line 115: this isn't called anywhere, but neither is it translated into `/integration`.  i think running this won't pass for the same reason `testHandleLookup` fails?

integration/test/Test/Demo.hs
```diff
@@ -174,22 +175,10 @@ testIndependentESIndices = do
...
-    [aDynDomain, anotherDynDomain] <- pure dynDomains
-    _ <- Internal.createFedConn anotherDynDomain (Internal.FedConn aDynDomain "full_search")
-    threadDelay 2_000_000
-
-    u1 <- randomUser aDynDomain def
-    u2 <- randomUser anotherDynDomain def
-    uid2 <- objId u2
-    Internal.refreshIndex anotherDynDomain
-
-    bindResponse (Public.searchContacts u1 (u2 %. "name") anotherDynDomain) $ \resp -> do
```
fisx, line 187: did you remove this search intentionally?

fisx (approved):

services/brig/test/integration/Federation/End2end.hs
akshaymankar: Deleted it now.
The test already existed in the new test suite: https://github.com/wireapp/wire-server/blob/1c6f7463d9eacfe68cdc402952e41213869267c0/integration/test/Test/Brig.hs#L144-L160

integration/test/Test/Demo.hs
akshaymankar: Yes, the test is supposed to show that federation works. Using search here is just making the test more complicated.

integration/test/Test/Brig.hs
akshaymankar: Create a PR then.

PR #3589, opened 2023-09-19
WPB-4748 create a way to inspect background notifications queue

battermann:
https://wearezeta.atlassian.net/browse/WPB-4748

## Checklist


tools/rabbitmq-consumer/src/RabbitMQConsumer/Lib.hs
```diff
@@ -0,0 +1,203 @@
...
+  conn <- openConnection' opts.host opts.port opts.vhost opts.username opts.password
+  chan <- openChannel conn
+  qos chan 0 1 False
+  done <- newEmptyMVar
+  runTimerAsync done opts.timeoutSec
+  case opts.cmd of
+    Head -> void $ consumeMsgs chan opts.queue Ack (printHead done opts)
+    Interactive -> void $ consumeMsgs chan opts.queue Ack (interactive done opts)
+    DropHead dhOpts -> void $ consumeMsgs chan opts.queue Ack (dropHead done opts dhOpts)
+  takeMVar done
```
akshaymankar, line 40: Seems wrong to run timer like this for interactive session. Things could timeout when the user is just looking at the JSON.

tools/rabbitmq-consumer/src/RabbitMQConsumer/Lib.hs
```diff
@@ -0,0 +1,203 @@
...
+      void $ tryPutMVar done ()
+
+    displayMessage :: Opts -> Message -> String
+    displayMessage opts msg =
+      intercalate
+        "\n"
+        [ "vhost: " <> cs opts.vhost,
+          "queue: " <> cs opts.queue,
+          "timestamp: " <> show msg.msgTimestamp,
+          "received message: " <> BL.unpack msg.msgBody
```
akshaymankar, line 84: We should try to decode the body as a `Value` and pretty print the json.

akshaymankar (commented):
Looks good overall, some minor comments inline.
Apart from those, could you please add this to the `executablesMap` in `nix/wire-server.nix` so we keep building the docker image for this?
https://github.com/wireapp/wire-server/blob/5f246fbbb96a6e5cdfeaae0e89e9227d6c1d4e62/nix/wire-server.nix#L73-L87

cabal.project
```diff
@@ -119,7 +119,7 @@ package polysemy-wire-zoo
     ghc-options: -Werror
 package proxy
     ghc-options: -Werror
-package rabbitmq-consumer
+package mlsstats
```
akshaymankar: Looks like a mistake

akshaymankar:
Error in CI

```
FATA[0001] trying to reuse blob sha256:4a660d37c91f870fb044f7ac64fd807cd5990d298288183e9b97cbbb228c5d3f at destination: checking whether a blob sha256:4a660d37c91f870fb044f7ac64fd807cd5990d298288183e9b97cbbb228c5d3f exists in quay.io/wire/rabbitmq-consumer: authentication required 
```

Probably means that the new image needs to be made publically readable and also the bot needs rights to be able to push it.

tools/rabbitmq-consumer/rabbitmq-consumer.cabal
```diff
@@ -0,0 +1,84 @@
...
+    , bytestring
+    , imports
+    , network
+    , optparse-applicative
+    , text
+    , types-common
+    , wire-api
+    , wire-api-federation
+
+  default-extensions:
```
elland, line 42: At some point we should go around cleaning up these extensions since a lot of them are on by default with GHC2021, right?

tools/rabbitmq-consumer/src/RabbitMQConsumer/Lib.hs
```diff
@@ -0,0 +1,232 @@
...
+      void $ tryPutMVar done ()
+
+    displayMessage :: Opts -> Message -> String
+    displayMessage opts msg =
+      intercalate
+        "\n"
+        [ "vhost: " <> cs opts.vhost,
+          "queue: " <> cs opts.queue,
+          "timestamp: " <> show msg.msgTimestamp,
+          "received message: \n" <> BL.unpack (maybe msg.msgBody encodePretty (decode @BackendNotification' msg.msgBody))
```
akshaymankar, line 92: We should handle the decode failure here and print the JSON as is with a warning that it failed to parse and the parse error. We can use `eitherDecode` to get the parse error.

akshaymankar: Sorry needed more coffee :coffee:

akshaymankar (approved):
:rocket:

PR #3596, opened 2023-09-21
[WPB-4556] document internal user creation

fisx:
https://wearezeta.atlassian.net/browse/WPB-4556

## Checklist

### DIFF
--- a/hack/bin/create_test_team_admins.sh
+++ b/hack/bin/create_test_team_admins.sh
@@ -12,6 +12,9 @@ USAGE="
 This bash script can be used to create active team admin users and
 their teams.
 
+This is the way to create teams if you have set
+'setRestrictUserCreation' to false in your 'values.yaml'.
+
 Note that this uses an internal brig endpoint.  It is not exposed over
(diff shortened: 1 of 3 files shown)


hack/bin/create_test_team_admins.sh
```diff
@@ -12,6 +12,9 @@ USAGE="
 This bash script can be used to create active team admin users and
 their teams.
 
+This is the way to create teams if you have set
+'setRestrictUserCreation' to false in your 'values.yaml'.
```
paulwire, line 16: should this not be "setRestrictUserCreation" to true - or am I confused here?

paulwire (commented):
Left a comment, regarding the change in create_test_team_admins.sh

hack/bin/create_test_team_admins.sh
fisx: nope. you're right, thanks!  aea091d255423880d37731466b0b7f2d2f8bed0d

paulwire (approved):

PR #3602, opened 2023-09-22
federator-client: Use a new http2 connection on every request

smatting:
Create a new http2 connection in every federator client request instead of using a shared connection via http2-manager.
This is to workaround the issue discovered here https://wearezeta.atlassian.net/browse/WPB-4787


battermann (approved):

battermann:
Changelog is missing

PR #3604, opened 2023-09-25
WPB-1906 - Unverified users can no longer create assets

lepsa:
When users upload assets, ensure that they are verified

Updated tests where these changes broke them.
Added a new test to ensure that a random UID that isn't verified can't upload assets.

## Checklist


mdimjasevic:
Just to help you out with CI, here's a sample of a failing test, but there are a lot more for exactly the same reason:

```
Cargohold

  API Integration

    simple

      roundtrip:                                            FAIL

        Exception: Parse failed: AesonException "Error in $: parsing TestSetup.IntegrationConfig(IntegrationConfig) failed, key \"brig\" not found"

        Falling back to environment variables

          error, called at src/Util/Test.hs:45:3 in types-common-0.16.0-9X9YMCJ9AH3FjwSZfkabep:Util.Test

        Use -p '/API Integration.simple.roundtrip/' to rerun this test only.
```

I haven't looked at it, but at first it sounds to me like you could add JSON golden tests for the culprit type.

lepsa:
> Just to help you out with CI, here's a sample of a failing test, but there are a lot more for exactly the same reason:
> 
> ```
> Cargohold
> 
>   API Integration
> 
>     simple
> 
>       roundtrip:                                            FAIL
> 
>         Exception: Parse failed: AesonException "Error in $: parsing TestSetup.IntegrationConfig(IntegrationConfig) failed, key \"brig\" not found"
> 
>         Falling back to environment variables
> 
>         CallStack (from HasCallStack):
> 
>           error, called at src/Util/Test.hs:45:3 in types-common-0.16.0-9X9YMCJ9AH3FjwSZfkabep:Util.Test
> 
>         Use -p '/API Integration.simple.roundtrip/' to rerun this test only.
> ```
> 
> I haven't looked at it, but at first it sounds to me like you could add JSON golden tests for the culprit type.

I've found another helm chart that might need to be updated, and it isn't listed in the PR guidelines.
`charts/cargohold/templates/tests/configmap.yaml`
It is also missing the federator config, so I've added that too.
Now we wait to see if this was the problem.

libs/wire-api/src/Wire/API/Error/Cargohold.hs
```diff
@@ -38,5 +40,9 @@ type instance MapError 'AssetTooLarge = 'StaticError 413 "client-error" "Asset t
 
 type instance MapError 'InvalidLength = 'StaticError 400 "invalid-length" "Invalid content length"
 
+type instance MapError 'Unverified = 'StaticError 403 "unverified" "Unverified"
+
+type instance MapError 'UserNotFound = 'StaticError 404 "not-found" "User not found"
+
```
akshaymankar, line 46: Semantically this should be a 403 since the HTTP operation is not on the user resource but on the asset resource.

libs/wire-api/src/Wire/API/Error/Cargohold.hs
```diff
@@ -38,5 +40,9 @@ type instance MapError 'AssetTooLarge = 'StaticError 413 "client-error" "Asset t
 
 type instance MapError 'InvalidLength = 'StaticError 400 "invalid-length" "Invalid content length"
 
+type instance MapError 'Unverified = 'StaticError 403 "unverified" "Unverified"
```
akshaymankar, line 43:
```suggestion
type instance MapError 'Unverified = 'StaticError 403 "unverified-user" "Unverified user"
```

services/brig/test/integration/API/User/Account.hs
```diff
@@ -1454,7 +1454,10 @@ testDeleteAnonUser brig = do
 
 testDeleteWithProfilePic :: Brig -> CargoHold -> Http ()
 testDeleteWithProfilePic brig cargohold = do
-  uid <- userId <$> createAnonUser "anon" brig
+  -- A random local part, with nice, email-friendly characters
+  localPart <- cs . show <$> liftIO UUID.nextRandom
+  -- Users need to be verified if they want to upload assets, so email it is!
+  uid <- userId <$> createUserWithEmail "anon" (Email localPart "example.test") brig
```
akshaymankar, line 1460: We should use the `randomEmail` utilitiy here.

services/brig/test/integration/Federation/End2end.hs
```diff
@@ -523,6 +523,11 @@ testSendMessageToRemoteConv brig1 brig2 galley1 galley2 cannon1 = do
           ( OtrMessage bobClient aliceClient (toBase64Text msgText) (Just "")
           )
 
+data NotificationException = NotificationException
+  deriving (Show, Typeable)
+
+instance Exception NotificationException
+
```
akshaymankar, line 530: Is this unused?

services/cargohold/src/CargoHold/API/Public.hs
```diff
@@ -132,13 +144,36 @@ mkAssetLocation key =
...
+  Endpoint ->
   id ->
   AssetSource ->
   Handler (Asset, AssetLocation Relative)
-uploadAssetV3 pid req = do
+uploadAssetV3 (Endpoint h p) pid req = do
   let principal = mkPrincipal pid
+      baseUrl = BaseUrl Http (T.unpack h) (fromIntegral p) ""
+  clientEnv <- liftIO $ newManager defaultManagerSettings <&> \mgr -> ClientEnv mgr baseUrl Nothing defaultMakeClientRequest
+  let getUser uid = runClientM (client' uid) clientEnv
```
akshaymankar, line 155: We should do all of this at boot time for cargohold and keep it in the environment. This way we can avoid having to pass this `Endpoint` here and having to thread it through all the functions above.

services/cargohold/test/integration/API.hs
```diff
@@ -131,6 +132,20 @@ testSimpleRoundtrip = do
...
+  let rets = [minBound ..]
+  let sets = def : map (\r -> def & V3.setAssetRetention ?~ r) rets
+  mapM_ simpleUpload sets
+  where
+    simpleUpload sets = do
+      uid <- liftIO $ Id <$> nextRandom
+      let bdy = (applicationText, "Hello World")
+      void $
+        uploadSimple (path "/assets/v3") uid sets bdy
+          <!! const 403 === statusCode
```
akshaymankar, line 147: All new tests should go into the new test framework.

services/cargohold/test/integration/API/Util.hs
```diff
@@ -18,30 +18,59 @@
...
+  r <- post (Bilge.host eHost . Bilge.port ePort . path "/i/users" . json p)
+  pure
+    . readNote "unable to parse Location header"
+    . C.unpack
+    $ getHeader' "Location" r
+  where
+    mkEmail loc dom = do
+      uid <- nextRandom
+      pure $ loc <> "+" <> UUID.toText uid <> "@" <> dom
+
```
akshaymankar, line 73: We created the new test framework so we can avoid this kind of wholesale copying.

akshaymankar (changes requested):

libs/wire-api/src/Wire/API/Error/Cargohold.hs
lepsa: Good catch. I'll update it and any tests that depend on it.

services/brig/test/integration/API/User/Account.hs
lepsa: Will do!

services/brig/test/integration/Federation/End2end.hs
lepsa: Yes, I think it was from an attempt to work with Multiverb

services/cargohold/test/integration/API/Util.hs
lepsa: This is needed to support existing legacy tests. The change to the API breaks them, as they were using random IDs that weren't marked as verified. This function sets up a new user to be verified.

services/cargohold/src/CargoHold/API/Public.hs
lepsa: Done, I've moved the creation of the ClientEnv into the environment for the Handler.

services/cargohold/test/integration/API.hs
lepsa: I've moved this over.

integration/test/Test/AssetUpload.hs
```diff
@@ -0,0 +1,18 @@
...
+import API.BrigInternal
+import API.Cargohold
+import SetupHelpers
+import Testlib.Prelude
+
+testAssetUploadUnverifiedUser :: HasCallStack => App ()
+testAssetUploadUnverifiedUser = do
+  user <- randomUser OwnDomain $ def {activate = False}
+  bindResponse (uploadAsset user) $ \resp -> do
+    resp.status `shouldMatchInt` 403
```
akshaymankar, line 12: We should also test unknown user case since we also added code for that.

services/brig/test/integration/API/User/Account.hs
```diff
@@ -1454,7 +1454,10 @@ testDeleteAnonUser brig = do
 
 testDeleteWithProfilePic :: Brig -> CargoHold -> Http ()
 testDeleteWithProfilePic brig cargohold = do
-  uid <- userId <$> createAnonUser "anon" brig
+  -- A random local part, with nice, email-friendly characters
```
akshaymankar, line 1457:
```suggestion
```

services/cargohold/src/CargoHold/API/Public.hs
```diff
@@ -136,9 +143,28 @@ uploadAssetV3 ::
...
+      users <- either (const $ throwE userNotFound) pure resp
+      maybe
+        (throwE unverified)
+        (const (pure ()) . userIdentity . accountUser)
+        (listToMaybe users)
+    _ -> pure ()
   asset <- V3.upload principal (getAssetSource req)
   pure (fmap tUntagged asset, mkAssetLocation @tag (asset ^. assetKey))
+  where
+    client' uid =
```
akshaymankar, line 161:
```suggestion
    getUserById uid =
```

It  might even make sense to just move the whole `runClientM` and `clientM` thing in here. Perhaps even start createing an "internal  brig client" in `wire-api`, so we don't have to do this adhoc servant stuff.

akshaymankar (changes requested):

integration/test/Test/AssetUpload.hs
lepsa: Done and done

libs/wire-api/src/Wire/API/Error/Cargohold.hs
```diff
@@ -26,6 +26,8 @@ data CargoholdError
   | AssetTooLarge
   | InvalidLength
   | NoMatchingAssetEndpoint
+  | Unverified
```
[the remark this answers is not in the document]
lepsa: Not at all!

services/brig/test/integration/API/User/Account.hs
lepsa: Done

services/cargohold/src/CargoHold/App.hs
```diff
@@ -106,7 +110,10 @@ newEnv o = do
   ama <- initAws (o ^. Opt.aws) lgr mgr
   multiIngressAWS <- initMultiIngressAWS lgr mgr
   let loc = toLocalUnsafe (o ^. Opt.settings . Opt.federationDomain) ()
-  pure $ Env ama met lgr mgr h2mgr def o loc multiIngressAWS
+      (Endpoint h p) = o ^. brig
+      baseUrl = BaseUrl Http (T.unpack h) (fromIntegral p) ""
+  clientEnv <- liftIO $ newManager defaultManagerSettings <&> \m -> ClientEnv m baseUrl Nothing defaultMakeClientRequest
```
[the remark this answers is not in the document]
lepsa: Why not indeed. Done.

services/cargohold/src/CargoHold/API/Public.hs
lepsa: Done, and I've moved the `runClientM` around.

services/cargohold/src/CargoHold/API/Public.hs
```diff
@@ -136,9 +143,28 @@ uploadAssetV3 ::
...
   let principal = mkPrincipal pid
+      getUser uid = runClientM (client' uid) clientEnv
+  case principal of
+    V3.UserPrincipal uid -> do
+      resp <- liftIO $ getUser uid
+      users <- either (const $ throwE userNotFound) pure resp
+      maybe
+        (throwE unverified)
+        (const (pure ()) . userIdentity . accountUser)
+        (listToMaybe users)
```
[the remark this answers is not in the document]
lepsa: Yes, I think I did a placeholder for that and got to writing the tests. I'll fix that.

I don't think that the user status is enough, as a user could be active without being verified. The ticket specifically talks 
about email verification.

`isNewUserEphemeral` is part of the checks that run when creating a new user, and having a phone number set is enough to be classed as non-ephemeral. Non-ephemeral users are registered as Active. This is checked as part of `newAccount`, which is in turn used by `createUser` and the "register" route.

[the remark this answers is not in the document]
lepsa: I've reached out to the ticket reporter to clarify what sort of user verification is desired.

lepsa:
> @lepsa this is ready for review again, correct?

There is an update on the ticket that I need to add into this, but probably good to go by the time you wake up.

services/cargohold/test/integration/API/Util.hs
lepsa: I think that would be a good idea, but would likely warrant its own ticket.

services/cargohold/src/CargoHold/API/Public.hs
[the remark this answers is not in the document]
lepsa: `iGetUserStatus` is indeed the way to go, good call.

lepsa:
> > @lepsa this is ready for review again, correct?
> 
> There is an update on the ticket that I need to add into this, but probably good to go by the time you wake up.

This should be good to go. I'm waiting on local tests to complete, but they are taking their sweet time.

services/cargohold/src/CargoHold/App.hs
```diff
@@ -106,7 +110,10 @@ newEnv o = do
   ama <- initAws (o ^. Opt.aws) lgr mgr
   multiIngressAWS <- initMultiIngressAWS lgr mgr
   let loc = toLocalUnsafe (o ^. Opt.settings . Opt.federationDomain) ()
-  pure $ Env ama met lgr mgr h2mgr def o loc multiIngressAWS
+      (Endpoint h p) = o ^. brig
+      baseUrl = BaseUrl Http (T.unpack h) (fromIntegral p) ""
+      clientEnv = ClientEnv mgr baseUrl Nothing defaultMakeClientRequest
+  pure $ Env ama met lgr mgr h2mgr def o loc multiIngressAWS clientEnv
```
elland, line 116: Nitpick: maybe we could use more readable names 😅

services/cargohold/test/integration/API/Util.hs
elland: Can we paste the ticket here for posterity?

akshaymankar: I created the ticket: https://wearezeta.atlassian.net/browse/WPB-5382

akshaymankar:
@lepsa All the servant machinery in the middle of domain code was bothering me, so I refactored the client out. There are only two functions left there which do absolute minimum. I think we can make it 1 if we use the generic servant stuff, but I think its a debate for another time.

Apart from this I made some small changes which @elland had requested.

Let me know if this looks OK to you. If so we can merge it tomorrow.

lepsa:
@akshaymankar Those changes look good to me. Simplifying is going to be for the best.

PR #3608, opened 2023-09-25
[WPB-700] Pass along server errors between Galley and Stern

elland:
https://wearezeta.atlassian.net/browse/WPB-700

## Checklist

### DIFF
--- a/services/brig/src/Brig/API/Internal.hs
+++ b/services/brig/src/Brig/API/Internal.hs
@@ -72,7 +72,7 @@ import Data.Qualified
 import Data.Set qualified as Set
 import Data.Time.Clock.System
 import Imports hiding (head)
-import Network.Wai.Routing hiding (toList, trace)
+import Network.Wai.Routing hiding (toList)
 import Network.Wai.Utilities as Utilities
 import Polysemy
 import Servant hiding (Handler, JSON, addHeader, respond)
--- a/services/brig/src/Brig/API/Internal.hs
+++ b/services/brig/src/Brig/API/Internal.hs
@@ -72,7 +72,7 @@ import Data.Qualified
 import Data.Set qualified as Set
 import Data.Time.Clock.System
 import Imports hiding (head)
-import Network.Wai.Routing hiding (toList)
+import Network.Wai.Routing hiding (toList, trace)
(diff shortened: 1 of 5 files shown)


services/brig/src/Brig/API/Internal.hs
```diff
@@ -72,7 +72,7 @@ import Data.Qualified
 import Data.Set qualified as Set
 import Data.Time.Clock.System
 import Imports hiding (head)
-import Network.Wai.Routing hiding (toList)
+import Network.Wai.Routing hiding (toList, trace)
```
fisx, line 75:
```suggestion
import Network.Wai.Routing hiding (toList)
```

tools/stern/src/Stern/Intra.hs
```diff
@@ -454,18 +455,18 @@ getTeamBillingInfo :: TeamId -> Handler (Maybe TeamBillingInfo)
...
         ( method GET
             . Bilge.paths ["i", "team", toByteString' tid, "billing"]
         )
-  case Bilge.statusCode r of
-    200 -> Just <$> parseResponse (mkError status502 "bad-upstream") r
+  case Bilge.statusCode resp of
+    200 -> Just <$> parseResponse (mkError status502 "bad-upstream") resp
     404 -> pure Nothing
-    _ -> throwE (mkError status502 "bad-upstream" "bad response")
+    _ -> throwE (mkError status502 "bad-upstream" (errorMessage resp))
```
fisx, line 469: i thought the idea was to re-throw the upstream error message?  isn't that more helpful than wrapping that in a 5xx?

fisx (commented):
looks reasonable, but of course i have comments.  :)  also: are there more end-points in stern that need this treatment?

tools/stern/src/Stern/Intra.hs
elland: I thought we just didn't want to lose the meta information from the error response, but still re-throw upstream issues as 502s.

elland:
> looks reasonable, but of course i have comments. :) also: are there more end-points in stern that need this treatment?

Didn't find any others :)

fisx (approved):

PR #3610, opened 2023-09-25
Fix MLS message notification bug

pcapriotti:
https://wearezeta.atlassian.net/browse/WPB-3867

This PR fixes a bug where some clients do not see MLS message notifications when fetching them with a `client` query parameter. The problem is that MLS message notifications are sent to gundeck as a single push, which will in general contain different `Recipient` objects for different clients of the same users. This makes the entries conflict with one another in cassandra, and therefore only one client ends up being stored in the database.

~The fix here is "condensing" consecutive `Recipient` items relating to the same user by joining them together into a single `Recipient` object. This works because because recipient lists are generated by user.~

This is now fixed by overloading `newMessagePush` so that it can work both with simple `(User, Client)` pairs - employed by Proteus messaging code - and with actual `Recipient` objects (i.e. one user, multiple clients) - used by MLS.

## TODO

- [x] Fix bug for remote notifications

## Checklist

### DIFF
--- a/services/galley/src/Galley/API/Federation.hs
+++ b/services/galley/src/Galley/API/Federation.hs
@@ -27,6 +27,7 @@ import Data.ByteString.Conversion (toByteString')
 import Data.Domain (Domain)
 import Data.Id
 import Data.Json.Util
+import Data.List1 (List1 (..))
 import Data.Map qualified as Map
 import Data.Map.Lens (toMapOf)
 import Data.Qualified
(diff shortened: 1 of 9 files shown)


services/galley/src/Galley/API/Federation.hs
```diff
@@ -803,6 +804,8 @@ onMLSMessageSent domain rmm =
             Event (tUntagged rcnv) rmm.subConversation rmm.sender rmm.time $
               EdMLSMessage (fromBase64ByteString rmm.message)
 
+      traceM $ "got remote message: " <> show rmm
+
```
smatting, line 808: Leftofter

smatting (changes requested):

services/galley/src/Galley/API/Federation.hs
pcapriotti: Fixed, thanks

mdimjasevic (approved):

PR #3611, opened 2023-09-25
WPB-4835 call-between-web-and-android-cant-be-established-when-one-backend-is-not-available

battermann:
https://wearezeta.atlassian.net/browse/WPB-4835

Implementation according to https://wearezeta.atlassian.net/wiki/spaces/ENGINEERIN/pages/897712162/Use+Case+Fetching+the+clients+of+a+user

## Checklist

### DIFF
--- a/integration/test/Test/Client.hs
+++ b/integration/test/Test/Client.hs
@@ -4,7 +4,6 @@ module Test.Client where
 
 import API.Brig
 import API.Brig qualified as API
-import API.BrigInternal qualified as API
 import API.Gundeck
 import Control.Lens hiding ((.=))
 import Control.Monad.Codensity
@@ -63,21 +62,10 @@ testListClientsIfBackendIsOffline = do
     resp.status `shouldMatchInt` 200
     resp.json %. "qualified_user_map" `shouldMatch` expectedResponse
 
+  -- we don't even have to start the backend, but we have to take the resource so that it doesn't get started by another test
   runCodensity (acquireResources 1 resourcePool) $ \[downBackend] -> do
-    downUser <- runCodensity (startDynamicBackend downBackend mempty) $ \_ -> do
-      do
-        let domains = [ownDomain, otherDomain, downBackend.berDomain]
-        sequence_
-          [ API.createFedConn x (API.FedConn y "full_search")
-            | x <- domains,
-              y <- domains,
-              x /= y
-          ]
-
-      downUser <- randomUser downBackend.berDomain def
-      connectUsers ownUser1 downUser
-      connectUsers ownUser2 downUser
-      pure (downUser)
+    rndUsrId <- randomId
+    let downUser = (object ["domain" .= downBackend.berDomain, "id" .= rndUsrId])
 
     bindResponse (listUsersClients ownUser1 [ownUser1, ownUser2, downUser]) $ \resp -> do
       resp.status `shouldMatchInt` 200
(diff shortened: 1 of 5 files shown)


integration/test/Test/Client.hs
```diff
@@ -32,3 +41,44 @@ testClientLastActive = do
...
+          [ API.createFedConn x (API.FedConn y "full_search")
+            | x <- domains,
+              y <- domains,
+              x /= y
+          ]
+
+      downUser <- randomUser downBackend.berDomain def
+      connectUsers ownUser1 downUser
+      connectUsers ownUser2 downUser
+      pure (downUser)
```
akshaymankar, line 80: This user doesn't even need to exist if the backend is down anyway. Might save time in actually having to start the backend.

integration/test/Test/Client.hs
```diff
@@ -32,3 +41,44 @@ testClientLastActive = do
...
+
+    bindResponse (listUsersClients ownUser1 [ownUser1, ownUser2, downUser]) $ \resp -> do
+      resp.status `shouldMatchInt` 200
+      resp.json %. "qualified_user_map" `shouldMatch` expectedResponse
```
akshaymankar, line 84: Is this OK for old clients? As in what do they do if they don't see some users? Do they have code to handle this? If not we should consider adding an extra field here which indicates partial failure and doing this in the next API version.

services/brig/src/Brig/API/Client.hs
```diff
@@ -133,13 +132,18 @@ lookupPubClientsBulk :: [Qualified UserId] -> ExceptT ClientError (AppT r) (Qual
...
   pure $ QualifiedUserMap (Map.union localUserClientMap remoteUserClientMap)
+  where
+    getRemoteClients :: Map Domain [UserId] -> AppT r (Map Domain (UserMap (Set PubClient)))
+    getRemoteClients uids = do
+      results <-
+        traverse
+          ( \(d, ids) -> fmap (d,) <$> runExceptT (getUserClients d (GetUserClients ids))
+          )
+          (Map.toList uids)
+      pure $ Map.fromList (rights results)
```
akshaymankar, line 146: It maybe worth logging the errors?

integration/test/Test/Client.hs
marcoconti83: Before this PR, the request will return 503. After the PR, the request will return a subset of the users.
No matter how the clients are able to handle a subset of the users (and they seem to be doing it properly already), the 503 was causing a DOS anyway that would prevent calling from working completely, so we are not making things worse :)

akshaymankar: Alright, as far as clients don't think the users are deleted :smile:

akshaymankar (approved):

integration/test/Test/Client.hs
akshaymankar: Maybe we should add this information to the description of the endpoint tho.

libs/wire-api/src/Wire/API/Routes/Public/Brig.hs
```diff
@@ -920,6 +920,7 @@ type ClientAPI =
     :<|> Named
            "list-clients-bulk@v2"
            ( Summary "List all clients for a set of user ids"
+               :> Description "If a backend is unreachable, the clients from this backend will be omitted from the response"
```
akshaymankar, line 923:
```suggestion
               :> Description "If a backend is unreachable, the clients from that backend will be omitted from the response"
```

akshaymankar (approved):

PR #3612, opened 2023-09-26
WPB-3797 do not cache federation remote domain config

lepsa:
# Changes
Removing the domain caching and sync threads from background worker and federator. Galley already had this code removed.

## Notes
Background worker is deriving the list of domains from the Rabbit queues, and I find this strange. This code is coming in from `develop`, so I've left it alone as changing it over to pull directly from Brig causes problems in the integration tests.

## Checklist


services/federator/src/Federator/Response.hs
```diff
@@ -145,7 +149,7 @@ type AllEffects =
      ServiceStreaming,
      Input RunSettings,
      Input Http2Manager, -- needed by Remote
-     Input FederationDomainConfigs, -- needed for the domain list.
+     Input FedUp.FederationDomainConfigs, -- needed for the domain list and federation policy.
```
elland, line 152: 😂

elland (approved):
LGTM, let's see what CI thinks.

services/federator/src/Federator/Response.hs
lepsa: Probably part of a pile of changes I made and then backed out.

PR #3624, opened 2023-09-29
WPB-5017: Adding a test config file to the PR guidelines.

lepsa:
## Changes

Adding a testing config file to the PR guidelines.

## Checklist


docs/src/developer/developer/pr-guidelines.md
```diff
@@ -81,7 +81,7 @@ Add to `charts/nginz/values.yaml`, under the `galley` section:
 If a PR adds new configuration options for say brig, the following files need to be edited:
 
 * [ ] The parser under `services/brig/src/Brig/Options.hs`
-* [ ] The integration test config: `services/brig/brig.integration.yaml`
+* [ ] The integration test config: `services/brig/brig.integration.yaml` and `charts/cargohold/templates/tests/configmap.yaml`
```
fisx, line 84: ah, wait, this was already documented, but you have to read the context carefully.  this list is an example for the context where you made changes to brig.  for cargohold, most items in this list will change.

do you have a suggestion how to make this more clear?  maybe introduce variables $Service and $service to use in the list instead of `brig`?

fisx (changes requested):

docs/src/developer/developer/pr-guidelines.md
lepsa: Variables would work, I think. I'll sit with this a bit and see what I can come up with.

pcapriotti (approved):
This is fine, but you could also add the item to the previous list, and simply use `brig` instead of `$SERVICE`, continuing the example.

fisx (approved):

PR #3630, opened 2023-10-04
Align connectUsers to mls branch

pcapriotti:
There have been some diverging changes on the mls branch related to the way users are created and connected in tests. This aligns the version on develop with that on mls.

## Checklist

 - [x] No CHANGELOG entry

### DIFF
--- a/integration/test/SetupHelpers.hs
+++ b/integration/test/SetupHelpers.hs
@@ -64,7 +64,7 @@ createTeamMember inviter tid = do
       <&> addJSONObject registerJSON
   getJSON 201 =<< submit "POST" registerReq
 
-connectUsers ::
+connectUsers2 ::
   ( HasCallStack,
     MakesValue alice,
     MakesValue bob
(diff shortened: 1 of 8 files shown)


mdimjasevic (approved):

integration/test/SetupHelpers.hs
```diff
@@ -64,23 +64,30 @@ createTeamMember inviter tid = do
...
+connectUsers2 ::
   ( HasCallStack,
     MakesValue alice,
     MakesValue bob
   ) =>
   alice ->
   bob ->
   App ()
-connectUsers alice bob = do
+connectUsers2 alice bob = do
```
akshaymankar, line 75: Can we please name it `connectTwoUsers` or `connect2Users`.

integration/test/SetupHelpers.hs
```diff
@@ -64,23 +64,30 @@ createTeamMember inviter tid = do
...
-createAndConnectUsers d1 d2 = do
-  [u1, u2] <- for [d1, d2] (flip randomUser def)
-  connectUsers u1 u2
-  pure (u1, u2)
+connectUsers :: HasCallStack => [Value] -> App ()
+connectUsers users = traverse_ (uncurry connectUsers2) $ do
+  t <- tails users
+  (a, others) <- maybeToList (uncons t)
+  b <- others
+  pure (a, b)
```
akshaymankar, line 84: Actually, with @fisx we decided that we should never have to do this. I was noticing that we use this unnecessarily everywhere. This operation is fairly expesive, so we should try to avoid it. We were able to remove it from all places. Do you have an example of where this is useful?

pcapriotti: It's just so that we don't have to think about connections at all. It's probably pretty rare that we want some connections to be absent (and in that case we can establish connections by hand). In all other cases, the possibly unnecessary connections are probably not adding much to the overall test runtime. Have you found tests that became noticeably faster after making the connections more efficient?

akshaymankar: The other day I saw people doing this for like 6 users, when all we wanted was 1 user to be connected to other 5 users. This just makes sure we call backend 10 times instead of 42 times. I imagine not doing this in a few tests would make things faster. I didn't test this however.
I think in most cases we need 1 user to be connected to everyone who is to be added to a conversation. Other connections are just not needed. So, a compelete graph of users is a special case rather than the other way.

pcapriotti: Connecting 6 users takes 12 vs. 30 requests, not 10 vs. 42. The most I've seen in the integration suite is 5 users, where it's 10 vs. 20. I've tries that test (`testAddReachableWithUnreachableRemoteUsers`) and the difference in runtime seems to be on the edge of being statistically undetectable (probably around 100ms, given a few dozen runs). I'd say it's not worth optimising this for such small numbers of users.

mdimjasevic: I agree with Paolo here. This is test code so we don't have to be that precise. The benefit of Paolo's approach is that it is very simple to use, making it easy to write tests and it avoids dull test setup work, and as he tested the performance, it's borderline slower.

To me this doesn't seem to be a place to improve performance of our tests.

PR #3634, opened 2023-10-06
[feat] improve type safety for Named, servantify brig internal route

MangoIV:
https://wearezeta.atlassian.net/browse/WPB-1224
- improve type safety of Named by making it possible to rule out weakly typed arguments to the type (e.g. Type)
- servantify the internal route for querying the teams API for servant

## Checklist

### DIFF
--- a/libs/wire-api/src/Wire/API/Routes/Internal/Brig.hs
+++ b/libs/wire-api/src/Wire/API/Routes/Internal/Brig.hs
@@ -498,12 +498,9 @@ instance ToSchema NewKeyPackageRef where
   schema =
     object "NewKeyPackageRef" $
       NewKeyPackageRef
-        <$> nkprUserId
-          .= field "user_id" schema
-        <*> nkprClientId
-          .= field "client_id" schema
-        <*> nkprConversation
-          .= field "conversation" schema
+        <$> nkprUserId .= field "user_id" schema
+        <*> nkprClientId .= field "client_id" schema
+        <*> nkprConversation .= field "conversation" schema
 
 data NewKeyPackage = NewKeyPackage
   { nkpConversation :: Qualified ConvId,
@@ -516,10 +513,8 @@ instance ToSchema NewKeyPackage where
   schema =
     object "NewKeyPackage" $
       NewKeyPackage
-        <$> nkpConversation
-          .= field "conversation" schema
-        <*> nkpKeyPackage
-          .= field "key_package" schema
+        <$> nkpConversation .= field "conversation" schema
+        <*> nkpKeyPackage .= field "key_package" schema
 
 data NewKeyPackageResult = NewKeyPackageResult
   { nkpresClientIdentity :: ClientIdentity,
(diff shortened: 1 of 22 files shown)


libs/wire-api/src/Wire/API/Routes/Internal/Brig.hs
```diff
@@ -664,6 +669,81 @@ type TeamsAPI =
...
+    "suspend-team"
+    ( "teams"
+        :> Capture "tid" TeamId
+        :> "suspend"
+        :> Post
+             '[Servant.JSON]
+             NoContent
+    )
+
+type UnSuspendTeam =
```
elland, line 718: `UnsuspendTeam` makes more sense to me. The current spelling makes it look like a typical Haskell unwrap of newtypes.

libs/wire-api/src/Wire/API/Routes/Named.hs
```diff
@@ -33,7 +33,9 @@ import Servant.Client.Core (clientIn)
 import Servant.OpenApi
 
 -- | See http://docs.wire.com/developer/developer/servant.html#named-and-internal-route-ids-in-swagger
-newtype Named name x = Named {unnamed :: x}
+type Named name = UntypedNamed (IsStronglyTyped name)
+
+newtype UntypedNamed name x = Named {unnamed :: x}
```
elland, line 38: I think some comments explaining what and why here might be good for posterity :)

elland (commented):
How do you feel about moving the api handler definitions to the same file where we define the routes?

libs/wire-api/src/Wire/API/Routes/Internal/Brig.hs
MangoIV: haha, you're right, I will change it :D

MangoIV:
> How do you feel about moving the api handler definitions to the same file where we define the routes?

@elland the file I would move to is already quite large and the `Brig.Team.API` file contains the public routes as well, so that was my reason to not move it, if you wish though, I will move it over. 

Maybe also consider moving the API definition itself to `Brig.Team.API`, similarly how it is done with the [exported `servantAPI`](https://github.com/wireapp/wire-server/blob/c15986afbdbf38db81ed67fad465cc98d882ffd3/services/brig/src/Brig/Team/API.hs#L81)

libs/wire-api/src/Wire/API/Routes/Internal/Brig.hs
```diff
@@ -664,6 +669,81 @@ type TeamsAPI =
...
+        :> Get '[Servant.JSON] FoundInvitationCode
+    )
+
+newtype FoundInvitationCode = FoundInvitationCode User.InvitationCode
+  deriving stock (Eq, Show, Generic)
+  -- TODO: is this correct?
+  deriving newtype (S.ToSchema)
+
+instance ToJSON FoundInvitationCode where
+  toJSON (FoundInvitationCode c) = Aeson.object ["code" Aeson..= c]
```
MangoIV, line 705: I also would like to ask if this is correct, especially the `S.ToSchema` part which does not adhere to the `ToJSON` instance but I don't know how / if they should be related.

elland (approved):

MangoIV:
*finally the tests have passed!*

PR #3638, opened 2023-10-10
[WPB-3138] Split brig-schema into lib and exec.

elland:
https://wearezeta.atlassian.net/browse/WPB-3138

## Checklist


mdimjasevic:
A changelog is missing.

services/brig/src/Brig/Schema/V_FUTUREWORK.hs
```diff
@@ -17,7 +17,7 @@
 -- You should have received a copy of the GNU Affero General Public License along
 -- with this program. If not, see <https://www.gnu.org/licenses/>.
 
-module V_FUTUREWORK
+module Brig.Schema.V_FUTUREWORK
```
mdimjasevic, line 20: This module sounds like it needs some love. Can you please create a ticket to address it?

services/brig/src/Brig/App.hs
```diff
@@ -157,7 +158,7 @@ import Wire.API.User.Identity (Email)
 import Wire.API.User.Profile (Locale)
 
 schemaVersion :: Int32
-schemaVersion = 79
```
mdimjasevic, line 160: Was this one wrong? Because there's a migration with version 80.

mdimjasevic (approved):

services/brig/src/Brig/App.hs
mdimjasevic: I just hope there won't be a problem in any installation due to this, as in a wrong thing was in place and it worked, but this made it right, but actually broke it.

PR #3640, opened 2023-10-10
[WPB-4547] Attach the reason for a member to leave a conversation to the leave event

MangoIV:
https://wearezeta.atlassian.net/browse/WPB-4547
## Checklist


.envrc
```diff
@@ -50,4 +50,6 @@ export INTEGRATION_DYNAMIC_BACKENDS_POOLSIZE=3
 # Keep these in sync with deploy/dockerephmeral/init.sh
 export AWS_REGION="eu-west-1"
 export AWS_ACCESS_KEY_ID="dummykey"
-export AWS_SECRET_ACCESS_KEY="dummysecret"
\ No newline at end of file
+export AWS_SECRET_ACCESS_KEY="dummysecret"
+
+ulimit -n 10420
```
pcapriotti, line 55: Why? Can you maybe add it to your .local file?

services/galley/src/Galley/API/Teams.hs
```diff
@@ -15,6 +15,7 @@
 -- You should have received a copy of the GNU Affero General Public License along
 -- with this program. If not, see <https://www.gnu.org/licenses/>.
 {-# LANGUAGE LambdaCase #-}
+{-# LANGUAGE PartialTypeSignatures #-}
```
pcapriotti, line 18: Leftover?

services/galley/src/Galley/API/Update.hs
```diff
@@ -1170,6 +1170,11 @@ removeMemberQualified lusr con qcnv victim =
       qcnv
       victim
 
+-- | if the public member leave api was called, we can assume that
+--   it was called by a user
+pattern EdMembersLeaveRemoved :: QualifiedUserIdList -> EventData
+pattern EdMembersLeaveRemoved l = EdMembersLeave EdReasonRemoved l
```
pcapriotti, line 1176: Can't this just be a function?

pcapriotti (approved):
Looks good to me. However, it seems the "user deleted" and "user removed from team" reasons are conflated into one. That's probably fine, but maybe they should both be called "user deleted" rather than removed from team.

Minor comments follow.

services/galley/src/Galley/API/Update.hs
MangoIV: sure, it can be, but I thought this would be useful to mark it as a pattern so you can 
- use it both ways
- you show it's not doing anything extra (in contrast to e.g. a smart constructor) 

will change it if you wish.

services/galley/src/Galley/API/Teams.hs
MangoIV: Oh yes, I will remove it

MangoIV:
> Looks good to me. However, it seems the "user deleted" and "user removed from team" reasons are conflated into one. That's probably fine, but maybe they should both be called "user deleted" rather than removed from team.
> 
> Minor comments follow.

So I will change it to "user deleted"? Or should I add another reason? I tried to tightly cohere with the ticket.

services/galley/src/Galley/API/Update.hs
pcapriotti: Nah, it's fine. Up to you.

MangoIV:
For transparency: it was internally decide to rename to field to something more appropriate.

PR #3647, opened 2023-10-12
[WPB-4928] Stop using Servant client to enqueue federation notifications

mdimjasevic:
This stops using the Servant client to enqueue federation notifications. It also reorganises the federation API notification endpoints such that they are separated more from the synchronous endpoints. The notification endpoints rely on a type alias that always has `EmptyResponse` as the response type.

Tracked by https://wearezeta.atlassian.net/browse/WPB-4928.

## Checklist

### DIFF
--- a/libs/wire-api-federation/src/Wire/API/Federation/API.hs
+++ b/libs/wire-api-federation/src/Wire/API/Federation/API.hs
@@ -101,6 +101,7 @@ fedClient = clientIn (Proxy @api) (Proxy @m)
 fedQueueClient ::
   forall tag api.
   ( HasNotificationEndpoint tag,
+    -- FUTUREWORK: Include this API constraint and get it working
     -- api ~ NotificationAPI tag (NotificationComponent tag),
     HasEmptyResponse api,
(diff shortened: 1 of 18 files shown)


libs/wire-api-federation/src/Wire/API/Federation/API.hs
```diff
@@ -94,14 +99,31 @@ fedClient ::
...
 
 fedQueueClient ::
-  forall (comp :: Component) (name :: Symbol) m api.
-  ( HasEmptyResponse api,
-    HasFedEndpoint comp api name,
-    HasClient m api,
-    m ~ FedQueueClient comp
+  forall tag api.
+  ( HasNotificationEndpoint tag,
+    -- api ~ NotificationAPI tag (NotificationComponent tag),
```
mdimjasevic, line 104: When I include this constraint, I get an error about the non-injectivity of a type family and I don't know how to fix it.

battermann (approved):
LGTM overall, but I don't know about the non-injectivity error either.

PR #3650, opened 2023-10-13
[WPB-5042] upgrade nixpkgs to upgrade haskell-language-server

MangoIV:
https://wearezeta.atlassian.net/browse/WPB-5042

- update to a new nixpkgs version
- update hls 
- change overrides where needed 
- ~~added new possibility to use nom~~ addressed in #3687 
- replaced old tests
- pinned ormolu 

## Checklist


pcapriotti (approved):
Looks good. I didn't look at the nix changes much, though. Minor comments below.

integration/test/Testlib/Types.hs
```diff
@@ -201,7 +202,18 @@ data ClientIdentity = ClientIdentity
...
+    (ToJSON)
+    via -- we want to change these so they're compatible with the members field
+    -- coming from a getConversation
+    CustomJSON
+      [ FieldLabelModifier (Rename "user" "user_id"),
+        FieldLabelModifier (Rename "client" "client_id")
+      ]
+      ClientIdentity
+
+data Mod
```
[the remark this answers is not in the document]
MangoIV: oups :eyes: I wonder whether it did not fail the CI...

integration/test/Testlib/Types.hs
```diff
@@ -201,7 +202,18 @@ data ClientIdentity = ClientIdentity
...
+  deriving stock (Show, Eq, Ord, Generic)
+  deriving
+    (ToJSON)
+    via -- we want to change these so they're compatible with the members field
+    -- coming from a getConversation
+    CustomJSON
+      [ FieldLabelModifier (Rename "user" "user_id"),
+        FieldLabelModifier (Rename "client" "client_id")
+      ]
+      ClientIdentity
```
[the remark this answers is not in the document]
MangoIV: I had to come up with a different fix but I removed the instance.

MangoIV:
*finally* it is green :)))

nix/manual-overrides.nix
```diff
@@ -2,34 +2,50 @@
...
   bytestring-arbitrary = hlib.markUnbroken (hlib.doJailbreak hsuper.bytestring-arbitrary);
+  bytestring-conversion = hlib.markUnbroken (hsuper.bytestring-conversion);
   openapi3 = hlib.markUnbroken (hlib.dontCheck hsuper.openapi3);
   cql = hlib.appendPatch (hlib.markUnbroken hsuper.cql) (fetchpatch {
     url = "https://gitlab.com/twittner/cql/-/merge_requests/11.patch";
     sha256 = "sha256-qfcCRkKjSS1TEqPRVBU9Ox2DjsdGsYG/F3DrZ5JGoEI=";
   });
   hashtables = hsuper.hashtables_1_3;
+  # in case everything breaks with the hashable update, use this 
+  # hashable = hsuper.callHackage "hashable" "1.4.2.0" {};
```
akshaymankar, line 15: Can we make sure things don't break before we merge them?

nix/manual-overrides.nix
```diff
@@ -2,34 +2,50 @@
...
+  template = hlib.markUnbroken hsuper.template;
   type-errors = hlib.dontCheck hsuper.type-errors;
+  th-abstraction = hsuper.th-abstraction_0_5_0_0;
+  th-desugar = hlib.doJailbreak hsuper.th-desugar;
   wai-middleware-prometheus = hlib.doJailbreak hsuper.wai-middleware-prometheus;
   wai-predicates = hlib.markUnbroken hsuper.wai-predicates;
 
+  # FIXME: the test "should allow accepting a certificate without a trailing dot" fails
+  # but only in the nix setup
+  http2-manager = hlib.dontCheck hsuper.http2-manager;
```
akshaymankar, line 47: Maybe we should ignore this one test in nix instead of ignoring all tests. There is no other place where tests from this repo get run.

.envrc
```diff
@@ -16,7 +16,11 @@ env_dir=./.env
 [[ -d "$layout_dir" ]] || mkdir -p "$layout_dir"
 
 if [[ ! -d "$env_dir" || ! -f "$layout_dir/nix-rebuild" || "$store_paths" != $(< "$layout_dir/nix-rebuild" ) ]]; then
-  nix-build ./nix -A wireServer.devEnv --out-link ./.env
+  if [[ "${USE_NOM}" == "1" ]]; then
```
flokli, line 19: nix-output-monitor is coming from outside the development environment, if the user did add it to their system config / user profile.

I'd probably invert the logic a bit - use `nom-build` if it's in `$PATH`, but let people override behaviour by setting `USE_NOM` to `0` if they want to opt-out.

flokli: (and this can probably be moved into its own PR, so it can land independently of the GHC upgrade)

MangoIV: I've added it to the devShell, it seems very non-nix-y to depend on the state of the environment to do something, idk; additionally idk if people want nom all the time, most times the entire environment is built and people want to just evaluate the nix and then be dropped in the devShell. Nom just displays a tiny stopwatch while evaluating so perhaps it will look like "it's not doing something". 
But if you think that it would make more sense like that, I'll change it. :)

flokli: But you're not guaranteed to be in the devshell already (assuming a clean checkout).

Hence the proposal - pick nom from `$PATH`, if people have it in their environment and didn't explicitly opt out with `USE_NOM=0`, but not add it to the dev environment (as it's likely already there)

MangoIV: Makes sense. Yes I will implement it like that then 😌

integration/test/MLS/Util.hs
```diff
@@ -573,30 +571,55 @@ consumeMessage cid mmp ws = do
...
+      if isNotif
+        then do
+          msg <- n %. "payload.0.data" & asByteString >>= showMessage cid
+          sender <- msg `lookupField` "message.content.sender" `catch` \(_ :: AssertionFailure) -> pure Nothing
+          let backendSender = object ["External" .= Number 0]
+          pure $ sender /= Just backendSender
+        else pure False
+
+consumeMessage1 :: ClientIdentity -> ByteString -> App ByteString
+consumeMessage1 cid msgData =
```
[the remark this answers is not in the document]
MangoIV: it is what it was before so it is consistent with the past, I just restored it, I agree the name is weird, thought. I will change it.

nix/overlay.nix
```diff
@@ -69,10 +69,20 @@ self: super: {
...
 
-  nginz = super.nginx.override {
+  nginz = (super.nginx.overrideAttrs rec {
+    version = "1.22.1";
+    src = super.fetchurl {
+      url = "https://nginx.org/download/nginx-${version}.tar.gz";
+      hash = "sha256-nrszOp6CuVKs0+K0rrHU/2QG9ySRurbNn+afDepzfzE=";
+    };
+  }).override {
+    src = null;
```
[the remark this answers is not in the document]
MangoIV: if you mean the `null`, I think this was because otherwise it would already fail at the override stage; I can try and remove it.

akshaymankar (approved):
The link to the issue of nginx pinning to be added in subsequent PR.

nix/overlay.nix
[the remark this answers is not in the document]
MangoIV: https://wearezeta.atlassian.net/jira/software/c/projects/WPB/issues/WPB-5360

MangoIV:
https://wearezeta.atlassian.net/jira/software/c/projects/WPB/issues/WPB-5360

PR #3652, opened 2023-10-13
Add tool to aggregate and push test statistics from junit/ant XML reports

stefanwire:
## Checklist


tools/test-stats/Main.hs
```diff
@@ -0,0 +1,171 @@
...
+  let connInfo =
+        ConnectInfo
+          { connectHost = opts.postgresqlHost,
+            connectPort = opts.postgresqlPort,
+            connectUser = opts.postgresqlUser,
+            connectPassword = opts.postgresqlPassword,
+            connectDatabase = opts.postgresqlDatabase
+          }
+  currentTime <- getCurrentTime
+  bracket (connect connInfo) (close) $ \conn -> do
```
stefanwire, line 149: Is the linter not complaining about the extra pair of parentheses around `close`?

stefanwire (commented):
@akshaymankar Legit from my side with one comment. Also, I cannot approve my own PR.

tools/test-stats/Main.hs
akshaymankar: I think we removed that lint warning: https://github.com/wireapp/wire-server/blob/58bcc0c03c99519442ce7979dff12a4bf2cd6d9a/.hlint.yaml#L29-L30
I will fix it though.

akshaymankar (approved):

PR #3654, opened 2023-10-16
[WPB-4981] replace unclaimed keypackages atomically

MangoIV:
https://wearezeta.atlassian.net/browse/WPB-4981
## Checklist

### DIFF
--- a/integration/test/Test/MLS/KeyPackage.hs
+++ b/integration/test/Test/MLS/KeyPackage.hs
@@ -124,7 +124,7 @@ testReplaceKeyPackages = do
     checkCount def 4
     checkCount suite 0
 
-  when False $ do
+  do
     -- replenish key packages for the second ciphersuite
     void $
       replicateM 5 (fmap fst (generateKeyPackage alice1 suite))
--- a/integration/test/Test/MLS/KeyPackage.hs
+++ b/integration/test/Test/MLS/KeyPackage.hs
@@ -56,10 +56,8 @@ testKeyPackageCount cs = do
     resp.status `shouldMatchInt` 200
     resp.json %. "count" `shouldMatchInt` 0
 
-  setMLSCiphersuite cs
-
   let count = 10
-  kps <- map fst <$> replicateM count (generateKeyPackage alice1)
+  kps <- map fst <$> replicateM count (generateKeyPackage alice1 cs)
   void $ uploadKeyPackages alice1 kps >>= getBody 201
 
   bindResponse (countKeyPackages cs alice1) $ \resp -> do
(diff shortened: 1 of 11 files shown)


pcapriotti (approved):

services/brig/src/Brig/Data/MLS/KeyPackage.hs
```diff
@@ -142,6 +144,22 @@ deleteKeyPackages u c suite refs =
...
     deleteQuery = "DELETE FROM mls_key_packages WHERE user = ? AND client = ? AND cipher_suite = ? AND ref in ?"
 
+deleteAllKeyPackages ::
+  (MonadClient m, MonadUnliftIO m) =>
+  UserId ->
+  ClientId ->
+  [CipherSuiteTag] ->
+  m ()
+deleteAllKeyPackages u c suites =
+  pooledForConcurrentlyN_ 16 suites $ \suite ->
```
MangoIV, line 154: Why do we need this?

integration/test/Test/MLS/KeyPackage.hs
```diff
@@ -68,10 +66,79 @@ testKeyPackageCount cs = do
...
+          ks <- resp.json %. "key_packages" & asList
+          k <- assertOne ks
+          k %. "key_package_ref"
+
+    refs `shouldMatchSet` claimed
+
+    checkCount def 4
+    checkCount suite 0
+
+  when False $ do
```
MangoIV, line 127: This never runs, no?

MangoIV (commented):
Sorry i couldn't be of more help with this. The bug with the MLSstate took me way too long :/

services/brig/src/Brig/Data/MLS/KeyPackage.hs
pcapriotti: We need to make multiple queries, because the ciphersuite is part of the partition key. This is what we usually do in such cases.

integration/test/Test/MLS/KeyPackage.hs
pcapriotti: Oops :sweat_smile:. Thanks for catching that.

services/brig/src/Brig/Data/MLS/KeyPackage.hs
MangoIV: Alright, thank you, have you somehow hit performance boundaries when not doing it like this or is it just for good measure?

pcapriotti: I'm not sure, I'm just doing it here to keep it consistent with all the other cases.

PR #3664, opened 2023-10-19
[fix] Flaky test for leaving self-conv on MLS

elland:
Due to message queue the assertion on no events became flaky.

### DIFF
--- a/services/galley/test/integration/API/MLS.hs
+++ b/services/galley/test/integration/API/MLS.hs
@@ -1733,17 +1733,15 @@ testSelfConversationLeave :: TestM ()
 testSelfConversationLeave = do
   alice <- randomQualifiedUser
   runMLSTest $ do
-    clients@(creator : others) <- traverse createMLSClient (replicate 3 alice)
+    (creator : others) <- traverse createMLSClient (replicate 3 alice)
     traverse_ uploadNewKeyPackage others
     (_, qcnv) <- setupMLSSelfGroup creator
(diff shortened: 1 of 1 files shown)


pcapriotti (approved):
Looks good. This should be fixed once we fix the general MLS test flakiness, so maybe add a FUTUREWORK about restoring that assertion once that's done?

services/galley/test/integration/API/MLS.hs
```diff
@@ -1733,17 +1733,15 @@ testSelfConversationLeave :: TestM ()
...
-        deleteMemberQualified (qUnqualified alice) alice qcnv
-          !!! do
-            const 403 === statusCode
-            const (Just "invalid-op") === fmap Wai.label . responseJsonError
-      WS.assertNoEvent (1 # WS.Second) wss
+    liftTest $
+      deleteMemberQualified (qUnqualified alice) alice qcnv
+        !!! do
+          const 403 === statusCode
+          const (Just "invalid-op") === fmap Wai.label . responseJsonError
```
akshaymankar, line 1744: I guess lack of notification is not a good test anyway, We should probably assert that the conversation exists and is usable by the user.

akshaymankar: Ofc feel free to do it later, no need to block the release over it :smile:

elland: I'll wait for @pcapriotti 's improvements

PR #3666, opened 2023-10-20
WPB-5098 Backend-to-backend OpenApi Docs

battermann:
https://wearezeta.atlassian.net/browse/WPB-5098

## Checklist


charts/nginz/values.yaml
```diff
@@ -162,6 +162,10 @@ nginx_conf:
       disable_zauth: true
       envs:
       - staging
+    - path: /api-federation/swagger-ui
```
fisx, line 165: can you update https://docs.wire.com/understand/api-client-perspective/swagger.html to mention this?

libs/wire-api-federation/src/Wire/API/Federation/Domain.hs
```diff
@@ -58,3 +59,6 @@ instance
 
 originDomainHeaderName :: IsString a => a
 originDomainHeaderName = fromString $ symbolVal (Proxy @OriginDomainHeaderName)
+
+instance (HasOpenApi api) => HasOpenApi (OriginDomainHeader :> api) where
```
fisx, line 63: i'm not sure, but shouldn't `OriginDomainHeader` also be documented somehow?  If in doubt, add something to the description?

libs/wire-api/src/Wire/API/Conversation/Action.hs
```diff
@@ -89,6 +89,40 @@ instance ToJSON SomeConversationAction where
         actionJSON = fromMaybe A.Null $ schemaOut (conversationActionSchema sb) action
      in A.object ["tag" A..= tag, "action" A..= actionJSON]
 
+instance S.ToSchema SomeConversationAction where
```
fisx, line 92: i guess it's ok to not use schema-profunctor, but then can you add a roundtrip- and/or unit-test?

fisx (approved):

libs/wire-api/src/Wire/API/Conversation/Action.hs
battermann: Roundtrip tests for the `ToJSON` and `FromJSON` instances, which I haven't touched, already exist. The `ToSchema` instance is just for the API docs.

PR #3670, opened 2023-10-23
Integration tests flake when assuming federation ingress is up.

elland:
https://wearezeta.atlassian.net/browse/WPB-4810

## Checklist

### DIFF
--- a/integration/test/Testlib/ModService.hs
+++ b/integration/test/Testlib/ModService.hs
@@ -311,15 +311,49 @@ startBackend resource overrides services = do
           whenM (doesDirectoryExist path) $ removeDirectoryRecursive path
 
   let modifyEnv env = env {serviceMap = Map.insert resource.berDomain serviceMap env.serviceMap}
+      mkReq srv = do
+        req <- baseRequest domain srv Unversioned "/i/status"
+        checkStatus <- appToIO $ do
+          res <- submit "GET" req
+          pure (res.status `elem` [200, 204])
+        eith <- liftIO (E.try checkStatus)
+        pure $ either (\(_e :: HTTP.HttpException) -> False) id eith
 
   Codensity $ \action -> local modifyEnv $ do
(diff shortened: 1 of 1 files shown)


integration/test/Testlib/ModService.hs
```diff
@@ -365,22 +399,15 @@ logToConsole colorize prefix hdl = do
           `E.catch` (\(_ :: E.IOException) -> pure ())
   go
 
-waitUntilServiceUp :: HasCallStack => String -> Service -> App ()
-waitUntilServiceUp domain = \case
+retryRequestUntil :: HasCallStack => (Service -> App Bool) -> Service -> App ()
+retryRequestUntil mkReq = \case
```
akshaymankar, line 403: We could just accept an `App Bool` instead of `(Service -> App Bool)` and then also accepting a `Service` just to pass it to that function. This way we can achieve two things:
1. Rename the first `mkReq` function as `checkServiceUp`, so it is more clear what request it makes
2. We can avoid having to pass the awkward `FederatorInternal` as a service name when we're trying to see if the ingress is working fine.

integration/test/Testlib/ModService.hs
```diff
@@ -311,7 +311,7 @@ startBackend resource overrides services = do
           whenM (doesDirectoryExist path) $ removeDirectoryRecursive path
 
   let modifyEnv env = env {serviceMap = Map.insert resource.berDomain serviceMap env.serviceMap}
-      mkReq srv = do
+      checkServiceIsUp srv = do
```
akshaymankar, line 314: This function should return `pure True` for `Ngniz`

elland: 😅

pcapriotti (approved):
The `federation-denied` check isn't great, as we discussed. Maybe it's worth making a ticket to remember to improve error generation.

PR #3672, opened 2023-10-23
Fix extra remove proposal bug

pcapriotti:
We were sending external remove proposals for each client of a user that was kicked out of a conversation following a remove commit. This was caused by some overgeneralisation of the mechanism that removes clients from subconversations when a user is deleted from the main.

I haven't added a test for this behaviour, but the upcoming PR #3671 is able to catch this problem (which is how I discovered it in the first place), so it should be covered when that one is merged.

## Checklist

### DIFF
--- a/services/galley/src/Galley/API/MLS/Removal.hs
+++ b/services/galley/src/Galley/API/MLS/Removal.hs
@@ -19,6 +19,7 @@ module Galley.API.MLS.Removal
   ( createAndSendRemoveProposals,
     removeExtraneousClients,
     removeClient,
+    RemoveUserIncludeMain (..),
     removeUser,
   )
 where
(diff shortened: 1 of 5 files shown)


mdimjasevic (approved):

services/galley/src/Galley/API/MLS/Removal.hs
```diff
@@ -170,6 +196,10 @@ removeClient lc qusr c = do
     let getClients = fmap (cid,) . cmLookupIndex cid . (.members)
     removeClientsWithClientMapRecursively (qualifyAs lc mlsConv) getClients qusr
 
+data RemoveUserIncludeMain
```
elland, line 199: Could we have some comments explaining what this means?

pcapriotti: Added some docs. Hopefully that clarifies it.

PR #3673, opened 2023-10-24
[WPB-5103] Add users to MLS conversation when some backends are unreachable

mdimjasevic:
The PR implements no new functionality. It just adds tests that confirm that the expected behavior of adding members to a conversation is the same as in Proteus in presence of unreachable backends.

Tracked by https://wearezeta.atlassian.net/browse/WPB-5103.

## Checklist

### DIFF
--- a/integration/test/Test/MLS/Unreachable.hs
+++ b/integration/test/Test/MLS/Unreachable.hs
@@ -1,0 +1,75 @@
+-- This file is part of the Wire Server implementation.
+--
+-- Copyright (C) 2023 Wire Swiss GmbH <opensource@wire.com>
+--
+-- This program is free software: you can redistribute it and/or modify it under
+-- the terms of the GNU Affero General Public License as published by the Free
+-- Software Foundation, either version 3 of the License, or (at your option) any
+-- later version.
+--
+-- This program is distributed in the hope that it will be useful, but WITHOUT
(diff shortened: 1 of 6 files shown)


integration/test/Test/MLS/Unreachable.hs
```diff
@@ -0,0 +1,98 @@
...
+    void $ withWebSocket bob $ \ws -> do
+      void $ createAddCommit alice1 [bob] >>= sendAndConsumeCommitBundle
+      awaitMatch 10 isMemberJoinNotif ws
+    mp <- createAddCommit alice1 [charlie]
+    pure (mp, thirdDomain)
+
+  -- try adding Charlie now that his backend is unreachable
+  bindResponse (postMLSCommitBundle addCommit.sender (mkBundle addCommit)) $ \resp -> do
+    resp.status `shouldMatchInt` 533
+    (resp.json %. "unreachable_backends" & asList) `shouldMatch` [d]
```
battermann, line 48: why not `shouldMatchSet` as below?

battermann (approved):

integration/test/Test/MLS/Unreachable.hs
mdimjasevic: For no particular reason.

I believe this way it is fine because it's only one list element so there's no concern about what element comes where.

PR #3678, opened 2023-10-30
Update group state after application messages

pcapriotti:
After an application message the ratchet is updated, therefore we need to save the updated group state so that future messages are generated correctly.

This PR includes an mls-test-cli update. The new mls-test-cli version modifies the `message` command to include both `group-in` and `group-out` options, as other similar commands already do.

https://wearezeta.atlassian.net/browse/WPB-5002

## Checklist

### DIFF
--- a/changelog.d/5-internal/multiple-messages
+++ b/changelog.d/5-internal/multiple-messages
@@ -1,0 +1,1 @@
+Update group state after application message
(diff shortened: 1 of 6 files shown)


MangoIV (approved):
Looks good to me :+1:

PR #90, opened 2017-10-12
Optional YAML configuration for Brig

elverkilde:
Adds optional YAML parsing for loading the configuration, the old commandline options are still supported. Only brig and brig-integration is implemented, brig-schema and brig-index needs to be done.

PR #88, opened 2017-10-11
Issue 581 v2

tiago-loureiro:
Replaces https://github.com/wireapp/wire-server/pull/73

 * Allows members of the same binding to create 1to1 conversations inside a team
 * Prevents members of the same binding team to create connections (they are unnecessary and can create some confusion)

PR #3680, opened 2023-10-30
Fix galley DB migrations

battermann:
The migrations `Remove member_client` and `Truncate mls_group_member_client` both had the version 88, so the latter was overridden by the former in the meta table.

## Checklist


services/galley/src/Galley/Schema/V89_TruncateMLSGroupMemberClient.hs
```diff
@@ -28,6 +28,6 @@ import Text.RawString.QQ
 -- could've been. This migration just deletes old data. This could break some
 -- conversations/users in unknown ways. But those are most likely test users.
 migration :: Migration
-migration = Migration 88 "Truncate mls_group_member_client" $ do
+migration = Migration 89 "Truncate mls_group_member_client" $ do
```
akshaymankar, line 31: Does this mean that this migration will run again on some instances?

battermann: Will try putting the conflicting migrations into 1 file, as discuessed.

services/galley/src/Galley/Schema/Run.hs
```diff
@@ -87,8 +87,8 @@ import Galley.Schema.V84_MLSSubconversation qualified as V84_MLSSubconversation
 import Galley.Schema.V85_MLSDraft17 qualified as V85_MLSDraft17
 import Galley.Schema.V86_TeamFeatureMlsMigration qualified as V86_TeamFeatureMlsMigration
 import Galley.Schema.V87_TeamFeatureSupportedProtocols qualified as V87_TeamFeatureSupportedProtocols
-import Galley.Schema.V88_TruncateMLSGroupMemberClient qualified as V88_TruncateMLSGroupMemberClient
-import Galley.Schema.V89_RemoveMemberClient qualified as V89_RemoveMemberClient
+import Galley.Schema.V88_RemoveMemberClient qualified as V89_RemoveMemberClient
+import Galley.Schema.V89_TruncateMLSGroupMemberClient qualified as V88_TruncateMLSGroupMemberClient
```
mdimjasevic, line 91: In the PR description you say: "The migrations Remove member_client and Truncate mls_group_member_client both had the version 88". At least the module names had the numbers right. However, why swapping their numbers?

I see further down that both had `Migration 88`, but I am not sure what the consequence of that is/was.

battermann: Both were run, but the last one overwrote the entry in the meta table.

mdimjasevic:
@battermann , furthermore, can you test this locally?

In other words, can we populate the local database with some dummy data (in an automated way) and then run the old migrations first, put the database aside, and then repeat that, but with the new migrations from this PR in separation and compare the outcome?

mdimjasevic:
And one more note. If you'll be squashing the two migrations into one, my hunch is it'd be wise to make new no-op migrations #88 and #89 and then have a migration #90 that applies the old migrations #88 and #89 in an idempotent way as a new migration #90. I'm no DB expert, but my hunch is this would make sense to avoid any breakage.

mdimjasevic (approved):

battermann:
Note that the change in `cassandra-schema.cql` comes from `make git-add-cassandra-schema` wasn't run for the previous migration(s).

PR #3682, opened 2023-10-31
wire-server chart: Disable integration chart by default

smatting:
Fixup to #3171. This PR updates the `wire-server` chart so that it doesn't include the `integration` chart by default.
Without this PR the [integration](https://github.com/wireapp/wire-server/tree/36bae6e2ce25b3cddeb9c7d72b2b48fb9d89b270/charts/integration) chart is installed, which creates an superfluous kubernetes objects, including an (internal) ingress for a subdomain of `svc.cluster.local`.

Disabling works because of [this tag](https://github.com/wireapp/wire-server/blob/f87b3b39803d79906e79ba573eed2df46b02e9ae/charts/wire-server/requirements.yaml#L131C1-L131C1)

## Checklist

### DIFF
--- a/changelog.d/5-internal/disable-integration-chart
+++ b/changelog.d/5-internal/disable-integration-chart
@@ -1,0 +1,1 @@
+Disable `integration` subchart of `wire-server` by default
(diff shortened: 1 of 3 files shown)


supersven (approved):
LGTM :+1: 

(Let's wait for the CI before merging.)

PR #3686, opened 2023-11-01
use crate2nix for some of our rust crates

flokli:
The current workflow for building rust crates is error-prone. It uses FOD, which can break in very subtle ways, and it's only visible when you either want to update the package, or some FOD got GC'ed from your store (or binary cache).

crate2nix allows a better workflow: given a source containing Cargo.toml and Cargo.lock, it can produce a `Cargo.nix` file (and `crate-hashes.json`), which will create per-crate-granularity nix derivation, allowing faster rebuilds of partial updates, as well as less flakyness when it comes to FODs.

`cryptobox` required some hacks, as `cabal` (or our custom plumbing of calling cabal2Nix) expects a single output. I tried to work my way around this with emitting a pkg-config file, but couldn't get it to work, so now emit a single-output derivation (but the underlying build still uses crate2nix).

`libzauth-c` didn't build at all, it failed with [this errorI(https://stackoverflow.com/questions/73273886/why-do-i-get-prehashsignature-is-not-implemented-for-signature-error-when-usin). Instead of trying to push down certain feature flags, I tried bumping our `jwt-simple` pin to the most recent commit on this branch (we should be having multiple versions of this in flight anyways), and it built. :tada: 

`mls-test-cli` doesn't work with crate2nix yet, due to a transitive build dependency being too crazy about `#cfg` usage in its `Cargo.toml` - I opened an upstream issue for that. Added a comment linking to the issue there.

For `rusty_jwt_tools`, I can't even create a `Cargo.nix`. It seems like the workaround in https://github.com/wireapp/rusty-jwt-tools/pull/54#issuecomment-1490530012 is not sufficient for this mode of operation. We should really get a Cargo.lock file into the project root. Added a comment there, too.

## Checklist


nix/pkgs/cryptobox/README.md
```diff
@@ -0,0 +1,7 @@
+# How to update
+
+```
+crate2nix source add git https://github.com/wireapp/cryptobox-c --rev $your-rev
+crate2nix generate
+```
+
```
MangoIV, line 7: can we (eventually) integrate this into `./hack/bin/generate-local-nix-packages.sh` such that the overhead for non-nix-affine people is lower?

MangoIV (approved):
looks good to me :)

nix/pkgs/cryptobox/README.md
flokli: It doesn't need to be regenerated, only if you consciously update the rev you want to use.

Also, the process isn't 100% automatic, you might need to adjust the nix expression, for example in case native build dependencies need to be added if it fails to build.

I'd rather keep this (more readable) documentation than just being hidden somewhere in a script doing another thing.

MangoIV: > It doesn't need to be regenerated, only if you consciously update the rev you want to use.

oh yeah; these are external to wire-server anyways, yeah I think it makes sense like this :)

flokli: That's not entirely true - a change in `Cargo.toml` `libzauth-c` would indeed update `Cargo.nix`.

I want to tackle this in a more general fashion in a followup, once the other two crates are migrated.

PR #3691, opened 2023-11-03
Swagger docs: new line after fed call tag

battermann:
## Checklist

before:

image

after:

image


supersven (approved):
LGTM :+1: 

I you feel like it, you could also try to use `<p>...</p>` to give the HTML more structure. However, `<br>` is for sure an improvement to `\n`.

battermann:
> LGTM 👍
> 
> I you feel like it, you could also try to use `<p>...</p>` to give the HTML more structure. However, `<br>` is for sure an improvement to `\n`.

Yes, the problem is that each federated call is appended separately to the complete description. In order to put them all in a list or paragraph, we'd have to parse the existing description and insert each fed call at the right place. Which is not worth the effort IMO.

PR #3700, opened 2023-11-08
http2-manager: Allow requests to work depite other requests being stuck on server side

akshaymankar:
## Checklist


akshaymankar:
One part of the problem is in the server side. The server hardcodes the number of workers to 3, but the max concurrency is 100. So the client gets stuck without any warning. This might get resolved by https://github.com/kazu-yamamoto/http2/issues/98

PR #3708, opened 2023-11-10
WBP-5133 External partners search restriction enforced by backend

battermann:
https://wearezeta.atlassian.net/browse/WPB-5133

This PR enforces the team member search restrictions of an external partner by the backend.

It's an alternative work around that doesn't need any client work.

However, I am not sure how this scales for very large teams, or teams with a very large amount of conversations.

## UPDATE

The behavior was changed to members endpoint only returning the invitee and self user for external partners.

## Checklist

### DIFF
--- a/integration/test/Test/Brig.hs
+++ b/integration/test/Test/Brig.hs
@@ -3,6 +3,7 @@ module Test.Brig where
 import API.Brig qualified as BrigP
 import API.BrigInternal qualified as BrigI
 import API.Common (randomName)
+import API.Galley
 import API.Galley qualified as Galley
 import Data.Aeson.Types hiding ((.=))
 import Data.Set qualified as Set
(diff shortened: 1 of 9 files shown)


integration/test/Test/Brig.hs
```diff
@@ -18,14 +19,45 @@ import Testlib.Prelude
...
+  bindResponse (Galley.getTeamMembers partner tid) $ \resp -> do
+    resp.status `shouldMatchInt` 200
+    assertContainsUserIds resp [owner, tm1, partner]
+  where
+    assertContainsUserIds :: Response -> [Value] -> App ()
+    assertContainsUserIds resp users = do
+      members <- resp.json %. "members" & asList
+      userIds <- for members (\m -> m %. "user")
+      expected <- for users objId
+      userIds `shouldMatchSet` expected
```
akshaymankar, line 59: It looks a bit awkward that we're testing galley features in the module called `Test.Brig`. Perhaps we should just move this half of the test to the other module? 
Although I think separating these "integration" tests by where most of the code lives is not very nice. We should probably name our testing modules according to the domain.

services/galley/src/Galley/API/Teams.hs
```diff
@@ -493,10 +495,22 @@ getTeamMembers ::
...
+    else do
+      -- If the user does not have the SearchContacts permission (e.g. the external partner),
+      -- we only return the team members they are in a conversation with plus the invitee
+      teamCids <- fmap (\tConv -> tConv ^. conversationId) <$> E.getTeamConversations tid
+      userTeamCids <- E.selectConversations uid teamCids
+      convs <- E.getConversations userTeamCids
+      let lConvMems = Conv.lmId <$> (convs >>= convLocalMembers)
+      let invitee = member ^. invitation <&> fst
+      let uids = nub (maybe (uid : lConvMems) (: uid : lConvMems) invitee)
+      E.selectTeamMembersPaginated tid uids mState mLimit <&> toTeamMembersPage member
```
akshaymankar, line 513: This is not very safe to do for large teams. I think we should create another index for these things.

akshaymankar (changes requested):
I think we're gonna run into trouble if very big teams start inviting externals. So let's think a bit more about this problem.

services/galley/src/Galley/API/Teams.hs
```diff
@@ -493,10 +493,18 @@ getTeamMembers ::
...
   let mState = C.PagingState . LBS.fromStrict <$> (mbPagingState >>= mtpsState)
   let mLimit = fromMaybe (unsafeRange Public.hardTruncationLimit) mbMaxResults
-  E.listTeamMembers @CassandraPaging tid mState mLimit <&> toTeamMembersPage member
+  if member `hasPermission` SearchContacts
+    then E.listTeamMembers @CassandraPaging tid mState mLimit <&> toTeamMembersPage member
+    else do
+      -- If the user does not have the SearchContacts permission (e.g. the external partner),
+      -- we only return the person who invited them and the self user.
+      let invitee = member ^. invitation <&> fst
+      let uids = maybe [uid] (: [uid]) invitee
```
akshaymankar, line 506: Nit: Feel free to ignore.
```suggestion
      let uids = uid : maybeToList invitee
```

akshaymankar (approved):
Looks great :rocket:

PR #3709, opened 2023-11-13
Avoid generating client IDs with leading zeros

pcapriotti:
Leading zeros in client IDs expose an issue with the protobuf messaging API, since there client IDs are represented as uint64. This commit works around th eissue by removing any leading zeros when generating a new client ID.

## Checklist

### DIFF
--- a/libs/wire-api/src/Wire/API/User/Client/Prekey.hs
+++ b/libs/wire-api/src/Wire/API/User/Client/Prekey.hs
@@ -40,6 +40,7 @@ import Data.ByteString.Conversion (toByteString')
 import Data.Id
 import Data.OpenApi qualified as S
 import Data.Schema
+import Data.Text qualified as T
 import Data.Text.Ascii (encodeBase16)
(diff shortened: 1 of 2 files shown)


libs/wire-api/src/Wire/API/User/Client/Prekey.hs
```diff
@@ -70,6 +71,7 @@ instance ToSchema Prekey where
 clientIdFromPrekey :: Prekey -> ClientId
 clientIdFromPrekey =
   ClientId
+    . T.dropWhile (== '0')
```
MangoIV, line 74:
```suggestion
  ClientId
    -- FIXME the clients expect UInts on the protobufs, 
    -- We will eventually not need this change anymore 
    . T.dropWhile (== '0')
```

MangoIV (approved):
Looks good to me

mdimjasevic (approved):

pcapriotti:
Superseded by https://github.com/wireapp/wire-server/pull/3713.

PR #3711, opened 2023-11-13
[WPB-5376] Migrated from cryptonite to crypton.

elland:
Includes a modification of password hashing to be more lower level, which sets the groundwork for supporting multiple hashing algos.

The core of the PR is the `Password.hs` file, most of the remaining changes are about bumping and pinning dependency versions.


libs/wire-api/src/Wire/API/Password.hs
```diff
@@ -56,14 +103,89 @@ genPassword =
...
+showT :: Show a => a -> Text
+showT = Text.pack . show
+{-# INLINE showT #-}
+
+-- | Decodes a base64 'Text' to a regular 'ByteString' (if possible)
+from64 :: Text -> Maybe ByteString
+from64 = toMaybe . B64.decode . Text.encodeUtf8
+  where
+    toMaybe = either (const Nothing) Just
+{-# INLINE from64 #-}
```
MangoIV, line 191: I'm not entirely sure why these are exported from here.

nix/haskell-pins.nix
```diff
@@ -99,18 +106,25 @@ let
         sha256 = "0dgizj1kc135yzzqdf5l7f5ax0qpvrr8mxvg7s1dbm01cf11aqzn";
       };
     };
+    HaskellNet-SSL = {
+      src = fetchgit {
+        url = "https://github.com/MangoIV/HaskellNet-SSL";
+        rev = "c2844b63a39f458ffbfe62f2ac824017f1f84453";
+        sha256 = "sha256-1mu/yEAWr3POY4MHRomum0DDvs5Qty1JvP3v5GS2u64=";
+      };
+    };
```
MangoIV, line 115: I should probably open a PR, right?

nix/haskell-pins.nix
```diff
@@ -182,12 +196,12 @@ let
...
-    # This can be removed once postie 0.6.0.3 (or later) is in nixpkgs
+    # This can be removed once postie with TLS 1.9 is on nixpkgs.
     postie = {
       src = fetchgit {
-        url = "https://github.com/alexbiehl/postie.git";
-        rev = "c92702386f760fcaa65cd052dc8114889c001e3f";
-        sha256 = "sha256-yiw6hg3guRWS6CVdrUY8wyIDxoqfGjIVMrEtP+Fys0Y=";
+        url = "https://github.com/wireapp/postie.git";
+        rev = "43b6d1d21d56e567077c194d49efb92e777e7628";
+        sha256 = "sha256-DKugy4EpRsSgaGvybdh2tLa7HCtoxId+7RAAAw43llA=";
```
MangoIV, line 204: Is there an open PR for this? (this is our own fork no?)

nix/haskell-pins.nix
```diff
@@ -234,6 +261,18 @@ let
...
       sha256 = "sha256-AdxxKWXdUjZiHLDj6iswMWpycs7mFB8eKhBR4ljF6kk=";
     };
+    # http-client = {
+    #   version = "0.7.15";
+    #   sha256 = "sha256-EGvMlwZSka9H90XXZWGVYtZecloWkj0efFLyB8Eew2k=";
+    # };
+    # http-client-tls = {
+    #   version = "0.3.6.3";
+    #   sha256 = "sha256-CB8cnaDV4Vmdq0ct3bQsYfqKtGPRRATPRA649AzmcaU=";
+    # };
```
MangoIV, line 271: do we still need these?

nix/manual-overrides.nix
```diff
@@ -42,6 +42,11 @@ hself: hsuper: {
 
   http2-manager = hlib.enableCabalFlag hsuper.http2-manager "-f-test-trailing-dot";
 
+  crypton-connection = hlib.markUnbroken hsuper.crypton-connection;
+  # Patched dependency on crypton-connection
+  HaskellNet-SSL = hlib.markUnbroken hsuper.HaskellNet-SSL;
```
MangoIV, line 47: if we override this dependency anyways, I don't think we need to mark it unbroken

nix/manual-overrides.nix
```diff
@@ -56,6 +61,8 @@ hself: hsuper: {
   # Explicitly enable haddock because cabal2nix disables it for packages with
   # internal libraries
   cql-io = hlib.doHaddock (hlib.dontCheck hsuper.cql-io);
+  connection = null;
```
MangoIV, line 64: have you tried removing this? I don't think, it's needed, if anything, set it to `crypton-connection`, perhaps?

libs/wire-api/src/Wire/API/Password.hs
```diff
@@ -14,38 +15,84 @@
...
-    verifyPassword,
-  )
-where
+module Wire.API.Password where
+
+-- ( Password,
+--   genPassword,
+--   mkSafePassword,
+--   verifyPassword,
+-- )
```
mdimjasevic, line 26: Leftover?

MangoIV (approved):
Really nice, some minor comments! :rocket:

libs/wire-api/src/Wire/API/Password.hs
```diff
@@ -14,38 +15,84 @@
...
 
 instance Cql Password where
   ctype = Tagged BlobColumn
 
-  fromCql (CqlBlob lbs) = pure . Password . EncryptedPass $ toStrict lbs
+  fromCql (CqlBlob lbs) = pure . Password . Text.decodeUtf8 . toStrict $ lbs
   fromCql _ = Left "password: expected blob"
 
-  toCql = CqlBlob . fromStrict . getEncryptedPass . fromPassword
+  toCql = CqlBlob . fromStrict . Text.encodeUtf8 . fromPassword
```
mdimjasevic, line 54: How does this interact with existing data in the DB?

libs/wire-api/src/Wire/API/Password.hs
```diff
@@ -56,14 +103,89 @@ genPassword =
...
+-- | Same as 'show' but works on 'Text'
+showT :: Show a => a -> Text
+showT = Text.pack . show
+{-# INLINE showT #-}
+
+-- | Decodes a base64 'Text' to a regular 'ByteString' (if possible)
+from64 :: Text -> Maybe ByteString
+from64 = toMaybe . B64.decode . Text.encodeUtf8
+  where
+    toMaybe = either (const Nothing) Just
```
mdimjasevic, line 190:
```suggestion
from64 = hush . B64.decode . Text.encodeUtf8
```

libs/wire-api/src/Wire/API/Password.hs
```diff
@@ -14,38 +15,84 @@
...
+{-# OPTIONS_GHC -Wno-incomplete-uni-patterns #-}
 
-module Wire.API.Password
-  ( Password,
-    genPassword,
-    mkSafePassword,
-    verifyPassword,
-  )
-where
+module Wire.API.Password where
```
mdimjasevic, line 20: Haven't looked into details, but probably you want to export only the smart constructors so as to avoid opening a possibility to construct an arbitrary `Password` by hand that might be invalid.

mdimjasevic (commented):
Overall, this looks good to me. I left a few inlined comments.

What is not clear to me is what is the http-client issue. In the ticket you have this:

    The only current hiccup is our fork of http-client, that is lagging over 3 years behind upstream, but since the fork doesn’t change that much, it adds support to reusing connections, it should be fine to sync with upstream.

However, I can't see any change related to that in this PR. Should it have been part of the PR?

libs/wire-api/src/Wire/API/Password.hs
MangoIV: (maybe also rename the `Password` constructor to `UnsafeMkPassword`?)

elland: Good catch. I was exporting more to test that things were compatible with the previous hash function. Forgot to constrain it again 👍

nix/haskell-pins.nix
elland: https://github.com/alexbiehl/postie/pull/4

libs/wire-api/src/Wire/API/Password.hs
elland: It is still the same algorithm, it should be 100% the same 🙏

libs/wire-api/src/Wire/API/Password.hs
elland: I'm not sure it's worth adding the `errors` package as a dependency for this. 🤔

libs/wire-api/src/Wire/API/Password.hs
mdimjasevic: Do we have any golden and roundtrip tests for this? They'd make me feel easier about this change :)

libs/wire-api/src/Wire/API/Password.hs
MangoIV: there's also one that does the same thing for `MonadError`, I'd guess we already depend on that.

MangoIV: https://hackage.haskell.org/package/either-5.0.2/docs/Data-Either-Combinators.html#v:rightToMaybe

libs/wire-api/src/Wire/API/Password.hs
elland: I think I can property test it 👍 let's see

mdimjasevic (approved):
Thanks for the password tests!

PR #3712, opened 2023-11-13
Reduce the size for CI image by getting rid of 2 GHCs

akshaymankar:
apply-refact refers to the GHC that it builds with by using the GHC.Paths module. This tool is actually not required in CI.

ormolu was referring to GHC and all its haskell dependencies as "propagatedBuildInputs". Using `hlib.justStaticExecutables` we can get rid of these.

## Checklist


akshaymankar:
This reduces the compressed image size by ~500 MB. But it reduces the exploded image size by more than 6 GB.

supersven (approved):
:+1:

PR #3716, opened 2023-11-14
[feat] add support for ghc-flakr's hs-run executable

MangoIV:
This PR is according to our discussion in C&D from November 14, 2023.

Mind that the pinning is done by `ghc-flakr` itself to ensure caching. 

https://github.com/wireapp/ghc-flakr

## Checklist


smatting (approved):
let's give this a try :+1: It's probably only going to be useful once we include `wire-api` as a dependency. I'm wondering if we should maybe inline the flake into `wire-server` itself

MangoIV:
> It's probably only going to be useful once we include wire-api as a dependency

@smatting I can do that right now :relieved:

PR #3722, opened 2023-11-16
Fix shell.nix

supersven:
The previous attribute path didn't exist and led to error:

```
➜  wire-server git:(develop) nix-shell --pure shell.nix
error: attribute 'wireServer' missing

       at /home/sven/src/wire-server/shell.nix:1:1:

            1| (import ./nix).pkgs.wireServer.devShell
             | ^
            2|
(use '--show-trace' to show detailed location information)
```


supersven:
We're convinced that this file should be deleted instead.

PR #3723, opened 2023-11-16
[fix] stern/backoffice conference calling TTL

battermann:
## Checklist


elland (approved):

battermann:
> maybe also add an integration test?

It's not possible because this is solely a UI problem, not allowing text input.

PR #3726, opened 2023-11-16
Delete shell.nix

supersven:
It's broken. And, the official way to get a nix env for this project is to use direnv.


MangoIV (approved):
👍

PR #3732, opened 2023-11-21
WBP-5388 restrict contact search results according to team federation policy

battermann:
## Checklist


libs/wire-api-federation/src/Wire/API/Federation/API/Brig.hs
```diff
@@ -44,7 +44,7 @@ import Wire.API.UserMap (UserMap)
 import Wire.API.Util.Aeson (CustomEncoded (..))
 import Wire.Arbitrary (GenericUniform (..))
 
-newtype SearchRequest = SearchRequest {term :: Text}
+data SearchRequest = SearchRequest {term :: Text, from :: Maybe TeamId, onlyInTeams :: Maybe [TeamId]}
```
mdimjasevic, line 47: Can you please break this into multiple lines and add Haddock for the new fields?

mdimjasevic (commented):
Why not include golden tests for `FederationRestriction`? They are tested indirectly via `FedeerationDomainConfig`, but at least I search for these golden tests when I want to see how JSON looks like for a given type, and in a few months time I might already forget there are golden tests for this larger type.

integration/test/Test/Search.hs
```diff
@@ -0,0 +1,222 @@
...
+      resp.status `shouldMatchInt` 200
+      docs <- resp.json %. "documents" >>= asList
+      case docs of
+        [] -> assertFailure "Expected a non empty result, but got an empty one"
+        doc : _ -> objQid doc `shouldMatch` objQid u2
+
+--------------------------------------------------------------------------------
+-- FEDERATION SEARCH WITH TEAM RESTRICTIONS
+
+data Restriction = AllowAll | TeamAllowed | TeamNotAllowed
```
mdimjasevic, line 69: What type from `wire-api` does this map to?

integration/test/Test/Search.hs
```diff
@@ -0,0 +1,222 @@
...
+  deriving (Eq, Ord, Show)
+
+data FedUserSearchTestCase = FedUserSearchTestCase
+  { searchPolicy :: String,
```
mdimjasevic, line 73: Why not make this an enumerated type instead of a string?

services/brig/src/Brig/API/Federation.hs
```diff
@@ -118,11 +120,14 @@ sendConnectionAction originDomain NewConnectionRequest {..} = do
...
 getUserByHandle ::
-  Member GalleyProvider r =>
+  ( Member GalleyProvider r,
+    Member FederationConfigStore r
+  ) =>
   Domain ->
   Handle ->
   ExceptT Error (AppT r) (Maybe UserProfile)
 getUserByHandle domain handle = do
+  -- TODO: should we do the team federation check here, too?
```
mdimjasevic, line 130: Confirm with someone from the product team, but I think the check should be performed here.

mdimjasevic (commented):
Here is a partial review. I have to leave now and I will continue tomorrow.

integration/test/API/BrigInternal.hs
```diff
@@ -63,7 +65,13 @@ instance ToJSON FedConn where
     Aeson.object
       [ "domain" .= d,
         "search_policy" .= s,
-        "restriction" .= r
+        "restriction"
+          .= maybe
+            (Aeson.object ["tag" .= "allow_all", "value" .= Aeson.Null])
+            ( \teams ->
+                Aeson.object ["tag" .= "restrict_by_team", "value" .= Aeson.Array (Vector.fromList (Aeson.String . cs <$> teams))]
```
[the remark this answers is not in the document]
battermann: I think it converts a String into a ByteString in this case. Remember that in the integration tests we do not have access to wire-api. However, Aeson.toJSON works, too.

integration/test/Test/Search.hs
```diff
@@ -0,0 +1,222 @@
...
+    uidD2 <- objId u2
+
+    bindResponse (BrigP.searchContacts u1 (u2 %. "name") d2) $ \resp -> do
+      resp.status `shouldMatchInt` 200
+      docs <- resp.json %. "documents" >>= asList
+      case docs of
+        [] -> assertFailure "Expected a non empty result, but got an empty one"
+        doc : _ -> doc %. "id" `shouldMatch` uidD2
+
+testRemoteUserSearchExactHandle :: HasCallStack => App ()
```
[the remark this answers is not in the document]
battermann: Yes, there is such a test, part of the test cases for federated team search.

integration/test/Test/Search.hs
battermann: To keep it simple.

integration/test/Test/Search.hs
battermann: No type directly, `FederationRestriction` is the closest.

services/brig/src/Brig/API/Federation.hs
battermann: I'll remove this comment and create a ticket instead. This should be a separate PR, IMO.

battermann: After looking into this, I think, that it is not necessary to apply the team federation policy here. The result of the design was that we needed to prevent connections and search in case of a team federation that is not allowed. The `getUserByHandle` is part of a federated handler and cannot be called by clients directly anyway. So this should be fine. Let me know if you don't agree.

integration/test/API/BrigInternal.hs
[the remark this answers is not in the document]
mdimjasevic: Yeah, I forgot this is for a `[String]` value, and not a `[Data.Id.TeamId]` value. The `Data.Id` module is not in `wire-api`, but in `types-common`, though. However, maybe this is not to be used in the `integration` package too, but I'm not sure.

services/brig/src/Brig/Effects/FederationConfigStore/Cassandra.hs
```diff
@@ -0,0 +1,240 @@
...
+
+getFederationConfig' :: MonadClient m => [FederationDomainConfig] -> Domain -> m (Maybe FederationDomainConfig)
+getFederationConfig' cfgs rDomain = do
+  let mFromCfgFile = find ((== rDomain) . domain) cfgs
+  mCnf <- retry x1 (query1 q (params LocalQuorum (Identity rDomain)))
+  case mCnf of
+    Nothing -> pure mFromCfgFile
+    Just (p, rInt) -> do
+      r <- toRestriction rDomain rInt
+      pure $ Just $ FederationDomainConfig rDomain p r
```
mdimjasevic, line 103:
```suggestion
getFederationConfig' cfgs rDomain = case find ((== rDomain) . domain) cfgs of
  Just cfg -> pure . Just $ cfg -- the configuration from the file has precedence
  Nothing -> do
    mCnf <- retry x1 (query1 q (params LocalQuorum (Identity rDomain)))
    pure $ fmap (\(p, rInt) -> FederationDomainConfig rDomain p (toRestriction rDomain rInt)) mCnf
```

mdimjasevic (commented):
Continuing tomorrow, but here's what I have so far.

services/brig/src/Brig/API/Federation.hs
```diff
@@ -32,6 +31,8 @@ import Brig.API.User qualified as API
 import Brig.App
 import Brig.Data.Connection qualified as Data
 import Brig.Data.User qualified as Data
+import Brig.Effects.FederationConfigStore (FederationConfigStore)
+import Brig.Effects.FederationConfigStore qualified as FederationConfigStore
```
mdimjasevic, line 35:
```suggestion
import Brig.Effects.FederationConfigStore qualified as E
```

services/brig/src/Brig/API/Internal.hs
```diff
@@ -38,12 +37,13 @@ import Brig.Code qualified as Code
...
 import Brig.Data.Client qualified as Data
 import Brig.Data.Connection qualified as Data
-import Brig.Data.Federation qualified as Data
 import Brig.Data.MLS.KeyPackage qualified as Data
 import Brig.Data.User qualified as Data
 import Brig.Effects.BlacklistPhonePrefixStore (BlacklistPhonePrefixStore)
 import Brig.Effects.BlacklistStore (BlacklistStore)
 import Brig.Effects.CodeStore (CodeStore)
+import Brig.Effects.FederationConfigStore (AddFederationRemoteResult (..), AddFederationRemoteTeamResult (..), FederationConfigStore, UpdateFederationResult (..))
+import Brig.Effects.FederationConfigStore qualified as FederationConfigStore
```
mdimjasevic, line 46:
```suggestion
import Brig.Effects.FederationConfigStore qualified as E
```

services/brig/src/Brig/API/Internal.hs
```diff
@@ -223,121 +223,63 @@ federationRemotesAPI =
...
-    & maybe id (\v cfg -> cfg {strategy = v}) ms
-    & (\cfg -> cfg {remotes = nub $ db <> mf})
-    & maybe id (\v cfg -> cfg {updateInterval = min 1 v}) mu
-    & pure
-
-updateFederationRemote :: Domain -> FederationDomainConfig -> ExceptT Brig.API.Error.Error (AppT r) ()
+    AddFederationRemoteDivergingConfig cfg ->
+      throwError . fedError . FederationUnexpectedError $
+        "keeping track of remote domains in the brig config file is deprecated, but as long as we \
+        \do that, adding a domain with different settings than in the config file is nto allowed.  want "
```
mdimjasevic, line 260:
```suggestion
        \do that, adding a domain with different settings than in the config file is not allowed.  want "
```

mdimjasevic (commented):
I'm done with reviewing. I've left cosmetic comments (such as importing an effect module), but also bigger ones. In some places I am not sure the logic is correct.

mdimjasevic:
I forgot to add one thing. Yesterday I thought it is fine to leave that checking of duplicates in the configuration file for a follow-up PR. But now that I'm done with seeing the whole PR, I'd ask to change it in this PR as it's too messy and error-prone to keep it the way it is. One step in the right direction that would make it much simpler is to run the interpreter with a `Map Domain FederationDomainConfig` value instead of a `[FederationDomainConfig]` value.

integration/test/Test/Search.hs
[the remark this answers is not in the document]
battermann: Yes. In fact this test and the one above are covered by the new test cases.

services/brig/src/Brig/Effects/FederationConfigStore/Cassandra.hs
```diff
@@ -0,0 +1,240 @@
...
+  -- https://docs.wire.com/understand/federation/backend-communication.html#configuring-remote-connections,
+  -- http://docs.wire.com/developer/developer/federation-design-aspects.html#configuring-remote-connections-dev-perspective
+  remotes <-
+    (<>)
+      <$> getFederationRemotes
+      <*> remotesListFromCfgFile cfgs
+
+  defFederationDomainConfigs
+    & maybe id (\v cfg -> cfg {strategy = v}) mFedStrategy
+    & (\cfg -> cfg {remotes = remotes})
```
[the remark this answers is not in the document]
battermann: This code existed before this PR, but yes, granted.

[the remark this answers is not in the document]
battermann: This is the alternative:

```haskell
  pure $
    FederationDomainConfigs
      { strategy = fromMaybe defFederationDomainConfigs.strategy mFedStrategy,
        remotes = remotes,
        updateInterval = defFederationDomainConfigs.updateInterval
      }
```

I am not sure what is better.

services/brig/src/Brig/Effects/FederationConfigStore/Cassandra.hs
```diff
@@ -0,0 +1,240 @@
...
+
+getFederationConfigs' :: forall m. (MonadClient m) => Maybe FederationStrategy -> [FederationDomainConfig] -> m FederationDomainConfigs
+getFederationConfigs' mFedStrategy cfgs = do
+  -- FUTUREWORK: we should solely rely on `db` in the future for remote domains; merging
+  -- remote domains from `cfg` is just for providing an easier, more robust migration path.
+  -- See
+  -- https://docs.wire.com/understand/federation/backend-communication.html#configuring-remote-connections,
+  -- http://docs.wire.com/developer/developer/federation-design-aspects.html#configuring-remote-connections-dev-perspective
+  remotes <-
+    (<>)
```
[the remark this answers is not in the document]
battermann: I think, because we guard the create and update of configs, we know that they must be consistent when we read them.

services/brig/src/Brig/User/API/Search.hs
```diff
@@ -73,18 +78,26 @@ search searcherId searchTerm maybeDomain maybeMaxResults = do
   -- looked up again.
   ensurePermissionsOrPersonalUser searcherId [SearchContacts]
   federationDomain <- viewFederationDomain
+  mSearcherTeamId <- lift $ wrapClient $ DB.lookupUserTeam searcherId
   let queryDomain = fromMaybe federationDomain maybeDomain
   if queryDomain == federationDomain
```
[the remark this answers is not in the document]
battermann: Yeah, this is pre PR code. Didn't really want to change too much that is not PR related.

services/brig/src/Brig/User/API/Search.hs
```diff
@@ -73,18 +78,26 @@ search searcherId searchTerm maybeDomain maybeMaxResults = do
...
     msg (val "searchRemotely")
       ~~ field "domain" (show domain)
       ~~ field "searchTerm" searchTerm
-  searchResponse <- Federation.searchUsers domain (FedBrig.SearchRequest searchTerm) !>> fedError
+  mFedCnf <- lift $ liftSem $ FederationConfigStore.getFederationConfig domain
+  let onlyInTeams = case restriction <$> mFedCnf of
+        Just FederationRestrictionAllowAll -> Nothing
+        Just (FederationRestrictionByTeam teams) -> Just teams
+        -- if we are not federating at all, we also do not allow to search any remote teams
+        Nothing -> Just []
```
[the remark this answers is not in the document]
battermann: Nothing means no restriction.

mdimjasevic (approved):
Ship it! :ship:

PR #3736, opened 2023-11-22
Migrate away from our http-client fork, use upstream.

elland:
https://wearezeta.atlassian.net/browse/WPB-5677

Changed HTTP back to use upstream. Instead of hooking into `withConnection` we use the `vpCallback` to do fingerprint verification instead.


libs/bilge/src/Bilge/IO.hs
```diff
@@ -162,17 +162,17 @@ instance MonadIO m => MonadHttp (SessionT m) where
...
+      toBilgeResponse bodyReader WaiTest.SResponse {WaiTest.simpleStatus, WaiTest.simpleHeaders} oReq =
         Client.Response
           { responseStatus = simpleStatus,
             -- I just picked an arbitrary version; shouldn't matter.
             responseVersion = http11,
             responseHeaders = simpleHeaders,
             responseBody = bodyReader,
-            responseOriginalRequest = originalReq,
             Client.responseCookieJar = mempty,
-            Client.responseClose' = Client.ResponseClose $ pure ()
```
MangoIV, line 175: these changes can probably be reverted...

services/brig/src/Brig/App.hs
```diff
@@ -400,29 +405,56 @@ initHttp2Manager = do
...
-  SSL.contextAddOption ctx SSL_OP_NO_SSLv3
-  SSL.contextSetCiphers ctx rsaCiphers
-  -- We use public key pinning with service providers and want to
-  -- support self-signed certificates as well, hence 'VerifyNone'.
-  SSL.contextSetVerificationMode ctx SSL.VerifyNone
-  SSL.contextSetDefaultVerifyPaths ctx
-  mgr <-
+initExtGetManager :: IO ([Fingerprint Rsa] -> IO Manager)
+initExtGetManager =
+  pure $ \fingerprints -> do
```
MangoIV, line 410: this is the only thing that makes it `IO`, should probably remove it and lift it at the appropriate place.

services/brig/src/Brig/App.hs
```diff
@@ -400,29 +405,56 @@ initHttp2Manager = do
...
+    ctx <- SSL.context
+    Just sha <- getDigestByName "SHA256"
+
+    SSL.contextAddOption ctx SSL_OP_NO_SSLv2
+    SSL.contextAddOption ctx SSL_OP_NO_SSLv3
+    SSL.contextSetCiphers ctx rsaCiphers
+    -- We use public key pinning with service providers and want to
+    -- support self-signed certificates as well, hence 'VerifyNone'.
```
MangoIV, line 418: this comment looks like it's out of date?

services/brig/src/Brig/Provider/RPC.hs
```diff
@@ -72,16 +71,17 @@ data ServiceError
...
     rs <- lift $
       wrapHttp $
         recovering x3 httpHandlers $
           const $
             liftIO $
-              withVerifiedSslConnection (verifyFingerprints fprs) man reqBuilder $
-                \req ->
-                  Http.httpLbs req man
+              Http.withConnection req man $
+                \_conn -> Http.httpLbs req man
```
MangoIV, line 84: why can we drop this `withVerifiedSSLConnection`?

MangoIV: aha, because we now do the work in `initExtManager`

libs/ssl-util/src/Ssl/Util.hs
```diff
@@ -169,45 +162,3 @@ verifyRsaFingerprint d = verifyFingerprint $ \pk ->
...
--- [Note: Hostname verification]
--- Ideally, we would like to perform proper hostname verification, which
--- is not done automatically by OpenSSL [1]. However, the necessary APIs
--- are not yet available via HsOpenSSL. Note though that public key pinning
--- is already supposed to thwart attacks based on a lack of or incorrect
--- hostname verification (see [2] for many common attacks and mistakes).
---
--- [1] https://wiki.openssl.org/index.php/Hostname_validation
--- [2] https://www.cs.utexas.edu/~shmat/shmat_ccs12.pdf
-
```
MangoIV, line 182: this is reference elsewhere

services/brig/src/Brig/App.hs
```diff
@@ -400,29 +405,56 @@ initHttp2Manager = do
...
+    SSL.contextSetVerificationMode
+      ctx
+      SSL.VerifyPeer
```
MangoIV, line 421: could this be an issue? that we don't have the VerifyNone anymore here?

integration/integration.cabal
```diff
@@ -51,7 +51,6 @@ common common-all
     MultiWayIf
     NamedFieldPuns
     NoImplicitPrelude
-    NoImportQualifiedPost
```
fisx, line 54: tripped over this as well yesterday.  :)  please don't remove this.  and maybe add a `-- (sic!)` so it won't be almost removed a third time?

https://github.com/wireapp/wire-server/pull/3793

MangoIV: alright :)

integration/integration.cabal
```diff
@@ -70,6 +70,7 @@ common common-all
     UndecidableInstances
     ViewPatterns
 
+-- (sic!)
```
fisx, line 73: hm, ormolu seems to have moved this way out of reach...

maybe you can add a comment at the top of the block mentioning the extension?  not as helpful, but better than this.  :)

services/galley/src/Galley/Env.hs
```diff
@@ -60,45 +66,62 @@ data Env = Env
...
   Ssl.contextAddOption ctx SSL_OP_NO_TLSv1
   Ssl.contextSetCiphers ctx rsaCiphers
+  Ssl.contextSetVerificationMode
+    ctx
+    Ssl.VerifyPeer
+      { vpFailIfNoPeerCert = True,
+        vpClientOnce = False,
+        vpCallback =
+          Just
+            ( \_b ctx509store -> do
```
fisx, line 94: would it make sense to give this function an name and use it in both places?  (i'm assuming it's the same as in `Brig.App`?  if not that would be an even better reason to factor out the *shared* code.  to make it transparent where the differences are.)

fisx (approved):

services/galley/src/Galley/Env.hs
MangoIV: I think this was a wart before, apparently it is quite hard to find a sensible place because these are different packages and the functions differ ever so slightly.

fisx: move it to ssl-utils?

and as i said: if they have subtle differences, then those differences should be factored out to make them visible.

or maybe at least cross-link the two places and explicitly state that we're too lazy to make this pretty?  :)

MangoIV: yeah, if you think `ssl-utils` is suited for this, then perhaps it might be a good idea to move it there if possible.

MangoIV: I think this is addressed.

elland:
🥳

PR #3741, opened 2023-11-23
WPB-5382 - Migrating tests from Cargohold into the new integration test suite.

lepsa:
https://wearezeta.atlassian.net/browse/WPB-5382

Migrating tests from Cargohold into the new integration test suite.

All but 3 tests have been ported over.

2 of the 3 non-ported tests are checking error strings that occur at service start, and are suppressed by `startDynamicBackends`. As these tests don't touch the API in any way, and backend spinup code in `integration` isn't designed for this style of test, I've left them be.
The last test that I haven't ported is due to the error condition that it is testing. This test checks for response codes when an asset ID can be found, but the asset itself cannot be streamed. This test is currently performed with a mock federation server, and to move it into `integration` would require someway of breaking the S3 storage instance for the backend. Something such as creating the asset, then altering it in such a way as to cause the S3 instance, cargohold, or both to fail. As such, I'm leaving the test as it exists in place.

## Checklist


services/cargohold/test/integration/API.hs
```diff
@@ -22,324 +21,40 @@ module API (tests) where
...
-                frComponent = Cargohold,
-                frRPC = "get-asset",
-                frBody = Aeson.encode (GetAsset uid key Nothing)
-              }
-          ]
-
+-- This test doesn't fit in well with the `integration` style of tests. This is
+-- because setting up the asset to have the weird state where the asset key exists,
+-- but somehow the asset itself can't be downloaded is a bit beyond what the API
+-- integration setup allows. So this specific test can stay here for now.
```
pcapriotti, line 57: I've added a small mocking utility to `integration`: https://github.com/wireapp/wire-server/pull/3742/commits/75889a5c30bff06d3885705b8c497f3afe53c80f. Maybe we can extend it to support this use case. In another PR, though.

integration/test/SetupHelpers.hs
```diff
@@ -163,6 +166,15 @@ createMLSOne2OnePartner domain other convDomain = loop
...
 
+-- Copied from `src/CargoHold/API/V3.hs` and inlined to avoid pulling in `types-common`
+randomToken :: HasCallStack => App String
+randomToken = map mkUrlSafe . unpack . B64Url.encode <$> liftIO (getRandomBytes 16)
+  where
+    mkUrlSafe :: Char -> Char
+    mkUrlSafe '/' = '_'
+    mkUrlSafe '+' = '-'
+    mkUrlSafe c = c
+
```
pcapriotti, line 177: You can use the URL-safe version of Base64 directly: https://hoogle.zinfra.io/file/nix/store/cq3gj4j8yicrmwgl5gr2jak51qg6p6il-base64-bytestring-1.2.1.0-doc/share/doc/base64-bytestring-1.2.1.0/html/Data-ByteString-Base64-URL.html.

pcapriotti (approved):
LGTM

services/cargohold/test/integration/API.hs
lepsa: Ah, that is very interesting then. I can look into that, but yes a new ticket once that is merged I think.

integration/test/SetupHelpers.hs
lepsa: Ah nice. Don't know how I missed that one.

PR #3743, opened 2023-11-23
fix: TURN tests failing b/c of fsnotify polling failing to detect changes.

elland:


MangoIV (approved):
❤️ sorry for causing the issues.

mdimjasevic (commented):
For me this breaks lots of tests locally. Now instead of 2 integration tests in Brig failing, I have 38 failing with this message:

```
Exception: user error (Error: couldn't start native file manager: initINotify: resource exhausted (Too many open files))
```
Do I have to tweak an environment variable to fix this?

PR #3749, opened 2023-11-28
WPB-5667: Updating integration tests to better handle comments and haddock.

lepsa:
Using `haskell-src-exts` to pick out all of the tests based on the `test` prefix, and to pair them up with the haddock comments.

This allows for correct handling of block commented tests which would be picked up under the previous system. However this comes with a limitation that we can't use post qualified imports, as `haskell-src-exts` cannot parse them.

Added two commented out tests. These test the build system itself, as neither of them should end up in the generated source for the integration binary.

## Checklist


integration/integration.cabal
```diff
@@ -25,6 +26,7 @@ common common-all
 
   default-extensions:
     NoImplicitPrelude
+    NoImportQualifiedPost
```
lepsa, line 29: This is required so that ormolo doesn't rewrite all of the imports and break haskell-src-exts' parsing.

This comment is moved around in sanitize-pr, and is otherwise very annoynig.

integration/test/Test/AccessUpdate.hs
```diff
@@ -27,7 +27,18 @@ import SetupHelpers
 import Testlib.Prelude
 import Testlib.ResourcePool
 
--- @SF.Federation @SF.Separation @TSFI.RESTfulAPI @S2
+-- These two commented out tests exist to test the Setup.hs code.
+-- Both of these tests should not appear in the output.
+
+-- testBar :: HasCallStack => App ()
```
lepsa, line 33: These could go anywhere, this was the file I had open when I thought of this test.

integration/test/Test/AccessUpdate.hs
```diff
@@ -27,7 +27,18 @@ import SetupHelpers
...
+-- testBar = pure ()
+
+{-
+testBaz :: HasCallStack => App ()
+testBaz = pure ()
+-}
+
+-- | @SF.Federation @SF.Separation @TSFI.RESTfulAPI @S2
```
battermann, line 41: @mythsunwind is your tool to still in place and does it handle the additional pipe here?

integration/Setup.hs
```diff
@@ -52,41 +56,102 @@ collectTests roots = do
...
+              Symbol (_, cs) n -> (n, extractComment <$> cs)
+       in if "test" `isPrefixOf` n'
+            then
+              let (summary, rest) = collectDescription comments
+               in pure (n', summary, rest)
+            else Nothing
+    absolutePath = pkgRoot </> fn
+    -- All of the haskell-src-exts supported extensions that we are using.
+    -- Several that are in the cabal file couldn't be directly copied over,
+    -- but they aren't cauing trouble at the moment.
```
battermann, line 111:
```suggestion
    -- but they aren't causing trouble at the moment.
```

battermann (commented):
LGTM

battermann (approved):

PR #3756, opened 2023-11-29
WPB-4887 increased ingress payload size from 256k to 512k

battermann:
https://wearezeta.atlassian.net/browse/WPB-4887

## Checklist

### DIFF
--- a/charts/nginz/values.yaml
+++ b/charts/nginz/values.yaml
@@ -53,7 +53,7 @@ nginx_conf:
   - /search/top
   - /search/common
 
-  default_client_max_body_size: "256k"
+  default_client_max_body_size: "512k"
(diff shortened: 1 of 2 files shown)


charts/nginz/values.yaml
```diff
@@ -53,7 +53,7 @@ nginx_conf:
   - /search/top
   - /search/common
 
-  default_client_max_body_size: "256k"
+  default_client_max_body_size: "512k"
```
supersven, line 56: A comment how this "magic number" was calculated would be great (though not mandatory.) If I remember this correctly, there was some specific request for a specific amount of group members? :thinking:

supersven (approved):

PR #3762, opened 2023-12-04
Federation API versioning

pcapriotti:
This PR implements the missing pieces to make federation API versioning possible for synchronous endpoints. Version information is communicated via the usual header (`X-Wire-API-Version`) and federator takes care of propagating the header. Federation endpoints can now be given different version "tags" in order to distinguish multiple endpoints with non-overlapping version ranges on the same path.

It also moves the MLS federation endpoints coming from the old `mls` branch to version 1.

https://wearezeta.atlassian.net/browse/WPB-185

## Checklist

### DIFF
--- a/libs/wire-api-federation/src/Wire/API/Federation/API.hs
+++ b/libs/wire-api-federation/src/Wire/API/Federation/API.hs
@@ -48,6 +48,7 @@ import Wire.API.Federation.API.Galley
 import Wire.API.Federation.BackendNotifications
 import Wire.API.Federation.Client
 import Wire.API.Federation.Component
+import Wire.API.Federation.Endpoint
 import Wire.API.Federation.HasNotificationEndpoint
 import Wire.API.MakesFederatedCall
 import Wire.API.Routes.Named
@@ -77,8 +78,13 @@ type HasUnsafeFedEndpoint comp api name = 'Just api ~ LookupEndpoint (FedApi com
 -- 'Wire.API.MakesFederatedCall.exposeAnnotations' for a better understanding
 -- of the information flow here.
 fedClient ::
-  forall (comp :: Component) (name :: Symbol) m (showcomp :: Symbol) api x.
-  (AddAnnotation 'Remote showcomp name x, showcomp ~ ShowComponent comp, HasFedEndpoint comp api name, HasClient m api, m ~ FederatorClient comp) =>
+  forall (comp :: Component) name m (showcomp :: Symbol) api x.
+  ( AddAnnotation 'Remote showcomp (FedPath name) x,
+    showcomp ~ ShowComponent comp,
+    HasFedEndpoint comp api name,
+    HasClient m api,
+    m ~ FederatorClient comp
+  ) =>
   Client m api
 fedClient = clientIn (Proxy @api) (Proxy @m)
 
--- a/libs/wire-api-federation/src/Wire/API/Federation/API/Brig.hs
+++ b/libs/wire-api-federation/src/Wire/API/Federation/API/Brig.hs
@@ -85,6 +85,7 @@ type BrigApi =
     -- (handles can be up to 256 chars currently)
     :<|> FedEndpoint "search-users" SearchRequest SearchResponse
     :<|> FedEndpoint "get-user-clients" GetUserClients (UserMap (Set PubClient))
+    :<|> FedEndpoint (Versioned 'V0 "get-mls-clients") MLSClientsRequestV0 (Set ClientInfo)
     :<|> FedEndpoint "get-mls-clients" MLSClientsRequest (Set ClientInfo)
     :<|> FedEndpoint "send-connection-action" NewConnectionRequest NewConnectionResponse
     :<|> FedEndpoint "claim-key-packages" ClaimKeyPackageRequest (Maybe KeyPackageBundle)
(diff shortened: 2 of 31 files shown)


libs/wire-api-federation/src/Wire/API/Federation/API.hs
```diff
@@ -71,14 +72,19 @@ type HasUnsafeFedEndpoint comp api name = 'Just api ~ LookupEndpoint (FedApi com
 
 -- | Return a client for a named endpoint.
 --
--- This function introduces an 'AddAnnotation' constraint, which is
+-- this function introduces an 'addannotation' constraint, which is
```
mdimjasevic, line 75: Have you accidentally changed the cases in this line?

libs/wire-api-federation/src/Wire/API/Federation/API/Brig.hs
```diff
@@ -85,6 +85,7 @@ type BrigApi =
     -- (handles can be up to 256 chars currently)
     :<|> FedEndpoint "search-users" SearchRequest SearchResponse
     :<|> FedEndpoint "get-user-clients" GetUserClients (UserMap (Set PubClient))
+    :<|> FedEndpoint (Versioned 'V0 "get-mls-clients") MLSClientsRequestV0 (Set ClientInfo)
```
mdimjasevic: Why is it needed in V0? Isn't MLS available only as of V1?

libs/wire-api/src/Wire/API/Routes/Version/Wai.hs
```diff
@@ -86,5 +86,5 @@ removeVersionHeader req =
 addVersionHeader :: Version -> Request -> Request
 addVersionHeader v req =
   req
-    { requestHeaders = (versionHeader, toByteString' (fromEnum v)) : requestHeaders req
+    { requestHeaders = (versionHeader, toByteString' (versionInt v :: Int)) : requestHeaders req
```
mdimjasevic, line 89: Nice! This crossed my mind the other day. Once we start dropping old versions, we won't be able to rely on the `Enum` instance any more.

services/federator/test/unit/Test/Federator/ExternalServer.hs
```diff
@@ -196,7 +198,7 @@ requestGalleySuccess =
           . runInputConst noClientCertSettings
           . runInputConst scaffoldingFederationDomainConfigs
           $ callInward Galley (RPC "get-conversations") aValidDomain (CertHeader cert) request
-      let expectedCall = Call Galley "/federation/get-conversations" "\"foo\"" aValidDomain
+      let expectedCall = Call Galley "/federation/get-conversations" [] "\"foo\"" aValidDomain
```
mdimjasevic, line 201: I'm reviewing commit by commit, but would it make sense to have a test with a non-empty list for headers?

services/brig/test/integration/Util.hs
```diff
@@ -197,7 +198,10 @@ runFedClient (FedClient mgr ep) domain =
       let req' = Servant.defaultMakeClientRequest burl req
        in req'
             { HTTP.requestHeaders =
-                HTTP.requestHeaders req' <> [(originDomainHeaderName, toByteString' originDomain)]
+                HTTP.requestHeaders req'
+                  <> [ (originDomainHeaderName, toByteString' originDomain),
+                       (versionHeader, toByteString' (versionInt (maxBound :: Version)))
```
mdimjasevic, line 203: Will this `maxBound` approach work even once we start dropping old versions?

services/galley/test/integration/TestSetup.hs
```diff
@@ -163,5 +170,8 @@ runFedClient (FedClient mgr ep) domain =
       let req' = Servant.defaultMakeClientRequest burl req
        in req'
             { HTTP.requestHeaders =
-                HTTP.requestHeaders req' <> [(originDomainHeaderName, toByteString' originDomain)]
+                HTTP.requestHeaders req'
+                  <> [ (originDomainHeaderName, toByteString' originDomain),
+                       (versionHeader, toByteString' (versionInt (maxBound :: Version)))
```
mdimjasevic, line 175: Will this `maxBound` approach work even once we start dropping old versions?

mdimjasevic (approved):
Looks good! I left some minor comments.

One thing I don't understand is why you need the `"get-mls-clients"` endpoint in V0 too. Why is it different from other MLS endpoints, which got introduced only in V1?

services/brig/test/integration/Util.hs
pcapriotti: I think so, the assumption is that we are always testing with the maximum available version. If this assumption changes, we'll have to change this code, but I think it's fair.

pcapriotti:
> Looks good! I left some minor comments.
> 
> One thing I don't understand is why you need the `"get-mls-clients"` endpoint in V0 too. Why is it different from other MLS endpoints, which got introduced only in V1?

No particular reason, it's just the way things evolved. The endpoint has been there on `develop` the whole time, but with a different request body. Of course we could have simply deleted it at any point (as we have done for some of the others), but we didn't, so I took it as an opportunity to test the versioning system. It's all pretty vacuous at this point, because the old endpoint is never going to be used in practice, but it's better than nothing.

services/federator/test/unit/Test/Federator/ExternalServer.hs
pcapriotti: Alright, I'll try to add a test with headers.

libs/wire-api-federation/src/Wire/API/Federation/API.hs
pcapriotti: I guess I did. Fixed now.

PR #3764, opened 2023-12-05
[WPB-5603] Fix the team member deleted event reason

mdimjasevic:
As reported by a client developer, the event reason when a team member is deleted was incorrectly "member removed", when it should have been "member deleted" (from a team). This PR fixes it.

Tracked by https://wearezeta.atlassian.net/browse/WPB-5603.

## Checklist

### DIFF
--- a/libs/wire-api/src/Wire/API/Conversation/Action.hs
+++ b/libs/wire-api/src/Wire/API/Conversation/Action.hs
@@ -37,7 +37,6 @@ import Data.Aeson qualified as A
 import Data.Aeson.KeyMap qualified as A
 import Data.Id
 import Data.Kind
-import Data.List.NonEmpty qualified as NonEmptyList
 import Data.OpenApi qualified as S
 import Data.Qualified (Qualified)
 import Data.Schema hiding (tag)
@@ -49,6 +48,7 @@ import Wire.API.Conversation.Action.Tag
 import Wire.API.Conversation.Protocol (ProtocolTag)
 import Wire.API.Conversation.Role
 import Wire.API.Event.Conversation
+import Wire.API.Event.LeaveReason
 import Wire.API.MLS.SubConversation
 import Wire.Arbitrary (Arbitrary (..))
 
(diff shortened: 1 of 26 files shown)


libs/wire-api/test/golden/Test/Wire/API/Golden/Generated/Event_conversation.hs
```diff
@@ -35,6 +35,7 @@ import Wire.API.Conversation.Code
 import Wire.API.Conversation.Role
 import Wire.API.Conversation.Typing
 import Wire.API.Event.Conversation
+import Wire.API.Event.LeaveReason
```
fisx, line 38: also add a manual golden test for `ConversationRemoveMembers`?

libs/wire-api/src/Wire/API/Event/LeaveReason.hs
```diff
@@ -0,0 +1,46 @@
...
+import Data.Schema
+import Imports
+import Wire.Arbitrary
+
+--  | The reason for a member to leave
+--    There are three reasons
+--    - the member has left on their own
+--    - the member was removed from the team
+--    - the member was removed by another member
+data EdMemberLeftReason
```
fisx, line 29: Moving and renaming of this data type could have gone to a separate PR (but I'm very bad at following through with this idea myself :).

libs/wire-api/src/Wire/API/Conversation/Action.hs
```diff
@@ -130,11 +130,7 @@ conversationActionSchema SConversationLeaveTag =
     "ConversationLeave"
     (S.description ?~ "The action of some users leaving a conversation on their own")
     $ pure ()
-conversationActionSchema SConversationRemoveMembersTag =
-  objectWithDocModifier
-    "ConversationRemoveMembers"
-    (S.description ?~ "The action of some users being removed from a conversation")
-    $ field "targets" (nonEmptyArray schema)
+conversationActionSchema SConversationRemoveMembersTag = schema
```
fisx, line 133: nice :)

fisx (approved):
LGTM!  I think concourse is having trouble keeping the federation network running and connected, good luck with that...  :/

pcapriotti (approved):

libs/wire-api/test/golden/Test/Wire/API/Golden/Generated/Event_conversation.hs
mdimjasevic: Thanks for spotting this! I added them in 26b485c7f.

libs/wire-api/src/Wire/API/Event/LeaveReason.hs
mdimjasevic: True. It covers the majority of the files changed in this PR, only for importing the module. Still, I made it as part of this PR given that it's a simple change and a tiny new module.

PR #3765, opened 2023-12-06
WPB-4888: Implement request tracing across federation

battermann:
https://wearezeta.atlassian.net/browse/WPB-4888

## Checklist

### DIFF
--- a/libs/wai-utilities/src/Network/Wai/Utilities/Server.hs
+++ b/libs/wai-utilities/src/Network/Wai/Utilities/Server.hs
@@ -361,7 +361,6 @@ onError g m r k e = liftIO $ do
   when (code >= 500) $
     either Prm.incCounter (counterIncr (path "net.errors")) `mapM_` m
   flushRequestBody r
-  Log.flush g
   k (jsonResponseToWai resp)
 
 -- | Log an 'Error' response for debugging purposes.
--- a/libs/wai-utilities/src/Network/Wai/Utilities/Server.hs
+++ b/libs/wai-utilities/src/Network/Wai/Utilities/Server.hs
@@ -361,6 +361,7 @@ onError g m r k e = liftIO $ do
   when (code >= 500) $
     either Prm.incCounter (counterIncr (path "net.errors")) `mapM_` m
   flushRequestBody r
+  Log.flush g
   k (jsonResponseToWai resp)
 
 -- | Log an 'Error' response for debugging purposes.
--- a/libs/wire-api-federation/src/Wire/API/Federation/BackendNotifications.hs
+++ b/libs/wire-api-federation/src/Wire/API/Federation/BackendNotifications.hs
@@ -33,7 +33,7 @@ data BackendNotification = BackendNotification
     -- pusher. This also makes development less clunky as we don't have to
     -- create a sum type here for all types of notifications that could exist.
     body :: RawJson,
-    requestId :: RequestId
+    requestId :: Maybe RequestId
   }
   deriving (Show, Eq)
 
@@ -54,7 +54,7 @@ instance FromJSON BackendNotification where
       <*> o .: "targetComponent"
       <*> o .: "path"
       <*> (RawJson . TL.encodeUtf8 <$> o .: "body")
-      <*> o .: "requestId"
+      <*> o .:? "requestId"
 
 type BackendNotificationAPI = Capture "name" Text :> ReqBody '[JSON] RawJson :> Post '[JSON] EmptyResponse
 
--- a/libs/wire-api-federation/src/Wire/API/Federation/BackendNotifications.hs
+++ b/libs/wire-api-federation/src/Wire/API/Federation/BackendNotifications.hs
@@ -7,6 +7,7 @@ import Control.Exception
 import Control.Monad.Except
 import Data.Aeson
 import Data.Domain
+import Data.Id (RequestId)
 import Data.Map qualified as Map
 import Data.Text qualified as Text
 import Data.Text.Lazy.Encoding qualified as TL
@@ -31,7 +32,8 @@ data BackendNotification = BackendNotification
     -- this body, which could be very large and completely useless to the
     -- pusher. This also makes development less clunky as we don't have to
     -- create a sum type here for all types of notifications that could exist.
-    body :: RawJson
+    body :: RawJson,
+    requestId :: Maybe RequestId
   }
   deriving (Show, Eq)
 
@@ -41,7 +43,8 @@ instance ToJSON BackendNotification where
       [ "ownDomain" .= notif.ownDomain,
         "targetComponent" .= notif.targetComponent,
         "path" .= notif.path,
-        "body" .= TL.decodeUtf8 notif.body.rawJsonBytes
+        "body" .= TL.decodeUtf8 notif.body.rawJsonBytes,
+        "requestId" .= notif.requestId
       ]
 
 instance FromJSON BackendNotification where
@@ -51,6 +54,7 @@ instance FromJSON BackendNotification where
       <*> o .: "targetComponent"
       <*> o .: "path"
       <*> (RawJson . TL.encodeUtf8 <$> o .: "body")
+      <*> o .:? "requestId"
 
 type BackendNotificationAPI = Capture "name" Text :> ReqBody '[JSON] RawJson :> Post '[JSON] EmptyResponse
 
@@ -70,8 +74,8 @@ sendNotification env component path body =
       runFederatorClient env . void $
         clientIn (Proxy @BackendNotificationAPI) (Proxy @(FederatorClient c)) (withoutFirstSlash path) body
 
-enqueue :: Q.Channel -> Domain -> Domain -> Q.DeliveryMode -> FedQueueClient c a -> IO a
-enqueue channel originDomain targetDomain deliveryMode (FedQueueClient action) =
+enqueue :: Q.Channel -> Maybe RequestId -> Domain -> Domain -> Q.DeliveryMode -> FedQueueClient c a -> IO a
+enqueue channel requestId originDomain targetDomain deliveryMode (FedQueueClient action) =
   runReaderT action FedQueueEnv {..}
 
 routingKey :: Text -> Text
@@ -127,7 +131,8 @@ data FedQueueEnv = FedQueueEnv
   { channel :: Q.Channel,
     originDomain :: Domain,
     targetDomain :: Domain,
-    deliveryMode :: Q.DeliveryMode
+    deliveryMode :: Q.DeliveryMode,
+    requestId :: Maybe RequestId
   }
 
 data EnqueueError = EnqueueError String
--- a/libs/wire-api-federation/src/Wire/API/Federation/BackendNotifications.hs
+++ b/libs/wire-api-federation/src/Wire/API/Federation/BackendNotifications.hs
@@ -33,7 +33,7 @@ data BackendNotification = BackendNotification
     -- pusher. This also makes development less clunky as we don't have to
     -- create a sum type here for all types of notifications that could exist.
     body :: RawJson,
-    requestId :: Maybe RequestId
+    requestId :: RequestId
   }
   deriving (Show, Eq)
 
@@ -54,7 +54,7 @@ instance FromJSON BackendNotification where
       <*> o .: "targetComponent"
       <*> o .: "path"
       <*> (RawJson . TL.encodeUtf8 <$> o .: "body")
-      <*> o .:? "requestId"
+      <*> o .: "requestId"
 
 type BackendNotificationAPI = Capture "name" Text :> ReqBody '[JSON] RawJson :> Post '[JSON] EmptyResponse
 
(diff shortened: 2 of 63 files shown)


libs/wire-api-federation/src/Wire/API/Federation/BackendNotifications.hs
```diff
@@ -31,7 +32,8 @@ data BackendNotification = BackendNotification
     -- this body, which could be very large and completely useless to the
     -- pusher. This also makes development less clunky as we don't have to
     -- create a sum type here for all types of notifications that could exist.
-    body :: RawJson
+    body :: RawJson,
+    requestId :: Maybe RequestId
```
mdimjasevic, line 36: Why is this an optional field?

services/federator/test/unit/Test/Federator/InternalServer.hs
```diff
@@ -148,7 +148,7 @@ federatedRequestFailureAllowList =
         . runInputConst settings
         . runInputConst (FederationDomainConfigs AllowDynamic [FederationDomainConfig (Domain "hello.world") FullSearch FederationRestrictionAllowAll] 10)
         . interpretMetricsEmpty
-        $ callOutward targetDomain Brig (RPC "get-user-by-handle") request
+        $ callOutward Nothing targetDomain Brig (RPC "get-user-by-handle") request
```
mdimjasevic, line 151: Kind of the same as above.

battermann: Also not in scope, or I am missing something.

services/federator/test/unit/Test/Federator/Remote.hs
```diff
@@ -89,7 +89,7 @@ mkTestCall sslCtx hostname port = do
     . discoverLocalhost hostname port
     . assertNoError @DiscoveryFailure
     . interpretRemote
-    $ discoverAndCall (Domain "localhost") Brig "test" [] mempty
+    $ discoverAndCall Nothing (Domain "localhost") Brig "test" [] mempty
```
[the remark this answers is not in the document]
battermann: It's simply of no use in the unit tests. It's for debugging/tracing requests in prod/staging envs.

libs/wire-api-federation/src/Wire/API/Federation/BackendNotifications.hs
mdimjasevic: Where is this actually set? I get that in theory when parsing a header it might be missing, but my understanding is we should make it mandatory, i.e., part of every request. If that is the case, then it'd be nice if this can also be reflected in types.

battermann: It's set by nginx and it's written into the `Env` as part of the servant middleware.

mdimjasevic: OK, I thought so. I'd prefer if we abort in the middleware if there's no request ID, but maybe that's too much.

I'll leave this up to you.

services/federator/test/unit/Test/Federator/Remote.hs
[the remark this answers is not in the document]
mdimjasevic: How about adding a test that asserts there is a request ID?

mdimjasevic (approved):
It looks fine to me, but I'd recommend adding a test that asserts there is a request ID.

libs/wire-api-federation/src/Wire/API/Federation/BackendNotifications.hs
battermann: I hear you. The only purpose of the request ID is debugging. If we made it a required header, we would have to change all of our integration tests. IMO this is not justified.

libs/wai-utilities/src/Network/Wai/Utilities/Server.hs
```diff
@@ -361,6 +361,7 @@ onError g m r k e = liftIO $ do
   when (code >= 500) $
     either Prm.incCounter (counterIncr (path "net.errors")) `mapM_` m
   flushRequestBody r
+  Log.flush g
```
akshaymankar, line 364: Why is this necessary?

libs/wire-api-federation/src/Wire/API/Federation/BackendNotifications.hs
```diff
@@ -51,6 +54,7 @@ instance FromJSON BackendNotification where
       <*> o .: "targetComponent"
       <*> o .: "path"
       <*> (RawJson . TL.encodeUtf8 <$> o .: "body")
+      <*> o .: "requestId"
```
akshaymankar, line 57: This is dangerous. The parser needs to be backwards compatible so any old things which are queued will still get processed after a restart.

libs/wire-api-federation/src/Wire/API/Federation/HasNotificationEndpoint.hs
```diff
@@ -46,10 +47,11 @@ fedNotifToBackendNotif ::
   KnownSymbol (NotificationPath tag) =>
   KnownComponent (NotificationComponent k) =>
   ToJSON (Payload tag) =>
+  RequestId ->
   Domain ->
   Payload tag ->
   BackendNotification
-fedNotifToBackendNotif ownDomain payload =
+fedNotifToBackendNotif mReqId ownDomain payload =
```
akshaymankar, line 54:
```suggestion
fedNotifToBackendNotif reqId ownDomain payload =
```

libs/wire-api-federation/src/Wire/API/Federation/HasNotificationEndpoint.hs
```diff
@@ -60,5 +62,6 @@ fedNotifToBackendNotif ownDomain payload =
         { ownDomain = ownDomain,
           targetComponent = componentVal @(NotificationComponent k),
           path = path,
-          body = body
+          body = body,
+          requestId = mReqId
```
akshaymankar, line 66:
```suggestion
          requestId = reqId
```

services/brig/src/Brig/Run.hs
```diff
@@ -156,10 +160,15 @@ type ServantCombinedAPI =
...
-  let reqid = maybe def RequestId $ lookupRequestId req
-  mkapp reqid req cont
+lookupRequestIdMiddleware :: Logger -> (RequestId -> Wai.Application) -> Wai.Application
+lookupRequestIdMiddleware logger mkapp req cont = do
+  case lookupRequestId req of
+    Just rid -> do
+      mkapp (RequestId rid) req cont
+    Nothing -> do
+      localRid <- RequestId . cs . UUID.toText <$> UUID.nextRandom
+      Log.info logger $ "request-id" .= localRid ~~ "request" .= (show req) ~~ msg (val "generated a new request id for local request")
```
akshaymankar, line 170: I think it would be nicer if log basics of a request like method and path instead of logging the whole request. It could contain sensitive info in headers. This way the logs also become searchable by those things.

services/cannon/src/Cannon/Types.hs
```diff
@@ -120,7 +119,7 @@ runCannon' :: Env -> Cannon a -> IO a
 runCannon' e c = runReaderT (unCannon c) e
 
 lookupReqId :: Request -> RequestId
-lookupReqId = maybe def RequestId . lookup requestIdName . requestHeaders
+lookupReqId = RequestId . fromMaybe "N/A" . lookup requestIdName . requestHeaders
```
akshaymankar, line 122: Why not generate it like in brig?

services/cargohold/src/CargoHold/Run.hs
```diff
@@ -93,6 +97,15 @@ mkApp o = Codensity $ \k ->
...
         )
         r
+        cont
+
+    lookupReqId :: Logger -> Wai.Request -> IO RequestId
+    lookupReqId l r = case lookup requestIdName $ Wai.requestHeaders r of
+      Just rid -> pure $ RequestId rid
+      Nothing -> do
+        localRid <- RequestId . cs . UUID.toText <$> UUID.nextRandom
+        Log.info l $ "request-id" .= localRid ~~ "request" .= (show r) ~~ msg (val "generated a new request id for local request")
```
akshaymankar, line 107: Same as brig, probably better to log individual items from the request rather than the whole request.

services/federator/src/Federator/ExternalServer.hs
```diff
@@ -114,8 +116,8 @@ server ::
 server mgr intPort interpreter =
   API
     { status = Health.status mgr "internal server" intPort,
-      externalRequest = \component rpc remoteDomain remoteCert ->
-        Tagged $ \req respond -> runCodensity (interpreter (callInward component rpc remoteDomain remoteCert req)) respond
+      externalRequest = \component rpc mReqId remoteDomain remoteCert ->
+        Tagged $ \req respond -> runCodensity (interpreter (callInward component rpc (fromMaybe (RequestId "N/A") mReqId) remoteDomain remoteCert req)) respond
```
akshaymankar, line 120: We should generate a request id here if we don't find one.

services/federator/src/Federator/InternalServer.hs
```diff
@@ -89,8 +91,8 @@ server ::
 server mgr extPort interpreter =
   API
     { status = Health.status mgr "external server" extPort,
-      internalRequest = \remoteDomain component rpc ->
-        Tagged $ \req respond -> runCodensity (interpreter (callOutward remoteDomain component rpc req)) respond
+      internalRequest = \mReqId remoteDomain component rpc ->
+        Tagged $ \req respond -> runCodensity (interpreter (callOutward (fromMaybe (RequestId "N/A") mReqId) remoteDomain component rpc req)) respond
```
akshaymankar, line 95: Same as external server, we should generate a request id here.

services/galley/src/Galley/Run.hs
```diff
@@ -106,24 +108,33 @@ mkApp opts =
...
+        cont
 
-    lookupReqId :: Request -> RequestId
-    lookupReqId = maybe def RequestId . lookup requestIdName . requestHeaders
+    lookupReqId :: Logger -> Request -> IO RequestId
+    lookupReqId l r = case lookup requestIdName $ requestHeaders r of
+      Just rid -> pure $ RequestId rid
+      Nothing -> do
+        localRid <- RequestId . cs . UUID.toByteString <$> UUID.nextRandom
+        Log.info l $ "request-id" .= localRid ~~ "request" .= (show r) ~~ msg (val "generated a new request id for local request")
```
akshaymankar, line 136: Same as above.

services/gundeck/src/Gundeck/Monad.hs
```diff
@@ -182,7 +181,7 @@ runDirect e m =
             )
 
 lookupReqId :: Request -> RequestId
-lookupReqId = maybe def RequestId . lookup requestIdName . requestHeaders
+lookupReqId = RequestId . fromMaybe "N/A" . lookup requestIdName . requestHeaders
```
akshaymankar, line 184: Same as cannon and federator, we should generate a req id here.

services/proxy/src/Proxy/Proxy.hs
```diff
@@ -61,4 +60,4 @@ reqIdMsg = ("request" .=) . unRequestId
 {-# INLINE reqIdMsg #-}
 
 lookupReqId :: Request -> RequestId
-lookupReqId = maybe def RequestId . lookup requestIdName . requestHeaders
+lookupReqId = RequestId . fromMaybe "N/A" . lookup requestIdName . requestHeaders
```
akshaymankar, line 63: We should generate a req id here.

tools/stern/src/Stern/App.hs
```diff
@@ -71,7 +70,7 @@ newEnv :: Opts -> IO Env
 newEnv o = do
   mt <- Metrics.metrics
   l <- Log.mkLogger (O.logLevel o) (O.logNetStrings o) (O.logFormat o)
-  Env (mkRequest $ O.brig o) (mkRequest $ O.galley o) (mkRequest $ O.gundeck o) (mkRequest $ O.ibis o) (mkRequest $ O.galeb o) l mt def
+  Env (mkRequest $ O.brig o) (mkRequest $ O.galley o) (mkRequest $ O.gundeck o) (mkRequest $ O.ibis o) (mkRequest $ O.galeb o) l mt (RequestId "N/A")
```
akshaymankar, line 73: I don't see this generated anywhere.

akshaymankar (changes requested):
Comments inline. Two things I would consider blockers for merging:
1. The BackendNotification parser needs to be backwards compatible at least for a couple of releases.
2.  We shouldn't log the whole request, picking fields to log would be best.

libs/wai-utilities/src/Network/Wai/Utilities/Server.hs
battermann: It's not, it sneaked in there somehow... good catch!

akshaymankar (approved):

PR #3766, opened 2023-12-07
[WPB-5810] Fix the service provider endpoints that return no body

mdimjasevic:
There are a couple of endpoints that have no response body and that were written using Servant's `Verb` mechanism that matches on the `Accept` header. The PR fixes this.

Tracked by https://wearezeta.atlassian.net/browse/WPB-5810.

## Checklist

### DIFF
--- a/changelog.d/3-bug-fixes/WPB-5810
+++ b/changelog.d/3-bug-fixes/WPB-5810
@@ -1,0 +1,1 @@
+Do not match on the `Accept` header for service provider endpoints with no response body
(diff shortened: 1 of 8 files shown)


pcapriotti (approved):
LGTM. Have you checked the behaviour after the patch?

mdimjasevic:
> LGTM. Have you checked the behaviour after the patch?

I wrote a test and `MultiVerb` seems to match on the `Accept` header. It's a bit worrisome we haven't noticed this so far, meaning we had no single test for this.

For now I just use `text/plain` and `application/json` as valid `Accept` header values. It's a workaround. I'd be happy to fix the behavior of `MultiVerb` in a follow-up PR.

PR #3771, opened 2023-12-12
Hack /client/:cid/nonce to also return a Content-Length Header

akshaymankar:
This will get around https://github.com/yesodweb/wai/issues/956 But there are other endpoints so, perhaps we shouldn't do this.

## Checklist


akshaymankar:
superseded by #3775

PR #3774, opened 2023-12-12
[WPB-5389] Guard user connection requests by team-level federation settings

mdimjasevic:
This PR introduces guarding of user connection requests by team-level federation settings.

Tracked by https://wearezeta.atlassian.net/browse/WPB-5389.

## Checklist

The following pre-existing 13 integration tests in Brig need to be dropped (as they've been failing due to application code changes in this PR) and rewritten in the `integration` package:
- [x] federation/POST /federation/on-user-deleted-connections : 200
- [x] post /users/connections-status/v2 : All connections
- [x] Remote connections: limits
- [x] Remote connections: send then cancel
- [x] Remote connections: block, remote cancels, then accept
- [x] Remote connections: block then accept
- [x] Remote connections: ignore, remote cancels, then accept
- [x] Remote connections: ignore then accept
- [x] Remote connections: mutual Connect - remote action then local action
- [x] Remote connections: mutual Connect - local action then remote action
- [x] post /connections (bad user)
- [x] connection/post /connections
- [x] post /clients/:cid/access-token/success

### DIFF
--- a/integration/test/SetupHelpers.hs
+++ b/integration/test/SetupHelpers.hs
@@ -240,7 +240,7 @@ getOne2OneConversation user1 user2 cnvState = do
         -- The conversation type 2 is for 1-to-1 conversations. Type 3 is for
         -- the connection conversation, which is the state of the conversation
         -- before the connection is fully established.
-        t <- (== (toConvType cnvState)) <$> (c %. "type" & asInt)
+        t <- (== toConvType cnvState) <$> (c %. "type" & asInt)
         others <- c %. "members.others" & asList
         qIds <- for others (%. "qualified_id")
         pure $ qIds == users && t


integration/test/Test/Federation.hs
```diff
@@ -128,7 +128,7 @@ testNotificationsForOfflineBackends = do
 
       -- FUTUREWORK: Uncomment after fixing this bug: https://wearezeta.atlassian.net/browse/WPB-3664
       -- void $ awaitNotification downUser1 downClient1 (Just newMsgNotif) 1 isOtherUser2LeaveUpConvNotif
-      -- void $ awaitNotification otherUser otherClient (Just newMsgNotif) 1 isDelUserLeaveDownConvNotif
+      void $ awaitNotification otherUser otherClient (Just newMsgNotif) isDelUserLeaveDownConvNotif
```
mdimjasevic, line 131: Why this change? The test is now failing. Also, it feels like this change doesn't belong to this PR.

MangoIV: oh yeah; perhaps I can should revert it, I just looked at the PR  above and it is merged, so the test should pass and I thought I'd just give it a shot and if it doesn't pass, then I comment it again...

MangoIV:
https://wearezeta.atlassian.net/browse/WPB-3664

MangoIV:
https://github.com/wireapp/wire-server/pull/3591

This PR should have fixed this line (see above)

```haskell
void $ awaitNotification otherUser otherClient (Just newMsgNotif) isDelUserLeaveDownConvNotif
```

Perhaps we should have a look at it again? 

cc @pcapriotti

integration/test/Test/Connection.hs
```diff
@@ -0,0 +1,403 @@
...
+    deleteUser charlie
+
+    -- charlie is on their local backend, so asking should be instant
+    getConnection charlie alice `bindResponse` \resp ->
+      resp.status `shouldMatchInt` 404
+
+    -- for alice, charlie is on the remote backend, so the status change
+    -- may not be instant
+    getConnection alice charlie `waitForResponse` \resp ->
+      resp.status `shouldMatchInt` 404
```
mdimjasevic, line 94: I suppose this works too, but we used a different approach so far. In particular, Alice would wait for a notification on her web socket that Charlie got deleted. Once she has that, she is sure that `getConnection alice charlie` returns a 404.

MangoIV: does this work if the connection is pending?

mdimjasevic: I don't think it does as there are no notifications for this state.

I'd say it's fine to do it this way; I was just saying so far we have done it differently, but that's no argument against doing something new like you did here.

integration/test/Test/Connection.hs
```diff
@@ -0,0 +1,403 @@
...
+
+  void $ putConnection alice bob "cancelled" >>= getBody 200
+  assertConnectionStatus alice bob "cancelled"
+
+testConnectionLimits :: HasCallStack => App ()
+testConnectionLimits = do
+  let connectionLimit = 16
+
+  alice <- randomUser OwnDomain def
+  [charlie1, charlie2, charlie3, charlie4] <- replicateM 4 do
```
battermann, line 245: We've got a helper to create multiple users: `createUsers $ replicate 4 OtherDomain`

integration/test/Test/Connection.hs
```diff
@@ -0,0 +1,403 @@
...
+    resp.status `shouldMatchInt` 201
+
+testNonFederatingRemoteTeam :: HasCallStack => App ()
+testNonFederatingRemoteTeam =
+  withFederatingBackendsAllowDynamic $ \(domainA, domainB, _) -> do
+    sequence_
+      [ createFedConn domainA (FedConn domainB defSearchPolicy Nothing),
+        createFedConn domainB (FedConn domainA defSearchPolicy Nothing)
+      ]
+    void $ updateFedConn domainA domainB (FedConn domainB defSearchPolicy $ Just [])
```
battermann, line 317: Why not create the policy the way we want it right away instead of first creating it and then updating it?

integration/test/Test/Connection.hs
```diff
@@ -0,0 +1,403 @@
...
+  withFederatingBackendsAllowDynamic $ \(domainA, domainB, _) -> do
+    sequence_
+      [ createFedConn domainA (FedConn domainB defSearchPolicy Nothing),
+        createFedConn domainB (FedConn domainA defSearchPolicy Nothing)
+      ]
+    alice <- randomUser domainA def
+    bob <- randomUser domainB def {API.BrigInternal.team = True}
+
+    -- Alice's backend federates with Bob's team
+    void $ updateFedConn domainA domainB (FedConn domainB defSearchPolicy $ Just [])
```
battermann, line 337: As above I would prefer to create the fedration policy the way we want right away.

libs/wire-api/src/Wire/API/Routes/Public/Brig.hs
```diff
@@ -971,6 +971,7 @@ type ConnectionAPI =
     :<|> Named
            "create-connection"
            ( Summary "Create a connection to another user"
+               :> MakesFederatedCall 'Brig "get-users-by-ids"
```
battermann, line 974: Should `CanThrow TeamsNotFederating` be added here?

services/brig/test/integration/API/User/Connection.hs
```diff
@@ -62,7 +60,7 @@ tests ::
   FedClient 'Galley ->
   DB.ClientState ->
   TestTree
-tests cl _at opts p b _c g fedBrigClient fedGalleyClient db =
+tests cl _at p b _c g fedBrigClient _fedGalleyClient db =
```
battermann, line 63: remove `_fedGalleyClient`?

services/brig/test/integration/Federation/End2end.hs
```diff
@@ -592,7 +592,15 @@ claimRemoteKeyPackages brig1 brig2 = do
     Set.map (\e -> (e.user, e.client)) bundle.entries
       @?= Set.fromList [(bob, c) | c <- bobClients]
 
-testRemoteTypingIndicator :: Brig -> Brig -> Galley -> Galley -> Cannon -> Cannon -> Http ()
+testRemoteTypingIndicator ::
```
battermann, line 595: This seems unrelated to this PR.

battermann (approved):
- Awesome that some tests from the old integration suite were ported!
- I would suggest to add tests for non-team users, too
- Looks good otherwise, only very marginal comments

integration/test/SetupHelpers.hs
```diff
@@ -194,3 +194,54 @@ withFederatingBackendsAllowDynamic k = do
...
+-- | Fetch the one-to-one conversation between the two users that is in one of
+-- two possible states.
+getOne2OneConversation :: HasCallStack => Value -> Value -> One2OneConvState -> App Value
+getOne2OneConversation user1 user2 cnvState = do
+  l <- getAllConvs user1
+  let isWith users c = do
+        -- The conversation type 2 is for 1-to-1 conversations. Type 3 is for
+        -- the connection conversation, which is the state of the conversation
+        -- before the connection is fully established.
+        t <- (==) <$> (c %. "type" & asInt) <*> pure (toConvType cnvState)
```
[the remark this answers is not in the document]
mdimjasevic: I copy-pasted this from the `Conversation` module where I had it.

To get the `type` field out of `c`, you need to run in the `App` monad that we have for tests.

integration/test/Test/Connection.hs
mdimjasevic: I forgot the details, but the type wrangling seemed a bit challenging for me to do it right away, hence I did it this way. A concrete suggestion on how to change this is more than welcome so I finally learn to do it nicely!

libs/wire-api/src/Wire/API/Routes/Public/Brig.hs
mdimjasevic: I don't think so. In Galley, we specify each individual error that a handler can throw and these are statically checked. In Brig, we have an ADT that a handler can throw, and these by the end of execution get converted to `StdError`, which is the code-label-message triple error response.

services/brig/test/integration/API/User/Connection.hs
mdimjasevic: I thought of that, but then left it (in case someone needs it). Now that I think about it again, we won't ever update these tests, so no one will re-introduce the need for that federation client; hence I'll remove it.

services/brig/test/integration/Federation/End2end.hs
mdimjasevic: This is a left-over from debugging this test as it was initially failing.

I tend to split the screen vertically and then these long signatures don't fit into a half of the screen, and it gets annoying having to move the cursor/slider to see what are the rest of the arguments.

If you insist, I'll remove it from this PR, but I'd argue the change makes the code more ergonomic.

integration/test/Test/Connection.hs
MangoIV: Iirc this does more than I'm allowed to do here. (To make the test replicate the old behaviour)

integration/test/SetupHelpers.hs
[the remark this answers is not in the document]
battermann: Yeah, I saw later that it was just moved here. Anyway, something like this `t <- (== toConvType cnvState) <$> (c %. "type" & asInt)` should work as well.

libs/wire-api/src/Wire/API/Routes/Public/Brig.hs
battermann: I think it is useful for clients to know which status and labels to expect so that they can handle them.

services/brig/test/integration/Federation/End2end.hs
battermann: All good.

libs/wire-api/src/Wire/API/Routes/Public/Brig.hs
mdimjasevic: I agree, but with Brig we're not there (yet). Brig is in a half-finished state when it comes to migrating to Polysemy. With Polysemy we have the ability to list only specific errors an endpoint can return and have them checked by the compiler (as done in Galley). If I manually add this annotation to the endpoint, it will not be requested by the endpoint handler, and then we're back to manually annotating Swagger with errors an endpoint can return; such documentation is out of sync and out of date in no time.

Or am I getting this wrong?

MangoIV: I think the ADT is specification enough, everything else has too much overhead.

MangoIV: I'm not even aware; is that ADT somewhere added to Swagger?

integration/test/SetupHelpers.hs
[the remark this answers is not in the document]
mdimjasevic: I implemented it as you suggested. It's easier to read.

libs/wire-api/src/Wire/API/Routes/Public/Brig.hs
mdimjasevic: No. These ADTs are not reflected in Swagger. In this case, we have the `ConnectionError` error type, but before the endpoint returns, this error that is passed around within the `ConnectionM r` monad (just an `ExceptT`), is eventually converted to an `StdError`, e.g.:

```haskell
API.createConnection lself conn target !>> connError
```
where:
```haskell
connError :: ConnectionError -> Error
```

integration/test/Test/Connection.hs
battermann: I meant change it to just this:

```
    sequence_
      [ createFedConn domainA (FedConn domainB defSearchPolicy $ Just []),
        createFedConn domainB (FedConn domainA defSearchPolicy Nothing)
      ]
```

PR #3777, opened 2023-12-13
.envrc: don't enable flakes extra-experimental-feature

flokli:
There's nothing in this repository using flakes, and setting this causes my system to just block entirely:

```
❯ cd dev/wire/wire-server
direnv: error /home/flokli/dev/wire/wire-server/.envrc is blocked. Run `direnv allow` to approve its content
❯ direnv allow
direnv: loading ~/dev/wire/wire-server/.envrc
🔧 Building environment
warning: unknown experimental feature 'flakes'
warning: unknown experimental feature 'flakes'
⏱ 4sdirenv: ([/nix/store/0lf6mkfaqx4zmsni9yys650qx573zfcw-direnv-2.33.0/bin/direnv export zsh]) is taking a while to execute. Use CTRL-C to give up.
⏱ 8s^C
^C^X^C^C^C
```

## Checklist

### DIFF
--- a/.envrc
+++ b/.envrc
@@ -13,7 +13,7 @@ store_paths=$(echo "$nix_files" | xargs nix-store --add ./nix)
 layout_dir=$(direnv_layout_dir)
 env_dir=./.env
 
-export NIX_CONFIG='extra-experimental-features = nix-command flakes'
+export NIX_CONFIG='extra-experimental-features = nix-command'
 
 [[ -d "$layout_dir" ]] || mkdir -p "$layout_dir"
 
--- a/hack/bin/bombon.hs
+++ b/hack/bin/bombon.hs
@@ -1,4 +1,4 @@
-#!/usr/bin/env -S nix -Lv run github:wireapp/ghc-flakr/ecb1f45f1549e06c92d71164e305ce501eb0e36e
+#!/usr/bin/env -S nix --extra-experimental-features flakes -Lv run github:wireapp/ghc-flakr/ecb1f45f1549e06c92d71164e305ce501eb0e36e
 {-# LANGUAGE BlockArguments #-}
 {-# LANGUAGE ImportQualifiedPost #-}
 {-# LANGUAGE OverloadedStrings #-}


MangoIV:
This was requested by @elland to make using hs-run easier (it uses flakes).

MangoIV:
I guess if you get stuck, it would be better to remove it from the envrc altogether and load it only on demand?

flokli:
There's a single occurence in the `hack/bin/bombon.hs` shell script, which uses `nix run`:

```
1:#!/usr/bin/env -S nix -Lv run github:wireapp/ghc-flakr/ecb1f45f1549e06c92d71164e305ce501eb0e36e
```

@elland could you (and everyone invoking `hack/bin/bombon.hs` enable this on your system-wide config?

MangoIV:
I think it's easier to not have it automatically load in `wire-server` and instead just use it explicitly in a `devShell` or in the shebang and then pass the argument along, if you don't have the feature enabled?

flokli:
Good point, lemme just update this PR to change the shebang.

MangoIV:
@flokli do you have some special setup that flakes feature makes you get stuck? I don't think it should have any effect as long as it's not used...

flokli:
I assume Nix behaves a bit stupidly if it encounters an experimental feature it doesn't know about, or nom does, but yes, I do have some patches, and my Nix doesn't know about the `flakes` experimental feature.

However, I also think we should only be adding stuff we actually need. I moved the invocation to the shebang in `hack/bin/bombon.hs` (which I can't test by myself).

I'd argue this script could also be rewritten in bash, saving us a bunch of startup to evaluate nixpkgs, compile the binary and haskell interpreter. The only thing it does is invoking `nix build -f nix wireServer.allLocalPackagesBom` (which could also be rewritten to `nix-build nix -A wireServer.allLocalPackagesBom`), and trhen invoke the `gh` binary - but I lack context on this script and it's a bit too far out of scope for this PR, which does the minimal thing - unblocking my shell while hopefully not breaking someone others' workflow ;-)

MangoIV:
yeah, the script is a proof of concept.

flokli:
closing in favor of #3778.

PR #3782, opened 2023-12-15
WPB-5312

battermann:
https://wearezeta.atlassian.net/browse/WPB-5312

## Checklist

### DIFF
--- a/changelog.d/5-internal/WPB-5312
+++ b/changelog.d/5-internal/WPB-5312
@@ -1,0 +1,1 @@
+Improve error logs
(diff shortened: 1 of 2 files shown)


mdimjasevic (approved):
It looks good.

Have you tried it out? I hope this doesn't result in potentially incomprehensible logs.

battermann:
> It looks good.
> 
> Have you tried it out? I hope this doesn't result in potentially incomprehensible logs.

Here is an example:

```
[federator@example.com] E, request=N/A, code=500, label=server-error, "Server Error. bääääääääääääääääääääääm CallStack (from HasCallStack):   error, called at src/Federator
/InternalServer.hs:111:8 in federator-1.0.0-inplace:Federator.InternalServer"  
```

smatting (approved):
Nice! Now, how can we solve the problem that we never have stack traces? In "integration" we solved the problem by attaching the stack trace to the exception object, iirc

battermann:
> Nice! Now, how can we solve the problem that we never have stack traces? In "integration" we solved the problem by attaching the stack trace to the exception object, iirc

I am not sure, did you checkout the example above? It has a stack trace.

smatting:
@battermann ah, didn't see. cool!

PR #3786, opened 2023-12-18
Introduce NotificationSubsystem to push notifications

akshaymankar:
https://wearezeta.atlassian.net/browse/WPB-5985

Remaning work for after holidays:

- [x] Use the subsystem is brig
- [x] Ponder about whether is makes sense to send notifications to bots from the notification subsystem
    Probably it does, but if we do it now this PR will never end. Perhaps this is better dealt with when we're explicitly looking into bots.
- [x] Figure out how types in gundeck-types can be more aligned with the subsystem
   There are two types which cause some contention: `Push` and `Recipient`, there are two versions of these types in `Wire.NotificationSubsystem` and `Gundeck.Types.V2`. This is most likely an artefact of migration from V1 to V2. I think this can be done but will require too many changes. We should consider doing this some other time. 

## Checklist


services/galley/src/Galley/API/Push.hs
```diff
@@ -80,25 +78,25 @@ newMessagePush botMap mconn mm userOrBots event =
...
-  push (toPush mp)
+  push $ maybeToList $ toPush mp
   for_ mqcnv $ \qcnv ->
     if tDomain loc /= qDomain qcnv
       then unless (null botMembers) $ do
         warn $ Log.msg ("Ignoring messages for local bots in a remote conversation" :: ByteString) . Log.field "conversation" (show qcnv)
       else deliverAndDeleteAsync (qUnqualified qcnv) (map (,event) botMembers)
 
-toPush :: MessagePush -> Maybe Push
+toPush :: MessagePush -> Maybe PushToUser
```
mdimjasevic, line 96: Does this need a new name given the return type change?

services/galley/src/Galley/API/Teams/Features.hs
```diff
@@ -200,16 +200,18 @@ pushFeatureConfigEvent ::
...
+      P.warn $
+        Log.field "action" (Log.val "Features.pushFeatureConfigEvent")
+          . Log.field "feature" (Log.val (toByteString' . Event._eventFeatureName $ event))
+          . Log.field "team" (Log.val (cs . show $ tid))
+          . Log.msg @Text "Fanout limit exceeded. Events will not be sent."
+    else do
+      let recipients = membersToRecipients Nothing (memList ^. teamMembers)
+      push $
+        maybeToList $
+          (newPush Nothing (toJSONObject event) recipients)
```
mdimjasevic, line 214: These two don't have the same behavior. In the old ones we'd always have a push. Why the change?

mdimjasevic (commented):
Overall this looks good, but a couple of places left me wondering if the change keeps the behavior; I left such comments inlined.

services/galley/src/Galley/API/Push.hs
akshaymankar: The `PushToUser` type has a bad name. I want to try and take out polymorphism from the `PushTo` type, I think its not needed. Once that happens `PushToUser` can just be `Push`. I want to unify all of these `Push` types. I think they create more confusion than anything.

services/galley/src/Galley/API/Teams/Features.hs
akshaymankar: In the old code we were sending and argument of type `ListType` to `push`. In the logic for `push` if it saw it was a `ListTruncated` it used to just not push anything. I found this really odd and just removed this `ListType` and made this place make a decision about whether or not things should be pushed.
It turned out that this was the only place from where we could send a `ListTrucated`. 

I think this is a better approach because it is not a job of the notification subsystem to say that if a notification cannot be sent to all the people, noone gets it. It should be decided case by case.

mdimjasevic (commented):
Overall, I think this is a good approach.

Perhaps I'm failing to see the obvious thing, but conceptually to me this looks exactly the same as the initial efforts to Polysemize Spar, Galley and Brig. Likely the devil is in the details. Therefore, I'd appreciate if you could summarize how this differs from the initial efforts, because the reason we're working on these subsystems is because we're not happy with the outcome of the initial efforts. So it'd be good if we have a reasoning framework that would assure us we're doing a good and different thing as we go, and that we will end up in some place better than before.

MangoIV:
@mdimjasevic: Correct me if I'm wrong but I think the key difference is that we try to focus on avoiding scope creep to make interpreting effects in its own scope possible, e.g. if we need data from another service, we don't just access its database but request that data from its subsystem directly.

akshaymankar:
I think the big difference between the new approach and the old one is that we're going to implement business logic in the interpreters of the effects. So far we've only implemented wrappers around simple functions which accessed HTTP or Cassandra or some other thing. By doing this the functions will still have a few `Member` constraints but hopefully they will be something closely related to the domain of Wire rather than the thing they wrap. For instance, I think it is much nicer to see `NotificationSubsystem` than `GundeckAccess`. With gundeck it is actually slightly harder to appreciate this because we've been strict with that service to only keep notification specific things in it. But perhaps `BrigAccess` will become a few things like `UserSubsystem`, `ClientSubsystem`, `ConnectionSubsystem`, etc. 

This will also help hide some information about how exactly things are implemented. For instance in this PR you can see the place which calls `pushNotificationsSlowly` doesn't also get a `Member Delay r` constraint added to it.  As a result we should also see a significant drop in the `Member (Error ...) r` constraints because they will now be part of the interpretation. 

This has a drawback that we won't be able to track all errors from the place they are thrown all the way up to the API unless the errors are also made part of the Subsystem interface. But I don't consider this a drawback, as some errors which we consider significant enough, we will make them part of the interface and so they will show up in the API, other errors will be as if they were thrown in IO, they will produce whatever HTTP status code with whatever message, they would just not be bloating up the constraints on the functions.

@mdimjasevic I have written about the differences a bit here: https://wearezeta.atlassian.net/browse/WPB-4907. Please let me know if I missed something.

mdimjasevic:
> I think the big difference between the new approach and the old one is that we're going to implement business logic in the interpreters of the effects.
...
> @mdimjasevic I have written about the differences a bit here: https://wearezeta.atlassian.net/browse/WPB-4907. Please let me know if I missed something.

Ok, that makes sense. Thanks for the explanation!

I'm eager to see how this evolves!

libs/wire-subsystems/src/Wire/NotificationSubsystem/Internal.hs
```diff
@@ -0,0 +1,45 @@
...
+    _pushRecipients :: NonEmpty Recipient,
+    pushJson :: Object,
+    _pushApsData :: Maybe ApsData
+  }
+  deriving stock (Eq, Generic, Show)
+  deriving (Arbitrary) via GenericUniform Push
+
+makeLenses ''Push
+
+data NotificationSubsystem m a where
```
akshaymankar, line 38: TODO: Write about this subsystem so a new developer can understand why it exists and what it is supposed to do.

services/brig/src/Brig/User/EJPD.hs
```diff
@@ -60,7 +67,7 @@ ejpdRequest includeContacts (EJPDRequestBody handles) = do
       let uid = userId target
 
       ptoks <-
-        PushTok.tokenText . view PushTok.token <$$> wrapHttp (Intra.lookupPushToken uid)
+        PushTok.tokenText . view PushTok.token <$$> liftSem (NotificationSubsystem.getPushTokens uid)
```
battermann, line 70: I personally find this a bit long. Should we agree on a short form like we do for effects with `E`? Maybe just `Sub`, `Notifs`, `Notifications`, etc. ...?

services/galley/src/Galley/API/Action.hs
```diff
@@ -1111,9 +1112,10 @@ pushTypingIndicatorEvents ::
...
 pushTypingIndicatorEvents qusr tEvent users mcon qcnv ts = do
   let e = Event qcnv Nothing qusr tEvent (EdTyping ts)
-  for_ (newPushLocal ListComplete (qUnqualified qusr) (ConvEvent e) (userRecipient <$> users)) $ \p ->
-    push1 $
-      p
-        & pushConn .~ mcon
-        & pushRoute .~ RouteDirect
-        & pushTransient .~ True
+  for_ (newPushLocal (qUnqualified qusr) (toJSONObject e) (userRecipient <$> users)) $ \p ->
+    pushNotifications
```
battermann, line 1116: Should we qualify this everywhere by convention, maybe?
Maybe such a convention can help when reading the code to see right away if it's a local function or a subsystem API call.

services/galley/src/Galley/API/Teams/Features.hs
```diff
@@ -199,16 +199,18 @@ pushFeatureConfigEvent ::
...
-  for_
-    (newPush (memList ^. teamMemberListType) Nothing (FeatureConfigEvent event) recipients)
-    push1
+  if ((memList ^. teamMemberListType) == ListTruncated)
+    then do
+      P.warn $
+        Log.field "action" (Log.val "Features.pushFeatureConfigEvent")
+          . Log.field "feature" (Log.val (toByteString' . Event._eventFeatureName $ event))
+          . Log.field "team" (Log.val (cs . show $ tid))
+          . Log.msg @Text "Fanout limit exceeded. Events will not be sent."
```
battermann, line 208: :+1:

battermann (commented):
This is really tough to review. Maybe it's a good idea to pair on these subsystem refactorings in the future, to compensate for the shortcomings of having to review such a large diff.

In general, I like the approach very much.

Would it be possible to reflect the separation of types, API, and implementation in the structure of wire-subsystems a bit better? E.g. something along these lines (not literally, but just to get an idea):

```
├── effects
│   ├── GundeckAccess.hs
│   └── RPC.hs
├── interpreters
│   └── MockNotifications.hs
├── subsystems
│   ├── Conversations.hs
│   └── Notifications.hs
└── types
    └── Push.hs
```

services/galley/src/Galley/API/Action.hs
akshaymankar: Its a bit too much to follow conventions like this. I would like it if we could just not have to qualify these. I will see if I can get rid of the collisions in other places.

services/brig/src/Brig/User/EJPD.hs
akshaymankar: `Notifs` and `Notifications` would collide with those types, `E` and `Sub` doesn't convey much. I will try to get rid of these qualified imports and use them as a last resort.

akshaymankar:
> This is really tough to review. Maybe it's a good idea to pair on these subsystem refactorings in the future, to compensate for the shortcomings of having to review such a large diff.

Fair point. Let's do this from next one onwards.



> Would it be possible to reflect the separation of types, API, and implementation in the structure of wire-subsystems a bit better? E.g. something along these lines (not literally, but just to get an idea):
> 
> ```
> ├── effects
> │   ├── GundeckAccess.hs
> │   └── RPC.hs
> ├── interpreters
> │   └── MockNotifications.hs
> ├── subsystems
> │   ├── Conversations.hs
> │   └── Notifications.hs
> └── types
>     └── Push.hs
> ```

Sure, I will give it a shot.

services/brig/src/Brig/API/Client.hs
```diff
@@ -462,9 +475,17 @@ pubClient c =
       pubClientClass = clientClass c
     }
 
-legalHoldClientRequested :: UserId -> LegalHoldClientRequest -> (AppT r) ()
+legalHoldClientRequested ::
+  ( Member (Embed HttpClientIO) r,
+    Member NotificationSubsystem r,
+    Member Async r,
+    Member TinyLog r
+  ) =>
```
MangoIV, line 483: will these `Async` constraints stay? I think it's a bit odd that we capture the semantics of this handler being asynchronous in the top level constraint. Ideally this should be abstracted away as especially access to `Async` is something that should be thoroughly tested, no?

services/brig/src/Brig/API/Internal.hs
```diff
@@ -613,21 +696,36 @@ deleteFromPhonePrefixH prefix = lift $ NoContent <$ API.phonePrefixDelete prefix
...
-      lift $ wrapHttpClient $ Intra.onUserEvent uid Nothing (UserUpdated ((emptyUserUpdatedData uid) {eupSSOId = Just ssoid}))
+      lift $ liftSem $ Intra.onUserEvent uid Nothing (UserUpdated ((emptyUserUpdatedData uid) {eupSSOId = Just ssoid}))
       pure UpdateSSOIdSuccess
     else pure UpdateSSOIdNotFound
 
-deleteSSOIdH :: UserId -> (Handler r) UpdateSSOIdResponse
+deleteSSOIdH ::
+  ( Member (Embed HttpClientIO) r,
+    Member NotificationSubsystem r,
+    Member Async r,
```
MangoIV, line 719: there are so many `Async` constraints that I do no know what thing actually uses `Async`, perhaps we should just test that thing and run the Async where it's happening?

libs/wire-subsystems/test/unit/Wire/NotificationSubsystem/InterpreterSpec.hs
```diff
@@ -0,0 +1,280 @@
...
+spec :: Spec
+spec = describe "NotificationSubsystem.Interpreter" do
+  describe "pushImpl" do
+    it "chunks and sends all notifications" do
+      let mockConfig =
+            NotificationSubsystemConfig
+              { fanoutLimit = toRange $ Proxy @30,
+                chunkSize = 12,
+                slowPushDelay = 0
+              }
```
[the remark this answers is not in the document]
akshaymankar: I made them `Natural` is there any incorrect input you can imagine?

libs/wire-subsystems/test/unit/Wire/NotificationSubsystem/InterpreterSpec.hs
```diff
@@ -0,0 +1,280 @@
...
+    GundeckAPIAccess.GetPushTokens uid -> do
+      liftIO $ expectationFailure $ "Unexpected call to GundeckAPI: GetPushTokens " <> show uid
+      error "impossible"
+
+waitUntilPushes :: IORef [a] -> Int -> IO [a]
+waitUntilPushes pushesRef n = do
+  ps <- readIORef pushesRef
+  if length ps >= n
+    then pure ps
+    else threadDelay 1000 >> waitUntilPushes pushesRef n
```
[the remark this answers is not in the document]
akshaymankar: This thread delay ensures that this function yields to other work as it is really just waiting for other threads to do work.

services/brig/src/Brig/API/Client.hs
akshaymankar: Good observation. We added another action to the effect which sends notifs async so we could get rid of all this noise.

services/brig/src/Brig/API/User.hs
```diff
@@ -849,57 +936,51 @@ revokeIdentity key = do
...
+    Member TinyLog r
   ) =>
   List1 UserId ->
   AccountStatus ->
-  ExceptT AccountStatusError m ()
+  ExceptT AccountStatusError (AppT r) ()
 changeAccountStatus usrs status = do
   ev <- mkUserEvent usrs status
-  lift $ mapConcurrently_ (update ev) usrs
+  lift $ liftSem $ unsafePooledMapConcurrentlyN_ 16 (update ev) usrs
```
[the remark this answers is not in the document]
MangoIV: `mapConcurrently_` (the non-pooled version) can be captured with the `Async` effect, it's something like `sequenceConcurrently . fmap` 
https://hackage.haskell.org/package/polysemy-1.9.1.3/docs/Polysemy-Async.html#v:sequenceConcurrently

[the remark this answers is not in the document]
akshaymankar: I find it a bit odd to do this without pooling because this can be called when all users of a team are to be suspended. So, we shouldn't be doing this anyway.

libs/wire-subsystems/src/Wire/NotificationSubsystem/Interpreter.hs
```diff
@@ -0,0 +1,170 @@
...
+  CleanupUser uid -> GundeckAPIAccess.userDeleted uid
+  UnregisterPushClient uid cid -> GundeckAPIAccess.unregisterPushClient uid cid
+  GetPushTokens uid -> GundeckAPIAccess.getPushTokens uid
+
+data NotificationSubsystemConfig = NotificationSubsystemConfig
+  { fanoutLimit :: Range 1 HardTruncationLimit Int32,
+    chunkSize :: Natural,
+    -- | Microseconds
+    slowPushDelay :: Natural,
+    requestId :: RequestId
```
battermann, line 54:
```suggestion
    slowPushDelayMs :: Natural,
    requestId :: RequestId
```

battermann (approved):
LGTM

libs/wire-subsystems/src/Wire/NotificationSubsystem/Interpreter.hs
akshaymankar: Using `DiffTime` everywhere. There was actually a bug here, we were defaulting to 20s instead of 20ms in galley. Hopefully this helps us solve these bugs.

libs/wire-subsystems/test/unit/Wire/NotificationSubsystem/InterpreterSpec.hs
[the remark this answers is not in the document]
akshaymankar: Made it `DiffTime` now. Do you have a test case in mind?

PR #3792, opened 2024-01-04
remove geoip (dead) code

stefanwire:
## Checklist


charts/brig/templates/deployment.yaml
```diff
@@ -42,50 +42,11 @@ spec:
...
-                key: licenseKey
-          - name: GEOIPUPDATE_EDITION_IDS
-            valueFrom:
-              secretKeyRef:
-                name: brig-geoip
-                key: editionIds
-          - name: GEOIPUPDATE_FREQUENCY
-            value: "0" # Setting this to 0 makes the script only run geoipupdate once.
-      {{- end }}
-        {{- if eq (include "includeSecurityContext" .) "true" }}
```
fisx, line 84: weird that the second `if` is not included in the body of the first one, but removing it fixes that i guess...

charts/integration/templates/integration-integration.yaml
```diff
@@ -83,7 +83,7 @@ spec:
   restartPolicy: Never
 
   initContainers:
-  # FUTUREWORK: reproduce brig's initContainers (for geoip)
+  # FUTUREWORK: reproduce brig's initContainers (despite geoip removed?)
```
fisx, line 86:
```suggestion
```

i think leaving this in causes more confusion than it is worth?  your call, though.

fisx (approved):

charts/brig/templates/deployment.yaml
stefanwire: Was wondering about the same question. As far as I see, it would be inconsequent to leave this `if` remain, since it seems to be there for geoip only.

PR #3793, opened 2024-01-05
YesImportQualifiedPost in /integration

fisx:
https://wearezeta.atlassian.net/browse/WPB-5803

We should either support this extension in both /integration and everywhere else, or nowhere.  I'm very much in favor of "everywhere".  Note that the extension allows for the old *and* the new syntax.

(I'm also curious: why was it disabled?)

## Checklist

### DIFF
--- a/integration/integration.cabal
+++ b/integration/integration.cabal
@@ -51,7 +51,6 @@ common common-all
     MultiWayIf
     NamedFieldPuns
     NoImplicitPrelude
-    NoImportQualifiedPost
     OverloadedLabels
     OverloadedRecordDot
     PackageImports


smatting (changes requested):
git blame leads me to https://github.com/wireapp/wire-server/pull/3749, specifically this comment https://github.com/wireapp/wire-server/pull/3749/files#diff-02ec43b3da1fa3f480f864ab97c43eeefe39e6e0487f245a1a5672f0282e10e7R112

fisx:
ah, ok.  makes kinda sense...

sorry for the noise.

PR #3797, opened 2024-01-10
[WPB-5883] Feature flag for a limited event fanout

mdimjasevic:
To maintain compatibility with old clients that do not implement the limited even fanout feature when a team member is deleted, this PR puts the limited fanout behind a feature flag. The flag is disabled by default to maintain compatibility with old clients that rely on the full fanout.

Tracked by https://wearezeta.atlassian.net/browse/WPB-5883.

## Checklist

- [X] Have two integration tests (the enabled case already exists), each having a different flag value (enabled vs. disabled), that assert on events that a team owner and other team members get, but also non-team users get.
  - [X] For the disabled case, create another Git branch that is based on the parent of the commit corresponding to PR #3434. Write an integration test that asserts on the events when a team member is deleted. Once the test is
    done, port it to the current PR branch. That way we make sure the behavior is as before. Otherwise it would be hard to know what are the expected semantics given that PR #3434 made some changes without writing tests.
- [X] Have a simple test just to make sure the feature flag's endpoints for getting and setting work.
- [x] Fix old Galley integration test "send billing events to owners even in large teams"

### DIFF
--- a/integration/test/Test/Conversation.hs
+++ b/integration/test/Test/Conversation.hs
@@ -720,28 +720,6 @@ testDeleteTeamMemberLimitedEventFanout = do
       ms <- resp.json %. "members" & asList
       forM ms $ (%. "user")
 
-testLeaveConversationSuccess :: HasCallStack => App ()
-testLeaveConversationSuccess = do
-  [alice, bob, chad, dee] <- createUsers [OwnDomain, OwnDomain, OtherDomain, OtherDomain]
-  [aClient, bClient] <- forM [alice, bob] $ \user ->
-    objId $ bindResponse (addClient user def) $ getJSON 201
-  startDynamicBackends [def] $ \[dynDomain] -> do
-    eve <- randomUser dynDomain def
-    eClient <- objId $ bindResponse (addClient eve def) $ getJSON 201
-    forM_ [bob, chad, dee, eve] $ connectTwoUsers alice
-    conv <-
-      postConversation
-        alice
-        ( defProteus
-            { qualifiedUsers = [bob, chad, dee, eve]
-            }
-        )
-        >>= getJSON 201
-    void $ removeMember chad conv chad >>= getBody 200
-    assertLeaveNotification chad conv alice aClient chad
-    assertLeaveNotification chad conv bob bClient chad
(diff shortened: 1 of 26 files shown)


services/galley/src/Galley/API/Teams.hs
```diff
@@ -978,12 +986,29 @@ deleteTeamMember' lusr zcon tid remove mBody = do
...
+        getFeatureStatus @LimitedEventFanoutConfig DontDoAuth tid
+          >>= ( \case
+                  FeatureStatusEnabled -> E.getBillingTeamMembers tid
+                  FeatureStatusDisabled -> do
+                    let filterFromMembers list =
+                          view userId <$> filter (`hasPermission` SetBilling) (list ^. teamMembers)
+                    filterFromMembers <$> getTeamMembersForFanout tid
+              )
+            . wsStatus
+      Journal.teamUpdate tid sizeAfterDelete $ filter (/= remove) toNotify
```
mdimjasevic, line 999: Now that I look at it, maybe this wasn't necessary at all. I'll run the tests in a version where this is left out.

mdimjasevic: Resolved in a commit "Revert the change to the billing team update notification".

integration/test/Test/Conversation.hs
```diff
@@ -660,36 +660,101 @@ testDeleteTeamConversationWithUnreachableRemoteMembers = do
...
+  where
+    assertConvLeaveNotif :: MakesValue leaverId => WebSocket -> leaverId -> App ()
+    assertConvLeaveNotif ws leaverId = do
+      n <- awaitMatch isConvLeaveNotif ws
+      nPayload n %. "data.qualified_user_ids.0" `shouldMatch` leaverId
+      nPayload n %. "data.reason" `shouldMatch` "user-deleted"
+    getMembers tid usr = bindResponse (getTeamMembers usr tid) $ \resp -> do
+      resp.status `shouldMatchInt` 200
+      ms <- resp.json %. "members" & asList
+      forM ms $ (%. "user")
```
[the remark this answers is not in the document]
mdimjasevic: Done in 31afbb5e9.

stefanwire (approved):

PR #3802, opened 2024-01-11
WPB-6099 Bump the version of rusty-jwt-tools in wire server

battermann:
https://wearezeta.atlassian.net/browse/WPB-6099
(it also resolves https://wearezeta.atlassian.net/browse/WPB-5804)

It can be merged once the associated CC release is rolled out to the clients. When in doubt, ask @beltram.

## Checklist


services/brig/src/Brig/Effects/JwtTools.hs
```diff
@@ -26,6 +26,10 @@ data JwtTools m a where
     Proof ->
     -- | The qualified client ID associated with the currently logged on user
     ClientIdentity ->
+    -- | The user's handle
+    Text ->
```
elland, line 30: I thought we had a Handle type.

services/brig/src/Brig/Effects/JwtTools.hs
```diff
@@ -46,14 +50,16 @@ makeSem ''JwtTools
 
 interpretJwtTools :: Member (Embed IO) r => Sem (JwtTools ': r) a -> Sem r a
 interpretJwtTools = interpret $ \case
-  GenerateDPoPAccessToken pr ci n uri method skew ex now pem ->
+  GenerateDPoPAccessToken pr ci h t n uri method skew ex now pem ->
```
elland, line 53: nit: good opportunity to use readable names for this, what do you think?

elland (approved):
I think the main concern has been raised by @akshaymankar already in chat. From my POV this seems like a by design break, but it's good to be sure. 👍

PR #3806, opened 2024-01-15
Revert brig memory setup back to 512mb

elland:
https://wearezeta.atlassian.net/browse/WPB-5676

After some local and CI based testing, it seems that whatever was causing the OOM issue for bring after upgrading GHC to 9.4 is no longer present. Considering the sheer scope of changes since then and the dependencies we updated or swapped since then, and as discussed within the team, I'm setting this up to revert the memory quota back to the previous value, so we can see how it behaves on staging.


mdimjasevic (approved):
Maybe a changelog entry and reference to the PR that bumped the memory?

PR #3811, opened 2024-01-17
WPB-6162 update x 509 verification with new client identity format

battermann:
## Checklist


stefanwire (approved):

rohan-wire:
Why was the ! percent-encoded? `!` was selected specifically because it does not need to be escaped in the userpart of the URL according to the RFC. 


(userinfo is unreserved, sub-delims, or percent encoded. sub-delims includes `!`)
https://datatracker.ietf.org/doc/html/rfc3986#appendix-A

PR #3815, opened 2024-01-17
WPB-6012 create new API version v6

battermann:
https://wearezeta.atlassian.net/browse/WPB-6012

## Checklist


stefanwire (approved):

integration/test/Testlib/Env.hs
```diff
@@ -99,7 +99,7 @@ mkGlobalEnv cfgFile = do
         gDomain1 = intConfig.backendOne.originDomain,
         gDomain2 = intConfig.backendTwo.originDomain,
         gDynamicDomains = (.domain) <$> Map.elems intConfig.dynamicBackends,
-        gDefaultAPIVersion = 5,
+        gDefaultAPIVersion = 6,
```
mdimjasevic, line 102: A good catch! How did you find it? By running tests or by some other means? Let's add this to the version bump checklist!

services/brig/src/Brig/API/Public.hs
```diff
@@ -239,6 +240,11 @@ internalEndpointsSwaggerDocsAPI ::
   PortNumber ->
   S.OpenApi ->
   Servant.Server (VersionedSwaggerDocsAPIBase service)
+internalEndpointsSwaggerDocsAPI service examplePort swagger (Just (VersionNumber V6)) =
+  swaggerSchemaUIServer $
+    swagger
+      & adjustSwaggerForInternalEndpoint service examplePort
+      & cleanupSwagger
```
mdimjasevic, line 247: I'm looking at the definition for V0..V3 and they are all `emptySwagger`. Is that so because we had no Servant-based Swagger docs for those versions?

mdimjasevic (approved):
Looks good. Let's extend the documentation for this process in a follow-up PR!

integration/test/Testlib/Env.hs
battermann: I looked at the previous PR for finalizing V4.

services/brig/src/Brig/API/Public.hs
battermann: As just discussedd, the internal API is not versioned, and therefore only the latest stable V5 and the development version V6 point to the internal API docs. 0-5 should point to empty swagger.

PR #3825, opened 2024-01-22
Improve usage of http-manager (fixes for fingerprint verification)

MangoIV:
reuse manager whenever possible but drop it in legalhold scenarios without using a fork

the crux of the `http-client` change can probably be broken down to these lines: 
```haskell
  manF <- view extGetManager
  man <- liftIO $ manF fprs
```
because we only initialize the manager as soon as we have the fingerprints (per request), we don't reuse connections, ever but create a new manager per request. 

there's a fundamental issue where receiving fingerprints per request modifies the ssl context, which is per manager. the previous approach avoided this, by verifying directly in the request and not by letting `http-client` run a callback.

## Checklist

### DIFF
--- a/services/galley/src/Galley/Env.hs
+++ b/services/galley/src/Galley/Env.hs
@@ -26,6 +26,7 @@ import Data.Id
 import Data.Metrics.Middleware
 import Data.Misc (Fingerprint (..), HttpsUrl, Rsa)
 import Data.Range
+import Debug.Trace
 import Galley.Aws qualified as Aws
 import Galley.Options
 import Galley.Options qualified as O
@@ -70,6 +71,7 @@ makeLenses ''Env
 -- TODO: somewhat duplicates Brig.App.initExtGetManager
 initExtEnv :: IORef [Fingerprint Rsa] -> IO Manager
 initExtEnv fingerprints = do
+  traceM "initExtEnv"
   ctx <- Ssl.context
   Ssl.contextAddOption ctx SSL_OP_NO_SSLv2
   Ssl.contextAddOption ctx SSL_OP_NO_SSLv3
--- a/services/galley/src/Galley/Env.hs
+++ b/services/galley/src/Galley/Env.hs
@@ -54,11 +54,11 @@ data Env = Env
     _applog :: Logger,
     _manager :: Manager,
     _http2Manager :: Http2Manager,
-    _federator :: Maybe Endpoint, -- FUTUREWORK: should we use a better type here? E.g. to avoid fresh connections all the time?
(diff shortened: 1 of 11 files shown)


MangoIV:
my reasoning for this to be correct: 
it would be a major bug in `http-client` and `HsOpenSSL` if it wasn't.

MangoIV:
before our update of http-client, we had a function with the following type signature:
makeVerifiedRequestWithManager :: Http.Manager -> ([Fingerprint Rsa] -> SSL.SSL -> IO ()) -> Fingerprint Rsa -> HttpsUrl -> (Http.Request -> Http.Request) -> App (Http.Response LC8.ByteString)
it receives a callback that modifies an SSL that is run per request the SSL, however, is part of the state of the Manager, meaning that there can be race conditions of the kind of
- make request with fingerprint list a
- make request with fingerprint list b concurrently
- before a can verify, b replaces the fingerprints
- verification fails, although the fingerprints were correct
(or, even worse, it succeeds even though they were correct)
our update actually (accidentally) avoided this, by passing the callback to creation of the manager, which is more correct but makes it so that we create a new manager per request.
My question is hence as follows:
How can we avoid updating the context in that way wihile also being able to set fingerprints on per request basis. Afaict there's two solutions:
- attach the fingerprints only to the manager and change the way the fingerprints are handed from handing to request to handing to manager
- expand the list of fingerprints, don't replace them

services/galley/src/Galley/External/LegalHoldService.hs
```diff
@@ -162,7 +162,7 @@ makeLegalHoldServiceRequest tid reqBuilder = do
           legalHoldServiceFingerprint = fpr,
           legalHoldServiceToken = serviceToken
         } = lhSettings
-  makeVerifiedRequest fpr baseUrl $ mkReqBuilder serviceToken
+  makeVerifiedRequestFreshManager fpr baseUrl $ mkReqBuilder serviceToken
   where
```
MangoIV, line 166: this is probably wrong.

MangoIV:
It is graen!

elland (approved):
Looks like this addresses the chicken-egg problem we overlooked with the verification callback 👍, CI seems consistently green now, which is a good sign.

services/galley/src/Galley/Env.hs
```diff
@@ -54,11 +54,11 @@ data Env = Env
     _applog :: Logger,
     _manager :: Manager,
     _http2Manager :: Http2Manager,
-    _federator :: Maybe Endpoint, -- FUTUREWORK: should we use a better type here? E.g. to avoid fresh connections all the time?
+    _federator :: Maybe Endpoint, -- FUTUREWORK: should we use a better type ----- LegalHold.testLHClaimKeys01 FAIL (34.01 s) -----here? E.g. to avoid fresh connections all the time?
```
MangoIV, line 57: lmao what

PR #3830, opened 2024-01-24
Clean up LH tests

fisx:
https://wearezeta.atlassian.net/browse/WPB-6254

## Checklist

### DIFF
--- a/integration/test/Testlib/PTest.hs
+++ b/integration/test/Testlib/PTest.hs
@@ -24,3 +24,37 @@ instance HasTests x => HasTests (Ciphersuite -> x) where
       [ mkTests m (n <> "[suite=" <> suite.code <> "]") s f (x suite)
         | suite <- allCiphersuites
       ]
+
+instance HasTests x => HasTests (Bool -> x) where
+  mkTests m n s f x =
+    mconcat
+      [ mkTests m (n <> "[" <> show b1 <> "]") s f (x b1)
+        | b1 <- [minBound ..]
+      ]
+
+instance HasTests x => HasTests ((Bool, Bool) -> x) where
+  mkTests m n s f x =
+    mconcat
+      [ mkTests m (n <> "[" <> show b1 <> "," <> show b2 <> "]") s f (x (b1, b2))
+        | b1 <- [minBound ..],
+          b2 <- [minBound ..]
+      ]
+
+instance HasTests x => HasTests ((Bool, Bool, Bool) -> x) where
+  mkTests m n s f x =
+    mconcat
+      [ mkTests m (n <> "[" <> show b1 <> "," <> show b2 <> "," <> show b3 <> "]") s f (x (b1, b2, b3))
+        | b1 <- [minBound ..],
+          b2 <- [minBound ..],
+          b3 <- [minBound ..]
+      ]
+
+instance HasTests x => HasTests ((Bool, Bool, Bool, Bool) -> x) where
+  mkTests m n s f x =
+    mconcat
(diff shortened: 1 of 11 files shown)


fisx:
@smatting i might break the spirit of the package by introducing API.BrigCommon, but i did my best not to.  could you take a look?

integration/test/Testlib/PTest.hs
```diff
@@ -24,3 +24,37 @@ instance HasTests x => HasTests (Ciphersuite -> x) where
...
+        | b1 <- [minBound ..]
+      ]
+
+instance HasTests x => HasTests ((Bool, Bool) -> x) where
+  mkTests m n s f x =
+    mconcat
+      [ mkTests m (n <> "[" <> show b1 <> "," <> show b2 <> "]") s f (x (b1, b2))
+        | b1 <- [minBound ..],
+          b2 <- [minBound ..]
+      ]
```
smatting, line 41: maybe this can be made even more general, extending to everything bounded and enumerable and showable? And maybe for the tests something descriptive instead of bools avoiding boolean blindness.

I have no opinion on the module structure

smatting: Actually looking at the HastTest, once instance for single enumerable thing would suffice. The recursive instances would give an instance for functions taking multiple arguments instead of a tuple, which is also nicer

fisx: i love the idea, but ghc begs to differ:

```
No instance for (Enum ((Bool, Bool), (Bool, Bool)))
```

and this is after i messed around with newtypes to resolve overlapping instance issues:

```
newtype WithArg arg x = WithArg (arg -> x)

instance (HasTests x, Enum arg, Bounded arg, Show arg) => HasTests (WithArg arg x) where
  mkTests m n s f (WithArg x) =
    mconcat
      [ mkTests m (n <> "[" <> show arg <> "]") s f (x arg)
        | arg <- [minBound ..]
      ]
```

so, i say it's too awkward to pursue really?

fisx: like this?  https://github.com/wireapp/wire-server/pull/3830/commits/3cc3a60e0ea814abff8a8599b37bc11b38eb9b89

not sure if this is better or worse than before.  before was...  simpler.  :)

smatting: I love it!;) Also the implementation is just 10 lines now, right? There is leftover code from the previous deisgn.

smatting (approved):
LGTM, except for the leftover code

integration/test/Testlib/PTest.hs
MangoIV: there's a class that does exactly this btw https://hackage.haskell.org/package/universe-base-1.1.3.1/docs/Data-Universe-Class.html#t:Universe

MangoIV:
```hs
module HasTests where

import Data.Universe.Class
import Data.Foldable

class HasTests a where
  execute :: a -> IO ()

instance HasTests (IO ()) where
  execute = id

instance (Universe a, HasTests b) => HasTests (a -> b) where
  execute f = traverse_ (execute . f) (universe @a)

instance {-# OVERLAPPING #-} Universe a => HasTests (a -> IO ()) where
  execute f = traverse_ f (universe @a)

aTest :: ((Bool, Bool), (Bool, Bool)) -> IO ()
aTest = print 

bTest :: Either ((Bool, Bool), (Either (Bool, Bool) Bool, Either Bool Bool)) Bool -> IO ()
bTest = print
```

fisx: `Universe` is maybe a bit more concise, yes.  Is there any other benefit I'm missing?

PR #3831, opened 2024-01-24
[WPB-183] Version federation API queue notifications

mdimjasevic:
Tracked by https://wearezeta.atlassian.net/browse/WPB-183.

## Checklist


libs/wire-api-federation/src/Wire/API/Federation/Version.hs
```diff
@@ -66,6 +81,107 @@ instance ToSchema VersionInfo where
...
+-- the upper boundary.
+versionRangeToBoundaries :: VersionRange -> (Version, Version)
+versionRangeToBoundaries AllVersions = (minBound @Version, maxBound @Version)
+versionRangeToBoundaries (FromVersion fv) = (fv, maxBound @Version)
+versionRangeToBoundaries (UntilVersion uv) = (minBound @Version, pred uv)
+versionRangeToBoundaries (FromUntilVersion fv uv) = (fv, pred uv)
+
+-- | Checks if a version is within a given version range.
+isWithinRange :: Version -> VersionRange -> Bool
+isWithinRange v vr =
```
fisx, line 163:
```suggestion
inVersionRange v vr =
```

libs/wire-api-federation/src/Wire/API/Federation/Version.hs
```diff
@@ -66,6 +81,107 @@ instance ToSchema VersionInfo where
...
+  let (lo, hi) = versionRangeToBoundaries vr
+   in lo <= v && v <= hi
```
fisx, line 165: this means that either `from` or `until` (i keep forgetting which is inclusive) change their semantics to being inclusive.  not sure how i would resolve this.  probably go with the practice of the client api versioning combinators we all are already used to?

libs/wire-api-federation/src/Wire/API/Federation/Version.hs
```diff
@@ -66,6 +81,107 @@ instance ToSchema VersionInfo where
...
+
+-- | For a version range of a local backend and for a set of versions that a
+-- remote backend supports, compute the newest version supported by both. The
+-- remote versions are given as integers as the range of versions supported by
+-- the remote backend can include a version unknown to the local backend. If
+-- there is no version in common, the return value is 'Nothing'.
+versionInCommon :: VersionRange -> Set Int -> Maybe Version
+versionInCommon localVersions remoteVersions =
```
fisx, line 173:
```suggestion
latestCommonVersion localVersions remoteVersions =
```

libs/wire-api-federation/src/Wire/API/Federation/Version.hs
```diff
@@ -66,6 +81,107 @@ instance ToSchema VersionInfo where
...
+  foldl' f Nothing (Set.map inRange remoteVersions)
+  where
+    inRange :: Int -> Maybe Version
+    inRange i = do
+      v <- intToVersion i
+      guard (v `isWithinRange` localVersions) $> v
+    f :: Maybe Version -> Maybe Version -> Maybe Version
```
fisx, line 180:
```suggestion

    f :: Maybe Version -> Maybe Version -> Maybe Version
```

libs/wire-api-federation/src/Wire/API/Federation/Version.hs
```diff
@@ -66,6 +81,107 @@ instance ToSchema VersionInfo where
...
+   in lo <= v && v <= hi
+
+-- | For a version range of a local backend and for a set of versions that a
+-- remote backend supports, compute the newest version supported by both. The
+-- remote versions are given as integers as the range of versions supported by
+-- the remote backend can include a version unknown to the local backend. If
+-- there is no version in common, the return value is 'Nothing'.
+versionInCommon :: VersionRange -> Set Int -> Maybe Version
+versionInCommon localVersions remoteVersions =
+  foldl' f Nothing (Set.map inRange remoteVersions)
```
fisx, line 174: You could also compute a `Set Int` from localVersions and then take the maximum of the intersection of the two sets, but I don't know if that's easier.

services/background-worker/src/Wire/BackendNotificationPusher.hs
```diff
@@ -70,16 +72,31 @@ pushNotification runningFlag targetDomain (msg, envelope) = do
...
+        -- we are trying to cleanup workers for a shutdown, this will call will block
         -- we are trying to cleanup workers for a shutdown, this will call will block
+        -- we are trying to cleanup workers for a shutdown, this will call will block
+        -- and prevent the message from being sent out as we are tearing down resources.
+        -- and prevent the message from being sent out as we are tearing down resources.
         -- and prevent the message from being sent out as we are tearing down resources.
         -- This removes one way that a message might be delivered twice.
+        -- This removes one way that a message might be delivered twice.
+        -- This removes one way that a message might be delivered twice.
         UnliftIO.bracket_ (takeMVar runningFlag) (putMVar runningFlag ()) go
```
fisx, line 92: how did that happen?  :)

services/background-worker/src/Wire/BackendNotificationPusher.hs
```diff
@@ -70,16 +72,31 @@ pushNotification runningFlag targetDomain (msg, envelope) = do
...
   where
     go :: AppT IO ()
-    go = case A.eitherDecode @BackendNotification (Q.msgBody msg) of
+    -- go = case A.eitherDecode @BackendNotification (Q.msgBody msg) of
+    go = case A.eitherDecode @(PayloadBundle _) (Q.msgBody msg) of
       Left e -> do
+        -- TODO(md): try to parse this as an old 'BackendNotification' to
+        -- maintain backward compatibility.
```
fisx, line 99: you can do it like this:

```suggestion
    go = case (A.eitherDecode @(PayloadBundle _) (Q.msgBody msg),
               A.eitherDecode @BackendNotification (Q.msgBody msg)) of
      (Right legacy, _) -> ...
      (Left e, Right legacy) -> ...
      ...
```

This way you don't run the legacy parser unless bundle-parsing failed, and it's pretty concise and readable.

services/galley/src/Galley/API/Action.hs
```diff
@@ -897,7 +897,9 @@ notifyConversationAction tag quid notifyOrigDomain con lconv targets action = do
         -- because quid's backend will update local state and notify its users
         -- itself using the ConversationUpdate returned by this function
         if notifyOrigDomain || tDomain ruids /= qDomain quid
-          then fedQueueClient @'OnConversationUpdatedTag update $> Nothing
+          then do
+            env <- ask
+            fedQueueClientBundle (toBundle @'OnConversationUpdatedTag env update) $> Nothing
```
fisx, line 902: `OnConversationUpdatedTag` is the fed end-point in its most recent (locally available) version, right?  And for older versions of the same end-point, you need different tags?  How does `toBundle` know where to find those older tags?

If I'm right and this is a problem, maybe using a [`Dynamic`](https://hoogle.zinfra.io/file/nix/store/5x7yh702gmyiwhbiq1qbybknyk1fl4ys-ghc-9.4.8-doc/share/doc/ghc/html/libraries/base-4.17.2.1/Data-Dynamic.html) list in the bundle is part of the answer?  But then you still need to figure out how to guess the tags that are in there.

If this makes no sense, I'll think about it more tomorrow.  :)

fisx (commented):
Looks good so far!  I nit-picked around a little.

libs/wire-api-federation/src/Wire/API/Federation/Version.hs
mdimjasevic: Well spottted! I'm making the upper bound exclusive to make it aligned with the `Until` combinator.

libs/wire-api-federation/src/Wire/API/Federation/Version.hs
mdimjasevic: Thanks! Adopting the change in an upcoming commit.

services/background-worker/src/Wire/BackendNotificationPusher.hs
mdimjasevic: Implemented it, though in a bit different way.

libs/wire-api-federation/src/Wire/API/Federation/Version.hs
```diff
@@ -83,108 +90,57 @@ versionInfo = VersionInfo (toList supportedVersions)
...
+
+untilVersions :: Version -> VersionRange
+untilVersions v = VersionRange Nothing (Just v)
+
+enumVersionRange :: VersionRange -> Set Version
+enumVersionRange =
+  Set.fromList . \case
+    (VersionRange Nothing Nothing) -> [minBound ..]
+    (VersionRange (Just l) Nothing) -> [l ..]
+    (VersionRange Nothing (Just u)) -> [minBound .. u]
```
mdimjasevic, line 133: This includes the upper limit. Is that your intent?

fisx: nope, good catch, thanks!  :)

libs/wire-api-federation/src/Wire/API/Federation/Version.hs
```diff
@@ -83,108 +90,57 @@ versionInfo = VersionInfo (toList supportedVersions)
...
-versionRangeToBoundaries (UntilVersion uv) = (minBound @Version, pred uv)
-versionRangeToBoundaries (FromUntilVersion fv uv) = (fv, pred uv)
-
--- | Checks if a version is within a given version range.
-inVersionRange :: Version -> VersionRange -> Bool
-inVersionRange v vr =
-  let (lo, hi) = versionRangeToBoundaries vr
-   in lo <= v && v < hi
+allVersions :: VersionRange
+allVersions = VersionRange Nothing Nothing
```
mdimjasevic, line 120: You haven't solved that problem you've brought up, namely that you can define one value in two ways, right? This is the same semantically as `VersionRange (Just V0) Nothing`.

fisx: also good catch!  [fixed](https://github.com/wireapp/wire-server/pull/3831/commits/601d298b00a987caaf77f9cd1dabcab3aca0612b)!

libs/wire-api-federation/src/Wire/API/Federation/BackendNotifications.hs
```diff
@@ -127,6 +190,12 @@ ensureQueue chan queue = do
 newtype FedQueueClient c a = FedQueueClient (ReaderT FedQueueEnv IO a)
   deriving (Functor, Applicative, Monad, MonadIO, MonadReader FedQueueEnv)
 
+reqOrigin :: FedQueueClient c (RequestId, Domain)
+reqOrigin = do
```
fisx, line 194: I would just inline this.  (It's two lines instead of one, but shorter ones, so I'm not sure you've even saved *any* characters?)

libs/wire-api-federation/src/Wire/API/Federation/Version.hs
```diff
@@ -36,6 +57,9 @@ versionInt :: Version -> Int
 versionInt V0 = 0
 versionInt V1 = 1
 
+intToVersion :: Int -> Maybe Version
+intToVersion intV = find (\v -> versionInt v == intV) [minBound .. maxBound]
```
fisx, line 61:
```suggestion
intToVersion intV = find (\v -> versionInt v == intV) [minBound ..]
```

same thing, though...

services/background-worker/src/Wire/BackendNotificationPusher.hs
```diff
@@ -78,32 +81,91 @@ pushNotification runningFlag targetDomain (msg, envelope) = do
...
+                Log.err $
+                  Log.msg (Log.val "Failed to get supported API versions, the notification will be ignored")
+                    . Log.field "domain" (domainText targetDomain)
+                    . Log.field "error" (displayException e)
+                throwM e
+              Right vi -> pure . Set.fromList . fmap versionInt . vinfoSupported $ vi
+        case mostRecentTuple bodyVersions (notifications bundle) remoteVersions of
+          Nothing ->
+            -- TODO(md): do more severe logging warning the site operator
+            Log.err $
```
fisx, line 153: I would make this Fatal or Critical, whatever the loudest level.

fisx, line 153: this probably creates a lot of traffic on the log service.  we should probably think about logging this only once every minute or so?  not sure if that's feasible.

pcapriotti: Done

pcapriotti: This is a pretty serious condition, I'm not sure it's worth adding complexity and obfuscating the logs to avoid some potential noise.

services/galley/src/Galley/Effects/BackendNotificationQueueAccess.hs
```diff
@@ -23,5 +22,49 @@ data BackendNotificationQueueAccess m a where
...
+    Functor f,
+    Member (Error FederationError) r,
+    Member BackendNotificationQueueAccess r
+  ) =>
+  Q.DeliveryMode ->
+  f (Remote x) ->
+  (Remote x -> FedQueueClient c a) ->
+  Sem r [Remote a]
+enqueueNotificationsConcurrentlyBuckets m r q =
+  send (EnqueueNotificationsConcurrentlyBuckets m r q) >>= either throw pure
```
akshaymankar, line 70: This looks absurd, the only exception that can happen is federation not being configured. Should we just remove these `Either`s from the API of this effect and just throw them in IO?

services/galley/src/Galley/Intra/BackendNotificationQueue.hs
```diff
@@ -78,6 +80,18 @@ enqueueNotificationsConcurrently m xs f = do
...
+  Q.DeliveryMode ->
+  f (Remote x) ->
+  (Remote x -> FedQueueClient c a) ->
+  ExceptT FederationError App [Remote a]
+enqueueNotificationsConcurrentlyBuckets m xs f = do
+  chanVar <- getChannel
+  lift $ pooledForConcurrentlyN 8 (toList xs) $ \r ->
+    qualifyAs r
+      <$> enqueueSingleNotification (tDomain r) m chanVar (f r)
+
```
akshaymankar, line 94: Looks wrong, the other one is bucketing, not this one.

services/galley/test/integration/API.hs
```diff
@@ -242,7 +241,6 @@ tests s =
               test s "existing has password, requested has password - 409" postCodeWithPasswordExistsWithPasswordRequested
             ],
           test s "remove user with only local convs" removeUserNoFederation,
-          test s "remove user with local and remote convs" removeUser,
```
akshaymankar, line 245: Where did these tests go? I don't see them being added to the integration package. Were they already ported?

akshaymankar (changes requested):
Comments inline.

services/background-worker/src/Wire/BackendNotificationPusher.hs
```diff
@@ -78,32 +82,99 @@ pushNotification runningFlag targetDomain (msg, envelope) = do
...
+                fcEnv = FederatorClientVersionedEnv {..}
+            liftIO $ either throwM pure =<< sendNotification fcEnv notif.targetComponent notif.path notif.body
+            lift $ ack envelope
+            metrics <- asks backendNotificationMetrics
+            withLabel metrics.pushedCounter (domainText targetDomain) incCounter
+            withLabel metrics.stuckQueuesGauge (domainText targetDomain) (flip setGauge 0)
+
+-- | Find the pair that maximises b.
+pairedMaximumOn :: Ord b => (a -> b) -> [a] -> (a, b)
+pairedMaximumOn f = maximumBy (compare `on` snd) . map (id &&& f)
```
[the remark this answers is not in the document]
pcapriotti: I don't think a test would have a lot of value here. The types are general enough that there is essentially only one implementation. Well, two, since there is a `minimum` version, but hopefully it's pretty clear which one this is. Not sure how a test would make that clearer.

services/galley/src/Galley/Effects/BackendNotificationQueueAccess.hs
pcapriotti: Why absurd? The only error is not being able to talk to the queue. What's wrong with having those errors as `Either`?

services/galley/src/Galley/Intra/BackendNotificationQueue.hs
pcapriotti: This is named according to the scheme used by synchronous notifications. This function assumes that the inputs are already bucketed by remote. The other one does the bucketing automatically.

services/galley/test/integration/API.hs
pcapriotti: As far as I could see, these tests are redundant. All of the functionality (except the assertions on federated RPCs) is being tested in the new integration.

services/background-worker/src/Wire/BackendNotificationPusher.hs
```diff
@@ -78,32 +82,99 @@ pushNotification runningFlag targetDomain (msg, envelope) = do
...
+            -- 'BackendNotification's anymore.
+            ceFederator <- asks (.federatorInternal)
+            ceHttp2Manager <- asks http2Manager
+            let ceOriginDomain = notif.ownDomain
+                ceTargetDomain = targetDomain
+                ceOriginRequestId = fromMaybe (RequestId "N/A") notif.requestId
+                cveEnv = FederatorClientEnv {..}
+                cveVersion = Just V0 -- V0 is assumed for non-versioned queue messages
+                fcEnv = FederatorClientVersionedEnv {..}
+            liftIO $ either throwM pure =<< sendNotification fcEnv notif.targetComponent notif.path notif.body
```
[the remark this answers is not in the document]
pcapriotti: Done.

pcapriotti:
I think I addressed all the issues. Let me know if I missed anything.

mdimjasevic:
I approve this PR, but given that I started the PR, GitHub wouldn't let me approve it. @pcapriotti , can you approve it for me?

pcapriotti (approved):

akshaymankar (approved):

PR #3835, opened 2024-01-24
Re-add http-client extension for connection re-use

elland:
But disable it for LH calls like it was before.


elland:
We've decided for a simpler approach #3825

PR #3839, opened 2024-01-26
fixing grepinclude references for docs.wire.com and adjusting nix build context; updating TLS documentation

jschumacher-wire:
## Checklist

### DIFF
--- a/nix/default.nix
+++ b/nix/default.nix
@@ -57,7 +57,7 @@ let
         # GrepInclude snippets in the docs refer to files under ../charts/ and ../services/, 
         # so we need to copy these too before building.
         # FUTUREWORK: perhaps there is a nicer way to copy everything that does not need 3 separate lines,
-        # however the statement `../` inside `cp -rH ${pkgs.nix-gitignore.gitignoreSource [] ../}* .` is not valid.
+        # however the statement `../` inside `cp -rH $#{pkgs.nix-gitignore.gitignoreSource [] ../}* .` is not valid.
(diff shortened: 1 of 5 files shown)


smatting (approved):
looks good, a bit unorthodox to (temporarily) copy everything, but hey if it works :+1:

nix/default.nix
```diff
@@ -52,9 +52,13 @@ let
         nativeBuildInputs = docsPkgs ++ [ pkgs.gnumake ];
       }
       ''
-        cp -rH ${pkgs.nix-gitignore.gitignoreSource [] ../docs}/* .
-        chmod -R +w ./src
-        cp ${../CHANGELOG.md} ./src/changelog/changelog.md
+        mkdir docs charts services
+        cp -rH ${pkgs.nix-gitignore.gitignoreSource [] ../docs}/* docs/
+        cp -rH ${pkgs.nix-gitignore.gitignoreSource [] ../charts}/* charts/
```
jschaul, line 57: You could add a comment for the future reader of this code. e.g.:

```suggestion
        # GrepInclude snippets in the docs refer to files under ../charts/ and ../services/, 
        # so we need to copy these too before building.
        # FUTUREWORK: perhaps there is a nicer way to copy everything that does not need 3 separate lines,
        # however the statement `../` inside `cp -rH ${pkgs.nix-gitignore.gitignoreSource [] ../}* .` is not valid.
        cp -rH ${pkgs.nix-gitignore.gitignoreSource [] ../charts}/* charts/
```

jschaul (approved):
Adding a comment is nice I think, otherwise LGTM.

PR #3841, opened 2024-01-26
Refactor/use GitHub forks

marcoconti83:
Updated references to git repositories to wireapp forks


smatting:
Note to self: the nix hashes need to be adjusted

stefanwire (approved):

PR #3847, opened 2024-01-29
Revert "Improve usage of http-manager (fixes for fingerprint verification)"

MangoIV:
Reverts wireapp/wire-server#3825

I have looked at our logs and for ~ 8 hours we get tons of name resolution errors. there's exactly one PR whose logs haven't been reaped yet and that one doesn't have the name resolution errors, I do have to assume that it's this PR. 

The CI was green for 4 runs in a row but didn't catch this...


MangoIV:
it actually doesn't change anything, so it's unrelated.

MangoIV:
@elland

PR #3849, opened 2024-01-31
Federation-v0 setup in integration tests

pcapriotti:
This PR finishes the setup of the federation-v0 backend in integration tests and makes it possible to write tests that use it.

It also makes federator ingresses use cert-manager in CI, so they can share an issuer with the federation-v0 instance and get certificates that are suitable for talking to each other.

https://wearezeta.atlassian.net/browse/WPB-183

## Checklist


hack/helmfile.yaml
```diff
@@ -118,13 +118,14 @@ releases:
     chart: '../.local/charts/nginx-ingress-services'
     values:
       - './helm_vars/nginx-ingress-services/values.yaml.gotmpl'
-      - './helm_vars/nginx-ingress-services/certificates-namespace1.yaml'
```
supersven, line 121: `certificates-namespace1.yaml` and `certificates-namespace2.yaml` are still in `.gitignore`. I think they can be removed there, too.

supersven (approved):
LGTM :+1: (Only a couple of small remarks.)

PR #3852, opened 2024-01-31
Use http-client fork again

akshaymankar:
The fork has been updated with latest http-client. The new logic we wrote for verifying the pinned public key fingerprints (for bots and legalhold) is wrong, so we go back to old ways of verifying these fingerprints.

## Checklist

 - [x] Add a new entry in an appropriate subdirectory of `changelog.d` - This is majorly a revert so the changelogs also get reverted.

### DIFF
--- a/nix/haskell-pins.nix
+++ b/nix/haskell-pins.nix
@@ -116,6 +116,20 @@ let
       };
     };
 
+    http-client = {
+      src = fetchgit {
+        url = "https://github.com/wireapp/http-client";
+        rev = "e6beaccdcc8fac892d2437ebbff029fef3551a13";
+        sha256 = "sha256-z47GlT+tHsSlRX4ApSGQIpOpaZiBeqr72/tWuvzw8tc=";
+      };
+      packages = {
+        "http-client" = "http-client";
+        "http-client-tls" = "http-client-tls";
+        "http-client-openssl" = "http-client-openssl";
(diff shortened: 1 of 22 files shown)


nix/haskell-pins.nix
```diff
@@ -116,6 +116,20 @@ let
...
+        rev = "e6beaccdcc8fac892d2437ebbff029fef3551a13";
+        sha256 = "sha256-z47GlT+tHsSlRX4ApSGQIpOpaZiBeqr72/tWuvzw8tc=";
+      };
+      packages = {
+        "http-client" = "http-client";
+        "http-client-tls" = "http-client-tls";
+        "http-client-openssl" = "http-client-openssl";
+        "http-conduit" = "http-conduit";
+      };
+    };
```
MangoIV, line 131: should I open a PR against our master so this is visible better?

MangoIV (approved):
Looks good to me.

nix/haskell-pins.nix
akshaymankar: Its actually here: https://github.com/wireapp/http-client/pull/5
I am waiting for the CI to be green and us to verify that bots work again. Once that happens we can merge that PR and update the revision here.

akshaymankar:
Didn't wait for CI because the last commit didn't even change the sha256 of the http-client fork.

PR #3860, opened 2024-01-31
commenting out the docs as per the ticket

arthurwolf:
## Checklist

### DIFF
--- a/docs/src/understand/sft.md
+++ b/docs/src/understand/sft.md
@@ -94,6 +94,9 @@ Conferencing in a federated environment assumes that each domain participating i
 conference will use an SFT in its own domain. The SFT in the caller's domain is called
 the `anchor SFT`.
 
+<!-- Commenting out Multi-SFT architecture as-per WPB-2566 -->
+
+<!-- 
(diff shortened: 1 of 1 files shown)


rohan-wire (changes requested):
I made a number of changes to reflect that multi-SFT is only supported in the configuration of the last diagram. I proposed a new diagram based on the first diagram in the original page. 

1) Please redraw `multi-sft-concept.png` in a consistent style.
2) Please save the original first diagram (`multi-sft-noturn.png`), as Wire will want to support the direct SFT to SFT (sans Federation TURN server) model in the future for any customer other than Bund. This was actually implemented but removed as M2 shipped. I don't think there is any plan to support the TURN server to TURN server with DTLS muxing, but I thought we would get a review from Dusan and/or Chris before deleting `img/multi-sft-turn.png`
3) Please remove all references to Restund in this page and its diagrams and replace them with "TURN server". Please do not reference coTURN in any of the understanding sections. We should only reference coTURN or Restund if absolutely referring to something like migrating from TURN server to the other.

PR #3869, opened 2024-02-01
[fix] use the correct API in the integration tests

MangoIV:
## Checklist

### DIFF
--- a/integration/test/API/BrigInternal.hs
+++ b/integration/test/API/BrigInternal.hs
@@ -1,5 +1,6 @@
 module API.BrigInternal where
 
+import API.Brig (AddClient (..))
 import API.Common
 import qualified Data.Aeson as Aeson
 import Data.Aeson.Types (Pair)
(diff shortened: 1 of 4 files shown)


integration/test/API/BrigInternal.hs
```diff
@@ -27,6 +28,28 @@ instance Default CreateUser where
...
       }
 
+iAddClient ::
+  (HasCallStack, MakesValue user) =>
+  user ->
+  AddClient ->
+  App Response
+iAddClient user args = do
+  uid <- objId user
+  req <- baseRequest user Brig Unversioned $ "/i/clients/" <> uid
```
MangoIV, line 38: I'm not entirely sure why we were using `Unversioned` here, either, perhaps @pcapriotti  you can verify?

MangoIV: nvm, I guess this is just because internal APIs are unversioned in general, :D

PR #3871, opened 2024-02-05
Unblock release.

fisx:
https://wearezeta.atlassian.net/browse/WPB-6392

### DIFF
--- a/integration/test/Test/LegalHold.hs
+++ b/integration/test/Test/LegalHold.hs
@@ -165,39 +165,12 @@ testLHMessageExchange (TaggedBool clients1New) (TaggedBool clients2New) (TaggedB
                 (False, False, False, False) ->
                   -- no LH in the picture
                   check 201 Nothing
-                (False, False, False, True) ->
-                  -- sender has old clients and not given consent, recipient has LH device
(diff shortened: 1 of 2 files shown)


integration/test/Test/LegalHold.hs
```diff
@@ -210,6 +183,7 @@ testLHMessageExchange (TaggedBool clients1New) (TaggedBool clients2New) (TaggedB
                 (True, True, True, True) ->
                   -- everybody happy with LH
                   check 201 Nothing
+                _ -> pure ()
```
stefanwire, line 186: No explicit 201 in the general case?

```suggestion
                _ -> check 201 Nothing
```

stefanwire (approved):
I would rather leave the explicit cases with their error codes as future work comments in the code. I remember how much time it took to sort out which error should be raised in what case. That would be redundant with Git, but the next guy dealing with this logic might not even be aware of the work done so far. And eventually, we want this logic to be in place, that is, as soon as the clients could catch up.

fisx:
> And eventually, we want this logic to be in place, that is, as soon as the clients could catch up.

I'll make a revert PR as soon as this is merged, and link it to https://wearezeta.atlassian.net/browse/WPB-6393, ok?

PR #3880, opened 2024-02-12
Added reason to bad gateway error for RPCs.

elland:


battermann (approved):

services/brig/src/Brig/Provider/RPC.hs
```diff
@@ -97,7 +97,7 @@ createBot scon new = do
       extReq scon ["bots"]
         . method POST
         . Bilge.json new
-    onExc ex = lift (extLogError scon ex) >> throwE ServiceUnavailable
+    onExc ex = lift (extLogError scon ex) >> throwE (ServiceUnavailableWith $ show ex)
```
battermann, line 100: maybe use https://hoogle.zinfra.io/file/nix/store/5x7yh702gmyiwhbiq1qbybknyk1fl4ys-ghc-9.4.8-doc/share/doc/ghc/html/libraries/ghc-9.4.8/GHC-Utils-Exception.html#v:displayException instead of show?

PR #3885, opened 2024-02-13
fix integration-cleanup.sh to match prefix only

stefanwire:
The `-f` filter is a regex and should match the prefix `test-`, thus,
the regex should be `^test-`. Without `^`, the search string is looked
up in the entire release name.

## Checklist


supersven:
The follow-up ticket is: https://wearezeta.atlassian.net/browse/WPB-6572

supersven (approved):
LGTM :+1:

PR #3895, opened 2024-02-20
change externallyCreated to useFakeS3 and change its depth

arthurwolf:
## Checklist

### DIFF
--- a/charts/nginx-ingress-services/values.yaml
+++ b/charts/nginx-ingress-services/values.yaml
@@ -100,7 +100,7 @@ service:
   s3:
     externalPort: 9000
     serviceName: fake-aws-s3
-    externallyCreated: false # See note below
+  useFakeS3: false # See note below
   teamSettings:
     externalPort: 8080
   accountPages:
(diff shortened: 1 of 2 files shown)


charts/nginx-ingress-services/values.yaml
```diff
@@ -142,9 +142,8 @@ config:
 #
 # For Services:
 # service:
-#   s3:
-#     externallyCreated: true
-#     ^ externallyCreated might be useful if S3 access is provided by
+#   useFakeS3: true
+#     ^ useFakeS3 might be useful if S3 access is provided by
```
julialongtin, line 146:
```suggestion
#     ^ useFakeS3 should be enabled if S3 access is to be provided by fake-aws-s3, inside of the kubernetes cluster.
```

julialongtin: this paragraph needs reworked. tl;dr: fakeAwsS3 should be used when s3 is to be provided by fake-aws-s3 within the cluster. when it is something outside of the cluster (like minio-external), we should leave this setting off. this setting creates a fake-aws-s3 service inside of the cluster, which should be what is referred to in the brig configuration.

julialongtin (approved):

PR #3899, opened 2024-02-26
Fix openapi docs of UTCTime and UTCTimeMillis

pcapriotti:
Fix openapi docs inconsistency with datetime fields. We have two kinds of datetime fields: with and without a millisecond component. They were both documented as having the millisecond component.

https://wearezeta.atlassian.net/browse/WPB-6529

## Checklist

 - [x] Add note to "API changes" page


stefanwire (approved):

supersven (approved):
LTGM :+1:

PR #3902, opened 2024-02-26
Handle async exceptions so Control-C works as expected

elland:
Integration tests weren't being killed on control-c, when investigating a proposed fix we realised that tests were running sequentially on the main thread, causing the issue with exception handling. Since running them in parallel increases flakiness, I'm holding off on correcting this until we do some of the mitigation work we have planned.

### DIFF
--- a/integration/test/Testlib/ModService.hs
+++ b/integration/test/Testlib/ModService.hs
@@ -386,7 +386,7 @@ retryRequestUntil :: HasCallStack => App Bool -> String -> App ()
 retryRequestUntil reqAction err = do
   isUp <-
     retrying
-      (limitRetriesByCumulativeDelay (7 * 1000 * 1000) (fibonacciBackoff (300 * 1000)))
+      (limitRetriesByCumulativeDelay (4 * 1000 * 1000) (fibonacciBackoff (200 * 1000)))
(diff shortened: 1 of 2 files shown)


pcapriotti:
On second thought, maybe we need to limit the number of concurrent tests. Otherwise we might try to use too many file descriptors or overwhelm some of the services.

MangoIV:
https://hackage.haskell.org/package/unliftio-0.2.25.0/docs/UnliftIO-Async.html#v:pooledForConcurrentlyN

MangoIV:
I used this successfully in similar situations

elland:
@MangoIV the problem with the UnliftIO alternative is that the async exception doesn't properly kill the threads, leaving us with the same IO leakage as before. One option is to keep the async handler in place, another might be something like async-pooled. 🤔

pcapriotti (approved):

elland:
@pcapriotti @MangoIV it seems running the tests in parallel at the moment increase our flakiness too much. I would for now just fix the Control-C handling and wait until our other flakiness mitigating changes to parallelise the tests.

pcapriotti:
That's fine, but we probably want to look into running the tests asynchronously at some point.

PR #3907, opened 2024-02-27
Welcome notification bug

pcapriotti:
This implements collecting welcome notifications by user before sending them to gundeck. Otherwise, gundeck would attempt to store one entry for each client, and one would overwrite the others.

We had already fixed a similar bug for MLS messages (https://github.com/wireapp/wire-server/pull/3610), but neglected to apply a similar fix to welcome messages.

https://wearezeta.atlassian.net/browse/WPB-6852

## Checklist

### DIFF
--- a/services/galley/src/Galley/API/MLS/Welcome.hs
+++ b/services/galley/src/Galley/API/MLS/Welcome.hs
@@ -96,7 +96,7 @@ sendLocalWelcomes qcnv qusr con now welcome lclients = do
         map (\(u, cs) -> Recipient u (RecipientClientsSome (List1 cs)))
           . Map.assocs
           . foldr
-            (uncurry (\u c -> Map.insertWith (<>) u (pure c)))
+            (\(u, c) -> Map.insertWith (<>) u (pure c))
             mempty
           $ tUnqualified lclients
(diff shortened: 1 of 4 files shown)


services/galley/src/Galley/API/MLS/Welcome.hs
```diff
@@ -88,9 +92,18 @@ sendLocalWelcomes ::
...
 sendLocalWelcomes qcnv qusr con now welcome lclients = do
+  -- only create one notification per user
+  let rcpts =
+        map (\(u, cs) -> Recipient u (RecipientClientsSome (List1 cs)))
+          . Map.assocs
+          . foldr
+            (uncurry (\u c -> Map.insertWith (<>) u (pure c)))
+            mempty
+          $ tUnqualified lclients
+  traceM $ "sending welcomes to: " <> show rcpts
```
stefanwire, line 103: `traceM` left

stefanwire (approved):

services/galley/src/Galley/API/MLS/Welcome.hs
pcapriotti: Thanks, fixed.

PR #3908, opened 2024-02-27
added a configuration option for IP binding for coturn, and example u…

arthurwolf:
…se (commented) in the values file

## Checklist

### DIFF
--- a/charts/coturn/values.yaml
+++ b/charts/coturn/values.yaml
@@ -26,6 +26,10 @@ coturnTurnListenPort: 3478
 coturnMetricsListenPort: 9641
 coturnTurnTlsListenPort: 5349
 
+# If you need to specify which IP Coturn should bind to.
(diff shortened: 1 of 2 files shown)


charts/coturn/values.yaml
```diff
@@ -108,3 +108,7 @@ livenessProbe:
 readinessProbe:
   timeoutSeconds: 5
   failureThreshold: 5
+
+# If you need to specify which IP Coturn should bind to.
```
julialongtin, line 112: Could you move this whole block up to line 25, so that it is next to listening port?

julialongtin (approved):

charts/coturn/values.yaml
arthurwolf: @julialongtin moved as requested. requesting review so it can be merged.

PR #3873, opened 2024-02-05
Fix: (de-)serialization of client capabilities.

fisx:
https://wearezeta.atlassian.net/browse/WPB-6400

## Checklist

 - [x] Add a new entry in an appropriate subdirectory of `changelog.d`
 - [x] Read and follow the [PR guidelines](https://docs.wire.com/developer/developer/pr-guidelines.html)

PR #3909, opened 2024-02-28
Serialisation of client capabilities (take 2)

pcapriotti:
Fix an oddity in the serialisation of capabilities in the Client structure.

This PR replaces #3873 and simplifies the implementation. No new types are added, and the difference in serialisation is obtained using the Versioned wrapper, which we are already using for versioned changes in serialisation.

The change in the internal API between brig and galley is not backward compatible, so this can cause temporary failures during upgrades. The invariant that brig's clients are a superset of galley's clients will be maintained.

**Note**: #3904 is a previous attempt at this fix, which overlooked the fact that the `Client` structure is used in user events, so it was reverted.

https://wearezeta.atlassian.net/browse/WPB-6400

## Checklist


libs/wire-api/src/Wire/API/User/Client.hs
```diff
@@ -373,17 +379,17 @@ instance Swagger.ToSchema UserClientsFull where
 
 instance ToJSON UserClientsFull where
   toJSON =
-    toJSON . Map.foldrWithKey' fn Map.empty . userClientsFull
+    toJSON . Map.foldrWithKey' f Map.empty . userClientsFull
```
fisx, line 382: it's not a big deal, ignore this if you like.  but in general (a) i don't like single-character identifiers, even locally; (b) i want us all to avoid edit wars.

integration/test/Test/Notifications.hs
```diff
@@ -89,3 +91,18 @@ testInvalidNotification = do
     void $
       getNotifications user def {since = Just notifId}
         >>= getJSON 404
+
+-- | Check that client-add notifications use the V5 format:
+-- @
+--   "capabilities": { "capabilities": [..] }
+-- @
```
fisx, line 98:
```suggestion
-- | Check that client-add notifications use the V5 format:
-- @
--   "capabilities": { "capabilities": [..] }
-- @
--
-- Migration plan: clients must be able to parse both old and new schema starting from V6.  Once V5 is deprecated, the backend can start sending notifications in the new form.
```

[see also](https://wearezeta.atlassian.net/wiki/pages/diffpagesbyversion.action?pageId=1035632650&selectedPageVersions=5&selectedPageVersions=4)

fisx (approved):

libs/wire-api/src/Wire/API/User/Client.hs
pcapriotti: You don't like single-character identifiers, but two-character ones are ok? That's a bit of a strange stance. But sure, I'll revert it.

mdimjasevic:
@pcapriotti , in an attempt to get the PR through the CI, I took the liberty to rebase the PR on top of `deveop`. I hope you don't mind.

mdimjasevic:
As far as I can tell, now we're getting actual test failures: https://concourse.ops.zinfra.io/builds/49403780

pcapriotti:
> As far as I can tell, now we're getting actual test failures: https://concourse.ops.zinfra.io/builds/49403780

Thanks for rebasing. Yes, that's right. It should be fixed now.

PR #3912, opened 2024-02-29
Refactor and test user event serialisation

pcapriotti:
https://wearezeta.atlassian.net/browse/WPB-6892

## Checklist


libs/brig-types/src/Brig/Types/User/Logging.hs
battermann, line 1: This file should be removed.

libs/wire-api/src/Wire/API/UserEvent.hs
```diff
@@ -0,0 +1,451 @@
...
+                  _ClientRemoved
+                  (field "client" (idObjectSchema schema))
+              )
+          EventTypeConnection ->
+            tag
+              _ConnectionEvent
+              ( ConnectionUpdated
+                  <$> ucConn .= field "connection" schema
+                  <*> ucName .= maybe_ (optField "user" (object "UserName" (field "name" schema)))
+              )
```
battermann, line 398: A lot of nesting, but I guess it's fine :sweat_smile:

services/brig/src/Brig/API/Connection.hs
```diff
@@ -124,17 +125,16 @@ createConnectionToLocalUser self conn target = do
       Created <$> insert Nothing Nothing
   where
     insert :: Maybe UserConnection -> Maybe UserConnection -> ExceptT ConnectionError (AppT r) UserConnection
-    insert s2o o2s = lift $ do
+    insert _s2o _o2s = lift $ do
```
battermann, line 128: Remove parameters, maybe?

battermann (approved):
I would request to remove the empty file. The rest is just opinion and looks good otherwise. Approved in advance.

libs/brig-types/src/Brig/Types/User/Logging.hs
pcapriotti: Ah right, thanks for catching this.

libs/wire-api/src/Wire/API/UserEvent.hs
pcapriotti: It follows the structure of the type. I don't really see anything wrong with this.

PR #3914, opened 2024-02-29
WPB-6577 Fix user creation conflict in SCIM

battermann:
For a description of the manual test approach, see comments of https://wearezeta.atlassian.net/browse/WPB-6577

## Checklist

### DIFF
--- a/services/spar/src/Spar/Scim/User.hs
+++ b/services/spar/src/Spar/Scim/User.hs
@@ -66,6 +66,7 @@ import Polysemy.Input
 import qualified SAML2.WebSSO as SAML
 import Spar.App (getUserByUrefUnsafe, getUserByUrefViaOldIssuerUnsafe, getUserIdByScimExternalId)
 import qualified Spar.App
+import Spar.Intra.BrigApp as Intra
 import qualified Spar.Intra.BrigApp as Brig
 import Spar.Options
 import Spar.Scim.Auth ()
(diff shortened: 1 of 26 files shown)


services/spar/src/Spar/Scim/User.hs
```diff
@@ -750,6 +781,9 @@ deleteScimUser tokeninfo@ScimTokenInfo {stiTeam, stiIdP} uid =
           -- (because that owner won't be managed by SCIM in the first place), but if it ever becomes
           -- possible, we should do a check here and prohibit it.
           unless (userTeam brigUser == Just stiTeam) $
+            -- users from other teams get you a 404.
+            -- users from other teams get you a 404.
+
```
fisx, line 786:
```suggestion
```

services/spar/src/Spar/Sem/ScimExternalIdStore/Spec.hs
```diff
@@ -45,15 +46,17 @@ propsForInterpreter interpreter extract lower = do
     prop "insert/lookup" $ prop_insertLookup (Just $ show . void . extract) lower
     prop "insert/insert" $ prop_insertInsert (Just $ show . void . extract) lower
 
+-- FUTUREWORK: Add prop tests for missing operations
```
fisx, line 49: have you given this a try?  if not i can do that.

services/spar/src/Spar/Sem/ScimExternalIdStore.hs
```diff
@@ -22,19 +22,25 @@ module Spar.Sem.ScimExternalIdStore
...
 import Polysemy.Check (deriveGenericK)
+import Spar.Scim.Types
 import Wire.API.User.Identity (Email)
+import Wire.API.User.Scim
 
 data ScimExternalIdStore m a where
   Insert :: TeamId -> Email -> UserId -> ScimExternalIdStore m ()
   Lookup :: TeamId -> Email -> ScimExternalIdStore m (Maybe UserId)
   Delete :: TeamId -> Email -> ScimExternalIdStore m ()
+  InsertStatus :: TeamId -> ValidExternalId -> UserId -> ScimUserCreationStatus -> ScimExternalIdStore m ()
```
fisx, line 42:
```suggestion
  -- NB: the fact that we are using `Email` in some cases here and `ValidExternalId` in others has historical reasons (this table was only used for non-saml accounts in the past, now it is used for *all* scim-managed accounts).  the interface would work equally well with just `Text` here (for unvalidated scim external id).
  InsertStatus :: TeamId -> ValidExternalId -> UserId -> ScimUserCreationStatus -> ScimExternalIdStore m ()
```

services/spar/src/Spar/Scim/User.hs
```diff
@@ -530,7 +546,22 @@ createValidScimUser tokeninfo@ScimTokenInfo {stiTeam} vsu@(ST.ValidScimUser veid
...
+      lift $ ScimExternalIdStore.insertStatus stiTeam veid buid ScimUserCreated
       pure storedUser
+  where
+    incompleteUserCreationCleanUp :: UserId -> Scim.ScimError -> Scim.ScimHandler (Sem r) ()
+    incompleteUserCreationCleanUp buid e = do
+      -- something went wrong while storing the user in brig
+      -- we can try clean up now, but if brig is down, we can't do much
+      -- maybe retrying the user creation in brig is also an option?
+      -- after clean up we rethrow the error so the handler returns the correct failure
```
fisx, line 558: what about *not* rethrowing, but looking up the status again once or twice?  that'd make for swifter cleanup, especially when the batch job only runs once a day.

fisx (commented):
looks good!  i have one actual change request, otherwise looks good!  let me know if you want to discuss this.

services/spar/src/Spar/Scim/User.hs
battermann: It only rethrows if `deleteScimUser` fails. AFAICT the deletion is synchronous.

services/spar/src/Spar/Sem/ScimExternalIdStore/Spec.hs
battermann: You can go ahead, but maybe better in a separate PR.

fisx (approved):

PR #3916, opened 2024-03-04
coturn cert-reloader sidecar config: process name should not contain the path (helm chart)

mastaab:
The sidecar only looks for the process name.

## Checklist


akshaymankar (approved):
:rocket:

PR #3919, opened 2024-03-04
Make federator error logs more informative.

fisx:
https://wearezeta.atlassian.net/browse/WPB-6998

## Checklist

### DIFF
--- a/services/federator/src/Federator/Remote.hs
+++ b/services/federator/src/Federator/Remote.hs
@@ -63,10 +63,10 @@ data RemoteError
   deriving (Show)
 
 instance AsWai RemoteError where
-  toWai (RemoteError _ _ e) =
-    federationRemoteHTTP2Error e
-  toWai (RemoteErrorResponse _ _ status body) =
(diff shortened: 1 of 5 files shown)


services/federator/src/Federator/Remote.hs
```diff
@@ -63,10 +63,10 @@ data RemoteError
   deriving (Show)
 
 instance AsWai RemoteError where
-  toWai (RemoteError _ _ e) =
-    federationRemoteHTTP2Error e
-  toWai (RemoteErrorResponse _ _ status body) =
-    federationRemoteResponseError status body
+  toWai (RemoteError target msg err) =
+    federationRemoteHTTP2Error (cs $ show (target, msg)) err
```
akshaymankar, line 67: The error type contains `errorData` we should use that instead of stringifying everything. If we really want to stringify this, can we use something better than a tuple?

akshaymankar (changes requested):

akshaymankar (approved):

PR #3929, opened 2024-03-06
Deflake integration tests

fisx:
https://wearezeta.atlassian.net/browse/WPB-6985

## Checklist


fisx:
Using `Chan` for this is not particularly elegant or efficient, and it's not easy to establish that the code added here is deadlock-free.  [streamly-process](https://streamly.composewell.com/) seems like a nice alternative that *may* *possibly* make the `Chan`s, the `LineOrEOF` type, and most of the error handling unnecessary, but I'm not sure how to duplicate output streams here?

integration/test/Test/SelfTest.hs
```diff
@@ -0,0 +1,65 @@
...
+--
+-- This program is distributed in the hope that it will be useful, but WITHOUT
+-- ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+-- FOR A PARTICULAR PURPOSE. See the GNU Affero General Public License for more
+-- details.
+--
+-- You should have received a copy of the GNU Affero General Public License along
+-- with this program. If not, see <https://www.gnu.org/licenses/>.
+
+module Test.SelfTest where
```
akshaymankar, line 18: There is also Test.Demo, which also contains bunch of self tests, we should consolidate these.

integration/test/Testlib/ModService.hs
```diff
@@ -353,44 +347,54 @@ withProcess resource overrides service = do
...
           (_, Just stdoutHdl, Just stderrHdl, ph) <- createProcess (proc exe ["-c", tempFile]) {cwd = cwd, std_out = CreatePipe, std_err = CreatePipe}
-          let prefix = "[" <> execName <> "@" <> domain <> "] "
-          let colorize = fromMaybe id (lookup execName processColors)
-          void $ forkIO $ logToConsole colorize prefix stdoutHdl
-          void $ forkIO $ logToConsole colorize prefix stderrHdl
+          (out1, out2) <- mkChans stdoutHdl
+          (err1, err2) <- mkChans stderrHdl
+          void $ forkIO $ logChanToConsole execName domain out1
+          void $ forkIO $ logChanToConsole execName domain err1
+          putMVar serviceHandleMVar (out2, err2, ph)
```
akshaymankar, line 362: As discussed it is probably best to do this without the MVar. I will push the code in a bit.

fisx (commented):
thanks @akshaymankar, changes look good!

akshaymankar (approved):
Let's merge it we read in flake-news what's the problem :newspaper_roll:

stefanwire (approved):

PR #3942, opened 2024-03-11
[feat] use new script for sbom generation

MangoIV:
https://wearezeta.atlassian.net/browse/SEC-596
## Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -551,9 +551,7 @@ helm-template-%: clean-charts charts-integration
 # directory as the script that is being executed)
 .PHONY: upload-bombon
 upload-bombon:
-	pushd ./hack/bin	 
-	./bombon.hs -- \
+	cd ./hack/bin	&& ./bombon.hs -- \
 		--project-version $(HELM_SEMVER) \
 		--api-key $(DEPENDENCY_TRACK_API_KEY) \
 		--auto-create
-	popd
--- a/Makefile
+++ b/Makefile
@@ -546,11 +546,14 @@ helm-template-%: clean-charts charts-integration
 	./hack/bin/helm-template.sh $(*)
 
 # Ask the security team for the `DEPENDENCY_TRACK_API_KEY` (if you need it)
+# changing the directory is necessary because of some quirkiness of how 
+# runhaskell / ghci behaves (it doesn't find modules that aren't in the same 
+# directory as the script that is being executed)
 .PHONY: upload-bombon
 upload-bombon:
-	nix build -f nix wireServer.allLocalPackagesBom -o "bill-of-materials.$(HELM_SEMVER).json"
-	./hack/bin/bombon.hs -- \
-		--bom-filepath "./bill-of-materials.$(HELM_SEMVER).json" \
+	pushd ./hack/bin	 
+	./bombon.hs -- \
 		--project-version $(HELM_SEMVER) \
 		--api-key $(DEPENDENCY_TRACK_API_KEY) \
 		--auto-create
+	popd
--- a/Makefile
+++ b/Makefile
@@ -551,7 +551,7 @@ helm-template-%: clean-charts charts-integration
 # directory as the script that is being executed)
 .PHONY: upload-bombon
 upload-bombon:
-	cd ./hack/bin	&& ./bombon.hs -- \
+	cd ./hack/bin && ./bombon.hs -- \
 		--project-version $(HELM_SEMVER) \
 		--api-key $(DEPENDENCY_TRACK_API_KEY) \
 		--auto-create
--- a/changelog.d/5-internal/SEC-596
+++ b/changelog.d/5-internal/SEC-596
@@ -1,1 +1,1 @@
-use new script Sbom.hs to generate the wire-server sbom
+Create a new script (`Sbom.hs`) to generate the wire-server sbom (bill of material) file.
--- a/changelog.d/5-internal/SEC-596
+++ b/changelog.d/5-internal/SEC-596
@@ -1,0 +1,1 @@
+use new script Sbom.hs to generate the wire-server sbom
--- a/hack/bin/Sbom.hs
+++ b/hack/bin/Sbom.hs
@@ -29,7 +29,6 @@ import Data.ByteString.Lazy (LazyByteString)
 import Data.ByteString.Lazy qualified as BSL
 import Data.ByteString.Lazy.Char8 qualified as C8L
 import Data.Containers.ListUtils (nubOrd, nubOrdOn)
-import Data.Functor
 import Data.Functor.Identity
 import Data.Map (Map)
 import Data.Map qualified as M
@@ -58,6 +57,20 @@ data License = MkLicense
   }
   deriving stock (Eq, Ord, Show, Generic)
   deriving anyclass (FromJSON, ToJSON)
+
+sadSbomMeta :: Text -> Text -> [Text] -> SBomMeta Identity
+sadSbomMeta drvPath outPath directDeps 
+  = MkSBomMeta 
+  { drvPath = drvPath 
+  , outPath = Identity outPath 
+  , directDeps = Identity directDeps 
+  , description = Nothing 
+  , homepage = Nothing 
+  , licenseSpdxId = []
+  , name = Nothing 
+  , typ = Nothing 
+  , urls = [] 
+  , version = Nothing }
 
 data SBomMeta f = MkSBomMeta
   { drvPath :: Text,
@@ -229,7 +242,9 @@ discoverSBom outP metaDb = do
           Nothing -> \x -> do
             T.putStrLn ("no meta found for drv: " <> deriver <> "\ntrying approximate match")
             x >>= maybe
-              do \m -> T.putStrLn ("no approximate match found for: " <> deriver) $> m
+              do \m -> do 
+                     T.putStrLn ("no approximate match found for: " <> deriver) 
+                     pure $ M.insert k (sadSbomMeta deriver k deps) m
               do \match -> pure . M.insert k (proxyToIdentity match)
               do approximateMatch deriver metaDb
           Just pmeta -> fmap $ M.insert k $ proxyToIdentity pmeta
--- a/hack/bin/Sbom.hs
+++ b/hack/bin/Sbom.hs
@@ -15,37 +15,37 @@
 {-# OPTIONS_GHC -Wno-incomplete-uni-patterns #-}
 
 {-
-- the only place that has the data we need about the package is the evaluated nix code, i.e. before 
-  writhing the derivation; this is where we have `meta` and friends to get the data we need 
-- say we now want to build a dependency tree; the issue is to find all dependencies of the derivation. 
-  this is hard because 
-  - there are normal input attrs that the builder will have a look at but also 
-  - string contexts like 
-    ```nix 
-    x = /* bash */ '' 
-      cp ${pkgs.bla}/bin $out 
+- the only place that has the data we need about the package is the evaluated nix code, i.e. before
+  writhing the derivation; this is where we have `meta` and friends to get the data we need
+- say we now want to build a dependency tree; the issue is to find all dependencies of the derivation.
+  this is hard because
+  - there are normal input attrs that the builder will have a look at but also
+  - string contexts like
+    ```nix
+    x = /* bash */ ''
+      cp ${pkgs.bla}/bin $out
     '';
     ```
     would ignore dependencies on `pkgs.bla`
-- we can build the dependency graph independently (without knowing about the meta) but we somehow need 
+- we can build the dependency graph independently (without knowing about the meta) but we somehow need
   to obtain the meta itself
-- people don't always have a complete package set but more commonly are hand assembling things; we need 
+- people don't always have a complete package set but more commonly are hand assembling things; we need
   to give the possibility to build meta "databases" from package sets
-- we need to trace which dependencies are missing when querying the meta database against them 
-- collecting the meta also poses some issue 
+- we need to trace which dependencies are missing when querying the meta database against them
+- collecting the meta also poses some issue
   - nixpkgs is not a tree, but a more general graph
-  - it also not a DAG but it has loops 
-  - this means more specifically that we cannot without care recurse into it 
-  - even if we only recurse very shallowly, we soon start running out of memory, this means we probably need 
+  - it also not a DAG but it has loops
+  - this means more specifically that we cannot without care recurse into it
+  - even if we only recurse very shallowly, we soon start running out of memory, this means we probably need
     to do some on the fly filtering by "actual" dependencies
-  - this is similarly and issue, because it means that for every package we have to evaluate the entirety 
-    of the package set instead of being able to keep and persist the database 
-  - a more clean solution would probably be to at each time we recurse, a derivation that does the evaluation 
+  - this is similarly and issue, because it means that for every package we have to evaluate the entirety
+    of the package set instead of being able to keep and persist the database
+  - a more clean solution would probably be to at each time we recurse, a derivation that does the evaluation
     and outputs a JSON that can later be read
 
-how this relates to bombon: 
+how this relates to bombon:
 - bombon uses  a more coarse grained approach
-- this builds a metadata "database" i.e. is two pass 
+- this builds a metadata "database" i.e. is two pass
 - see the corresponding nix code in ./nix
 -}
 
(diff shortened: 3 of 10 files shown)


hack/bin/Sbom.hs
```diff
@@ -0,0 +1,349 @@
...
+      sbom'manufacture = "wire",
+      sbom'supplier = Nothing,
+      sbom'licenses = ["AGPL-3.0-or-later"]
+    }
+
+-- FUTUREWORK(mangoiv): we can also have
+--
+-- - qualifiers: extra qualifying data for a package such as an OS, architecture, a distro, etc. Optional and type-specific.
+-- - subpath: extra subpath within a package, relative to the package root. Optional.
+-- - use heuristics based approach to finding original repositories for packages, e.g. pkg:hackage....
```
emil-wire, line 158: @MangoIV would supporting more "original repositories" essentially mean code like this?

````...
      | any (maybe False (T.isInfixOf "hackage.haskell.org")) meta.urls = "hackage"
      | any (maybe False (T.isInfixOf "crates.io")) meta.urls = "cargo"
````

Comment (MangoIV) [reply hack/bin/Sbom.hs:158]: yeah I mean if we expect to have the urls not contain any cross references to these domains even though they don't originate for that, that's correct, yes.

Comment (MangoIV) [reply hack/bin/Sbom.hs:158]: the issue is that this approach is horrendously slow, I will use something like AhoCorasick if we want to match on more repos.

File: hack/bin/Sbom.hs [hunk 2]
```diff
@@ -0,0 +1,349 @@
...
+      <*> val
+      .: "type"
+      <*> val
+      .: "urls"
+      <*> val
+      .: "version"
+      <*> pure Proxy
+      <*> pure Proxy
+  parseJSON invalid = typeMismatch "Object" invalid
+
```
MangoIV, line 119: I hate that ormolu does this

fisx: you can use `(`, `)` to make this behave better.  but then hlint might complain about that...  :roll_eyes:

hack/bin/Sbom.hs
```diff
@@ -0,0 +1,349 @@
...
+{-# LANGUAGE OverloadedRecordDot #-}
+{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE RecordWildCards #-}
+{-# LANGUAGE StrictData #-}
+{-# LANGUAGE UndecidableInstances #-}
+{-# LANGUAGE ViewPatterns #-}
+{-# OPTIONS_GHC -Wall #-}
+{-# OPTIONS_GHC -Wno-incomplete-uni-patterns #-}
+
+module Sbom where
```
fisx, line 17:
```suggestion
-- | see https://github.com/nikstur/bombon for more background.
module Sbom where
```

hack/bin/bombon.hs
```diff
@@ -1,18 +1,23 @@
-#!/usr/bin/env -S nix -Lv run github:wireapp/ghc-flakr/99fe5a331fdd37d52043f14e5c565ac29a30bcb4
+#!/usr/bin/env -S nix -Lv run github:wireapp/ghc-flakr/6311bb166bf835d4a587fe1661b86c9a1426f212
 {-# LANGUAGE DataKinds #-}
+{-# LANGUAGE LambdaCase #-}
+{-# OPTIONS_GHC -Wall #-}
```
fisx, line 4:
```suggestion
{-# OPTIONS_GHC -Wall #-}

-- See https://github.com/nikstur/bombon for more background.
```

it would also be helpful to explain here in one or two lines what bombon.hs and Sbom.hs do, and how they relate to each other.

hack/bin/Sbom.hs
MangoIV: I’ve used do in the past but I was fearing that people would be mad if I did 😅

hack/bin/bombon.hs
MangoIV: Good idea, thank you. I have written some paragraphs and I guess it makes sense to put them somewhere in the rep.

MangoIV: Do you have an idea where I could put a loose markdown file that I could link to? 😅

(I’m guessing the confluence doesn’t make much sense as it isn’t public)

elland: There's a docs folder but I'd inline it if possible, or at least add something in-place. If the write up is too long, you can link it here instead.

Makefile
```diff
@@ -546,11 +546,12 @@ helm-template-%: clean-charts charts-integration
...
 # Ask the security team for the `DEPENDENCY_TRACK_API_KEY` (if you need it)
+# changing the directory is necessary because of some quirkiness of how 
+# runhaskell / ghci behaves (it doesn't find modules that aren't in the same 
+# directory as the script that is being executed)
 .PHONY: upload-bombon
 upload-bombon:
-	nix build -f nix wireServer.allLocalPackagesBom -o "bill-of-materials.$(HELM_SEMVER).json"
-	./hack/bin/bombon.hs -- \
-		--bom-filepath "./bill-of-materials.$(HELM_SEMVER).json" \
+	cd ./hack/bin	&& ./bombon.hs -- \
```
supersven, line 554:
```suggestion
	cd ./hack/bin && ./bombon.hs -- \
```
`<TAB>` isn't needed here.

changelog.d/5-internal/SEC-596
```diff
@@ -0,0 +1 @@
+use new script Sbom.hs to generate the wire-server sbom
```
supersven, line 1:
```suggestion
Create a new script (`Sbom.hs`) to generate the wire-server sbom (bill of material) file.
```

nix/wire-server.nix
```diff
@@ -442,6 +442,29 @@ let
...
+    src = pkgs.fetchFromGitHub {
+      owner = "interlynk-io";
+      repo = "sbomqs";
+      rev = "v${version}";
+      hash = "sha256-+y7+xi+E8kjGUjhIRKNk6ogcQMP+Dp39LrL66B1XdrQ=";
+    };
+
+    vendorHash = "sha256-V6k7nF2ovyl4ELE8Cqe/xjpmPAKI0t5BNlssf41kd0Y=";
+
+    #ldflags = [ "${envLdflags}" ];
```
supersven, line 458: Do we still need this comment?

nix/default.nix
```diff
@@ -7,7 +7,7 @@ let
       # All wire-server specific packages
       (import ./overlay.nix)
       (import ./overlay-docs.nix)
-      (self: super: { lib = super.lib // (import sources.bombon).lib.${super.system}; })
+      # (self: super: { lib = super.lib // (import sources.bombon).lib.${super.system}; })
```
supersven, line 10: Is this commented out? Do we still need it?

nix/all-toplevel-derivations.nix
```diff
@@ -0,0 +1,54 @@
+{ lib
+, pkgSet
+, fn
+, recursionDepth
+, keyFilter
+, ...
+}:
+let
```
supersven, line 8: I think a comment about what this is supposed to do and a rough description of the algorithm would be good.

nix/all-toplevel-derivations.nix
```diff
@@ -0,0 +1,54 @@
...
+          in
+          builtins.deepSeq meta (
+            builtins.trace ("reached leaf: " + toString set)
+              (
+                if meta.success
+                then [ meta.value ]
+                else builtins.trace "package didn't evaluate" [ ]
+              )
+          )
+        else if depth >= recursionDepth
```
supersven, line 32: Hm, is this correct? Shouldn't you always check on the recursion depth (and not just as fallback)? :thinking:

nix/all-toplevel-derivations.nix
```diff
@@ -0,0 +1,54 @@
...
+, pkgSet
+, fn
+, recursionDepth
+, keyFilter
+, ...
+}:
+let
+  go = depth: set':
+    let
+      canSet = builtins.tryEval set';
```
supersven, line 11: Hm, I've got troubles to understand why you've chosen the name `canSet` here. :thinking: 

Wouldn't this be better named the `evaluatablePkgsSet` or `evaluatableSet`? :thinking:

hack/bin/Sbom.hs
```diff
@@ -0,0 +1,376 @@
...
+{-# LANGUAGE RecordWildCards #-}
+{-# LANGUAGE StrictData #-}
+{-# LANGUAGE UndecidableInstances #-}
+{-# LANGUAGE ViewPatterns #-}
+{-# OPTIONS_GHC -Wall #-}
+{-# OPTIONS_GHC -Wno-incomplete-uni-patterns #-}
+
+{-
+- the only place that has the data we need about the package is the evaluated nix code, i.e. before
+  writhing the derivation; this is where we have `meta` and friends to get the data we need
```
supersven, line 19:
```suggestion
  writing the derivation; this is where we have `meta` and friends to get the data we need
```

hack/bin/Sbom.hs
```diff
@@ -0,0 +1,376 @@
...
+- people don't always have a complete package set but more commonly are hand assembling things; we need
+  to give the possibility to build meta "databases" from package sets
+- we need to trace which dependencies are missing when querying the meta database against them
+- collecting the meta also poses some issue
+  - nixpkgs is not a tree, but a more general graph
+  - it also not a DAG but it has loops
+  - this means more specifically that we cannot without care recurse into it
+  - even if we only recurse very shallowly, we soon start running out of memory, this means we probably need
+    to do some on the fly filtering by "actual" dependencies
+  - this is similarly and issue, because it means that for every package we have to evaluate the entirety
```
supersven, line 41:
```suggestion
  - this is similarly an issue, because it means that for every package we have to evaluate the entirety
```

nix/wire-server.nix
```diff
@@ -442,6 +442,29 @@ let
     };
   }) cabal-plan;
 
+  sbomqs = pkgs.buildGoModule rec {
```
supersven, line 445: Wouldn't this better reside as dedicated `.nix` file in `nix/pkgs`? :thinking:

nix/pkg-info.nix
```diff
@@ -0,0 +1,60 @@
...
+  # if thing is a list, map the function, else apply f to thing and return a singleton of
+  # it
+  mapOrSingleton = f: x:
+    if isList x
+    then map f x
+    else [ (f x) ];
+
+  # things to save from the src attr (the derivation that was created by a fetcher)
+  srcInfo = {
+    urls = (pkg.src.urls or (trc "package didn't have src or url" pkg [ ])) ++ [ (pkg.src.url or null) ];
```
supersven, line 20: `trc "package didn't have src or url" pkg [ ]` isn't the empty list one parameter too much? :thinking:

supersven: (As far as I understand `trc` takes two arguments.)

hack/bin/Sbom.hs
```diff
@@ -0,0 +1,376 @@
...
+  mconcat
+    [ "pkg:",
+      repo,
+      "/",
+      fromMaybe (runIdentity meta.outPath) meta.name,
+      maybe "" ("@" <>) meta.version
+    ]
+  where
+    repo
+      | any (maybe False (T.isInfixOf "hackage.haskell.org")) meta.urls = "hackage"
```
supersven, line 197: Is this really the expected package format: `pkg:hackage/svens-package@.0.0.42`?
I mean: You won't find anything with this URL on the internet, right? :thinking:

hack/bin/Sbom.hs
```diff
@@ -0,0 +1,376 @@
...
+      <*> do val .: "type"
+      <*> do val .: "urls"
+      <*> do val .: "version"
+      <*> pure Proxy
+      <*> pure Proxy
+  parseJSON invalid = typeMismatch "Object" invalid
+
+type SBom = Map Text (SBomMeta Identity)
+
+type MetaDB = Map Text (SBomMeta Proxy)
```
supersven, line 149: Unfortunately, I don't understand why we need to have `SBomMeta` with `outPath :: Proxy Text` and `outPath :: Proxy Text` :thinking: 

Could you please explain why `outPath :: Text` isn't good enough?

MangoIV: Because it’s just not there before. Proxy a is just a weird way of saying Const ()

nix/wire-server.nix
MangoIV: Good idea. Will do that.

nix/all-toplevel-derivations.nix
MangoIV: Yuh the naming is awful. I will switch to one of your proposed ones 😅

nix/all-toplevel-derivations.nix
MangoIV: But we only recur in the else branch? I’m missing the issue with this.

nix/default.nix
MangoIV: Nope it should be removed. Good catch.

nix/wire-server.nix
MangoIV: Nope. This is just something generated by the thing i used to generate this derivation. (It’s pretty cool, it’s called nix-init and it works fairly well for go programs)

emil-wire (approved):

Makefile
MangoIV: i have no idea, how a `<TAB>` got here, I don't even have this character in my editor. (tabs get replaced by two spaces...)

nix/all-toplevel-derivations.nix
MangoIV: done

nix/pkg-info.nix
MangoIV: yes, it takes two arguments but it returns a function! ;)

hack/bin/Sbom.hs
MangoIV: should I replace `Proxy` with `Const ()` to make it more clear?

MangoIV:
@supersven thank you for your review; I have tried to address your issues, would you happen to have time to take one more glance at it? :3 

Thank you!

supersven:
> @supersven thank you for your review; I have tried to address your issues, would you happen to have time to take one more glance at it? :3
> 
> Thank you!

@MangoIV I'm at it :smile:

hack/bin/Sbom.hs
```diff
@@ -0,0 +1,376 @@
...
+        ]
+
+buildMetaDB :: [Meta] -> MetaDB
+buildMetaDB = foldMap \MkSBomMeta {..} -> [(drvPath, MkSBomMeta {..})]
+
+discoverSBom :: FilePath -> MetaDB -> IO SBom
+discoverSBom outP metaDb = do
+  canonicalOutP <- canonicalizePath =<< getSymbolicLinkTarget outP
+  info <- pathInfo canonicalOutP
+  let go :: (Text, (Text, [Text])) -> IO SBom -> IO SBom
```
supersven, line 270:
```suggestion
  let go :: PathInfo -> IO SBom -> IO SBom
```

I'm not sure if we really need `PathInfo` as type synonym. However, as it's there, we can use it consequently :)

hack/bin/Sbom.hs
```diff
@@ -0,0 +1,376 @@
...
+import System.Process
+
+data License = MkLicense
+  { id :: Maybe Text,
+    name :: Maybe Text
+  }
+  deriving stock (Eq, Ord, Show, Generic)
+  deriving anyclass (FromJSON, ToJSON)
+
+sadSbomMeta :: Text -> Text -> [Text] -> SBomMeta Identity
```
supersven, line 96:
```suggestion
emptySbomMeta :: Text -> Text -> [Text] -> SBomMeta Identity
```
or
```suggestion
notFoundSbomMeta :: Text -> Text -> [Text] -> SBomMeta Identity
```
?

Hm, could be that I don't get the name :thinking: However, if this is named "sad" because the search algorithm is sad because it didn't find anything, then I would say be can name this better :smiley:

MangoIV: yeah, it's because it didn't find anything and the information it can provide is sad :(

nix/default.nix
```diff
@@ -79,7 +78,6 @@ let
         pkgs.entr
       ] ++ docsPkgs;
     };
-  mls-test-cli = pkgs.mls-test-cli;
-  rusty-jwt-tools = pkgs.rusty-jwt-tools;
```
supersven, line 83: Do we inherit the `rusty-jwt-tools` implicitly? Or, are they not required anymore?

MangoIV: it's unused.

MangoIV: image

nix/wire-server.nix
```diff
@@ -462,9 +462,35 @@ let
...
+    let
+      mk = pkg:
+        import ./pkg-info.nix {
+          inherit pkg;
+          inherit (pkgs) lib hostPlatform writeText;
+        };
+      out = import ./all-toplevel-derivations.nix {
+        inherit (pkgs) lib;
+        fn = mk;
+        recursionDepth = 2;
```
supersven, line 479: Is there a special reason to choose `2` here? In this case we should document it.

nix/all-toplevel-derivations.nix
supersven: Ah, I see.

nix/wire-server.nix
MangoIV: yes, after three it doesn't fit into memory.

supersven: Does it still contain everything Emil needs to see?

supersven: This :arrow_up: is actually my last main concern. Once we got this answered, I'll approve this PR.

MangoIV: yes; there's only a couple of bootstrapping things missing; I explicitly put in package sets that we need.

supersven: Cool :sunglasses:

supersven (approved):

nix/wire-server.nix
MangoIV: I mean cool is debatable. Nix not being able to work with the equivalent of 3 MB of JSON is concerning

PR #3947, opened 2024-03-13
smallstep-accomp: Resolve proxy target on request (#3946)

supersven:
Usually, proxy targets are resolved when nginx is started. This can lead to strange behavior if the target either doesn't exist (yet) or the DNS entry changes while nginx is running.

This little trick with the indirection via a variable should trigger the lookup(s) while nginx is running. The default behavior of the `resolver` directive is to update the target according to its TTL in the configured DNS server.

N.B. this a cherry-pick / backport of https://github.com/wireapp/wire-server/pull/3946
Ticket: https://wearezeta.atlassian.net/browse/WPB-6822

## Checklist


jschumacher-wire (approved):
Looks good.

PR #3957, opened 2024-03-15
clarify the smallstep-accomp values in the readme

mastaab:
we added annotations with https://github.com/wireapp/wire-server/pull/3955 but forgot to add this to the readme.md

PR #3958, opened 2024-03-15
clarify the smallstep-accomp values in the helm chart readme (#3957)

mastaab:
upstream from q1-2024


supersven (approved):
The cherry pick should work as well :+1:

PR #3959, opened 2024-03-15
WPB-6717 ES Credentials

battermann:
## Checklist

### DIFF
--- a/docs/src/developer/reference/config-options.md
+++ b/docs/src/developer/reference/config-options.md
@@ -853,3 +853,25 @@ accessible to services (and not the private key.)
 
 The corresponding Cassandra options are described in Cassandra's documentation:
 [client_encryption_options](https://cassandra.apache.org/doc/stable/cassandra/configuration/cass_yaml_file.html#client_encryption_options)
+
+## Configure Elasticsearch basic authentication
+
+When the wire backend is configured to work against a custom Elasticsearch instance, it may be desired to enable basic authentication for the internal communication between the wire backend and the ES instance. To do so the Elasticsearch credentials can be set in wire-server's secrets as follows:
+
+```yaml
+brig:
+  secrets:
+    elasticsearch:
+      username: elastic
+      password: changeme
+```
+
+In some cases an additional Elasticsearch instance is needed (e.g. for index migrations). To configure credentials for the additional ES instance add the secret as follows:
+
+```yaml
+brig:
+  secrets:
+    elasticsearchAdditional:
+      username: elastic
+      password: changeme
+```
--- a/docs/src/developer/reference/config-options.md
+++ b/docs/src/developer/reference/config-options.md
@@ -861,13 +861,13 @@ When the Wire backend is configured to work against a custom Elasticsearch insta
 ```yaml
 brig:
   secrets:
-    elasticsearch: |-
+    elasticsearch:
       username: elastic
       password: changeme
 
 elasticsearch-index:
   secrets:
-    elasticsearch: |-
+    elasticsearch:
       username: elastic
       password: changeme
 ```
@@ -877,7 +877,7 @@ In some cases an additional Elasticsearch instance is needed (e.g. for index mig
 ```yaml
 brig:
   secrets:
-    elasticsearchAdditional: |-
+    elasticsearchAdditional:
       username: elastic
       password: changeme
 ```
--- a/docs/src/developer/reference/config-options.md
+++ b/docs/src/developer/reference/config-options.md
@@ -856,7 +856,7 @@ The corresponding Cassandra options are described in Cassandra's documentation:
 
 ## Configure Elasticsearch basic authentication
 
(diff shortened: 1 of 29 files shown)


battermann:
I am running into the problem currently, that our docker ES version does not support the security features: https://www.elastic.co/guide/en/elasticsearch/reference/current/security-settings.html which are necessary to properly test basic auth. (xpack security is included in versions >= 7)

Newer version, however, are not compatible with our current ES code, namely types cannot be provided in put mapping requests, unless the include_type_names query parameter is set to true. So far, I haven’t found a way to pass this query parameter with our current version of bloodhound.

smatting:
Shouldn't this be reflected in the docs somewhere? For example in the brig chart it's not obvious in what format `.elasticsearch` has to be, or that this parameter even exists.

battermann:
> Shouldn't this be reflected in the docs somewhere? For example in the brig chart it's not obvious in what format `.elasticsearch` has to be, or that this parameter even exists.

Absolutely, I'll add it.

docs/src/developer/reference/config-options.md
```diff
@@ -853,3 +853,25 @@ accessible to services (and not the private key.)
 
 The corresponding Cassandra options are described in Cassandra's documentation:
 [client_encryption_options](https://cassandra.apache.org/doc/stable/cassandra/configuration/cass_yaml_file.html#client_encryption_options)
+
+## Configure Elasticsearch basic authentication
+
+When the wire backend is configured to work against a custom Elasticsearch instance, it may be desired to enable basic authentication for the internal communication between the wire backend and the ES instance. To do so the Elasticsearch credentials can be set in wire-server's secrets as follows:
```
supersven, line 859:
```suggestion
When the Wire backend is configured to work against a custom Elasticsearch instance, it may be desired to enable basic authentication for the internal communication between the wire backend and the ES instance. To do so the Elasticsearch credentials can be set in wire-server's secrets as follows:
```

I think it's (uppercase) `Wire`, because it's a name.

libs/types-common/src/Util/Options.hs
```diff
@@ -79,7 +79,12 @@ urlPort u = do
 makeLenses ''AWSEndpoint
 
 newtype FilePathSecrets = FilePathSecrets FilePath
-  deriving (Eq, Show, FromJSON)
+  deriving (Eq, Show, FromJSON, IsString)
+
+initCredentials :: (FromJSON a) => FilePathSecrets -> IO a
```
supersven, line 84: I'm wondering if we couldn't make potential callers' lifes easier by using `MonadIO` here :thinking:

services/brig/test/integration/Index/Create.hs
```diff
@@ -84,7 +86,9 @@ testCreateIndexWhenPresent brigOpts = do
     Left e -> fail $ "Invalid ES URL: " <> show esURL <> "\nerror: " <> show e
     Right esURI -> do
       indexName <- ES.IndexName . Text.pack <$> replicateM 20 (Random.randomRIO ('a', 'z'))
-      ES.withBH HTTP.defaultManagerSettings (ES.Server esURL) $ do
+      mgr <- liftIO $ HTTP.newManager HTTP.defaultManagerSettings
+      let bEnv = (ES.mkBHEnv (ES.Server esURL) mgr) {ES.bhRequestHook = ES.basicAuthHook (ES.EsUsername "elastic") (ES.EsPassword "changeme")}
```
supersven, line 90: This block exists three times. We could consider to extract a function to reduce duplication.

supersven (approved):
LGTM :+1: Only a few nitpicks.

services/brig/src/Brig/Options.hs
```diff
@@ -623,7 +623,11 @@ data Settings = Settings
     setOAuthRefreshTokenExpirationTimeSecsInternal :: !(Maybe Word64),
     -- | The maximum number of active OAuth refresh tokens a user is allowed to have.
     -- use `setOAuthMaxActiveRefreshTokens` as the getter function which always provides a default value
-    setOAuthMaxActiveRefreshTokensInternal :: !(Maybe Word32)
+    setOAuthMaxActiveRefreshTokensInternal :: !(Maybe Word32),
+    -- | Elasticsearch credentials
+    setElasticsearchCredentials :: !(Maybe FilePathSecrets),
+    -- | Credentials for additional ES index (maily used for migrations)
+    setElasticsearchAdditionalCredentials :: !(Maybe FilePathSecrets)
```
akshaymankar, line 630: Why is this not part of `ElasticSearchOpts`?

services/brig/test/resources/elasticsearch-credentials.yaml
```diff
@@ -0,0 +1,2 @@
+username: "elastic"
+password: "changeme"
```
akshaymankar, line 2: It feels wrong to have this file because for RabbitMQ we have environment variables. We should be consistent.

akshaymankar: Discussed in chat. there are other secrets which are configured in files. So we're already not consistent.
To avoid spending more time on this, we can make sure the helm charts are configured in a way that ensures we can change our minds on this later.

akshaymankar (approved):

PR #3962, opened 2024-03-18
New shell script hack to create session tokens (and logout if you like).

fisx:
This happened as a side effect of looking into https://wearezeta.atlassian.net/browse/WPB-7073, but I've always wanted to have something very basic like this for getting session tokens.

### DIFF
--- a/hack/bin/get-session-token
+++ b/hack/bin/get-session-token
@@ -1,0 +1,39 @@
+#!/usr/bin/env bash
+
+#set -x
+set -e -o pipefail
+
+WIRE_COOKIEJAR=/tmp/get-session-token.cookiejar
+
+#WIRE_HOST=https://prod-nginz-https.wire.com
+#WIRE_USER='...'
+#WIRE_PASSWORD='...'
+
+# run eg. './create_test_team_admins.sh -c', and fill user and password from the output:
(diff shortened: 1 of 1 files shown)


hack/bin/get-session-token
```diff
@@ -0,0 +1,44 @@
+#!/usr/bin/env bash
+
+#set -x
+set -e -o pipefail
+
+WIRE_COOKIEJAR=/tmp/get-session-token.cookiejar
+
+#WIRE_HOST=https://prod-nginz-https.wire.com
+#WIRE_USER='...'
+#WIRE_PASSWORD='...'
```
elland, line 10: A bit more documentation would be nice.

fisx: it's a hack.  it's meant to be used by editing the code, so you'll have to read and understand most of it anyway.  so adding docs would feel to me like adding noise.  what exactly do you have in mind?  maybe links to the swagger docs of the end-points called here?

elland: Thinking about this some more, how about an example flow we could copy paste to save ourselves even more time? 😄

fisx:
CI can be considered green; this PR has nothing to do with integration tests.

elland (approved):

PR #3965, opened 2024-03-19
Revert http2 pin.

fisx:
Removed a pin that was introduced in 8f823f2517f, as it only made things worse.

## Checklist

 - ~~[ ] Add a new entry in an appropriate subdirectory of `changelog.d`~~

### DIFF
--- a/nix/haskell-pins.nix
+++ b/nix/haskell-pins.nix
@@ -141,7 +141,7 @@ let
     };
 
     # We forked to add a handler for the ConnectionIsClosed signal
-    # since it was threated as a halting exception instead of a 
+    # since it was threated as a halting exception instead of a
     # clean exit.
     http2 = {
       src = fetchgit {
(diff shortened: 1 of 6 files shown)


nix/haskell-pins.nix
```diff
@@ -141,15 +141,17 @@ let
...
-    };
+    #
+    # (fisx 2024-03-19: took out this pin; the fork was even less stable.)
+    #http2 = {
+    #  src = fetchgit {
+    #    url = "https://github.com/wireapp/http2";
+    #    rev = "9cad270779bbcd9e6297b9ff05a4a7eb83bca069";
+    #    sha256 = "sha256-c+PzfZZUxo/tE8oH1ZmKwbUMAq34kGD1OoBWorNLG38=";
+    #  };
+    #};
```
akshaymankar, line 154: Just delete it. Igor is working on things anyway.

elland (approved):
No need to comment out that which can be deleted ;)

PR #3966, opened 2024-03-19
[WPB-5687] more legalhold tests

MangoIV:
https://wearezeta.atlassian.net/browse/WPB-5687
it appears I had overlooked to port some of the legalhold tests that were still failing

- [x] legalhold status in user profile
- [x] "Cannot create conversation with both LH activated and non-consenting users" testCannotCreateGroupWithUsersInConflict,
- [x] image

## Checklist


MangoIV:
I have completely removed `testGroupConvInvitationHandlesLHConflicts` as it was a dejavu, this is tested multiple times across the file, I will port it if I missed something and this tests otherwise untested behaviour but I think this is the same as in other tests.

services/galley/test/integration/API/Teams/LegalHold.hs
```diff
@@ -395,39 +394,6 @@ testNoConsentCannotBeInvited = do
...
-  -- team without legalhold
-  (peer :: UserId, teamPeer) <- createBindingTeam
-  peer2 <- (^. Team.userId) <$> addUserToTeam peer teamPeer
-
-  do
-    postConnection userLHNotActivated peer !!! const 201 === statusCode
-    void $ putConnection peer userLHNotActivated Conn.Accepted <!! const 200 === statusCode
-
-    postConnection userLHNotActivated peer2 !!! const 201 === statusCode
-    void $ putConnection peer2 userLHNotActivated Conn.Accepted <!! const 200 === statusCode
```
MangoIV, line 551: do we know why this works? I don't get it

mdimjasevic: I am not sure what exactly you're talking about, but maybe I can help with this. Can you clarify what you mean by "this"?

MangoIV: We’re trying to post a conversation to Alice after only connecting Bob with Charlie and Debora (see new code, there it’s more readable), the new code errors saying Bob is not connected (to Alice) which makes sense but I don’t get why it doesn’t in the old test? Perhaps in the old test team members were somehow implicitly connected already?

MangoIV:
I don't think we need a further changelog entry as this is a follow up PR.

services/galley/test/integration/API/Teams/LegalHold.hs
MangoIV: the reason is, that even though `alice` and `bob` are automatically connected, because they're in a team, you still have to specify the team the conversation belongs to, s.t. they count as connected; I don't quite understand what happens if bob and alice each join a conversation that was created of a foreign team owner, will this always error with "not connected"

akshaymankar (commented):
Why do all these tests need a dynamic backend?

integration/test/Notifications.hs
```diff
@@ -115,6 +116,13 @@ isMemberJoinNotif n = fieldEquals n "payload.0.type" "conversation.member-join"
 isConvLeaveNotif :: MakesValue a => a -> App Bool
 isConvLeaveNotif n = fieldEquals n "payload.0.type" "conversation.member-leave"
 
+isConvLeaveNotifWithLeaver :: (MakesValue user, MakesValue a) => user -> a -> App Bool
+isConvLeaveNotifWithLeaver user n =
+  fieldEquals n "payload.0.type" "conversation.member-leave"
+    &&~ (n %. "payload.0.data.user_ids.0") `isEqual` (user %. "id")
+
+-- fieldEquals n "payload.0.data.user_ids" users
+
```
akshaymankar, line 125:
```suggestion
```

integration/test/Test/LegalHold.hs
```diff
@@ -14,6 +14,7 @@
 --
 -- You should have received a copy of the GNU Affero General Public License along
 -- with this program. If not, see <https://www.gnu.org/licenses/>.
+{-# OPTIONS_GHC -Wwarn #-}
```
akshaymankar, line 17:
```suggestion
```

akshaymankar (commented):
I also see that you left some of the LH tests in the old test suite, why so?

MangoIV:
> Why do all these tests need a dynamic backend?

because otherwise the rest of the services don't see the mock LH service; perhaps they don't need it though, might just be a skill issue on my behalf :3 do you have a pointer at some place where this is done without dynamic backend? 

> I also see that you left some of the LH tests in the old test suite, why so?

they don't flake.

akshaymankar:
> because otherwise the rest of the services don't see the mock LH service; perhaps they don't need it though, might just be a skill issue on my behalf :3 do you have a pointer at some place where this is done without dynamic backend?

If you try to access `wire-server-integration-integration.<namespace>.pod.cluster.local` from brig/galley it should get to any port listening in the integration test pod.

> they don't flake.

Given these flakes are for galley trying to talk to the dynamic service that we spin up, I think they might flake anyway. So, instead of going through this cycle of moving them in batches, IMO it is best if we move all of them.

MangoIV:
it's still 1400 lines of code to port, this is work that gets tedious after a while and it's hard to concentrate for such a long time, I would prefer if I could do it in batches, ordered by urgency...

services/galley/test/integration/API/Teams/LegalHold.hs
mdimjasevic: I'm sorry, I still don't quite understand the issue. Can you write a test for your question and see what you get in response?

In general, all users within a team are connected by default, without them having to directly establish pair-wise connections.

Your rewrite of the test isn't quite faithful. A regular group conversation is not the same as a team conversation; they follow different rules. In your tests you create a plain group conversation, whereas you should be setting a team as the owner of a conversation. You can do that by setting the `team` parameter of the `defProteus` argument.

akshaymankar:
> it's still 1400 lines of code to port, this is work that gets tedious after a while and it's harder for concentration, I think this work might benefit from doing in batches ordered by urgency.

Sure, IMO the urgency is the same, which tests fail is not consistent across runs. If you find it easier to do it in batches, that's fine too. I think we shouldn't close the ticket until all of these are migrated away.

MangoIV:
That’s fine, let’s keep it open. These tests should be the ones that are in the hundreds of failures, not the ones that only failed recently.

services/galley/test/integration/API/Teams/LegalHold.hs
MangoIV: yes, that was the issue and that's why it didn't work, I found the issue yesterday morning; the problem was that the old combinator set that option kinda implicitly. :)

integration/test/Notifications.hs
```diff
@@ -177,9 +189,36 @@ isUserConnectionNotif = notifTypeIsEqual "user.connection"
...
 isConnectionNotif status n =
-  (&&)
-    <$> nPayload n %. "type" `isEqual` "user.connection"
-    <*> nPayload n %. "connection.status" `isEqual` status
+  nPayload n %. "type" `isEqual` "user.connection"
+    &&~ nPayload n %. "connection.status" `isEqual` status
+
+-- NB:
+-- (&&) <$> (print "hello" *> pure False) <*> fail "bla" === _|_
+-- runMaybeT $  (lift (print "hello") *> MaybeT (pure Nothing)) *> lift (fail "bla") === pure Nothing
```
fisx, line 197:
```suggestion
  -- NB:
  -- (&&) <$> (print "hello" *> pure False) <*> fail "bla" === _|_
  -- runMaybeT $  (lift (print "hello") *> MaybeT (pure Nothing)) *> lift (fail "bla") === pure Nothing
  nPayload n %. "type" `isEqual` "user.connection"
    &&~ nPayload n %. "connection.status" `isEqual` status
```

integration/test/Notifications.hs
```diff
@@ -177,9 +189,36 @@ isUserConnectionNotif = notifTypeIsEqual "user.connection"
...
+-- | make Bool lazy
+liftBool :: Functor f => f Bool -> BoolT f
+liftBool = MaybeT . fmap (bool Nothing (Just ()))
+
+-- | make Bool strict
+unliftBool :: Functor f => BoolT f -> f Bool
+unliftBool = fmap isJust . runMaybeT
+
+-- | lazy (&&)
+(&&~) :: App Bool -> App Bool -> App Bool
```
fisx, line 208: it's fine to introduce this locally, but you probably shouldn't export it, then?

integration/test/Test/LegalHold.hs
```diff
@@ -605,172 +593,315 @@ testLHNoConsentBlockOne2OneConv
...
+              LegalholderIsAdmin -> (alice, tidAlice, bob, "wire_member")
+              BothAreAdmins -> (alice, tidAlice, bob, "wire_admin")
+              PeerIsAdmin -> (bob, tidBob, alice, "wire_member")
+
+        let createConv = defProteus {qualifiedUsers = [invitee], newUsersRole = inviteeRole, team = Just tidInviter}
+        postConversation inviter createConv `bindResponse` \resp -> do
+          resp.json %. "members.self.conversation_role" `shouldMatch` "wire_admin"
+          case admin of
+            BothAreAdmins -> resp.json %. "members.others.0.conversation_role" `shouldMatch` "wire_admin"
+            _ -> resp.json %. "members.others.0.conversation_role" `shouldMatch` "wire_member"
```
fisx, line 791:
```suggestion
          resp.json %. "members.others.0.conversation_role" `shouldMatch` case admin of
            LegalholderIsAdmin -> "wire_member"
            BothAreAdmins -> "wire_admin"
            PeerIsAdmin -> "wire_member"

```

integration/test/Test/LegalHold.hs
```diff
@@ -605,172 +593,315 @@ testLHNoConsentBlockOne2OneConv
...
-            for_ [(alice, alicec, lastNotifAlice'), (bob, bobc, lastNotifBob')] \(user, client, lastNotif) -> do
-              awaitNotification user client (Just lastNotif) isUserConnectionNotif >>= \notif ->
-                notif %. "payload.0.connection.status" `shouldMatchOneOf` ["sent", "pending", "accepted"]
-
-            sendMessageFromBobToAlice (const [alicec]) \resp -> do
-              resp.status `shouldMatchInt` 201
-
-            sendMessageFromBobToAlice (\device -> [device]) \resp -> do
-              resp.status `shouldMatchInt` 412
+testLHNoConsentRemoveFromGroup :: GroupConvAdmin -> App ()
```
fisx, line 771:
```suggestion
-- | If a member of an existing conversation is assigned a LH device, users are removed from
-- the conversation until policy conflicts are resolved.  as to who gets to stay: admins win
-- against members, and local members win against remote members.
testLHNoConsentRemoveFromGroup :: GroupConvAdmin -> App ()
```

fisx (approved):
looks good!  please consider my nit-picks before you merge.

integration/test/Notifications.hs
MangoIV: I think it makes sense to export it, it may be needed elsewhere? I would also encourage users to use this instead of `liftA2 (&&)`

MangoIV:
I have applied your suggestions @fisx with some slight deviations from your doc proposals.

integration/test/Notifications.hs
fisx: fine, but then please don't export it from `/integration/test/Notifications.hs`.

MangoIV: Alright!

PR #3978, opened 2024-04-02
[WPB-5990] Consolidate logic of building a user profile

akshaymankar:
`connectedProfile` and `publicProfile` did the same thing no matter what, then
the logic to display email or not was part of Brig itself. It makes more sense
if we make the email logic part of a single function which makes the profile.
This way we could write tests for it.

Not having these two thing removes the need to get relations between requesting
user and the viewed users, which saves 1 DB query per viewed user.

https://wearezeta.atlassian.net/browse/WPB-5990

## Checklist

 - [x] ~Add a new entry in an appropriate subdirectory of `changelog.d`~ No changelog, refactor


pcapriotti (approved):
Looks good.

PR #3982, opened 2024-04-05
early exit when deleting already gone client

stefanwire:
Ticket: https://wearezeta.atlassian.net/browse/WPB-7416

## Checklist

### DIFF
--- a/services/galley/src/Galley/API/Clients.hs
+++ b/services/galley/src/Galley/API/Clients.hs
@@ -47,6 +47,7 @@ import Polysemy
 import Polysemy.Error
 import Polysemy.Input
 import Polysemy.TinyLog qualified as P
+import System.Logger.Message
 import Wire.API.Conversation hiding (Member)
 import Wire.API.Federation.API
 import Wire.API.Federation.API.Galley
@@ -111,12 +112,20 @@ rmClientH ::
   UserId ::: ClientId ->
   Sem r Response
 rmClientH (usr ::: cid) = do
-  lusr <- qualifyLocal usr
-  let nRange1000 = toRange (Proxy @1000) :: Range 1 1000 Int32
-  firstConvIds <- Query.conversationIdsPageFrom lusr (GetPaginatedConversationIds Nothing nRange1000)
-  goConvs nRange1000 firstConvIds lusr
-
-  E.deleteClient usr cid
+  clients <- E.getClients [usr]
+  if (cid `elem` clientIds usr clients)
+    then do
+      lusr <- qualifyLocal usr
+      let nRange1000 = toRange (Proxy @1000) :: Range 1 1000 Int32
+      firstConvIds <- Query.conversationIdsPageFrom lusr (GetPaginatedConversationIds Nothing nRange1000)
+      goConvs nRange1000 firstConvIds lusr
+      E.deleteClient usr cid
+    else
+      P.debug
+        ( field "user" (idToText usr)
+            . field "client" (clientToText cid)
+            . msg (val "rmClientH: client already gone")
+        )
   pure empty
   where
     goConvs :: Range 1 1000 Int32 -> ConvIdsPage -> Local UserId -> Sem r ()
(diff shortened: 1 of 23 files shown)


services/galley/src/Galley/Cassandra/Client.hs
```diff
@@ -60,13 +62,24 @@ eraseClients user = retry x5 (write Cql.rmClients (params LocalQuorum (Identity
...
   Sem (ClientStore ': r) a ->
   Sem r a
 interpretClientStoreToCassandra = interpret $ \case
-  GetClients uids -> embedClient $ lookupClients uids
-  CreateClient uid cid -> embedClient $ updateClient True uid cid
-  DeleteClient uid cid -> embedClient $ updateClient False uid cid
-  DeleteClients uid -> embedClient $ eraseClients uid
-  UseIntraClientListing -> embedApp . view $ options . settings . intraListing
+  GetClients uids -> do
+    debug (msg (val "ClientStore.GetClients"))
```
supersven, line 72: I'm wondering if we shouldn't reduce the duplication and transform this

```
debug (msg (val "ClientStore.GetClients"))
```

into e.g.

```
logEffect "ClientStore.GetClients"

logEffect :: Member TinyLog r -> String -> Sem r a
logEffect = debug . msg . val
```
:thinking: 

However, this is likely a matter of taste. (I was able to read your approach, too :wink: )

services/galley/src/Galley/Cassandra/SubConversation.hs
```diff
@@ -124,20 +126,40 @@ listSubConversations cid = do
       )
 
 interpretSubConversationStoreToCassandra ::
-  Members '[Embed IO, Input ClientState] r =>
+  ( Member (Embed IO) r,
```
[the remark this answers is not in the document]
stefanwire: No rule, but the compiler won't find redundant constraints in `Members`. I think by now it's common sense to expand these `Members` constraints to individual `Member` constraints and not introduce new `Members` ones.

services/galley/src/Galley/Cassandra/Client.hs
stefanwire: Can do.

services/galley/src/Galley/Cassandra/SubConversation.hs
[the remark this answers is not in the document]
supersven: Good to know. I wasn't aware of this.

services/galley/src/Galley/Cassandra/Client.hs
stefanwire: Done.

supersven (approved):
Rubberstamping for Zulu (The PR to `master` has already been approved: https://github.com/wireapp/wire-server/pull/3985)

services/galley/src/Galley/API/Clients.hs
```diff
@@ -111,12 +112,20 @@ rmClientH ::
...
+      lusr <- qualifyLocal usr
+      let nRange1000 = toRange (Proxy @1000) :: Range 1 1000 Int32
+      firstConvIds <- Query.conversationIdsPageFrom lusr (GetPaginatedConversationIds Nothing nRange1000)
+      goConvs nRange1000 firstConvIds lusr
+      E.deleteClient usr cid
+    else
+      P.debug
+        ( field "user" (idToText usr)
+            . field "client" (clientToText cid)
+            . msg (val "rmClientH: client already gone")
```
mdimjasevic, line 127: While at it, why not rename the function too? We have the "H" suffix (I guess "H" for handler) in the old `wai` routes days. We have been dropping them in the migration to Servant.

```suggestion
            . msg (val "rmClient: client already gone")
```

mdimjasevic:
Could this debugging effort be merged to a non-release branch and then deploy that branch to environments of interest for debugging?

stefanwire:
> Could this debugging effort be merged to a non-release branch and then deploy that branch to environments of interest for debugging?

Following the discussion on _Backend_, I will merge the logs. They should not be viewed as efforts to debug a single issue, but rather seen as an effort to debug future issues on live systems. This should remain in place until we settle on a better way to achieve the same or better observability of our services.

services/galley/src/Galley/API/Clients.hs
stefanwire: I would generally agree, but cosmetics should be done on `develop` only. I try to keep the PRs added to two branches to the minimal change with maximal code overlap.

PR #3988, opened 2024-04-09
[chore] enforce Imports in brig

MangoIV:
- make import Imports implicit using cabal mixins
- replace qualified imports of Import with qualified imports of Prelude
- remove warnings on unnecessary packages
- remove brig Paths_ module

## Checklist

### DIFF
--- a/services/brig/brig.cabal
+++ b/services/brig/brig.cabal
@@ -19,6 +19,13 @@ extra-source-files:
   docs/swagger.md
 
 common common-all
+  -- NOTE(mangoiv): we cannot use the -Wunused-packages here because 
+  -- cabal doesn't recognise that we need `base` to be able  to 
+  -- hide it
+  mixins: 
+    , base hiding (Prelude)
+    , imports (Imports as Prelude)
+    , imports
   default-language:   GHC2021
   ghc-options:
     -O2 -Wall -Wincomplete-uni-patterns -Wincomplete-record-updates
@@ -51,7 +58,6 @@ common common-all
     MultiParamTypeClasses
     MultiWayIf
     NamedFieldPuns
-    NoImplicitPrelude
     NumericUnderscores
     OverloadedLabels
     OverloadedRecordDot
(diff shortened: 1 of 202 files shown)


MangoIV:
To do the thing: 

`find . -name "*.hs" -type f | xargs sed -i -e 's/import Imports qualified/import Prelude qualified/'` 
`find . -name "*.hs" -type f | xargs sed -i -e 's/import Imports hiding/import Prelude hiding/'`
`find . -name "*.hs" -type f | xargs sed -i -e '/^import Imports$/d'`
**MUST COME AFTER THE ABOVE:**
`find . -name "*.hs" -type f | xargs sed -i -e 's/Imports/Prelude/'`

MangoIV:
> i don't see this as a thing we need, especially if we pay for it in any way. one way in which we pay is that it makes things less transparent and straight-forward. more hidden moving parts. i don't like hidden moving parts.

I think I don't understand this; what is hidden and moving? this is a more or less well-known, documented feature of `cabal` https://cabal.readthedocs.io/en/3.4/cabal-package.html#pkg-field-mixins , the dev can just use the `Prelude` as if it was *the* `Prelude`. Obviously it would have to be documented in the developer documentation.

MangoIV:
> no strong opinions here, probably not a terrible thing? :)

we don't use it, it doesn't make sense to generate it, it just needs  additional compile time.

fisx:
> > i don't see this as a thing we need, especially if we pay for it in any way. one way in which we pay is that it makes things less transparent and straight-forward. more hidden moving parts. i don't like hidden moving parts.
> 
> I think I don't understand this; what is hidden and moving? this is a more or less well-known, documented feature of `cabal` https://cabal.readthedocs.io/en/3.4/cabal-package.html#pkg-field-mixins , the dev can just use the `Prelude` as if it was _the_ `Prelude`. Obviously it would have to be documented in the developer documentation.

maybe i didn't use the best words, let me try again.  every time somebody touches the code that you've been touching here, whether they are new to the code base or have been here forever, will cause some extra mental load to figure out what you did there.  it's not the default, and it is less straight-forward than `-XNoImplicitPreude` and `import Imports`.  it's more time people have to spend loading in concepts.

writing something always imposes a cost on the reader.

this is all about a general mindset i'm trying to explain (and follow, not always successfully): don't make changes just because it's easy to make those changes, and don't under-estimate the workload for the team to follow those changes.  maybe we can find a better example for what i mean at some point in the future?

and if people want to chime in and tell me this is all fine and go ahead magnus i won't fight!  :)

fisx:
way too much text, sorry!

MangoIV:
thank you, very sensible :)

PR #4000, opened 2024-04-15
[WPB-8712] use treefmt for everything

MangoIV:
https://wearezeta.atlassian.net/browse/WPB-8712
## Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -181,7 +181,7 @@ services: init install
 
 # formats everything according to treefmt rules
 # this may take a while (5 minutes) on first run but should be instant on 
-# any subsequent run
+# any subsequent run except after you have changed files.
 .PHONY: format
 format:
 	treefmt
--- a/Makefile
+++ b/Makefile
@@ -129,6 +129,7 @@ devtest:
 .PHONY: sanitize-pr
 sanitize-pr:
 	./hack/bin/generate-local-nix-packages.sh
+	make hlint-inplace-pr
 	make format
 	make git-add-cassandra-schema
 	@git diff-files --quiet -- || ( echo "There are unstaged changes, please take a look, consider committing them, and try again."; exit 1 )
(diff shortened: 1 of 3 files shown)


MangoIV:
I think I want to readd some `make hlint-apply` rule or somth

changelog.d/5-internal/WPB-8712
```diff
@@ -0,0 +1 @@
+move formatting and linting of haskell files to treefmt, remove some of the now unneded rules
```
fisx, line 1:
```suggestion
move formatting and linting of haskell files to treefmt, remove some of the now unneeded rules
```

Makefile
```diff
@@ -192,35 +179,27 @@ check-local-nix-derivations: regen-local-nix-derivations
 services: init install
 	$(MAKE) -C services/nginz
 
-# formats all Haskell files (which don't contain CPP)
+# formats everything according to treefmt rules
+# this may take a while (5 minutes) on first run but should be instant on 
+# any subsequent run
```
fisx, line 184:
```suggestion
# any subsequent run except after you have changed files.
```

...  right?  this is nice, though.

Makefile
```diff
@@ -129,9 +129,8 @@ devtest:
 .PHONY: sanitize-pr
 sanitize-pr:
 	./hack/bin/generate-local-nix-packages.sh
-	make formatf
 	make hlint-inplace-pr
-	make hlint-check-pr  # sometimes inplace has been observed not to do its job very well.
+	make format
```
fisx, line 133:
```suggestion
	make format
	make formatc  # (at some point in the past, this has caught ormolu screwing up with inplace)
```

this *might* be fixed now with the `--check-inplace` argument, but can we be certain?

but if you prefer, we can remove this and see if it happens again.

MangoIV:
there's one tradeoff to adding hlint to this is that it will make the initial lint very slow (will take ~5 minutes) but after then the cache kicks in and only changed files are re-linted.

fisx (approved):

Makefile
MangoIV: `--check-idempotence`? i have removed `formatc`, i think...

Makefile
MangoIV: yeah, depends on what instant means to you ;D

Makefile
MangoIV: i was hoping that ormolu by now is idempotent or at least the idempotency check works reliably :eyes:

PR #4002, opened 2024-04-15
Optimize getting all feature configs

akshaymankar:
https://wearezeta.atlassian.net/browse/WPB-8713

## Checklist


services/galley/src/Galley/Cassandra/TeamFeatures.hs
```diff
@@ -1,3 +1,5 @@
+{-# LANGUAGE TemplateHaskell #-}
```
elland, line 1: Note: let's not forget the cleanup.

PR #4004, opened 2024-04-16
Increase 2FA timeout to unflake test

battermann:
The test `testLoginVerify6DigitExpiredCodeFails` turned out to flake during a local run, which could be reproduced by running it 100x using `replicateM_` (in sequence).
Setting the verification timeout from 1 sec to 2 sec seemed to remove the flakiness.
We could just merge this or spend some more time trying to understand better what is going on exactly?


supersven (approved):
LGTM :+1: 

Thanks a lot for taking care of flaky tests!

PR #4014, opened 2024-04-23
[WPB-8764] Backport changes from `q1-2024` to `develop` (Apr 23, 2024)

mdimjasevic:
This includes:

- #3896
- #3905
- #3918
- #3923
- #3927
- #3952
- #3957

Trackedy by https://wearezeta.atlassian.net/browse/WPB-8764.

## Checklist

 - [x] Add a new entry in an appropriate subdirectory of `changelog.d`
 - [x] Read and follow the [PR guidelines](https://docs.wire.com/developer/developer/pr-guidelines.html)

PR #4015, opened 2024-04-24
Remove sftd from list of published charts

smatting:
Followup to https://github.com/wireapp/wire-server/pull/4014


akshaymankar (approved):
:rocket:

PR #4024, opened 2024-04-26
WPB-8702 Retrieve in database users that can only login with Phone Numbers & SMS to login

battermann:
## Checklist

### DIFF
--- a/tools/db/phone-users/default.nix
+++ b/tools/db/phone-users/default.nix
@@ -29,7 +29,6 @@ mkDerivation {
   libraryHaskellDepends = [
     aeson
     aeson-pretty
-    base
     bytestring
     cassandra-util
     conduit


cabal.project
```diff
@@ -48,6 +48,7 @@ packages:
   , tools/db/inconsistencies/
   , tools/db/migrate-sso-feature-flag/
   , tools/db/move-team/
+  , tools/db/phone-users/
```
akshaymankar, line 51: Please add this block somewhere below:

```cabal
package phone-users
  ghc-options: -Werror
```

changelog.d/5-internal/WPB-8702
```diff
@@ -0,0 +1 @@
+Tool to determine number of phone-only users
```
akshaymankar, line 1:
```suggestion
Add tool to determine number of phone-only users
```

tools/db/phone-users/src/PhoneUsers/Lib.hs
```diff
@@ -0,0 +1,172 @@
...
+import Options.Applicative
+import PhoneUsers.Types
+import qualified System.IO as SIO
+import qualified System.Logger as Log
+import Wire.API.Team.Feature (FeatureStatus (FeatureStatusDisabled, FeatureStatusEnabled))
+import Wire.API.User (AccountStatus (Active))
+
+lookupClientsLastActiveTimestamps :: ClientState -> UserId -> IO [Maybe UTCTime]
+lookupClientsLastActiveTimestamps client u = do
+  runClient client $ runIdentity <$$> retry x1 (query selectClients (params LocalQuorum (Identity u)))
```
akshaymankar, line 39: I think we shouldn't use quorum reads here, they are more expensive and its not terribly important that this data is super up to date.

pcapriotti (approved):

tools/db/phone-users/src/PhoneUsers/Lib.hs
```diff
@@ -0,0 +1,172 @@
...
+  where
+    selectClients :: PrepQuery R (Identity UserId) (Identity (Maybe UTCTime))
+    selectClients = "SELECT last_active from clients where user = ?"
+
+readUsers :: ClientState -> ConduitM () [UserRow] IO ()
+readUsers client =
+  transPipe (runClient client) (paginateC selectUsersAll (paramsP LocalQuorum () 200) x5)
```
akshaymankar, line 46: You probably want a higher page size, smaller this number is, slower the whole operation will be. I think we typically use 1000.

tools/db/phone-users/src/PhoneUsers/Lib.hs
```diff
@@ -0,0 +1,172 @@
...
+    select =
+      "select conference_calling from team_features where team_id = ?"
+
+process :: Maybe Int -> ClientState -> ClientState -> IO Result
+process limit brigClient galleyClient =
+  runConduit $
+    readUsers brigClient
+      .| Conduit.mapM (\chunk -> SIO.hPutStr stderr "." $> chunk)
+      .| Conduit.concat
+      .| (maybe (Conduit.filter (const True)) Conduit.take limit)
```
akshaymankar, line 67: Why the limit?

tools/db/phone-users/src/PhoneUsers/Lib.hs
```diff
@@ -0,0 +1,172 @@
...
+  where
+    select :: PrepQuery R (Identity TeamId) (Identity FeatureStatus)
+    select =
+      "select conference_calling from team_features where team_id = ?"
+
+process :: Maybe Int -> ClientState -> ClientState -> IO Result
+process limit brigClient galleyClient =
+  runConduit $
+    readUsers brigClient
+      .| Conduit.mapM (\chunk -> SIO.hPutStr stderr "." $> chunk)
```
akshaymankar, line 65: Does this print a dot for every user? In past we just logged the number of retrieved users.

tools/db/phone-users/src/PhoneUsers/Lib.hs
```diff
@@ -0,0 +1,172 @@
...
+      .| Conduit.mapM (getUserInfo brigClient galleyClient)
+      .| Conduit.foldMap infoToResult
+
+getUserInfo :: ClientState -> ClientState -> UserRow -> IO UserInfo
+getUserInfo brigClient galleyClient ur = do
+  if not $ isCandidate
+    then pure NoPhoneUser
+    else do
+      -- should we give C* a little break here and add a small threadDelay?
+      threadDelay 200
```
akshaymankar, line 77: Its probably not necessary, but please run it on staging and check. 0.2 ms is barely a break anyway.

tools/db/phone-users/src/PhoneUsers/Lib.hs
```diff
@@ -0,0 +1,172 @@
...
+              Just tid -> do
+                isPaying <- isPayingTeam galleyClient tid
+                pure $
+                  if isPaying
+                    then ActiveTeamUser Free
+                    else ActiveTeamUser Paid
+            putStrLn ""
+            putStrLn $ "active phone user found: " <> show apu
+            print ur
+            putStrLn $ "last active time stamps: " <> show lastActiveTimeStamps
```
akshaymankar, line 100: Why not use the logger that was created? This way things can even be printed as JSON then we can easily answer more questions about the users than just the final `Result`.

akshaymankar (commented):
I think we shouldn't use quorum reads. Apart from that other comments I made are minor.

tools/db/phone-users/src/PhoneUsers/Lib.hs
battermann: for test runs :)

battermann: it's optional

tools/db/phone-users/src/PhoneUsers/Lib.hs
battermann: dot for every page, I found this useful to get some feedback on the progress of the app

PR #4027, opened 2024-04-29
Add more fingerprint verification tests for bots

akshaymankar:
https://wearezeta.atlassian.net/browse/WPB-6350

## Checklist


integration/test/Test/Bot.hs
```diff
@@ -0,0 +1,142 @@
...
+ - signed the cert. (a cert can only have one signatory, a CA or self)
+ -
+ - this test succeeds if the signature verification fails (because it's not self signed), however,
+ - even if Brig starts to do signature verification, the test would still succeed, because brig
+ - doesn't know (or trust) the CA, anyway, even if it does signature verification.
+ -
+ - For this test to make sense, we would have to make sure that the brig we're testing against
+ - *would* trust the CA, *if* it did verification, because only in that case it would now succeed
+ - with verification and not return a "PinInvalidCert" error.
+ - -}
```
MangoIV, line 40: due to this, I'm a bit stuck on this PR, most importantly I don't know how I would distinguish between failing due to not knowing the CA and failing due to not doing CA verification without modifying Brig. 

This additionally implies, that I do know how to distinguish the cases of 
- root CA
- intermediate CA
- leaf cert (which is conceptually the same as above)

akshaymankar: > For this test to make sense, we would have to make sure that the brig we're testing against
 *would* trust the CA, *if* it did verification, because only in that case it would now succeed
 with verification and not return a "PinInvalidCert" error.

If brig starts doing some CA verification tomorrow, the people will hopefully write a test to make sure that that verification works. This test failing would just start meaning that the unknown CAs are not trusted. So, IMO this test is good as it is.

akshaymankar: Maybe we can call this test `testBotUnknownRootCA` anyway.

MangoIV: This test would start failing if the CAs were made trusted *and* if the signature verification would be done, at the same time, so if we implemented this and then went back to modify the test such that brig starts trusting these CAs, then it would start failing.

MangoIV:
how many "hi ci"s can we score until it finally doesn't flake?

MangoIV:
this PR is truly cursed.

integration/test/Testlib/ModService.hs
```diff
@@ -204,10 +204,10 @@ startDynamicBackend resource beOverrides = do
     setLogLevel =
       def
         { sparCfg = setField "saml.logLevel" ("Warn" :: String),
-          brigCfg = setField "logLevel" ("Warn" :: String),
+          brigCfg = setField "logLevel" ("Debug" :: String),
           cannonCfg = setField "logLevel" ("Warn" :: String),
           cargoholdCfg = setField "logLevel" ("Warn" :: String),
-          galleyCfg = setField "logLevel" ("Warn" :: String),
+          galleyCfg = setField "logLevel" ("Debug" :: String),
           gundeckCfg = setField "logLevel" ("Warn" :: String),
```
MangoIV, line 211: remove

MangoIV:
I have a near 100% flake rate, what’s wrong with this PR? Is it possible that this test somehow influences others? There’s one think that takes particularly long, is it possible that this is why other tests fail?

stefanwire (approved):

MangoIV (approved):
Looks good to me but please add a link to the PR in the Haskell pins. Thank you!

integration/test/Testlib/Certs.hs
```diff
@@ -0,0 +1,146 @@
...
+publicKeyToString :: RSA.PublicKey -> String
+publicKeyToString = toPem . pubKeyToPEM . PubKeyRSA
+
+-- | order: publickey, private key
+keyPairToString :: RSAKeyPair -> (String, String)
+keyPairToString = bimap publicKeyToString privateKeyToString
+
+-- | the minimum key size is hard coded to be 256 bytes (= 2048 bits)
+--
+-- TODO(mangoiv): key generation takes an (actually multiple) eternities
```
akshaymankar, line 41: Has this TODO been resolved?

nix/haskell-pins.nix
```diff
@@ -266,6 +266,15 @@ let
...
     };
+
+    HsOpenSSL = {
+      src = fetchgit {
+        url = "https://github.com/wireapp/HsOpenSSL";
+        rev = "2d7afef71ecdc2edd579dde072b75674da1fb730";
+        sha256 = "sha256-wLjDOMokKELVH30bYNyDAYT3ux3pdMq4sIeMZLQOSG8=";
+      };
+    };
+
```
akshaymankar, line 277: There's been a new release of HsOpenSSL already with this! So, this could be moved to hackagePins.

https://github.com/haskell-cryptography/HsOpenSSL/pull/91#issuecomment-2161667646

nix/manual-overrides.nix
```diff
@@ -64,6 +64,8 @@ hself: hsuper: {
   # (these are fine)
   # -----------------
   # Make hoogle static to reduce size of the hoogle image
+  cryptostore = hlib.addBuildDepends (hlib.dontCheck (hlib.appendConfigureFlags hsuper.cryptostore [ "-fuse_crypton" ]))
+    [ hself.crypton hself.crypton-x509 hself.crypton-x509-validation ];
```
akshaymankar, line 68: The comment above this belongs to the line below this.

libs/ssl-util/src/Ssl/Util.hs
```diff
@@ -33,16 +31,17 @@ module Ssl.Util
...
 import Data.Dynamic (fromDynamic)
 import Data.Time.Clock (getCurrentTime)
 import Imports
 import Network.HTTP.Client.Internal
 import OpenSSL.BN (integerToMPI)
-import OpenSSL.EVP.Digest (Digest, digestLBS)
-import OpenSSL.EVP.PKey (SomePublicKey, toPublicKey)
+import OpenSSL.EVP.Digest (digestLBS)
+import OpenSSL.EVP.Internal
+import OpenSSL.EVP.PKey
```
akshaymankar, line 44: I think this should be reverted given there was no actual code change in this file, looks like we depend on some Internal module for no reason.

integration/test/Test/LegalHold.hs
```diff
@@ -109,11 +109,11 @@ testLHMessageExchange ::
   TaggedBool "clients1New" ->
   TaggedBool "clients2New" ->
   App ()
+-- We used to throw LegalholdConflictsOldClients if clients didn't have LH capability, but we
+-- don't do that any more because that broke things.
+-- Related: https://github.com/wireapp/wire-server/pull/4056
```
akshaymankar, line 114: Weird place to put a comment, can we leave it as it was please?

integration/test/Testlib/Certs.hs
stefanwire: Yes, by using constant p and q for RSA keys. Should be fine for tests.

akshaymankar: Let's delete the comment then :)

PR #4042, opened 2024-05-07
[WPB-5360] update the nginz nginx module to support nginx 1.26

MangoIV:
https://wearezeta.atlassian.net/browse/WPB-5360
turns out it was not that hard after all, I could trace back our compile issue to a refactoring done in this commit
https://github.com/nginx/nginx/commit/3aef1d693f3cc431563a7e6a6aba6a34e5290f03?diff=split&w=1 and that was quite easy to fix

## Checklist

### DIFF
--- a/nix/overlay.nix
+++ b/nix/overlay.nix
@@ -76,10 +76,10 @@ self: super: {
   };
 
   nginz = (super.nginx.overrideAttrs rec {
-    version = "1.22.1";
+    version = "1.26.0";
     src = super.fetchurl {
       url = "https://nginx.org/download/nginx-${version}.tar.gz";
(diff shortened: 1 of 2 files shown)


nix/overlay.nix
```diff
@@ -76,10 +76,10 @@ self: super: {
   };
 
   nginz = (super.nginx.overrideAttrs rec {
-    version = "1.22.1";
+    version = "1.26.0";
     src = super.fetchurl {
       url = "https://nginx.org/download/nginx-${version}.tar.gz";
-      hash = "sha256-nrszOp6CuVKs0+K0rrHU/2QG9ySRurbNn+afDepzfzE=";
+      hash = "sha256-0ubIQ51sbbUBXY6qskcKtSrvhae/NjGCh5l34IQ3BJc=";
```
MangoIV, line 82: this override can also go in the next nixpkgs update. yay.

PR #4070, opened 2024-05-29
WPB-8824 MLS E2EID CRL proxy flag

battermann:
## Checklist


docs/src/developer/reference/config-options.md
```diff
@@ -334,6 +334,10 @@ When a client first tries to fetch or renew a certificate, they may need to logi
 
 The client enrolls using the Automatic Certificate Management Environment (ACME) protocol [RFC 8555](https://www.rfc-editor.org/rfc/rfc8555.html). The `acmeDiscoveryUrl` parameter must be set to the HTTPS URL of the ACME server discovery endpoint for this team. It is of the form "https://acme.{backendDomain}/acme/{provisionerName}/discovery". For example: `https://acme.example.com/acme/provisioner1/discovery`.
 
+`useProxyOnMobile` is an optional field. If `true`, mobile clients should use the CRL proxy. If missing, null or false, mobile clients should not use the CRL proxy.
+
+`crlProxy` contains the URL to the CRL proxy. (Not that this field is optional in the server config, but mandatory when the team feature is updated via the team feature API.)
```
stefanwire, line 339:
```suggestion
`crlProxy` contains the URL to the CRL proxy. (Note that this field is optional in the server config, but mandatory when the team feature is updated via the team feature API.)
```

docs/src/understand/team-feature-settings.md
```diff
@@ -94,6 +94,10 @@ When a client first tries to fetch or renew a certificate, they may need to logi
 
 The client enrolls using the Automatic Certificate Management Environment (ACME) protocol [RFC 8555](https://www.rfc-editor.org/rfc/rfc8555.html). The `acmeDiscoveryUrl` parameter must be set to the HTTPS URL of the ACME server discovery endpoint for this team. It is of the form "https://acme.{backendDomain}/acme/{provisionerName}/discovery". For example: `https://acme.example.com/acme/provisioner1/discovery`.
 
+`useProxyOnMobile` is an optional field. If `true`, mobile clients should use the CRL proxy. If missing, null or false, mobile clients should not use the CRL proxy.
+
+`crlProxy` contains the URL to the CRL proxy. (Not that this field is optional in the server config, but mandatory when the team feature is updated via the team feature API.)
```
stefanwire, line 99:
```suggestion
`crlProxy` contains the URL to the CRL proxy. (Note that this field is optional in the server config, but mandatory when the team feature is updated via the team feature API.)
```

stefanwire (approved):
Why did we choose to have a `Bool` and a `Maybe URL`? Is there a use-case in which a `Just URL` has to be remembered by Cassandra even when the `Bool` is `False`? Otherwise, the `Maybe URL` could implicitly cover the `Bool`'s value by means of `isJust`.

battermann:
> Why did we choose to have a `Bool` and a `Maybe URL`? Is there a use-case in which a `Just URL` has to be remembered by Cassandra even when the `Bool` is `False`? Otherwise, the `Maybe URL` could implicitly cover the `Bool`'s value by means of `isJust`.

It is only a `Maybe` to make the API backwards compatible.

PR #4071, opened 2024-05-29
[WPB-8943] ghc 9.4 -> 9.6, nixpkgs bump

MangoIV:
https://wearezeta.atlassian.net/browse/WPB-8943

## Checklist


MangoIV:
TODO: point hsaml2 to upstream again (PR go merged already)

MangoIV:
this is a chonky chonk of a PR

libs/wire-api/src/Wire/API/User/IdentityProvider.hs
```diff
@@ -208,13 +207,17 @@ instance ToSchema RawIdPMetadata where
...
+      $ NamedSchema (Just "IdPMetadataInfo")
+      $ mempty
+      & properties
+      .~ properties_
+        & minProperties
+      ?~ 1
+        & maxProperties
+      ?~ 1
+        & type_
+      ?~ OpenApiObject
```
akshaymankar, line 220: Another place where something broke with fixity.

libs/wire-api/src/Wire/API/OAuth.hs
```diff
@@ -164,11 +167,14 @@ data OAuthClient = OAuthClient
...
-        <*> (.name) .= field "application_name" schema
-        <*> (.redirectUrl) .= field "redirect_url" schema
+    object "OAuthClient"
+      $ OAuthClient
+      <$> (.clientId)
+      .= field "client_id" schema
+        <*> (.name)
+      .= field "application_name" schema
+        <*> (.redirectUrl)
+      .= field "redirect_url" schema
```
akshaymankar, line 177: Something broke with fixity here

nix/haskell-pins.nix
```diff
@@ -66,24 +66,25 @@ let
...
     cryptobox-haskell = {
       src = fetchgit {
         url = "https://github.com/wireapp/cryptobox-haskell";
         rev = "7546a1a25635ef65183e3d44c1052285e8401608";
-        sha256 = "0dgizj1kc135yzzqdf5l7f5ax0qpvrr8mxvg7s1dbm01cf11aqzn";
+        hash = "sha256-9mMVgmMB1NWCPm/3inLeF4Ouiju0uIb/92UENoP88TU=";
       };
     };
 
+    # FIXME(mangoiv): should be merged https://github.com/wireapp/saml2-web-sso/pull/86
```
akshaymankar, line 82: Approved now.

nix/manual-overrides.nix
```diff
@@ -34,30 +37,50 @@ hself: hsuper: {
...
+      rec {
+        # we don't care for the patches for mingw and for 1.0.19
+        patches = [ ];
+        version = "1.0.18";
+        src = fetchurl {
+          url = "https://download.libsodium.org/libsodium/releases/${old.pname}-${version}.tar.gz";
+          hash = "sha256-b1BEkLNCpPikxKAvybhmy++GItXfTlRStGvhIeRmNsE=";
+        };
+      }
+    )));
```
akshaymankar, line 51: Why not put this in haskell-pins.nix?

akshaymankar: Also, it would be helpful to document what problems 1.0.19 causes.

MangoIV: > Why not put this in haskell-pins.nix?

because the library itself doesn't change, it's just the actual `libsodium` that gets pinned - `libsodium` the c library has moved on upstream but many (all) haskell libraries are only compatible with 1.0.18.

nix/haskell-pins.nix
MangoIV: thank you <3

libs/wire-api/src/Wire/API/OAuth.hs
MangoIV: investigating

fisx:
good thing we have [/tools/rebase-onto-formatter.sh](https://github.com/wireapp/wire-server/blob/5ef2be02eea0b0147cd5bfea65b7a1ba4e4ed058/tools/rebase-onto-formatter.sh)...

for me the "semantic" commit contains quite a lot of noise, too.  did something go wrong there?

MangoIV:
all files that are in the semantic parts actually contain changes, it is possible that they are reformatted because my editor just does that :) It shouldn't be that bad though ;)

MangoIV:
> good thing we have [/tools/rebase-onto-formatter.sh](https://github.com/wireapp/wire-server/blob/5ef2be02eea0b0147cd5bfea65b7a1ba4e4ed058/tools/rebase-onto-formatter.sh)...

I don't use this tool, I don't really have a need for it :3

MangoIV:
you just do `git rebase -i HEAD^^^` then move the commit you want to ammend to the semantic part up and squash it into it.

nix/haskell-pins.nix
```diff
@@ -276,19 +267,15 @@ let
...
+    http2 = {
+      version = "4.1.4";
+      sha256 = "sha256-r4Bu0vourKMkBO1cPeJVszSbAqHopmkv9EeTHcaTfuo=";
     };
-    # dependency of hoogle
-    safe = {
-      version = "0.3.20";
-      sha256 = "sha256-PGwjhrRnkH8cLhd7fHTZFd6ts9abp0w5sLlV8ke1yXU=";
+
+    # warp is not compatible with
```
elland, line 275: missing end of sentence?

elland (approved):
Only one comment issue came to mind.

PR #4079, opened 2024-06-03
[WPB-9076] Galley OOM killed

elland:
https://wearezeta.atlassian.net/browse/WPB-9076


MangoIV:
so cool! thank you for documenting this!

docs/src/developer/developer/profiling.md
```diff
@@ -0,0 +1,40 @@
+# Profiling Wire
+
+This documents one possible avenue to profile Wire Server, be it as a whole or individual services such as Galley or Brig.
+
+For that we are using `cabal.project.local` to define that we want to build with profiling on. added at the top of the file:
+
+```haskell
+profiling: True
+benchmarks: True
```
akshaymankar, line 9: Why do we need benchmarks?

elland, line 9:
```suggestion
```

MangoIV:
https://well-typed.com/blog/2023/03/prof-late/

can we use speedscope for this? I have looked in the PR and the RTS flags you applied look a bit different - they use `+RTS -pj`

MangoIV:
@akshaymankar do we want to change the RTS options in the cabal file or do we want to add a modification to the nix builds such that we can deploy a profile build only where we want?

PR #4080, opened 2024-06-04
WPB-9469 retrieve the count of users that have have not consented to data usage permissions on web

battermann:
## Checklist


battermann:
We don't need to keep this in our code base, so I will close it.

PR #4083, opened 2024-06-05
coturn: Add IP configuration options

smatting:
This PR adds configuration options in the `coturn` helm chart for settings
- `relay-ip`
- `external-ip`
- `prometheus-ip`
- `federation-listening-ip`

It also adds a `__COTURN_HOST_IP__` template variable which will be substitued by the host ip of the `k8s` node that is running coturn.

This PR is backwards-compatible to `values.yaml` files written for version of this helm charts without this PR: new variables have defaults that match their previous hardcoded value.

## Checklist

### DIFF
--- a/charts/coturn/templates/configmap-coturn-conf-template.yaml
+++ b/charts/coturn/templates/configmap-coturn-conf-template.yaml
@@ -30,7 +30,9 @@ data:
     listening-ip={{ default "__COTURN_EXT_IP__" .Values.coturnTurnListenIP }}
     listening-port={{ .Values.coturnTurnListenPort }}
     relay-ip={{ default "__COTURN_EXT_IP__" .Values.coturnTurnRelayIP }}
+    {{- if .Values.coturnTurnExternalIP }}
     external-ip={{ default "__COTURN_EXT_IP__" .Values.coturnTurnExternalIP }}
+    {{- end }}
     realm=dummy.io
     no-stun-backward-compatibility
     secure-stun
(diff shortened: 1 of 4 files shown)


charts/coturn/values.yaml
```diff
@@ -26,9 +26,11 @@ coturnTurnListenPort: 3478
 coturnMetricsListenPort: 9641
 coturnTurnTlsListenPort: 5349
 
-# If you need to specify which IP Coturn should bind to.
-# This will typically be the IP of the kubenode.
-# coturnTurnListenIP: "182.168.22.133"
+# coturnTurnListenIP: "1.2.3.4" # can also be __COTURN_EXT_IP__, __COTURN_POD_IP__,__COTURN_HOST_IP__
```
supersven, line 29: Some comment explaining the difference between `...ListenIp`, `...RelayIp` and `...ExternalIp` would be helpful.

smatting: Wouldn't that just reproduce the [docs fo coturn](https://github.com/coturn/coturn/tree/master/docs)?

charts/coturn/templates/configmap-coturn-conf-template.yaml
```diff
@@ -29,14 +29,17 @@ data:
     ## turn, stun.
     listening-ip={{ default "__COTURN_EXT_IP__" .Values.coturnTurnListenIP }}
     listening-port={{ .Values.coturnTurnListenPort }}
-    relay-ip=__COTURN_EXT_IP__
+    relay-ip={{ default "__COTURN_EXT_IP__" .Values.coturnTurnRelayIP }}
+    {{- if .Values.coturnTurnExternalIP }}
+    external-ip={{ default "__COTURN_EXT_IP__" .Values.coturnTurnExternalIP }}
+    {{- end }}
```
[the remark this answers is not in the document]
smatting: The motivation behind this was to keep backwards-compatible with values.yaml made for the pre-PR version of the helm chart: `external-ip` was not in the configmap before

supersven (approved):

charts/coturn/values.yaml
supersven: Yes, but IMHO otherwise admins would need to understand `coturn` itself and not just look at our Helm charts. :thinking:

PR #4085, opened 2024-06-06
Proposal: Define metrics using `Prometheus.unsafeRegister` instead of having the metrics-core wrapper

akshaymankar:
The `metrics-core` package provided a wrapper around metrics which allowed for metrics reuse using 3 `IORef (HashMap Text <metric-type>)` values. This makes the code much safer than the default way of using the Promehtheus library but also doesn't allow for having Labels and makes using metrics clunkier.  It is possible to make these lookups type level and get safe Labels, but it just feels wrong to put so much work into this given metrics are not _super_ important part of the business logic and they are never reused from different places. It is easier to the just use `unsafeRegister` once on a top level binding.

This PR is a result of me getting carried away and seeing how long it would take me to rework everything, it took ~4 hours. It is completely fine to throw it away and my feeling won't be hurt if we decide this is not a good idea (please, really don't worry about it).

More work on this PR:
- [x] ~Replace all dots in the metric names with underscores, they aren't supported by prometheus and get replaced at runtime~ I tried this, the dots are helpful for people who read the code, so let's leave them.
- [x] Verify that all the top level binds have a `NOINLINE` pragma

https://wearezeta.atlassian.net/browse/WPB-9680

## Checklist


fisx (approved):
if we forget to inline a metric, it will always only remember the last update, and the internal STVar in prometheus will grow lineary with the number of updates. but i'd expect this to get exposed during testing & QA.

otherwise good! much nicer code :) looking forward to rewriting it again and moving it to wire-subsystems.

PR #4088, opened 2024-06-08
Add weeder (dead code elimination tool) to dev environment

fisx:
https://wearezeta.atlassian.net/browse/WPB-9667

i've run weeder from a PR locally with success (>1400 weeds, but many of them are false positives).  now we only need to add that patch to our nix envs.  (we should probably not run it on ci for now, until we have fine-tuned weeder.toml and eliminated all the legitimate weed.)

the purpose of this PR is:

- add weeder to the dev env so people can run it and fine-tune it

the purpose of this PR is *NOT*:

- add weeder to the CI (see https://github.com/NorfairKing/weeder-nix, but different PR)
- remove any weed
- have a complete list of roots in the weeder.toml
- push all the buttons that make weeder as strict as possible

## Checklist


nix/manual-overrides.nix
```diff
@@ -70,4 +70,5 @@ hself: hsuper: {
   types-common-journal = hlib.addBuildTool hsuper.types-common-journal protobuf;
   wire-api = hlib.addBuildTool hsuper.wire-api mls-test-cli;
   wire-message-proto-lens = hlib.addBuildTool hsuper.wire-message-proto-lens protobuf;
+  # weeder = hlib.addBuildTool hsuper.weeder;
```
elland, line 73: dead code?

fisx: not done yet!  :)

fisx:
This doesn't install the pinned version of weeder, but the one from hackage (2.7).

cabal.project
```diff
@@ -62,117 +62,7 @@ packages:
...
-package wire-api-federation
-    ghc-options: -Werror
-package wire-message-proto-lens
-    ghc-options: -Werror
-package wire-subsystems
-    ghc-options: -Werror
-package zauth
-    ghc-options: -Werror
-package fedcalls
+package *
```
fisx, line 65: bad

Makefile
```diff
@@ -155,7 +151,11 @@ ghcid:
 
 # Used by CI
 .PHONY: lint-all
-lint-all: formatc hlint-check-all check-local-nix-derivations treefmt-check
+lint-all: formatc hlint-check-all check-local-nix-derivations treefmt-check # weeder (does not work on CI yet)
+
+.PHONY: weeder
+weeder:
+	time weeder -N || echo -e '\n\n*** make sure you have added this to your cabal.project.local and make clean before building wire-server:\n\npackage *\n  ghc-options: -fwrite-ide-info\n\n'
```
akshaymankar, line 158:
```suggestion
	weeder -N || echo -e '\n\n*** make sure you have added this to your cabal.project.local and make clean before building wire-server:\n\npackage *\n  ghc-options: -fwrite-ide-info\n\n'
```

Also it seems like we're gonna print this message on any failure, my guess is that it will definitely be confusing in future.

nix/overlay.nix
```diff
@@ -89,6 +89,12 @@ self: super: {
     ];
   };
 
+  haskellPackages = super.haskellPackages.override {
+    overrides = hself: hsuper: {
+      weeder = self.haskell.lib.dontCheck (hself.callPackage ./pkgs/weeder { });
+    };
```
akshaymankar, line 95: We should write the reason we're using a fork and link to a PR if we have expectation that this is temporary.

weeder.toml
```diff
@@ -0,0 +1,6 @@
+# weeder intro and further reading: https://github.com/ocharles/weeder?tab=readme-ov-file#weeder
+roots = ["^Main.main$", "^Paths_.*\\..*"]
+type-class-roots = true  # `root-instances` is more precise, but requires more config maintenance.
+
+# TODO: unused-types = true
+# TODO: type-class-roots = false, and see how bad it gets
```
akshaymankar, line 6: Forgotten todos?

weeder.toml
```diff
@@ -0,0 +1,6 @@
+# weeder intro and further reading: https://github.com/ocharles/weeder?tab=readme-ov-file#weeder
+roots = ["^Main.main$", "^Paths_.*\\..*"]
```
akshaymankar, line 2: I just ran weeder and I think this is not complete.

Makefile
```diff
@@ -127,11 +127,7 @@ devtest:
 	ghcid --command 'cabal repl integration' --test='Testlib.Run.mainI []'
 
 .PHONY: sanitize-pr
-sanitize-pr:
-	./hack/bin/generate-local-nix-packages.sh
-	make formatf
-	make hlint-inplace-pr
-	make hlint-check-pr  # sometimes inplace has been observed not to do its job very well.
```
akshaymankar, line 134: Why did you change this? Now it will run `formatc` instead of `foramtf` and the linting will be slower.

fisx: fair point, but i think the change should be in the other direction: [08c94df](https://github.com/wireapp/wire-server/pull/4088/commits/08c94df8645f4e5a973db60efcb7c9b4dc8c6ba4)

akshaymankar: I don't think it is a good idea. I'd say we should let CI be more thorough and we can keep having a little shortcut for our dev experience.
Because this is becoming a discussion, perhaps we should create another ticket/PR for this.

fisx: more good points.  give me one last chance: https://github.com/wireapp/wire-server/pull/4088/commits/5a6b9ef168c0a2e0b8bbc25e58c34453e0117cbb

akshaymankar (approved):

PR #4098, opened 2024-06-19
WPB 2690

smatting:
charts/coturn: support putting coturn into 'drain' mode when terminating pods, denying new incoming client connections. This speeds up graceful coturn restarts significantly. See related PR https://github.com/wireapp/coturn/pull/15 and https://github.com/wireapp/coturn/pull/12

drive-by fixes:
- use dumb-init to start the container to correctly handle signal handling in general
- write process PID to a writable file when running as non-root user

## Checklist

### DIFF
--- a/charts/coturn/templates/configmap-coturn-conf-template.yaml
+++ b/charts/coturn/templates/configmap-coturn-conf-template.yaml
@@ -26,6 +26,8 @@ data:
     ## don't turn on coturn's cli.
     no-cli
 
+    pidfile="/var/tmp/turnserver.pid"
+
     ## turn, stun.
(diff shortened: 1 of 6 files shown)


jschaul (approved):

charts/coturn/templates/configmap-coturn-conf-template.yaml
```diff
@@ -26,6 +26,8 @@ data:
     ## don't turn on coturn's cli.
     no-cli
 
+    pidfile="/var/tmp/turnserver.pid"
+
```
smatting, line 30: This was added only to prevent a warning

PR #4112, opened 2024-06-26
 catchErrors middleware: Don't create responses for ThreadKilled errors 

akshaymankar:
Also don't log anything. This happens when warp decides to reap a worker thread,
it could happen because the client has already closed the connection. Creating
responses and logging seems unnecessary.

https://wearezeta.atlassian.net/browse/WPB-7161

## Checklist

### DIFF
--- a/libs/wai-utilities/src/Network/Wai/Utilities/Server.hs
+++ b/libs/wai-utilities/src/Network/Wai/Utilities/Server.hs
@@ -243,7 +243,7 @@ errorHandlers =
   -- prevent us from logging the error cleanly later
   [ Handler $ \(x :: JSONResponse) -> pure (Right x),
     Handler $ \(x :: Wai.Error) -> pure (Left x),
-    -- warp throws 'ThreadKilled' when the client is gone or when it thinks its
+    -- warp throws 'ThreadKilled' when the client is gone or when it thinks it's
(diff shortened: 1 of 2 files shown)


libs/wai-utilities/src/Network/Wai/Utilities/Server.hs
```diff
@@ -238,12 +237,23 @@ catchErrorsWithRequestId getRequestId l app req k =
...
 -- | Standard handlers for turning exceptions into appropriate
 -- 'Error' responses.
-errorHandlers :: (Applicative m) => [Handler m (Either Wai.Error JSONResponse)]
+errorHandlers :: [Handler IO (Either Wai.Error JSONResponse)]
 errorHandlers =
   -- a Wai.Error can be converted to a JSONResponse, but doing so here would
   -- prevent us from logging the error cleanly later
   [ Handler $ \(x :: JSONResponse) -> pure (Right x),
     Handler $ \(x :: Wai.Error) -> pure (Left x),
+    -- warp throws 'ThreadKilled' when the client is gone or when it thinks its
```
pcapriotti, line 246:
```suggestion
    -- warp throws 'ThreadKilled' when the client is gone or when it thinks it's
```

pcapriotti (approved):
:shipit:

PR #4115, opened 2024-07-01
[DO NOT MERGE] check opensearch compatibility

amitsagtani97:
https://wearezeta.atlassian.net/browse/WPB-9850

This is a test pr and won't be merged. 

## Checklist

### DIFF
--- a/charts/elasticsearch-ephemeral/templates/es.yaml
+++ b/charts/elasticsearch-ephemeral/templates/es.yaml
@@ -32,16 +32,16 @@ spec:
           value: "single-node"
         - name: "action.auto_create_index"
           value: ".watches,.triggered_watches,.watcher-history-*,pod-*,node-*"
-        - name: "xpack.security.enabled"
-          value: "true"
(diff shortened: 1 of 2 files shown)


supersven:
I'm wondering if this shouldn't be the `opensearch-ephemeral` (while we keep the `elasticsearch-ephemeral` untouched.)

The pitfall I'm seeing here is that while we and customers are using ElasticSearch in production, this PR would lead to running tests with OpenSearch.

amitsagtani97:
> I'm wondering if this shouldn't be the `opensearch-ephemeral` (while we keep the `elasticsearch-ephemeral` untouched.)
> 
> The pitfall I'm seeing here is that while we and customers are using ElasticSearch in production, this PR would lead to running tests with OpenSearch.

Yes, that's the plan. I will create a new chart for experimenting.

amitsagtani97:
Tested the Opensearch compatibility on a WIAB test env. closing this pr.

PR #4116, opened 2024-07-01
add profiling options to nix and cabal files of brig and galley

MangoIV:
## Checklist

### DIFF
--- a/charts/brig/templates/deployment.yaml
+++ b/charts/brig/templates/deployment.yaml
@@ -1,5 +1,5 @@
 apiVersion: apps/v1
-kind: Deployment
+kind: StatefulSet
 metadata:
   name: brig
   labels:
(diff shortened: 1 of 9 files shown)


supersven:
To enable the external storage on EKS-based clusters, this needs to be set in the deployment's Helm values of `brig` and `galley`:

```yaml
profLogs:
  enabled: true
  storageClassName: gp2
``

File: charts/brig/templates/deployment.yaml [hunk 1]
```diff
@@ -167,3 +178,14 @@ spec:
...
+  {{- if .Values.profLogs.enabled }}
+  volumeClaimTemplates:
+    - metadata:
+        name: prof-logs
+      spec:
+        accessModes: [ "ReadWriteOnce" ]
+        storageClassName: {{ .Values.profLogs.storageClassName }}
+        resources:
+          requests:
+            storage: "10Gi"
```
elland, line 190: I'm wondering if that's enough. I got close to 9gb in 12-13 hours of profiling. Maybe we do `50gi`?

MangoIV: no. None of the available tools can handle event logs of > 2G graciously. If we go farther above we won’t get anything useful out of it. That’s why we increase the sample interval from .1 to 1 seconds

MangoIV: Also your profile didn’t do the right thing, unfortunately, the ones that we need produce ~2GB per hour.

MangoIV:
not planned

PR #4117, opened 2024-07-02
WPB-6954 Set SFT username's shared field according to team setting

battermann:
## Checklist

### DIFF
--- a/services/brig/src/Brig/Calling/API.hs
+++ b/services/brig/src/Brig/Calling/API.hs
@@ -57,29 +57,37 @@ import Polysemy.Error qualified as Polysemy
 import System.Logger.Class qualified as Log
 import System.Random.MWC qualified as MWC
 import Wire.API.Call.Config qualified as Public
+import Wire.API.Team.Feature (AllFeatureConfigs (afcConferenceCalling), FeatureStatus (FeatureStatusDisabled, FeatureStatusEnabled), wsStatus)
+import Wire.GalleyAPIAccess (GalleyAPIAccess, getAllFeatureConfigsForUser)
 import Wire.Network.DNS.SRV (srvTarget)
 
 -- | ('UserId', 'ConnId' are required as args here to make sure this is an authenticated end-point.)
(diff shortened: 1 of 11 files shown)


services/brig/src/Brig/Calling/API.hs
```diff
@@ -96,20 +104,26 @@ handleNoTurnServers (Left NoTurnServers) = do
...
+    pure $ case ccStatus of
+      FeatureStatusEnabled -> True
+      FeatureStatusDisabled -> False
   eitherConfig <-
     (dropTransport <$$>)
       . lift
       . liftSem
       . Polysemy.runError
-      $ newConfig env discoveredServers Nothing Nothing Nothing HideAllSFTServers CallsConfigDeprecated
+      $ newConfig env discoveredServers Nothing Nothing Nothing HideAllSFTServers CallsConfigDeprecated shared
```
stefanwire, line 126: Pretty sure that we don't create these credentials for the non-V2 endpoint. So, we could simply pass `False` and no test should fail.

services/brig/test/unit/Test/Brig/Calling.hs
```diff
@@ -323,7 +323,7 @@ testSFTStaticV2NoStaticUrl = do
       . ignoreLogs
       . interpretSFTInMemory mempty
       . throwErrorInIO @_ @NoTurnServers
-      $ newConfig env (Discovered turnUri) Nothing (Just sftEnv) (Just . unsafeRange $ 2) ListAllSFTServers (CallsConfigV2 Nothing)
+      $ newConfig env (Discovered turnUri) Nothing (Just sftEnv) (Just . unsafeRange $ 2) ListAllSFTServers (CallsConfigV2 Nothing) True
```
stefanwire, line 326: Perhaps add a test with `False` here and test that `s` is changed accordingly. I was admittedly sloppy when I adjusted these tests last time. 😞

stefanwire (approved):

PR #4118, opened 2024-07-02
Fix repeated with-rtsopts option

pcapriotti:
GHC does not support repeated `--with-rtsopts` options, and it simply applies the last one. This means many of the baked-in options were actually not being passed, including `-N` for some of the services and `-T` for cannon.

## Checklist

### DIFF
--- a/services/federator/federator.cabal
+++ b/services/federator/federator.cabal
@@ -205,7 +205,7 @@ executable federator
   ghc-options:
     -O2 -Wall -Wincomplete-uni-patterns -Wincomplete-record-updates
     -Wpartial-fields -fwarn-tabs -optP-Wno-nonportable-include-path
-    -threaded "-with-rtsopts=-N1 -T" -rtsopts -Wredundant-constraints
+    -threaded "-with-rtsopts=-N -T" -rtsopts -Wredundant-constraints
     -Wunused-packages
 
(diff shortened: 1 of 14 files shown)


MangoIV (approved):
LGTM

services/federator/federator.cabal
```diff
@@ -205,8 +205,8 @@ executable federator
   ghc-options:
     -O2 -Wall -Wincomplete-uni-patterns -Wincomplete-record-updates
     -Wpartial-fields -fwarn-tabs -optP-Wno-nonportable-include-path
-    -threaded -with-rtsopts=-N1 -with-rtsopts=-T -rtsopts
-    -Wredundant-constraints -Wunused-packages
+    -threaded "-with-rtsopts=-N1 -T" -rtsopts -Wredundant-constraints
+    -Wunused-packages
```
akshaymankar, line 209: I wonder why we specify `-N1` here:
1. Its the default anyway
2. What's the reason behind this, perhaps would be good to document. Seems to have been added without any specified reason: https://github.com/wireapp/wire-server/commit/47684fec021bfbd78f89e0ac8f70223929f40181

MangoIV: seems like it had been N1 as a default when history started in 2017.

PR #4119, opened 2024-07-03
[chore] replace cabal.project.local template and update cabal.project

MangoIV:
## Checklist


fisx (approved):

akshaymankar:
`package: *` applies to all packages not just local pacakges. I think this would fail if protobuf stuff which is not provided by nix has any warnings.

MangoIV:
can we not overwrite in case we run into warnings for these two packages in future? i.e. 
package * in general and then package proto-lens-protoc ghc-options -Wwarn?

MangoIV:
in fact, I deleted my cabal store and am trying to build right now, it doesn't look as though it's problematic.

MangoIV:
so protolens-protoc compiles without warnings.

MangoIV:
currently proto-lens-protoc and hspec-discover do not emit warnings. I have protected us against them failing in future by adding a -Wwarn flag to them. I think this should resolve the objections.

MangoIV:
There’s even a better option: just don’t use the package stanza at all: https://cabal.readthedocs.io/en/stable/cabal-project-description-file.html#package-configuration-options

fisx (approved):

PR #4121, opened 2024-07-03
Introduce VerificationCodeSubsystem

akshaymankar:
https://wearezeta.atlassian.net/browse/WPB-8891

## Checklist


libs/wire-subsystems/test/unit/Wire/MockInterpreters/Random.hs
```diff
@@ -0,0 +1,37 @@
...
+    pure x
+  NDigitNumber n -> withStatefulGen $ randomR (0, 10 ^ n - 1)
+
+runRandomPure :: InterpreterFor Random r
+runRandomPure = evalState defaulGen . randomToStatefulStdGen . raiseUnder
+
+defaulGen :: StdGen
+defaulGen = mkStdGen 0xBAD
+
+withStatefulGen :: (Member (State StdGen) r) => (StdGen -> (a, StdGen)) -> Sem r a
```
pcapriotti, line 32: This could be generalised to any type `s` instead of `StdGen`.

pcapriotti (approved):
LGTM

libs/wire-subsystems/test/unit/Wire/MockInterpreters/Random.hs
akshaymankar: I was wondering why there is no `modify`, but thought probably because of lack of atomicity guarantees. So I decided to write a specific one. So I didn't bother generalizing it in this module.

pcapriotti: What do you mean? There is a `modify`, but this one is an analogue of the `StateT` constructor. As a rule of thumb, I think it's good to turn types like `StdGen` above into type variables, because that limits the ways they could be used incorrectly in the implementation. Not a huge concern here though, obviously.

PR #4125, opened 2024-07-05
Federation V1 test setup

pcapriotti:
This sets up a federation-v1 backend in docker-compose and turns some of the federation tests into API compatibility tests. We ran into issues when trying to run `testRemoteUserGetsDeleted` on the legacy environments, for different reasons:
- open federation is broken in fed v0
- the refresh interval in fed v1 is too high.
So for the time being that test is not used to test API compatibility.

This PR also changes the default behaviour of the docker compose script. Now the federation-v0 (resp. the new federation-v1 backend) is not started unless the variable `ENABLE_FEDERATION_V0` (resp. `ENABLE_FEDERATION_V1`) is set to 1. The script can now also take optional arguments, which are passed directly to docker-compose.

https://wearezeta.atlassian.net/browse/WPB-9001

## Checklist


elland (approved):
🚀

stefanwire (approved):

dm17:
It is great to see recent this progress towards this federation plan! It would be an absolute boon since:
-- it is easy for anyone to install via their app store / F-DROID (its there!)
-- it would be easy to enable "add servers" in the client
-- anyone can connect via the official servers and then send each other private (federated) servers (via wire) to join
-- actually is E2EE on linux desktop unlike telegram (https://tdesktop.fail/)

This makes the flow / management easy for expert users who want to self-host wire, and no harder than any other messenger app to then get users onto their federated server!

PR #4130, opened 2024-07-08
WPB-10058 delete phone column in brig's user table

battermann:
https://wearezeta.atlassian.net/browse/WPB-10058

## Checklist


changelog.d/5-internal/WPB-10058
```diff
@@ -0,0 +1,2 @@
+- DB migration for dropping `phone` column from `user` table
+- Tool for deleting phone keys from `user_keys`
```
fisx, line 2: will this run automatically with every upgrade? should this be mentioned here or not?
is important that the first remove-phone release is deployed first if you dont like 5xx errors?

fisx (approved):

changelog.d/5-internal/WPB-10058
akshaymankar: I'd argue these belong to features section. Also, making these bullet points will cause the generated markdown to be wrong (or at least a little weird) because each file becomes a bullet point. So just create two files if you need two bullet points.

akshaymankar:
Perhaps there should also be docs about how to run this tool.

tools/db/remove-phone-keys/src/RemovePhoneKeys/Lib.hs
```diff
@@ -0,0 +1,85 @@
...
+  where
+    logEvery :: Int -> IntSum -> AppT IO ()
+    logEvery i s =
+      when (unIntSum s `mod` i == 0) $ Log.info $ "deleted_keys" .= show s
+
+run :: AppT IO ()
+run = do
+  ctr <- liftIO $ newIORef 0
+  -- deleting from user_keys while paginating through it is probably not a good idea
+  -- therefore we read all keys first and then delete them
```
mdimjasevic, line 74: Do you know the size of this table, at least approximately? I hope we can read them all at once into memory.

mdimjasevic (approved):
Looks good to me, assuming we can read all the phone keys at once into memory.

tools/db/remove-phone-keys/src/RemovePhoneKeys/Lib.hs
akshaymankar: Did we do some tests with this, I think the pagination might be smarter than we think.

battermann: I tested on staging with > 20M keys in mem which worked fine. Do you think that is sufficient?

changelog.d/5-internal/WPB-10058
mdimjasevic: Thanks, guys! Addressed in commit 96c15f09e.

battermann: done

PR #4140, opened 2024-07-11
Concurrently fetch user profiles from the DB

akshaymankar:
The endpoint typically takes 0.5-1 second to respond with ~200 users. It is called by the webapp on boot for every known user (not sure why) making UX very slow. 

Here is a chart from prod showing response times for this particular endpoint:

image

Local tests show about 70% reduction in response time.

## Checklist

### DIFF
--- a/changelog.d/5-internal/optimize-list-users
+++ b/changelog.d/5-internal/optimize-list-users
@@ -1,0 +1,1 @@
+Optimize getting a lot of users by concurrently getting target users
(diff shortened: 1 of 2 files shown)


changelog.d/5-internal/optimize-list-users
```diff
@@ -0,0 +1,3 @@
+Optimize getting a lot of users by:
+1. Getting details of requesting user only once
```
akshaymankar, line 2: This is not quite correct

elland (approved):
Not a fan of magic numbers but fine by me.

PR #4149, opened 2024-07-15
[WPB-8707] Remove phone functionality in the development client API version

mdimjasevic:
The PR removes phone numbers and phone functionality from the development client API version, namely from v6.

A couple of up-for-debate choices have been made, and I'm open to discuss them:

1. The `phone` key was dropped from Brig's `team_invitation` DB table. Previous releases were still reading this column. As there is no Team Settings UI action to enter a phone number, so the column was never written to;  hence not reading from the column will not miss to read actual phone numbers. Therefore, during deployment this will lead to benign 5xx errors. Related to this, the `"phone"` field for endpoints that get one or more team invitations is silently ignored and the field has been removed from the request body of the v6 endpoint to send a team invitation.
2. The `"phone"` field of the request body to `POST /register` is now ignored instead of giving an error.

Tracked by https://wearezeta.atlassian.net/browse/WPB-8707.

## Checklist

 - [x] Remove the `POST /login/send` endpoint,
 - [x] Remove the `PUT /self/phone` endpoint,
 - [x] Remove the `DELETE /self/phone` endpoint,
 - [x] Modify the `GET /self` endpoint to remove the `phone` key from the response,
 - [x] Modify the `GET /activate` endpoint to remove the `phone` key from the response,
 - [x] Modify the `POST /activate/send` endpoint to remove the `phone` key from the request body, i.e., no ability to send an activation code via SMS,
 - [x] Drop the `phone` field in the response to the `GET /teams/invitations/info` endpoint,
 - [x] Drop the `phone` field in the response to the `GET /teams/{tid}/invitations` endpoint,
 - [x] Drop the `phone` field in the response to the `GET /teams/{tid}/invitations/{iid}` endpoint,
 - [x] Drop the `phone` field from the request body of `POST /teams/{tid}/invitations` endpoint,
 - [x] Modify the `POST /register` endpoint to remove the `phone` key both from the request and the response,
 - [x] Modify the `POST /login` endpoint by removing phone-related keys, namely `phone` and `code`, from the request body,
 - [x] Modify the `POST /activate` endpoint by removing the `phone` key from the request body,


libs/wire-api/src/Wire/API/Routes/Public/Brig.hs
```diff
@@ -1432,9 +1476,32 @@ type AuthAPI =
...
                     (Respond 200 "OK" LoginCodeTimeout)
            )
+    :<|> Named
+           "login-v5"
+           ( "login"
+               :> Until 'V6
+               :> Summary "Authenticate a user to obtain a cookie and first access token"
+               :> Description "Logins are throttled at the server's discretion"
+               :> MakesFederatedCall 'Brig "send-connection-action"
+               :> ReqBody '[JSON] LoginV5
```
battermann, line 1486: Same as above... and more below, will not comment on all

battermann (approved):
Looks good. If there is a good reason not to use the `VersionedReqBody` wrapper, I'd like to know, and then it's good to go. Otherwise, it might be a sensible change?

PR #4152, opened 2024-07-15
[fix] API version check comes before method check

battermann:
Because the API version check is done via a version header and servant's order of checks while resolving the routes is fixed, it can happen that an endpoint with the same path, but different method is found even if it has a different version.

This change is a hack to enforce the version check to come first.


elland (approved):

mdimjasevic:
Federation tests were failing in the CI and I suppose for no good reason. I used the opportunity to rebase the PR on top of `develop` and to retrigger the CI.

PR #4153, opened 2024-07-15
Stop leaking ES error info.

elland:
https://wearezeta.atlassian.net/browse/WPB-6865

## Checklist


battermann:
I have manually reproduced and tested that the change works.

This implementation removes any exception message from the errors returned from wire-server when there is an internal error, not only the one mentioned in the ticket. These internal errors are always logged.

pcapriotti (approved):

PR #4158, opened 2024-07-17
WPB-10207 Match cipher suite tag in query parameters against key packages on replacing key packages

stefanwire:
https://wearezeta.atlassian.net/browse/WPB-10207

## Checklist

### DIFF
--- a/integration/test/API/Brig.hs
+++ b/integration/test/API/Brig.hs
@@ -349,11 +349,8 @@ deleteKeyPackages cid kps = do
   req <- baseRequest cid Brig Versioned ("/mls/key-packages/self/" <> cid.client)
   submit "DELETE" $ req & addJSONObject ["key_packages" .= kps]
 
-replaceKeyPackages :: ClientIdentity -> [Ciphersuite] -> [ByteString] -> App Response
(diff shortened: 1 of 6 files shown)


integration/test/API/Brig.hs
```diff
@@ -350,13 +350,16 @@ deleteKeyPackages cid kps = do
   submit "DELETE" $ req & addJSONObject ["key_packages" .= kps]
 
 replaceKeyPackages :: ClientIdentity -> [Ciphersuite] -> [ByteString] -> App Response
-replaceKeyPackages cid suites kps = do
+replaceKeyPackages cid suites kps = replaceKeyPackages' cid (Just suites) kps
+
+replaceKeyPackages' :: ClientIdentity -> Maybe [Ciphersuite] -> [ByteString] -> App Response
```
pcapriotti, line 355: The convention we've been sticking to in cases like this one is to create a data type say `ReplaceKeyPackages` with a `Default` instance and redefine `replaceKeyPackages` as `ClientIdentity -> ReplaceKeyPackages -> App Response`, so we can have something that resembles named arguments. Having primed method names is confusing and doesn't really scale.

pcapriotti (approved):
LGTM. Minor comment below.

battermann (approved):

PR #4163, opened 2024-07-18
[WPB-10230] Prevent creating a user with no identity

mdimjasevic:
There was never an assertion nor a test checking that a new user has to have an identity. This PR fixes that.

Tracked by https://wearezeta.atlassian.net/browse/WPB-10230.

## Checklist

### DIFF
--- a/changelog.d/3-bug-fixes/WPB-10230
+++ b/changelog.d/3-bug-fixes/WPB-10230
@@ -1,0 +1,1 @@
+Disallow creating a user with no identity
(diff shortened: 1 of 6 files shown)


mdimjasevic:
I am closing this PR. A number of integration tests in Brig fail, which is an indicator something is off here. A use case for no-identity accounts is guest rooms, and that is a justified case. No-identity accounts expire and are eventually deleted by the server. For more details see the bug report ticket.

PR #4164, opened 2024-07-22
One to one SFT feature flag config

pcapriotti:
This adds a config field to the `conferenceCalling` flag. It also reorganises the way flags are fetched from the database, so that the logic for computing default values is shared between the code for fetching all features and the code for fetching a single one.

https://wearezeta.atlassian.net/browse/WPB-10235

## Checklist


elland (approved):

services/galley/src/Galley/Cassandra/GetAllTeamFeatureConfigs.hs
```diff
@@ -40,6 +40,7 @@ data AllTeamFeatureConfigsRow = AllTeamFeatureConfigsRow
     -- conference calling
     conferenceCalling :: Maybe FeatureStatus,
     conferenceCallingTtl :: Maybe FeatureTTL,
+    conferenceCallingSftForOne2One :: Maybe Bool,
```
fisx, line 42: there is an old rule that we *always* use int instead of bool in cassandra, because it is more extensible.  not sure we'll ever need that here, or how.  just saying :)

libs/wire-api/src/Wire/API/Team/Feature.hs
```diff
@@ -360,6 +360,9 @@ instance (Arbitrary cfg, IsFeatureConfig cfg) => Arbitrary (WithStatusPatch cfg)
 -- if we switch to `unlocked`, we auto-enable the feature, and if we switch to locked, we
 -- auto-disable it.  But we need to change the API to force clients to use `lockStatus`
 -- instead of `status`, current behavior is just wrong.
+--
+-- FUTUREWORK(paolo): why delete this type? it's used as the request body of
```
fisx, line 404: fair.  i don't know how deletion follows from my rant about lock status.  feel free to remove both FUTUREWORKs, or better yet move the rant to somewhere suitable, eg, the lock status type.

integration/test/Test/FeatureFlags/User.hs
```diff
@@ -0,0 +1,70 @@
...
+
+  -- set user value for both users
+  for_ [alice, bob] $ \u -> do
+    void $ I.putFeatureForUser u featureName (object ["status" .= "disabled"]) >>= getBody 200
+    config <- I.getFeatureForUser u featureName >>= getJSON 200
+    config %. "status" `shouldMatch` "disabled"
+    config %. "ttl" `shouldMatch` "unlimited"
+
+    -- this config is just made up by brig, it does not reflect the actual value
+    -- that will be returned to the user
```
fisx, line 35: i don't understand.  are you saying the internal end-point returns a different config from the public end-point for the same user?  isn't that a bad thing?

integration/test/Test/FeatureFlags/User.hs
```diff
@@ -0,0 +1,70 @@
...
+
+import qualified API.BrigInternal as I
+import API.Galley
+import qualified API.GalleyInternal as I
+import SetupHelpers
+import Test.FeatureFlags.Util
+import Testlib.Prelude
+
+testFeatureConferenceCallingForUser :: App ()
+testFeatureConferenceCallingForUser = do
```
fisx, line 11: are there tests for lock_status for conf calling?

libs/wire-api/src/Wire/API/Team/Feature.hs
```diff
@@ -272,7 +273,7 @@ setTTL ttl (WithStatusBase s ls c _) = WithStatusBase s ls c (pure ttl)
 setWsTTL :: FeatureTTL -> WithStatus cfg -> WithStatus cfg
 setWsTTL = setTTL
 
-type WithStatus (cfg :: Type) = WithStatusBase Identity cfg
+type WithStatus = WithStatusBase Identity
```
fisx, line 315: i liked the unreduced variant better.

libs/wire-api/src/Wire/API/Team/Feature.hs
```diff
@@ -238,6 +244,36 @@ data WithStatusBase (m :: Type -> Type) (cfg :: Type) = WithStatusBase
...
 
+--------------------------------------------------------------------------------
+-- DbFeature
+
+-- | Feature data stored in the database, as a function of its default values.
+newtype DbFeature cfg = DbFeature
+  {unDbFeature :: WithStatusNoLock cfg -> WithStatusNoLock cfg}
+
+instance Semigroup (DbFeature cfg) where
+  DbFeature f <> DbFeature g = DbFeature (f . g)
```
fisx, line 256: are you sure you want `f` to write over `g`?

fisx (dismissed):

cassandra-schema.cql
```diff
@@ -1190,6 +1190,8 @@ CREATE TABLE galley_test.team_features (
     app_lock_inactivity_timeout_secs int,
     app_lock_status int,
     conference_calling int,
+    conference_calling_lock_status int,
```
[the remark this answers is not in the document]
pcapriotti: I don't understand. Nothing has changed here in terms of semantics. Locked has always meant: "you can't change it and it returns the default". It's just that before you weren't able to lock it, and now you can. What does it mean "it acts like lock status"?

integration/test/API/BrigInternal.hs
```diff
@@ -244,10 +244,37 @@ getEJPDInfo dom handles mode = do
...
+  submit "GET" req
+
+-- | http://staging-nginz-https.zinfra.io/api-internal/swagger-ui/brig/#/brig/put_i_users__uid__features_conferenceCalling
+putFeatureForUser ::
+  (HasCallStack, MakesValue user, MakesValue config) =>
+  user ->
+  String ->
+  config ->
+  App Response
+putFeatureForUser user featureName config = do
```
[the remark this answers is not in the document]
pcapriotti: Yeah, we only added the integration helper.

libs/wire-api/src/Wire/API/Team/Feature.hs
```diff
@@ -1,3 +1,4 @@
+{-# LANGUAGE ApplicativeDo #-}
```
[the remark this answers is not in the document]
pcapriotti: Feel free to make a PR.

services/brig/src/Brig/Calling/API.hs
```diff
@@ -60,7 +60,7 @@ import Polysemy
 import Polysemy.Error qualified as Polysemy
 import System.Logger.Class qualified as Log
 import Wire.API.Call.Config qualified as Public
-import Wire.API.Team.Feature (AllFeatureConfigs (afcConferenceCalling), FeatureStatus (FeatureStatusDisabled, FeatureStatusEnabled), wsStatus)
+import Wire.API.Team.Feature
```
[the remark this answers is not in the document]
pcapriotti: When I get a conflict or an error in one of those lines, I fix it by simply removing the import list.

integration/test/Test/FeatureFlags/User.hs
pcapriotti: It is a bad thing, but it was already there. The problem is that brig returns a whole `WithStatus` object, despite only having information about the status flag. We can try to improve this with the coming refactoring.

libs/wire-api/src/Wire/API/Team/Feature.hs
pcapriotti: I made this change because I wanted to use `WithStatus` as an argument of a higher-rank type constructor. I gave up with that approach, so this could be reverted, but I'm curious: why do you like the other one better? This is usable in more cases, and perfectly equivalent otherwise.

cassandra-schema.cql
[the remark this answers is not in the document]
pcapriotti: The problem was that some internal workflows were assuming the lock status to be set to "unlocked" and were simply operating at the level of "status". This change would have broken such a workflow.

services/galley/src/Galley/Cassandra/GetAllTeamFeatureConfigs.hs
pcapriotti: Switched to `Int`.

integration/test/Test/FeatureFlags/User.hs
pcapriotti: There are now.

libs/wire-api/src/Wire/API/Team/Feature.hs
pcapriotti: I added a note explaining this lock/disable business.

libs/wire-api/src/Wire/API/Team/Feature.hs
pcapriotti: I'm not sure which one is better. I actually think it would be more confusing to invert the order here.

libs/wire-api/src/Wire/API/Team/Feature.hs
pcapriotti: I'll leave this in, if you don't mind too much. It's the normal way to define a type synonym.

akshaymankar (approved):
:rocket:

PR #4167, opened 2024-07-23
[WPB-8892] Add block list operations to the user subsystem

battermann:
Refactor the code in Brig by moving the email block-listing actions into a separate effect in the `wire-subsystems` library.

Tracked by https://wearezeta.atlassian.net/browse/WPB-8892.

## Checklist


mdimjasevic (approved):
Looks good! I should get more familiar with the concept of mini-backend and how the interpreter for the block-listing effect fits into this idea.

PR #4175, opened 2024-07-30
nginx-zauth-module: Do not try to read the access token from query params

akshaymankar:
https://wearezeta.atlassian.net/browse/WPB-5260

Depends on https://github.com/wireapp/wire-webapp/pull/17842. There needs to be another PR made to wire-web-packages to further stop using the tokens.

## Checklist


akshaymankar:
:warning: This breaks websockets because browsers don't allow adding HTTP headers to requests. See https://stackoverflow.com/a/77060459

PR #4176, opened 2024-07-31
Enable manual usage of locally running wire-server 

akshaymankar:
Webapp repository needed a little change for this to work: https://github.com/wireapp/wire-webapp/pull/17844

## Checklist

### DIFF
--- a/docs/src/developer/developer/building.md
+++ b/docs/src/developer/developer/building.md
@@ -218,3 +218,26 @@ After all containers are up you can use these Makefile targets to run the tests 
    ```
 
    `TASTY_NUM_THREADS` can also be set to other values, it defaults to number of cores available.
+
+## How to run the webapp locally against locally running backend
+
+1. Clone the webapp from: https://github.com/wireapp/wire-webapp
+2. Install these depedencies needed for the webapp:
+   1. nodejs
+   2. yarn
+   3. mkcert
+3. Copy .env.localhost to .env and uncomment the local section
+4. Run the webapp using:
+   ```bash
+   yarn
+   yarn start
+   ```
+4. From wire-server repo start the dependencies using:
+   ```bash
+   ./deploy/dockerephemeral/run.sh
+   ```
+5. From wire-server repo start the backend using:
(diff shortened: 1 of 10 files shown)


integration/test/Testlib/RunServices.hs
```diff
@@ -57,6 +77,36 @@ main = do
...
+            )
             [backendA, backendB]
         liftIO run
+
+manualTestingOverrides :: ServiceOverrides
+manualTestingOverrides =
+  let smtpEndpoint = object ["host" .= "localhost", "port" .= (2500 :: Int)]
+      authSettings =
+        object
+          [ "userTokenTimeout" .= (4838400 :: Int),
```
supersven, line 93: I'm usually too lazy to calculate and would write a `let` or `where` binding for such numbers. E.g. `fiftysixDays = 4838400`.

But, that's probably a matter of taste :smile:

supersven (approved):
Some minor nitpicks. Other than that: LGTM :+1:

docs/src/developer/developer/building.md
```diff
@@ -218,3 +218,26 @@ After all containers are up you can use these Makefile targets to run the tests
    ```
 
    `TASTY_NUM_THREADS` can also be set to other values, it defaults to number of cores available.
+
+## How to run the webapp locally against locally running backend
+
+1. Clone the webapp from: https://github.com/wireapp/wire-webapp
+2. Install these depedencies needed for the webapp:
```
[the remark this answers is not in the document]
akshaymankar: Yes, I was thinking, but then I thought clearly noone would maintain that. So, its probably fine to just `nix shell nixpkgs#nodejs ...`

integration/test/Testlib/RunServices.hs
akshaymankar: I copy pasted these from default values of the helm charts and I have no clue what they correspond to, I'd wager noone would care about this until they testing something specific at which point they'd probably just change the values temporarily.

PR #4181, opened 2024-08-02
Feature flag refactoring (part 1)

pcapriotti:
First stab at refactoring feature flags and removing code duplication.

## Highlights of changes: 
 
- Improved naming slightly. Features types are now called `Feature`, `LockableFeature` and `LockableFeaturePatch`.
- TTL fields were removed from all records. They remain in the API, but are always hardcoded to be `unlimited`.
- Turned `AllFeatures` into an extensible record type.
- Removed `WithStatusBase` barbie.
- Deleted obsolete `computeFeatureConfigForTeamUser`.
- Abstracted `getFeature` and `setFeature`.
- Abstracted `getAllTeamFeatures`.

https://wearezeta.atlassian.net/browse/WPB-10323

## Checklist


libs/wire-api/src/Wire/API/Team/Feature.hs
```diff
@@ -602,40 +536,24 @@ instance ToSchema LockStatusResponse where
...
-        Nothing -> forgetLock defStatus
-        Just fs -> fs
-  where
-    lockStatus = fromMaybe (wsLockStatus defStatus) mLockStatusDb
+genericComputeFeature :: forall cfg. LockableFeature cfg -> DbFeature cfg -> LockableFeature cfg
+genericComputeFeature defFeature dbFeature =
+  let feat = applyDbFeature dbFeature defFeature
+   in case feat.lockStatus of
+        LockStatusLocked -> defFeature {lockStatus = LockStatusLocked}
+        LockStatusUnlocked -> feat
```
MangoIV, line 556: Maybe just a heads up to other reviewers looking at this (besides me), this logic needs to be equivalent to the old one. I think it makes sense like @pcapriotti explained, but yeah, more eyes can't hurt.

libs/wire-api/src/Wire/API/Team/Feature.hs
```diff
@@ -1271,120 +1265,107 @@ instance Cass.Cql FeatureStatus where
...
+--
+-- TODO(mangoiv): generalize this to be useable with schema profunctor
+class HObjectSchema c xs where
+  hobjectSchema :: (forall cfg. (c cfg) => ObjectSchema SwaggerDoc (f cfg)) -> ObjectSchema SwaggerDoc (NP f xs)
+
+instance HObjectSchema c '[] where
+  hobjectSchema _ = pure Nil
+
+instance (HObjectSchema c xs, c x) => HObjectSchema c ((x :: Type) : xs) where
+  hobjectSchema f = (:*) <$> hd .= f <*> tl .= hobjectSchema @c @xs f
```
MangoIV, line 1319: This should probably be moved to `schema-profunctor` or a corresponding utility module.
```suggestion
-- | object schema for nary products
class HObjectSchema c xs where
  hobjectSchema :: (forall cfg. (c cfg) => ObjectSchema SwaggerDoc (f cfg)) -> ObjectSchema SwaggerDoc (NP f xs)

instance HObjectSchema c '[] where
  hobjectSchema _ = pure Nil

instance (HObjectSchema c xs, c x) => HObjectSchema c ((x :: Type) : xs) where
  hobjectSchema f = (:*) <$> hd .= f <*> tl .= hobjectSchema @c @xs f
```

libs/wire-api/src/Wire/API/Team/Feature.hs
```diff
@@ -1271,120 +1265,107 @@ instance Cass.Cql FeatureStatus where
...
+
+instance (NpUpdate x xs) => NpUpdate x (y : xs) where
+  npUpdate' p x (y :* xs) = y :* npUpdate' p x xs
+
+instance (TypeError ('ShowType x :<>: 'Text " not found")) => NpUpdate x '[] where
+  npUpdate' = error "npUpdate': someone naughty removed the type error constraint"
+
+-- | Update the first field of a given type in an @'NP' f xs@.
+npUpdate :: forall x f xs. (NpUpdate x xs) => f x -> NP f xs -> NP f xs
+npUpdate = npUpdate' (Proxy @x)
```
MangoIV, line 1366: as already said above, these should probably be moved to another module.

libs/wire-subsystems/test/unit/Wire/MiniBackend.hs
```diff
@@ -1,3 +1,5 @@
+{-# OPTIONS_GHC -Wno-ambiguous-fields #-}
+
```
MangoIV, line 2: do we need this one? there doesn't seem to be any change here. 
```suggestion
```

services/galley/src/Galley/API/Internal.hs
```diff
@@ -72,7 +72,7 @@ import Polysemy
 import Polysemy.Error
 import Polysemy.Input
 import Polysemy.TinyLog qualified as P
-import Servant hiding (JSON, WithStatus)
+import Servant hiding (JSON)
```
MangoIV, line 75:
```suggestion
import Servant
```

MangoIV: (I expanded all of the things and it doesn't appear as though there's any other use of `JSON`)

services/galley/src/Galley/API/Teams/Features/Get.hs
MangoIV, line 1: this is so nice

services/galley/src/Galley/Cassandra/FeatureTH.hs
```diff
@@ -0,0 +1,49 @@
...
+    ]
+  where
+    mkPattern = foldr (\x y -> PromotedConsT `AppT` x `AppT` y) PromotedNilT
+
+    mkTuple [] = ConT ''()
+    mkTuple [v] = ConT ''Identity `AppT` v
+    mkTuple vs =
+      let n = length vs
+       in foldl' AppT (TupleT n) vs
+
```
MangoIV, line 47:
```suggestion
-- | generates some of the remaining @SOP.Generic@ instances as orphans
--   it is cut off at 50 on purpose to reduce compilation times
--   you may increase up to 64 which is the number at which you 
--   you should probably start fixing cql instead. 
```

services/galley/src/Galley/Cassandra/MakeFeature.hs
```diff
@@ -0,0 +1,440 @@
...
+    NP Maybe (FeatureRow cfg) ->
+    DbFeature cfg
+  mkFeature = foldMap dbFeatureStatus . hd
+
+  unmkFeature :: LockableFeature cfg -> NP Maybe (FeatureRow cfg)
+  default unmkFeature ::
+    (FeatureRow cfg ~ '[FeatureStatus]) =>
+    LockableFeature cfg ->
+    NP Maybe (FeatureRow cfg)
+  unmkFeature feat = Just feat.status :* Nil
```
MangoIV, line 48: maybe these should be factored out and called `defaultUnmkFeature` and `defaultMkFeature` to avoid duplicating the default code for the locked versions of things

services/galley/src/Galley/Cassandra/MakeFeature.hs
```diff
@@ -0,0 +1,440 @@
...
+import Galley.Cassandra.Instances ()
+import Generics.SOP
+import Imports hiding (Generic, Map)
+import Wire.API.Conversation.Protocol (ProtocolTag)
+import Wire.API.MLS.CipherSuite
+import Wire.API.Team.Feature
+
+-- | This is necessary in order to convert an @NP f xs@ type to something that
+-- CQL can understand.
+$(generateTupleP)
```
MangoIV, line 28:
```suggestion
$generateTupleP
```

tools/stern/test/integration/API.hs
```diff
@@ -1,7 +1,8 @@
 {-# LANGUAGE OverloadedRecordDot #-}
+{-# OPTIONS_GHC -Wno-ambiguous-fields #-}
+{-# OPTIONS_GHC -Wno-incomplete-patterns #-}
+{-# OPTIONS_GHC -Wno-orphans #-}
```
MangoIV, line 4: afaiu you didn't add any orphans and the other warning doesn't seem like we want to keep it, do we? 
```suggestion
```

MangoIV (commented):
kind of a big one, was really nice working on it :)

libs/wire-api/src/Wire/API/Event/FeatureConfig.hs
```diff
@@ -47,20 +47,20 @@ instance Arbitrary Event where
...
+                arbitrary @(LockableFeature AppLockConfig) <&> toJSON,
+                arbitrary @(LockableFeature FileSharingConfig) <&> toJSON,
+                arbitrary @(LockableFeature ClassifiedDomainsConfig) <&> toJSON,
+                arbitrary @(LockableFeature ConferenceCallingConfig) <&> toJSON,
+                arbitrary @(LockableFeature SelfDeletingMessagesConfig) <&> toJSON,
+                arbitrary @(LockableFeature GuestLinksConfig) <&> toJSON,
+                arbitrary @(LockableFeature SndFactorPasswordChallengeConfig) <&> toJSON,
+                arbitrary @(LockableFeature SearchVisibilityInboundConfig) <&> toJSON,
+                arbitrary @(LockableFeature MLSConfig) <&> toJSON,
+                arbitrary @(LockableFeature ExposeInvitationURLsToTeamAdminConfig) <&> toJSON
```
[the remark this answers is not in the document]
pcapriotti: I don't think `hcFoldMap` can be used here, since we don't have any NP or SOP to fold over. I changed this so that it uses the `Features` list by explicit recursion.

libs/wire-api/src/Wire/API/Team/Feature.hs
```diff
@@ -189,9 +182,8 @@ import Wire.Arbitrary (Arbitrary, GenericUniform (..))
 -- 12. Add a section to the documentation at an appropriate place
 -- (e.g. 'docs/src/developer/reference/config-options.md' (if applicable) or
 -- 'docs/src/understand/team-feature-settings.md')
-class IsFeatureConfig cfg where
+class (Default cfg, Default (LockableFeature cfg)) => IsFeatureConfig cfg where
```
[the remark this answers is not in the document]
pcapriotti: It's just to keep the constraints equivalent to what we had before. Before there was a `defFeatureStatus` method in the class itself.

libs/wire-api/src/Wire/API/Team/Feature.hs
```diff
@@ -296,153 +264,119 @@ data DbFeatureWithLock cfg = DbFeatureWithLock
...
-        <*> wsbConfig .= maybe_ (optField "config" schema)
-        <*> wsbTTL .= maybe_ (optField "ttl" schema)
+      LockableFeaturePatch
+        <$> (.status) .= maybe_ (optField "status" schema)
+        <*> (.lockStatus) .= maybe_ (optField "lockStatus" schema)
+        <*> (.config) .= maybe_ (optField "config" schema)
+        <* const FeatureTTLUnlimited
+          .= optField
+            "ttl"
+            (schema :: ValueSchema NamedSwaggerDoc FeatureTTL)
```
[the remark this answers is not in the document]
pcapriotti: I've added an explanation to the PR description.

libs/wire-api/src/Wire/API/Team/Feature.hs
pcapriotti: I thought about it, but this seems to be quite specific to features, since it only makes sense when an `NP` is built out of distinct types. I'm not sure it's worth introducing a special abstraction just for this case.

services/galley/src/Galley/Cassandra/MakeFeature.hs
pcapriotti: I'm not sure how that would work. Features with a lock status as well as status cannot really share much code, since they have a different type, so need a different pattern match. The duplication is minimal anyway, so I think this is fine.

tools/stern/test/integration/API.hs
pcapriotti: There is an orphan instance there.

libs/wire-api/src/Wire/API/Routes/Internal/LegalHold.hs
```diff
@@ -23,7 +23,7 @@ import Data.OpenApi (OpenApi)
 import Data.OpenApi.Lens
 import Data.Proxy
 import Imports
-import Servant.API hiding (Header, WithStatus)
+import Servant.API hiding (Header)
```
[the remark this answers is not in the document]
pcapriotti: Fixed

libs/wire-api/src/Wire/API/Team/Feature.hs
```diff
@@ -91,7 +77,12 @@ module Wire.API.Team.Feature
     MlsMigrationConfig (..),
     EnforceFileDownloadLocationConfig (..),
     LimitedEventFanoutConfig (..),
-    AllFeatures (..),
+    Features,
+    AllFeatures,
+    NpProject (..),
+    npProject,
+    NpUpdate (..),
+    npUpdate,
```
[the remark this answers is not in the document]
pcapriotti: Done

libs/wire-api/src/Wire/API/Routes/Internal/LegalHold.hs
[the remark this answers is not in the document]
pcapriotti: Fixed

libs/wire-api/src/Wire/API/Team/Feature.hs
```diff
@@ -1,8 +1,13 @@
 {-# LANGUAGE ApplicativeDo #-}
 {-# LANGUAGE GeneralizedNewtypeDeriving #-}
+{-# LANGUAGE PartialTypeSignatures #-}
+{-# LANGUAGE QuantifiedConstraints #-}
 {-# LANGUAGE RecordWildCards #-}
 {-# LANGUAGE StrictData #-}
 {-# LANGUAGE TemplateHaskell #-}
+{-# LANGUAGE UndecidableSuperClasses #-}
+{-# LANGUAGE NoStarIsType #-}
+{-# OPTIONS_GHC -Wno-ambiguous-fields #-}
```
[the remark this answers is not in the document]
pcapriotti: I also removed UndecidableSuperClasses, and specialised `LiftForF`, since it was only used twice anyway.

libs/wire-subsystems/test/unit/Wire/MiniBackend.hs
pcapriotti: Yeah, it's needed because `status` is now abiguous.

services/galley/src/Galley/API/Internal.hs
pcapriotti: Fixed

services/galley/src/Galley/Cassandra/MakeFeature.hs
pcapriotti: Removed brackets and added some docs.

libs/wire-api/src/Wire/API/Event/FeatureConfig.hs
```diff
@@ -42,30 +42,31 @@ data Event = Event
...
+
+instance
+  ( IsFeatureConfig cfg,
+    ToSchema cfg,
+    Arbitrary cfg,
+    AllArbitraryFeatures cfgs
+  ) =>
+  AllArbitraryFeatures (cfg : cfgs)
+  where
+  allArbitraryFeatures = arbitraryFeature @cfg : allArbitraryFeatures @cfgs
```
MangoIV, line 62: I think this might be possible with recursion on `SList` which reifies the type level list. You also need a csontraint on all elements, so perhaps smth like `cpara_SList` might work this time.

services/galley/src/Galley/Cassandra/MakeFeature.hs
```diff
@@ -0,0 +1,449 @@
...
+factorI :: forall f xs ys. (AllZip (IsF f) xs ys) => NP f xs -> NP I ys
+factorI Nil = Nil
+factorI (x :* xs) = I x :* factorI xs
+
+-- | Convert @NP I [f x1, ..., f xn]@ to @NP f [x1, ..., xn]@.
+--
+-- See 'factorI'.
+unfactorI :: forall f xs ys. (AllZip (IsF f) xs ys) => NP I ys -> NP f xs
+unfactorI Nil = Nil
+unfactorI (I x :* xs) = x :* unfactorI xs
```
MangoIV, line 444: it might be possible to do this with `htrans`. I thought about it a bit more which made me wonder why this is not in the library but I think it is. This seems to be exactly like htrans in that given a constraint that relates the element types (this is `IsF f`) you can make a natural transformation on the functor. I thought that `hfromI` or `htoI` would be the ones we want but it's not quite that because they require a `Coercible` constraint on `I` and `f` which is not what we want.

libs/wire-api/src/Wire/API/Team/Feature.hs
```diff
@@ -26,33 +27,16 @@ module Wire.API.Team.Feature
...
     LockStatus (..),
-    WithStatusBase (..),
     DbFeature (..),
-    DbFeatureWithLock (..),
+    dbFeatureLockStatus,
     dbFeatureStatus,
-    dbFeatureTTL,
     dbFeatureConfig,
     dbFeatureModConfig,
-    WithStatus,
```
elland, line 36: 👏

elland (approved):
If @MangoIV has no pending comments, LGTM.

PR #4183, opened 2024-08-05
[feat] bump nixpkgs - allow cabal-install 3.12

MangoIV:
## Checklist

 - n/a ~~Add a new entry in an appropriate subdirectory of `changelog.d`~~

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -238,11 +238,11 @@ add-license:
 
 .PHONY: treefmt
 treefmt:
-	treefmt
+	treefmt -u debug
 
 .PHONY: treefmt-check
 treefmt-check:
-	treefmt --fail-on-change
+	treefmt --fail-on-change -u debug
 
 #################################
 ## docker targets
--- a/integration/integration.cabal
+++ b/integration/integration.cabal
@@ -13,7 +13,7 @@ build-type:    Custom
 custom-setup
   setup-depends:
     , base
-    , Cabal
+    , Cabal >= 3.12
     , containers
     , directory
     , filepath
--- a/integration/integration.cabal
+++ b/integration/integration.cabal
@@ -13,7 +13,7 @@ build-type:    Custom
 custom-setup
   setup-depends:
     , base
-    , Cabal >= 3.12
+    , Cabal             >=3.12
     , containers
     , directory
     , filepath
--- a/nix/default.nix
+++ b/nix/default.nix
@@ -28,7 +28,7 @@ let
   docsPkgs = [
     pkgs.plantuml
     pkgs.texlive.combined.scheme-full
-    (pkgs.python3.withPackages
+    (pkgs.python310.withPackages
       (ps: with ps; [
         myst-parser
         rst2pdf
--- a/services/federator/test/integration/Main.hs
+++ b/services/federator/test/integration/Main.hs
@@ -15,12 +15,8 @@
 -- You should have received a copy of the GNU Affero General Public License along
 -- with this program. If not, see <https://www.gnu.org/licenses/>.
 
-module Main
-  ( main,
-  )
-where
+module Main (main) where
 
-import Control.Concurrent.Async
 import Imports
 import OpenSSL (withOpenSSL)
 import System.Environment (withArgs)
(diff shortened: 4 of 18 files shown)


integration/integration.cabal
```diff
@@ -13,7 +13,7 @@ build-type:    Custom
 custom-setup
   setup-depends:
     , base
-    , Cabal
+    , Cabal             >=3.12
```
akshaymankar, line 16: Why is this constraint needed?

MangoIV: because otherwise cabal multi-repl doesn't work.

MangoIV: should I put a note there?

services/spar/test-integration/Main.hs
```diff
@@ -40,8 +40,8 @@ import System.Random (randomRIO)
 import Test.Hspec
 import Test.Hspec.Core.Format
 import Test.Hspec.Core.Runner
-import Test.Hspec.JUnit
 import Test.Hspec.JUnit.Config.Env
+import Test.Hspec.JUnit.Formatter as JUnit
```
MangoIV, line 44: This is awesome - they deprecate the module but don't offer alternatives for the functions provided from there ... tempted to just ignore the deprecation warnings. what do you think @akshaymankar

integration/integration.cabal
akshaymankar: Isn't there only one version of Cabal available in the nix env? The project as is doesn't depend on cabal multi-repl right?

services/spar/test-integration/Main.hs
akshaymankar: I think you're supposed to use `Test.Hspec.JUnit.Format.junit`

integration/integration.cabal
MangoIV: in theory yes, in practice no, Cabal is a boot library and hence belongs to the GHC, hence nixpkgs installs two versions, 3.12 and 3.10 (which comes with GHC and cannot be removed), that then leads to proto-lens tool to select the wrong (old) version because it is actually constrained to Cabal <3.12 for no particular reason. It's all a bit messed up, if we had purely nixpkgs deps, then this wouldn't cause any issues because cabal doesn't actually care about how these deps have been installed...

MangoIV: This is the reason why cabal multi-repl is not enabled by default in HLS. By fixing this for us though, it removes the bug wrt duplicate constraints and it also improves HLS because of multi-repl. I think adding this constraint (which is provisioned by the nix env anyway) is worth this improvement.

services/spar/test-integration/Main.hs
MangoIV: ahah, that makes sense, thank you <3

MangoIV:
I think this is ready! <3

services/federator/test/integration/Main.hs
```diff
@@ -15,50 +15,26 @@
...
-      }
-  where
-    checksAndJUnitFormatter :: JUnitConfig -> FormatConfig -> IO Format
-    checksAndJUnitFormatter junitConfig config = do
-      junit <- junitFormat junitConfig config
-      let checksFormatter = fromJust (lookup "checks" $ configAvailableFormatters defaultConfig)
-      checks <- checksFormatter config
-      pure $ \event -> do
-        concurrently_ (junit event) (checks event)
+  withArgs hspecArgs . hspec . add junitConfig $ mkspec env
```
akshaymankar, line 37: This doesn't seem equivalent to the old code. Old code was creating a combined formatter which would print to console and create the junit xml report. This looks like it will do one of those but not both.

akshaymankar (changes requested):

Makefile
```diff
@@ -238,11 +238,11 @@ add-license:
 
 .PHONY: treefmt
 treefmt:
-	treefmt
```
MangoIV, line 241: This is because treefmt was very chatty about files it's not allowed to format

nix/default.nix
```diff
@@ -28,7 +28,7 @@ let
   docsPkgs = [
     pkgs.plantuml
     pkgs.texlive.combined.scheme-full
-    (pkgs.python3.withPackages
```
MangoIV, line 31: this is because nixpkgs not too long ago bumped the default python version and now many python packages are broken

akshaymankar (approved):

PR #4207, opened 2024-08-13
move JWK encoding of removal keys to API v7

stefanwire:
https://wearezeta.atlassian.net/browse/WPB-10263

## Checklist


pcapriotti (changes requested):
I don't think we should change a finalised API version.

pcapriotti (approved):
This is fine for now. I would consider adding a `format` query parameter in v7 to enable the old behaviour, though. Or even remove the new endpoint, and augment the old one with the query parameter (defaulting to the old format). That would be backwards compatible so we wouldn't need to use versioning at all.

PR #4214, opened 2024-08-16
Read sftToken from secrets.yaml

amitsagtani97:
https://wearezeta.atlassian.net/browse/WPB-10302

## Checklist

### DIFF
--- a/charts/brig/values.yaml
+++ b/charts/brig/values.yaml
@@ -100,6 +100,10 @@ config:
     legalholdUserTokenTimeout: 4838400
     legalholdAccessTokenTimeout: 900
   sft:
+    sftBaseDomain: sft.wire.example.com
+    sftSRVServiceName: sft
+    sftDiscoveryIntervalSeconds: 10
+    sftListLength: 20
     sftToken:
       ttl: 120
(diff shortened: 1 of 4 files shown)


charts/brig/values.yaml
```diff
@@ -99,6 +99,10 @@ config:
     providerTokenTimeout: 900
     legalholdUserTokenTimeout: 4838400
     legalholdAccessTokenTimeout: 900
+  sft:
```
smatting, line 102: This block doesn't have any conditions: even if secrets.sftTokenSecret is not set it would configure with a secret pointing to an non-existing file. This block should probably have the same condition as the secret file.

smatting (changes requested):
Since there are no integration tests for this, how was this tested?

smatting (commented):
This PR changes the default configuration of brig. It would even affect the wire cloud configuration which doesn't seem to configure the "sft" setting. Would the example values in the values.yaml not break existing deployments that overwrite the example settings in the "sft" block with correct real values?

If so we should change it so that the helm chart remains backwards compatible.

smatting (approved):
LGTM

PR #4218, opened 2024-08-19
[WPB-8887] initial impl for GetBy* account queries as Effect and interpreter

MangoIV:
https://wearezeta.atlassian.net/browse/WPB-8887

- new Effect operation  GetAccountBy in UserSubsystem
- new record GetBy
- new stores ActivationCodeStore and InvitationCodeStore
- new sql quasiquoter in cassandra-util
- some more Ord instances derived
- new function tSplit for the use with ViewPatterns

## Checklist


libs/wire-subsystems/src/Wire/InvitationCodeStore.hs
```diff
@@ -60,16 +64,23 @@ recordInstance ''StoredInvitationByTeam
...
+  LookupInvitationCodesByEmail :: EmailAddress -> InvitationCodeStore m [StoredInvitationByTeam]
 
 makeSem ''InvitationCodeStore
 
-lookupInvitationByEmail :: (Member InvitationCodeStore r, Member TinyLog r) => Email -> Sem r (Maybe StoredInvitation)
-lookupInvitationByEmail email = runMaybeT do
-  MkStoredInvitationByTeam {teamId, invId} <- MaybeT $ lookupSingleInvitationCodeByEmail email
-  MaybeT $ lookupInvitation teamId invId
+-- TODO: account for show/hide?
+lookupInvitationByEmail :: (Member InvitationCodeStore r, Member TinyLog r) => ShowOrHideInvitationUrl -> EmailAddress -> Sem r (Maybe StoredInvitation)
```
MangoIV, line 73: I changed this on purpose. ShowOrHideUrl is not part of the store. It should be a pure function outside of the store.

libs/wire-api/src/Wire/API/Team/Invitation.hs
```diff
@@ -1,4 +1,6 @@
+{-# LANGUAGE RebindableSyntax #-}
```
MangoIV, line 1: this should not be necessary.

libs/types-common/src/Wire/Timeout.hs
```diff
@@ -0,0 +1,32 @@
...
+  { timeoutDiff :: NominalDiffTime
+  }
+  deriving newtype (Eq, Enum, Ord, Num, Real, Fractional, RealFrac, Show)
+
+instance Read Timeout where
+  readsPrec i s =
+    case readsPrec i s of
+      [(x :: Int, s')] -> [(Timeout (fromIntegral x), s')]
+      _ -> []
+
```
elland, line 23: See if we can remove it, if yes we can use this type to replace the other Timeout type.

libs/wire-api/src/Wire/API/Team/Invitation.hs
```diff
@@ -67,27 +68,27 @@ instance ToSchema InvitationRequest where
...
-  { inTeam :: TeamId,
-    inRole :: Role,
-    inInvitation :: InvitationId,
-    inCreatedAt :: UTCTimeMillis,
-    -- | this is always 'Just' for new invitations, but for
+  { team :: TeamId,
+    role :: Role,
+    invitationId :: InvitationId,
+    createdAt :: UTCTimeMillis,
+    -- | this is always 'Just' for newvitations, but for
```
elland, line 86:
```suggestion
    -- | this is always 'Just' for new invitations, but for
```

libs/wire-subsystems/src/Wire/ActivationCodeStore/Cassandra.hs
```diff
@@ -0,0 +1,39 @@
...
+import Imports
+import OpenSSL.EVP.Digest
+import Polysemy
+import Polysemy.Embed
+import Wire.API.User.Activation
+import Wire.ActivationCodeStore
+import Wire.UserKeyStore (EmailKey, emailKeyUniq)
+
+interpretActivationCodeStoreToCassandra :: (Member (Embed IO) r) => ClientState -> InterpreterFor ActivationCodeStore r
+interpretActivationCodeStoreToCassandra casClient =
```
elland, line 16: Split this from the PR.

libs/wire-subsystems/src/Wire/InvitationCodeStore.hs
```diff
@@ -0,0 +1,124 @@
...
+  }
+  deriving (Show, Eq, Generic)
+  deriving (Arbitrary) via (GenericUniform StoredInvitation)
+
+recordInstance ''StoredInvitation
+
+data StoredInvitationInfo = MkStoredInvitationInfo
+  { teamId :: TeamId,
+    invitationId :: InvitationId,
+    code :: InvitationCode
```
elland, line 59: Let's try dropping it and see what breaks?

libs/wire-subsystems/src/Wire/InvitationCodeStore.hs
```diff
@@ -0,0 +1,124 @@
...
+    -- TODO(mangoiv): maybe we can drop this last element
+  }
+  deriving (Show, Eq, Generic)
+  deriving (Arbitrary) via (GenericUniform StoredInvitationInfo)
+
+recordInstance ''StoredInvitationInfo
+
+data PaginatedResult a
```
elland, line 67: TODO: use this type elsewhere we have paginated results.

libs/wire-subsystems/src/Wire/InvitationCodeStore.hs
```diff
@@ -0,0 +1,124 @@
...
+import Wire.API.Team.Invitation (Invitation (inviteeEmail))
+import Wire.API.Team.Invitation qualified as Public
+import Wire.API.Team.Role (Role, defaultRole)
+import Wire.API.User (EmailAddress, InvitationCode, Name)
+import Wire.Arbitrary (Arbitrary, GenericUniform (..))
+import Wire.Sem.Logger qualified as Log
+import Wire.Timeout
+
+data StoredInvitation = MkStoredInvitation
+  { teamId :: TeamId,
```
elland, line 42: TODO: Private constructor?

libs/wire-subsystems/src/Wire/InvitationCodeStore.hs
```diff
@@ -0,0 +1,124 @@
...
+data InvitationCodeStore :: Effect where
+  InsertInvitation :: InvitationId -> TeamId -> Role -> UTCTime -> Maybe UserId -> EmailAddress -> Maybe Name -> Timeout -> InvitationCodeStore m StoredInvitation
+  LookupInvitation :: TeamId -> InvitationId -> InvitationCodeStore m (Maybe StoredInvitation)
+  LookupInvitationInfo :: InvitationCode -> InvitationCodeStore m (Maybe StoredInvitationInfo)
+  LookupInvitationCodesByEmail :: EmailAddress -> InvitationCodeStore m [StoredInvitationInfo]
+  -- | invariant: page size is 100
+  LookupInvitationsPaginated :: Maybe (Range 1 500 Int32) -> TeamId -> Maybe InvitationId -> InvitationCodeStore m (PaginatedResult [StoredInvitation])
+  CountInvitations :: TeamId -> InvitationCodeStore m Int64
+  DeleteInvitation :: TeamId -> InvitationId -> InvitationCodeStore m ()
+  DeleteInvitations :: TeamId -> InvitationCodeStore m ()
```
elland, line 83: DeleteTeamInvitations? DeleteAllTeamInvitations?

libs/wire-subsystems/src/Wire/InvitationCodeStore/Cassandra.hs
```diff
@@ -0,0 +1,188 @@
...
+      LookupInvitationsPaginated mSize tid miid -> embed $ lookupInvitationsPaginatedImpl mSize tid miid
+      CountInvitations tid -> embed $ countInvitationsImpl tid
+      DeleteInvitation tid invId -> embed $ deleteInvitationImpl tid invId
+      DeleteInvitations tid -> embed $ deleteInvitationsImpl tid
+
+insertInvitationImpl ::
+  --   ( Log.MonadLogger m,
+  --     MonadClient m
+  --   ) =>
+  --   ShowOrHideInvitationUrl ->
```
elland, line 38:
```suggestion
```

libs/wire-subsystems/src/Wire/InvitationCodeStore/Cassandra.hs
```diff
@@ -0,0 +1,188 @@
...
+          { teamId = teamId,
+            role = Just role,
+            invitationId = invId,
+            createdAt = now,
+            createdBy = uid,
+            email = email,
+            name = name,
+            code = code
+          }
+  -- TODO: see how we can improve this
```
elland, line 62: Ticket to rework invitations?

libs/wire-subsystems/src/Wire/InvitationCodeStore.hs
```diff
@@ -0,0 +1,124 @@
...
+  deriving stock (Eq, Ord, Show, Functor, Foldable)
+
+----------------------------
+
+data InvitationCodeStore :: Effect where
+  InsertInvitation :: InvitationId -> TeamId -> Role -> UTCTime -> Maybe UserId -> EmailAddress -> Maybe Name -> Timeout -> InvitationCodeStore m StoredInvitation
+  LookupInvitation :: TeamId -> InvitationId -> InvitationCodeStore m (Maybe StoredInvitation)
+  LookupInvitationInfo :: InvitationCode -> InvitationCodeStore m (Maybe StoredInvitationInfo)
+  LookupInvitationCodesByEmail :: EmailAddress -> InvitationCodeStore m [StoredInvitationInfo]
+  -- | invariant: page size is 100
```
elland, line 79:
```suggestion
  -- | Range is page size, it defaults to 100
```

libs/wire-subsystems/src/Wire/UserStore.hs
```diff
@@ -63,6 +63,9 @@ data UserStore m a where
   --   an email address or phone number.
   IsActivated :: UserId -> UserStore m Bool
   LookupLocale :: UserId -> UserStore m (Maybe (Maybe Language, Maybe Country))
+  -- | Look up accounts given user ids. For the purpose of the DB, Users and accounts are identical, so this
+  -- returns a @['StoredUser']@
+  LookupAccounts :: [UserId] -> UserStore m [StoredUser]
```
elland, line 68: Far apart from `GetUser` and different name makes it a bit confusing.

libs/wire-subsystems/src/Wire/InvitationCodeStore.hs
fisx: ...  or the other way around?

changelog.d/5-internal/wpb-8887
```diff
@@ -0,0 +1 @@
+Migrated user invitation DB access to Subsystems.
```
fisx, line 1: isn't this PR primarily about getAccountsBy in the user subsystem, and this merely requires the invitation store?

libs/wire-subsystems/src/Wire/InvitationCodeStore/Cassandra.hs
```diff
@@ -0,0 +1,184 @@
...
+    cqlInvitation :: PrepQuery W (TeamId, InvitationId) ()
+    cqlInvitation = "DELETE FROM team_invitation where team = ? AND id = ?"
+    cqlInvitationInfo :: PrepQuery W (Identity InvitationCode) ()
+    cqlInvitationInfo = "DELETE FROM team_invitation_info WHERE code = ?"
+    cqlInvitationEmail :: PrepQuery W (EmailAddress, TeamId) ()
+    cqlInvitationEmail = "DELETE FROM team_invitation_email WHERE email = ? AND team = ?"
+    lookupInvitationCodeEmail :: Client (Maybe (InvitationCode, EmailAddress))
+    lookupInvitationCodeEmail = retry x1 (query1 cqlInvitationCodeEmail (params LocalQuorum (teamId, invId)))
+    cqlInvitationCodeEmail :: PrepQuery R (TeamId, InvitationId) (InvitationCode, EmailAddress)
+    cqlInvitationCodeEmail = "SELECT code, email FROM team_invitation WHERE team = ? AND id = ?"
```
fisx, line 171:
```suggestion
    cqlInvitationInfo :: PrepQuery W (Identity InvitationCode) ()
    cqlInvitationInfo = "DELETE FROM team_invitation_info WHERE code = ?"
    
    cqlInvitationEmail :: PrepQuery W (EmailAddress, TeamId) ()
    cqlInvitationEmail = "DELETE FROM team_invitation_email WHERE email = ? AND team = ?"
    
    lookupInvitationCodeEmail :: Client (Maybe (InvitationCode, EmailAddress))
    lookupInvitationCodeEmail = retry x1 (query1 cqlInvitationCodeEmail (params LocalQuorum (teamId, invId)))
    
    cqlInvitationCodeEmail :: PrepQuery R (TeamId, InvitationId) (InvitationCode, EmailAddress)
    cqlInvitationCodeEmail = "SELECT code, email FROM team_invitation WHERE team = ? AND id = ?"
```

libs/wire-subsystems/src/Wire/InvitationCodeStore/Cassandra.hs
```diff
@@ -0,0 +1,184 @@
...
+deleteInvitationsImpl :: TeamId -> Client ()
+deleteInvitationsImpl teamId =
+  runConduit $
+    paginateC cqlSelect (paramsP LocalQuorum (Identity teamId) 100) x1
+      .| Conduit.mapM_ (pooledMapConcurrentlyN_ 16 (deleteInvitationImpl teamId . runIdentity))
+  where
+    cqlSelect :: PrepQuery R (Identity TeamId) (Identity InvitationId)
+    cqlSelect = "SELECT id FROM team_invitation WHERE team = ? ORDER BY id ASC"
+
+--------------------------------
```
fisx, line 182:
```suggestion
```

?

libs/wire-subsystems/src/Wire/UserStore/Cassandra.hs
```diff
@@ -180,3 +186,12 @@ activatedSelect = "SELECT activated FROM user WHERE id = ?"
 
 localeSelect :: PrepQuery R (Identity UserId) (Maybe Language, Maybe Country)
 localeSelect = "SELECT language, country FROM user WHERE id = ?"
+
+accountsSelect :: PrepQuery R (Identity [UserId]) (TupleType StoredUser)
+accountsSelect =
```
fisx, line 191: remove and make `selectUser` above cover both use cases?

libs/types-common/src/Wire/Timeout.hs
fisx: I just compiled everything without the FromJSON instance and it turns out we need it in `Brig.Options`.

Further, our other `Timeout` type contains a `Word64`, and this one contains `NominalDiffTime`, so we shouldn't touch this topic in this PR...

libs/types-common/src/Wire/Timeout.hs
```diff
@@ -0,0 +1,32 @@
+module Wire.Timeout
```
fisx, line 1: This module adds `Wire.*` to this package, which is a bit messy.  Add it to `Util` instead?

libs/wire-subsystems/test/unit/Wire/MockInterpreters/InvitationCodeStore.hs
```diff
@@ -0,0 +1,40 @@
...
+
+inMemoryInvitationCodeStoreInterpreter ::
+  forall r.
+  ( Member (State (Map (TeamId, InvitationId) StoredInvitation)) r,
+    Member (State (Map (InvitationCode) StoredInvitationInfo)) r
+  ) =>
+  InterpreterFor InvitationCodeStore r
+inMemoryInvitationCodeStoreInterpreter = interpret \case
+  InsertInvitation invitationId teamId role' createdAt' createdBy email name _timeout -> do
+    code <- todo -- InvitationCode . encodeBase64Url <$> unsafeCoerce (randBytes 24)
```
fisx, line 24: why not use `mkInvitationCode`? it does exactly this.

services/brig/src/Brig/API/User.hs
```diff
@@ -143,6 +140,7 @@ import Wire.DeleteQueue
 import Wire.EmailSubsystem
 import Wire.Error
 import Wire.GalleyAPIAccess as GalleyAPIAccess
+import Wire.InvitationCodeStore qualified as Store
```
fisx, line 143:
```suggestion
import Wire.InvitationCodeStore qualified as InvitationCodeStore
```

services/brig/src/Brig/API/User.hs
```diff
@@ -409,37 +415,38 @@ createUser new = do
...
       ok <- lift $ liftSem $ claimKey uk uid
       unless ok $
         throwE RegisterErrorUserKeyExists
-      let minvmeta :: (Maybe (UserId, UTCTimeMillis), Role)
-          minvmeta = ((,inCreatedAt inv) <$> inCreatedBy inv, Team.inRole inv)
-      added <- lift $ liftSem $ GalleyAPIAccess.addTeamMember uid (Team.iiTeam ii) minvmeta
+      let minvmeta :: Maybe (UserId, UTCTimeMillis)
+          minvmeta = (,inv.createdAt) <$> inv.createdBy
+          role :: Role
+          role = fromMaybe defaultRole inv.role
```
fisx, line 431: how come this wasn't here before?

services/brig/src/Brig/Team/API.hs
```diff
@@ -110,10 +121,14 @@ teamSizePublic uid tid = do
...
-getInvitationCode :: TeamId -> InvitationId -> (Handler r) FoundInvitationCode
+getInvitationCode ::
+  (Member Store.InvitationCodeStore r) =>
+  TeamId ->
+  InvitationId ->
+  (Handler r) FoundInvitationCode
 getInvitationCode t r = do
-  code <- lift . wrapClient $ DB.lookupInvitationCode t r
-  maybe (throwStd $ errorToWai @'E.InvalidInvitationCode) (pure . FoundInvitationCode) code
+  inv <- lift . liftSem $ Store.lookupInvitation t r
```
fisx, line 130: `lookupInvitationCode` changed its name. maybe i should go through all the changes in the implementation that you did during the move, and see if they are all semantics-preserving?

services/brig/src/Brig/Team/API.hs
```diff
@@ -230,69 +251,163 @@ createInvitation' tid mUid inviteeRole mbInviterUid fromEmail body = do
...
+  let toInvitations is = mapM (toInvitation showInvitationUrl) is
+  lift (liftSem $ Store.lookupInvitationsPaginated mSize tid startingId) >>= \case
+    PaginatedResultHasMore storedInvs -> do
+      invs <- toInvitations storedInvs
+      pure $ InvitationList invs True
+    PaginatedResult storedInvs -> do
+      invs <- toInvitations storedInvs
+      pure $ InvitationList invs False
+
+-- \| brig used to not store the role, so for migration we allow this to be empty and fill in the
```
fisx, line 310:
```suggestion
-- | brig used to not store the role, so for migration we allow this to be empty and fill in the
```

services/brig/src/Brig/Team/API.hs
```diff
@@ -230,69 +251,163 @@ createInvitation' tid mUid inviteeRole mbInviterUid fromEmail body = do
...
+  TeamId ->
+  Maybe InvitationId ->
+  Maybe (Range 1 500 Int32) ->
+  (Handler r) Public.InvitationList
+listInvitations uid tid startingId mSize = do
   ensurePermissions uid tid [AddTeamMember]
   showInvitationUrl <- lift $ liftSem $ GalleyAPIAccess.getExposeInvitationURLsToTeamAdmin tid
-  rs <- lift $ wrapClient $ DB.lookupInvitations showInvitationUrl tid start (fromMaybe (unsafeRange 100) mSize)
-  pure $! Public.InvitationList (DB.resultList rs) (DB.resultHasMore rs)
+  let toInvitations is = mapM (toInvitation showInvitationUrl) is
```
fisx, line 301: inline?

services/brig/src/Brig/Team/API.hs
```diff
@@ -230,69 +251,163 @@ createInvitation' tid mUid inviteeRole mbInviterUid fromEmail body = do
...
+  lift (liftSem $ Store.lookupInvitationsPaginated mSize tid startingId) >>= \case
+    PaginatedResultHasMore storedInvs -> do
+      invs <- toInvitations storedInvs
+      pure $ InvitationList invs True
+    PaginatedResult storedInvs -> do
+      invs <- toInvitations storedInvs
+      pure $ InvitationList invs False
```
fisx, line 308: Is this the pagination type newly introduced in this PR? Why not `Cassandra.Page`? It's already there, and it allows for nicer code here because it lets you get to the page and the `hasMore` flag in any order.

services/brig/src/Brig/Team/API.hs
```diff
@@ -230,69 +251,163 @@ createInvitation' tid mUid inviteeRole mbInviterUid fromEmail body = do
...
+
+-- \| brig used to not store the role, so for migration we allow this to be empty and fill in the
+-- default here.
+toInvitation ::
```
fisx, line 312: why is this new here?  i mean, why wasn't it needed before?

services/brig/src/Brig/Team/API.hs
```diff
@@ -230,69 +251,163 @@ createInvitation' tid mUid inviteeRole mbInviterUid fromEmail body = do
...
+        role = fromMaybe defaultRole storedInv.role,
+        invitationId = storedInv.invitationId,
+        createdAt = storedInv.createdAt,
+        createdBy = storedInv.createdBy,
+        inviteeEmail = storedInv.email,
+        inviteeName = storedInv.name,
+        inviteeUrl = url
+      }
+
+mkInviteUrl ::
```
fisx, line 332: why is this new here? i mean, why wasn't it needed before?

services/brig/src/Brig/Team/API.hs
```diff
@@ -230,69 +251,163 @@ createInvitation' tid mUid inviteeRole mbInviterUid fromEmail body = do
...
+  (Handler r) (Maybe Public.Invitation)
 getInvitation uid tid iid = do
-  ensurePermissions uid tid [AddTeamMember]
-  showInvitationUrl <- lift $ liftSem $ GalleyAPIAccess.getExposeInvitationURLsToTeamAdmin tid
-  lift $ wrapClient $ DB.lookupInvitation showInvitationUrl tid iid
+  invitationM <- lift . liftSem $ Store.lookupInvitation tid iid
+  case invitationM of
+    Nothing -> pure Nothing
+    Just invitation -> do
+      ensurePermissions uid tid [AddTeamMember]
```
fisx, line 372: moving this here changes behavior in case permission is not granted *and* the invitation is not available.

why change it in the first place?  if this is about runtime performance: please don't :)

services/brig/src/Brig/Team/API.hs
```diff
@@ -230,69 +251,163 @@ createInvitation' tid mUid inviteeRole mbInviterUid fromEmail body = do
...
+  (Handler r) Public.Invitation
 getInvitationByCode c = do
-  inv <- lift . wrapClient $ DB.lookupInvitationByCode HideInvitationUrl c
-  maybe (throwStd $ errorToWai @'E.InvalidInvitationCode) pure inv
+  inv <- lift . liftSem $ Store.lookupInvitationByCode c
+  maybe (throwStd $ errorToWai @'E.InvalidInvitationCode) (pure . Store.invitationFromStored Nothing) inv
 
-headInvitationByEmail :: EmailAddress -> (Handler r) Public.HeadInvitationByEmailResult
-headInvitationByEmail e = do
+-- FIXME(mangoiv): This should not be in terms of store
```
fisx, line 385: why not? what else? what does this function do?

services/brig/src/Brig/Team/API.hs
```diff
@@ -302,15 +417,17 @@ suspendTeam ::
...
   (Handler r) NoContent
 suspendTeam tid = do
-  Log.info $ Log.msg (Log.val "Team suspended") ~~ Log.field "team" (toByteString tid)
+  lift $ liftSem $ Log.info $ Log.msg (Log.val "Team suspended") ~~ Log.field "team" (toByteString tid)
   changeTeamAccountStatuses tid Suspended
-  lift $ wrapClient $ DB.deleteInvitations tid
-  lift $ liftSem $ GalleyAPIAccess.changeTeamStatus tid Team.Suspended Nothing
+  lift . liftSem $ do
+    Store.deleteInvitations tid
+    GalleyAPIAccess.changeTeamStatus tid Team.Suspended Nothing
```
fisx, line 430: how is this different from line 311/427 above?

libs/wire-subsystems/src/Wire/InvitationCodeStore.hs
MangoIV: Please let’s not do this. Just name the constructor something that makes it clear like “UnsafeMk…”, if we must?

tools/stern/stern.cabal
```diff
@@ -186,11 +186,8 @@ test-suite stern-tests
 
 executable stern-integration
   main-is:            Main.hs
-
-  -- cabal-fmt: expand test/integration
```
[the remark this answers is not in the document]
MangoIV: It produces incorrect results which confuses cabal

libs/wire-subsystems/src/Wire/UserSubsystem/Interpreter.hs
```diff
@@ -495,3 +507,67 @@ checkHandlesImpl check num = reverse <$> collectFree [] check num
...
+                (Deleted, _, _) -> valid
+                (Ephemeral, _, _) -> valid
+
+  accsByIds :: [UserAccount] <-
+    wither accountValid
+      =<< lookupAccounts (nubOrd $ handleUserIds <> getByUserIds)
+
+  accsByEmail <- flip foldMap getByEmail \email -> do
+    let ek = mkEmailKey email
+    mactiveUid <- lookupKey ek
```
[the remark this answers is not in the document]
MangoIV: catMaybes below gets rid of Maybe?

services/brig/src/Brig/Team/API.hs
MangoIV: I was careful to and Igor also had a look at it but more eyes can’t hurt. 

The reason why I did this is because lookupInvitation was two functions in a trench coat. splitting it up simplified things and separation of concerns.

services/brig/src/Brig/Team/API.hs
MangoIV: Two reasons: stores are abstract interfaces and should ideally not depend on Cassandra, I know it does in practice but we should at least  try and make it as separated as possible and at least not have it depend on the Cassandra packages, afaiu. 

Secondly, @elland doesnt like Bool, which is fair 😅

services/brig/src/Brig/Team/API.hs
MangoIV: This is the other function which was in the trench coat I mentioned earlier.

services/brig/src/Brig/Team/API.hs
MangoIV: We split up functions that did multiple things at once to reduce interetwingledness

services/brig/src/Brig/Team/API.hs
MangoIV: I’m sure it wasn’t runtime performance, probably needs to get looked at if this changes behaviour.

services/brig/src/Brig/API/User.hs
elland: It was somewhere else, we've streamlined it.

services/brig/src/Brig/Team/API.hs
elland: Less readable, and it's called in 2 places.

services/brig/src/Brig/Team/API.hs
elland: 1. it was partially somewhere else
2. despite show or hide url being a galley setting, it was passed around up to the DB.

services/brig/src/Brig/Team/API.hs
elland: Moved from elsewhere.

tools/stern/stern.cabal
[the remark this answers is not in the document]
elland: It populates with files from the folder.

libs/wire-subsystems/src/Wire/InvitationCodeStore.hs
elland: That might also be an idea 🤔

services/brig/src/Brig/Team/API.hs
fisx: ah, cool. maybe next time you can be a little more explicit about the reasons, then you won't have to handle me separately :)

changelog.d/5-internal/wpb-8887
elland, line 1:
```suggestion
Implemented GetBy* for complex account queries.
```

services/brig/src/Brig/Team/API.hs
elland: That is odd, seems we're calling both the old function and the new subsystem one. I was under the impression we were _deleting_ old functions as we moved to to Polysemy....

elland: I'm adding comments. This one changes the team status in Galley, the other one updates all users from the given team.

services/brig/src/Brig/Team/API.hs
elland: Yeah it changes the behaviour, before even if there were no invitations, we'd get an error on permission failing, now we're leaking information.

fisx:
didn't we decide to leave account stuff out of this PR, and make it only about Invitations?

changelog.d/5-internal/wpb-8887
```diff
@@ -0,0 +1 @@
+Implemented GetBy* for complex account queries.
```
fisx, line 1: that's not accurate any more. same with PR description

fisx:
> didn't we decide to leave account stuff out of this PR, and make it only about Invitations?

we sorted that out: the part of activation codes that is needed for getAccountsBy is already here, the rest will follow in a separate PR.

libs/wire-subsystems/src/Wire/PasswordResetCodeStore/Cassandra.hs
```diff
@@ -79,6 +76,14 @@ genPhoneCode =
   PasswordResetCode . unsafeFromText . pack . printf "%06d"
     <$> liftIO (randIntegerZeroToNMinusOne 1000000)
 
+codeDeleteImpl :: (MonadClient m) => PasswordResetKey -> m ()
+codeDeleteImpl prk =
+  retry x5
+    . write codeDeleteQuery
+    . params LocalQuorum
+    . Identity
+    $ prk
```
MangoIV, line 85:
```suggestion
-- FUTUREWORK(fisx,elland): this should be replaced by a method in a 
-- future auth subsystem
codeDeleteImpl :: (MonadClient m) => PasswordResetKey -> m ()
codeDeleteImpl prk =
  retry x5
    . write codeDeleteQuery
    . params LocalQuorum
    . Identity
    $ prk
```

libs/wire-subsystems/src/Wire/PasswordResetCodeStore/Cassandra.hs
```diff
@@ -19,6 +19,8 @@
 module Wire.PasswordResetCodeStore.Cassandra
   ( passwordResetCodeStoreToCassandra,
     interpretClientToIO,
+    -- Temporary measure until we create AuthSubsystem
```
MangoIV, line 22:
```suggestion
```

libs/wire-subsystems/src/Wire/UserSubsystem.hs
```diff
@@ -53,11 +53,30 @@ instance Default UserProfileUpdate where
         supportedProtocols = Nothing
       }
 
+-- | how to get an account for a user
+data GetBy = MkGetBy
+  { -- | whether or not to include ending invitations in the lookups }
```
MangoIV, line 58:
```suggestion
  { -- | whether or not to include ending invitations in the lookups
```

MangoIV: why no `-XStrictData` for this module - are there any types here that require laziness?

libs/wire-subsystems/src/Wire/UserSubsystem.hs
```diff
@@ -53,11 +53,30 @@ instance Default UserProfileUpdate where
...
+    includePendingInvitations :: !Bool,
+    -- | get accounds by 'UserId's
+    getByUserIds :: ![UserId],
+    -- | get accounds by 'Email's
+    getByEmail :: ![EmailAddress],
+    -- | get accounds by their 'Handle'
```
MangoIV, line 64:
```suggestion
    -- | get accounts by 'UserId's
    getByUserIds :: ![UserId],
    -- | get accounts by 'Email's
    getByEmail :: ![EmailAddress],
    -- | get accounts by their 'Handle'
```

libs/wire-subsystems/src/Wire/UserSubsystem.hs
```diff
@@ -71,6 +90,7 @@ data UserSubsystem m a where
   CheckHandles :: [Handle] -> Word -> UserSubsystem m [Handle]
   -- | parses a handle, this may fail so it's effectful
   UpdateHandle :: Local UserId -> Maybe ConnId -> UpdateOriginType -> Text {- use Handle here? -} -> UserSubsystem m ()
+  -- TODO(mangoiv): this can probably go in favour of 'GetAccountsBy'
```
MangoIV: have we looked at whether this can go? If we want to postpone it, maybe it should be a `FUTUREWORK`? 
```suggestion
  -- FUTUREWORK(mangoiv): this can probably go in favour of 'GetAccountsBy'
```

libs/wire-subsystems/src/Wire/UserSubsystem.hs
```diff
@@ -96,3 +116,11 @@ getUserProfile luid targetUser =
...
 getLocalUserProfile targetUser =
   listToMaybe <$> getLocalUserProfiles ((: []) <$> targetUser)
+
+getLocalUserAccount :: (Member UserSubsystem r) => Local UserId -> Sem r (Maybe UserAccount)
+getLocalUserAccount uid =
+  listToMaybe
+    <$> getAccountsBy
+      ( qualifyAs uid $
+          def {getByUserIds = [tUnqualified uid]}
+      )
```
MangoIV, line 124: this looks very similar to the thing I said above - maybe this should be the impl of that effect op? what am I missing?

services/brig/src/Brig/API/Internal.hs
```diff
@@ -278,7 +282,9 @@ authAPI =
   Named @"legalhold-login" (callsFed (exposeAnnotations legalHoldLogin))
     :<|> Named @"sso-login" (callsFed (exposeAnnotations ssoLogin))
     :<|> Named @"login-code" getLoginCode
-    :<|> Named @"reauthenticate" reauthenticate
+    -- We qualify in place to avoid changing the internal API too much
+    -- FUTUREWORK?
+    :<|> Named @"reauthenticate" (\uid reauth -> qualifyLocal uid >>= \luid -> reauthenticate luid reauth)
```
MangoIV, line 287: I don't understand this comment - can you clarify? 

Maybe we can also decide whether this should be `FUTUREWORK` or not

services/brig/src/Brig/API/OAuth.hs
```diff
@@ -67,7 +68,7 @@ import Wire.Sem.Now qualified as Now
 internalOauthAPI :: ServerT I.OAuthAPI (Handler r)
 internalOauthAPI =
   Named @"create-oauth-client" registerOAuthClient
-    :<|> Named @"get-oauth-client" getOAuthClientById
+    :<|> Named @"i-get-oauth-client" getOAuthClientById
```
MangoIV, line 71: I don't understand this change - can you clarify why it has been done?

libs/wire-subsystems/src/Wire/UserSubsystem/Interpreter.hs
```diff
@@ -495,3 +493,67 @@ checkHandlesImpl check num = reverse <$> collectFree [] check num
...
+    ac <- lookupActivationCode ek
+    let muidFromActivationKey = ac >>= fst
+    res <- getUsers (nubOrd $ catMaybes [mactiveUid, muidFromActivationKey])
+    pure $
+      map
+        storedToAcc
+        if includePendingInvitations
+          then res
+          else filter (\acc -> acc.status /= Just PendingInvitation) res
+  pure (nubOrd $ accsByIds <> accsByEmail)
```
MangoIV, line 559: all of these should ideally be `Set` - but I don't know if it's a good idea to add a `FUTUREWORK`...

libs/wire-subsystems/test/unit/Wire/MiniBackend.hs
```diff
@@ -123,16 +134,23 @@ data MiniBackend = MkMiniBackend
     users :: [StoredUser],
     userKeys :: Map EmailKey UserId,
     passwordResetCodes :: Map PasswordResetKey (PRQueryData Identity),
-    blockList :: [EmailKey]
+    blockList :: [EmailKey],
+    activationCodes :: Map EmailKey (Maybe UserId, ActivationCode),
+    invitationInfos :: Map InvitationCode StoredInvitationInfo,
+    invitations :: Map (TeamId, InvitationId) StoredInvitation
   }
```
MangoIV, line 141: I still don't like that we're doing all these adhoc `Map`s...

libs/wire-subsystems/test/unit/Wire/MiniBackend.hs
```diff
@@ -352,9 +370,29 @@ interpretMaybeFederationStackState maybeFederationAPIAccess localBackend teamMem
...
+
+liftActivationCodeStoreState :: (Member (State MiniBackend) r) => Sem (State (Map EmailKey (Maybe UserId, ActivationCode)) : r) a -> Sem r a
+liftActivationCodeStoreState = interpret \case
+  Polysemy.State.Get -> gets (.activationCodes)
+  Put newAcs -> modify $ \b -> b {activationCodes = newAcs}
+
+liftInvitationCodeStoreState :: (Member (State MiniBackend) r) => Sem (State (Map (TeamId, InvitationId) StoredInvitation) : r) a -> Sem r a
+liftInvitationCodeStoreState = interpret \case
+  Polysemy.State.Get -> gets (.invitations)
+  Put newInvs -> modify $ \b -> b {invitations = newInvs}
```
MangoIV, line 394: same for this - I think this should all be replaced by handing something like a `Lens'` to a more generic function - but maybe something to think about in the future ;)

libs/wire-subsystems/test/unit/Wire/MockInterpreters/UserStore.hs
```diff
@@ -17,6 +19,7 @@ inMemoryUserStoreInterpreter ::
   InterpreterFor UserStore r
 inMemoryUserStoreInterpreter = interpret $ \case
   GetUser uid -> gets $ find (\user -> user.id == uid)
+  GetUsers TODO -> TODO
```
MangoIV, line 22: oversight?

libs/wire-subsystems/test/unit/Wire/MockInterpreters/UserSubsystem.hs
```diff
@@ -1,15 +1,11 @@
...
-import Wire.UserKeyStore
 import Wire.UserSubsystem
 
 userSubsystemTestInterpreter :: [UserAccount] -> InterpreterFor UserSubsystem r
-userSubsystemTestInterpreter initialUsers =
+userSubsystemTestInterpreter _initialUsers =
   interpret \case
-    GetLocalUserAccountByUserKey localUserKey -> case (tUnqualified localUserKey) of
-      EmailKey _ email -> pure $ find (\u -> userEmail u.accountUser == Just email) initialUsers
     _ -> error $ "userSubsystemTestInterpreter: implement on demand"
```
MangoIV, line 11: why was this removed?

MangoIV (commented):
feel free to ignore the nitpicks, as always, maybe the more important bit is that tests are missing, maybe an oversight. 

It would also be nice if e.g. weeding would be done in a separate PR in the future :)

libs/wire-subsystems/test/unit/Wire/MockInterpreters/UserSubsystem.hs
fisx: as you requested in a TODO, this is not an operation of the user subsystem effect any more.  if it was used in any tests, they will fail and we'll look into that.

libs/wire-subsystems/test/unit/Wire/MockInterpreters/UserStore.hs
fisx: yes :)

libs/types-common/default.nix
```diff
@@ -40,6 +40,7 @@
 , quickcheck-instances
 , random
 , schema-profunctor
+, scientific
```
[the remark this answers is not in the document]
fisx: are you doing the CI's job now?  :)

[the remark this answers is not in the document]
MangoIV: as good as I can :D

libs/wire-subsystems/src/Wire/PasswordResetCodeStore/Cassandra.hs
elland: Why remove this?

libs/wire-subsystems/src/Wire/UserSubsystem/Interpreter.hs
```diff
@@ -514,47 +516,51 @@ getAccountsByImpl (tSplit -> (domain, MkGetBy {includePendingInvitations, includ
...
-    res <- getUsers (nubOrd $ catMaybes [mactiveUid, muidFromActivationKey])
-    pure $
-      map
-        storedToAcc
-        if includePendingInvitations
-          then res
-          else filter (\acc -> acc.status /= Just PendingInvitation) res
+    getUsers (nubOrd . catMaybes $ [mactiveUid])
+      <&> map storedToAcc
+      >>= (filterPendingInvitations >=> filterPendingActivation)
```
MangoIV, line 564: :golf:

libs/wire-subsystems/src/Wire/UserSubsystem/Interpreter.hs
```diff
@@ -1,4 +1,6 @@
 {-# LANGUAGE RecordWildCards #-}
+{-# OPTIONS_GHC -Wno-redundant-constraints #-}
+{-# OPTIONS_GHC -Wno-unused-imports #-}
```
MangoIV, line 3: leftover?

MangoIV:
@elland is this PR ready (modulo tests) or should I hold on wrt reviewing?

libs/wire-subsystems/src/Wire/UserSubsystem.hs
```diff
@@ -115,6 +117,17 @@ getLocalUserProfile :: (Member UserSubsystem r) => Local UserId -> Sem r (Maybe
 getLocalUserProfile targetUser =
   listToMaybe <$> getLocalUserProfiles ((: []) <$> targetUser)
 
+getLocalUserAccountUnverified :: (Member UserSubsystem r) => Local UserId -> Sem r (Maybe UserAccount)
```
fisx, line 120: is this a good name? or a good function? use `getAccountsBy` instead?

if you want to keep it, maybe call it `getLocalUserAccountIncludePending`?

libs/wire-subsystems/test/unit/Wire/MockInterpreters/UserSubsystem.hs
```diff
@@ -7,9 +7,22 @@ import Wire.API.User
...
+        filter
+          ( \u ->
+              mailKeyFrom u
+                `elem` getBy.getByEmail
+          )
+          initialUsers
     _ -> error $ "userSubsystemTestInterpreter: implement on demand"
+
+mailKeyFrom :: ExtendedUserAccount -> EmailKey
+mailKeyFrom acc =
```
fisx, line 24: users with invitation pending may need to be searched by their *unvalidated* email address in the extended account. not sure.

services/brig/src/Brig/Data/Activation.hs
```diff
@@ -120,15 +134,18 @@ activateKey k c u = verifyCode k c >>= pickUser >>= activate
           pure . Just $ EmailActivated uid (emailKeyOrig key)
       -- if the key is the same, we only want to update our profile
       | otherwise = do
-          lift (runM (passwordResetCodeStoreToCassandra @m @'[Embed m] (E.codeDelete (mkPasswordResetKey uid))))
+          -- Temporary measure until we create AuthSubsystem
```
fisx, line 137:
```suggestion
```

redundant. this is already stated in the fun def.

libs/wire-subsystems/src/Wire/UserSubsystem.hs
```diff
@@ -53,11 +53,33 @@ instance Default UserProfileUpdate where
         supportedProtocols = Nothing
       }
 
+-- | how to get an account for a user
+data GetBy = MkGetBy
+  { -- | whether or not to include ending invitations in the lookups
+    includePendingInvitations :: !Bool,
+    -- -- | whether or not to include users with unverified identities
+    -- includePendingActivations :: !Bool,
```
fisx, line 61:
```suggestion
```

(unless we need this somewhere)

services/brig/test/integration/API/Team.hs
```diff
@@ -731,7 +732,8 @@ testInvitationPaging opts brig = do
               === statusCode
         (invs, more) <- (ilInvitations &&& ilHasMore) <$> responseJsonError r
         if more
-          then (invs :) <$> getPages (count + step) (fmap inInvitation . listToMaybe . reverse $ invs) step
+          then -- TODO: improve
+            (invs :) <$> getPages (count + step) (fmap (.invitationId) . listToMaybe . reverse $ invs) step
```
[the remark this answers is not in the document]
elland: I've amended the comment to say `improve readability`, maybe a futurework.

services/brig/test/integration/API/Team.hs
```diff
@@ -338,6 +338,8 @@ testInvitationEmailLookupRegister brig = do
   email <- randomEmail
   (owner, tid) <- createUserWithTeam brig
   let invite = stdInvitationRequest email
+  -- This incidentally also tests that sending multiple
+  -- invites from the same team results in last-invite-wins scenario
```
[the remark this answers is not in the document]
elland: No idea, but changing the behaviour is not in scope. I hope we do get a ticket to rework invitations.

libs/wire-subsystems/test/unit/Wire/MiniBackend.hs
elland: Yes but out of scope.

libs/wire-subsystems/test/unit/Wire/MiniBackend.hs
```diff
@@ -72,6 +80,24 @@ import Wire.UserSubsystem
 import Wire.UserSubsystem.Error
 import Wire.UserSubsystem.Interpreter
 
+newtype PendingNotEmptyIdentityStoredUser = PendingNotEmptyIdentityStoredUser StoredUser
+  deriving (Show, Eq)
+
+instance Arbitrary PendingNotEmptyIdentityStoredUser where
+  arbitrary = do
+    user <- arbitrary `suchThat` \user -> isJust user.identity
```
MangoIV, line 88: why not create a user identity first, then update the generated user with that identity? That way our tests don't slow down because of unnecessary discards. it also increases coverage.

libs/wire-subsystems/test/unit/Wire/MiniBackend.hs
```diff
@@ -72,6 +80,24 @@ import Wire.UserSubsystem
...
+    pure $ PendingNotEmptyIdentityStoredUser (user {status = Just PendingInvitation})
+
+newtype NotPendingEmptyIdentityStoredUser = NotPendingEmptyIdentityStoredUser StoredUser
+  deriving (Show, Eq)
+
+-- TODO: make sure this is a valid state
+instance Arbitrary NotPendingEmptyIdentityStoredUser where
+  arbitrary = do
+    user <- arbitrary `suchThat` \user -> isNothing user.identity
```
MangoIV, line 97: same here

libs/wire-subsystems/test/unit/Wire/MockInterpreters/InvitationCodeStore.hs
```diff
@@ -0,0 +1,41 @@
...
+
+inMemoryInvitationCodeStoreInterpreter ::
+  forall r.
+  ( Member (State (Map (TeamId, InvitationId) StoredInvitation)) r,
+    Member (State (Map (InvitationCode) StoredInvitationInfo)) r
+  ) =>
+  InterpreterFor InvitationCodeStore r
+inMemoryInvitationCodeStoreInterpreter = interpret \case
+  InsertInvitation (MkInsertInvitation invitationId teamId role' createdAt' createdBy email name) _timeout -> do
+    code <- unsafeCoerce mkInvitationCode
```
MangoIV, line 25: please remove this `unsafeCoerce`. I don't think it's necessary, especially without a type annotation or anything.

libs/wire-subsystems/test/unit/Wire/MockInterpreters/UserSubsystem.hs
```diff
@@ -7,9 +7,22 @@ import Wire.API.User
...
+  case acc.account.accountUser.userIdentity of
+    Just (EmailIdentity mail) -> mkEmailKey mail
+    Just (SSOIdentity _ (Just mail)) -> mkEmailKey mail
+    _ -> error "Why are we testing users without emails for this?"
```
MangoIV, line 28: is just the `Nothing` case missing here? (at least that's what I infer from the message?) if so, can we make that explicit?

services/brig/src/Brig/API/User.hs
```diff
@@ -112,9 +107,8 @@ import Data.List1 as List1 (List1, singleton)
 import Data.Misc
 import Data.Qualified
 import Data.Range
-import Data.Time.Clock (UTCTime, addUTCTime)
 import Data.UUID.V4 (nextRandom)
-import Imports
+import Imports hiding (local)
```
MangoIV, line 111:
```suggestion
import Imports
```

services/brig/src/Brig/User/Auth.hs
```diff
@@ -282,32 +292,48 @@ newAccess uid cid ct cl = do
...
       throwE $
         if pending
           then LoginPendingActivation
           else LoginFailed
     Just uid -> pure uid
 
-validateLoginId :: (MonadReader Env m) => LoginId -> ExceptT LoginError m (Either EmailKey Handle)
+validateLoginId :: (Applicative m) => LoginId -> m (Either EmailKey Handle)
 validateLoginId (LoginByEmail email) = (pure . Left . mkEmailKey) email
 validateLoginId (LoginByHandle h) = (pure . Right) h
```
MangoIV, line 316: let's make this a pure function. 
```suggestion
validateLoginId :: LoginId -> Either EmailKey Handle
validateLoginId (LoginByEmail email) = Left $ mkEmailKey email
validateLoginId (LoginByHandle h) = Right h
```

services/brig/src/Brig/Data/Activation.hs
```diff
@@ -105,7 +105,7 @@ activateKey k c u = wrapClientE (verifyCode k c) >>= pickUser >>= activate
     activate :: (EmailKey, UserId) -> ExceptT ActivationError (AppT r) (Maybe ActivationEvent)
     activate (key, uid) = do
       luid <- qualifyLocal uid
-      a <- lift (liftSem $ User.getLocalUserAccount luid) >>= maybe (throwE invalidUser) pure
+      a <- lift (liftSem $ User.getLocalUserAccount luid False) >>= maybe (throwE invalidUser) pure
```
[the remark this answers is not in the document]
fisx: (i was wrong, next line explains it: this function is only called in flows where the user is active.)

services/brig/src/Brig/API/Internal.hs
```diff
@@ -278,7 +285,9 @@ authAPI =
   Named @"legalhold-login" (callsFed (exposeAnnotations legalHoldLogin))
     :<|> Named @"sso-login" (callsFed (exposeAnnotations ssoLogin))
     :<|> Named @"login-code" getLoginCode
-    :<|> Named @"reauthenticate" reauthenticate
+    -- We qualify in place to avoid changing the internal API too much
+    -- FUTUREWORK?
```
fisx, line 289:
```suggestion
    -- Qualify in place, because the domain is provided in a non-standard way on this end-point, and we don't want to mess with that interface.
```

fisx:
now `getExtendedAccountsBy` works as expected for some of the places where it's called, but the tests in wire-subsystem don't pass any more.

we may need more flags steering the incoherent behavior in just the right places, like `includeIdentityLessAccounts :: Bool`, and who knows what else.

but first, we need to collect all the places where `getExtendedAccountsBy` is called (and that's *lots*), and trace what the old behavior was.

@MangoIV still think this PR is worth saving?  :)P

MangoIV:
> think this PR is worth saving

it’s not on me to the decide, it’s not my time that goes away without result

services/brig/src/Brig/API/Internal.hs
fisx: i thought i fixed this and pushed the change, but it's gone. maybe an overly aggressive rebase?

anyway: cf7ec8b3f20dfdc8056d23127982e94af806247d

libs/wire-subsystems/test/unit/Wire/MockInterpreters/InvitationCodeStore.hs
```diff
@@ -0,0 +1,42 @@
+{-# LANGUAGE RecordWildCards #-}
+{-# OPTIONS_GHC -Wwarn #-}
+
+module Wire.MockInterpreters.InvitationCodeStore where
```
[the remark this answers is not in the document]
fisx: separate PR!

libs/wire-subsystems/test/unit/Wire/MiniBackend.hs
fisx: non-trivial (identity is not a field of stored user, but a function in several fields).

separate PR!

libs/wire-subsystems/test/unit/Wire/MiniBackend.hs
fisx: separate PR!

elland (approved):
image

libs/wire-subsystems/src/Wire/InvitationCodeStore/Cassandra.hs
```diff
@@ -0,0 +1,192 @@
...
+deleteInvitationsImpl teamId =
+  runConduit $
+    paginateC cqlSelect (paramsP LocalQuorum (Identity teamId) 100) x1
+      .| Conduit.mapM_ (pooledMapConcurrentlyN_ 16 (deleteInvitationImpl teamId . runIdentity))
+  where
+    cqlSelect :: PrepQuery R (Identity TeamId) (Identity InvitationId)
+    cqlSelect = "SELECT id FROM team_invitation WHERE team = ? ORDER BY id ASC"
+
+mkInvitationCode :: IO InvitationCode
+mkInvitationCode = InvitationCode . encodeBase64Url <$> randBytes 24
```
MangoIV, line 192: https://hoogle.zinfra.io/file/nix/store/ll4lg104j02cpv5li2zcx603rbdkd8gd-polysemy-wire-zoo-0.1.0-doc/share/doc/polysemy-wire-zoo-0.1.0/html/Wire-Sem-Random.html

This will also resolve our issues with `unsafePerformIO`. `unsafePerformIO` is dangerous there because, depending on optimizations, profiling, ghc version and other factors, GHC could decide to e.g. float out the `let` binding and hence not create a new invitation code.

elland: Seems like a neat FUTUREWORK for when we do invitation subsystems.

MangoIV: `unsafePerformIO` has valid usecases but willingly breaking referential transparency is not one of them. If you only want one invitation code, put it in a separate function, put a `NOINLINE` pragma but if you want new ones every time, then it has to be in `IO`. I don't get why this is problematic to do, it's a small change and we already have all the glue ready.

fisx: We only use `unsafePerformIO` in tests; in fact, the code where it is used is weed. Also there is a comment in the unsafe line that we should check that this actually works.

Good enough for now?

MangoIV: The problem is that there is no "actually works" here since GHC might just decide differently depending on how it's configured or just randomly in some future ghc version. 

If it's weed, then why include it in the PR? I was always asked to just stash stuff that is not used anyway.

fisx (approved):

PR #4220, opened 2024-08-20
WPB-2568 Fix shellcheck linting problems on all shell scripts

smatting:
This PR activates shellcheck linting for almost all shell scripts.

It also fixes various suggestions on the shell scripts that are now shellchecked.

Tracked by https://wearezeta.atlassian.net/browse/WPB-2568

Current sprint: https://wearezeta.atlassian.net/browse/WPB-9757

## Checklist


akshaymankar:
It is annoying that this wouldn't respect gitignore and could overwrite files in dist-newstyle, etc. I am looking into why it the git walker doesn't work. So far I have found this:

Running `treefmt --no-cache -vv --walk=git -f shellcheck` prints this:

```
DEBU format | shellcheck: executing: /home/axeman/workspace/wire-server/.env/bin/shellcheck docs/convert/convert.sh docs/convert/revert.sh hack/bin/helm_overrides.sh
INFO format | shellcheck: 3 file(s) processed in 21.992731ms
traversed 6499 files
emitted 6499 files for processing
```

But if we ask git about how many files it has by running `git ls-files | wc -l`, it says: 6671.

When using the `filesystem` walker it says this:

```
DEBU format | shellcheck: executing: /home/axeman/workspace/wire-server/.env/bin/shellcheck changelog.d/mk-cleanup.sh deploy/dockerephemeral/db-migrate/brig-index.sh deploy/dockerephemeral/db-migrate/brig-schema.sh deploy/dockerephemeral/db-migrate/galley-schema.sh deploy/dockerephemeral/db-migrate/gundeck-schema.sh deploy/dockerephemeral/db-migrate/spar-schema.sh deploy/dockerephemeral/init_vhosts.sh deploy/dockerephemeral/run.sh docs/convert/convert.sh docs/convert/revert.sh hack/bin/cabal-run-all-tests.sh hack/bin/cabal-run-integration.sh hack/bin/certchain.sh hack/bin/create_team.sh hack/bin/create_team_request_code.sh hack/bin/find-latest-docker-tag.sh hack/bin/gen-certs.sh hack/bin/generate-local-nix-packages.sh hack/bin/helm_overrides.sh hack/bin/integration-cleanup.sh hack/bin/integration-logs-relevant-bits.sh hack/bin/integration-test.sh hack/bin/kind-upload-image.sh hack/bin/kind-upload-images.sh hack/bin/kubectl-get-debug-info.sh hack/bin/mk-drv.sh hack/bin/nix-hls.sh hack/bin/oauth_test.sh hack/bin/register_idp.sh hack/bin/set-wire-server-image-version.sh hack/bin/update.sh hack/bin/upload-image.sh hack/bin/upload-images.sh integration/scripts/integration-dynamic-backends-brig-index.sh integration/scripts/integration-dynamic-backends-db-schemas.sh integration/scripts/integration-dynamic-backends-s3.sh integration/scripts/integration-dynamic-backends-ses.sh integration/scripts/integration-dynamic-backends-vhosts.sh libs/http2-manager/test/resources/gen-certs.sh services/background-worker/fill_rabbit.sh services/background-worker/shutdown_test.sh services/federator/test/resources/unit/gen-certs.sh services/federator/test/resources/unit/gen-multidomain-certs.sh services/gen-aws-conf.sh services/proxy/test/scripts/proxy-test.sh tools/hlint.sh
INFO format | shellcheck: 46 file(s) processed in 602.981369ms
traversed 41030 files
emitted 41030 files for processing
formatted 46 files (0 changed) in 1.517s
```

I didn't want us to find weird problems in future so I did a a timeboxed stare at treefmt's code and found that [this code](https://github.com/numtide/treefmt/blob/1427cc2d98d5270f5720e5edac1f34e10c602876/walk/git.go#L54-L57) basically ignores any file which doesn't have the mode `100644`, which means it excludes executable files causing our shellscripts to not be found :facepalm:

akshaymankar:
Can we please not do the changes to treefmt args and instead do either of these things:
1. Wait for https://github.com/numtide/treefmt/pull/389 to be merged and released and fly without treefmt for shell scripts until then
2. Patch treefmt in nix with the bug fix

Thanks a lot for fixing all the shellscripts tho :heart:

brianmcgee:
I cut a new release, which includes the fix for this: https://github.com/numtide/treefmt/releases/tag/v2.0.5.

You can wait until it propagates through Hydra, or add `treefmt` as a source (looks like you're using `niv`)  and use the tag directly in the short-term: https://github.com/numtide/treefmt/releases/tag/v2.0.5.

hack/bin/cabal-run-tests.sh
```diff
@@ -14,8 +14,8 @@ fi
 for cabal in $(find "$TOP_LEVEL" -name "$pattern" | grep -v dist-newstyle); do
   # This is required because some tests (e.g. golden tests) must be run from
   # the package root.
-  cd "$(dirname $cabal)"
-  package="$(basename ${cabal%.*})"
+  cd "$(dirname "$cabal")"
+  package="$(basename "${cabal%.*}")"
```
julialongtin, line 18: this looks wrong to me. too many quotes.

hack/bin/create_test_user.sh
```diff
@@ -47,20 +47,21 @@ fi;
...
-        -d'{"email":"'$EMAIL'","password":"'$PASSWORD'","name":"demo"}')
+        -d'{"email":"'"$EMAIL"'","password":"'"$PASSWORD"'","name":"demo"}')
 
     UUID=$(echo "$CURL_OUT" | tail -1 | sed 's/.*\"id\":\"\([a-z0-9-]*\)\".*/\1/')
 
     if [ "$CSV" == "false" ]
-        then echo -e "Succesfully created a user with email: "$EMAIL" and password: "$PASSWORD
-        else echo -e $UUID","$EMAIL","$PASSWORD
+        then echo -e "Succesfully created a user with email: ""$EMAIL"" and password: ""$PASSWORD"
+        else echo -e "$UUID"",""$EMAIL"",""$PASSWORD"
```
julialongtin, line 65: do you really need to enter and exit the quoted state for the commas here? i don't believe they will be considered part of the variable names...

hack/bin/cabal-run-tests.sh
smatting: a bit intuitive having quotes inside quotes, but it is correct https://stackoverflow.com/questions/46376773/bash-is-subshell-output-implicitly-quoted

services/spar/test-scim-suite/run.sh
```diff
@@ -10,32 +10,37 @@ SCIM_TEST_SUITE_BRIG_PORT=8082
 function create_team_and_scim_token {
     TOP_LEVEL="$( cd "$( dirname "${BASH_SOURCE[0]}" )/../../.." && pwd )"
 
-    IFS=',' read -r -a creds <<< $($TOP_LEVEL/hack/bin/create_test_team_admins.sh -c)
+    IFS=',' read -r -a creds <<< "$("$TOP_LEVEL/hack/bin/create_test_team_admins.sh" -c)"
```
julialongtin, line 13: this looks sus to me, too.

julialongtin (changes requested):
some of these changes with quotes look suspicious to me...

smatting:
Randomly requested review from people I know that know bash

akshaymankar (approved):
:sparkles:

hack/bin/serve-charts.sh
```diff
@@ -2,16 +2,18 @@
...
+: "${HELM_SERVER_PORT:=4001}"
 
 # get rid of all helm repositories
 # We need to deal with helm repo list failing because of https://github.com/helm/helm/issues/10028
 (helm repo list -o json || echo '[]') | jq -r '.[] | .name' | xargs -I% helm repo remove %
 
-cd "$(dirname "$BASH_SOURCE[0]")/../../.local/charts"
-for chart in $@; do
+SCRIPT_DIR=$(cd -- "$(dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
+cd "$SCRIPT_DIR/../../.local/charts"
```
lwille, line 12: **thought:** This seems a bit convoluted to me: we temporarily change directory just to print its name, then cd into that again.

**suggestion:** I think using [`dirname`](https://explainshell.com/explain/1/dirname) and [`realpath`](https://explainshell.com/explain/1/realpath) should do the same thing, are easier to understand and thus common practice:

```suggestion
SCRIPT_DIR=$(realpath "$(dirname "${BASH_SOURCE[0]}")")
cd "$SCRIPT_DIR/../../.local/charts"
```

tools/nginz_disco/nginz_disco.sh
```diff
@@ -40,16 +41,19 @@ function valid_ipv6() {
 function upstream() {
     name=$1
     port=${2:-'8080'}
-    ips=$(dig +short +retries=3 +search ${name} | sort)
+    ips=$(dig +short +retries=3 +search "${name}" | sort)
     unset servers
-    for ip in ${ips[@]}; do
-        if valid_ipv4 $ip || valid_ipv6 $ip; then
+    for ip in "${ips[@]}"; do
```
lwille, line 46: **issue:** this is not equivalent. Test:
```
ips=$(dig +short +retries=3 +search "cloudflare.com" | sort) # ips has 2 elements
for ip in "${ips[@]}"; do echo $ip; done
```

this prints `104.16.132.229 104.16.133.229` in one line instead of two lines.

tools/sftd_disco/sftd_disco.sh
```diff
@@ -30,23 +28,23 @@ function valid_url() {
...
 function upstream() {
     name=$1
-    port=${2:-'8585'}
-    entries=$(dig +short +retries=3 +search SRV ${name} | sort)
+    entries=$(dig +short +retries=3 +search SRV "${name}" | sort)
     unset servers
     comma=""
     IFS=$'\n'
-    for entry in ${entries[@]}; do
+    for entry in "${entries[@]}"; do
```
lwille, line 35: **issue:** same thing here: the array is expanded inside the quotes, so you just get one `entry` with all records inside.

**suggestion:** don't quote `${entries[@]}`, however use quoting when using `$entry` (as you did down the line).

```suggestion
    for entry in ${entries[@]}; do
```

lwille (changes requested):
I might have found some issues, can you have a look?

hack/bin/serve-charts.sh
smatting: the solution is from https://stackoverflow.com/questions/59895/how-do-i-get-the-directory-where-a-bash-script-is-located-from-within-the-script It probably went through a lot of discussion, so I wouldn't wanna change it

tools/sftd_disco/sftd_disco.sh
smatting: Thank you for checking! No, actually array quoted arrays are still splitting, and shellcheck suggests it https://www.shellcheck.net/wiki/SC2068
The problem here is that entries is not array, but a string. Applied your suggestsions in https://github.com/wireapp/wire-server/pull/4220/commits/88b16c876cfb10ca69969812616476fd33112c13

PR #4242, opened 2024-09-16
[WPB-10772] Make it impossible for a user under legalhold to join an MLS conversation

MangoIV:
Two things to do: 
- [x] prevent a third party from fetching key packages from a user under legalhold 
- [x] prevent a user under legalhold to add itself to a group
## Checklist


MangoIV:
```haskell
-- ---------
-- WPB-10783
-- ---------
testMLSThenLegalhold :: (HasCallStack) => App ()
testMLSThenLegalhold = do
  -- scenario 1:
  -- if charlie is in any MLS conversation, he cannot approve to be put under legalhold
  (charlie, tid, []) <- createTeam OwnDomain 1
  [charlie1] <- traverse (createMLSClient def) [charlie]
  void $ createNewGroup charlie1
  void $ createAddCommit charlie1 [charlie] >>= sendAndConsumeCommitBundle

  legalholdWhitelistTeam tid charlie >>= assertStatus 200
  withMockServer def lhMockApp \lhDomAndPort _chan -> do
    postLegalHoldSettings tid charlie (mkLegalHoldSettings lhDomAndPort) >>= assertStatus 201
    requestLegalHoldDevice tid charlie charlie >>= assertSuccess
    approveLegalHoldDevice tid (charlie %. "qualified_id") defPassword
      `bindResponse` assertLabel 409 "mls-legalhold-not-allowed"
```

changelog.d/2-features/WPB-10772
```diff
@@ -0,0 +1,5 @@
+makes it impossible for a user to join an MLS conversation while already under legalhold (at least pending)
+
+This implies two things: 
+- if a user is under legalhold they cannot ever join an MLS conversation, not even an MLS self conversation 
+- a user has to reject an to be put under legalhold when it wants to join MLS conversation (not explicitly acknowledging is not enough)
```
pcapriotti, line 5: You're missing a word there.

libs/types-common/src/Data/Qualified.hs
```diff
@@ -127,6 +129,16 @@ foldQualified loc f g q
...
 
+pattern Local :: forall a x. forall. Local a -> (Local x, Qualified a)
+pattern Local loc <- ((\(loc, q) -> (tDomain loc == qDomain q, qUnqualified q <$ loc)) -> (True, loc))
+
+pattern Remote :: forall a x. forall. Remote a -> (Local x, Qualified a)
+pattern Remote loc <- ((\(loc, q) -> (tDomain loc == qDomain q, qTagUnsafe q)) -> (False, loc))
+
+{-# COMPLETE Local, Remote #-}
+
+-- pattern Remote :: Remote a -> (Local x, Qualified a)
```
pcapriotti, line 140: left over?

libs/types-common/src/Data/Qualified.hs
```diff
@@ -127,6 +129,16 @@ foldQualified loc f g q
   | otherwise =
       g (qTagUnsafe q)
 
+pattern Local :: forall a x. forall. Local a -> (Local x, Qualified a)
```
pcapriotti, line 132: Minor point, but would it make sense to use an ad-hoc data type here instead of a pair? Pattern matching on a pair and getting `Local` or `Remote` values is a bit strange. Maybe something like:
```haskell
case q `RelativeTo` loc of
  Local x -> ...
  Remote x -> ...
```
might be a bit more readable than the equivalent with a pair.

services/galley/src/Galley/API/MLS/Commit/InternalCommit.hs
```diff
@@ -162,6 +162,8 @@ processInternalCommit senderIdentity con lConvOrSub ciphersuite ciphersuiteUpdat
                       Right clientInfo -> do
                         let allClients = Set.map ciId clientInfo
                         let allMLSClients = Set.map ciId (Set.filter ciMLS clientInfo)
+                        -- TODO(mangoiv): check if any legalhold clients exist and abort if any
+
```
pcapriotti, line 166: left over?

pcapriotti (approved):
Looks good.

libs/types-common/src/Data/Qualified.hs
MangoIV: should it be `type RelativeTo = (,)`?

pcapriotti: I was thinking more `data RelativeTo a b = RelativeTo (Qualified a) (Local b)`.

MangoIV: ahhh maybe it's fine actually, yeah, let's go with that :)

PR #4244, opened 2024-09-17
Work around legacy integration test resource leak.

fisx:
## Checklist


pcapriotti (commented):
Shouldn't we find the resource leak instead of patching it like this?

pcapriotti (approved):
I guess this is fine, since these tests are legacy and maybe they will go away at some point.

PR #4251, opened 2024-09-18
[WPB-10708] personal account to own team

fisx:
Add brig endpoint `/upgrade-personal-to-team` to create a new team and make the user an owner. Sending confirmation emails will be done in a [followup PR](https://github.com/wireapp/wire-server/pull/4253).

https://wearezeta.atlassian.net/browse/WPB-10708

## Checklist

### DIFF
--- a/services/brig/src/Brig/API/User.hs
+++ b/services/brig/src/Brig/API/User.hs
@@ -20,6 +20,7 @@
 -- TODO: Move to Brig.User.Account
 module Brig.API.User
   ( -- * User Accounts / Profiles
+    upgradePersonalToTeam,
     createUser,
     createUserSpar,
     createUserInviteViaScim,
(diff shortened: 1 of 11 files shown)


fisx:
the ci failure is an unrelated flake, running again.

services/brig/src/Brig/API/User.hs
```diff
@@ -253,6 +256,43 @@ createUserSpar new = do
...
+    createUserTeam <- do
+      liftSem $ GalleyAPIAccess.createTeam uid (bnuTeam bNewTeam) tid
+      let BindingNewTeam newTeam = bNewTeam.bnuTeam
+      pure $ CreateUserTeam tid (fromRange (newTeam ^. newTeamName))
+
+    wrapClient $ updateUserTeam uid tid
+    liftSem $ Intra.sendUserEvent uid Nothing (teamUpdated uid tid)
+    initAccountFeatureConfig uid
+
+    pure $! createUserTeam
```
battermann, line 294: Do we need to think about what can go wrong here if this fails in the middle of it and gets retried?
AFAICT the critical point is after the team ID of the user is updated because then it cannot be re-triggered. In that case, events might not get sent and `initAccountFeatureConfig` might not get called. How bad is this?

If it fails after team is created but before user team is updated, and then retried, a re-trigger should work AFAICT, there will just be an unused zombie team lying around in the DB.

battermann (approved):
Only concern is about robustness/idempotency of the handler, see comment. Otherwise LGTM.

services/brig/src/Brig/API/User.hs
pcapriotti: The same questions would apply to the code that creates a team owner from scratch, no? The order of operations is pretty much the same. As usual, in the absence of proper atomicity and transactions, I'm not sure we can ever come up with good answers to these questions.

PR #4252, opened 2024-09-18
[chore] Replace pattern synomyn with ADT.

elland:
ADTs are simpler, easier to reason about and cover this specific case better, without us having to override completeness checks with pragmas.


supersven (approved):
Looks like an improvement to me :+1:

PR #4254, opened 2024-09-18
[chore] Removed implicit failures in favour of explicit error handling.

elland:


services/galley/src/Galley/API/MLS/Message.hs
```diff
@@ -222,22 +218,24 @@ postMLSCommitBundleToLocalConv qusr c conn bundle ctype lConvOrSubId = do
...
           -- should be a server error
-          Right resp <- runError @(Tagged TeamMemberNotFound ()) $ getUserStatus luid tid (tUnqualified luid)
-          case resp.ulhsrStatus of
-            UserLegalHoldPending -> throwS @MLSLegalholdIncompatible
-            UserLegalHoldEnabled -> throwS @MLSLegalholdIncompatible
-            UserLegalHoldDisabled -> pure ()
-            UserLegalHoldNoConsent -> pure ()
+          resp <- runError @(Tagged TeamMemberNotFound ()) $ getUserStatus luid tid (tUnqualified luid)
+          case resp of
+            Left _ -> throw $ InternalErrorWithDescription "Server error. Team member must have vanished with the legal hold check"
```
elland, line 233:
```suggestion
            Left _ -> throw $ InternalErrorWithDescription "Server error. Team member must have vanished during the legal hold check"
```

services/brig/src/Brig/API/MLS/KeyPackages.hs
```diff
@@ -124,17 +127,20 @@ claimLocalKeyPackages qusr skipOwn suite target = do
...
-        -- the user has to first reject to be put under legalhold
-        -- before they can join conversations again
-        case resp.ulhsrStatus of
-          UserLegalHoldPending -> throwE ClientLegalHoldIncompatible
-          UserLegalHoldEnabled -> throwE ClientLegalHoldIncompatible
-          UserLegalHoldDisabled -> pure ()
-          UserLegalHoldNoConsent -> pure ()
+      mSu <- lift $ liftSem $ getUser (tUnqualified target)
+      case mSu of
+        Nothing -> pure () -- Legalhold is a team feature.
```
elland, line 132:
```suggestion
        Nothing -> pure () -- No user, no conflict.
```

services/brig/src/Brig/API/MLS/KeyPackages.hs
```diff
@@ -124,17 +127,20 @@ claimLocalKeyPackages qusr skipOwn suite target = do
...
-        -- the user has to first reject to be put under legalhold
-        -- before they can join conversations again
-        case resp.ulhsrStatus of
-          UserLegalHoldPending -> throwE ClientLegalHoldIncompatible
-          UserLegalHoldEnabled -> throwE ClientLegalHoldIncompatible
-          UserLegalHoldDisabled -> pure ()
-          UserLegalHoldNoConsent -> pure ()
+      mSu <- lift $ liftSem $ getUser (tUnqualified target)
+      case mSu of
+        Nothing -> pure () -- No user, no conflict.
```
MangoIV, line 132: when is it a legal state to not have a stored user for a user id here?

services/galley/src/Galley/API/Federation.hs
```diff
@@ -629,17 +627,19 @@ sendMLSCommitBundle remoteDomain msr = handleMLSMessageErrors $ do
   (ctype, qConvOrSub) <- getConvFromGroupId ibundle.groupId
   when (qUnqualified qConvOrSub /= msr.convOrSubId) $ throwS @'MLSGroupConversationMismatch
   -- this cannot throw the error since we always pass the sender which is qualified to be remote
-  Just resp <-
-    runErrorS @MLSLegalholdIncompatible $
-      postMLSCommitBundle
-        loc
-        (tUntagged @QRemote sender)
```
MangoIV, line 636: this type application was placed here to prevent future changes from introducing errors. It is only safe to assume that we can discard the error when the sender is actually remote. since tUntagged works on local and remote, a future changed may go unchecked without this type application. 

also mind that while it's a minor thing, there's no need to convert `map` to `fmap`. it is more to type, more work for the specializer (longer compile times, possible missed specialisations and hence slower runtime, although the latter probably doesn't happen with `fmap`) and it obfuscates which type it is being applied to.

MangoIV (commented):
I think explicitly converting to an internal error is a nice touch. 
I don't think it makes things nicer to read with little to no benefit at runtime but that's a matter of taste.

services/brig/src/Brig/API/MLS/KeyPackages.hs
fisx: we might race an account deletion here. not likely, but doesn't do much damage to handle it, no?

services/galley/src/Galley/API/MLS/Message.hs
```diff
@@ -222,22 +218,24 @@ postMLSCommitBundleToLocalConv qusr c conn bundle ctype lConvOrSubId = do
...
           -- should be a server error
-          Right resp <- runError @(Tagged TeamMemberNotFound ()) $ getUserStatus luid tid (tUnqualified luid)
-          case resp.ulhsrStatus of
-            UserLegalHoldPending -> throwS @MLSLegalholdIncompatible
-            UserLegalHoldEnabled -> throwS @MLSLegalholdIncompatible
-            UserLegalHoldDisabled -> pure ()
-            UserLegalHoldNoConsent -> pure ()
+          resp <- runError @(Tagged TeamMemberNotFound ()) $ getUserStatus luid tid (tUnqualified luid)
+          case resp of
+            Left _ -> throw $ InternalErrorWithDescription "Server error. Team member must have vanished during the legal hold check"
```
fisx, line 233: i guess in the spirit of `Nothing -> pure ()` above, we could throw the more confusing but less loud MLSLegalholdIncompatible here?

but i am leaning towards @MangoIV and think internal errors that are supposed to be caught by tests shouldn't take up so much space in the code.

...  and in the comments of the related PRs, for that matter...  :-)  so i'll shut up.

services/galley/src/Galley/API/Federation.hs
elland: @MangoIV stealing your comment and adding it to the code 👍🏼

services/galley/src/Galley/API/MLS/Message.hs
elland: `MLSLegalholdIncompatible` is a semantic error, the internal error here is an internal error, right? State changed from under us midway through a function. :)

services/brig/src/Brig/API/MLS/KeyPackages.hs
```diff
@@ -121,20 +124,23 @@ claimLocalKeyPackages qusr skipOwn suite target = do
         uncurry (KeyPackageBundleEntry (tUntagged target) c)
           <$> wrapClientM (Data.claimKeyPackage target c suite)
 
-    assertUserNotLegalholded :: ExceptT ClientError (AppT r) ()
-    assertUserNotLegalholded = do
+    assertUserNotUnderLegalHold :: ExceptT ClientError (AppT r) ()
+    assertUserNotUnderLegalHold = do
```
fisx, line 128: `legalheld`?  :)

fisx (approved):

services/brig/src/Brig/API/MLS/KeyPackages.hs
MangoIV: Then it’s better to get a 500 here than some undefined behaviour from further down?

services/galley/src/Galley/API/MLS/Message.hs
MangoIV: > but i am leaning towards @MangoIV and think internal errors that are supposed to be caught by tests shouldn't take up so much space in the code.

this was exactly the intention my idea was that they should be explained in the code for whenever they occur and then not take much space at runtime or at compile time as they’re not part of the usual usage. :)

PR #4265, opened 2024-09-23
Fix feature flag defaults

pcapriotti:
This fixes the way the `DbFeature` is computed for `mlsMigration` and `enforceFileDownloadLocation`. Before, we were using a `null` cassandra value to represent both the absence of configuration and the presence of a `Nothing` value. The consequence was that the default would be accidentally overwritten by the `Nothing` value in some cases.

The fix is to use an explicit "null" value to represent `Nothing`. For dates, we use the POSIX timestamp origin. For the download location, we use an empty string.

Also, this PR contains a refactoring and reorganization of the test module for feature flags. It has been split into several modules, one per feature, and the various utilities have been simplified and made more uniform across features.

https://wearezeta.atlassian.net/browse/WPB-11100

## Checklist


charts/galley/values.yaml
```diff
@@ -93,6 +93,12 @@ config:
         defaults:
           lockStatus: unlocked
           status: enabled
+      enforceFileDownloadLocation:
+        defaults:
+          lockStatus: locked
+          status: disabled
+          config:
+            enforcedTimeoutSeconds: ""
```
fisx, line 101: why not `0`?

services/galley/src/Galley/API/Teams/Features.hs
```diff
@@ -428,6 +428,15 @@ instance SetFeatureConfig MlsMigrationConfig where
...
     pure feat
 
-instance SetFeatureConfig EnforceFileDownloadLocationConfig
+instance SetFeatureConfig EnforceFileDownloadLocationConfig where
+  type
+    SetFeatureForTeamConstraints EnforceFileDownloadLocationConfig r =
+      (Member (Error TeamFeatureError) r)
+
+  prepareFeature _ feat = do
+    -- empty download location is not allowed
```
fisx, line 437:
```suggestion
    -- empty download location is not allowed
    -- this is consistent with all other features, and least surprising for clients
```

fisx: i guess for the same reason this check should be added to the timestamps for `MlsMigrationConfig`?

fisx (approved):

charts/galley/values.yaml
pcapriotti: Oops, that's the wrong config (I must have copy-pasted from `selfDeletingMessages`). I've turned it into an empty config now.

PR #4268, opened 2024-09-25
[WPB-10659] Test notifications for personal user to team user migration

fisx:
Also contains small changes to the integration test suite:
1. Allow looking up elements from the end in JSON array using negative numbers. So something like `"notifications.-1"` will get the last notification.
2. Separate context from failure message in the `AssertionFailure`.  (I (@akshaymankar) wasn't aware it was appended to message earlier and implemented a separate one, then realized it was the case, so I removed appending to message part). 

## Checklist

### DIFF
--- a/integration/test/API/Galley.hs
+++ b/integration/test/API/Galley.hs
@@ -720,3 +720,9 @@ setTeamFeatureConfigVersioned versioned user team featureName payload = do
 -- | http://staging-nginz-https.zinfra.io/v6/api/swagger-ui/#/default/get_feature_configs
 getFeaturesForUser :: (HasCallStack, MakesValue user) => user -> App Response
 getFeaturesForUser user = baseRequest user Galley Versioned "feature-configs" >>= submit "GET"
+
+getTeamNotifications :: (HasCallStack, MakesValue user) => user -> Maybe String -> App Response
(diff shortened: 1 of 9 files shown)


integration/test/API/Galley.hs
```diff
@@ -720,3 +720,9 @@ setTeamFeatureConfigVersioned versioned user team featureName payload = do
 -- | http://staging-nginz-https.zinfra.io/v6/api/swagger-ui/#/default/get_feature_configs
 getFeaturesForUser :: (HasCallStack, MakesValue user) => user -> App Response
 getFeaturesForUser user = baseRequest user Galley Versioned "feature-configs" >>= submit "GET"
+
+getTeamNotifications :: (HasCallStack, MakesValue user) => user -> Maybe String -> App Response
```
fisx, line 724:
```suggestion
-- | https://staging-nginz-https.zinfra.io/v6/api/swagger-ui/#/default/get_teams_notifications
getTeamNotifications :: (HasCallStack, MakesValue user) => user -> Maybe String -> App Response
```

fisx (commented):
can't formally approve as it's my PR, but :+1: :shipit:

akshaymankar (approved):

PR #4272, opened 2024-09-26
[WPB-10308] Use RabbitMQ queues for notifications

battermann:
This is not to be merged before we make a release 5.6 per https://wearezeta.atlassian.net/browse/WPB-11239. The release happened on Oct 31, 2024: https://github.com/wireapp/wire-server/releases/tag/v2024-10-30

## Checklist

 - [x] find `error`, `undefined`, `TODO`, and fix them
 - [x] ~~brig currently loads a conn into Env and keeps it even after it dies. we should load an MVar into the env and replace broken chans in a top-level loop (like the old delete users queue in brig).~~ (this point was based on a misunderstanding.)


services/cannon/src/Cannon/Types.hs
```diff
@@ -117,6 +119,9 @@ wsenv = Cannon $ do
   r <- asks reqId
   pure $ WS.setRequestId r e
 
+stableRabbitmqConn :: Cannon (MVar (Maybe Connection))
+stableRabbitmqConn = Cannon $ asks stableRabbitmqConn_
```
akshaymankar, line 123: There is a limit on number of channels we can open per connection, so this solution will likely break because of that. I think it might be best to go back to separate connections per client for now because of two reasons:
- I think I misjudged that openning ~20k connections will break rabbitmq, perhaps we just need to run many nodes.
- For this we should probably implement some smart connection pooling, which would take some time and look very different from this, so perhaps we can split this work up like that.

integration/test/Test/Events.hs
```diff
@@ -0,0 +1,373 @@
...
+    x <- k eventsChan ackChan
+
+    -- Ensure all the acks are sent before closing the websocket
+    isAckChanEmpty <-
+      retrying
+        (limitRetries 5 <> constantDelay 10_000)
+        (\_ isEmpty -> pure $ not isEmpty)
+        (\_ -> atomically $ isEmptyTChan ackChan)
+    unless isAckChanEmpty $ do
+      putStrLn $ colored yellow $ "The ack chan is not empty after 50ms, some acks may not make it to the server"
```
mdimjasevic, line 275: Is this to stay in the test?

integration/test/Test/Events.hs
```diff
@@ -0,0 +1,373 @@
...
+
+    void $ tryPutMVar closeWS ()
+
+    timeout 1_000_000 (wait wsThread) >>= \case
+      Nothing ->
+        putStrLn $ colored yellow $ "The websocket thread did not close after waiting for 1s"
+      Just () -> pure ()
```
mdimjasevic, line 282: Is this to stay in the test?

services/gundeck/src/Gundeck/Push.hs
```diff
@@ -123,11 +145,105 @@ instance MonadMapAsync Gundeck where
...
+        rsingleton0 = rcast . rsingleton
+
+        rnil1 :: forall x. Range 0 1 [x]
+        rnil1 = rcast rnil
+
+        partitionToRange :: These a b -> (Range 0 1 [a], Range 0 1 [b])
+        partitionToRange = \case
+          (This a) -> (rsingleton0 a, rnil1)
+          (That b) -> (rnil1, rsingleton0 b)
+          (These a b) -> (rsingleton0 a, rsingleton0 b)
```
mdimjasevic, line 211: Do these belong in `Data.Range`?

mdimjasevic (approved):

PR #4273, opened 2024-09-26
Bump nixpkgs

smatting:
Motivated by https://github.com/wireapp/wire-server/pull/4220

## Checklist


smatting:
Building of hoogle docs fails because of 
```
hoogle> Using Parsec parser
hoogle> Configuring hoogle-5.0.18.4...
hoogle> CallStack (from HasCallStack):
hoogle>   withMetadata, called at libraries/Cabal/Cabal/src/Distribution/Simple/Utils.hs:368:14 in Cabal-3.10.3.0:Distribution.Simple.Utils
hoogle> Error: Setup: Encountered missing or private dependencies:
hoogle> crypton-connection ==0.3.2
```

The application however builds

nix/patches/hoogle.patch
MangoIV, line 1: Please use fetchPatch if possible.

nix/wire-server.nix
```diff
@@ -518,7 +518,7 @@ in
       pkgs.netcat
       pkgs.niv
       pkgs.haskellPackages.apply-refact
-      (pkgs.python310.withPackages
+      (pkgs.python3.withPackages
```
MangoIV, line 521: Have you tried building the docs? I remember there was something broken.

nix/patches/hoogle.patch
MangoIV: Why does doJailBreak not work? Build type custom?

nix/wire-server.nix
smatting: this is not the python that is used by the docs, it's not used by anything really. mainly a convience if you want to write python scripts in hack/bin

nix/patches/hoogle.patch
smatting: ah right, that amounts to the same thing. done in https://github.com/wireapp/wire-server/pull/4273/commits/2ffa78b1a8553b29366ebb7c9110ae959b565da2

MangoIV: there's some pathological cases where it doesn't work, thought you'd run into this with hoogle, I wouldn't have been surprised :P

MangoIV (approved):
we're good, I think

PR #4277, opened 2024-09-30
[WPB-9859] Support OAUTH SMTP authentication

stefanwire:
https://wearezeta.atlassian.net/browse/WPB-9859

## Checklist

### DIFF
--- a/services/brig/src/Brig/Options.hs
+++ b/services/brig/src/Brig/Options.hs
@@ -120,16 +120,16 @@ data EmailSMTPCredentials
   = -- | username and password file
     EmailSMTPBasicAuth !Text !FilePathSecrets
   | -- | username and token file
-    EmailSMTPXAUTH2 !Text !FilePathSecrets
+    EmailSMTPXOAUTH2 !Text !FilePathSecrets
   deriving (Show, Generic)
 
 instance FromJSON EmailSMTPCredentials where
   parseJSON = withObject "smtpCredentials" $ \v ->
     v .:? "smtpAuth" .!= ("basic" :: String) >>= \case
-      "xauth2" ->
-        EmailSMTPXAUTH2
+      "xoauth2" ->
+        EmailSMTPXOAUTH2
           <$> v .: "smtpUsername"
-          <*> v .: "smtpXAUTH2Token"
+          <*> v .: "smtpXOAUTH2ClientCredential"
       _ ->
         EmailSMTPBasicAuth
           <$> v .: "smtpUsername"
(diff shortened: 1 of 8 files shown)


services/brig/src/Brig/Options.hs
```diff
@@ -116,17 +116,24 @@ data EmailAWSOpts = EmailAWSOpts
...
+data EmailSMTPCredentials
+  = -- | username and password file
+    EmailSMTPBasicAuth !Text !FilePathSecrets
+  | -- | username and token file
+    EmailSMTPXAUTH2 !Text !FilePathSecrets
   deriving (Show, Generic)
 
-instance FromJSON EmailSMTPCredentials
+instance FromJSON EmailSMTPCredentials where
+  parseJSON = withObject "smtpCredentials" $ \v ->
```
fisx, line 127:
```suggestion
  parseJSON = withObject "EmailSMTPCredentials" $ \v ->
```

i like the habit of using the haskell type as description verbatim.

services/brig/src/Brig/Options.hs
```diff
@@ -116,17 +116,24 @@ data EmailAWSOpts = EmailAWSOpts
...
+    v .:? "smtpAuth" .!= ("basic" :: String) >>= \case
+      "xauth2" ->
+        EmailSMTPXAUTH2
+          <$> v .: "smtpUsername"
+          <*> v .: "smtpXAUTH2Token"
+      _ ->
+        EmailSMTPBasicAuth
+          <$> v .: "smtpUsername"
+          <*> v .: "smtpPassword"
```
fisx, line 136: this is fine, but you could also:

```suggestion
    v .:? "smtpAuth" >>= \case
      Just "xauth2" ->
        EmailSMTPXAUTH2
          <$> v .: "smtpUsername"
          <*> v .: "smtpXAUTH2Token"
      _ -> -- this covers the `Just "basicAuth"` case and the fallback for everything else.
        EmailSMTPBasicAuth
          <$> v .: "smtpUsername"
          <*> v .: "smtpPassword"
```

(not tested)

(or should there be a type error for everything but `Nothing` in the fallback rule?)

fisx (commented):
looks good so far i think (assuming you're adding what we just discussed later today).

i agree that we should also write a test in /integration. as an easy way to commit the test to the PR without breaking things when the credentials evaporate, we can deactivate the test by default, printing an info "if you want this test to run, you need to provide credentials in the source code and remove the `pending`").

stefanwire:
The ticket has been rejected.

PR #4280, opened 2024-10-02
[WPB-11101] Refactor invitation store

mdimjasevic:
This PR improves abstraction over DB interaction relating to team invitations. An internal type is not exposed anymore and some effect actions are updated as a result.

Tracked by https://wearezeta.atlassian.net/browse/WPB-11101.

## Checklist

### DIFF
--- a/libs/wire-subsystems/src/Wire/InvitationStore.hs
+++ b/libs/wire-subsystems/src/Wire/InvitationStore.hs
@@ -18,7 +18,7 @@
 {-# LANGUAGE StrictData #-}
 {-# LANGUAGE TemplateHaskell #-}
 
-module Wire.InvitationCodeStore where
+module Wire.InvitationStore where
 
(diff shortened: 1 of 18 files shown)


libs/wire-subsystems/src/Wire/InvitationStore.hs
```diff
@@ -18,25 +18,21 @@
 {-# LANGUAGE StrictData #-}
 {-# LANGUAGE TemplateHaskell #-}
 
-module Wire.InvitationCodeStore where
+module Wire.InvitationStore where
```
battermann, line 21: :+1:

battermann (approved):

libs/wire-subsystems/src/Wire/InvitationStore.hs
mdimjasevic: The store is not about codes, hence the name shouldn't have "code" in it.

PR #4282, opened 2024-10-07
Misc(tm)

fisx:
Add `make c` variant without treefmt; rm stack files; source comments with cross-references to redundant code.

(This PR has no ticket because it takes longer to author one than to just do this.)

## Checklist

 - [X] ~~Add a new entry in an appropriate subdirectory of `changelog.d`~~


Makefile
```diff
@@ -82,7 +82,10 @@ cabal.project.local:
 
 # Usage: make c package=brig test=1
 .PHONY: c
-c: treefmt
+c: treefmt c-fast
```
supersven, line 85: Now that I see this, I'm wondering if our approach to squeeze everything into one makefile target fits to the idea of `make`. I.e. if someone wants to compile and format in one go, why don't they call e.g. `make c format`? :thinking: 

(I'm fine with your change though, because your intention probably was to keep the dev's interface stable.)

supersven (approved):

PR #4292, opened 2024-10-11
[WPB-11368] Test for export team member CSV

battermann:
In preparation for https://github.com/wireapp/wire-server/pull/4293, this moves the CSV export test to the new integration suite.

https://wearezeta.atlassian.net/browse/WPB-11368

## Checklist

### DIFF
--- a/integration/test/Test/Teams.hs
+++ b/integration/test/Test/Teams.hs
@@ -286,12 +286,12 @@ testTeamMemberCsvExport = do
     handle <- randomHandle
     putHandle m handle >>= assertSuccess
     setField "handle" handle m
+      >>= setField "role" (if m == owner then "owner" else "member")
 
   memberMap :: Map.Map String Value <- fmap Map.fromList $ for (modifiedMembers) $ \m -> do
     uid <- m %. "id" & asString
     pure (uid, m)
 
-  print $ Map.keys memberMap
   bindResponse (getTeamMembersCsv owner tid) $ \resp -> do
     resp.status `shouldMatchInt` 200
     let rows = sort $ tail $ B8.lines $ resp.body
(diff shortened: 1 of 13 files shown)


integration/test/Test/Teams.hs
```diff
@@ -276,3 +280,51 @@ testUpgradePersonalToTeamAlreadyInATeam = do
...
+        now <- formatTime defaultTimeLocale "%Y-%m-%d" <$> liftIO getCurrentTime
+        take 10 (parseField 4) `shouldMatch` now
+        parseField 5 `shouldMatch` (ownerMember %. "handle")
+      parseField 7 `shouldMatch` "wire"
+      parseField 9 `shouldMatch` "foo"
+      parseField 12 `shouldMatch` (mem %. "num_clients")
+  where
+    unquote :: String -> String
+    unquote ('\'' : x) = x
+    unquote x = x
```
fisx, line 330: there could be more `'` in some of those strings.

```suggestion
    unquote :: String -> String
    unquote = init . tail
```

integration/test/Test/Teams.hs
```diff
@@ -276,3 +280,51 @@ testUpgradePersonalToTeamAlreadyInATeam = do
...
+      let parseField = unquote . read . B8.unpack . (cols !!)
+
+      parseField 0 `shouldMatch` (mem %. "name")
+      parseField 1 `shouldMatch` (mem %. "handle")
+      parseField 2 `shouldMatch` (mem %. "email")
+      role <- mem %. "role" & asString
+      parseField 3 `shouldMatch` role
+      when (role /= "owner") $ do
+        now <- formatTime defaultTimeLocale "%Y-%m-%d" <$> liftIO getCurrentTime
+        take 10 (parseField 4) `shouldMatch` now
```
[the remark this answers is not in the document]
pcapriotti: Yeah, that is a problem, but I hope it's rare enough that it doesn't matter.

integration/test/Test/Teams.hs
pcapriotti: I'm not sure what you're saying here. For whatever reason, we escape these values by adding a quote at the beginning when they start with some "forbidden character". It doesn't make a lot of sense to me, but that's what the exporter was doing before, so I didn't change it. This function simply undoes that escaping.

fisx: sorry, i was reading it wrong. all good!

fisx (approved):

PR #4293, opened 2024-10-14
Refactor CSV export

pcapriotti:
This adds `last_active` and `status` columns to the team CSV export. It also includes a refactoring of the export handler. I manually tested performance of the new handler, and for 500 users it's very similar to before despite the extra query. For larger teams it should be comparatively faster.

We have had a discussion on whether or not to make a new version of this endpoint. Adding columns at the end of the previous CSV format is quite unlikely to break scripts, so in the end I decided *not* to make a new version and keep things simple. In the future, we might consider further improving the export functionality by making it asynchronous for large teams, and possibly making columns user-selectable.

https://wearezeta.atlassian.net/browse/WPB-11368

## Checklist

### DIFF
--- a/services/galley/src/Galley/API/Teams/Export.hs
+++ b/services/galley/src/Galley/API/Teams/Export.hs
@@ -82,10 +82,7 @@ getUserRecord cache member = do
     let mFromInvitation = snd <$> member ^. invitation
     case mFromInvitation of
       Just ts -> pure $ Just ts
-      Nothing -> do
-        -- TODO: make this a single user query
-        suis <- Spar.lookupScimUserInfos [uid]
-        pure $ listToMaybe suis >>= suiCreatedOn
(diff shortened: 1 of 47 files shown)


services/galley/src/Galley/API/Teams/Export.hs
```diff
@@ -0,0 +1,180 @@
...
+  Sem r TeamExportUser
+getUserRecord cache member = do
+  let uid = member ^. userId
+  export <- getUserExportData uid >>= noteS @TeamMemberNotFound
+  mCreatedOn <- do
+    let mFromInvitation = snd <$> member ^. invitation
+    case mFromInvitation of
+      Just ts -> pure $ Just ts
+      Nothing -> do
+        -- TODO: make this a single user query
```
mdimjasevic, line 86: Is this still to be done or you've changed your mind?

PR #4296, opened 2024-10-17
[WPB-1220] servantify proxy internal

fisx:
Servantifying the whole of proxy would have been less awkward, but there are some questions as to which services to support, and for each of them servantification is non-trivial.

On the bright side Proxy.Run doesn't need to be touched again for servantivication of the rest.

## Checklist

### DIFF
--- a/libs/metrics-wai/src/Data/Metrics/Middleware/Prometheus.hs
+++ b/libs/metrics-wai/src/Data/Metrics/Middleware/Prometheus.hs
@@ -17,6 +17,7 @@
 
 module Data.Metrics.Middleware.Prometheus
   ( waiPrometheusMiddleware,
+    waiPrometheusMiddlewarePaths,
     normalizeWaiRequestRoute,
   )
 where
@@ -33,12 +34,17 @@ import Network.Wai.Routing.Route (Routes, prepare)
 -- This middleware requires your servers 'Routes' because it does some normalization
 -- (e.g. removing params from calls)
 waiPrometheusMiddleware :: (Monad m) => Routes a m b -> Wai.Middleware
-waiPrometheusMiddleware routes =
+waiPrometheusMiddleware routes = waiPrometheusMiddlewarePaths $ treeToPaths $ prepare routes
+
+-- | Helper function that should only be needed as long as we have wai-routing code left in
+-- proxy: run `treeToPaths` on old routing tables and `routeToPaths` on the servant ones, and
+-- feed both to this function.
+waiPrometheusMiddlewarePaths :: Paths -> Wai.Middleware
+waiPrometheusMiddlewarePaths paths =
   Promth.prometheus conf . instrument (normalizeWaiRequestRoute paths)
   where
     -- See Note [Raw Response]
     instrument = Promth.instrumentHandlerValueWithFilter Promth.ignoreRawResponses
-    paths = treeToPaths $ prepare routes
     conf =
       Promth.def
         { Promth.prometheusEndPoint = ["i", "metrics"],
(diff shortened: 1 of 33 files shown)


services/proxy/src/Proxy/API/Internal.hs
```diff
@@ -15,29 +15,19 @@
...
+import Proxy.Proxy qualified
+import Servant
+import Wire.API.Routes.MultiVerb
+import Wire.API.Routes.Named (Named (Named))
 
-sitemap :: Env -> Routes a Proxy ()
-sitemap e = do
-  Public.sitemap e
-  routesInternal
+type InternalAPI = Named "status" ("i" :> "status" :> MultiVerb 'GET '[Servant.JSON] '[RespondEmpty 200 "OK"] ())
```
supersven, line 30: Don't we need a route for the HTTP HEAD requests anymore? Or, is it implicitly generated? :thinking: 

The old version had:
```haskell
routesInternal = do
  head "/i/status" (continue $ const (pure empty)) true
...
```

fisx: yes, it's implicit.

libs/metrics-wai/src/Data/Metrics/Middleware/Prometheus.hs
```diff
@@ -33,12 +34,17 @@ import Network.Wai.Routing.Route (Routes, prepare)
 -- This middleware requires your servers 'Routes' because it does some normalization
 -- (e.g. removing params from calls)
 waiPrometheusMiddleware :: (Monad m) => Routes a m b -> Wai.Middleware
-waiPrometheusMiddleware routes =
+waiPrometheusMiddleware routes = waiPrometheusMiddlewarePaths $ treeToPaths $ prepare routes
```
supersven, line 37: I'm a bit surprised by our lack of test coverage for `proxy` :roll_eyes: 

Wouldn't this ticket have been a chance to make the beast testable and ensure we're not breaking anything? :thinking: (I may miss some details something, though.)

libs/metrics-wai/src/Data/Metrics/Middleware/Prometheus.hs
```diff
@@ -33,12 +34,17 @@ import Network.Wai.Routing.Route (Routes, prepare)
...
+
+-- | Helper function that should only be needed as long as we have wai-routing code left in
```
supersven, line 39: I'm wondering if such long Haddock titles are rendered nicely? :thinking:

libs/metrics-wai/src/Data/Metrics/Middleware/Prometheus.hs
fisx: i would argue "out of scope", this is only about internal end-points, and i only changed Run.hs because it seemed the most straight-forward way to do this at the time. it's tricky to test these without api keys for the resp. services. and the public routes really haven't changed in this PR, just wrapped a little.

but i agree to your point.  hm, what to do?

libs/types-common/src/Data/Id.hs
```diff
@@ -418,6 +419,9 @@ newtype RequestId = RequestId
       ToBytes
     )
 
+defRequestId :: RequestId
```
supersven, line 422: You may use this here as well: https://github.com/wireapp/wire-server/blob/develop/services/proxy/src/Proxy/Env.hs#L65

supersven: I'm wondering if `N/A` wouldn't be better communicated with a `Maybe RequestId`. :thinking: 

However, this is beyond the scope of this PR.

libs/metrics-wai/src/Data/Metrics/Middleware/Prometheus.hs
supersven: Argh, my fault, sorry!

Haddock has no notion of a title. (I wrongly thought these would be in the first line.)

supersven (approved):

PR #4299, opened 2024-10-22
WPB-11502 remove transitive-anns

battermann:
## Checklist


libs/wire-api/src/Wire/API/Routes/Internal/Galley.hs
```diff
@@ -57,8 +56,7 @@ import Wire.API.Team.SearchVisibility
 import Wire.API.User.Client
 
 type LegalHoldFeaturesStatusChangeFederatedCalls =
```
elland, line 58: Why is this type still necessary?

elland (approved):
👏

libs/wire-api/src/Wire/API/Routes/Internal/Galley.hs
battermann: It's not. Just a left over... will remove...

PR #4300, opened 2024-10-22
[chore] Weed out dead code buildup, add weeding step to sanitize-pr.

elland:
As a first step, this adds weeding to the sanitize-pr step, next step would be making it part of our CI linting step.

I also sneaked in a small change around using verification code subsystems, we've had the effect for a while, but it was unused. 

## Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -56,10 +56,10 @@ rabbit-clean:
 # Clean
 .PHONY: full-clean
 full-clean: clean
+	make rabbit-clean
 	rm -rf ~/.cache/hie-bios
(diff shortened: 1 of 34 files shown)


Makefile
```diff
@@ -56,10 +56,10 @@ rabbit-clean:
 # Clean
 .PHONY: full-clean
 full-clean: clean
+	make rabbit-clean
```
elland, line 59: We call it before we delete the env, otherwise rabbitmqadmin is not available.

fisx (approved):

PR #4309, opened 2024-10-25
Fix swagger

fisx:
dc889c0305eac14fd42a1b1ddd163fcf26271e54 is the interesting commit in this PR.  Tested this with and without the patch on a blocked version, and the fix does the right thing.

## Checklist


pcapriotti (approved):
Ah, so the specialisation type family was duplicated? Probably my mistake, sorry.

battermann (approved):

PR #4311, opened 2024-10-25
Move docs from docs.wire.com to generated helper page served by brig.

fisx:
Also updated docs.wire.com accordingly.

(I categorized this as "bug" because the data on docs.wire.com was out of date and the data on the brig page was incomplete.)

## Checklist


services/brig/src/Brig/API/Public.hs
```diff
@@ -244,17 +245,72 @@ versionedSwaggerDocsAPI Nothing = allroutes (throwError listAllVersionsResp)
...
+                  renderLink "swagger-ui" ("/" <> v <> "/api/swagger-ui") <> "; ",
+                  renderLink "swagger.json" ("/" <> v <> "/api/swagger.json"),
+                  "<br>"
+                ]
+                | v <- versionToLByteString <$> [minBound :: Version ..]
+              ]
+
+        internal :: [LByteString]
+        internal =
+          [ "<h2>Internal (not version-controlled)</h2>",
```
battermann, line 279:
```suggestion
          [ "<h2>Internal (not versioned)</h2>",
```
Version-controlled sounds like it refers to source code rather than an API.

battermann (approved):
A minor comment, but not so important.

PR #4316, opened 2024-10-29
[WPB-8881] Move email update and remove operations to effects

mdimjasevic:
This is a refactoring PR about user email update and remove operations.

Tracked by https://wearezeta.atlassian.net/browse/WPB-8881.

## Checklist

### DIFF
--- a/libs/wire-subsystems/src/Wire/UserStore.hs
+++ b/libs/wire-subsystems/src/Wire/UserStore.hs
@@ -74,6 +74,7 @@ data UserStore m a where
   GetActivityTimestamps :: UserId -> UserStore m [Maybe UTCTime]
   GetRichInfo :: UserId -> UserStore m (Maybe RichInfoAssocList)
   GetUserAuthenticationInfo :: UserId -> UserStore m (Maybe (Maybe Password, AccountStatus))
+  DeleteEmail :: UserId -> UserStore m ()
 
 makeSem ''UserStore
 
--- a/libs/wire-subsystems/src/Wire/UserStore.hs
+++ b/libs/wire-subsystems/src/Wire/UserStore.hs
@@ -54,6 +54,7 @@ data UserStore m a where
   GetIndexUsersPaginated :: Int32 -> Maybe PagingState -> UserStore m (PageWithState IndexUser)
   GetUsers :: [UserId] -> UserStore m [StoredUser]
   UpdateUser :: UserId -> StoredUserUpdate -> UserStore m ()
+  UpdateEmailUnvalidated :: UserId -> EmailAddress -> UserStore m ()
   UpdateUserHandleEither :: UserId -> StoredUserHandleUpdate -> UserStore m (Either StoredUserUpdateError ())
   DeleteUser :: User -> UserStore m ()
   -- | This operation looks up a handle but is guaranteed to not give you stale locks.
--- a/libs/wire-subsystems/src/Wire/UserSubsystem.hs
+++ b/libs/wire-subsystems/src/Wire/UserSubsystem.hs
@@ -201,9 +201,9 @@ getLocalUserAccountByUserKey :: (Member UserSubsystem r) => Local EmailKey -> Se
 getLocalUserAccountByUserKey q@(tUnqualified -> ek) =
   listToMaybe <$> getAccountsByEmailNoFilter (qualifyAs q [emailKeyOrig ek])
 
--- | Call 'changeEmail' and process result: if email changes to itself, succeed,
--- if not, send validation email.
-changeSelfEmail ::
+-- | Call 'createEmailChangeToken' and process result: if email changes to
+-- itself, succeed, if not, send validation email.
+requestEmailChange ::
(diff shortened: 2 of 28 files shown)


services/brig/src/Brig/API/Handler.hs
```diff
@@ -117,6 +121,13 @@ brigErrorHandlers logger reqId =
 -------------------------------------------------------------------------------
 -- Utilities
 
+-- | Run a 'UserSubsystemError' error and lift to the handler type.
+liftUserSubsystemError ::
+  Sem ((P.Error UserSubsystemError) : r) a ->
```
fisx, line 126: why do you want to eliminate this effect here?  shouldn't it be `(Member (P.Error userSubsystemError) r) => Sem r a -> Handler r a`?

libs/wire-subsystems/src/Wire/UserSubsystem.hs
```diff
@@ -160,6 +160,11 @@ data UserSubsystem m a where
   InternalUpdateSearchIndex :: UserId -> UserSubsystem m ()
   InternalFindTeamInvitation :: Maybe EmailKey -> InvitationCode -> UserSubsystem m StoredInvitation
   GetUserExportData :: UserId -> UserSubsystem m (Maybe TeamExportUser)
+  -- | Throws errors in the interpreter.
+  RemoveEmail :: Local UserId -> UserSubsystem m ()
```
fisx, line 164: i would make this a helper function that uses the UserSubsystem effect constraint to call RemoveEmailEither.  i can't think of a reason to keep the flexibility here and pay with the extra complexity inside the effect.  plus it's not going to hard to change it back in the future, should the need arise.

also it is more modular: the error that `RemoveEmail` throws won't depend on the UserSubsystem interpreter, but on whichever interpreter is plugged in to throw.  (not sure i'm making sense to myself, but i think the point i'm trying to make is valid.  maybe let's call on monday?)

services/brig/src/Brig/API/Handler.hs
mdimjasevic: We improved this by dropping the helper.

libs/wire-subsystems/src/Wire/UserSubsystem.hs
mdimjasevic: I believe the changes to it are in line with what you had in mind, but please confirm.

libs/wire-subsystems/test/unit/Wire/MockInterpreters/UserStore.hs
```diff
@@ -74,6 +81,7 @@ inMemoryUserStoreInterpreter = interpret $ \case
   GetActivityTimestamps _ -> pure []
   GetRichInfo _ -> error "rich info not implemented"
   GetUserAuthenticationInfo _uid -> error "Not implemented"
+  DeleteEmail _uid -> error "Not implemented"
```
fisx, line 84: should we write tests that fail because of these?

libs/wire-subsystems/src/Wire/UserStore.hs
```diff
@@ -54,6 +54,7 @@ data UserStore m a where
   GetIndexUsersPaginated :: Int32 -> Maybe PagingState -> UserStore m (PageWithState IndexUser)
   GetUsers :: [UserId] -> UserStore m [StoredUser]
   UpdateUser :: UserId -> StoredUserUpdate -> UserStore m ()
+  UpdateEmailUnvalidated :: UserId -> EmailAddress -> UserStore m ()
```
fisx, line 57: do we need a DeleteEmailUnvalidated?  or does this happen via TTL?  or implicitly when something else happens?

fisx (approved):

services/brig/src/Brig/API/Internal.hs
```diff
@@ -554,19 +559,29 @@ changeSelfEmailMaybeSend ::
...
   ) =>
   UserId ->
   MaybeSendEmail ->
   EmailAddress ->
   UpdateOriginType ->
   (Handler r) ChangeEmailResponse
 changeSelfEmailMaybeSend u ActuallySendEmail email allowScim = do
-  API.changeSelfEmail u email allowScim
+  lusr <- qualifyLocal u
+  timeout <- asks (.settings.activationTimeout)
```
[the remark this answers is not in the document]
mdimjasevic: Has this been resolved by commit a1648f1ac?

libs/wire-subsystems/src/Wire/UserStore.hs
fisx: ah, see `updateEmailAndDeleteEmailUnvalidate` and other places.  we do delete unvalidated emails, just not in the subsystems yet.

libs/wire-subsystems/test/unit/Wire/MockInterpreters/UserStore.hs
fisx: separate PR, see jira ticket.

PR #4317, opened 2024-10-30
 Detect and flag missed notifications from RabbitMQ

elland:
https://wearezeta.atlassian.net/browse/WPB-11167

## Checklist


services/background-worker/src/Wire/BackendDeadUserNotificationWatcher.hs
```diff
@@ -0,0 +1,98 @@
...
+  -- This ensures that we receive notifications 1 by 1 which ensures they are
+  -- delivered in order.
+  lift $ Q.qos chan 0 1 False
+
+  -- TODO: replace bare string with the constant from the rabbitmq PR.
+  let queueName = "dead-user-notifications"
+
+  void . lift $ Q.declareQueue chan Q.newQueue {Q.queueName = queueName}
+  void $ QL.consumeMsgs chan queueName Q.Ack $ \(_msg, envelope) ->
+    let vals = Text.splitOn "." envelope.envRoutingKey
```
akshaymankar, line 37: Might be best to use the `x-first-death-queue` header instead of the routing key, this way we don't have to worry about what routing key the message was published as.

services/background-worker/src/Wire/BackendDeadUserNotificationWatcher.hs
```diff
@@ -0,0 +1,98 @@
...
+     in case vals of
+          [uidText, cidText] -> do
+            m <- runMaybeT $ do
+              uid <- hoistMaybe $ hush $ parseIdFromText uidText
+              cid <- hoistMaybe $ fromByteString $ Text.encodeUtf8 cidText
+              pure (uid, cid)
+            (uid, cid) <- maybe handleParseErrors pure m
+            markAsNeedsFullSync cassandra uid cid
```
akshaymankar, line 45: We're likely to get transient notifications here, we should ignore them.

services/background-worker/src/Wire/BackendDeadUserNotificationWatcher.hs
```diff
@@ -0,0 +1,98 @@
...
+
+startWatcher :: DomainConsumerMap -> Q.Channel -> AppT IO ()
+startWatcher _consumersRef chan = do
+  markAsWorking BackendDeadUserNoticationWatcher
+
+  cassandra <- asks (.cassandra)
+
+  -- This ensures that we receive notifications 1 by 1 which ensures they are
+  -- delivered in order.
+  lift $ Q.qos chan 0 1 False
```
akshaymankar, line 30: Why is the order important here? This would definitely affect the throughput.

services/background-worker/src/Wire/BackendDeadUserNotificationWatcher.hs
pcapriotti: Yeah, that's what we're doing now.

services/background-worker/src/Wire/BackendDeadUserNotificationWatcher.hs
pcapriotti: What do you mean?

akshaymankar: We push transient messages with ttl of 0. This probably means that most of them get dead lettered, unless there is an active consumer at that point. But we shouldn't count these as missed notifications because they're meant to be missed if the client is not online. 

Another idea would be to make them go to some other DLX which just drops them, so we don't have to worry about it here.

They're pushed here: https://github.com/wireapp/wire-server/blob/b74b7e67abd4f3f603c3b7b956e535577e4e853e/services/gundeck/src/Gundeck/Push.hs#L296-L307

services/background-worker/src/Wire/BackendDeadUserNotificationWatcher.hs
elland: Leftover from copying from the other bg worker. Gone now.

mdimjasevic (approved):
Please add a changelog.

:rocket:

elland:
Let's add the changelog to the parent.

PR #4324, opened 2024-10-31
[WPB-11196] Generate schemas for <service>.yaml

fisx:
## Checklist

### DIFF
--- a/libs/extended/src/Servant/API/Extended.hs
+++ b/libs/extended/src/Servant/API/Extended.hs
@@ -19,14 +19,18 @@
 -- errors instead of plaintext.
 module Servant.API.Extended where
 
+import Data.Bifunctor
 import Data.ByteString
 import Data.ByteString.Lazy qualified as BL
 import Data.EitherR (fmapL)
 import Data.Kind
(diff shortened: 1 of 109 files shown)


libs/extended/src/Servant/API/Extended.hs
```diff
@@ -116,3 +120,18 @@ instance
...
+    "application"
+      M.// "yaml"
+      M./: ("charset", "utf-8")
+      NE.:| ["application" M.// "yaml"]
+
+instance {-# OVERLAPPABLE #-} (ToJSON a) => MimeRender YAML a where
+  mimeRender _ = fromStrict . Y.encode
+
+instance {-# OVERLAPPABLE #-} (FromJSON a) => MimeUnrender YAML a where
+  mimeUnrender _ = first show . Y.decodeEither' . toStrict
```
MangoIV, line 137: Are you sure you want to make this `OVERLAPPABLE`? I think making every instance that would overlap `OVERLAPPING` is the better UX as you wouldn't want to *accidentally* write a new instance for more specific type.

fisx:
stale, closed.

PR #4329, opened 2024-11-04
integration: Allow MLS State to track multiple conversations

akshaymankar:
## Checklist

### DIFF
--- a/integration/test/Test/MLS.hs
+++ b/integration/test/Test/MLS.hs
@@ -165,20 +165,22 @@ testMixedProtocolAddUsers secondDomain suite = do
   [bob, charlie] <- replicateM 2 (randomUser secondDomain def)
   connectUsers [alice, bob, charlie]
 
-  convId <-
-    postConversation alice defProteus {qualifiedUsers = [bob], team = Just tid}
-      >>= getJSON 201
-      >>= objConvId
+  convId <- do
+    convId <-
+      postConversation alice defProteus {qualifiedUsers = [bob], team = Just tid}
+        >>= getJSON 201
+        >>= objConvId
 
-  bindResponse (putConversationProtocol bob convId "mixed") $ \resp -> do
-    resp.status `shouldMatchInt` 200
+    bindResponse (putConversationProtocol bob convId "mixed") $ \resp -> do
+      resp.status `shouldMatchInt` 200
+
+    bindResponse (getConversation alice (convIdToQidObject convId)) $ \resp -> do
+      resp.status `shouldMatchInt` 200
+      resp.json %. "epoch" `shouldMatchInt` 0
+      objConvId resp.json
 
   [alice1, bob1] <- traverse (createMLSClient suite def) [alice, bob]
-
-  bindResponse (getConversation alice (convIdToQidObject convId)) $ \resp -> do
-    resp.status `shouldMatchInt` 200
-    resp.json %. "epoch" `shouldMatchInt` 0
-    createGroup suite alice1 convId
+  createGroup suite alice1 convId
 
   void $ uploadNewKeyPackage suite bob1
 
@@ -201,20 +203,21 @@ testMixedProtocolUserLeaves secondDomain = do
   bob <- randomUser secondDomain def
   connectUsers [alice, bob]
 
-  convId <-
-    postConversation alice defProteus {qualifiedUsers = [bob], team = Just tid}
-      >>= getJSON 201
-      >>= objConvId
+  convId <- do
+    convId <-
+      postConversation alice defProteus {qualifiedUsers = [bob], team = Just tid}
+        >>= getJSON 201
+        >>= objConvId
 
-  bindResponse (putConversationProtocol bob convId "mixed") $ \resp -> do
-    resp.status `shouldMatchInt` 200
+    bindResponse (putConversationProtocol bob convId "mixed") $ \resp -> do
+      resp.status `shouldMatchInt` 200
+
+    bindResponse (getConversation alice (convIdToQidObject convId)) $ \resp -> do
+      resp.status `shouldMatchInt` 200
+      objConvId resp.json
 
   [alice1, bob1] <- traverse (createMLSClient def def) [alice, bob]
-
-  bindResponse (getConversation alice (convIdToQidObject convId)) $ \resp -> do
-    resp.status `shouldMatchInt` 200
-    createGroup def alice1 convId
-
+  createGroup def alice1 convId
   void $ uploadNewKeyPackage def bob1
 
   mp <- createAddCommit alice1 convId [bob]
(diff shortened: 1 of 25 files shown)


pcapriotti:
I haven't looked at this in detail, but I have a few questions.

What is the motivation for having multiple conversations? Conversations are mostly independent from one another, so any test that contains two conversations can always be written in a way where the two conversations exist one after the other.

Also, there seem to be multiple changes here that are not related to allowing multiple conversations. Why are we switching to a specific type for conversation parameters, instead of keeping with the existing conversion based on JSON values? Why is the ciphersuite being passed explicitly instead of using the one in the state?

akshaymankar:
> What is the motivation for having multiple conversations? Conversations are mostly independent from one another, so any test that contains two conversations can always be written in a way where the two conversations exist one after the other.

In performance tests we'd like to setup many users and then create different conversations with the same set of users. Having this assumption and explicit state management was getting a bit too much, so we decided to do this.

> Why are we switching to a specific type for conversation parameters, instead of keeping with the existing conversion based on JSON values?

The JSON values could be either a subconversation, a conversation or a domain-id pair in object form, this was getting a bit too confusing to be used as the key for the map in the state, so we decided to create the `ConvId` type.

> Why is the ciphersuite being passed explicitly instead of using the one in the state?

We figured it was less confusing than to call `modifyMLSState` in the tests, sometimes some Util functions were doing this and keeping all that in mind while trying to make sense of the assertions was getting a bit confusing.

integration/test/Test/MLS/One2One.hs
```diff
@@ -316,18 +321,19 @@ one2OneScenarioConvDomain One2OneScenarioRemoteConv = OtherDomain
 
 testMLSOne2One :: (HasCallStack) => Ciphersuite -> One2OneScenario -> App ()
 testMLSOne2One suite scenario = do
-  setMLSCiphersuite suite
+  -- setMLSCiphersuite suite
```
mdimjasevic, line 324: A leftover.

integration/test/Test/MLS.hs
```diff
@@ -416,115 +433,115 @@ testRemoteAddUser = do
 
 testRemoteRemoveClient :: (HasCallStack) => Ciphersuite -> App ()
 testRemoteRemoveClient suite = do
-  setMLSCiphersuite suite
+  -- setMLSCiphersuite suite
```
mdimjasevic, line 436: A leftover.

integration/test/Test/MLS.hs
```diff
@@ -416,115 +433,115 @@ testRemoteAddUser = do
...
-    parsedMsg <- showMessage alice1 mlsMsg
+    parsedMsg <- showMessage suite alice1 mlsMsg
     let leafIndexBob = 1
     parsedMsg %. "message.content.body.Proposal.Remove.removed" `shouldMatchInt` leafIndexBob
     parsedMsg %. "message.content.sender.External" `shouldMatchInt` 0
 
 testRemoteRemoveCreatorClient :: (HasCallStack) => Ciphersuite -> App ()
 testRemoteRemoveCreatorClient suite = do
-  setMLSCiphersuite suite
+  -- setMLSCiphersuite suite
```
mdimjasevic, line 465: A leftover.

integration/test/Test/MLS.hs
```diff
@@ -567,30 +584,31 @@ testRemoveClientsIncomplete :: (HasCallStack) => App ()
...
+  void $ createAddCommit alice1 convId [bob] >>= sendAndConsumeCommitBundle
+  mp <- createRemoveCommit alice1 convId [bob1]
 
   err <- postMLSCommitBundle mp.sender (mkBundle mp) >>= getJSON 409
   err %. "label" `shouldMatch` "mls-client-mismatch"
 
 testAdminRemovesUserFromConv :: (HasCallStack) => Ciphersuite -> App ()
 testAdminRemovesUserFromConv suite = do
-  setMLSCiphersuite suite
+  -- setMLSCiphersuite suite
```
mdimjasevic, line 598: A leftover.

integration/test/Test/MLS/SubConversation.hs
```diff
@@ -184,150 +192,167 @@ testLeaveSubConv leaver = do
...
-      mems `shouldMatchSet` ((renameField "id" "user_id" <=< make) `traverse` [alice1, charlie1, charlie2])
+  addUsersToFailureContext [("alice", alice), ("bob", bob), ("charlie", charlie)] $ do
+    [alice1, bob1, bob2, charlie1, charlie2] <- traverse (createMLSClient def def) [alice, bob, bob, charlie, charlie]
+    traverse_ (uploadNewKeyPackage def) [bob1, bob2, charlie1, charlie2]
+    convId <- createNewGroup def alice1
+
+    _ <- createAddCommit alice1 convId [bob, charlie] >>= sendAndConsumeCommitBundle
+
+    -- save the state of the parent group
+    -- parentState <- getMLSState
```
mdimjasevic, line 247: Is this to be removed?

integration/test/Test/MLS/SubConversation.hs
```diff
@@ -184,150 +192,167 @@ testLeaveSubConv leaver = do
...
+    -- switch to the subgroup
+    let subConvName = "conference"
+    createSubConv def convId alice1 subConvName
+    subConvId <- getSubConvId alice convId "conference"
+
+    for_ [bob1, bob2, charlie1, charlie2] \c ->
+      createExternalCommit subConvId c Nothing >>= sendAndConsumeCommitBundle
+    -- save the state of the subgroup and switch to the parent context
+    -- childState <- getMLSState <* setMLSState parentState
```
mdimjasevic, line 256: Is this to be removed?

integration/test/Test/MLS/SubConversation.hs
```diff
@@ -184,150 +192,167 @@ testLeaveSubConv leaver = do
...
+      removeCommitEvents %. "events.0.type" `shouldMatch` "conversation.member-leave"
+      removeCommitEvents %. "events.0.data.reason" `shouldMatch` "removed"
+      removeCommitEvents %. "events.0.from" `shouldMatch` alice1.user
+
+      for_ wss \ws -> do
+        n <- awaitMatch isConvLeaveNotif ws
+        n %. "payload.0.data.reason" `shouldMatch` "removed"
+        n %. "payload.0.from" `shouldMatch` alice1.user
+
+      -- setMLSState childState
```
mdimjasevic, line 277: A leftover.

mdimjasevic (approved):
This looks good! There are a few lines with old commented out code in the tests that should be removed.

integration/test/Testlib/Run.hs
```diff
@@ -112,7 +114,7 @@ runTests tests mXMLOutput cfg = do
   runCodensity (mkGlobalEnv cfg) $ \genv ->
     withAsync displayOutput $ \displayThread -> do
       -- Currently 4 seems to be stable, more seems to create more timeouts.
-      report <- fmap mconcat $ pooledForConcurrentlyN 4 tests $ \(qname, _, _, action) -> do
+      report <- fmap mconcat $ pooledForConcurrentlyN 16 tests $ \(qname, _, _, action) -> do
```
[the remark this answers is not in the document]
akshaymankar: Oh yeah this is unintended, I made it 16 to make it faster on my computer.

integration/test/Testlib/ModService.hs
```diff
@@ -338,6 +342,22 @@ waitUntilServiceIsUp domain srv =
...
+        eith <- liftIO (E.try checkStatus)
+        pure $ either (\(_e :: HTTP.HttpException) -> False) id eith
+      externalStatus = do
+        checkStatus <- appToIO $ do
+          extPort <- fromIntegral . (.port) . (.federatorExternal) <$> getServiceMap domain
+          res <- submit "GET" req {HTTP.port = extPort}
+          pure (res.status `elem` [200, 204])
+        eith <- liftIO (E.try checkStatus)
+        pure $ either (\(_e :: HTTP.HttpException) -> False) id eith
+  (&&) <$> internalStatus <*> externalStatus
```
akshaymankar, line 360: I had added this out of desperation yesterday, it is not needed.

PR #4336, opened 2024-11-11
Preserve old invitation behaviour in v6

pcapriotti:
This PR introduces a new field `allowExisting` in `InvitationRequest`, which is true by default in v7, but false by default in earlier versions. The field controls whether an invitation follows the new behaviour of sending team invitations to existing users.

https://wearezeta.atlassian.net/browse/WPB-12177

## Checklist


battermann (approved):
LGTM

PR #4342, opened 2024-11-18
Drain rabbitmq consumers slowly from Cannon

elland:
https://wearezeta.atlassian.net/browse/WPB-11458

In case Cannon gets terminated, we want to drain the consumers, like we do with websocket queues.

## Checklist


services/cannon/src/Cannon/Run.hs
```diff
@@ -76,8 +77,10 @@ run o = withTracer \tracer -> do
   cassandra <- defInitCassandra (o ^. cassandraOpts) g
   e <-
     mkEnv ext o cassandra g
-      <$> D.empty 128
-      <*> newManager defaultManagerSettings {managerConnCount = 128}
+      <$> D.empty connectionLimit
+      -- TODO: think about how we're storing these connections, perhaps list is not efficient enough
```
mdimjasevic, line 81: TODO?

services/cannon/src/Cannon/RabbitMqConsumerApp.hs
```diff
@@ -3,22 +3,78 @@
...
+  cancel timeoutAction
+  Log.info e.logg $ Log.msg (Log.val "Draining complete")
+  where
+    closeConn :: Log.Logger -> Key -> Q.Connection -> IO ()
+    closeConn l key conn = do
+      Log.info l $
+        Log.msg (Log.val "closing rabbitmq connection")
+          . Log.field "key" (show key)
+      Q.closeConnection conn
+      void $ D.remove key e.rabbitConnections
```
mdimjasevic, line 62: What happens if Cannon crashes in between these two?

elland: If it completely crashes, then there's no issue, it's not like the in-memory data will be preserved. Since this is bookkeeping for the graceful termination. If it starts graceful termination between these two steps, the worst case is trying to close a connection twice, which should be a no-op.

PR #4356, opened 2024-12-03
[WPB-11973] Freeze API version 7, create dev API version 8.

fisx:
## Checklist

 - [x] Add a new entry in an appropriate subdirectory of `changelog.d`
 - [x] Read and follow https://docs.wire.com/developer/developer/api-versioning.html#version-bump-checklist
 - [x] Read and follow the [PR guidelines](https://docs.wire.com/developer/developer/pr-guidelines.html)

PR #4363, opened 2024-12-05
[WPB-10324] Drop legacy notification endpoints

mdimjasevic:
Tracked by https://wearezeta.atlassian.net/browse/WPB-10324.

## Checklist


changelog.d/1-api-changes/WPB-10324
```diff
@@ -0,0 +1 @@
+Drop legacy notification endpoints from client API at version 9
```
fisx, line 1: did you also create the confluence page and add it there? seems like a good idea to me, especially since we might all completely forget by the time this becomes relevant.

integration/test/Test/Version.hs
```diff
@@ -43,9 +45,11 @@ testVersion (Versioned' v) = withModifiedBackend
       domain <- resp.json %. "domain" & asString
       federation <- resp.json %. "federation" & asBool
 
-      -- currently there is only one development version
-      -- it is however theoretically possible to have multiple development versions
-      length dev `shouldMatchInt` 1
+      -- currently there are three development versions
+      --
+      -- it is however theoretically possible to have a different number of
```
fisx, line 50:
```suggestion
      -- it is however theoretically (and practically) possible to have a different number of
```

libs/wire-api/src/Wire/API/Routes/Version.hs
```diff
@@ -100,6 +100,8 @@ versionInt V4 = 4
 versionInt V5 = 5
 versionInt V6 = 6
 versionInt V7 = 7
+versionInt V8 = 8
+versionInt V9 = 9
```
fisx, line 104: this will conflict with https://github.com/wireapp/wire-server/pull/4356

but maybe that's fine?

changelog.d/1-api-changes/WPB-10324
mdimjasevic: I haven't yet, as I am still not 100% sure what version this should be reflected in. The latest discussion in the ticket suggests this won't be done in version 9, but in a much later version and perhaps this PR can be closed, but let's wait and see how the discussion evolves.

libs/wire-api/src/Wire/API/Routes/Version.hs
mdimjasevic: Yes, that's totally fine. I am aware of this. There's no way to avoid conflicts if we work in parallel on related stuff. I see the PR has been merged in the PR so I'll rebase this PR to resolve the conflicts.

PR #4366, opened 2024-12-10
ClientAPI: Include unnesting of capabilties in V7, ensure consumable notifs are not part of the enum

akshaymankar:
- [x] The V7 swagger seems to have changed more than I expected so this needs to be looked into.
        This was because v6 swagger was committed as v7 by mistake.

## Checklist


services/brig/docs/swagger-v7.json
```diff
@@ -1998,6 +1763,206 @@
...
+            "type": "string"
+          }
+        },
+        "required": [
+          "team_id",
+          "team_name"
+        ],
+        "type": "object"
+      },
+      "Currency.Alpha": {
```
fisx, line 1781: why is it Currency.Alpha now and was Alpha before?

did the version of any relevant dependency change?  or did the namespace change so that this needs to be better qualified now?  (is qualifying even a thing here?)

and why did this happen now, and now in https://github.com/wireapp/wire-server/pull/4356 ?

fisx (commented):
it all looks pretty reasonable to me except for the swagger output.  hm...

integration/test/Test/Swagger.hs
```diff
@@ -127,3 +129,21 @@ testSwaggerLint sw = do
...
+        liftIO $ writeFile (tempdir </> filename <> ".json") content
+
+  for_ existingVersions $ \v -> do
+    bindResponse (BrigP.getSwaggerPublicAllJson v) $ \resp -> do
+      resp.status `shouldMatchInt` 200
+      writeSwagger (show v) resp
+  for_ (Set.toList internalApis) $ \a -> do
+    bindResponse (BrigP.getSwaggerInternalJson a) $ \resp -> do
+      resp.status `shouldMatchInt` 200
+      writeSwagger a resp
```
akshaymankar, line 149: This is not a real test, let's not add things in integration tests just because we need to execute stuff. But let's discuss this in #4367, which wouldn't block the release?

fisx (approved):

PR #4367, opened 2024-12-10
brig: Add endpoints to show real swagger

akshaymankar:
## Checklist

### DIFF
--- a/services/brig/src/Brig/API/Public.hs
+++ b/services/brig/src/Brig/API/Public.hs
@@ -95,6 +95,7 @@ import Polysemy.Input (Input)
 import Polysemy.TinyLog (TinyLog)
 import Servant hiding (Handler, JSON, addHeader, respond)
 import Servant qualified
+import Servant.OpenApi (HasOpenApi)
 import Servant.OpenApi.Internal.Orphans ()
 import Servant.Swagger.UI
 import System.Logger.Class qualified as Log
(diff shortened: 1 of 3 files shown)


services/brig/src/Brig/API/Public.hs
```diff
@@ -314,6 +318,46 @@ versionedSwaggerDocsAPI Nothing = allroutes (throwError listAllVersionsResp)
...
+        <> serviceSwagger @SparAPITag @v
+        <> serviceSwagger @CargoholdAPITag @v
+        <> serviceSwagger @CannonAPITag @v
+        <> serviceSwagger @GundeckAPITag @v
+        <> serviceSwagger @ProxyAPITag @v
+        <> serviceSwagger @OAuthAPITag @v
+    )
+      & S.info . S.title .~ "Wire-Server API"
+      & S.info . S.description ?~ $(embedText =<< makeRelativeToProject "docs/swagger.md")
+      & S.servers .~ [S.Server ("/" <> toUrlPiece V8) Nothing mempty]
```
fisx, line 358:
```suggestion
      & S.servers .~ [S.Server ("/" <> toUrlPiece v) Nothing mempty]
```

services/brig/src/Brig/API/Public/Swagger.hs
```diff
@@ -41,6 +42,8 @@ type SwaggerDocsAPIBase = SwaggerSchemaUI "swagger-ui" "swagger.json"
 
 type VersionedSwaggerDocsAPI = "api" :> Header VersionHeader VersionNumber :> SwaggerDocsAPIBase
 
+type RealVersionedSwaggerDocsAPI = "real-api" :> Header' '[Required, Servant.Strict] VersionHeader VersionNumber :> SwaggerDocsAPIBase
```
fisx, line 45: i like the idea, but i'm wondering if we should then even expose the unreal-api? instead we could move the frozen swagger files to unit tests and make sure things don't change in frozen versions, but always generate them dynamically.

i guess storing the files is faster, but how busy are these end-points?

fisx: also note that real-api will differ from api in old versions because missing lower version bounds.

PR #4373, opened 2024-12-13
switch default ciphersuite to 2

stefanwire:
https://wearezeta.atlassian.net/browse/WPB-15004

## Checklist

### DIFF
--- a/integration/test/Test/MLS/KeyPackage.hs
+++ b/integration/test/Test/MLS/KeyPackage.hs
@@ -28,7 +28,7 @@ testDeleteKeyPackages = do
 testKeyPackageMultipleCiphersuites :: App ()
 testKeyPackageMultipleCiphersuites = do
   let suite = def
-      altSuite = Ciphersuite "0x0005"
+      altSuite = Ciphersuite "0x0007"
   alice <- randomUser OwnDomain def
   [alice1, alice2] <- replicateM 2 (createMLSClientWithCiphersuites [suite, altSuite] def alice)
 
@@ -209,7 +209,7 @@ testUnsupportedCiphersuite = do
 testReplaceKeyPackages :: (HasCallStack) => App ()
 testReplaceKeyPackages = do
   let suite = def
-      altSuite = Ciphersuite "0x0005"
+      altSuite = Ciphersuite "0x0007"
       oldSuite = Ciphersuite "0x0001"
   alice <- randomUser OwnDomain def
   [alice1, alice2] <- replicateM 2 $ createMLSClientWithCiphersuites [suite, altSuite, oldSuite] def alice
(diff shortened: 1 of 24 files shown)


integration/test/Test/MLS/KeyPackage.hs
```diff
@@ -11,9 +11,10 @@ import Testlib.Prelude
 
 testDeleteKeyPackages :: App ()
 testDeleteKeyPackages = do
+  let suite = Ciphersuite "0x0001"
   alice <- randomUser OwnDomain def
-  alice1 <- createMLSClient def def alice
-  kps <- replicateM 3 (uploadNewKeyPackage def alice1)
+  alice1 <- createMLSClient suite def alice
+  kps <- replicateM 3 (uploadNewKeyPackage suite alice1)
```
akshaymankar, line 17: Why this change? I thought we wanted most tests to use `def` suite?

akshaymankar (commented):
We should add release notes about this change because if the operator doesn't notice this, the clients could get into a very confusing.

libs/wire-api/src/Wire/API/Conversation/Protocol.hs
```diff
@@ -109,7 +109,7 @@ optionalActiveMLSConversationDataSchema (Just v)
             schema
         <*> fmap (.epochTimestamp)
           .= field "epoch_timestamp" (named "EpochTimestamp" . nullable . unnamed $ utcTimeSchema)
-        <*> maybe MLS_128_DHKEMX25519_AES128GCM_SHA256_Ed25519 (.ciphersuite)
+        <*> maybe MLS_128_DHKEMP256_AES128GCM_SHA256_P256 (.ciphersuite)
```
akshaymankar, line 112: This looks like a mistake, we shouldn't change legacy defaults in older versions.

libs/wire-api/src/Wire/API/MLS/CipherSuite.hs
```diff
@@ -122,7 +123,13 @@ data CipherSuiteTag
   deriving (Arbitrary) via (GenericUniform CipherSuiteTag)
 
 defCipherSuite :: CipherSuiteTag
-defCipherSuite = MLS_128_DHKEMX25519_AES128GCM_SHA256_Ed25519
+defCipherSuite = MLS_128_DHKEMP256_AES128GCM_SHA256_P256
```
akshaymankar, line 126: This is unused now. As discussed on the call, we should delete this and make a v8 version of the only endpoint which uses `defaultCipherSuiteV7` which makes the ciphersuite mandatory.

akshaymankar: Also feel free to do this in a separate PR/ticket.

stefanwire: Now done in this PR.

akshaymankar (approved):
Looks good :rocket:

PR #4376, opened 2024-12-16
fix bug: consent request not correctly forwarded to galeb

smatting:
This PR fixes the following bug: requests to nginz of form `/v./consent/.*` are not forwaded forwarded to `galeb` with path `/consent/.*` but with path `/consent`

I also considered changing the rewrite rule from

```
rewrite ^/v[0-9]+({{ $location.path }}) $1;
```

to

```
rewrite ^/v[0-9]+({{ $location.path }}.*) $1;
```

to fix the bug, but opted against it, because it would make the paths configured in `values.yaml` less explicit.
I think it's better to be explicit in the config about the fact that requests may be of form `/consent/<foo>/<bar>/<baz>`.
This is also consistent with other parts in the configuration where we are explicit about the whole path instead of relying on the fact for a `~*` you may omit the tail.

## Checklist

Tracked by https://wearezeta.atlassian.net/browse/WPB-14613

### DIFF
--- a/changelog.d/3-bug-fixes/WPB-14613
+++ b/changelog.d/3-bug-fixes/WPB-14613
@@ -1,0 +1,1 @@
+Fix bug in nginz: `/consent/<foo>` requests not correctly forwarded to `galeb`.
--- a/charts/nginz/values.yaml
+++ b/charts/nginz/values.yaml
@@ -708,7 +708,7 @@ nginx_conf:
       versioned: false
       strip_version: true
     galeb:
-    - path: /consent
+    - path: /consent(.*)
       envs:
       - all
       disable_zauth: true


jschaul (commented):
>This PR fixes the following bug: requests to nginz of form /v./consent/.* are not forwaded forwarded to galeb with path /consent/.* but with path /consent

Wouldn't that block need a `versioned: true` if we want to also forward requests with a v.. prefix?

How did you test? What requests are clients making (example ones)?

This change is fine as-is; however perhaps versioned rewrites need to also be added, no?

smatting:
> Wouldn't that block need a `versioned: true` if we want to also forward requests with a v.. prefix?

But we don't want to forward versioned requests to galeb (I assume): See that all routes for `galeb` have `versioned: false` and `strip_version: true`.

> How did you test? What requests are clients making (example ones)?
I checked logs via 
```
{namespace=~"wire", app="nginz", container="nginz"} | pattern `<remote_addr> <_> "<timestamp>" "<method> <path> <_> <status> <bytes_sent> "<_>" "<user_agent>" <_> <_> <request_time> <_> - <user> <conn> <request_id> <_> "<_>"` | path =~"/v./consent.*"
```
I didn't test this change to be honest, for example the `ibis` upstream rule
```
    - path: /teams/([^/]*)/billing(.*)
      envs:
      - all
      versioned: false
      strip_version: true
```
has the same properties as this rule and is working fine.

PR #4388, opened 2024-12-23
integration: Use lsof instead of ss to find processes listening on a port

akshaymankar:
Also don't ignore exceptions when stopping dynamic backends.

## Checklist


supersven (approved):
Looks useful :+1:

PR #4392, opened 2024-12-27
[WPB-15208] rm redundant function re-definition.

fisx:
https://wearezeta.atlassian.net/browse/WPB-15208

## Checklist

### DIFF
--- a/integration/test/Test/Events.hs
+++ b/integration/test/Test/Events.hs
@@ -23,7 +23,7 @@ import Servant.API (AsApi, ToServant, toServant)
 import Servant.API.Generic (fromServant)
 import qualified Servant.Client as Servant
 import SetupHelpers
-import Testlib.Prelude hiding (assertNoEvent)
+import Testlib.Prelude
 import UnliftIO hiding (handle)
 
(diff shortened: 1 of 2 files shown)


integration/test/Test/Events.hs
```diff
@@ -622,17 +624,30 @@ instance ToJSON NoEvent where
...
+
+assertWebSocketDied :: (HasCallStack) => EventWebSocket -> App ()
+assertWebSocketDied ws = do
+  recpol <- do
+    timeOutSeconds <- asks (.timeOutSeconds)
+    pure $ limitRetriesByCumulativeDelay (timeOutSeconds * 1_000_000) (constantDelay 800_000)
+  recoverAll recpol $ \_ ->
+    assertNoEventHelper ws >>= \case
+      NoEvent -> assertFailure $ "WebSocket is still open"
+      WebSocketDied -> pure ()
```
fisx, line 650: If we get a `NoEvent`, and only then a `WWebSocketDied`, the previous version didn't work.

I'm trying magic here.  I don't expect it to work, but I'd like to be proven wrong.

supersven (approved):
LGTM :+1:

PR #4397, opened 2025-01-02
integration: Close FDs of dynamic backend processes after they stop

akshaymankar:
https://wearezeta.atlassian.net/browse/WPB-15156

Most of the services get a SIGKILL because we don't wait long enough for them to close gracefully. This is mostly fine in tests because we don't want to wait ~30s for each service. So making sure that the FDs don't leak should fix the problem with federator not being able to come up due to previous federator's FDs not being closed.

## Checklist


pcapriotti (approved):
LGTM

akshaymankar:
Superseded by https://github.com/wireapp/wire-server/pull/4398

PR #4398, opened 2025-01-06
 federator: Close sockets on SIGINT/SIGTERM

akshaymankar:
## Checklist

### DIFF
--- a/services/federator/src/Federator/Interpreter.hs
+++ b/services/federator/src/Federator/Interpreter.hs
@@ -106,13 +106,20 @@ serveServant ::
   (HasServer api '[], Metrics.RoutesToPaths api) =>
   Env ->
   Int ->
+  IORef [IO ()] ->
   ServerT api (Sem AllEffects) ->
   IO ()
-serveServant env port server = do
+serveServant env port cleanupsRef server = do
   let hoistApp :: RequestId -> Server api
       hoistApp rid =
         hoistServerWithContext (Proxy @api) (Proxy @'[]) (runFederator env rid) server
-  Warp.run port
+      registerCleanupAction cleanupAction =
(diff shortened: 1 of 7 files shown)


battermann (approved):

services/federator/src/Federator/Interpreter.hs
```diff
@@ -106,13 +106,20 @@ serveServant ::
...
+serveServant env port cleanupsRef server = do
   let hoistApp :: RequestId -> Server api
       hoistApp rid =
         hoistServerWithContext (Proxy @api) (Proxy @'[]) (runFederator env rid) server
-  Warp.run port
+      registerCleanupAction cleanupAction =
+        atomicModifyIORef' cleanupsRef $ \xs -> (cleanupAction : xs, ())
+  let settings =
+        Warp.setPort port
+          . Warp.setInstallShutdownHandler registerCleanupAction
```
fisx, line 120: [hoogle says](https://hoogle.zinfra.io/file/nix/store/kvw1cn7822shw82mmkyn7mjs2xavf0y5-warp-3.4.2-doc/share/doc/warp-3.4.2/html/Network-Wai-Handler-Warp.html#v:setInstallShutdownHandler) that we may want to [add a timeout](https://hoogle.zinfra.io/file/nix/store/kvw1cn7822shw82mmkyn7mjs2xavf0y5-warp-3.4.2-doc/share/doc/warp-3.4.2/html/Network-Wai-Handler-Warp.html#v:setGracefulShutdownTimeout)?

fisx (approved):
brig uses `runSettingsWithShutdown`, which looks very similar.  just wondering if you knew about that.  your way certainly looks like it's going to work as well.

fisx:
background-worker uses `runSettingsWithCleanup`, which is similar.  maybe we should check if we missed a variant to include in our code, hehe...

PR #4404, opened 2025-01-08
cannon: Create queue before responding to the websocket

akshaymankar:
This way tests can run more reliably, i.e. when they see that the websocket is functional, they queue already exists in RabbitMQ.

We can also be sure that when the request is upgraded to the websocket, events can start flowing. If the channel limit is reached, cannon throws HTTP 503 rejecting the websocket connection.

Also in this PR:
- Remove channel re-establishment code in Cannon as it is not used anyway and it was easier to ensure that the queue is created before we respond to the websocket.
- Add `Deflake` type to run a test multiple times
- Ensure that integration tests run the websocket assertions only after the websocket is functional.
- Small changes to websockets handling in integration tests so errors look nicer
- On RabbitMq channel close, send code `1001` (meaning `CLOSE_GOING_AWAY`) instead of `internal-error` as this is a normal thing.

## Checklist

### DIFF
--- a/charts/integration/templates/integration-integration.yaml
+++ b/charts/integration/templates/integration-integration.yaml
@@ -360,3 +360,5 @@ spec:
       value: "1"
     - name: ENABLE_FEDERATION_V1
       value: "1"
+    - name: TEST_INCLUDE
+      value: testConsumeTempEventsWithoutOwnClient
--- a/charts/integration/templates/integration-integration.yaml
+++ b/charts/integration/templates/integration-integration.yaml
@@ -360,5 +360,3 @@ spec:
       value: "1"
     - name: ENABLE_FEDERATION_V1
       value: "1"
-    - name: TEST_INCLUDE
-      value: testConsumeTempEventsWithoutOwnClient
--- a/hack/bin/integration-test.sh
+++ b/hack/bin/integration-test.sh
@@ -11,7 +11,7 @@ UPLOAD_LOGS=${UPLOAD_LOGS:-0}
 echo "Running integration tests on wire-server with parallelism=${HELM_PARALLELISM} ..."
 
 CHART=wire-server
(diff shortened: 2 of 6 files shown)


akshaymankar:
Ran that test 100 times twice: 
https://concourse.ops.zinfra.io/teams/main/pipelines/wire-server-pr/jobs/integration-tests-kube-ci/builds/4066
https://concourse.ops.zinfra.io/teams/main/pipelines/wire-server-pr/jobs/integration-tests-kube-ci/builds/4066.1

It failed 7 out 100 times before the fix: https://concourse.ops.zinfra.io/teams/main/pipelines/wire-server-pr/jobs/integration-tests-kube-ci/builds/4059.1

battermann (approved):

akshaymankar:
> The test suite now takes more than 15 mins, sounds like a reason to not merge this yet.

It was because a test was stuck due to us never expecting WS handshake to fail. Should be resolved now.

pcapriotti (commented):
I had originally left the channel retry mechanism in because my original intention was to integrate this pool into background worker (see https://wearezeta.atlassian.net/browse/WPB-14515). I haven't really understood why it was necessary to remove the retry code in this PR.

akshaymankar:
> I had originally left the channel retry mechanism in because my original intention was to integrate this pool into background worker (see https://wearezeta.atlassian.net/browse/WPB-14515). I haven't really understood why it was necessary to remove the retry code in this PR.

That's another thing I'm not sure why we want to do. I think the background worker can just work with one connection and a channel per connection. Also if we go down this way of reconnecting channels, we'll again get into trouble with sending acks to wrong channels.

The reason I removed it was it just made things easier to write, so instead of caling `createChannel` and then waiting on the MVar , we could just get the channel and run rest of the code with it.

pcapriotti (approved):

PR #4407, opened 2025-01-09
WPB-15030 Delete deprecated and unqualified endpoints from V8

battermann:
## Checklist

### DIFF
--- a/libs/wire-api/src/Wire/API/Routes/Public/Spar.hs
+++ b/libs/wire-api/src/Wire/API/Routes/Public/Spar.hs
@@ -1,5 +1,3 @@
-{-# OPTIONS_GHC -Wno-orphans #-}
-
 -- This file is part of the Wire Server implementation.
 --
 -- Copyright (C) 2022 Wire Swiss GmbH <opensource@wire.com>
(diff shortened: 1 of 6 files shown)


libs/wire-api/src/Wire/API/Routes/Public/Spar.hs
```diff
@@ -164,6 +167,10 @@ sparResponseURI Nothing =
 sparResponseURI (Just tid) =
   SAML.getSsoURI' (Proxy @APISSO) (Proxy @APIAuthResp) tid
 
+instance HasLink (Until 'V8 :> api) where
```
battermann, line 170: Not so sure about this. We need this instance because of `getSsoURI` (https://hoogle.zinfra.io/file/nix/store/kjcia6s20r3lw0nvl1h68mba1d8m0nwy-saml2-web-sso-0.20-doc/share/doc/saml2-web-sso-0.20/html/SAML2-WebSSO-SP.html#v:getSsoURI). But I don't know yet if this works. In any case, it should probably go into the Version module.

battermann: @fisx can we check this together?

PR #4408, opened 2025-01-09
charts{redis-cluster,rabbitmq}: Use OCI registry for upstream charts

akshaymankar:
As recommended here: https://github.com/bitnami/charts/issues/30110

The http repository started failing recently, upstream maintainers suggest that the OCI repository is the official repository and should be used.

## Checklist


battermann (approved):

akshaymankar:
This doesn't work because our charts are too old and they're not pushed to the OCI repos.

PR #4437, opened 2025-01-31
[WPB-14310] make existing registration api enterprise aware (cont.)

fisx:
## Checklist

### DIFF
--- a/libs/wire-api/src/Wire/API/EnterpriseLogin.hs
+++ b/libs/wire-api/src/Wire/API/EnterpriseLogin.hs
@@ -242,6 +242,21 @@ data DomainRegistration = DomainRegistration
   }
   deriving stock (Eq, Show)
 
+instance Arbitrary DomainRegistration where
+  arbitrary = do
+    dom :: Domain <- arbitrary
+    upd :: DomainRegistrationUpdate <- arbitrary
+    mbteam :: Maybe TeamId <- arbitrary
+    pure
+      DomainRegistration
+        { domain = dom,
+          authorizedTeam = mbteam,
+          domainRedirect = upd.domainRedirect,
+          teamInvite = upd.teamInvite,
+          dnsVerificationToken = Nothing,
+          authTokenHash = Nothing
+        }
+
 mkDomainRegistration :: Domain -> DomainRegistration
 mkDomainRegistration domain =
   DomainRegistration
--- a/libs/wire-subsystems/src/Wire/EnterpriseLoginSubsystem/Interpreter.hs
+++ b/libs/wire-subsystems/src/Wire/EnterpriseLoginSubsystem/Interpreter.hs
@@ -28,8 +28,9 @@ import Imports hiding (lookup)
 import Network.HTTP.Types.Method
 import Network.Mail.Mime (Address (Address), Mail (mailHeaders, mailParts, mailTo), emptyMail, plainPart)
 import Polysemy
-import Polysemy.Error
-import Polysemy.Input
+import Polysemy.Error (Error, note, throw)
+import Polysemy.Error qualified as Error
+import Polysemy.Input (Input, input, inputs, runInputConst, runInputSem)
 import Polysemy.TinyLog (TinyLog)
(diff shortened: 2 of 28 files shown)


libs/wire-api/src/Wire/API/EnterpriseLogin.hs
```diff
@@ -232,6 +232,61 @@ instance ToSchema DomainRegistrationResponse where
         <*> (.teamInvite) .= teamInviteObjectSchema
         <*> (.dnsVerificationToken) .= optField "dns_verification_token" (maybeWithDefault Aeson.Null schema)
 
+----------------------------------------------------------------------
+
+data DomainRegistration' = DomainRegistration'
+  { domain :: Domain,
+    settings :: DomainRegistrationSettings',
```
akshaymankar, line 239: Can we make this a `Maybe` and remove the `None` case? I think preauthorized can also be outsize of this, because its not really a setting, but its just a step in configuration. I think the main logic of wire-server (i.e. regsitrations and logins) are not affected by it being pre-authorized, it is treated the same as not having any config.

fisx: right, domain was a hickup.  and yes, maybe is nicer than none.  :+1:  bd36d474cb4d6e54880b17b888b11b26b7fdbd98

fisx: on second thought, the code is using the domain field in DomainRegistration a lot, and if we want to replace that with DomainRegistration', the code change in the logic would explode a little, and the new code wouldn't exactly be more concise.

i'll tentatively add the domain back in, let's see if i can convince you.

libs/wire-api/src/Wire/API/EnterpriseLogin.hs
```diff
@@ -232,44 +239,133 @@ instance ToSchema DomainRegistrationResponse where
...
+domainRegistrationFromRow DomainRegistrationRow {..} = do
+  settings <- domainRegistrationSettingsFromRow domainRedirect teamInvite authorizedTeam
+  Right DomainRegistration {..}
+
+domainRegistrationFromUpdate :: DomainRegistration -> DomainRegistrationUpdate -> Either String DomainRegistration
+domainRegistrationFromUpdate reg upd = do
+  newSettings <- domainRegistrationSettingsFromRow upd.domainRedirect upd.teamInvite reg.authorizedTeam
+  Right reg {settings = newSettings}
+
+-- | This type only server as a helper for json (and cql, for historical reasons).  App logic
```
fisx, line 343:
```suggestion
-- | This type only serves as a helper for json (and cql, for historical reasons).  App logic
```

libs/wire-subsystems/src/Wire/EnterpriseLoginSubsystem/Interpreter.hs
```diff
@@ -284,14 +285,15 @@ unauthorizeImpl ::
...
-    Locked -> throw EnterpriseLoginSubsystemOperationForbidden
-    SSO _ -> throw EnterpriseLoginSubsystemOperationForbidden
+  let new = old {settings = Nothing} :: DomainRegistration
+  case old.settings of
+    Nothing -> pure ()
+    Just DomainLocked -> throw EnterpriseLoginSubsystemOperationForbidden
+    Just DomainPreAuthorized -> audit old new *> upsert new
+    Just (DomainNoRegistration _) -> audit old new *> upsert new
+    Just (DomainForBackend _) -> audit old new *> upsert new
+    Just (DomainTeamInviteRestricted _) -> audit old new *> upsert new
```
fisx, line 295:
```suggestion
    Just (DomainTeamInviteRestricted _) -> pure ()
```

`DomainTeamInviteRestricted _` implies `domainRedirect == None`, and the old code says `None -> pure ()`.

fisx (commented):
LGTM!  (can't formally approve as i've opened this PR)

fisx:
stale, closed.

PR #4442, opened 2025-02-04
Pydio events

pcapriotti:
https://wearezeta.atlassian.net/browse/WPB-15965

## Checklist


hack/helm_vars/wire-server/values.yaml.gotmpl
```diff
@@ -342,6 +342,10 @@ galley:
           defaults:
             status: disabled
             lockStatus: locked
+        cells:
```
stefanwire, line 345: According to the [documentation](https://wearezeta.atlassian.net/wiki/spaces/PAD/pages/1800962144/2025-03-19+Wire+Cells+implementation+in+wire-server), this should be locked and disabled by default.

stefanwire (commented):
Preliminary comments after reading through the first files. Need to continue later.

services/galley/src/Galley/Cassandra/Queries.hs
```diff
@@ -212,28 +213,28 @@ updateTeamSplashScreen = {- `IF EXISTS`, but that requires benchmarking -} "upda
...
-      Maybe ReceiptMode,
-      Maybe ProtocolTag,
-      Maybe GroupId,
-      Maybe Epoch,
-      Maybe (Writetime Epoch),
-      Maybe CipherSuiteTag,
-      Maybe GroupConvType
-    )
-selectConv = "select type, creator, access, access_role, access_roles_v2, name, team, deleted, message_timer, receipt_mode, protocol, group_id, epoch, WRITETIME(epoch), cipher_suite, group_conv_type from conversation where conv = ?"
+type ConvRow =
```
battermann, line 216: would be nice to use https://hackage.haskell.org/package/cql-4.0.4/docs/Database-CQL-Protocol.html#v:recordInstance
but maybe too much for this PR

battermann (approved):
only some minor comments

services/galley/src/Galley/API/Cells.hs
```diff
@@ -0,0 +1,52 @@
...
+  getCellsState = getCellsState . Data.convMetadata
+
+instance HasCellsState Public.Conversation where
+  getCellsState = getCellsState . Public.cnvMetadata
+
+instance HasCellsState ConversationMetadata where
+  getCellsState = cnvmCellsState
+
+instance HasCellsState () where
+  getCellsState = def
```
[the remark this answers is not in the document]
pcapriotti: The default is Disabled, and it's not really meant to be changed. However, this `()` is used for situations in which we don't have any conversation metadata available, so it makes sense to use the default cells state. Right now this is only used for federated conversations, so it might not be needed in the future depending on how we decide to handle them.

services/galley/src/Galley/Cassandra/Queries.hs
pcapriotti: Yeah, let's do it in a separate PR.

PR #4449, opened 2025-02-11
Cleanup ingress classes after integration tests.

jschaul:
These objects used to linger forever, eventually exhausting etcd after many months of development.

## Checklist

### DIFF
--- a/hack/bin/integration-teardown-federation.sh
+++ b/hack/bin/integration-teardown-federation.sh
@@ -28,4 +28,7 @@ fi
 . "$DIR/helm_overrides.sh"
 helmfile --environment "$HELMFILE_ENV" --file "${TOP_LEVEL}/hack/helmfile.yaml" destroy --skip-deps --skip-charts --concurrency 0 || echo "Failed to delete helm deployments, ignoring this failure as next steps will the destroy namespaces anyway."
 
+kubectl delete ingressclass "nginx-$NAMESPACE_1" || true
+kubectl delete ingressclass "nginx-$NAMESPACE_2" || true
+
 kubectl delete namespace "$NAMESPACE_1" "$NAMESPACE_2"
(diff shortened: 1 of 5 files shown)


lwille (approved):

hack/bin/integration-teardown-federation.sh
```diff
@@ -28,4 +28,7 @@ fi
 . "$DIR/helm_overrides.sh"
 helmfile --environment "$HELMFILE_ENV" --file "${TOP_LEVEL}/hack/helmfile.yaml" destroy --skip-deps --skip-charts --concurrency 0 || echo "Failed to delete helm deployments, ignoring this failure as next steps will the destroy namespaces anyway."
 
+kubectl delete ingressclass "nginx-$NAMESPACE_1" || true
+kubectl delete ingressclass "nginx-$NAMESPACE_2" || true
+
```
akshaymankar, line 33: I think this shouldn't be needed if the line above worked. 
The line above doesn't work because some env vars are not set.
From CI output:

```
+ helmfile --environment default --file /tmp/build/e290497f/wire-server/hack/bin/../../hack/helmfile.yaml destroy --skip-deps --skip-charts --concurrency 0
in /tmp/build/e290497f/wire-server/hack/helmfile.yaml: failed to read helmfile.yaml: failed to load environment values file "helm_vars/common.yaml.gotmpl": failed to render [helm_vars/common.yaml.gotmpl], because of template: stringTemplate:7:28: executing "stringTemplate" at <requiredEnv "FEDERATION_CA_CERTIFICATE">: error calling requiredEnv: required env var `FEDERATION_CA_CERTIFICATE` is not set
+ echo 'Failed to delete helm deployments, ignoring this failure as next steps will the destroy namespaces anyway.'
```

PR #4459, opened 2025-02-18
Team Feature Config Migration

battermann:
This PR introduces a new storage system for team features and a corresponding migration.

https://wearezeta.atlassian.net/browse/WPB-16238

## Checklist

 - [x] Add a new entry in an appropriate subdirectory of `changelog.d`
 - [x] Read and follow the [PR guidelines](https://docs.wire.com/developer/developer/pr-guidelines.html)

PR #4470, opened 2025-02-25
Cleanup old team feature storage

pcapriotti:
This PR removes support for legacy team feature storage, and forces the use of the new dynamic feature table introduced in #4459.

Since #4459 and #4472 contain the data migration needed to fully migrate to the new storage, it is important that this PR not be merged before both of those PRs are included in a release.

https://wearezeta.atlassian.net/browse/WPB-16238

## Checklist


pcapriotti:
> So we are keeping the DB schema for now? I guess that is fine.

You mean the old table itself? Yeah, we have to keep it to prevent issues during deployment. We can get rid of it in one of the next releases.

PR #4472, opened 2025-02-26
Team feature migration tool

pcapriotti:
Team feature migration tool.

We decided that a data migration is not the best way to move feature configurations to the new table. Instead, this PR adds a new `migrate-features` tool that can be used to manually perform the migration.

https://wearezeta.atlassian.net/browse/WPB-16238

## Checklist


supersven:
We're using this Job manifest to run it on `staging`:

```yaml
apiVersion: batch/v1
kind: Job
metadata:
  name: migrate-features
spec:
  template:
    spec:
      containers:
        - name: migrate-features
          image: quay.io/wire/migrate-features:0.0.2-pr.10263
          args:
            [
              --cassandra-host-galley,
              galley-galley-eks-service.databases.svc.cluster.local,
              --cassandra-port-galley,
              "9042",
              --cassandra-keyspace-galley,
              galley,
            ]
      restartPolicy: Never
  backoffLimit: 4
```

If it succeeds, we may add this :point_up: to the docs.

services/galley/migrate-data/src/Run.hs
```diff
@@ -33,8 +32,7 @@ main = do
     l
     o
     [ V1_BackfillBillingTeamMembers.migration,
-      V3_BackfillTeamAdmins.migration,
-      V4_MigrateToDynamicFeatures.migration
+      V3_BackfillTeamAdmins.migration
```
akshaymankar, line 35:
```suggestion
      V3_BackfillTeamAdmins.migration
      -- V4 was deleted but it must've run in some environments. The next migration should be V5.
```

akshaymankar (commented):
Also needs docs, I'll add them

akshaymankar (approved):

PR #4464, opened 2025-02-19
Clean wire-server/docs and migrate to wire-docs repository WPB-16191

mohitrajain:
**Before merging it, we need to approve merge this PR:**
https://github.com/wireapp/wire-server/pull/4474
## Checklist

 - [x] Add a new entry in an appropriate subdirectory of `changelog.d`
 - [x] Read and follow the [PR guidelines](https://docs.wire.com/developer/developer/pr-guidelines.html)

PR #4487, opened 2025-03-07
Document Channels feature flag

supersven:
I'm not sure how this fits into the migration of docs (@mohitrajain ?). However, the doc files seem to only be moved, so this would be easy to copy somewhere else.

Ticket: https://wearezeta.atlassian.net/browse/WPB-16301

## Checklist

 - [ ] ~~Add a new entry in an appropriate subdirectory of `changelog.d`~~

### DIFF
--- a/docs/src/developer/reference/config-options.md
+++ b/docs/src/developer/reference/config-options.md
@@ -372,6 +372,48 @@ To get the public key run:
 openssl ec -in private.pem -pubout --out public.pem
 ```
 
+### Channels
+
+Channels are configured per team (as "team feature".) The configuration defines
+if the channels feature is activated and who can create or open channels.
+
+The configuration has these fields:
+
+- `lockStatus`: `locked` or `unlocked`
+- `status`: `enabled` or `disabled`
+- `config.allowed_to_create_channels`: _roles_
+- `config.allowed_to_open_channels`: _roles_
+
+Possible _roles_ are:
+
+| Value          | Who is that?                                    | UX description                    |
+| -------------- | ----------------------------------------------- | --------------------------------- |
+| `everyone`     | Partner (a.k.a. external), Member, Admin, Owner | Everyone in the team              |
+| `team-members` | Member, Admin, Owner                            | Team members (excluding external) |
+| `admins`       | Admin, Owner                                    | Only admins                       |
+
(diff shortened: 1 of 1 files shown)


mohitrajain:
@supersven I am keeping an eye on all the documentation changes and repeating them with my commits with original commit id in the message for example:
https://github.com/wireapp/wire-server/pull/4464/commits/959c931fd8cf055c64980377ccf36712cc874955 

Please feel free to merge the change, i will collect it and stage it in the PR:
https://github.com/wireapp/wire-server/pull/4464

stefanwire (approved):

docs/src/developer/reference/config-options.md
```diff
@@ -372,6 +372,48 @@ To get the public key run:
...
+- `config.allowed_to_create_channels`: _roles_
+- `config.allowed_to_open_channels`: _roles_
+
+Possible _roles_ are:
+
+| Value          | Who is that?                                    | UX description                    |
+| -------------- | ----------------------------------------------- | --------------------------------- |
+| `everyone`     | Partner (a.k.a. external), Member, Admin, Owner | Everyone in the team              |
+| `team-members` | Member, Admin, Owner                            | Team members (excluding external) |
+| `admins`       | Admin, Owner                                    | Only admins                       |
```
fisx, line 393: wayyy too late in the process, but why didn't we just use the roles we already have here?  they way they are implemented at the moment, they all contain each other, so "partner" would behave identically to "everyone", "member" to "team-members", and "admin" to "admins".  (and we could also set it to "owner" if we really wanted to.)

fisx (approved):

docs/src/developer/reference/config-options.md
supersven: IIRC these values were stated in the RFC. Now, they are described in the Tech Spec: https://wearezeta.atlassian.net/wiki/spaces/ENGINEERIN/pages/1758134359/Channels+feature+flag+and+config

Internally, these `ChannelPermissions` _should_ be mapped to `Roles` before we use them for any logic.

(BTW - This was the advice from @pcapriotti I told you about: Keep the `ToSchema` instances close to the expected JSON format. That's why there's no internal mapping to e.g. `Role`s in the `ToSchema` instance.)

PR #4503, opened 2025-03-24
Cells events - part 2

pcapriotti:
Followup to #4442:

- Send events when Cells state is pending.
- Add `cells` flag to `NewConv`

https://wearezeta.atlassian.net/browse/WPB-15965

## Checklist


integration/test/Test/Cells.hs
```diff
@@ -37,20 +37,43 @@ testCellsEvent = do
...
+  -- start watcher before creating conversation
+  q0 <- watchCellsEvents def
+  (alice, tid, _) <- createTeam OwnDomain 1
+  conv <- postConversation alice defProteus {team = Just tid, cells = True} >>= getJSON 201
+
+  let q = q0 {filter = isNotifConv conv} :: QueueConsumer
+
   event <- getMessage q %. "payload.0"
-  event %. "type" `shouldMatch` "conversation.member-join"
+  event %. "type" `shouldMatch` "conversation.create"
```
fisx, line 74: did you remove the `member-join` noise because it is tested elsewhere, or just because it doesn't belong into this test?  (or because it doesn't happen any more?  if so, why doesn't it?  and shouldn't the test make sure the queue is empty in that case?)

services/galley/src/Galley/API/Cells.hs
```diff
@@ -49,4 +49,4 @@ instance HasCellsState () where
 
 shouldPushToCells :: (HasCellsState a) => a -> EventType -> Bool
 shouldPushToCells st et =
-  isCellsConversationEvent et && getCellsState st == CellsReady
+  isCellsConversationEvent et && getCellsState st /= CellsDisabled
```
fisx, line 52:
```suggestion
  isCellsConversationEvent et && case getCellsState st of
    CellsDisabled -> False
    CellsPending -> True
    CellsReady -> True
```

(to make changes in the state type cause type errors in this place if it's overlooked)

services/gundeck/src/Gundeck/Push.hs
```diff
@@ -276,13 +276,18 @@ pushViaRabbitMq p = do
...
+
+  let cellPushes = filter (._pushIsCellsEvent) ps
+  unless (null cellPushes) $ case mQueue of
+    Nothing -> do
+      Log.err $ msg ("Cells events have been generated, but no Cells queue is configured in gundeck" :: ByteString)
+    Just queue -> traverse_ (pushToCells queue) cellPushes
 
 pushToCells :: (MonadPushAll m) => Text -> Push -> m ()
-pushToCells queue p = when p._pushIsCellsEvent $ do
+pushToCells queue p = do
```
fisx, line 290: what's behind this change?

fisx (approved):

integration/test/Test/Cells.hs
pcapriotti: This is just a funny-looking diff. The portion related to the new test got mixed with the end of the previous test.

services/gundeck/src/Gundeck/Push.hs
pcapriotti: I added a log message that should help diagnose configuration mistakes (e.g. forgetting to set the cells queue name). With this, we generate a single log message rather than one per push.

PR #4504, opened 2025-03-25
WPB-16388 allow users to add new members to private channels

battermann:
## Checklist


integration/test/Test/Channels.hs
```diff
@@ -198,10 +204,145 @@ testTeamAdminPermissions = do
...
+      mp <- createAddCommit userClient convId [userToAdd]
+      postMLSCommitBundle userClient (mkBundle mp) `bindResponse` \resp -> do
+        resp.status `shouldMatchInt` 500
+        resp.json %. "label" `shouldMatch` "federation-not-implemented"
+
+-- if the federation queue gets stuck, the second test run will fail
+-- therefore this test verifies that a notification that cannot be parsed by the remote
+-- backend does not block the queue
+testWithOldBackendVersion :: (HasCallStack) => App ()
+testWithOldBackendVersion = replicateM_ 2 do
```
pcapriotti, line 329: I think these should be gated by environment variables. You can use `AnyFedDomain` as a parameter to achieve that automatically.

pcapriotti (approved):
Looks good. Just a minor comment on the integration test with legacy backends.

fisx (approved):

emil-wire:
Reason for bypassing Sonar: it scanned the docs, we forgot to exclude them from the scan

PR #4519, opened 2025-04-04
Disallow duplicate clients in MLS groups

pcapriotti:
This is a first step to safeguard against broken clients that sometimes operate with new credentials without registering a new client to the backend. We simply check that duplicate clients are never introduced in group, either with internal or external commits.

Checking that the signature key of new leaf nodes always matches the one registered with the backend will be in a followup PR.

https://wearezeta.atlassian.net/browse/WPB-16786

## Checklist

### DIFF
--- a/integration/test/Test/MLS.hs
+++ b/integration/test/Test/MLS.hs
@@ -849,7 +849,7 @@ testBackendRemoveProposal suite domain = do
 testExternalCommitDuplicateClient :: (HasCallStack) => App ()
 testExternalCommitDuplicateClient = do
   alice <- randomUser OwnDomain def
-  creator : [other] <- traverse (createMLSClient def def) (replicate 2 alice)
+  [creator, other] <- traverse (createMLSClient def def) (replicate 2 alice)
   (_, conv) <- createSelfGroup def creator
   convId <- objConvId conv
   void $ createAddCommit creator convId [alice] >>= sendAndConsumeCommitBundle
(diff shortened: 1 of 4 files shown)


integration/test/Test/MLS.hs
```diff
@@ -845,3 +845,45 @@ testBackendRemoveProposal suite domain = do
...
+    setClientGroupState alice2 def
+    (kp, _) <- generateKeyPackage alice2 def
+
+    -- We cannot upload the new key package at this point, because the
+    -- signature key won't match. However, alice1 can still use it to craft an
+    -- add proposal.
+    mp <- createAddCommitWithKeyPackages alice1 convId [(alice2, kp)]
+    bindResponse (postMLSCommitBundle alice1 (mkBundle mp)) $ \resp -> do
+      resp.status `shouldMatchInt` 400
+      resp.json %. "label" `shouldMatch` "mls-protocol-error"
```
battermann, line 889:
```suggestion
  -- wipe key store
  setClientGroupState alice2 def
  (kp, _) <- generateKeyPackage alice2 def

  -- We cannot upload the new key package at this point, because the
  -- signature key won't match. However, alice1 can still use it to craft an
  -- add proposal.
  mp <- createAddCommitWithKeyPackages alice1 convId [(alice2, kp)]
  bindResponse (postMLSCommitBundle alice1 (mkBundle mp)) $ \resp -> do
    resp.status `shouldMatchInt` 400
    resp.json %. "label" `shouldMatch` "mls-protocol-error"
```

battermann (approved):

stefanwire (approved):

PR #4530, opened 2025-04-08
New section How To in Administration docs

Veki301:
## Checklist


mohitrajain:
@Veki301 We should add this file to wire-docs repo at this path:
https://github.com/wireapp/wire-docs/tree/main/src/how-to/administrate 
and also make an entry in README.md so that it can get indexed as well. 

wire-server/docs are reserved for backend documents that move with code directly.

PR #4536, opened 2025-04-10
[WPB-1227] rm dependencies wai-route, wai-predicates everywhere

fisx:
https://wearezeta.atlassian.net/browse/WPB-1227

## Checklist


fisx:
`pattern-trie` is no longer needed, but the dependency is still lurking somewhere.  maybe `generate-nix-packages.sh` isn't very good at removing dependencies from packages that have just gone, instead of their cabal file updated?  :thinking:

PR #4548, opened 2025-04-17
WIP cargohold stackit

jschaul:
Just a try

## Checklist

### DIFF
--- a/services/cargohold/cargohold.cabal
+++ b/services/cargohold/cargohold.cabal
@@ -87,6 +87,7 @@ library
       aeson                  >=2.0.1.0
     , amazonka               >=1.3.7
     , amazonka-s3            >=1.3.7
+    , amazonka-s3-streaming  >=1.3.7
     , attoparsec             >=0.12
     , auto-update            >=0.1.4
     , base                   >=4       && <5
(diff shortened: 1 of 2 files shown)


smatting:
I wasn't able to figure out the maximum number of lightweight threads the `warp` server will spawn. This could help to inform a good setting for the memory cargohold might need.

akshaymankar:
> I wasn't able to figure out the maximum number of lightweight threads the `warp` server will spawn. This could help to inform a good setting for the memory cargohold might need.

I think there is no maximum. Perhaps we should just use `net.connections` metrics from production to decide this?

akshaymankar (approved):
Looks good! 
Feel free to tweak the default memory requests and limits separately.

jschaul:
This PR changes the way assets are uploaded to S3, it makes use of [multi-part uploads API](https://docs.aws.amazon.com/AmazonS3/latest/userguide/mpuoverview.html) via the amazonka-s3-streaming library.

cargohold needs to be compatible with StorageGRID.
Without this PR cargohold receives 400 Bad request when uploading. It might be related to this comment:
: https://docs.netapp.com/us-en/storagegrid/s3/s3-rest-api-supported-operations-and-limitations.html#common-request-headers

Switching to multi-part uploads seems to fix the problem.

See also https://wearezeta.atlassian.net/browse/WPB-17236

jschaul:
It seems our concurrent uploads per second are not very high; so current limits should be okay. We should monitor and increase if we discover OOMkills.

PR #4556, opened 2025-04-25
WPB-17378 gundeck: Fix logic to split pushes between cassandra and rabbitmq

battermann:
https://wearezeta.atlassian.net/browse/WPB-17378

## Checklist

### DIFF
--- a/changelog.d/3-bug-fixes/WPB-17378
+++ b/changelog.d/3-bug-fixes/WPB-17378
@@ -1,0 +1,1 @@
+Prevent notifications from being send back to the sending client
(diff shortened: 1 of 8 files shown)


changelog.d/3-bug-fixes/WPB-17378
```diff
@@ -0,0 +1 @@
+Prevent notifications from being send back to the sending client
```
akshaymankar, line 1:
```suggestion
gundeck: Send notifications to temp clients only when there are no rabbitmq clients in the recipient list.  Before this, it was wrongly sending the notification to all recipients.
```

akshaymankar (approved):

PR #4558, opened 2025-04-30
Reset MLS conversations

pcapriotti:
This adds a new endpoint `/mls/reset-conversation` which can be used to restore an MLS group that ended up in an invalid state for any reason. After resetting, the conversation has the same users, but the corresponding MLS group gets a new group ID and resets to epoch 0 with no clients.

https://wearezeta.atlassian.net/browse/WPB-17337

## Checklist

### DIFF
--- a/integration/test/Test/MLS/Reset.hs
+++ b/integration/test/Test/MLS/Reset.hs
@@ -4,7 +4,6 @@ import API.Galley
 import MLS.Util
 import SetupHelpers
 import Testlib.Prelude
-import Testlib.VersionedFed
 
 testResetGroupConversation :: (HasCallStack) => Domain -> App ()
 testResetGroupConversation domain = do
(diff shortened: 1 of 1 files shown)


services/galley/src/Galley/API/MLS/Reset.hs
```diff
@@ -0,0 +1,166 @@
...
+    Member SubConversationStore r
+  ) =>
+  Qualified UserId ->
+  ConvType ->
+  Local ConvOrSubConvId ->
+  MLSReset ->
+  Sem r ()
+resetLocalMLSConversation qusr ctype lcnvOrSub reset =
+  case tUnqualified lcnvOrSub of
+    SubConv c s -> deleteLocalSubConversation qusr (c <$ lcnvOrSub) s reset -- TODO: extract
```
stefanwire, line 101: `TODO` or `FUTUREWORK`?

libs/wire-api/src/Wire/API/Routes/Public/Galley/Conversation.hs
```diff
@@ -608,7 +609,7 @@ type ConversationAPI =
                :> QualifiedCapture "cnv" ConvId
                :> "subconversations"
                :> Capture "subconv" SubConvId
-               :> ReqBody '[JSON] DeleteSubConversationRequest
```
stefanwire, line 611: Wouldn't this break older API versions?

pcapriotti: No, because the API is unchanged. Just the name of the type has changed.

services/galley/src/Galley/API/MLS/Reset.hs
pcapriotti: Ah, thanks for catching this. I wanted to refactor this slightly to avoid invoking the API handler in this case, but I'll move it to a follow-up PR.

stefanwire (approved):

PR #4564, opened 2025-05-05
Add a webapp url field to email domain registration (3)

supersven:
Actually, this contains only a small left-over refactoring.

Ticket: https://wearezeta.atlassian.net/browse/WPB-17032

## Checklist

 - [x] ~~Add a new entry in an appropriate subdirectory of `changelog.d`~~ (only internal changes)

### DIFF
--- a/libs/wire-api/src/Wire/API/EnterpriseLogin.hs
+++ b/libs/wire-api/src/Wire/API/EnterpriseLogin.hs
@@ -100,7 +100,7 @@ domainRedirectSchemaV8 =
 backendConfigSchemaV8 :: ObjectSchema SwaggerDoc (HttpsUrl, Maybe HttpsUrl)
 backendConfigSchemaV8 =
   (,)
-    <$> fst .= backendUrlSchema
+    <$> fst .= field "backend_url" schema
     -- API versions <= V9 ignore the WebApp URL
     <*> const Nothing .= pure Nothing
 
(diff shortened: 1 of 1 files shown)


libs/wire-api/src/Wire/API/EnterpriseLogin.hs
```diff
@@ -100,7 +100,7 @@ domainRedirectSchemaV8 =
 backendConfigSchemaV8 :: ObjectSchema SwaggerDoc (HttpsUrl, Maybe HttpsUrl)
 backendConfigSchemaV8 =
   (,)
-    <$> fst .= backendUrlSchema
+    <$> fst .= field "backend_url" schema
     -- API versions <= V9 ignore the WebApp URL
```
fisx, line 104: is this comment still correct?  (was it ever?  shouldn't it be `<9`?)

fisx: what is the point of this change?

fisx (approved):

libs/wire-api/src/Wire/API/EnterpriseLogin.hs
supersven: > is this comment still correct? (was it ever? shouldn't it be `<9`?)

Thanks for spotting this! :+1:  You're right it should be `<=V8` or `<V9` (which are equal constraints.)

The change itself is just inlining a public function which is only used here. There's no second call site, thus we don't need add this indirection.

PR #4570, opened 2025-05-09
More SAML login error details.

fisx:
https://wearezeta.atlassian.net/browse/WPB-17585

## Checklist

### DIFF
--- a/libs/saml2-web-sso/src/SAML2/WebSSO/API.hs
+++ b/libs/saml2-web-sso/src/SAML2/WebSSO/API.hs
@@ -124,12 +124,13 @@ parseAuthnResponseBody mbSPId base64 = do
   -- '=', and probably other noise, this seems the safe thing to do.  It is no less secure than
   -- rejecting some noise characters and ignoring others.
   let xmltxt :: LBS = EL.decodeLenient (cs base64 :: LBS)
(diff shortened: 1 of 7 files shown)


libs/saml2-web-sso/src/SAML2/WebSSO/API.hs
```diff
@@ -124,12 +124,13 @@ parseAuthnResponseBody mbSPId base64 = do
...
   -- rejecting some noise characters and ignoring others.
   let xmltxt :: LBS = EL.decodeLenient (cs base64 :: LBS)
+      extraErrInfo :: String = ST.unpack base64
 
   (signedAssertions, idp, status) <- do
     resp <-
       -- do not use `resp` as a result of `parseAuthnResponseBody`!  only use what comes back
       -- from `simpleVerifyAuthnResponse`!
-      either (throwError . BadSamlResponseXmlError . cs) pure $
+      either (throwError . BadSamlResponseXmlError . cs . (<> extraErrInfo)) pure $
```
akshaymankar, line 133: It could be hard to tell where the error stops and where the response starts. We should perhaps merge the strings with identifier text like `parseError: <parseError>, base64Response: <response>`

akshaymankar (approved):
Assuming its safe to print such details.

PR #4566, opened 2025-05-06
Fix docker-compose redis replication

supersven:
## Checklist

 - [ ] Add a new entry in an appropriate subdirectory of `changelog.d`
 - [x] Read and follow the [PR guidelines](https://docs.wire.com/latest/developer/developer/pr-guidelines.html)

PR #4559, opened 2025-04-30
Add a webapp url field to email domain registration (1)

supersven:
I'm splitting https://github.com/wireapp/wire-server/pull/4542 because it's growing out of bounds.

This PR cares about these endpoints (V9+):

- [`POST /get-domain-registration`](https://wearezeta.atlassian.net/wiki/spaces/ENGINEERIN/pages/150930346/Use+case+connect+to+a+custom+backend+via+domain+redirect#API-v9%2B)
- [`POST /domain-verification/{domain}/backend`](https://wearezeta.atlassian.net/wiki/spaces/ENGINEERIN/pages/1570832467/Email+domain+registration+and+configuration#POST-%2Fdomain-verification%2F%7Bdomain%7D%2Fbackend)

Ticket: https://wearezeta.atlassian.net/browse/WPB-17032

## Checklist

 - [x] Add a new entry in an appropriate subdirectory of `changelog.d`
 - [x] Read and follow the [PR guidelines](https://docs.wire.com/developer/developer/pr-guidelines.html)
 - [x] Run `make git-add-cassandra-schema` to update the cassandra schema documentation

PR #4525, opened 2025-04-07
[WPB-1216] servantify proxy (part 2, again)

fisx:
https://wearezeta.atlassian.net/browse/WPB-1216

## Checklist

 - [x] Add a new entry in an appropriate subdirectory of `changelog.d`
 - [x] Read and follow the [PR guidelines](https://docs.wire.com/developer/developer/pr-guidelines.html)

PR #4524, opened 2025-04-04
fix(redis-ephemeral): bump to 6.2.16

lwille:
related to WPB-15658 - we did bump the Helm chart, but it didn't contain the latest available image.

## Checklist

 - [x] Add a new entry in an appropriate subdirectory of `changelog.d`
 - [x] Read and follow the [PR guidelines](https://docs.wire.com/developer/developer/pr-guidelines.html)

PR #4573, opened 2025-05-12
Release 2025-05-12 - (expected chart version 5.15.0)

zebot:
# [2025-05-12] (Chart Release 5.15.0)

## API changes

* From API version V9 on, the `POST /domain-verification/{domain}/backend` and
  `POST /get-domain-registration` endpoints are adjusted to also carry the
  `webapp_url` in their payloads. The structure of these payloads changes as
  well: The former `backend_url` and the new `webapp_url` are now combined in one
  object in the `backend` field:

  ```json
  {
  ...
    "backend": {
      "config_url": "{url}",
      "webapp_url": "{url}"
    }
  }
  ```

  The same change is applied to the internal endpoints `PUT
  /i/domain-registration` and `GET /i/domain-registration`. (#4559)

* Add a new endpoint `/mls/reset-conversation` which can be used to restore an MLS group that ended up in an invalid state for any reason. After resetting, the conversation has the same users, but the corresponding MLS group gets a new group ID and resets to epoch 0 with no clients. (#4558)

## Features

* Team admins can create a channel without joining (#4527, #4553)

* Only forward one conversation create event to pydio (#4535, #4551)

* Fixed channel permissions for external partners. They are allowed to add members if they are channel admins or if add-permissions are set to everyone. (#4534)

* gundeck: Send notifications to temp clients only when there are no rabbitmq clients in the recipient list.  Before this, it was wrongly sending the notification to all recipients. (#4556)

* charts/{cannon,nginz}: Add the events endpoint to nginz config (#4540, #4540)

* Lazy streams were broken due to
  https://github.com/haskell-servant/servant/pull/1781 . So, in specific cases,
  the playload of a streamed response was realised in the application's memory
  instead of streaming it piecewise. (#4538)

* Use [multipart upload](https://docs.aws.amazon.com/AmazonS3/latest/userguide/mpuoverview.html) to S3 for all assets (#4548)

## Internal changes

* Remove wai-routes, wai-predicates deps from proxy and translate routing table and handlers to servant.  New integration tests for proxy featuring mock services. (#4525)

* Include untouched base64-encoded authentication response from http request body in error messages, not some intermediate parse result. (#4570)

* Move zauth logic from brig to wire-subsystems (part 1: Brig.ZAuth, simplify implementation) (#4479)

* Fix Redis replication in our docker-compose env (used to run integration tests
  locally): We allow only TLS connections to the Redis nodes. Thus, replication
  has to use TLS as well (by default it doesn't.) (#4566)

* Bump redis version used by redis-ephemeral to 6.2.16 (#4524)

* The backend now verifies that new leaf nodes occurring in an MLS commit match the signature key registered by the corresponding client (#4531)

## Federation changes

* Federation API version V2 is finalized. (#4546)


stefanwire:
Didn't pass our quality checks.

PR #4574, opened 2025-05-12
Wpb 16876 allow team admin to add users and user groups to a channel they are not in

battermann:
## Checklist

### DIFF
--- a/integration/test/Test/Channels.hs
+++ b/integration/test/Test/Channels.hs
@@ -411,3 +411,18 @@ testTeamAdminCanCreateChannelWithoutJoining = do
   postConversation owner defMLS {groupConvType = Just "channel", team = Just tid, skipCreator = Just True} `bindResponse` \resp -> do
     resp.status `shouldMatchInt` 201
     resp.json %. "members" `shouldMatch` ([] :: [Value])
+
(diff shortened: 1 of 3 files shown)


integration/test/Test/Channels.hs
```diff
@@ -408,6 +411,111 @@ testTeamAdminCanCreateChannelWithoutJoining = do
...
+consumingMessagesExpectNoMemberJoin mp = Codensity $ \k -> do
+  conv <- getMLSConv mp.convId
+  let oldClients = Set.delete mp.sender conv.members
+  let newClients = Set.delete mp.sender conv.newMembers
+  let clients =
+        map (,MLSNotificationMessageTag) (toList oldClients)
+          <> map (,MLSNotificationWelcomeTag) (toList newClients)
+
+  withWebSockets (map fst clients) $ \wss -> do
+    r <- k ()
```
supersven, line 520: Hm, I'm wondering if it's okay to call the continuation before some other computations. :thinking: (I really don't know.)

battermann: It has to be like that, because the continuation in this case is the sending of the commit bundle. This has to be done before notifications can be consumed (the other computations). But thanks for the comment, this lead to a small clean up. :)

PR #4588, opened 2025-05-20
WPB-17597 stern backoffice endpoint to generically lock or unlock any team feature

battermann:
https://wearezeta.atlassian.net/browse/WPB-17597

## Checklist

### DIFF
--- a/integration/test/Test/FeatureFlags/ConferenceCalling.hs
+++ b/integration/test/Test/FeatureFlags/ConferenceCalling.hs
@@ -1,17 +1,20 @@
 module Test.FeatureFlags.ConferenceCalling where
 
+import qualified API.GalleyInternal as Internal
+import qualified Data.Aeson as A
+import SetupHelpers (createTeam)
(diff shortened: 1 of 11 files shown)


integration/test/Test/FeatureFlags/ConferenceCalling.hs
```diff
@@ -24,3 +27,52 @@ testConferenceCalling access = do
     & addUpdate (confCalling def {sft = toJSON True})
     & addUpdate (confCalling def {sft = toJSON False})
     & addInvalidUpdate (confCalling def {sft = toJSON (0 :: Int)})
+
+checkPatchConferenceCalling ::
```
pcapriotti, line 31: Maybe this could be abstracted a bit to avoid copy-pasting the whole thing? One idea could be to pass a function that computes the expected status (or maybe the whole expected value of the feature).

pcapriotti (approved):
Looks good.

PR #4545, opened 2025-04-16
[WPB-16870] introduce user groups subsystem

fisx:
https://wearezeta.atlassian.net/browse/WPB-16870

this adds a wire subsystem for user groups with tests and in-memory interpreter.

coming up in future PRs:
- brig public api.
- postgres interpreter.

## Checklist

- [ ] coordinate with platform before you merge this PR, so that all the envs can be configured correctly in advance
 - [x] Add a new entry in an appropriate subdirectory of `changelog.d`
 - [X] ~~Read and follow the [PR guidelines](https://docs.wire.com/developer/developer/pr-guidelines.html) ([can't at the moment](https://wearezeta.atlassian.net/browse/WPB-17187))~~

PR #4591, opened 2025-05-21
[WPB-16870] user groups create and get

supersven:
Restore the reverted 61b9fda5578790bb0e1fd3337d02fafa1bc336a1 / https://github.com/wireapp/wire-server/pull/4545 (reverted by ddd0e33e9f0d93fdb97560c38e020da10a867b04 / https://github.com/wireapp/wire-server/pull/4590). 

We had to revert, because the test environments aren't prepared for it, yet. The Postgres password needs to be taken from a secret and handed to the corresponding Helm secret.

## Checklist
 - [x] coordinate with platform before you merge this PR, so that all the envs can be configured correctly in advance


fisx (approved):

charts/brig/values.yaml
```diff
@@ -78,6 +78,14 @@ config:
     # tlsCaSecretRef:
     #   name: <secret-name>
     #   key: <ca-attribute>
+
+  # Postgres connection settings.
+  # Don't forget to set the user's password in `secrets.pgPassword`
```
smatting, line 83: Maybe metion that all https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-PARAMKEYWORDS are posssible here, including pgpassfile.

supersven: Sorry, I've overlooked those. Adding them.

supersven: Okay, I've tried to explain this a bit more.

Now, I'm wondering if we really need `postgresqlPassword`. Or, if it cannot be defined via `passfile` (https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-CONNECT-PASSFILE) :thinking:

supersven: Thankfully, @battermann remembered the details: Providing the password file via a volume mount won't work (we tried that.) The issue is the `brig` runs with a random uid and the library (`libpq-connect`) expects an access right mask of `0600` on the password file.

I've now added this insight to the docs.

battermann: Yes, only `brig` runs with an unprivileged user, not a random one.

PR #4605, opened 2025-06-04
Fix: postgres migrations, added migration for foreign key constraint

battermann:
## Checklist


libs/wire-subsystems/src/Wire/UserGroupSubsystem/Interpreter.hs
```diff
@@ -169,7 +169,7 @@ deleteGroupImpl ::
   Sem r ()
 deleteGroupImpl deleter groupId =
   getTeamAsMember deleter >>= \case
-    Nothing -> pure ()
+    Nothing -> throw UserGroupNotFound
```
fisx, line 172: don't we usually make delete idempotent?

fisx: (hum, but the tests also assume it's *not* idempotent.)

PR #4615, opened 2025-06-13
notifications: limit RabbitMQ prefetch_count

supersven:
Limit the amount of unacknowledged messages the RabbitMQ notifications consumer
receives to `100`. Beyond this limit, received messages have to be acknowledged
to receive new ones. This prevents overloading the consumer with new messages.

Ticket: https://wearezeta.atlassian.net/browse/WPB-17896

## Checklist

### DIFF
--- a/integration/test/Test/Events.hs
+++ b/integration/test/Test/Events.hs
@@ -701,17 +701,19 @@ testQosLimit = do
 
   runCodensity (createEventsWebSocket alice (Just cid)) \ws -> do
     -- The order of message_count and user.client-add events is not forseeable.
-    -- Thus, we have to handle (ack) them in any order.
-    void $ assertClientAddOrMessageCountEvent ws cid (550, 550 + 1)
-    void $ assertClientAddOrMessageCountEvent ws cid (550, 550 + 1)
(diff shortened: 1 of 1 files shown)


integration/test/Test/Events.hs
```diff
@@ -679,23 +679,55 @@ testMessageCount = do
...
+      e %. "data.event.payload.0.client.id" `shouldMatch` cid
+      e %. "data.delivery_tag"
+    sendAck ws deliveryTag False
+
+    es <- consumeAllEventsNoAck ws
+    assertBool "First 100 events" $ length es == 100
+
+    forM_ es $ \e -> do
+      t <- e %. "type" & asString
+      if t == "message_count" then ackMessageCount ws else ackEvent ws e
```
battermann: How can this ever be a `message_count` event?

battermann (approved):

integration/test/Test/Events.hs
supersven: You're right: This can be simplified :+1:

supersven:
The test ran 8 times green in a row, so I'm considering it to be stable now.

PR #4618, opened 2025-06-16
Add TeamId to conversation.delete event

supersven:
Ticket: https://wearezeta.atlassian.net/browse/WPB-17738

## Checklist

### DIFF
--- a/integration/test/Test/Cells.hs
+++ b/integration/test/Test/Cells.hs
@@ -77,6 +77,24 @@ testCellsCreationEvent = do
 
   assertNoMessage q
 
+testCellsDeletionEvent :: (HasCallStack) => App ()
+testCellsDeletionEvent = do
+  -- start watcher before creating conversation
+  q0 <- watchCellsEvents def
+  (alice, tid, _) <- createTeam OwnDomain 1
+  conv <- postConversation alice defProteus {team = Just tid, cells = True} >>= getJSON 201
+  void $ deleteTeamConversation tid conv alice >>= assertSuccess
+
+  let q = q0 {filter = isConvDeleteNotif} :: QueueConsumer
+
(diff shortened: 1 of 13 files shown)


services/galley/src/Galley/API/Action.hs
```diff
@@ -862,9 +863,10 @@ updateLocalConversationUnchecked ::
   Local Conversation ->
   Qualified UserId ->
   Maybe ConnId ->
+  Maybe TeamId ->
   ConversationAction tag ->
   Sem r LocalConversationUpdate
-updateLocalConversationUnchecked lconv qusr con action = do
+updateLocalConversationUnchecked lconv qusr con tid action = do
```
battermann: I would prefer to use the team ID from the metadata of the `Conversation` value that is passed here. So that we do not have to thread the `tid` value through all the way from the `deleteTeamConverstion` handler. That would be less noisy and potentially reduce merge conflicts.

changelog.d/1-api-changes/add-teamid-conversation.delete
```diff
@@ -0,0 +1,2 @@
+Add the `TeamId` to the payload of `conversation.delete` events. The field's
+path is `data.teamId`.
```
battermann:
```suggestion
path is `team`.
```

battermann (changes requested):

services/galley/src/Galley/API/Action.hs
supersven: Thanks for the hint! (I wasn't aware that the `TeamID` is there as well) :+1:

supersven: I've adjusted the code now accordingly.

battermann (approved):

PR #4621, opened 2025-06-19
Fix spelling of `validateSAMLEmails` feature (backwards compat).

fisx:
## Checklist


changelog.d/0-release-notes/WPB-18206-scim-update-email-fix-spelling
```diff
@@ -0,0 +1 @@
+If you had *not* set the `validateSAMLEmails` feature in your local instance settings, the default will change from `disabled` to `enabled` with this release.  See also "bug fixes".  Please make sure that is what you want, or provide your own override.
```
battermann:
```suggestion
If you had *not* set the `validateSAMLEmails` feature in your local instance settings, the default will change from `disabled` to `enabled` with this release.  See also "bug fixes".  Please make sure that this is what you want, or provide your own override.
```

changelog.d/3-bug-fixes/WPB-18206-scim-update-email-fix-spelling
```diff
@@ -0,0 +1 @@
+The feature flag `validateSAMLEmails` has (almost) consistently been spelled with `E`, but in `values.yaml` it was consistent with the rest api, where it was spelled with `e`.  As a consequence, the default for people who didn't set it has been `disabled`, not `enabled`, as you might conclude from reading `values.yaml`.  This has been fixed, `E` is now used everywhere.
```
battermann: Hm, but the hard coded default in `libs/wire-api/src/Wire/API/Team/Feature.hs` is enabled. Which would be the fall back when not set (or set incorrectly), no?

libs/wire-api/src/Wire/API/Routes/Public/Spar.hs
```diff
@@ -130,7 +130,7 @@ type APIIDP =
     :<|> Named "idp-get-raw" (ZOptUser :> IdpGetRaw)
     :<|> Named "idp-get-all" (ZOptUser :> IdpGetAll)
     :<|> Named "idp-create@v7" (Until 'V8 :> ZOptUser :> IdpCreate) -- (change is semantic, see handler)
-    :<|> Named "idp-create" (ZOptUser :> IdpCreate)
+    :<|> Named "idp-create" (From 'V8 :> ZOptUser :> IdpCreate)
```
battermann, line 133: What does this have to do with this PR?

changelog.d/3-bug-fixes/WPB-18206-scim-update-email-fix-spelling
fisx: the default in /libs/wire-api is "disabled", the default in the charts is "enabled".  but since the default in the charts had a typo, it never had any effect, and the haskell default kicked in instead.

that's my understanding anyway?

changelog.d/0-release-notes/WPB-18206-scim-update-email-fix-spelling
fisx: i think they're both correct, but sure.

libs/wire-api/src/Wire/API/Routes/Public/Spar.hs
fisx: it's called a dry-by fix :-)  ("fix" because i don't like our practice of not providing lower bounds if we know them, but we can debate that of course.  should i debate first, and remove this until we've reached a conclusion?)

changelog.d/0-release-notes/WPB-18206-scim-update-email-fix-spelling
battermann: Ah, yes, the way I read it first sounded incorrect to me, but you are right.

libs/wire-api/src/Wire/API/Routes/Public/Spar.hs
battermann: Oh, yes, it should be like this. I missed the line above.

fisx:
note to self:

```
----- Spar.testSsoLoginNoSamlEmailValidation[validatesamlemails=false] FAIL (0.08 s) -----
assertion failure:
Actual:
404
Expected:
in range (200, 299) (including bounds)
```

PR #4624, opened 2025-06-20
disallow email change to blocked domain

supersven:
This changes the rules for blocked email domains:
- Users cannot activate their accounts with a blocked email (this stays the same)
- Users with blocked domains cannot be invited to teams
- Users cannot change their email address to one of a blocked domain

The pre-condition to merge is is that legitimate domains are not part of the blocked domains set anymore. In the past, legitimate domains were blocked to force users to be invited to teams. This should now be handled by the new enterprise login feature(s). 

Ticket: https://wearezeta.atlassian.net/browse/WPB-18324

## Checklist
 - [x] **Do not merge this before we've got a green light from Guillaume that all regular users (teams) have been migrated to enterprise login**


integration/test/API/Brig.hs
```diff
@@ -422,12 +422,21 @@ putSelfLocale caller locale = do
...
-  callerid <- asString $ caller %. "id"
-  req <- baseRequest caller Brig Versioned $ joinHttpPath ["users", callerid, "email"]
+putUserEmail :: (HasCallStack, MakesValue caller, MakesValue target) => caller -> target -> String -> App Response
+putUserEmail caller target emailAddress = do
+  uid <- asString $ target %. "id"
+  req <- baseRequest caller Brig Versioned $ joinHttpPath ["users", uid, "email"]
   submit "PUT" $ req & addJSONObject ["email" .= emailAddress]
 
+putSelfEmail :: (HasCallStack, MakesValue user) => user -> String -> String -> String -> App Response
+putSelfEmail user cookie token emailAddress = do
```
fisx, line 432: nice, much less...  weird.  :)

services/brig/src/Brig/Options.hs
```diff
@@ -729,43 +729,23 @@ parseUserFeatureStatus = withObject "UserFeatureStatus" $ \obj -> obj .: "status
...
-    --   broken by software or usage errors.  Example: is it even legal for the owner of an
-    --   email domain to keep users from signing up with wire using that email?  This is
-    --   possibly true for domains like @mystartup.com@, but what about
-    --   @globalmailhosting.com@?  Other example: next point.
-    --
-    -- * We could implement something more sophisticated involving 'TXT' DNS records that the
-    --   team admin needs to set up containing some secrets obtainable via team settings.
-    --   This is a lot more involved to implement, and very easy for coders or users to get
-    --   wrong.
+  { -- TODO: Update comment
```
fisx: wanna fix that?

supersven: Oh, that was some left-over :sweat_smile: The comment is up-to-date

services/brig/src/Brig/Options.hs
```diff
@@ -729,43 +729,21 @@ parseUserFeatureStatus = withObject "UserFeatureStatus" $ \obj -> obj .: "status
...
+    --   - Team invitations
+    --   - Changes of the account's email address
     domainsBlockedForRegistration :: DomainsBlockedForRegistration
   }
   deriving (Show, FromJSON, Generic)
 
 -- | See also: "Galley.API.CustomBackend", `galley.custom_backend`.
-newtype DomainsBlockedForRegistration = DomainsBlockedForRegistration [Domain]
-  deriving newtype (Show, FromJSON, Generic)
+newtype DomainsBlockedForRegistration = DomainsBlockedForRegistration (HashSet Domain)
```
fisx, line 745: did you check the boundary where lists are faster on the one side, and hashsets on the other?  ;-)  (i suppose this is better because we don't need to worry about block list growth for longer.)  :+1:

libs/wire-subsystems/test/unit/Wire/UserSubsystem/InterpreterSpec.hs
```diff
@@ -93,7 +96,7 @@ spec = describe "UserSubsystem.Interpreter" do
             mkUserIds domain users = map (flip Qualified domain . (.id)) users
             onlineUsers = mkUserIds onlineDomain onlineTargetUsers
             offlineUsers = mkUserIds offlineDomain offlineTargetUsers
-            config = UserSubsystemConfig visibility miniLocale False 100 undefined
+            config = UserSubsystemConfig visibility miniLocale False 100 undefined []
```
fisx: maybe switch to the `def { ... }` pattern like for `/integration/test/API/Brig.hs:AddUser`?

libs/wire-subsystems/test/unit/Wire/UserSubsystem/InterpreterSpec.hs
```diff
@@ -915,6 +918,31 @@ spec = describe "UserSubsystem.Interpreter" do
               === ( ChangeEmailResponseNeedsActivation,
                     [user {emailUnvalidated = Just updatedEmail}]
                   )
+    prop "Email change not allowed for blocked domains" $ do
+      \(locx :: Local ())
```
fisx: how can this have different values?  or why do you `generate` is with qc?

libs/wire-subsystems/test/unit/Wire/TeamInvitationSubsystem/InterpreterSpec.hs
```diff
@@ -180,3 +184,62 @@ spec = do
...
+
+    prop "try to invite to blocked domain" $
+      \(tid :: TeamId)
+       (preExistingPersonalAccount :: Maybe User)
+       (preExistingInviteeEmail :: EmailAddress)
+       (emailUsername :: EmailUsername)
+       (blockedDomains :: NonEmptyList Domain) -> do
+          let hasEmailIdentity user = case user.userIdentity of
+                Just (EmailIdentity _) -> True
+                _ -> False
```
fisx:
```suggestion
          let hasEmailIdentity user = isJust (emailIdentity user)
```

NB: this does not have the same semantics, but here we probably don't care either way; see `/libs/wire-api/src/Wire/API/User/Identity.hs:emailIdentity`.

libs/wire-subsystems/src/Wire/UserSubsystem.hs
```diff
@@ -284,6 +290,19 @@ requestEmailChange lusr email allowScim = do
         (activationCode adata)
         (Just (userLocale usr))
 
+    guardBlockedDomainEmail ::
```
fisx, line 292: this is duplicated in team invitation subsystem, but slightly differently.  maybe there should be a predicate `isEmailDomainBlocked` near...  hm, where could it go?  and is it enough code duplication to worry about it.

probably not...

services/brig/src/Brig/API/User.hs
```diff
@@ -804,6 +811,19 @@ sendActivationCode email loc = do
           _otherwise ->
             liftSem $ (maybe sendActivationMail (const sendEmailAddressUpdateMail) ident) em name aKey aCode loc'
 
+    guardBlockedDomainEmail ::
```
fisx, line 820: and here's a third one, hehe...

fisx (approved):
ci isn't happy yet, but it all looks good to me!

fisx:
tests: something wrong with the guard functions?  can't find anything.  or some domains are blocked in the setup, and didn't break the tests earlier because the blocking only applied to creating new accounts from scratch?

services/brig/src/Brig/Options.hs
supersven: One important detail probably is that the `HashSet` is constructed on application startup. I.e. we pay a price for the hashes, but only once.

My main concern was that we may iterate over a long list for every request. The corner cases with short lists are interesting on a theoretical level; however, both data structures would be good enough™ in practice for them.

libs/wire-subsystems/src/Wire/UserSubsystem.hs
supersven: The main issue here is, that the types (especially effect constraints) are different :thinking: I'm bot sure how to solve that without introducing accessor type classes over the effect's configs. :thinking: (And, that might be a bit too much added complexity for a bit of less code duplication :thinking: )

supersven:
> tests: something wrong with the guard functions? can't find anything. or some domains are blocked in the setup, and didn't break the tests earlier because the blocking only applied to creating new accounts from scratch?

I'd bet that the tests failed due to general CI flakiness issues :roll_eyes:

libs/wire-subsystems/test/unit/Wire/UserSubsystem/InterpreterSpec.hs
supersven: I thought that more variance would lead to better QC tests. However, I might have over done it :sweat_smile: I've now removed the variance over `locx`.

blackheaven (approved):

supersven:
Rebased to be prepared to merge this soon (not now.)

PR #4626, opened 2025-06-24
gundeck: Send push notifications to clients with `consumable-notifications` capability

akshaymankar:
https://wearezeta.atlassian.net/browse/WPB-18369

## Checklist

### DIFF
--- a/services/gundeck/src/Gundeck/Push.hs
+++ b/services/gundeck/src/Gundeck/Push.hs
@@ -80,7 +80,7 @@ import Wire.API.Presence qualified as Presence
 import Wire.API.Push.Token qualified as Public
 import Wire.API.Push.V2
 import Wire.API.User (UserSet (..))
-import Wire.API.User.Client (Client (..), ClientCapability (..), ClientCapabilityList (..), UserClientsFull (..), supportsConsumableNotifications)
+import Wire.API.User.Client (Client (..), UserClientsFull (..), supportsConsumableNotifications)
 
 push :: [Push] -> Gundeck ()
 push ps = do
--- a/services/gundeck/src/Gundeck/Push.hs
+++ b/services/gundeck/src/Gundeck/Push.hs
@@ -178,8 +178,7 @@ splitPush clientsFull p = do
               Set.filter (\c -> c.clientId `elem` toList cs) allClients
             RecipientClientsAll -> allClients
             RecipientClientsTemporaryOnly -> mempty
-          isClientForRabbitMq c = supportsConsumableNotifications c
-          (rabbitmqClients, legacyClients) = Set.partition isClientForRabbitMq relevantClients
+          (rabbitmqClients, legacyClients) = Set.partition supportsConsumableNotifications relevantClients
           rabbitmqClientIds = (.clientId) <$> Set.toList rabbitmqClients
           legacyClientIds = (.clientId) <$> Set.toList legacyClients
       case (rabbitmqClientIds, legacyClientIds) of
--- a/services/gundeck/src/Gundeck/Push.hs
+++ b/services/gundeck/src/Gundeck/Push.hs
@@ -33,7 +33,6 @@ where
 import Bilge qualified
 import Bilge.RPC qualified as Bilge
 import Control.Error
-import Control.Exception (ErrorCall (ErrorCall))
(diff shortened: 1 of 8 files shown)


services/gundeck/src/Gundeck/Push.hs
```diff
@@ -180,7 +178,7 @@ splitPush clientsFull p = do
               Set.filter (\c -> c.clientId `elem` toList cs) allClients
             RecipientClientsAll -> allClients
             RecipientClientsTemporaryOnly -> mempty
-          isClientForRabbitMq c = ClientSupportsConsumableNotifications `Set.member` c.clientCapabilities.fromClientCapabilityList
+          isClientForRabbitMq c = supportsConsumableNotifications c
```
supersven:
```suggestion
          isClientForRabbitMq = supportsConsumableNotifications
```
Can't we eta-reduce here?

akshaymankar: We can probably even get rid of the helper :smile:

services/gundeck/test/unit/MockGundeck.hs
```diff
@@ -539,16 +556,44 @@ handlePushCass Push {..}
   | _pushTransient = pure ()
 handlePushCass Push {..} = do
   forM_ _pushRecipients $ \(Recipient uid _ cids) -> do
+    clients <- Set.toList . Set.unions . Map.elems . (.userClientsFull) <$> mpaGetClients (Set.singleton uid)
+    let consumabeNotifClients = map (.clientId) $ filter supportsConsumableNotifications clients
```
supersven:
```suggestion
    let consumableNotifClients = map (.clientId) $ filter supportsConsumableNotifications clients
```
Typo.

services/gundeck/src/Gundeck/Push.hs
```diff
@@ -249,29 +257,36 @@ pushAllLegacy pushes = do
...
+  unless (psh ^. pushTransient) $
+    mpaRunWithBudget cost () $
+      mpaPushNative notif (psh ^. pushNativePriority) =<< nativeTargets psh rcps' alreadySent
+
+pushAllViaRabbitMq :: (MonadPushAll m, MonadMapAsync m, MonadNativeTargets m) => [NewNotification] -> m ()
+pushAllViaRabbitMq newNotifs = do
+  for_ newNotifs $ pushViaRabbitMq
+  mpaForkIO $ do
+    for_ newNotifs $ \newNotif ->
+      pushNativeWithBudget newNotif.nnNotification newNotif.nnPush []
```
supersven, line 282: Just to check my understanding: Is this the core of this PR; i.e. was the bug that there were no SNS notifications for clients listening to RabbitMQ? :thinking:

services/gundeck/test/unit/MockGundeck.hs
```diff
@@ -2,6 +2,7 @@
 {-# LANGUAGE RecordWildCards #-}
```
supersven, line 2: I'm wondering why you haven't had to touch any test to validate the changed behavior. Though, it can well be that I just don't understand the impact of the changes to the mock well enough. :thinking:

supersven (changes requested):
The questions should probably be solved before green lightening this PR. However, it's already looking pretty good.

services/gundeck/src/Gundeck/Push.hs
supersven: Beta reduction seems to be fitting as well :upside_down_face:

services/gundeck/src/Gundeck/Push.hs
akshaymankar: Yes, exactly.

services/gundeck/test/unit/MockGundeck.hs
akshaymankar: The test is just comparing the `MockState` after running the mock type instance and the real type instance with the `MockGundeck`. So, changing the mock instance to do what we expect was enough to make the test correct.

services/gundeck/test/unit/MockGundeck.hs
akshaymankar: Fixed.

supersven (approved):
LGTM :+1:

PR #4629, opened 2025-06-26
Revert change in spelling of validateSAML{e,E}mails feature.

fisx:
We missed two details when "fixing" this: (1) we broke `GET /v8/feature-configs` because we forgot to version-control the response body; (2) events.

Until we have answers to these two points, I'll revert this and we'll keep the typo for backwards compatibility.

## Checklist

### DIFF
--- a/libs/wire-api/src/Wire/API/Routes/Public/Galley/Feature.hs
+++ b/libs/wire-api/src/Wire/API/Routes/Public/Galley/Feature.hs
@@ -95,11 +95,10 @@ type family AllDeprecatedFeatureConfigAPI cfgs where
       :<|> AllDeprecatedFeatureConfigAPI cfgs
 
 type DeprecatedFeatureAPI =
-  FeatureStatusDeprecatedGet DeprecationNotice1 V2 SearchVisibilityAvailableConfig
-    :<|> FeatureStatusDeprecatedPut DeprecationNotice1 V2 SearchVisibilityAvailableConfig
-    :<|> FeatureStatusDeprecatedGet DeprecationNotice1 V2 ValidateSAMLEmailsConfig
-    :<|> FeatureStatusDeprecatedGet DeprecationNotice1 V9 ValidateSAMLEmailsConfig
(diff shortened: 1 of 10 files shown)


stefanwire (approved):

libs/wire-api/src/Wire/API/Routes/Public/Galley/Feature.hs
```diff
@@ -95,11 +95,10 @@ type family AllDeprecatedFeatureConfigAPI cfgs where
       :<|> AllDeprecatedFeatureConfigAPI cfgs
 
 type DeprecatedFeatureAPI =
-  FeatureStatusDeprecatedGet DeprecationNotice1 V2 SearchVisibilityAvailableConfig
-    :<|> FeatureStatusDeprecatedPut DeprecationNotice1 V2 SearchVisibilityAvailableConfig
-    :<|> FeatureStatusDeprecatedGet DeprecationNotice1 V2 ValidateSAMLEmailsConfig
-    :<|> FeatureStatusDeprecatedGet DeprecationNotice1 V9 ValidateSAMLEmailsConfig
-    :<|> FeatureStatusDeprecatedGet DeprecationNotice2 V2 DigitalSignaturesConfig
+  FeatureStatusDeprecatedGet DeprecationNotice1 SearchVisibilityAvailableConfig V2
```
supersven, line 98: What's the benefit of switching `untilVersion` and `feature`? :thinking:

fisx: this way it's in the same order everywhere.  (while i was staging this change, i though "maybe i shouldn't have done that, it worked fine before!", but then it was already staged...  :)

supersven: Well, if it aligns with other things: Fine by me :+1:

fisx:
> I'm wondering if we cannot provide a golden test to guard against such issues in future? 🤔

see today's watercooler discussion :)

PR #4630, opened 2025-06-26
WPB-17885 EphemeralUserCreation feature flag

battermann:
## Checklist

### DIFF
--- a/docs/src/developer/reference/config-options.md
+++ b/docs/src/developer/reference/config-options.md
@@ -247,12 +247,12 @@ fileSharing:
 
 These are all the possible combinations of `status` and `lockStatus`:
 
-| `status`   | `lockStatus`   |                                                   |
-|------------|----------------|---------------------------------------------------|
-| `enabled`  | `locked`       | Feature enabled, cannot be disabled by team admin |
-| `enabled`  | `unlocked`     | Feature enabled, can be disabled by team admin    |
-| `disabled` | `locked`       | Feature disabled, cannot be enabled by team admin |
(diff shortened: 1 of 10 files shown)


docs/src/developer/reference/config-options.md
```diff
@@ -946,6 +946,17 @@ brig:
       setAuditLogEmailRecipient: security@wire.com
 ```
 
+### Ephemeral User Creation
+
+Ephemeral user creation is enabled per default. To disable the feature set `setEphemeralUserCreationEnabled` to `false`:
```
supersven: I'm wondering if all readers of the docs know what an "ephemeral user" is. :thinking: Maybe, some short explanation would be good.

integration/test/Test/User.hs
```diff
@@ -381,3 +381,20 @@ testPasswordChange =
...
+testEphemeralUserCreation :: (HasCallStack) => TaggedBool "ephemeral-user-creation-enabled" -> App ()
+testEphemeralUserCreation (TaggedBool enabled) = do
+  withModifiedBackend
+    def
+      { brigCfg = setField "optSettings.setEphemeralUserCreationEnabled" enabled
+      }
+    $ \domain -> do
+      name <- randomName
+      req <- baseRequest domain Brig Versioned "/register"
+      bindResponse (submit "POST" $ req & addJSONObject ["name" .= name] & addHeader "X-Forwarded-For" "127.0.0.42") $ \resp -> do
```
supersven: These lines look pretty similar to `API.Brig.addUser`. I'm wondering if there's a reason to not use it? :thinking:

PR #4631, opened 2025-06-26
cannon: Replace message count mechanism with detecting the end of initial sync

akshaymankar:
https://wearezeta.atlassian.net/browse/WPB-18436

The message count recieved in the queue info gotten as response of declaring the
queue may include some expired messages which have not yet made it to the head
of the queue. This makes the message count mechanism useless to the clients for
detecting the end of initial sync.

This commit detects end of initial sync based on sending a transient
notification in the queue. If we recieve this notification back, it would mean
that the queue is empty and the initial sync is complete.

The timing of sending this notification must be carefully chosen, it is sent in
one of these two conditions:

  1. If the message count in the queue info is 0, the queue must be empty as the
  expired messages are already considered expired when they reach the head of
  the queue. So its a good time to publish our end-of-initial-sync notification.
  If the queue is indeed empty we'll recieve it and forward it to the client,
  letting it know that it can consider itself up to date.

  2. If number of unacked messages becomes 0 right after recieving an
  acknowledgement and the initial sync hasn't already completed, this could mean
  either the queue is empty or there is some delay in receiving the next message
  from RabbitMQ. Here we publish our end-of-initial-sync notification. If there
  was a delay in receiving the next message from RabbitMQ, the consumer thread
  will increase the count for unacked messages and we'll get another opportunity
  to check this. Otherwise, the consumer thread will recieve this
  end-of-initial-sync notification and forward it to the client.

## Checklist

### DIFF
--- a/changelog.d/2-features/end-of-initial-sync
+++ b/changelog.d/2-features/end-of-initial-sync
@@ -1,0 +1,1 @@
+cannon: Replace message count mechanism with detecting the end of initial sync
--- a/integration/test/Test/Events.hs
+++ b/integration/test/Test/Events.hs
@@ -53,13 +53,12 @@ testConsumeEventsOneWebSocket = do
   clientId <- objId client
 
   runCodensity (createEventsWebSocket alice (Just clientId)) $ \ws -> do
-    assertMessageCount_ ws
     deliveryTag <- assertEvent ws $ \e -> do
       e %. "type" `shouldMatch` "event"
       e %. "data.event.payload.0.type" `shouldMatch` "user.client-add"
       e %. "data.event.payload.0.client.id" `shouldMatch` clientId
       e %. "data.delivery_tag"
-    assertNoEvent_ ws
+    assertEndOfInitialSync ws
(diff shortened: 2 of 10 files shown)


services/cannon/src/Cannon/WS.hs
```diff
@@ -205,8 +207,14 @@ env ::
...
   RabbitMqPool ->
+  Int ->
   Env
-env leh lp gh gp = Env leh lp (Bilge.host gh . Bilge.port gp $ empty) (RequestId defRequestId)
+env externalHostname portnum gundeckHost gundeckPort logg manager websockets rabbitConnections rand clock drainOpts cassandra pool notificationTTL =
+  let upstream = (Bilge.host gundeckHost . Bilge.port gundeckPort $ empty)
+      reqId = RequestId defRequestId
+   in Env {..}
+
+-- Env leh lp  (RequestId defRequestId)
```
battermann: remove comment

battermann (approved):

PR #4632, opened 2025-06-26
[WPB-18631] Robustness against rabbitmq outage

fisx:
https://wearezeta.atlassian.net/browse/WPB-18631
(found working on https://wearezeta.atlassian.net/browse/WPB-18443)

## Checklist
 - [x] Wrote tests to motivate changes

### DIFF
--- a/services/background-worker/src/Wire/BackgroundWorker.hs
+++ b/services/background-worker/src/Wire/BackgroundWorker.hs
@@ -31,7 +31,7 @@ supervisor ::
   String ->
   -- | action
   AppT IO a ->
-  -- | the ascyn thread (result is `Nothing` if thread is cancelled)
+  -- | the async thread (result is `Nothing` if thread is cancelled)
   IO (Async (Maybe a))
 supervisor env workerName workerAction = async loop
   where


integration/test/Test/Events.hs
```diff
@@ -566,6 +622,7 @@ testChannelKilled = do
     -- Some times RabbitMQ still remembers connections from previous uses of the
     -- dynamic backend. So we wait to ensure that we kill connection only for our
     -- current.
+
```
supersven: Do we need this empty line?

services/background-worker/test/Test/Wire/BackendNotificationPusherSpec.hs
```diff
@@ -326,7 +327,7 @@ spec = do
           http2Manager = undefined
           statuses = undefined
           rabbitmqAdminClient = Just $ mockRabbitMqAdminClient mockAdmin
-          rabbitmqVHost = "test-vhost"
+          amqpEP = Q.AmqpEndpoint undefined undefined "test-vhost" undefined
```
supersven, line 330: Can't the whole `amqpEP` be `undefined` (instead of having almost al fields of `AmqpEndpoint` being undefined)? :thinking:

services/background-worker/src/Wire/BackendNotificationPusher.hs
```diff
@@ -339,18 +348,48 @@ startWorker rabbitmqOpts = do
...
   pure (chanRef, consumersRef)
+
+-- cleanup will run in a new thread when the signal is caught, so we need to use IORefs and
+-- specific exception types to message threads to clean up
+cancelWorker :: WorkerResult -> AppT IO ()
+cancelWorker (notifChanRef, notifConsumersRef) = do
+  -- Notification pusher thread
+  Log.info $ Log.msg (Log.val "Cancelling the notification pusher thread")
+  readIORef notifChanRef >>= traverse_ \chan -> do
+    Log.info $ Log.msg (Log.val "Got channel")
```
supersven: The channel's name might be interesting here as well.

services/background-worker/src/Wire/BackendNotificationPusher.hs
```diff
@@ -339,18 +348,48 @@ startWorker rabbitmqOpts = do
       void $
         async $
           liftIO $
-            openConnectionWithRetries env.logger rabbitmqOpts $
+            openConnectionWithRetries env.logger "BackendNotificationPusher" rabbitmqOpts $ -- can we find a bug here, too?
```
supersven: What about this? Shall we search a potential bug?

services/background-worker/test/Test/Wire/BackendNotificationPusherSpec.hs
fisx: i don't know, we did set rabbitmqVHost before, and i didn't see any reason to change that and deal with whatever happens?

services/background-worker/src/Wire/BackendNotificationPusher.hs
fisx: yes, but not in this PR.

services/background-worker/src/Wire/BackgroundWorker.hs
```diff
@@ -2,64 +2,77 @@
...
+-- restart.
+supervisor ::
+  forall a.
+  -- | for `runAppT` and logging
+  Env ->
+  -- | worker name
+  String ->
+  -- | action
+  AppT IO a ->
+  -- | the ascyn thread (result is `Nothing` if thread is cancelled)
```
supersven:
```suggestion
  -- | the async thread (result is `Nothing` if thread is cancelled)
```

services/background-worker/src/Wire/BackendNotificationPusher.hs
fisx: i can only find channelID, and that is only exported from Network.API.Internal, which is not exposed in the cabal file.  any hints where to find the name?

integration/test/Test/Events.hs
fisx: we do not :)

services/background-worker/src/Wire/BackendNotificationPusher.hs
supersven: Hm, maybe transform this into a `FUTUREWORK: ...` or a ticket? :thinking:

services/background-worker/src/Wire/BackendNotificationPusher.hs
supersven: Ah, that mess. Yeah, I remember we've tried that before ... :roll_eyes: 
Well, then, we may leave this for now.

services/background-worker/test/Test/Wire/BackendNotificationPusherSpec.hs
supersven: Finally, it's not really important. :smile: 

One could also ask if `AppT` isn't too big if we only need the `vhost`. Or, if `getRemoteDomains :: RabbitMqAdmin.AdminAPI (Servant.AsClientT IO) -> Text {- vhosy -} -> AppT IO [Domain]` wouldn't be sufficient. Hower, not our construction place right now :wink:

supersven (approved):
Unfortunately, it's pretty hard be ensure that all this is free of deadlocks, because there are locks in many places.

I haven't found any so far. So, I'm approving. :heavy_check_mark:

akshaymankar (changes requested):
Let's see if #4652 works.

akshaymankar:
Superseded by #4652

PR #4634, opened 2025-06-27
detached background worker 2nd attempt

supersven:
Actually, there isn't much we can do when a queue is externally deleted:
- There is a loop which handles the reconfiguration according to existing (!) queues
- Older federation backend versions (`V0` and `V1`) cannot be changed anymore.

To heal the symptoms of detached consumers, I've decreased the cleanup interval for local integration test runs to one Second. That should be a good tradeoff between the time to re-execute tests and not making to many bogus requests to RabbitMQ.

I've also added a test to ensure that reestablishing the consumer works.

Ticket: https://wearezeta.atlassian.net/browse/WPB-17247

## Checklist

### DIFF
--- a/services/background-worker/src/Wire/BackendNotificationPusher.hs
+++ b/services/background-worker/src/Wire/BackendNotificationPusher.hs
@@ -37,12 +37,11 @@ import Wire.BackgroundWorker.Options
 import Wire.BackgroundWorker.Util
 
 startPushingNotifications ::
-  IORef (Map Domain (Q.ConsumerTag, MVar ())) ->
   MVar () ->
   Q.Channel ->
   Domain ->
   AppT IO Q.ConsumerTag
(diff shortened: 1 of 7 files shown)


services/background-worker/src/Wire/BackendNotificationPusher.hs
```diff
@@ -41,7 +43,19 @@ startPushingNotifications ::
   AppT IO Q.ConsumerTag
 startPushingNotifications runningFlag chan domain = do
   lift $ ensureQueue chan domain._domainText
-  QL.consumeMsgs chan (routingKey domain._domainText) Q.Ack (void . pushNotification runningFlag domain)
+  QL.consumeMsgs'
```
fisx, line 46: This change is missing in the changelog.  maybe this?

```
Log rabbit-connection errors in notification pusher.
```

PR #4635, opened 2025-06-30
WPB-16886 Get channel data (team admin)

battermann:
## Checklist


libs/wire-api/src/Wire/API/Routes/Public/Galley/Conversation.hs
```diff
@@ -164,16 +164,28 @@ type ConversationAPI =
                :> MultiVerb1 'GET '[JSON] (VersionedRespond 'V5 200 "Conversation" ConversationV9)
            )
     :<|> Named
-           "get-conversation"
+           "get-conversation@v9"
            ( Summary "Get a conversation by ID"
                :> From 'V6
+               :> Until 'V9
```
akshaymankar:
```suggestion
               :> Until 'V10
```

This is a V10 change

libs/wire-api/src/Wire/API/Routes/Public/Galley/Conversation.hs
```diff
@@ -164,16 +164,28 @@ type ConversationAPI =
...
                :> CanThrow 'ConvAccessDenied
                :> ZLocalUser
                :> "conversations"
                :> QualifiedCapture "cnv" ConvId
                :> Get '[JSON] ConversationV9
            )
+    :<|> Named
+           "get-conversation"
+           ( Summary "Get a conversation by ID"
+               :> From 'V9
```
akshaymankar:
```suggestion
               :> From 'V10
```

integration/test/Test/Channels.hs
```diff
@@ -590,3 +590,18 @@ testAdminCanRemoveMemberWithoutJoining = do
       void $ mlsCliConsume convId def cid msgData
       r <- createPendingProposalCommit convId cid >>= sendAndConsumeCommitBundle
       shouldBeEmpty $ r %. "events"
+
```
akshaymankar, line 593: We should also add a test to ensure that admins can't get non-channel conversations.

services/galley/src/Galley/API/Query.hs
```diff
@@ -148,11 +151,65 @@ getUnqualifiedConversation ::
...
+  Local UserId ->
+  ConvId ->
+  Sem r Public.Conversation
+getUnqualifiedConversation lusr cnv = do
+  mChannel <- getChannel lusr cnv
+  case mChannel of
+    Just channel -> pure channel
+    Nothing ->
+      Mapping.conversationView (qualifyAs lusr ()) (Just lusr)
+        <$> getConversationAndCheckMembership (tUntagged lusr) (qualifyAs lusr cnv)
```
akshaymankar: Why do we get the conv data twice here in case it is not a channel? Perhaps we can change the `getConversationAndCheckMembership` to allow an admin to succeed the membership check?

akshaymankar (changes requested):

akshaymankar (approved):

PR #4636, opened 2025-06-30
Timeout for stuck websocket connections

pcapriotti:
This PR adds some cleanup logic for rabbitmq websocket connections that have been inactive for some configurable amount of time.

https://wearezeta.atlassian.net/browse/WPB-17895

## Checklist

### DIFF
--- a/services/cannon/src/Cannon/RabbitMqConsumerApp.hs
+++ b/services/cannon/src/Cannon/RabbitMqConsumerApp.hs
@@ -150,7 +150,7 @@ rabbitMQWebSocketApp uid mcid mSyncMarkerId e pendingConn =
         Log.info e.logg $
           Log.msg (Log.val "Closing websocket due to inactivity")
             . logClient
-        WS.sendCloseCode wsConn 1003 ("inactivity" :: ByteString)
+        WS.sendCloseCode wsConn 1002 ("inactivity" :: ByteString)
 
     handleClientMisbehaving wsConn =
       Handler $ \(err :: WebSocketServerError) -> do


stefanwire (approved):

services/cannon/src/Cannon/RabbitMqConsumerApp.hs
```diff
@@ -34,16 +35,24 @@ rabbitMQWebSocketApp uid mcid mSyncMarkerId e pendingConn = do
...
-    runWithChannel (chan, queueInfo) = bracket openWebSocket closeWebSocket $ \wsConn ->
+    runWithChannel (chan, queueInfo) = bracket openWebSocket closeWebSocket $ \conn ->
       ( do
+          activity <- newEmptyMVar
+          let wsConn =
+                WSConnection
+                  { inner = conn,
+                    activity,
+                    activityTimeout = 30000000, -- TODO
+                    pongTimeout = 30000000 -- TODO
```
akshaymankar: Pending TODOs. Perhaps these value should come from configs and maybe we can write a test then?

akshaymankar (commented):
Have you seen the function `withPingPong` from websockets? Perhaps we don't have to implement this.

pcapriotti:
> Have you seen the function `withPingPong` from websockets? Perhaps we don't have to implement this.

I've seen it, but I can't see how to use it to implement our desired behaviour (i.e. only submit pings after inactivity).

services/cannon/src/Cannon/RabbitMqConsumerApp.hs
pcapriotti: Something went wrong when I last rebased, it seems some commits got lost.

akshaymankar:
> I've seen it, but I can't see how to use it to implement our desired behaviour (i.e. only submit pings after inactivity). 

Hmm, perhaps its ok to just send pings all the time? The point of all this is to kill connections which have inactive clients.

pcapriotti:
Sorry about the broken PR. Should be ok to review now.

services/cannon/src/Cannon/RabbitMqConsumerApp.hs
```diff
@@ -111,6 +145,13 @@ rabbitMQWebSocketApp uid mcid mSyncMarkerId e pendingConn = do
                   . logClient
               WS.sendCloseCode wsConn 1003 ("websocket-failure" :: ByteString)
 
+    handleInactivity wsConn =
+      Handler $ \(_ :: InactivityTimeout) -> do
+        Log.info e.logg $
+          Log.msg (Log.val "Closing websocket due to inactivity")
+            . logClient
+        WS.sendCloseCode wsConn 1003 ("inactivity" :: ByteString)
```
akshaymankar:
```suggestion
        WS.sendCloseCode wsConn 1002 ("inactivity" :: ByteString)
```

https://developer.mozilla.org/en-US/docs/Web/API/CloseEvent/code#value

PR #4637, opened 2025-07-01
Move most API changes to V10

akshaymankar:
https://wearezeta.atlassian.net/browse/WPB-18530

## Checklist

 - [x] ~Add a new entry in an appropriate subdirectory of `changelog.d`~ No changelog, one already published changelog needed a change.

### DIFF
--- a/integration/test/Test/Swagger.hs
+++ b/integration/test/Test/Swagger.hs
@@ -13,7 +13,7 @@ import Testlib.Prelude
 import UnliftIO.Temporary
 
 existingVersions :: Set Int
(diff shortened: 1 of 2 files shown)


libs/wire-api/src/Wire/API/Routes/Public/Gundeck.hs
```diff
@@ -135,6 +135,7 @@ type TimeAPI =
     "get-server-time"
     ( Summary "Get the current server time"
         :> Description "Returns the current server time in UTC with seconds precision."
+        :> From 'V9
```
pcapriotti, line 138: There's no need to have this bound, because it's a new endpoint and won't show up in the old swagger anyway. But it's fine if you want to keep it.

PR #4639, opened 2025-07-02
WPB-17993 RabbitMQ 4.x compatibility

battermann:
https://wearezeta.atlassian.net/browse/WPB-17993

## Checklist

### DIFF
--- a/hack/helm_vars/rabbitmq/values.yaml.gotmpl
+++ b/hack/helm_vars/rabbitmq/values.yaml.gotmpl
@@ -1,23 +1,22 @@
-rabbitmq:
-  persistence:
-    enabled: false
-  auth:
-    username: {{ .Values.rabbitmqUsername }}
-    password: {{ .Values.rabbitmqPassword }}
-    tls:
-      enabled: true
-      failIfNoPeerCert: false
-      existingSecret: rabbitmq-certificate
-  service:
-    extraPorts:
-      - name: http-stats-ssl
-        port: 15671
-        protocol: TCP
-        targetPort: 15671
-  extraConfiguration: |-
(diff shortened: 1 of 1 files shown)


hack/helmfile.yaml.gotmpl
```diff
@@ -254,13 +254,15 @@ releases:
 
   - name: 'rabbitmq'
     namespace: '{{ .Values.namespace1 }}'
-    chart: '../.local/charts/rabbitmq'
+    chart: 'bitnami/rabbitmq'
+    version: '15.2.0'
```
supersven: Is there a reason to use `15.2.0` and not `15.2.5` or even a later version? To get RabbitMQ 4.0 (and not 4.1), chart version `15.5.3` might be a good choice.

deploy/dockerephemeral/docker-compose.yaml
```diff
@@ -318,7 +318,7 @@ services:
 
   rabbitmq:
     container_name: rabbitmq
-    image: rabbitmq:3.13.7-management-alpine
+    image: rabbitmq:4.1.1-management-alpine
```
supersven, line 321: Shouldn't this better refer to some `4.0` version? Ideally, being aligned with the version deployed on CI? :thinking:

hack/helmfile.yaml.gotmpl
supersven: image

deploy/dockerephemeral/docker-compose.yaml
battermann: Versions should be aligned, yes. Will make sure. Thanks!

PR #4643, opened 2025-07-03
charts/nginz: Remove unused default value

akshaymankar:
## Checklist


supersven (approved):
I've paired on this with @akshaymankar . The `doc` field is actually not used in the `nginz` chart. So, we can drop it :+1:

PR #4652, opened 2025-07-08
background-worker/dead-user-notifs-watcher: Reconnect when connection with RabbitMQ is broken

akshaymankar:
https://wearezeta.atlassian.net/browse/WPB-18631

Alternative approach for #4632 (Test is copied from there). 

## Checklist

### DIFF
--- a/libs/extended/src/Network/AMQP/Extended.hs
+++ b/libs/extended/src/Network/AMQP/Extended.hs
@@ -138,7 +138,7 @@ mkRabbitMqChannelMVar :: Logger -> AmqpEndpoint -> IO (MVar Q.Channel)
 mkRabbitMqChannelMVar l opts = do
   chanMVar <- newEmptyMVar
   connThread <-
-    async . openConnectionWithRetries l opts $
+    async . openConnectionWithRetries l opts Nothing $
       RabbitMqHooks
         { onNewChannel = \conn -> putMVar chanMVar conn >> forever (threadDelay maxBound),
           onChannelException = \_ -> void $ tryTakeMVar chanMVar,
@@ -160,9 +160,10 @@ withConnection ::
   (MonadIO m, MonadMask m) =>
   Logger ->
   AmqpEndpoint ->
+  Maybe Text ->
   (Q.Connection -> m a) ->
   m a
-withConnection l AmqpEndpoint {..} k = do
+withConnection l AmqpEndpoint {..} connName k = do
   -- Jittered exponential backoff with 1ms as starting delay and 1s as total
   -- wait time.
   let policy = limitRetriesByCumulativeDelay 1_000_000 $ fullJitterBackoff 1000
@@ -180,13 +181,13 @@ withConnection l AmqpEndpoint {..} k = do
           )
           ( const $ do
               Log.info l $ Log.msg (Log.val "Trying to connect to RabbitMQ")
-              connOpts <- mkConnectionOpts AmqpEndpoint {..}
+              connOpts <- mkConnectionOpts AmqpEndpoint {..} connName
               liftIO $ Q.openConnection'' connOpts
           )
   bracket getConn (liftIO . Q.closeConnection) k
 
-mkConnectionOpts :: (MonadIO m) => AmqpEndpoint -> m Q.ConnectionOpts
-mkConnectionOpts AmqpEndpoint {..} = do
+mkConnectionOpts :: (MonadIO m) => AmqpEndpoint -> Maybe Text -> m Q.ConnectionOpts
+mkConnectionOpts AmqpEndpoint {..} name = do
   mTlsSettings <- traverse (liftIO . (mkTLSSettings host)) tls
   (username, password) <- liftIO $ readCredsFromEnv
   pure
@@ -194,7 +195,8 @@ mkConnectionOpts AmqpEndpoint {..} = do
       { Q.coServers = [(host, fromIntegral port)],
         Q.coVHost = vHost,
         Q.coAuth = [Q.plain username password],
-        Q.coTLSSettings = fmap Q.TLSCustom mTlsSettings
+        Q.coTLSSettings = fmap Q.TLSCustom mTlsSettings,
+        Q.coName = name
       }
 
 -- | Connects with RabbitMQ and opens a channel. If the channel is closed for
@@ -205,9 +207,10 @@ openConnectionWithRetries ::
   (MonadIO m, MonadMask m, MonadBaseControl IO m) =>
   Logger ->
   AmqpEndpoint ->
+  Maybe Text ->
   RabbitMqHooks m ->
   m ()
-openConnectionWithRetries l AmqpEndpoint {..} hooks = do
+openConnectionWithRetries l AmqpEndpoint {..} connName hooks = do
   (username, password) <- liftIO $ readCredsFromEnv
   connectWithRetries username password
   where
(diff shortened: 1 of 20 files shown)


libs/extended/src/Network/AMQP/Extended.hs
```diff
@@ -180,21 +181,22 @@ withConnection l AmqpEndpoint {..} k = do
...
               Log.info l $ Log.msg (Log.val "Trying to connect to RabbitMQ")
-              connOpts <- mkConnectionOpts AmqpEndpoint {..}
+              connOpts <- mkConnectionOpts AmqpEndpoint {..} connName
               liftIO $ Q.openConnection'' connOpts
           )
   bracket getConn (liftIO . Q.closeConnection) k
 
-mkConnectionOpts :: (MonadIO m) => AmqpEndpoint -> m Q.ConnectionOpts
-mkConnectionOpts AmqpEndpoint {..} = do
+mkConnectionOpts :: (MonadIO m) => AmqpEndpoint -> Maybe Text -> m Q.ConnectionOpts
```
supersven, line 189: Is there a good reason to not force users of this function to provide a connection name? IMHO it's really helpful for keeping an overview of the connections in the RabbitMQ admin gui. I may be overlooking something, though. :thinking:

akshaymankar: I just propagated the `Maybe`, I think there are no places where we send a `Nothing`, but I thought its best to let the calling place make this decision.

services/background-worker/src/Wire/BackgroundWorker/Env.hs
```diff
@@ -29,8 +30,12 @@ type IsWorking = Bool
 -- | Eventually this will be a sum type of all the types of workers
 data Worker
   = BackendNotificationPusher
-  | BackendDeadUserNoticationWatcher
-  deriving (Show, Eq, Ord)
+  | DeadUserNotificationWatcher
+  deriving (Eq, Ord)
+
+instance Show Worker where
```
supersven: The drawback of overwriting `show` like this, is that one cannot `read` the result later. That's sometimes pretty useful (especially when things get nested.)
However, I see that this probably has no big impact here. Just wanted to mention, that I would likely have invented e.g. `workerAsMetricsString :: Worker -> String` and derived the `Show` instance.

services/background-worker/src/Wire/BackgroundWorker/Env.hs
```diff
@@ -127,10 +138,20 @@ instance (MonadIO m) => MonadLogger (AppT m) where
...
+markAsWorking worker = do
+  env <- ask
+  modifyIORef env.statuses $ Map.insert worker True
+  withLabel env.workerRunningGauge (Text.pack $ show worker) (flip setGauge 1)
+
+markAsNotWorking :: (MonadIO m, MonadMonitor m) => Worker -> AppT m ()
+markAsNotWorking worker = do
+  env <- ask
+  modifyIORef env.statuses (Map.insert worker False)
+  withLabel env.workerRunningGauge (Text.pack $ show worker) (flip setGauge 1)
```
supersven:
```suggestion
  withLabel env.workerRunningGauge (Text.pack $ show worker) (flip setGauge 0)
```

Is the `1` correct? (I would have expected `0`. I.e. to be the "opposite" of what we do for `markAsWorking`.)

services/background-worker/src/Wire/BackgroundWorker/Env.hs
```diff
@@ -127,10 +138,20 @@ instance (MonadIO m) => MonadLogger (AppT m) where
...
-  flip modifyIORef (Map.insert worker True) =<< asks statuses
-
-markAsNotWorking :: (MonadIO m) => Worker -> AppT m ()
-markAsNotWorking worker =
-  flip modifyIORef (Map.insert worker False) =<< asks statuses
+markAsWorking :: (MonadIO m, MonadMonitor m) => Worker -> AppT m ()
+markAsWorking worker = do
+  env <- ask
+  modifyIORef env.statuses $ Map.insert worker True
+  withLabel env.workerRunningGauge (Text.pack $ show worker) (flip setGauge 1)
```
supersven: Not a big deal, but we could also consider to invent `markAs :: (...) => Bool -> Double -> AppT m ()` to reduce duplication.

services/background-worker/test/Test/Wire/BackendNotificationPusherSpec.hs
```diff
@@ -331,6 +331,7 @@ spec = do
           backendNotificationsConfig = BackendNotificationsConfig 1000 500000 1000
 
       backendNotificationMetrics <- mkBackendNotificationMetrics
+      workerRunningGauge <- mkWorkerRunningGauge
```
supersven, line 334: Some assertion / test to check the current state of the gauge would be nice.

services/background-worker/src/Wire/DeadUserNotificationWatcher.hs
```diff
@@ -87,57 +85,32 @@ markAsNeedsFullSync cassandra uid cid = do
...
-  async (openConnection Nothing)
+  cleanupRef <- newIORef Nothing
+  markAsNotWorking DeadUserNotificationWatcher
+  -- We can fire and forget this thread because it keeps respawning itself using the 'onConnectionClosedHandler'.
+  void . async . openConnectionWithRetries env.logger amqp (Just "dead-user-notifications-watcher") $
+    RabbitMqHooks
+      { onNewChannel = \chan -> do
+          consumerTag <- startConsumer chan
+          atomicWriteIORef cleanupRef (Just (chan, consumerTag))
+          forever $ threadDelay maxBound,
```
supersven, line 102: Is waiting forever intended?

services/background-worker/src/Wire/BackgroundWorker/Env.hs
akshaymankar: Create a new function.

services/background-worker/src/Wire/BackgroundWorker/Env.hs
akshaymankar: Good catch, was copy-pasta.

services/background-worker/src/Wire/BackgroundWorker/Env.hs
akshaymankar: Done.

services/background-worker/src/Wire/DeadUserNotificationWatcher.hs
akshaymankar: Yes, added a comment.

services/background-worker/test/Test/Wire/BackendNotificationPusherSpec.hs
akshaymankar: This is tough to check because making this metric go to 0 means causing RabbitMQ outage for long enough that we can detect this change in metrics. I don't think we can do that easily.

supersven: Okay, testing the metrics is probably not crucial.

libs/extended/src/Network/AMQP/Extended.hs
supersven: As long as the prod consumers have names: Fine by me :smile:

PR #4656, opened 2025-07-10
UserSubsystem: Deduplicate user search results on exact handle match

akshaymankar:
https://wearezeta.atlassian.net/browse/WPB-17867

## Checklist

### DIFF
--- a/changelog.d/3-bug-fixes/dedup-search-results
+++ b/changelog.d/3-bug-fixes/dedup-search-results
@@ -1,0 +1,1 @@
+Avoid returning duplicate user search results when handle matches exactly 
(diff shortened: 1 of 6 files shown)


battermann (approved):
The fix makes sense, and the test too. What I cannot assess is if the mock interpreter for the user index behaves the same as the real user index. I trust your expertise here.

Personally, I would have reproduced the issue with an integration test and then fix it. But the unit test has a much better resource footprint of course, that is nice.

akshaymankar:
> The fix makes sense, and the test too. What I cannot assess is if the mock interpreter for the user index behaves the same as the real user index. I trust your expertise here.
> 
> Personally, I would have reproduced the issue with an integration test and then fix it. But the unit test has a much better resource footprint of course, that is nice.

I don't think it would behave exactly like elasticsearch, but I tried to encode the important parts of our query to ES. The way ES actually orders things is way more complicated and probably not worth replicating because from Wire's perspective the order of results doesn't matter so much.

PR #4657, opened 2025-07-10
Makefile: Login to docker hub before setting up integration tests

akshaymankar:
Only logs in if the credentials are provided.

https://wearezeta.atlassian.net/browse/WPB-18677

## Checklist

 - [x] ~Add a new entry in an appropriate subdirectory of `changelog.d` No changelog.


hack/bin/helm-oci-login.sh
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+if [[ -z "${DOCKER_HUB_USERNAME+x}" ]]; then
+    echo "Not logging in to docker hub as there are no credentials provided."
+else
+    helm registry login registry-1.docker.io --username "${DOCKER_HUB_USERNAME}" --password "${DOCKER_HUB_PASSWORD}"
```
supersven, line 8: The `registry-1` domain part looks a bit suspicious; like some temporary thing. Are you sure it shouldn't be e.g. `registry.hub.docker.com`?

hack/bin/helm-oci-login.sh
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+if [[ -z "${DOCKER_HUB_USERNAME+x}" ]]; then
```
supersven, line 5: I'm wondering if there's a benefit in expanding `$DOCKER_HUB_USERNAME` to `x` :thinking: 

However, it doesn't hurt.

akshaymankar: For bash reasons: https://stackoverflow.com/questions/3601515/how-to-check-if-a-variable-is-set-in-bash

hack/bin/helm-oci-login.sh
akshaymankar: This is the registry URL in all the bitnami index.

PR #4663, opened 2025-07-15
replace nginz disco with native upstream resolver WPB-15302

jschaul:
WPB-15302

- replace outside upstreams.txt file by an inline block, making use of 'resolve' keyword (available since nginx 1.27.3). Current nix packages pins nginx to 1.28.
```
nix repl --file nix/default.nix
nix-repl> pkgs.nginx.version
"1.28.0"
```
(but actually hardcoded nginx to 1.26 for some reason! - So this PR **also upgrades nginx from 1.26 to 1.28**)
- remove nginz_disco code, and sidecar deployment
- drive-by improvement: be more graceful when shutting down: longer graceperiod, and use quit not sigterm (to process in-flight requests before shutting down - hopefully less 502s that way)

## Checklist

### DIFF
--- a/integration/test/Testlib/ModService.hs
+++ b/integration/test/Testlib/ModService.hs
@@ -631,7 +631,7 @@ replaceUpstreamsInConfig nginxConf sm =
     -- }
     --
     -- Prerequisite is that the upstream  block itself does not contain block
-    -- delimitiers. AFAIK this usually holds.
+    -- delimiters. AFAIK this usually holds.
     removeUpstreamBlocks :: Text.Text
     removeUpstreamBlocks =
       replaceAll "" $


charts/nginz/templates/conf/_nginx.conf.tpl
```diff
@@ -189,8 +189,22 @@ http {
...
-  include {{ .Values.nginx_conf.upstream_config }};
+  resolver kube-dns.kube-system.svc.{{ $clusterDomain }} valid=30s ipv6=off;
+  resolver_timeout 5s;
+
+{{- $validUpstreams := include "valid_upstreams" . | fromJson }}
+{{- range $name, $_ := $validUpstreams }}
+  upstream {{ $name }} {
+      zone {{ $name }} 64k; # needed for dynamic DNS updates
+      least_conn;
+      keepalive 32;
```
smatting, line 202: where do these settings come from?

siberijah (approved):

charts/nginz/templates/conf/_nginx.conf.tpl
jschaul: They are kept from before, see https://github.com/wireapp/wire-server/blob/develop/tools/nginz_disco/nginz_disco.sh#L49-L55 which results in current upstream files looking like

```
upstream cargohold { 
	 least_conn; 
	 keepalive 32; 
	 server 10.100.213.88:8080 max_fails=3 weight=100; 
}
upstream galeb.integrations { 
	 least_conn; 
	 keepalive 32; 
	 server 10.100.18.166:8080 max_fails=3 weight=100; 
}
```

The only difference here is the addition of a `zone`, which according to the docs in https://nginx.org/en/docs/http/ngx_http_upstream_module.html (see 'resolve' keyword) is needed, as 
> The server group must reside in the [shared memory](https://nginx.org/en/docs/http/ngx_http_upstream_module.html#zone).

nix/overlay.nix
```diff
@@ -73,10 +73,10 @@ self: super: {
   };
 
   nginz = (super.nginx.overrideAttrs rec {
-    version = "1.26.0";
+    version = "1.28.0";
     src = super.fetchurl {
       url = "https://nginx.org/download/nginx-${version}.tar.gz";
-      hash = "sha256-0ubIQ51sbbUBXY6qskcKtSrvhae/NjGCh5l34IQ3BJc=";
+      hash = "sha256-xrXGsIbA3508o/9eCEwdDvkJ5gOCecccHD6YX1dv92o=";
```
supersven: Do we still need to override these attributes? `1.28.0` seems to be exactly what `nixpkgs` provides: https://github.com/NixOS/nixpkgs/blob/b3582c75c7f21ce0b429898980eddbbf05c68e55/pkgs/servers/http/nginx/stable.nix#L4

jschaul: You are right. Removed the override, commit pushed.

supersven (approved):

integration/test/Testlib/ModService.hs
```diff
@@ -605,34 +594,82 @@ server 127.0.0.1:{port} max_fails=3 weight=1;
...
+    -- remove blocks like:
+    --
+    -- upstream <name> {
+    --   ...
+    --   <config>
+    --   ...
+    -- }
+    --
+    -- Prerequisite is that the upstream  block itself does not contain block
+    -- delimitiers. AFAIK this usually holds.
```
pcapriotti:
```suggestion
    -- delimiters. AFAIK this usually holds.
```

pcapriotti (approved):
LGTM

PR #4668, opened 2025-07-17
integration-setup: Add explicit helm login to docker hub 

akshaymankar:
This is required because we have to run `helm dep update` manually before
running `helmfile` even if we run `helmfile` without `--skip-sync` because `helm
dep update` needs to happen in a certain order to make sure all nested local
dependencies are packaged correctly.

It was removed in #4664 because it was wrongly considered unnecessary. 

https://wearezeta.atlassian.net/browse/WPB-18677

## Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -444,7 +444,7 @@ libzauth:
 kube-integration:  kube-integration-setup kube-integration-test
 
 .PHONY: kube-integration-setup
-kube-integration-setup: charts-integration
+kube-integration-setup: charts-integration helm-oci-login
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -463,6 +463,9 @@ kube-integration-teardown:
 kube-integration-e2e-telepresence:
 	./services/brig/federation-tests.sh $(NAMESPACE)
 
+helm-oci-login:
```
supersven, line 467: Shouldn't this target be `.PHONY` as well? :thinking:

supersven (approved):
LGTM :+1:

PR #4670, opened 2025-07-17
Check consistency of group info

pcapriotti:
This PR adds an optional consistency check of the submitted group info on any commit. The ratchet tree extension is parsed and compared with the index map of the group after the commit is applied.

https://wearezeta.atlassian.net/browse/WPB-18394

## Checklist

### DIFF
--- a/services/galley/src/Galley/API/MLS/Proposal.hs
+++ b/services/galley/src/Galley/API/MLS/Proposal.hs
@@ -29,7 +29,6 @@ module Galley.API.MLS.Proposal
     -- * Proposal actions
     paAddClient,
     paRemoveClient,
-    paApply,
 
     -- * Types
     ProposalAction (..),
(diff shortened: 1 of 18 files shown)


eyeinsky:
The CI failed for yesterday's changes, a few of these I think were unrelated to what I did (fake-aws-sqs timed out, upload-hoogle-nix failed), and with one I broke the linter -- the linter thing I fixed and now waiting on CI.

services/galley/src/Galley/API/MLS/Proposal.hs
```diff
@@ -1,3 +1,5 @@
+{-# OPTIONS -Wwarn #-}
```
stefanwire: Was this added intentionally?

stefanwire (approved):

services/galley/src/Galley/API/MLS/Proposal.hs
pcapriotti: Nope, thanks for catching that :)

PR #4673, opened 2025-07-21
WPB-18562 Propagate a rate limit error with status 429 on internal call to reauthenticate

battermann:
## Checklist

### DIFF
--- a/integration/test/Test/Teams.hs
+++ b/integration/test/Test/Teams.hs
@@ -21,7 +21,7 @@ module Test.Teams where
 import API.Brig
 import qualified API.BrigInternal as I
 import API.Common
-import API.Galley (getTeam, getTeamMembers, getTeamMembersCsv, getTeamNotifications)
+import API.Galley (deleteTeamMember, getTeam, getTeamMembers, getTeamMembersCsv, getTeamNotifications)
 import qualified API.GalleyInternal as I
 import API.Gundeck
 import qualified API.Nginz as Nginz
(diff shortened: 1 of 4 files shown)


integration/test/Test/Teams.hs
```diff
@@ -464,3 +465,12 @@ testUpgradeGuestToTeamShouldFail = do
...
     resp.status `shouldMatchInt` 404
+
+testDeleteTeamUserRatelimitingIsPropagated :: (HasCallStack) => App ()
+testDeleteTeamUserRatelimitingIsPropagated = do
+  (owner, tid, mems) <- createTeam OwnDomain 10
+  -- this is eventually going to run into rate limiting of internal request `/i/users/:uid/reauthenticate`
+  statusCodes <- for mems $ \m -> do
+    bindResponse (deleteTeamMember tid owner m) $ \resp -> do
+      pure resp.status
+  Set.fromList statusCodes `shouldMatchSet` ([202, 429] :: [Int])
```
stefanwire, line 476: Should we make sure that 429 was returned at least once?

stefanwire (approved):

integration/test/Test/Teams.hs
battermann: This actually asserts that it's returned at least once. But I might not 100% understand your comment?

stefanwire: True, I misread. ✅

PR #4685, opened 2025-07-25
WPB-18190: Add route to delete collaborator from team

eyeinsky:
Create draft PR with how far I got.

## Checklist

 - [ ] Add a new entry in an appropriate subdirectory of `changelog.d`
 - [ ] Read and follow the [PR guidelines](https://docs.wire.com/latest/developer/developer/pr-guidelines.html)

PR #4694, opened 2025-07-30
WPB-18190: Add route to delete collaborator from team

blackheaven:
https://wearezeta.atlassian.net/browse/WPB-18190

## Checklist

PS: I did not see #4685, sorry


libs/wire-subsystems/src/Wire/TeamCollaboratorsSubsystem/Interpreter.hs
```diff
@@ -93,6 +94,43 @@ getAllTeamCollaboratorsImpl zUser team = do
...
+    Member NotificationSubsystem r
+  ) =>
+  Local UserId ->
+  UserId ->
+  TeamId ->
+  Sem r ()
+removeTeamCollaboratorImpl zUser user team = do
+  guardPermission (tUnqualified zUser) team TeamMember.RemoveTeamCollaborator InsufficientRights
+  Store.removeTeamCollaborator user team
+  -- TODO gdf remove O2O conversations
```
blackheaven: I'm not sure how to handle this one.

I have tried to move `services/galley/src/Galley/Effects/ConversationStore.hs` in `lib/wire-subsystems`, but Cassandra queries and many data-types are in Galley.

I'm also not sure how to list the O2O conversations of a `(TeamId, UserId)`.

I would be inclined to create a dedicate effects, with one operation and a Cassandra interpreter, having queries in the same file.

WDYT?

libs/wire-subsystems/src/Wire/ConversationsStore/Cassandra.hs
```diff
@@ -0,0 +1,20 @@
...
+interpretConversationsStoreCassandra casClient =
+  interpret $
+    runEmbedded (runClient casClient) . embed . \case
+      CloseConversationsFrom tid uid -> closeConversationsFromImpl tid uid
+
+closeConversationsFromImpl :: TeamId -> UserId -> Client ()
+closeConversationsFromImpl t u = retry x5 $ write conversationsUpdate (params LocalQuorum (t, Just u))
+  where
+    conversationsUpdate :: PrepQuery W (TeamId, Maybe UserId) ()
+    conversationsUpdate = "UPDATE conversation SET deleted = true WHERE team = ? AND author = ?"
```
blackheaven: This one is failing with:

```
[brig@example.com] E, IO Exception occurred, message=ResponseError {reHost = datacenter1:rack1:127.0.0.1:9042, reTrace = Nothing, reWarn = [], reCause = Invalid "unconfigured table conversation"}, request=abef79a6-8f9f-41b3-b674-036e4216db63
[brig@example.com] E, request=abef79a6-8f9f-41b3-b674-036e4216db63, code=500, label=server-error, "Server Error"
```

I feel like I'm in another key space, can I have some hints about it?

services/galley/src/Galley/ConversationsSubsystem.hs
```diff
@@ -0,0 +1,53 @@
...
+      page <- paginate conversationsFilter (params LocalQuorum (Identity convIds))
+      InternalPage (_, _, filteredConvIds) <- mkInternalPage page pure
+      pure $
+        map (\(convId, _team, _mCreator) -> convId) $
+          filter (\(_convId, team, mCreator) -> team == Just tid && mCreator == Just uid) filteredConvIds
+    conversationsFilter :: PrepQuery R (Identity [ConvId]) (ConvId, Maybe TeamId, Maybe UserId)
+    conversationsFilter = "select conv, team, creator from conversation where conv in ?"
+    performUpdate convIds = retry x5 $ write conversationsUpdate (params LocalQuorum (Identity convIds))
+    conversationsUpdate :: PrepQuery W (Identity [ConvId]) ()
+    conversationsUpdate = "update conversation set deleted = true where conv in ?"
```
pcapriotti: I don't think this is enough. This would leave the conversation in the database, as well as member entries. It's probably better to reuse the conversation store here, which already has a `deleteConversation` method.

services/galley/src/Galley/ConversationsSubsystem.hs
```diff
@@ -0,0 +1,53 @@
...
+  where
+    listConversationsIds pagingState =
+      fmap runIdentity <$> paginateWithState Cql.selectUserConvs (paramsPagingState LocalQuorum (Identity uid) 32 pagingState)
+    performFilter :: [ConvId] -> Client [ConvId]
+    performFilter convIds = do
+      page <- paginate conversationsFilter (params LocalQuorum (Identity convIds))
+      InternalPage (_, _, filteredConvIds) <- mkInternalPage page pure
+      pure $
+        map (\(convId, _team, _mCreator) -> convId) $
+          filter (\(_convId, team, mCreator) -> team == Just tid && mCreator == Just uid) filteredConvIds
```
pcapriotti: This doesn't look right to me, but maybe I'm not understanding something. In my understanding, we should:
* delete all one2one conversations with **unconnected** team members
* remove ourselves from team conversations.

The code here seems to only look for team conversations that were created by the collaborator, and then delete them. I can't see anything about one2one conversations. Regardless, I don't think it matters who created the conversation. When a collaboration is ended, the collaborator should simply be kicked from team conversations.

pcapriotti (commented):
Looks good overall, but I'm not sure about the logic for deleting conversations. See comments below.

services/galley/src/Galley/ConversationsSubsystem.hs
blackheaven: I have `deleteConversation` for `One2One` conversations type, and remove user-conversation relationship otherwise, let me know what you think.

blackheaven:
I cannot reproduce the error, which seems not related to my changes.

changelog.d/2-features/WPB-18190
```diff
@@ -0,0 +1 @@
+Allow member to be removed from a team.
```
battermann:
```suggestion
Allow collaborator to be removed from a team.
```

integration/test/Test/TeamCollaborators.hs
```diff
@@ -143,3 +143,19 @@ testImplicitConnectionNoCollaborator = do
     >>= assertSuccess
 
   postOne2OneConversation bob alice team0 "chit-chat" >>= assertLabel 403 "no-team-member"
+
+testRemoveMember :: (HasCallStack) => App ()
```
battermann: I think we should create a few team and 1:1 conversations and verify that the user is removed.

We should also verify that the user still exists and is still member of other non-team conersations etc.

blackheaven:
@battermann I have made the change to rely on the same mechanisms behind `rmUser`, feel free to have a look.

Thanks.

libs/wire-subsystems/src/Wire/TeamCollaboratorsSubsystem.hs
```diff
@@ -13,5 +13,6 @@ data TeamCollaboratorsSubsystem m a where
   GetAllTeamCollaborators :: Local UserId -> TeamId -> TeamCollaboratorsSubsystem m [TeamCollaborator]
   InternalGetTeamCollaborator :: TeamId -> UserId -> TeamCollaboratorsSubsystem m (Maybe TeamCollaborator)
   InternalGetTeamCollaborations :: UserId -> TeamCollaboratorsSubsystem m [TeamCollaborator]
+  InternalRemoveTeamCollaborator :: UserId -> TeamId -> TeamCollaboratorsSubsystem m ()
```
battermann: I am not sure about the naming convention here. Why the prefix `Internal`?

blackheaven: Mainly because we do not check permissions in it, should I drop it?

integration/test/Test/TeamCollaborators.hs
blackheaven: Sorry, I have missed it:

> yes, I would create a couple of conversations, team group conv, 1:1 conv, non-team conv, and then check membership after the collaborator is removed

blackheaven: TODO: find how to create team and non-team convs

blackheaven: @battermann I did not managed to create non-team conversation becoming one, but the remaining edge cases are checked let me know what you think.

services/galley/src/Galley/ConversationsSubsystem.hs
```diff
@@ -0,0 +1,58 @@
...
+      let extractConv = map (\(convId, _team, _convType) -> convId)
+      pure $
+        bimap extractConv extractConv $
+          partition (\(_convId, _team, convType) -> convType == One2OneConv) $
+            filter (\(_convId, team, _convType) -> team == Just tid) filteredConvIds
+    conversationsFilter :: PrepQuery R (Identity [ConvId]) (ConvId, Maybe TeamId, ConvType)
+    conversationsFilter = "select conv, team, type from conversation where conv in ?"
+    performConversationsRemoveUser convIds = retry x5 $ write conversationsRemoveUser (params LocalQuorum (uid, convIds))
+    conversationsRemoveUser :: PrepQuery W (UserId, [ConvId]) ()
+    conversationsRemoveUser = "delete from user where user = ? and conv in ?"
```
pcapriotti: Same issue as in https://github.com/wireapp/wire-server/pull/4708: we shouldn't use `IN` queries.

services/galley/src/Galley/ConversationsSubsystem.hs
```diff
@@ -0,0 +1,58 @@
...
+    performFilter convIds = do
+      page <- paginate conversationsFilter (params LocalQuorum (Identity convIds))
+      InternalPage (_, _, filteredConvIds) <- mkInternalPage page pure
+      let extractConv = map (\(convId, _team, _convType) -> convId)
+      pure $
+        bimap extractConv extractConv $
+          partition (\(_convId, _team, convType) -> convType == One2OneConv) $
+            filter (\(_convId, team, _convType) -> team == Just tid) filteredConvIds
+    conversationsFilter :: PrepQuery R (Identity [ConvId]) (ConvId, Maybe TeamId, ConvType)
+    conversationsFilter = "select conv, team, type from conversation where conv in ?"
```
pcapriotti: Same here. Instead, I think we should use the existing pagination internal API, like we do when deleting a user.

services/galley/src/Galley/ConversationsSubsystem.hs
blackheaven: dropped

services/galley/src/Galley/ConversationsSubsystem.hs
blackheaven: dropped, however, I'm not sur what you are referring to, there is nothing more in `rmUser`.

libs/wire-subsystems/src/Wire/TeamCollaboratorsSubsystem.hs
battermann: Ah got it. It's fine then.

services/brig/src/Brig/Team/API.hs
```diff
@@ -114,6 +114,7 @@ servantAPI =
     :<|> Named @"get-team-size" (\uid tid -> lift . liftSem $ teamSizePublic uid tid)
     :<|> Named @"accept-team-invitation" (\luid req -> lift $ liftSem $ acceptTeamInvitation luid req.password req.code)
     :<|> Named @"add-team-collaborator" (\zuid tid (NewTeamCollaborator uid perms) -> lift . liftSem $ createTeamCollaborator zuid uid tid perms)
+    :<|> Named @"remove-team-collaborator" (\zuid tid uid -> lift . liftSem $ GalleyAPIAccess.removeTeamMember zuid uid tid)
```
battermann: We should move this handler to galley, so that we do not need to make an extra unnecessary call from brig to galley.

blackheaven:
@battermann I have reworked the the design, I still have to rework the tests

libs/wire-api/src/Wire/API/Routes/Public/Galley/TeamMember.hs
```diff
@@ -207,6 +207,17 @@ type TeamMemberAPI =
                     "CSV of team members"
                     CSV
            )
+    :<|> Named
+           "remove-team-collaborator"
+           ( Summary "Remove a collaborator from the team."
+               :> From 'V11
```
battermann: should be V12 by now

libs/wire-api/src/Wire/API/Team/Member.hs
```diff
@@ -481,6 +481,7 @@ data HiddenPerm
   | NewTeamCollaborator
   | JoinRegularConversations
   | CreateApp
+  | RemoveTeamCollaborator
```
battermann, line 485: Is this used, I can't find it

integration/test/Test/TeamCollaborators.hs
```diff
@@ -156,4 +157,106 @@ testImplicitConnectionNoCollaborator = do
...
+testRemoveMemberInTeamConversation = do
+  (owner, team, [alice, bob]) <- createTeam OwnDomain 3
+
+  conv <-
+    postConversation
+      owner
+      defProteus {team = Just team, qualifiedUsers = [alice, bob]}
+      >>= getJSON 201
+
+  withWebSockets [owner, alice] $ \[wsOwner, wsAlice] -> do
```
battermann: should the collaborator themselves, also get an event?

services/galley/src/Galley/API/Teams.hs
```diff
@@ -1295,3 +1302,58 @@ checkAdminLimit :: (Member (ErrorS 'TooManyTeamAdmins) r) => Int -> Sem r ()
...
+  toNotify <-
+    handleImpossibleErrors $
+      getFeatureForTeam @LimitedEventFanoutConfig tid
+        >>= ( \case
+                FeatureStatusEnabled -> Left <$> E.getTeamAdmins tid
+                FeatureStatusDisabled -> Right <$> getTeamMembersForFanout tid
+            )
+          . (.status)
+  uncheckedDeleteTeamMember lusr Nothing tid rusr toNotify
+  internalRemoveTeamCollaborator rusr tid
```
battermann, line 1343: should we send the collaborator remove event here?

libs/wire-api/src/Wire/API/Event/Team.hs
```diff
@@ -159,6 +161,7 @@ data EventData
   | EdConvDelete ConvId
   | EdCollaboratorAdd UserId [CollaboratorPermission]
   | EdAppCreate UserId
+  | EdCollaboratorRemove UserId
```
battermann, line 164: this is not sent, is it?

services/galley/src/Galley/API/Teams.hs
```diff
@@ -929,27 +931,32 @@ removeFromConvsAndPushConvLeaveEvent lusr zcon tid remove = do
...
-              True
-              zcon
-              (qualifyAs lusr dc)
-              targets
-              ( ConversationRemoveMembers
-                  (pure . tUntagged . qualifyAs lusr $ remove)
-                  EdReasonDeleted
-              )
-              def
+        case () of
```
battermann: this is a funny construct, why not match on the `cnvmType` value directly?

libs/wire-api/src/Wire/API/Team/Member.hs
blackheaven: It is in `removeTeamCollaborator`, the handler.

libs/wire-api/src/Wire/API/Routes/Public/Galley/TeamMember.hs
blackheaven: updated, thanks.

services/galley/src/Galley/API/Teams.hs
blackheaven: added, thanks.

libs/wire-api/src/Wire/API/Event/Team.hs
blackheaven: added, thanks.

services/galley/src/Galley/API/Teams.hs
blackheaven: done, thanks

integration/test/Test/TeamCollaborators.hs
blackheaven: added

blackheaven:
>  error: interrupted by the user

@battermann I have tackled your comments, can you have another look please?

services/galley/src/Galley/API/Teams.hs
```diff
@@ -931,27 +933,31 @@ removeFromConvsAndPushConvLeaveEvent lusr zcon tid remove = do
...
-              (qualifyAs lusr dc)
-              targets
-              ( ConversationRemoveMembers
-                  (pure . tUntagged . qualifyAs lusr $ remove)
-                  EdReasonDeleted
-              )
-              def
+        case dc.metadata.cnvmType of
+          One2OneConv ->
+            E.deleteConversation dc.id_
```
battermann: I think this case needs to go into the `when (remove `isMember` dc.localMembers)` case below. Otherwise, all 1:1 conversations will be deleted.

services/galley/src/Galley/API/Teams.hs
```diff
@@ -1297,3 +1303,67 @@ checkAdminLimit :: (Member (ErrorS 'TooManyTeamAdmins) r) => Int -> Sem r ()
...
+  Local UserId ->
+  TeamId ->
+  UserId ->
+  Sem r ()
+removeTeamCollaborator lusr tid rusr = do
+  P.debug $
+    Log.field "targets" (toByteString rusr)
+      . Log.field "action" (Log.val "Teams.removeTeamCollaborator")
+  zusrMember <- E.getTeamMember tid (tUnqualified lusr)
+  void $ permissionCheck RemoveTeamMember zusrMember
```
battermann: shouldn't we use the new permission `RemoveTeamCollaborator` here?

services/galley/src/Galley/API/Teams.hs
```diff
@@ -1297,3 +1303,67 @@ checkAdminLimit :: (Member (ErrorS 'TooManyTeamAdmins) r) => Int -> Sem r ()
...
+          . (.status)
+  uncheckedDeleteTeamMember lusr Nothing tid rusr toNotify
+  internalRemoveTeamCollaborator rusr tid
+  now <- Now.get
+  let e = newEvent tid now (EdCollaboratorRemove rusr)
+  pushNotifications
+    [ def
+        { origin = Just $ tUnqualified lusr,
+          json = toJSONObject e,
+          recipients = [userRecipient $ tUnqualified lusr]
```
battermann: Here only the user that did the removal gets the notification. We should assess which other users need this. I can imagine other team owners and admins as well as the collaborator themselves.

battermann: Also there should be a test for those notifications.

services/galley/src/Galley/API/Teams.hs
blackheaven: Nice catch thanks, fixed

services/galley/src/Galley/API/Teams.hs
blackheaven: Nice catch thanks, changed

services/galley/src/Galley/API/Teams.hs
blackheaven: I have extended it to all the members of the team and added in the test.

I'm not sure to understand why the others collaborators should be notified, and not the regular members, can you elaborate on that please? Here on in a call.

battermann: There might be a scalability issue if we fan this out to all team members. And I also don't think it is required.
I am also not sure if admins and owners need this event as there should also be the member.leave event if I am not mistaken.
My main point was to suggest doing some research and figure out what the requirements for notifications are in this case. I can imagine that at least the collaborator that got removed should be informed.

libs/wire-api/src/Wire/API/Routes/Public/Galley/TeamMember.hs
```diff
@@ -207,6 +207,17 @@ type TeamMemberAPI =
...
+           "remove-team-collaborator"
+           ( Summary "Remove a collaborator from the team."
+               :> From 'V12
+               :> ZLocalUser
+               :> "teams"
+               :> Capture "tid" TeamId
+               :> "collaborators"
+               :> Capture "uid" UserId
+               :> MultiVerb1 'DELETE '[JSON] (RespondEmpty 200 "")
+           )
```
battermann, line 222: We should add the necessary `CanThrow` combinators to this endpoint so that the correct errors are shown in the swagger docs.

integration/test/Notifications.hs
```diff
@@ -171,66 +171,69 @@ isConvAccessUpdateNotif :: (HasCallStack, MakesValue n) => n -> App Bool
...
+isTeamMemberLeaveNotif :: (HasCallStack, MakesValue a) => a -> App Bool
 isTeamMemberLeaveNotif = notifTypeIsEqual "team.member-leave"
 
-isTeamCollaboratorAddedNotif :: (MakesValue a) => a -> App Bool
+isTeamCollaboratorAddedNotif :: (HasCallStack, MakesValue a) => a -> App Bool
 isTeamCollaboratorAddedNotif = notifTypeIsEqual "team.collaborator-add"
 
-isUserActivateNotif :: (MakesValue a) => a -> App Bool
+isTeamCollaboratorRemovededNotif :: (HasCallStack, MakesValue a) => a -> App Bool
+isTeamCollaboratorRemovededNotif = notifTypeIsEqual "team.collaborator-remove"
```
battermann: typo

services/galley/src/Galley/API/Teams.hs
```diff
@@ -1297,3 +1304,68 @@ checkAdminLimit :: (Member (ErrorS 'TooManyTeamAdmins) r) => Int -> Sem r ()
...
+  UserId ->
+  Sem r ()
+removeTeamCollaborator lusr tid rusr = do
+  P.debug $
+    Log.field "targets" (toByteString rusr)
+      . Log.field "action" (Log.val "Teams.removeTeamCollaborator")
+  zusrMember <- E.getTeamMember tid (tUnqualified lusr)
+  void $ permissionCheck RemoveTeamCollaborator zusrMember
+  toNotify <-
+    handleImpossibleErrors $
```
battermann:
```suggestion
```
I don't think this is necessary

services/galley/src/Galley/API/Teams.hs
```diff
@@ -1297,3 +1304,68 @@ checkAdminLimit :: (Member (ErrorS 'TooManyTeamAdmins) r) => Int -> Sem r ()
...
+    handleImpossibleErrors ::
+      Sem
+        ( ErrorS 'NotATeamMember
+            ': ErrorS 'TeamNotFound
+            ': r
+        )
+        a ->
+      Sem r a
+    handleImpossibleErrors action =
+      mapToDynamicError @'TeamNotFound (mapToDynamicError @'NotATeamMember action)
```
battermann: I don't think this is necessary. AFAICT the code this is applied to does not throw these errors. So it is redundant. Also it would just remove the static errors from showing in the constraints. But they are already part of the constraints. The dyn errors could still (if they occur) bubble up as dynamic errors and would need to be handled by the client. So regardless of the other objections, I don't think it is a good solution to the problem. (if it was a problem)

services/galley/src/Galley/API/Teams.hs
```diff
@@ -1297,3 +1304,68 @@ checkAdminLimit :: (Member (ErrorS 'TooManyTeamAdmins) r) => Int -> Sem r ()
...
+  uncheckedDeleteTeamMember lusr Nothing tid rusr toNotify
+  internalRemoveTeamCollaborator rusr tid
+  now <- Now.get
+  let e = newEvent tid now (EdCollaboratorRemove rusr)
+  members <- E.getTeamMembers tid
+  pushNotifications
+    [ def
+        { origin = Just $ tUnqualified lusr,
+          json = toJSONObject e,
+          recipients = members <&> \m -> Recipient (m ^. userId) RecipientClientsAll
```
battermann: I commented on this already in another place. Maybe we can agree on sending the notification to the collaborator that got removed for now? And if it turns out that it is needed for other users/clients we can still add it later.

battermann (changes requested):

services/galley/src/Galley/API/Teams.hs
blackheaven: It makes more sense, thanks, fixed.

libs/wire-api/src/Wire/API/Event/Team.hs
```diff
@@ -235,6 +240,10 @@ parseEventData AppCreate Nothing = fail "missing event data for type 'team.app-c
 parseEventData AppCreate (Just j) = do
   let f o = EdAppCreate <$> o .: "user"
   withObject "app create data" f j
+parseEventData CollaboratorRemove Nothing = fail "missing event data for type 'team.collaborator-remove"
```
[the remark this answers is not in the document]
blackheaven: fixed, thanks non-human.

services/galley/src/Galley/API/Teams.hs
blackheaven: dropped, thanks.

integration/test/Notifications.hs
blackheaven: I had to read it 5 times, thanks, fixed.

libs/wire-api/src/Wire/API/Routes/Public/Galley/TeamMember.hs
blackheaven: nice, thanks, how do you represent `FederationError`?

services/galley/src/Galley/API/Teams.hs
blackheaven: dropped thanks

services/galley/src/Galley/API/Teams.hs
blackheaven: I agree, changed :)

libs/wire-api/src/Wire/API/Routes/Public/Galley/TeamMember.hs
```diff
@@ -207,6 +207,18 @@ type TeamMemberAPI =
...
+           ( Summary "Remove a collaborator from the team."
+               :> CanThrow 'NotATeamMember
+               :> From 'V12
+               :> ZLocalUser
+               :> "teams"
+               :> Capture "tid" TeamId
+               :> "collaborators"
+               :> Capture "uid" UserId
+               :> MultiVerb1 'DELETE '[JSON] (RespondEmpty 200 "")
+           )
```
battermann, line 222:
```suggestion
    :<|> Named
           "remove-team-collaborator"
           ( Summary "Remove a collaborator from the team."
               :> CanThrow OperationDenied
               :> CanThrow 'NotATeamMember
               :> From 'V12
               :> ZLocalUser
               :> "teams"
               :> Capture "tid" TeamId
               :> "collaborators"
               :> Capture "uid" UserId
               :> MultiVerb1 'DELETE '[JSON] (RespondEmpty 200 "")
           )
```

battermann (approved):
LGTM now, only I suggest a minor change, regarding the endpoint errors.

PR #4695, opened 2025-07-31
Move account creation to UserStore

pcapriotti:
https://wearezeta.atlassian.net/browse/WPB-18186

## Checklist


libs/wire-subsystems/test/unit/Wire/MockInterpreters/UserStore.hs
```diff
@@ -22,6 +22,7 @@ inMemoryUserStoreInterpreter ::
   (Member (State [StoredUser]) r) =>
   InterpreterFor UserStore r
 inMemoryUserStoreInterpreter = interpret $ \case
+  CreateUser new _ -> modify $ (newStoredUserToStoredUser new :)
```
blackheaven: I'm surprised there's no hlint warning about it

blackheaven (approved):

libs/wire-subsystems/test/unit/Wire/MockInterpreters/UserStore.hs
akshaymankar: We have some of them suppressed because sometimes extra brackets and dollars make the code easier to read.

pcapriotti: Fixed.

pcapriotti:
@akshaymankar and I looked at the question of why the invariant on the identity field (mentioned [here](https://github.com/wireapp/wire-server/pull/4695/files#diff-d954ccb3e413f8a2f69a6b55f29a9d095140b98c2f4c99f965fea3b65d45e6e2R108)) doesn't actually seem to be preserved.

The issue is that a user is always created in a non-activated state, then, if the email activation code was provided in the request, it is activated afterwards. The user record always has email_unvalidated set to null, and email identity set to the actual email, regardless of the activation state, and the returned user object reflects that. This is close to ok for user created with an activation code, but definitely wrong for non-activated users, since it does not match what the user looks like when queried separately.

The plan is to leave this PR as is, since it is a pure refactoring that doesn't change the current behaviour. In a separate PR, we plan to change the API as follows:
- get rid of the non-activated case of user creation (clients can request an email activation code separately, *then* register)
- always set email_unvalidated
Since email_unvalidated is a field in the returned JSON object, and was set to Nothing before, we need to restrict this change to the current version of the API in order to maintain compatibility.

PR #4699, opened 2025-08-01
WPB-19804 Email template validation

battermann:
https://wearezeta.atlassian.net/browse/WPB-19804

## Checklist

### DIFF
--- a/hack/helm_vars/wire-server/values.yaml.gotmpl
+++ b/hack/helm_vars/wire-server/values.yaml.gotmpl
@@ -169,12 +169,6 @@ brig:
         smsActivationUrl: https://example.com/v/${code}
         passwordResetUrl: https://example.com/reset/?key=${key}&code=${code}
         deletionUrl: https://example.com/d/?key=${key}&code=${code}
-      provider:
-        homeUrl: https://example.com/
-        providerActivationUrl: https://example.com/verify/bot/?key=${key}&code=${code}
-        approvalUrl: https://example.com/provider/approve?key=${key}&code=${code}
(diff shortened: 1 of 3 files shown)


libs/wire-subsystems/src/Wire/EmailSubsystem/Template.hs
```diff
@@ -72,157 +71,28 @@ renderTextWithBranding tpl replace branding = render tpl (replace . branding)
...
-    memberWelcomeEmail :: !MemberWelcomeEmailTemplate,
-    newTeamOwnerWelcomeEmail :: !NewTeamOwnerWelcomeEmailTemplate
-  }
+renderHtmlWithBrandingSem :: (Member (Output Text) r) => Template -> Map Text Text -> Sem r Lazy.Text
+renderHtmlWithBrandingSem = renderWithBrandingSem HTML.text
+
+renderTextWithBrandingSem :: (Member (Output Text) r) => Template -> Map Text Text -> Sem r Lazy.Text
+renderTextWithBrandingSem = renderWithBrandingSem id
+
+renderWithBrandingSem :: (Member (Output Text) r) => (Text -> Text) -> Template -> Map Text Text -> Sem r Lazy.Text
```
fisx, line 81:
```suggestion
-- If a template field is not declared, do not replace it and drop key to `Output` effect.  This way we catch all errors, not just the first, and the caller gets to decide what to do with the error.
renderWithBrandingSem :: (Member (Output Text) r) => (Text -> Text) -> Template -> Map Text Text -> Sem r Lazy.Text
```

PR #4714, opened 2025-08-13
WPB-19257: Add member count to user group list

blackheaven:
https://wearezeta.atlassian.net/browse/WPB-19257

## Checklist

### DIFF
--- a/libs/schema-profunctor/src/Data/Schema.hs
+++ b/libs/schema-profunctor/src/Data/Schema.hs
@@ -907,6 +907,12 @@ instance ToSchema Bool where schema = genericToSchema
 
 instance ToSchema Natural where schema = genericToSchema
 
+instance (ToSchema a) => ToSchema (Identity a) where schema = dimap runIdentity Identity schema
+
+instance (ToSchema a) => ToSchema (Const a x) where schema = dimap getConst Const schema
+
+instance (S.ToSchema a, A.ToJSON a, A.FromJSON a) => ToSchema (Maybe a) where schema = genericToSchema
+
 declareSwaggerSchema :: SchemaP (WithDeclare d) v w a b -> Declare d
 declareSwaggerSchema = runDeclare . schemaDoc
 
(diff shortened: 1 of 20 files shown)


libs/schema-profunctor/src/Data/Schema.hs
```diff
@@ -907,6 +907,12 @@ instance ToSchema Bool where schema = genericToSchema
 
 instance ToSchema Natural where schema = genericToSchema
 
+instance (ToSchema a) => ToSchema (Identity a) where schema = dimap runIdentity Identity schema
+
+instance (ToSchema a) => ToSchema (Const a x) where schema = dimap getConst Const schema
+
+instance (S.ToSchema a, A.ToJSON a, A.FromJSON a) => ToSchema (Maybe a) where schema = genericToSchema
+
```
battermann: I am not sure how that behaves.
If we want to be consistent with the rest of our code base, we can just define the part in the ToSchema instance as follows:
```
<*> (.membersCount) .= maybe_ (optField "membersCount" schema)
```

battermann (approved):
I approve already, but please consider addressing my comment

libs/schema-profunctor/src/Data/Schema.hs
blackheaven: It's not needed anymore, dropping it

libs/wire-api/src/Wire/API/UserGroup.hs
```diff
@@ -130,6 +132,7 @@ instance ToSchema (UserGroup_ (Const ())) where
         <$> (.id_) .= field "id" schema
         <*> (.name) .= field "name" schema
         <*> (.members) .= pure mempty
+        <*> (.membersCount) .= field "membersCount" schema
```
battermann: shouldn't this be `maybe_ (optField ...`

battermann: I think this would be good so that the field is omitted when the flag is not provided

blackheaven: You're right, thanks

PR #4719, opened 2025-08-15
WPD-19575: Drop Cryptobox

blackheaven:
https://wearezeta.atlassian.net/browse/WPB-19575

## Checklist


libs/wire-api/src/Wire/API/User/Client/Prekey.hs
```diff
@@ -89,6 +97,93 @@ clientIdFromPrekey =
...
+-- * **Root Object**: CBOR Map (Major Type 5).
+-- * **Key 0** (@Unsigned Int@): Protocol Version or Algorithm Identifier.
+-- * **Key 1** (@Unsigned Int@): **The Prekey ID** (Target of extraction).
+-- * **Key 2** (@Map@): The raw Public Key material (often a nested COSE-like structure).
+--
+-- = References
+--
+-- * <https://www.rfc-editor.org/rfc/rfc8949.html RFC 8949>: Concise Binary Object Representation (CBOR)
+-- * <https://www.rfc-editor.org/rfc/rfc9052.html RFC 9052>: CBOR Object Signing and Encryption (COSE) - /Contextual/
+--
```
akshaymankar: Instead of this comment, can we please write  a type and parse things into it? Then the comment could just be a link to the spec.

This would also make the code much simpler because we're now parsing into a generic CBOR term.

libs/wire-api/src/Wire/API/User/Client/Prekey.hs
```diff
@@ -89,6 +97,93 @@ clientIdFromPrekey =
...
+-- == Errors
+--
+-- This function returns a 'Left String' in the following cases:
+--
+-- 1. The input is not valid base64 data.
+-- 2. The input is not valid CBOR data.
+-- 3. The root object is not a Map.
+-- 4. The key @1@ is missing from the Map.
+-- 5. The value associated with key @1@ is not an integer.
```
akshaymankar: This is gonna go stale as soon as someone touches it. If we really want this, we should create an error type. Otherwise, I'm also ok with just deleting this part.

libs/wire-api/src/Wire/API/User/Client/Prekey.hs
```diff
@@ -89,6 +100,95 @@ clientIdFromPrekey =
...
+-- * **Key 3** (@Map@): **Signed Prekey Public Key**.
+--   A medium-term key (e.g., Map {0: {0: Bytes(32)}}).
+-- * **Key 4** (@Null@): **One-Time Prekey**.
+--   This field is Null if the server is out of one-time keys for this user.
+--
+-- = References
+--
+-- * <https://www.rfc-editor.org/rfc/rfc8949.html RFC 8949>: Concise Binary Object Representation (CBOR)
+-- * <https://www.rfc-editor.org/rfc/rfc7748.html RFC 7748>: Elliptic Curves for Security (X25519)
+-- * <https://www.rfc-editor.org/rfc/rfc9052.html RFC 9052>: CBOR Object Signing and Encryption (COSE)
```
akshaymankar: I'm not sure why we're using COSE here. The Prekey decoder in the rust proteus library only has 3 things: https://github.com/wireapp/proteus/blob/b92dbc2d0c77105cae3911a7388acba05450a06d/src/internal/keys.rs#L204-L222

akshaymankar (approved):

PR #4728, opened 2025-08-25
WPB-19693 fix role filter in team search endpoint

battermann:
## Checklist

### DIFF
--- a/integration/test/Test/Search.hs
+++ b/integration/test/Test/Search.hs
@@ -243,8 +243,7 @@ testFederatedUserSearchForNonTeamUser = do
 
 testSearchTeam :: (HasCallStack) => App ()
 testSearchTeam = do
-  let n = 5
-  (owner, tid, m1 : m2 : m3 : m4 : _) <- createTeam OwnDomain n
+  (owner, tid, m1 : m2 : m3 : m4 : _) <- createTeam OwnDomain 5
   updateTeamMember tid owner m1 Owner >>= assertSuccess
   updateTeamMember tid owner m2 Member >>= assertSuccess
   updateTeamMember tid owner m3 Partner >>= assertSuccess
@@ -253,6 +252,10 @@ testSearchTeam = do
   bindResponse (BrigP.searchTeamAll owner) $ \resp -> do
     resp.status `shouldMatchInt` 200
     docs <- resp.json %. "documents" >>= asList
-    length docs `shouldMatchInt` n
+    length docs `shouldMatchInt` 5
   bindResponse (BrigP.searchTeam owner [("frole", "member")]) $ \resp -> do
-    printJSON resp.json
+    resp.status `shouldMatchInt` 200
+    docs <- resp.json %. "documents" >>= asList
+    for_ docs $ \doc -> do
+      doc %. "role" `shouldMatch` "member"
+    length docs `shouldMatchInt` 1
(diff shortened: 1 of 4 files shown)


eyeinsky:
This looks good to me! The tests are also easy to read for someone new to the codebase. :)

Just two questions to improve my understanding:
- was the bug in the `instance FromHttpApiData Role` than needed an explicit mapping?
- is the search index update related to the bug fix? I guess if it is then the index is updated before the search happens, but I can't find this relation in the code, I guess it was already present before the changes in this PR?

eyeinsky:
Also, I guess the https://wearezeta.atlassian.net/browse/WPB-11124 ticket can be closed now.

battermann:
> Also, I guess the https://wearezeta.atlassian.net/browse/WPB-11124 ticket can be closed now.

yes when this is merged

battermann:
> This looks good to me! The tests are also easy to read for someone new to the codebase. :)
> 
> Just two questions to improve my understanding:
> 
> * was the bug in the `instance FromHttpApiData Role` than needed an explicit mapping?
> * is the search index update related to the bug fix? I guess if it is then the index is updated before the search happens, but I can't find this relation in the code, I guess it was already present before the changes in this PR?

Actually, it is/was all:
- the parser didn't work
- the field in the index was neither set nor updated
- the role filter was ignored when doing queries.

integration/test/Test/Search.hs
```diff
@@ -237,3 +237,63 @@ testFederatedUserSearchForNonTeamUser = do
...
+    for_ docs $ \doc -> do
+      uid <- doc %. "id" & asString
+      role <- doc %. "role" & asString
+      case uid of
+        _ | uid == ownerId -> role `shouldMatch` "owner"
+        _ | uid == m1Id -> role `shouldMatch` "owner"
+        _ | uid == m2Id -> role `shouldMatch` "member"
+        _ | uid == m3Id -> role `shouldMatch` "partner"
+        _ | uid == m4Id -> role `shouldMatch` "admin"
+        _ -> assertFailure $ "Unexpected user id in search results: " <> uid
```
akshaymankar: We'll get better errors on failure if we use `shouldMatchSet`.

integration/test/Test/Search.hs
```diff
@@ -237,3 +237,63 @@ testFederatedUserSearchForNonTeamUser = do
...
+      ("member", [m2Id]),
+      ("partner", [m3Id])
+    ]
+    $ \(role, expectedIds) -> do
+      bindResponse (BrigP.searchTeam owner [("frole", role)]) $ \resp -> do
+        resp.status `shouldMatchInt` 200
+        docs <- resp.json %. "documents" >>= asList
+        length docs `shouldMatchInt` (length expectedIds)
+        for_ docs $ \doc -> do
+          doc %. "role" `shouldMatch` role
```
akshaymankar: We could use `expectedIds`, add the role to each of them and compare the lists using `shouldMatchSet`.

libs/wire-api/src/Wire/API/Routes/Internal/Brig/SearchIndex.hs
```diff
@@ -30,3 +31,11 @@ type ISearchIndexAPI =
...
         :> Post '[JSON] NoContent
     )
+    :<|> Named
+           "update-search-index"
+           ( Summary "force update of search index for a single user"
+               :> "index"
+               :> "update"
+               :> Capture "userId" UserId
+               :> Post '[JSON] NoContent
+           )
```
akshaymankar: Instead of force updating, we should use the write time of the role to decide the version of the doc.

akshaymankar (changes requested):
I think force updating on normal updates is not good when we can compute the write time of the role of the user.

PR #4730, opened 2025-08-25
charts/gundeck: add prestop hook

jschaul:
We currently get a few 500s sometime when gundeck restarts, where some current requests seem to get aborted mid-request. This is possibly due to terminating pods still getting some traffic.

https://wearezeta.atlassian.net/browse/WPB-19694

## Checklist

### DIFF
--- a/charts/gundeck/templates/deployment.yaml
+++ b/charts/gundeck/templates/deployment.yaml
@@ -27,6 +27,7 @@ spec:
         checksum/configmap: {{ include (print .Template.BasePath "/configmap.yaml") . | sha256sum }}
         checksum/secret: {{ include (print .Template.BasePath "/secret.yaml") . | sha256sum }}
     spec:
+      terminationGracePeriodSeconds: {{ .Values.terminationGracePeriodSeconds }}
       serviceAccountName: {{ .Values.serviceAccount.name }}
       topologySpreadConstraints:
         - maxSkew: 1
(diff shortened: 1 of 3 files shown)


charts/gundeck/templates/deployment.yaml
```diff
@@ -171,5 +172,9 @@ spec:
               scheme: HTTP
               path: /i/status
               port: {{ .Values.service.internalPort }}
+          lifecycle:
+            preStop:
+              exec:
+                command: ["sh", "-c", "sleep 10"]
```
supersven, line 178: Unfortunately, I'm not an K8s expert: Could you please explain why we need the `sleep` command? Why isn't it good enough to only increase `terminationGracePeriodSeconds`? 🤔

lwille: the default `terminationGracePeriodSeconds` is 30 seconds, which is plenty.

What this sleep would do is delaying the time between the pod entering the `Terminating` state, and thus being removed from any `Service` that sends traffic there.
However, `gundeck` will still accept any in-flight requests and process them, since it doesn't know about being terminated.

Shamelessly adapted from [SO](https://stackoverflow.com/a/76991296/204581) to visualize and explain the process:

> The full sequence is :
> 1. pod deletion is requested (state: `Terminating`)
> 2. preStop hook kicks in and `terminationGracePeriodSeconds` countdown starts :
>    * when `preStop` hook completes, `kubelet` sends a `SIGTERM` to the container
>    * if `preStop` hook isn't finished within `terminationGracePeriodSeconds` countdown, `kubelet` sends SIGKILL to the container
image

→ this means: 
* we delay the shutdown (`SIGTERM`) to gundeck to allow for the k8s API to remove it from its `Service` and process
* once the 10s counter is finished, the regular `SIGTERM` signal is sent, causing `gundeck` to gracefully shutdown
* if the whole thing takes longer than `terminationGracePeriodSeconds`, it's simply killed.

PR #4736, opened 2025-08-28
[DO NOT MERGE] test for reconnection to RabbitMQ

battermann:
## Checklist

### DIFF
--- a/deploy/dockerephemeral/docker-compose.yaml
+++ b/deploy/dockerephemeral/docker-compose.yaml
@@ -347,7 +347,7 @@ services:
       - "5672:5672"
       - "15671:15671"
       - "15672:15672"
-      - "8080:8080"
+      - "8081:8080"
     volumes:
       - ./haproxy.cfg:/usr/local/etc/haproxy/haproxy.cfg:ro
       - ./haproxy-logs:/var/log/haproxy
--- a/integration/test/Testlib/Run.hs
+++ b/integration/test/Testlib/Run.hs
@@ -196,23 +196,24 @@ runMigrations = do
       void $ liftIO $ waitForProcess ph
 
 deleteFederationV0AndV1Queues :: GlobalEnv -> IO ()
-deleteFederationV0AndV1Queues env = do
-  let testDomains = env.gDomain1 : env.gDomain2 : env.gDynamicDomains
-  putStrLn "Attempting to delete federation V0 queues..."
-  (mV0User, mV0Pass) <- readCredsFromEnvWithSuffix "V0"
-  fromMaybe (putStrLn "No or incomplete credentials for fed V0 RabbitMQ") $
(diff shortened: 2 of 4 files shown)


jschaul:
1. run the docker-compose services, either on latest commit; or on a previous commit using toxiproxy
2. start the test: `TEST_INCLUDE=testRabbitMQConnection make ci-safe package=integration`
3. when the prompt appears, break the connection in some way. Use either `./toxiproxy-rabbitmq-terminate.sh` or a way with haproxy (you need to first fix the configuration, somehow the haproxy setup isn't quite working yet on that latest commit
4. re-establish a connection (press enter in the toxiproxy script)
5. press enter in the integration test so it attempts to send another message. If the test is green then, well, you could not reproduce an issue.

integration/test/Test/Demo.hs
```diff
@@ -215,3 +219,28 @@ testLegacyFedFederationV2 fedDomainV2 = do
...
+  withWebSockets [alice, bob] $ \wss@[_, bobWs] -> do
+    void $ createAddCommit alice1 convId [alice, bob] >>= sendAndConsumeCommitBundle
+    for_ wss (awaitMatch isMemberJoinNotif)
+
+    void $ createApplicationMessage convId alice1 "test" >>= sendAndConsumeMessage
+    void $ awaitMatch isNewMLSMessageNotif bobWs
+
+    putStrLn "simulate RabbitMQ outage"
+    putStrLn "e.g.: docker stop rabbitmq"
+    putStrLn "and: docker start rabbitmq"
```
lwille, line 240: In this test, we seem to be discretely doing things:

sending + consuming
(kill connection + wait for reconnect)
sending + consuming

I think that's not really what happens in the background worker, or in gundeck for forwarding notifications. Those processes would be waiting for new AMQP messages all the time, and suddenly the disconnect would kick.

What happens if the RabbitMQ connection is killed while sending/consuming messages?

battermann: It's correct that we test that sending and receiving works, then kill RabbitMQ, restart it and wait for reconnect and then test sending and receiving again.

But I don't understand what you mean with "that's not what happens in background worker". The background worker is always connected to the queue, also when the outage starts. Or more precisely, it's a thread in the background worker, that should constantly and indefinitely try to reconnect or restart if killed.

Gundeck in not involved as from the logs we can see that it's the notification-pusher that while trying to establish a connection to RabbitMQ.

Trying to kill RabbitMQ _while_ bg worker is consuming messages is an interesting idea. What we do is, we kill the broker, while the bg is connected, however there are no messages being processed when the connection is killed, because the queue is empty. Killing the broker while the queue contains unprocessed messages is technically a bit difficult because when we take RabbitMQ down we also cannot produce messages. Still maybe worthwhile to try to find a workaround and test this?

PR #4742, opened 2025-09-01
WPB-18932 do not send events generated by the self client to own client

battermann:
https://wearezeta.atlassian.net/browse/WPB-18932

## Checklist


battermann:
I don't think this is the right approach. If we do this, we should have something similar to the connection ID in the `Push` event, e.g. an optional `ClientId` that we then filter for on gundeck side.

PR #4743, opened 2025-09-02
WPB-19807 User Group and Channel API stubs

battermann:
https://wearezeta.atlassian.net/browse/WPB-19807

## Checklist


libs/wire-api/src/Wire/API/Conversation/Pagination.hs
```diff
@@ -0,0 +1,94 @@
...
+      & S.type_ ?~ S.OpenApiString
+      & S.enum_ ?~ ["name", "created_at"]
+
+------------------------------
+
+defaultSortOrder :: SortBy -> SortOrder
+defaultSortOrder SortByName = Asc
+defaultSortOrder SortByCreatedAt = Desc
+
+makePrisms ''ConversationPage
```
blackheaven: Should not it be `makeLenses`?

blackheaven (approved):

libs/wire-api/src/Wire/API/Conversation/Pagination.hs
battermann: we don't need it. removed

PR #4747, opened 2025-09-02
ignore hls.json

supersven:
This file is used by my (Sven) editor setup. It's probably not of
general usefulness.

## Checklist

 - [X] Read and follow the [PR guidelines](https://docs.wire.com/latest/developer/developer/pr-guidelines.html)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -105,3 +105,6 @@ logs-integration
 
 # BOM file - https://github.com/wireapp/tom-bombadil
 sbom.json
+
+# HLS config for haskell-tools plugin (Neovim)
+hls.json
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -105,3 +105,6 @@ logs-integration
 
 # BOM file - https://github.com/wireapp/tom-bombadil
 sbom.json
+
+# HLS config for haskell-tools plugin (Neovim)
+hls.json
```
fisx, line 110: can't you add this to your `~/.gitignore`, as something like `**/hls.json`?  not sure about the syntax, but i would expect that to work after a handful of tries?

but i'm also fine with adding it here.

supersven: Thanks for the hint :+1:

PR #4750, opened 2025-09-03
Use servant auth in spar

eyeinsky:
## Checklist


eyeinsky:
This is now ready for review. There are a few TODO comments that should be checked on what to do (and then removed).

integration/test/API/Spar.hs
```diff
@@ -143,3 +143,12 @@ finalizeSamlLoginWithZHost domain mbZHost tid (SAML.SignedAuthnResponse authnres
   baseRequest domain Spar Versioned (joinHttpPath ["sso", "finalize-login", tid])
     >>= formDataBody [partLBS (cs "SAMLResponse") . EL.encode . XML.renderLBS XML.def $ authnresp]
     >>= \req -> submit "POST" (req & maybe id zHost mbZHost)
+
+checkAdminGetTeamId :: (HasCallStack, MakesValue caller) => caller -> SAML.IdPMetadata -> App Response
```
fisx:
```suggestion
createIdP :: (HasCallStack, MakesValue caller) => caller -> SAML.IdPMetadata -> App Response
```

i don't insist on "createIdP", but the current name doesn't seem to fit.

also, i like to add links to (in this case) `-- | https://staging-nginz-https.zinfra.io/v12/api/swagger-ui/#/default/idp-create` to these functions.

integration/test/Test/Spar.hs
```diff
@@ -430,6 +432,17 @@ testSparCreateTwoScimTokensForOneIdp = do
   tokens <- getScimTokens owner >>= getJSON 200 >>= (%. "tokens") >>= asList
   length tokens `shouldMatchInt` 0
 
+testCheckAdminGetTeamId :: (HasCallStack) => App ()
+testCheckAdminGetTeamId = do
+  (owner :: Value, tid :: String, [admin, regular] :: [Value]) <- createTeam OwnDomain 3
+  updateTeamMember tid owner admin Admin >>= assertSuccess
```
fisx: are you doing this because the role of `owner` is `Owner`, not `Admin`?  the only distinction between roles admin and owner is that our billing system ibis (java) is more impressed by owners, so we usually use `owner` instead of creating a member with `Admin`.

libs/metrics-wai/src/Data/Metrics/Servant.hs
```diff
@@ -134,6 +134,12 @@ instance RoutesToPaths (NoContentVerb method) where
 instance RoutesToPaths (Stream method status framing ct a) where
   getRoutes = []
 
+instance
+  (RoutesToPaths rest) =>
+  RoutesToPaths (AuthProtect _tag :> rest)
+  where
+  getRoutes = getRoutes @rest -- TODO: I guess it's ok to leave out the auth protect here?
```
fisx: yes.  this is only about the HTTP request *path*.

libs/wire-api/src/Wire/API/Routes/Public.hs
```diff
@@ -259,6 +260,9 @@ instance (HasOpenApi api) => HasOpenApi (ZAuthServant 'ZAuthUser _opts :> api) w
 instance (HasOpenApi api) => HasOpenApi (ZAuthServant 'ZLocalAuthUser opts :> api) where
   toOpenApi _ = addZAuthSwagger (toOpenApi (Proxy @api))
 
+instance (HasOpenApi api) => HasOpenApi (AuthProtect _tag :> api) where
+  toOpenApi _ = addZAuthSwagger (toOpenApi (Proxy @api)) -- TODO: do we want to show something in Swagger here?
```
fisx: if the `_tag` argument contains enough information so we can write an instance for one tag that also explains the policy (and i think that's the case?), then this would be nice.  :+1:

libs/wire-subsystems/src/Wire/UserSubsystem/Interpreter.hs
```diff
@@ -1079,3 +1080,12 @@ removeEmailEitherImpl lusr = runError $ do
 
 getUserTeamImpl :: (Member UserStore r) => UserId -> Sem r (Maybe TeamId)
 getUserTeamImpl = UserStore.getUserTeam
+
+-- TODO: should we inline this action in the single location we use this?
```
fisx: i leaning towards "don't do anything, keep the polysemy action".  the alternative would be to call the function in `Spar.Intra` directly, and i think it'd be against the idea of the effect.

also with `BrigAccess`, it'll be easier to refactor this after brig and spar get merged.  and it's a pattern we use not only for brig, but also for galley and maybe others in the wire-subsystems library, and that is supposed to be idiomatic.

libs/wire-api/src/Wire/API/Routes/Public/Spar.hs
```diff
@@ -125,11 +128,13 @@ type APIAuthResp =
         :> Post '[PlainText] Void
     )
 
+type instance AuthServerData (AuthProtect "zoptuser") = (UserId, TeamId)
```
fisx: can you avoid the orphan if you do this near the definition of `UserId` etc.?

is that a good idea?  orphans suck, but i'm not sure types-common should depend on servant.  (or does it already?)

it would be extra-nice if we could add the swagger instances to `Data.Id` as well, then it'd be harder to forget to update them if anything changes.

services/spar/src/Spar/API.hs
```diff
@@ -442,6 +449,20 @@ ssoSettings =
 ----------------------------------------------------------------------------
 -- IdPConfigStore API
 
+type Auth = AuthHandler Request (UserId, TeamId)
```
fisx: maybe this is (slightly) more confusing rather than helpful?  if you have no strong opinion i'd remove this alias, or (second best option) change it to `type Auth = AuthHandler Request` only.

services/spar/src/Spar/API.hs
```diff
@@ -442,6 +449,20 @@ ssoSettings =
...
+
+authHandler :: Env -> Auth
+authHandler ctx = mkAuthHandler $ \req -> (either throwError' pure =<<) $ runSparToHandler ctx $ runError $ do
+  bs <- maybe (throwSparSem $ SparNoPermission "No Z-User header") pure $ lookup "Z-User" (requestHeaders req)
+  uid <-  maybe (throwSparSem $ SparNoPermission "Can't parse Z-User header") pure $ fromByteString bs
```
fisx:
```suggestion
  uid <-  maybe (throwSparSem $ SparNoPermission "[internal error] Can't parse Z-User header") pure $ fromByteString bs
```

services/spar/src/Spar/API.hs
```diff
@@ -634,16 +655,15 @@ idpCreateV7 ::
...
   IdPMetadataInfo ->
   Maybe SAML.IdPId ->
   Maybe WireIdPAPIVersion ->
   Maybe (Range 1 32 Text) ->
   Sem r IdP
-idpCreateV7 zusr idpmeta mReplaces mApiversion mHandle = do
-  teamid <- Brig.getZUsrCheckPerm zusr CreateUpdateDeleteIdp
-  assertNoScimOrNoIdP teamid
-  idpCreate zusr idpmeta mReplaces mApiversion mHandle
+idpCreateV7 (zusr, tid) idpmeta mReplaces mApiversion mHandle = do
```
fisx: don't forget to also change `idpCreate`, which is the most recent version.  it is also called from here, but it in the future `idpCreateV7` will go away.  (i think you don't need the `TeamId` in `idpCreate`, but you need to add the `Auth` combinator.

fisx (changes requested):
looks good, almost done!  :)

fisx:
(oh, and could you check the boxes in the PR description?)

integration/test/Test/Spar.hs
eyeinsky: I didn't know how special the owner is and had just found a way to create and admin from surrounding tests. But if owner pretty much is an admin then yes, I could skip creating one explicitly.

libs/wire-api/src/Wire/API/Routes/Public/Spar.hs
eyeinsky: Hmm, I think the orphan instance can't be avoided, since `AuthServerData` and `AuthProtect` are both defined elsewhere. The `Data.Id` module does have `-fno-warn-orphans` set though, so it could be removed from Wire.API.Routes.Public.Spar when moved there. If it weren't for the orphans flag I'd even keep the instance where it is as it feels like it's more part of the API rather than the `UserId` type, but needing to keep the orphans flag there too is an open door for more.. So maybe still move then?

libs/wire-api/src/Wire/API/Routes/Public.hs
eyeinsky: Now that I look at it again, the `addZAuthSwagger` helper already adds the authorization mechanism to this endpoint. I think this change doesn't change anything for the outside user, it's just that the authentication happens before parsing of the body.

I wonder if I should replace `ZOptUser` with `AuthProtect "zoptuser"` in all locations in this API part: [/libs/wire-api/src/Wire/API/Routes/Public/Spar.hs](https://github.com/wireapp/wire-server/blob/c41d8ee4c91f51057bc2e3458f4095719e030f75/libs/wire-api/src/Wire/API/Routes/Public/Spar.hs#L129-L135), or leave this off for another day? It seems that the new `authHandler` essentially does the same work as the [`authorizeIdP`](https://github.com/wireapp/wire-server/blob/c41d8ee4c91f51057bc2e3458f4095719e030f75/services/spar/src/Spar/API.hs#L870-L888) function in the Spar service to authenticate admins?

eyeinsky:
@fisx, I'm ready for another round of review!

libs/wire-api/src/Wire/API/Routes/Public.hs
fisx: i just realized that `"zoptuser"` may also benefit from getting changed.  It's about making sure the user exists and is admin of a team, correct?  them maybe `"IsTeamAdmin"`?

The it's possible to drop the `ZOptUser` in this sub-api entirely and replace it with `AuthProtect "..."`, but note that the policy is different for every end-point.  Maybe better leave that for another PR?

fisx: > I think this change doesn't change anything for the outside user, it's just that the authentication happens before parsing of the body.

true, but it's still nice to extend the swagger coverage :-)

but maybe we should save the time.

libs/wire-api/src/Wire/API/Routes/Public/Spar.hs
fisx: no, if moving it doesn't avoid orphans it think you picked the right place for it.

fisx:
the ci failure looks very much like a flake, i'll bump it.

libs/types-common/src/Data/Id.hs
```diff
@@ -169,6 +170,8 @@ type ServiceId = Id 'Service
 
 type TeamId = Id 'Team
 
+type instance AuthServerData (AuthProtect "zoptuser") = (UserId, TeamId)
```
fisx: ah, you already moved it.  either way is good!  :)

libs/wire-api/src/Wire/API/Routes/Public.hs
eyeinsky: > true, but it's still nice to extend the swagger coverage :-)

But what do you mean by the swagger coverage? Both the previous authentication and the current  one require a bearer token which is then resolved to a `Z-User` HTTP header (AFAIK) and that is what the `addZAuthSwagger` does. I.e, for the API user both the old and new combinator require the same process.

libs/types-common/src/Data/Id.hs
eyeinsky: I'll move it back, I like the other location better! (though we should definitely continue to review every added orphan)

services/spar/src/Spar/API.hs
```diff
@@ -606,14 +628,14 @@ idpCreate ::
     Member IdPRawMetadataStore r,
     Member (Error SparError) r
   ) =>
-  Maybe UserId ->
+  (UserId, TeamId) ->
```
blackheaven: nitpick: can you curry it? or is it costly?

integration/test/Test/Spar.hs
```diff
@@ -430,6 +431,23 @@ testSparCreateTwoScimTokensForOneIdp = do
...
+    -- insuficient permissions for non-admin, both as expected.
+    resp.status `shouldMatchInt` 403
+    resp.json %. "label" `shouldMatch` "insufficient-permissions"
+
+testCheckAdminGetTeamId :: App ()
+testCheckAdminGetTeamId = mkTestCheckAdminGetTeamId Versioned
+
+testCheckAdminGetTeamIdV7 :: App ()
+testCheckAdminGetTeamIdV7 = mkTestCheckAdminGetTeamId (ExplicitVersion 7)
+
```
akshaymankar: This could also be written using `withAPIVersion`, then we can avoid adding `Versioned` to the request helper object.

services/spar/src/Spar/API.hs
eyeinsky: Technically it can be, but it would move the ugliness into the API type: since this is the handler for an API endpoint then having two separate arguments for it would mean that the API type would also consist of two parts like so: `AuthProtect "UserId" :> AuthProtect "TeamId" :> ...` (while currently it is `AuthProtect "zoptuser"`).

libs/wire-api/src/Wire/API/Routes/Public/Spar.hs
```diff
@@ -125,12 +128,14 @@ type APIAuthResp =
...
```
akshaymankar:
```suggestion
type instance AuthServerData (AuthProtect "TeamAdmin") = (UserId, TeamId)
```

I think we should use the tag to clarify that this is a TeamAdmin.

services/spar/src/Spar/API.hs
```diff
@@ -634,16 +656,15 @@ idpCreateV7 ::
...
   IdPMetadataInfo ->
   Maybe SAML.IdPId ->
   Maybe WireIdPAPIVersion ->
   Maybe (Range 1 32 Text) ->
   Sem r IdP
-idpCreateV7 zusr idpmeta mReplaces mApiversion mHandle = do
-  teamid <- Brig.getZUsrCheckPerm zusr CreateUpdateDeleteIdp
```
akshaymankar, line 644: This looks like a more granular check than the one in the auth handler. We should perhaps keep this.

eyeinsky: The auth check eventually calls this https://github.com/wireapp/wire-server/blob/eda172e3894feb4476fe23d1d27c960648126fef/libs/wire-subsystems/src/Wire/UserSubsystem/Interpreter.hs#L1090, which checks for the same permission (unless I misunderstand the comment).

libs/wire-api/src/Wire/API/Routes/Public/Spar.hs
eyeinsky: Did it! eda172e3894feb4476fe23d1d27c960648126fef

services/spar/src/Spar/API.hs
```diff
@@ -606,14 +628,14 @@ idpCreate ::
...
+  (UserId, TeamId) ->
   IdPMetadataInfo ->
   Maybe SAML.IdPId ->
   Maybe WireIdPAPIVersion ->
   Maybe (Range 1 32 Text) ->
   Sem r IdP
-idpCreate zusr (IdPMetadataValue rawIdpMetadata idpmeta) mReplaces (fromMaybe defWireIdPAPIVersion -> apiversion) mHandle = withDebugLog "idpCreateXML" (Just . show . (^. SAML.idpId)) $ do
-  teamid <- Brig.getZUsrCheckPerm zusr CreateUpdateDeleteIdp
+idpCreate (zusr, _tid) (IdPMetadataValue rawIdpMetadata idpmeta) mReplaces (fromMaybe defWireIdPAPIVersion -> apiversion) mHandle = withDebugLog "idpCreateXML" (Just . show . (^. SAML.idpId)) $ do
+  teamid <- Brig.getZUsrCheckPerm (Just zusr) CreateUpdateDeleteIdp
```
fisx: oops, i missed that: you should probably use `_tid` here instead of pulling it from brig again :)

eyeinsky: Ah, yes, it's more like *I* missed it!

This also seems like it has a cascading effect of not needing the `zusr` part of the argument :thinking:

eyeinsky: So after using the provided `tid` 6861fd9853f97e1cde09d06573c4cfd793584c84 the zusr was unused and cascaded into this 7b35e4bb3fab10a467fef5bb1f3f900bec2d458a. Does this make sense?

fisx: yes, it does to me.

(i would have kept the userid and discarded it in the handler, but you're right, your code is better.)

libs/wire-api/src/Wire/API/Routes/Public.hs
fisx: i meant that it'd be nice to have the access control policy documented.  but perhaps this is overdoing it?  not sure.

eyeinsky:
@fisx With help from Leif we got the test that I had prior deleted added back and fixed. The test returned 500 "server-error" because there was a Cassandra query that needed an additional Maybe to account for the case where the user existed but it didn't have a team column. The test is fixed here 12567d25164f4b31c3a25d169818ea50acbacc03.

Now everything should be done in this PR. :)

PR #4761, opened 2025-09-08
Minor fixes

eyeinsky:
Unrelated things I found while doing the feature PR, please double check if all of these are actually correct.

### DIFF
--- a/integration/test/Testlib/Run.hs
+++ b/integration/test/Testlib/Run.hs
@@ -1,6 +1,5 @@
 module Testlib.Run (main, mainI) where
 
-
 import Control.Concurrent
 import Control.Exception as E
 import Control.Monad


Makefile
```diff
@@ -100,10 +100,9 @@ endif
 	./hack/bin/cabal-install-artefacts.sh $(package)
 
 # ci here doesn't refer to continuous integration, but to cabal-run-integration.sh
-# Usage: make ci                        - build & run all tests, excluding integration
-#        make ci package=all            - build & run all tests, including integration
-#        make ci package=brig           - build brig & run "brig-integration"
-#        make ci package=integration    - build & run "integration"
+# Usage: make ci-fast                   - run all integration tests
+#        make ci-fast package=brig      - run integration test for package
```
fisx: why did you remove `package=all` (genuinely interesting) and `package=integration` (kind of a special case because the ideal is that this is the only integration test suite we have (everything else is legacy)?

fisx (approved):
nice!

fisx:
(i gave you approval, but please consider re-adding the Makefile lines i commented about)

Makefile
eyeinsky: I was wrong, fixed it now. (Now the only change in this area is "ci" => "ci-fast")

supersven (approved):
LGTM 👍 

Thanks a lot for cleaning thing up on the go ❤️

supersven:
PS: I've restarted the integration tests. Looks like a flake failed.

PR #4763, opened 2025-09-08
Sventennie/simplified user connecting qr code feature flag

supersven:
Add a feature flag for "simplified user connection requests" QR codes. This flag will only be interpreted by clients; backend behaviour does not change.

If you're interested, the QR code feature is described here: https://wearezeta.atlassian.net/wiki/spaces/PAD/pages/1289683040/QR+Profiles+-+Simplifying+connection+requests (I completely forgot that it existed :sweat_smile: )

Ticket: https://wearezeta.atlassian.net/browse/WPB-19751

## Checklist

 - [X] Read and follow the [PR guidelines](https://docs.wire.com/latest/developer/developer/pr-guidelines.html)


pcapriotti (approved):
LGTM

